PR #9410, opened 2021-09-09
Reset metrics before each task starts

### LINKED ISSUE #7558: `callback_metrics` should be reset when a new epoch start
## 🐛 Bug

Callback metrics should be destroyed whenever a new epoch/loop starts.

### To Reproduce

```python
def test_bug(tmpdir):
    from tests.helpers.boring_model import BoringModel

    class TestModel(BoringModel):

            assert not self.trainer.callback_metrics

            output = self.layer(batch)
            loss = self.loss(batch, output)
            self.log('train_loss', loss)
            return loss

            output = self.layer(batch)
            loss = self.loss(batch, output)
            self.log('val_loss', loss)

    model = TestModel()
    model.validation_epoch_end = None
    trainer = pl.Trainer(
        default_root_dir=tmpdir,
        max_epochs=3,
        num_sanity_val_steps=0,
        callbacks=[pl.callbacks.ModelCheckpoint(monitor='val_loss')]
    )

    trainer.validate(model)
    print(trainer.callback_metrics)

    model.validation_step = None
    trainer.fit(model)
```

### Expected behavior

The above case should throw an error because during `.fit` validation is disabled but since `trainer.validate` runs before it was disabled, the callback_metrics calculated during that loop stays and doesn't cause a `MisconfigurationException` during checkpointing. This is just one case where wrong metrics stored can cause some issues.

Another one can occur if `check_val_every_n_epoch>1` where if validation is not happening in every epoch, it still contains metrics from some old epoch which can cause some issues with `ModelCheckpoint/EarlyStopping/ReduceLROnPlateau`.

### Environment

 - PyTorch Version (e.g., 1.0): 1.8.1+cu101
 - OS (e.g., Linux): Linux
 - Build command you used (if compiling from source):
 - Python version: 3.7
 - CUDA/cuDNN version: 10.1
 - GPU models and configuration: Tesla T4

rohitgr7:
## What does this PR do?

Fixes #7558

## Before submitting

## PR review

## Did you have fun?


tchaton (approved):
LGMT !

awaelchli (commented):
Given the description of the PR and looking at the implementation, I think it's better these calls go directly into the trainer.run implementation, not the loops.

rohitgr7:
> Given the description of the PR and looking at the implementation, I think it's better these calls go directly into the trainer.run implementation, not the loops.

good point, thanks :)

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -278,14 +278,20 @@ def should_reset_tensors(self, fx: str) -> bool:
...
+        if trainer_metrics:
             # reset metrics
-            self._progress_bar_metrics = {}
-            self._logged_metrics = {}
-            self._callback_metrics = {}
-        assert self.trainer._results is not None
-        self.trainer._results.reset(metrics=metrics)
+            self.reset_metrics()
+
+        if self.trainer.state.fn != TrainerFn.PREDICTING:
```
awaelchli, line 291: can you explain this check?

carmocca: I can, the prediction loops do not define `_results` (they are `None`)

But i'd change the code to:


```python
if self.trainer._results is not None:
    self.trainer._results.reset(metrics=metrics)
```

Instead of the predict check and the assertion

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -278,14 +278,20 @@ def should_reset_tensors(self, fx: str) -> bool:
             is_first_batch = bool(self._batch_idx) + self._split_idx == 0
         return is_different_fx and is_first_batch
 
-    def reset(self, metrics: Optional[bool] = None) -> None:
-        if self.trainer.sanity_checking:
+    def reset_metrics(self) -> None:
```
carmocca, line 281: Do not extract this if it's not going to be called from the outside.

I think it'd be better having two functions, this one and `def reset_results(metrics)`

to avoid the `trainer_metrics` flag

pytorch_lightning/trainer/trainer.py
```diff
@@ -989,6 +989,9 @@ def _run(self, model: "pl.LightningModule") -> Optional[Union[_EVALUATE_OUTPUT,
         # ----------------------------
         # TRAIN
         # ----------------------------
+        # reset logger connector
```
carmocca, line 992: It is easier to put these reset calls directly in the trainer but IMO they should be managed by each loop, even if it's more verbose.

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: Yeah this is better, Will update this.. thanks @carmocca

pytorch_lightning/trainer/trainer.py
rohitgr7: @awaelchli suggested this. It was in the loop before. Any reason why they should be inside loop?

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: I think it can be called from outside, let's say someone wants to reset their metrics after every epoch, then they can just use this in one of their hooks. Right now it resets them at every run.

pytorch_lightning/trainer/trainer.py
carmocca: because the loops are the ones to control the flow of training, and they should be responsible for interacting with the logger connector, not the trainer.
Otherwise, the customization is limited.

For example, compare the existing usage of the `logger_connector` in the trainer vs the loops.

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
carmocca: That's why I suggest keeping it public but removing the `trainer_metrics` flag

pytorch_lightning/trainer/trainer.py
awaelchli: well here is some advice 😅 
in most cases where @carmocca's arguments stand against @awaelchli's arguments, @carmocca wins. So when @awaelchli suggests changes better wait until @carmocca reviews before you make the changes 😄

rohitgr7:
due to recent fault-tolerant changes, I had to move it back to the trainer. Since `logger_connector` is sort of handled by the trainer, this seems ok to me.

rohitgr7:
> > due to recent fault-tolerant changes, I had to move it back to the trainer
> 
> Can you elaborate on what are the changes and why they force moving it?

Training state is being restored before dispatch so adding these resets inside loops will reset the resultcollection set by resume_from_checkpoint.

CHANGELOG.md
```diff
@@ -352,6 +352,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `add_argparse_args` raising `TypeError` when args are typed as `typing.Generic` in Python 3.6 ([#9554](https://github.com/PyTorchLightning/pytorch-lightning/pull/9554))
 
 
+- Fixed error reporting in DDP process reconciliation when processes are launched by an external agent ([#9389](https://github.com/PyTorchLightning/pytorch-lightning/pull/9389))
```
awaelchli, line 355: don't forget to resolve the bad merge here :)

pytorch_lightning/loops/dataloader/prediction_loop.py
```diff
@@ -78,7 +78,7 @@ def reset(self) -> None:
         self.epoch_batch_indices = []
 
     def on_run_start(self) -> None:
-        """Calls ``on_predict_start`` hook."""
+        """Calls ``_on_predict_start`` hook."""
```
awaelchli, line 81: here too :)

awaelchli (approved):

Borda (approved):

PR #9413, opened 2021-09-09
Fix restoring training state during `trainer.fit` only

### LINKED ISSUE #9404: [checkpoint] Restoring training state doesn't check whether we're fitting
### Motivation

Optimizer & LR scheduler states only need to be restored during fitting. However, these are currently called unconditionally here in the checkpoint connector: 

https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L160-L161

https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L165

https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L1010-L1011

Which means we attempt to restore these states during any of `validate`/`test`/`predict`

### Pitch

Check if the trainer is fitting, and only if so restore the precision, optimizer, and LR scheduler states

### Additional context

<sub>

</sub>

rohitgr7:
## What does this PR do?

Fixes #9404
TODO: need to merge/add similar tests for datamodule and callbacks in another PR.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loops/base.py
```diff
@@ -198,16 +198,17 @@ def state_dict(self, destination: Optional[Dict] = None, prefix: Optional[str] =
 
     def load_state_dict(
         self,
-        state_dict: Dict,
+        state_dict: Optional[Dict] = None,
```
ananthsub, line 201: why does this need to be optional?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -187,40 +191,49 @@ def restore_loops(self) -> None:
...
-        expected_steps = self.trainer.num_training_batches / n_accum
-        if self.trainer.num_training_batches != 0 and self.trainer.global_step % expected_steps > 1:
-            rank_zero_warn(
-                "You're resuming from a checkpoint that ended mid-epoch."
-                " Training will start from the beginning of the next epoch."
-                " This can cause unreliable results if further training is done,"
-                " consider using an end of epoch checkpoint."
+        if self.trainer.state.fn == TrainerFn.FITTING:
+            self.trainer.fit_loop.global_step = self._loaded_checkpoint["global_step"]
+            self.trainer.fit_loop.current_epoch = self._loaded_checkpoint["epoch"]
```
ananthsub, line 196: @carmocca i'm not sure what the expectation is around resuming these properties from checkpoint, given `global_step` has implications on logging in every trainer function

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -187,40 +191,49 @@ def restore_loops(self) -> None:
...
+                    f" but you have set Trainer(max_epochs={self.trainer.max_epochs})."
+                )
+
+            # Division deals with global step stepping once per accumulated batch
+            # Inequality deals with different global step for odd vs even num_training_batches
+            self.trainer.accumulate_grad_batches = self.trainer.accumulation_scheduler.get_accumulate_grad_batches(
+                self.trainer.current_epoch
             )
+            n_accum = 1 if self.trainer.accumulate_grad_batches is None else self.trainer.accumulate_grad_batches
+            expected_steps = self.trainer.num_training_batches / n_accum
```
ananthsub, line 214: we should add a TODO here: `trainer.accumulate_grad_batches` can be a dict which will fail the division

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -187,40 +191,49 @@ def restore_loops(self) -> None:
...
         state_dict = self._loaded_checkpoint.get("loops")
-        if state_dict:
-            self.trainer.fit_loop.load_state_dict(state_dict["fit_loop"])
-            self.trainer.validate_loop.load_state_dict(state_dict["validate_loop"])
-            self.trainer.test_loop.load_state_dict(state_dict["test_loop"])
-            self.trainer.predict_loop.load_state_dict(state_dict["predict_loop"])
+
+        if state_dict is not None:
+            if self.trainer.state.fn == TrainerFn.FITTING:
+                self.trainer.fit_loop.load_state_dict(state_dict.get("fit_loop"))
```
ananthsub, line 227: not sure why we're using `state_dict.get(...)` here instead of direct index access. this would avoid the optional typing in the loops load state dict. shouldn't it be enforced that the state dict has a fit_loop entry if we're resuming for fitting?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1095,7 +1095,8 @@ def _pre_training_routine(self):
         # register auto-resubmit when on SLURM
         self.slurm_connector.register_slurm_signal_handlers()
 
-        self.checkpoint_connector.resume_end()
+        if self.state.fn == TrainerFn.FITTING:
+            self.checkpoint_connector.resume_end()
```
ananthsub, line 1099: why do we need this check on fitting for resume end?

tests/trainer/test_trainer.py
```diff
@@ -460,6 +510,54 @@ def on_load_checkpoint(self, _):
         assert next_model.num_on_load_checkpoint_called == 1
 
 
+def test_trainer_properties_resume_from_checkpoint(tmpdir):
```
ananthsub, line 513: is it possible for us to test this without relying on a callback?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: yep.. actually wanted to confirm this. Should it be enforced or not? Ideally, it should be but just wanted other's thoughts on this.

pytorch_lightning/trainer/trainer.py
rohitgr7: because `trainer.tune` calls it too. so just don't want it to delete the reloaded_state_dict in case of tuning. Also, `resume_start` is under the same condition as well.

pytorch_lightning/loops/base.py
rohitgr7: let's see if the below enforcement is required or not. Will update this, once that gets finalized.

tests/trainer/test_trainer.py
rohitgr7: yes, I can put all of these inside the lightning module's hooks, but just wanted to keep them separate just like it's done in other tests inside `models/test_restore`.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: okay.. not familiar yet with how dict work here. Is it like after nth epoch(key) start accumulating w.r.t to value?? and we can have a simple dict with different keys representing nth epoch and their corresponding accumulate_grad_batches???

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: Yes, I believe we should always reload the global_step, current_epoch. Users might have some logic within validation based on those.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -187,40 +191,49 @@ def restore_loops(self) -> None:
...
+                )
 
         state_dict = self._loaded_checkpoint.get("loops")
-        if state_dict:
-            self.trainer.fit_loop.load_state_dict(state_dict["fit_loop"])
-            self.trainer.validate_loop.load_state_dict(state_dict["validate_loop"])
-            self.trainer.test_loop.load_state_dict(state_dict["test_loop"])
-            self.trainer.predict_loop.load_state_dict(state_dict["predict_loop"])
+
+        if state_dict is not None:
```
tchaton, line 225: Neat !

tests/trainer/test_trainer.py
```diff
@@ -460,6 +510,54 @@ def on_load_checkpoint(self, _):
...
+
+    model = CustomBoringModel()
+    checkpoint_callback = ModelCheckpoint(dirpath=tmpdir, monitor="val_loss", save_last=True)
+    trainer_args = dict(
+        default_root_dir=tmpdir,
+        max_epochs=1,
+        limit_train_batches=7,
+        limit_val_batches=7,
+        logger=False,
+        callbacks=[checkpoint_callback],  # this performs the assertions
```
carmocca, line 535: the `ModelCheckpoint` performs the assertions?

tests/trainer/test_trainer.py
carmocca: Why not move it to `test_restore`?

pytorch_lightning/trainer/trainer.py
ananthsub: should we explicitly call this with tune in mind then? 
```suggestion
        if self.state.fn != TrainerFn.TUNING:
            self.checkpoint_connector.resume_end()
```

tests/trainer/test_trainer.py
rohitgr7: Because that one is under `models/` and this one is more trainer specific.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
ananthsub: > trainer.accumulate_grad_batches is set using accumulation_scheduler right above so it's always an int.

i don't think that's true: https://github.com/PyTorchLightning/pytorch-lightning/blob/181433458114523e83e079f6367c43e8f569ee5b/pytorch_lightning/trainer/connectors/training_trick_connector.py#L48

but let's discuss this in a separate issue/PR

tests/trainer/test_trainer.py
rohitgr7: Copy paste 😅.. will update

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: okay, looks like during init it isn't but during restore it is:
https://github.com/PyTorchLightning/pytorch-lightning/blob/181433458114523e83e079f6367c43e8f569ee5b/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L209-L213

pytorch_lightning/trainer/trainer.py
rohitgr7: will update.

rohitgr7:
Does anyone know how to fix the mypy issue here?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -169,12 +170,15 @@ def restore_training_state(self) -> None:
         if not self._loaded_checkpoint:
             return
 
-        # restore precision plugin (scaler etc.)
-        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
         # restore loops and their progress
         self.restore_loops()
 
-        self.restore_optimizers_and_schedulers()
+        if self.trainer.state.fn == TrainerFn.FITTING:
```
carmocca, line 176: you'll have to `assert self.trainer.state.fn is not None` before this equality check to fix mypy

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -169,12 +170,15 @@ def restore_training_state(self) -> None:
...
+            # restore precision plugin (scaler etc.)
+            self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
```
carmocca, line 178: Isn't it possible that we want to load this state for evaluation? precision can be enabled for them

pytorch_lightning/trainer/connectors/checkpoint_connector.py
carmocca: @rohitgr7 this was marked as resolved but I think it needs to stay

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -191,41 +195,50 @@ def restore_loops(self) -> None:
...
-            self.trainer.fit_loop.load_state_dict(state_dict["fit_loop"])
-            self.trainer.validate_loop.load_state_dict(state_dict["validate_loop"])
-            self.trainer.test_loop.load_state_dict(state_dict["test_loop"])
-            self.trainer.predict_loop.load_state_dict(state_dict["predict_loop"])
+
+        if state_dict is not None:
+            if self.trainer.state.fn == TrainerFn.FITTING:
+                self.trainer.fit_loop.load_state_dict(state_dict["fit_loop"])
+
+            if self.trainer.state.fn == TrainerFn.VALIDATING:
```
carmocca, line 234: This means that before you would only need to load once, but now you need to do it per trainer fn if you are using a checkpoint with existing data per loop.

Not sure if this is a good thing.

Can you add a test?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: yes, my bad.. forgot to commit the recent changes. updated

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: I think we have the reload for val_loop inside fitting too since evaluation is also a part for `trainer.fit`. For others, I guess it's fine since others are isolated, and no matter what this will be called everytime with `trainer.{test/validate/predict}`

rohitgr7: even before, it was called every time you call `trainer.{fit/test/validate/predict}`, but now let's say if `trainer.test` is not called then there is no point in loading the `test_loop` state. I am not sure about `val_loop`, whether to load it or not inside FITTING, since for a normal case, evaluation with `trainer.fit` goes over a complete epoch from scratch. do you think loading `val_loop` state dict inside FITTING makes sense?>

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: hmm. good point. done. thanks

pytorch_lightning/trainer/trainer.py
```diff
@@ -913,12 +913,13 @@ def tune(
...
         if self.state.fn == TrainerFn.FITTING:
-            self.checkpoint_connector.resume_start()
-        self.checkpoint_connector.restore_datamodule()
-        self.checkpoint_connector.restore_model()
-        # restore callback states
-        self.checkpoint_connector.restore_callbacks()
+            self.checkpoint_connector.restore_datamodule()
+            self.checkpoint_connector.restore_model()
+            # restore callback states
+            self.checkpoint_connector.restore_callbacks()
```
rohitgr7, line 922: FYI: this is required to at least load the resume_from_checkpoint and assign global_step and epoch during val/test as suggested by Thomas above. `reload_datamodule` and `restore_callbacks` are now reloaded during `.fit`. Please let me know whether it's the intended behavior or not.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -194,38 +199,49 @@ def restore_loops(self) -> None:
...
+                )
 
         state_dict = self._loaded_checkpoint.get("loops")
-        if state_dict:
-            self.trainer.fit_loop.load_state_dict(state_dict["fit_loop"])
-            self.trainer.validate_loop.load_state_dict(state_dict["validate_loop"])
-            self.trainer.test_loop.load_state_dict(state_dict["test_loop"])
-            self.trainer.predict_loop.load_state_dict(state_dict["predict_loop"])
+
+        if state_dict is not None:
```
tchaton, line 233: Let's move this at the top.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -194,38 +199,49 @@ def restore_loops(self) -> None:
...
-        n_accum = 1 if self.trainer.accumulate_grad_batches is None else self.trainer.accumulate_grad_batches
-        expected_steps = self.trainer.num_training_batches / n_accum
-        if self.trainer.num_training_batches != 0 and self.trainer.global_step % expected_steps > 1:
-            rank_zero_warn(
-                "You're resuming from a checkpoint that ended mid-epoch."
-                " Training will start from the beginning of the next epoch."
-                " This can cause unreliable results if further training is done,"
-                " consider using an end of epoch checkpoint."
+        assert self.trainer.state.fn is not None
+        if self.trainer.state.fn == TrainerFn.FITTING:
```
tchaton, line 203:
```suggestion
        if self.trainer.state.fn != TrainerFn.FITTING:
             return 
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -913,12 +913,13 @@ def tune(
 
     def _restore_modules_and_callbacks(self) -> None:
         # restore modules after setup
+        self.checkpoint_connector.resume_start()
+
         if self.state.fn == TrainerFn.FITTING:
```
tchaton, line 918:
```suggestion
        if self.state.fn != TrainerFn.FITTING:
             return 
```

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -986,8 +986,8 @@ def assert_checkpoint_log_dir(idx):
 
         trainer.test(model)
         # resume_from_checkpoint is resumed when calling `.fit`
-        assert trainer.global_step == 0
-        assert trainer.current_epoch == 0
+        assert trainer.global_step == 4
```
tchaton, line 989: Any idea why this is different ?

rohitgr7: https://github.com/PyTorchLightning/pytorch-lightning/pull/9413#discussion_r707093777
as you suggested.

CHANGELOG.md
```diff
@@ -385,9 +385,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed error reporting in DDP process reconciliation when processes are launched by an external agent ([#9389](https://github.com/PyTorchLightning/pytorch-lightning/pull/9389))
 
 
+- Fixed restoring training state during `trainer.fit` only ([#9413](https://github.com/PyTorchLightning/pytorch-lightning/pull/9413))
+
+
 - Fixed missing deepspeed distributed call ([#9540](https://github.com/PyTorchLightning/pytorch-lightning/pull/9540))
 
 
+
```
rohitgr7, line 394:
```suggestion
```

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -169,12 +170,16 @@ def restore_training_state(self) -> None:
...
-        # restore precision plugin (scaler etc.)
-        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
         # restore loops and their progress
         self.restore_loops()
 
-        self.restore_optimizers_and_schedulers()
+        # restore precision plugin (scaler etc.)
+        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
+
+        assert self.trainer.state.fn is not None
```
ananthsub, line 179: why do we need the asserts?

pytorch_lightning/trainer/trainer.py
```diff
@@ -913,8 +913,11 @@ def tune(
 
     def _restore_modules_and_callbacks(self) -> None:
         # restore modules after setup
-        if self.state.fn == TrainerFn.FITTING:
-            self.checkpoint_connector.resume_start()
+        self.checkpoint_connector.resume_start()
+
+        if self.state.fn != TrainerFn.FITTING:
+            return
+
```
ananthsub, line 920: why this swap?

rohitgr7: https://github.com/PyTorchLightning/pytorch-lightning/pull/9413#discussion_r711204473

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: mypy I guess
https://github.com/PyTorchLightning/pytorch-lightning/pull/9413#discussion_r711102946

tests/models/test_restore.py
```diff
@@ -107,7 +107,7 @@ def validation_step_end(self, outputs):
 def test_model_properties_resume_from_checkpoint(tmpdir):
     """Test that properties like `current_epoch` and `global_step` in model and trainer are always the same."""
     model = BoringModel()
-    checkpoint_callback = ModelCheckpoint(dirpath=tmpdir, monitor="val_loss", save_last=True)
+    checkpoint_callback = ModelCheckpoint(dirpath=tmpdir, save_last=True)
```
awaelchli, line 110: let's not change this, seems unrelated to this PR

tests/trainer/test_trainer.py
```diff
@@ -461,6 +511,55 @@ def on_load_checkpoint(self, _):
         assert next_model.num_on_load_checkpoint_called == 1
 
 
+def test_trainer_properties_resume_from_checkpoint(tmpdir):
```
awaelchli, line 514: not a fan of this test because it's relying on the callback system. we could more efficiently test with unit testing by doing something like this:

```python
checkpoint_connector = CheckpointConnector(trainer, ...) 
trainer.state.fn = FIT
checkpoint.connector.resume_start()
checkpoint_connector.restore_xyz()
assert trainer.current_epoch == ...
```


this has several benefits, but most importantly it makes it very clear we are testing the checkpoint connector and not the trainer (which has no business in restoring anything).

rohitgr7: yes, I can do that.. but still need optimizers, schedulers, and lighting module to be setup correctly which can be done withing trainer.fit/validate/test/ 
also, there are few calls that happen during `.fit` only which are required to be tested too. callback seems to be an ideal way to do this end-to-end testing.. ensuring the required functions are called correctly and required things are restored correctly.

CHANGELOG.md
```diff
@@ -443,6 +446,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed PyTorch Profiler not enabled for manual optimization ([#9316](https://github.com/PyTorchLightning/pytorch-lightning/pull/9316))
 - Fixed inspection of other args when a container is specified in `save_hyperparameters` ([#9125](https://github.com/PyTorchLightning/pytorch-lightning/pull/9125))
 - Fixed signature of `Timer.on_train_epoch_end` and `StochasticWeightAveraging.on_train_epoch_end` to prevent unwanted deprecation warnings ([#9347](https://github.com/PyTorchLightning/pytorch-lightning/pull/9347))
+- Fixed error reporting in DDP process reconciliation when processes are launched by an external agent ([#9389](https://github.com/PyTorchLightning/pytorch-lightning/pull/9389))
+- Fixed missing deepspeed distributed call ([#9540](https://github.com/PyTorchLightning/pytorch-lightning/pull/9540))
```
rohitgr7, line 450:
```suggestion
```

tests/checkpointing/test_model_checkpoint.py
awaelchli: if we are now restoring also the state when testing/validating, this will be a breaking change. Let's make sure this is documented in the changelog and the PR marked as breaking change.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -196,6 +201,15 @@ def restore_loops(self) -> None:
         self.trainer.fit_loop.global_step = self._loaded_checkpoint["global_step"]
         self.trainer.fit_loop.current_epoch = self._loaded_checkpoint["epoch"]
 
+        assert self.trainer.state.fn is not None
+        state_dict = self._loaded_checkpoint.get("loops")
+        if state_dict is not None and self.trainer.state.fn != TrainerFn.TUNING:
+            state_fn_loop = f"{self.trainer.state.fn}_loop"
```
tchaton, line 207: Mind adding a test for this change ?

tchaton (approved):
LGTM !

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -196,6 +201,15 @@ def restore_loops(self) -> None:
         self.trainer.fit_loop.global_step = self._loaded_checkpoint["global_step"]
         self.trainer.fit_loop.current_epoch = self._loaded_checkpoint["epoch"]
 
+        assert self.trainer.state.fn is not None
+        state_dict = self._loaded_checkpoint.get("loops")
+        if state_dict is not None and self.trainer.state.fn != TrainerFn.TUNING:
```
carmocca, line 206: Why this change? This has the drawback of losing traceability.

I don't expect anybody subclassing the trainer and implementing an entirely new base loop from scratch.

tests/checkpointing/test_model_checkpoint.py
carmocca:
```suggestion
        assert trainer.global_step == epochs * limit_train_batches
```

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -988,8 +988,8 @@ def assert_checkpoint_log_dir(idx):
...
+        assert trainer.current_epoch == 2
```
carmocca, line 992:
```suggestion
        assert trainer.current_epoch == epochs
```

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -988,8 +988,8 @@ def assert_checkpoint_log_dir(idx):
 
         trainer.test(model)
         # resume_from_checkpoint is resumed when calling `.fit`
```
carmocca, line 990: This comment no longer applies

```suggestion
```

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: but this just ensures that we load the state_dict only for the loop that is required based on TrainerFn. else whenever we call validate/test/fit... it will reload all the loop states every time.

carmocca: Sorry, I'm referring to using `getattr` over `self.trainer.fit_loop...`

rohitgr7: just to avoid all those repetitive if-else statements.

CHANGELOG.md
```diff
@@ -242,6 +242,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `HorovodPlugin.all_gather` to return a `torch.Tensor` instead of a list ([#9696](https://github.com/PyTorchLightning/pytorch-lightning/pull/9696))
 
 
+- Load `current_epoch` and `global_step` irrespective of trainer task ([#9413](https://github.com/PyTorchLightning/pytorch-lightning/pull/9413))
```
rohitgr7, line 245:
```suggestion
- Restore `current_epoch` and `global_step` irrespective of trainer task ([#9413](https://github.com/PyTorchLightning/pytorch-lightning/pull/9413))
```

tests/trainer/connectors/test_checkpoint_connector.py
```diff
@@ -126,3 +131,56 @@ def test_hpc_max_ckpt_version(tmpdir):
     assert trainer.checkpoint_connector.hpc_resume_path == str(tmpdir / "hpc_ckpt_33.ckpt")
     assert trainer.checkpoint_connector.max_ckpt_version_in_folder(tmpdir) == 33
     assert trainer.checkpoint_connector.max_ckpt_version_in_folder(tmpdir / "not" / "existing") is None
+
+
+@mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": "1"})
+@RunIf(min_torch="1.7.0")
+def test_loops_restore(tmpdir):
```
carmocca, line 138: This coudl be a simple unit-test where `restore_loops` is called and the different Loops are mocked to check that their respective `load_state_dict` functions were called.

tests/trainer/test_trainer.py
```diff
@@ -518,6 +521,106 @@ def on_load_checkpoint(self, _):
...
+
+        def on_test_start(self):
+            self._test_on_val_test_predict_tune_start()
+
+    for fn in ("tune", "fit", "validate", "test", "predict"):
+        model = CustomClassifModel()
+        dm = ClassifDataModule()
+        trainer_args.update(
+            {
+                "auto_scale_batch_size": True if fn == "tune" else False,
```
carmocca, line 616:
```suggestion
                "auto_scale_batch_size": fn == "tune"
```

tests/trainer/test_trainer.py
```diff
@@ -518,6 +521,106 @@ def on_load_checkpoint(self, _):
...
+            }
+        )
+
+        trainer = Trainer(**trainer_args)
+        getattr(trainer, fn)(model, datamodule=dm)
```
carmocca, line 621: nit: split this just in case one fails

```suggestion
        trainer_fn = getattr(trainer, fn)
        trainer_fn(model, datamodule=dm)
```

carmocca (commented):
Code looks good, some comments on the tests

tests/trainer/connectors/test_checkpoint_connector.py
rohitgr7: are we checking anywhere else whether loop's state_dict is restored correctly or not?? couldn't find it that's why added these assertions to check everything is working as expected.

carmocca: Yes, all the tests in `tests/loops/test_loops.py` do

rohitgr7: i see.. will update

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -196,6 +202,16 @@ def restore_loops(self) -> None:
         self.trainer.fit_loop.global_step = self._loaded_checkpoint["global_step"]
         self.trainer.fit_loop.current_epoch = self._loaded_checkpoint["epoch"]
 
+        assert self.trainer.state.fn is not None
```
ananthsub, line 205: nit: let's raise an explicit exception instead of relying on assert. ideally assert is only used inside of tests

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -196,6 +202,16 @@ def restore_loops(self) -> None:
...
+        state_dict = self._loaded_checkpoint.get("loops")
+        if state_dict is not None and self.trainer.state.fn != TrainerFn.TUNING:
+            loop_name = f"{self.trainer.state.fn}_loop"
+            loop_attr: Loop = getattr(self.trainer, loop_name)
```
ananthsub, line 209: what happens if `loop_attr` isn't set on the trainer? the exception will be thrown in L210 when `None` tries to load the state dict. i'm with @carmocca that this naming scheme is not as enforced and can break with renames

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: this is just to avoid mypy issues. https://github.com/PyTorchLightning/pytorch-lightning/pull/9413#discussion_r711102946

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: ok.. just trying to avoid those annoying if-else statements here.. I can do a simple dict mapping
```py
if state_dict is not None and self.trainer.state.fn != TrainerFn.TUNING:
    loop_attr_map = {TrainerFn.FITTING: trainer.fit_loop, TrainerFn.VALIDATING: trainer.validate_loop, ...}
    loop_attr: Loop = loop_attr_map[self.trainer.state.fn]
    loop_attr.load_state_dict(state_dict[loop_attr.__name__])
```

or just revert back to the if-else statements? @ananthsub ?

ananthsub: i like the explicit if-else statements because then github/editors can surface the references, which helps in traceability and refactoring.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -171,12 +172,16 @@ def restore_training_state(self) -> None:
...
             return
 
-        # restore precision plugin (scaler etc.)
-        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
         # restore loops and their progress
         self.restore_loops()
 
-        self.restore_optimizers_and_schedulers()
+        # restore precision plugin (scaler etc.)
+        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
```
awaelchli, line 179: can  we move this back before restore_loops(), just in case. We don't want to change the order of things when not necessary.

pytorch_lightning/trainer/trainer.py
```diff
@@ -953,8 +953,11 @@ def tune(
 
     def _restore_modules_and_callbacks(self) -> None:
         # restore modules after setup
-        if self.state.fn == TrainerFn.FITTING:
-            self.checkpoint_connector.resume_start()
+        self.checkpoint_connector.resume_start()
```
awaelchli, line 956: then can we move this outside to where _restore_modules_and_callbacks() is called?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1152,7 +1155,8 @@ def _pre_training_routine(self):
         # register signals
         self.signal_connector.register_signal_handlers()
 
-        self.checkpoint_connector.resume_end()
+        if self.state.fn != TrainerFn.TUNING:
+            self.checkpoint_connector.resume_end()
```
awaelchli, line 1159: Given the objective of this PR, this now needs to move outside and not only be called for training. (_pre_training_routine is only called for training)

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
rohitgr7: that would actually create a problem. if someone do this
```py
Trainer(resume_from_checkpoint=...)
trainer.validate()
trainer.fit()  # here resume_from_checkpoint will be gone since validate will call resume_end and will destroy resume_from_checkpoint
```
this is a bit misalignment and should not work like this, but I think unify_checkpoint PR will clean this up in a better way.

pytorch_lightning/trainer/trainer.py
rohitgr7: yeah make sense.. done

tests/trainer/connectors/test_checkpoint_connector.py
```diff
@@ -126,3 +130,51 @@ def test_hpc_max_ckpt_version(tmpdir):
...
+        limit_val_batches=2,
+        logger=False,
+        callbacks=[checkpoint_callback],
+        num_sanity_val_steps=0,
+    )
+    trainer = Trainer(**trainer_args)
+    trainer.validate(model)
+    trainer.test(model)
+    trainer.predict(model)
+    trainer.fit(model)
```
carmocca, line 154: Not a fan of this test as it tries to unit-test the `load_state_dict` behaviour but still runs the trainer for all these entry-points

I would either remove it or reduce the scope so it interacts with the connector directly without using the trainer

tests/trainer/test_trainer.py
```diff
@@ -518,6 +521,107 @@ def on_load_checkpoint(self, _):
...
+
+    for fn in ("tune", "fit", "validate", "test", "predict"):
+        model = CustomClassifModel()
+        dm = ClassifDataModule()
+        trainer_args.update(
+            {
+                "auto_scale_batch_size": fn == "tune",
+            }
+        )
+
```
carmocca, line 619:
```suggestion
        trainer_args["auto_scale_batch_size"] = fn == "tune"
```

tests/trainer/test_trainer.py
```diff
@@ -518,6 +521,107 @@ def on_load_checkpoint(self, _):
...
+            )
+
+        def _check_schedulers(self):
+            return all(
+                self._is_equal(self.trainer.lr_schedulers[i]["scheduler"].state_dict(), state_dict["lr_schedulers"][i])
+                for i in range(len(self.trainer.lr_schedulers))
+            )
+
+        def _check_model_state_dict(self):
+            for k in self.state_dict().keys():
```
carmocca, line 579:
```suggestion
            for k in self.state_dict():
```

tests/trainer/test_trainer.py
```diff
@@ -518,6 +521,107 @@ def on_load_checkpoint(self, _):
...
+        num_sanity_val_steps=0,
+    )
+    trainer = Trainer(**trainer_args)
+    trainer.fit(model, datamodule=dm)
+
+    resume_ckpt = str(tmpdir / "last.ckpt")
+    state_dict = torch.load(resume_ckpt)
+
+    trainer_args.update(
+        {"max_epochs": 3, "resume_from_checkpoint": resume_ckpt, "checkpoint_callback": False, "callbacks": []}
```
carmocca, line 553: why does this needs to pass an empty callback list?

rohitgr7: because trainer_args contains a checkpoint_callback inside callbacks few lines above. If I don't remove it, it will keep update last ckpt in trainer.fit will be override the original ckpt.

tests/trainer/test_trainer.py
```diff
@@ -518,6 +521,102 @@ def on_load_checkpoint(self, _):
         assert next_model.num_on_load_checkpoint_called == 1
 
 
+def test_trainer_properties_restore_resume_from_checkpoint(tmpdir):
```
carmocca, line 524: Move the test to `tests/models/test_restore.py`

carmocca (approved):

PR #9044, opened 2021-08-23
Move `init_ddp_connection` to distributed utilities

kaushikb11:
## What does this PR do?

It is duplicated in DDP & DDPSpawn plugins.

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

Issue #7324, opened 2021-05-03
Precision Plugins should be part of Training Type Plugins


ananthsub:
## 🚀 Feature

Precision Plugin should be a property of the training type plugin, not the accelerator. these can overlap, and interleaving sequences of operations across the two in the accelerator can be challenging to get fully right.

cc @justusschock @awaelchli @kaushikb11 @tchaton 

todo:  fill out in more detail

### Motivation

This would make management of things like this easier:
- https://github.com/PyTorchLightning/pytorch-lightning/issues/7271
- https://github.com/PyTorchLightning/pytorch-lightning/issues/7217

justusschock:
@ananthsub This was discussed quite a few times already now and I think the agreement was to change this, when we maybe also move the accelerator inside the training type plugin 

cc @SeanNaren

SeanNaren:
Thanks @ananthsub for making this. I recall doing an initial POC of this a few months back and realised there is quite some work to make this change when being careful of BW compat.

It would also make these 'shell' classes obsolete, like we see with deepspeed: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/plugins/precision/deepspeed_precision.py

all the above logic should live within the `TrainingTypePlugin` really. We do have to be careful though, I do like distinctions to remain (precision logic to be separate if possible).

carmocca:
It might be helpful to compare how the implementation of one training type or plugin would look if this was done. As it stands, it is not clear to me exactly what the benefits of this are

Issue #6094, opened 2021-02-20
API consistency: "val" vs "validation"


msegado:
### Motivation

As a new user learning PyTorch Lightning, I was surprised at the naming inconsistency between the `validation_step()` and `val_dataloader()` hooks. Searching through the codebase, it seems that *most* methods or getters/setters favor "validation"...

- `validation_step_end`
- `validation_step`
- `validation_epoch_end`
- `on_validation_start`
- `on_validation_model_train`
- `on_validation_model_eval`
- `on_validation_epoch_start`
- `on_validation_epoch_end`
- `on_validation_end`
- `on_validation_batch_start`
- `on_validation_batch_end`
- `init_validation_tqdm`
- `enable_validation`
- `disable_validation`
- `_on_validation_start_log`
- `_on_validation_epoch_start_log`
- `_on_validation_epoch_end_log`
- `_on_validation_end_log`
- `_on_validation_batch_start_log`
- `_on_validation_batch_end_log`

...while a handful use "val" instead:
- `val_transforms`
- `val_step_context`
- `val_dataloader`
- `val_batch_idx`
- `total_val_batches`
- `should_check_val_fx`
- `reset_val_dataloader`
- `reset_train_val_dataloaders`
- `num_seen_val_check_batches`

I wasn't able to see any obvious pattern in which methods got which naming convention, and think having a consistent convention throughout would make for a (slightly) cleaner and easier-to-learn library.

Related: https://github.com/PyTorchLightning/pytorch-lightning/pull/634

### Pitch

Standardize all validation-related methods (and properties, if any exist) to use the same naming convention; either "val" (for consistency with e.g. torchvision) or "validation" (for consistency with the bulk of the current Lightning API). Renamed methods would be marked as deprecated and removed in the next major version bump.

Borda:
cc: @PyTorchLightning/core-contributors ^^

s-rog:
If we were to implement this change, I would prefer `val` for its brevity.

justusschock:
We have the same in other parts as well... train vs. training, test vs testing

carmocca:
Poll:

msegado:
[...]

PR #9414, opened 2021-09-09
2/n Consolidate collective functions - collective base and subclasses

### LINKED ISSUE #7534: Offer collective functions under Lightning utilities
## 🚀 Feature
Lightning should offer a central place to use the collective functions provided here: https://pytorch.org/docs/stable/distributed.html#collective-functions

### Motivation

LightningModule code is usually agnostic to what device its running on or whether its running in a distributed training environment. However, there are times where the module does need to rely on collective functions. 

In Lightning, we currently have many places where these are offered:
- On this distributed object, which only supports `broadcast`: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/distributed/dist.py

- `reduce`, `barrier`, `broadcast`, `all_gather`, and `reduce_boolean_decision` are on the trainer's accelerator and training type plugin: 
- https://github.com/PyTorchLightning/pytorch-lightning/blob/233f252bb427c930be8e7ca56fe115b637278b8d/pytorch_lightning/accelerators/accelerator.py#L431-L455
https://github.com/PyTorchLightning/pytorch-lightning/blob/233f252bb427c930be8e7ca56fe115b637278b8d/pytorch_lightning/plugins/training_type/training_type_plugin.py#L78-L103

- more utilities for gathering tensors, all_gather, and sync_ddp here: https://github.com/PyTorchLightning/pytorch-lightning/blob/b9a52fa2ef31f12f6992ece18a033318ec551907/pytorch_lightning/utilities/distributed.py#L86-L217

- `all_gather` repeated again here on the lightning module, calling the trainer's accelerator functions: https://github.com/PyTorchLightning/pytorch-lightning/blob/233f252bb427c930be8e7ca56fe115b637278b8d/pytorch_lightning/core/lightning.py#L506-L532

Some of these call each other and the dependency isn't very clear now, so it is confusing for users which to go through.

### Pitch

1. Offer these utilities under a central place: `pytorch_lightning/utilities/collectives.py` for these utilities:
`barrier`, `all_gather`, `broadcast`, etc

These should be very thin wrappers over the PyTorch distributed functions, checking if torch.distributed is available and initialized. If not, we return what's expected for single-process training.

2. Update the callsites internally to use to these implementations

3. Mark existing functions as deprecated and slated for removal in v1.6

four4fish:
## What does this PR do?

Implementing proposal here https://docs.google.com/document/d/1e83FcZHHHsTmiUmNpgmPTBYjugZaC3pcZhd4VV9AuIM/edit

Steps:
1. [RFC] create pytorch_lightning/utilities/collective_util.py and move collective related utils from distributed.py to collective_util.py
2.  **2/n Consolidate collective functions - collective base and subclasses**. (this PR)
3. Integrate with all training_type_plugins (ddp, dp, fullshard and etc) and Accelerators

Fixes #7534 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/plugins/collective/__init__.py
+++ b/pytorch_lightning/plugins/collective/__init__.py
@@ -1,0 +1,18 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from pytorch_lightning.plugins.collective.collective_plugin import Collective  # noqa: F401
+from pytorch_lightning.plugins.collective.single_device_collective import SingleNodeCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.torch_collective import TorchCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.horovod_collective import HorovodCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.tpu_collective import TPUCollective  # noqa: F401
--- a/pytorch_lightning/plugins/collective/__init__.py
+++ b/pytorch_lightning/plugins/collective/__init__.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from pytorch_lightning.plugins.collective.collective_plugin import Collective  # noqa: F401
-from pytorch_lightning.plugins.collective.single_device_collective import SingleNodeCollective  # noqa: F401
-from pytorch_lightning.plugins.collective.torch_collective import TorchCollective  # noqa: F401
 from pytorch_lightning.plugins.collective.horovod_collective import HorovodCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.single_device_collective import SingleDeviceCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.torch_collective import TorchCollective  # noqa: F401
 from pytorch_lightning.plugins.collective.tpu_collective import TPUCollective  # noqa: F401
--- a/pytorch_lightning/plugins/collective/__init__.py
+++ b/pytorch_lightning/plugins/collective/__init__.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from pytorch_lightning.plugins.collective.collective_plugin import Collective  # noqa: F401
+from pytorch_lightning.plugins.collective.collective_plugin import CollectivePlugin  # noqa: F401
 from pytorch_lightning.plugins.collective.horovod_collective import HorovodCollective  # noqa: F401
 from pytorch_lightning.plugins.collective.single_device_collective import SingleDeviceCollective  # noqa: F401
 from pytorch_lightning.plugins.collective.torch_collective import TorchCollective  # noqa: F401
--- a/pytorch_lightning/plugins/collective/collective_plugin.py
+++ b/pytorch_lightning/plugins/collective/collective_plugin.py
@@ -12,16 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from abc import ABC, abstractmethod
-from typing import Any, Optional, Union
+from typing import Any, List, Optional, Union
 
 import torch
+
+from pytorch_lightning.utilities.distributed import ReduceOp
 
 
 class Collective(ABC):
     """Base class for collective functions for training type plugins."""
 
     @abstractmethod
-    def barrier(self, name: Optional[str] = None, *args, **kwargs) -> None:
+    def barrier(self, name: Optional[str] = None) -> None:
         """Forces all possibly joined processes to wait for each other."""
 
     @abstractmethod
@@ -29,11 +31,18 @@ class Collective(ABC):
         """Broadcasts an object to all processes."""
 
     @abstractmethod
-    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+    def all_gather(
+        self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False
+    ) -> Union[List[torch.Tensor], torch.Tensor]:
         """Perform a all_gather on all processes."""
 
     @abstractmethod
-    def reduce(self, tensor: Union[torch.Tensor, Any], *args: Any, **kwargs: Any) -> Union[torch.Tensor, Any]:
+    def reduce(
+        self,
+        tensor: Union[torch.Tensor, Any],
+        group: Optional[Any] = None,
+        reduce_op: Optional[Union[ReduceOp, str]] = "mean",
+    ) -> Union[torch.Tensor, Any]:
         """Reduces the given tensor (e.g. across GPUs/processes).
 
         Args:
@@ -42,6 +51,6 @@ class Collective(ABC):
             **kwargs: plugin-specific keyword arguments
         """
 
+    @abstractmethod
     def reduce_boolean_decision(self, decision: bool) -> bool:
         """Reduce the early stopping decision across all processes."""
-        return decision
--- a/pytorch_lightning/plugins/collective/collective_plugin.py
+++ b/pytorch_lightning/plugins/collective/collective_plugin.py
@@ -1,0 +1,47 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from abc import ABC, abstractmethod
+from typing import Any, Optional, Union
+
+import torch
+
+
+class Collective(ABC):
+    """Base class for collective functions for training type plugins."""
+
+    @abstractmethod
+    def barrier(self, name: Optional[str] = None, *args, **kwargs) -> None:
+        """Forces all possibly joined processes to wait for each other."""
+
+    @abstractmethod
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        """Broadcasts an object to all processes."""
+
+    @abstractmethod
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        """Perform a all_gather on all processes."""
+
+    @abstractmethod
+    def reduce(self, tensor: Union[torch.Tensor, Any], *args: Any, **kwargs: Any) -> Union[torch.Tensor, Any]:
+        """Reduces the given tensor (e.g. across GPUs/processes).
+
+        Args:
+            tensor: the tensor to sync and reduce
+            *args: plugin-specific positional arguments
+            **kwargs: plugin-specific keyword arguments
+        """
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
+        return decision
--- a/pytorch_lightning/plugins/collective/collective_plugin.py
+++ b/pytorch_lightning/plugins/collective/collective_plugin.py
@@ -19,22 +19,41 @@ import torch
 from pytorch_lightning.utilities.distributed import ReduceOp
 
 
-class Collective(ABC):
-    """Base class for collective functions for training type plugins."""
+class CollectivePlugin(ABC):
+    """Interface for collective functions for training type plugins.
+
+    Lightning collective supports communications between multiple processes and multiple nodes, provides routines such
+    as barrier, broadcast, all_gather, and reduce, reduce
+    """
 
     @abstractmethod
     def barrier(self, name: Optional[str] = None) -> None:
-        """Forces all possibly joined processes to wait for each other."""
+        """Synchronizes all processes which blocks processes until the whole group enters this function.
+
+        Args:
+            name: a str pass into barrier. Only torch xla respect this param
+        """
 
     @abstractmethod
     def broadcast(self, obj: object, src: int = 0) -> object:
-        """Broadcasts an object to all processes."""
+        """Broadcasts an object to all processes.
+
+        Args:
+            obj: the object to broadcast
+            src: source rank.
+        """
 
     @abstractmethod
     def all_gather(
-        self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False
+        self, tensor: torch.Tensor, process_group: Optional[Any] = None, sync_grads: bool = False
     ) -> Union[List[torch.Tensor], torch.Tensor]:
-        """Perform a all_gather on all processes."""
+        """Perform a all_gather on all processes.
+
+        Args:
+            tensor: the tensor to all_gather
+            process_group: the process group to gather results from
+            sync_grads: flag that allows users to synchronize gradients for all_gather op
+        """
 
     @abstractmethod
     def reduce(
@@ -47,10 +66,9 @@ class Collective(ABC):
 
         Args:
             tensor: the tensor to sync and reduce
+            process_group: the process group to reduce
+            reduce_op: the reduction operation. Defaults to 'mean'.
+                Can also be a string 'sum' or ReduceOp.
             *args: plugin-specific positional arguments
             **kwargs: plugin-specific keyword arguments
         """
-
-    @abstractmethod
-    def reduce_boolean_decision(self, decision: bool) -> bool:
-        """Reduce the early stopping decision across all processes."""
--- a/pytorch_lightning/plugins/collective/collective_plugin.py
+++ b/pytorch_lightning/plugins/collective/collective_plugin.py
@@ -20,10 +20,13 @@ from pytorch_lightning.utilities.distributed import ReduceOp
 
 
 class CollectivePlugin(ABC):
-    """Interface for collective functions for training type plugins.
+    """Interface for collective functions.
 
     Lightning collective supports communications between multiple processes and multiple nodes, provides routines such
-    as barrier, broadcast, all_gather, and reduce, reduce
+    as barrier, broadcast, all_gather, and reduce
+
+    .. note::
+        This API is experimental/in-beta and subject to change
     """
 
     @abstractmethod
@@ -53,6 +56,8 @@ class CollectivePlugin(ABC):
             tensor: the tensor to all_gather
             process_group: the process group to gather results from
             sync_grads: flag that allows users to synchronize gradients for all_gather op
+
+        Returns: a tensor (torch distributed) or a list of tensor (horovod)
         """
 
     @abstractmethod
--- a/pytorch_lightning/plugins/collective/collective_plugin.py
+++ b/pytorch_lightning/plugins/collective/collective_plugin.py
@@ -59,7 +59,7 @@ class CollectivePlugin(ABC):
     def reduce(
         self,
         tensor: Union[torch.Tensor, Any],
-        group: Optional[Any] = None,
+        process_group: Optional[Any] = None,
         reduce_op: Optional[Union[ReduceOp, str]] = "mean",
     ) -> Union[torch.Tensor, Any]:
         """Reduces the given tensor (e.g. across GPUs/processes).
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -39,8 +39,8 @@ class HorovodCollective(Collective):
     def join(self) -> None:
         """Horovod function that indicates that the rank finished processing data.
 
-        All ranks that did not call join() continue to process allreduce operations. This function blocks the Python thread
-        until all ranks join.
+        All ranks that did not call join() continue to process allreduce operations. This function blocks the Python
+        thread until all ranks join.
         """
         if self.on_gpu:
             hvd.join(self.local_rank)
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -11,21 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import io
-from typing import Any, Optional, Union
+from typing import Any, List, Optional, Union
 
 import torch
 
 from pytorch_lightning.plugins.collective import Collective
 from pytorch_lightning.utilities import _HOROVOD_AVAILABLE
+from pytorch_lightning.utilities.distributed import distributed_available
+from pytorch_lightning.utilities.distributed import group as dist_group
 from pytorch_lightning.utilities.distributed import ReduceOp
-from pytorch_lightning.utilities.types import _TPU_AVAILABLE
-
-if _TPU_AVAILABLE:
-    import torch_xla.core.xla_model as xm
-    from torch_xla.core.xla_model import rendezvous
-else:
-    xm, rendezvous = [None] * 4
 
 if _HOROVOD_AVAILABLE:
     import horovod.torch as hvd
@@ -42,48 +36,47 @@ class HorovodCollective(Collective):
         self._on_gpu = on_gpu
         self._local_rank = local_rank
 
-    def join(self):
+    def join(self) -> None:
         """Horovod function that indicates that the rank finished processing data.
 
         All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread
         until all ranks join.
         """
-        if self.on_gpu:
-            hvd.join(self.local_rank)
+        if self._on_gpu:
+            hvd.join(self._local_rank)
         else:
             hvd.join()
 
-    def barrier(self, name: Optional[str] = None) -> None:
-        if self.is_distributed:
-            rendezvous(name)
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
+        if distributed_available():
+            self.join()
 
     def broadcast(self, obj: object, src: int = 0) -> object:
-        if not self.is_distributed:
-            return obj
-        buffer = io.BytesIO()
-        torch.save(obj, buffer)
-        data = bytearray(buffer.getbuffer())
-        data_tensor = torch.tensor(data, device=self.root_device, dtype=torch.float)
-        data = xm.all_gather(data_tensor)
-        buffer = io.BytesIO(data.cpu().byte().numpy())
-        obj = torch.load(buffer)
+        obj = hvd.broadcast_object(obj, src)
         return obj
 
-    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
-        """
-        Function to gather a tensor from several distributed processes
-        Args:
-            tensor: tensor of shape (batch, ...)
-            group: not available with TPUs
-            sync_grads: not available with TPUs
-        Return:
-            A tensor of shape (world_size, batch, ...)
-        """
-        if isinstance(tensor, torch.Tensor) and tensor.dim() == 0:
-            tensor = tensor.unsqueeze(0)
-        return self._xm.all_gather(tensor)
+    def all_gather(
+        self, result: Union[torch.Tensor], group: Optional[Any] = dist_group.WORLD, sync_grads: bool = False
+    ) -> List[torch.Tensor]:
+        if group is not None and group != dist_group.WORLD:
+            raise ValueError("Horovod does not support allgather using a subcommunicator at this time. Unset `group`.")
 
-    def reduce(self, tensor, group: Optional[Any] = None, reduce_op: Optional[Union[ReduceOp, str]] = "mean"):
+        if len(result.shape) == 0:
+            # Convert scalars to single dimension tensors
+            result = result.reshape(1)
+
+        # sync and gather all
+        self.join()
+        gathered = hvd.allgather(result)
+        gathered_result = list(gathered.split(1, dim=0))
+        return gathered_result
+
+    def reduce(
+        self,
+        tensor: Union[torch.Tensor, Any],
+        group: Optional[Any] = None,
+        reduce_op: Optional[Union[ReduceOp, str]] = "mean",
+    ) -> Union[torch.Tensor, Any]:
         """Reduces a tensor from several distributed processes to one aggregated tensor.
 
         Args:
@@ -108,3 +101,7 @@ class HorovodCollective(Collective):
         # sync all processes before reduction
         self.join()
         return hvd.allreduce(tensor, op=reduce_op)
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
+        return decision
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -1,0 +1,110 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import io
+from typing import Any, Optional, Union
+
+import torch
+
+from pytorch_lightning.plugins.collective import Collective
+from pytorch_lightning.utilities import _HOROVOD_AVAILABLE
+from pytorch_lightning.utilities.distributed import ReduceOp
+from pytorch_lightning.utilities.types import _TPU_AVAILABLE
+
+if _TPU_AVAILABLE:
+    import torch_xla.core.xla_model as xm
+    from torch_xla.core.xla_model import rendezvous
+else:
+    xm, rendezvous = [None] * 4
+
+if _HOROVOD_AVAILABLE:
+    import horovod.torch as hvd
+
+
+class HorovodCollective(Collective):
+    """Collective interface for Horovod training type plugins."""
+
+    def __init__(
+        self,
+        on_gpu: Optional[bool] = False,
+        local_rank: Optional[int] = 0,
+    ):
+        self._on_gpu = on_gpu
+        self._local_rank = local_rank
+
+    def join(self):
+        """Horovod function that indicates that the rank finished processing data.
+
+        All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread
+        until all ranks join.
+        """
+        if self.on_gpu:
+            hvd.join(self.local_rank)
+        else:
+            hvd.join()
+
+    def barrier(self, name: Optional[str] = None) -> None:
+        if self.is_distributed:
+            rendezvous(name)
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        if not self.is_distributed:
+            return obj
+        buffer = io.BytesIO()
+        torch.save(obj, buffer)
+        data = bytearray(buffer.getbuffer())
+        data_tensor = torch.tensor(data, device=self.root_device, dtype=torch.float)
+        data = xm.all_gather(data_tensor)
+        buffer = io.BytesIO(data.cpu().byte().numpy())
+        obj = torch.load(buffer)
+        return obj
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        """
+        Function to gather a tensor from several distributed processes
+        Args:
+            tensor: tensor of shape (batch, ...)
+            group: not available with TPUs
+            sync_grads: not available with TPUs
+        Return:
+            A tensor of shape (world_size, batch, ...)
+        """
+        if isinstance(tensor, torch.Tensor) and tensor.dim() == 0:
+            tensor = tensor.unsqueeze(0)
+        return self._xm.all_gather(tensor)
+
+    def reduce(self, tensor, group: Optional[Any] = None, reduce_op: Optional[Union[ReduceOp, str]] = "mean"):
+        """Reduces a tensor from several distributed processes to one aggregated tensor.
+
+        Args:
+            tensor: the tensor to sync and reduce
+            group: the process group to gather results from. Defaults to all processes (world)
+            reduce_op: the reduction operation. Defaults to 'mean'/'avg'.
+                Can also be a string 'sum' to calculate the sum during reduction.
+
+        Return:
+            reduced value, except when the input was not a tensor the output remains is unchanged
+        """
+        if group is not None:
+            raise ValueError("Horovod does not support allreduce using a subcommunicator at this time. Unset `group`.")
+
+        if reduce_op in (None, "avg", "mean"):
+            reduce_op = hvd.Average
+        elif reduce_op in ("sum", ReduceOp.SUM):
+            reduce_op = hvd.Sum
+        else:
+            raise ValueError(f"unrecognized `reduce_op`: {reduce_op}")
+
+        # sync all processes before reduction
+        self.join()
+        return hvd.allreduce(tensor, op=reduce_op)
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -31,10 +31,10 @@ class HorovodCollective(Collective):
     def __init__(
         self,
         on_gpu: Optional[bool] = False,
-        local_rank: Optional[int] = 0,
+        local_rank: int = 0,
     ):
-        self._on_gpu = on_gpu
-        self._local_rank = local_rank
+        self.on_gpu = on_gpu
+        self.local_rank = local_rank
 
     def join(self) -> None:
         """Horovod function that indicates that the rank finished processing data.
@@ -42,8 +42,8 @@ class HorovodCollective(Collective):
         All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread
         until all ranks join.
         """
-        if self._on_gpu:
-            hvd.join(self._local_rank)
+        if self.on_gpu:
+            hvd.join(self.local_rank)
         else:
             hvd.join()
 
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -32,14 +32,14 @@ class HorovodCollective(Collective):
         self,
         on_gpu: Optional[bool] = False,
         local_rank: int = 0,
-    ):
+    ) -> None:
         self.on_gpu = on_gpu
         self.local_rank = local_rank
 
     def join(self) -> None:
         """Horovod function that indicates that the rank finished processing data.
 
-        All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread
+        All ranks that did not call join() continue to process allreduce operations. This function blocks the Python thread
         until all ranks join.
         """
         if self.on_gpu:
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -15,7 +15,7 @@ from typing import Any, List, Optional, Union
 
 import torch
 
-from pytorch_lightning.plugins.collective import Collective
+from pytorch_lightning.plugins.collective import CollectivePlugin
 from pytorch_lightning.utilities import _HOROVOD_AVAILABLE
 from pytorch_lightning.utilities.distributed import distributed_available
 from pytorch_lightning.utilities.distributed import group as dist_group
@@ -25,18 +25,18 @@ if _HOROVOD_AVAILABLE:
     import horovod.torch as hvd
 
 
-class HorovodCollective(Collective):
+class HorovodCollective(CollectivePlugin):
     """Collective interface for Horovod training type plugins."""
 
     def __init__(
         self,
-        on_gpu: Optional[bool] = False,
+        on_gpu: bool = False,
         local_rank: int = 0,
     ) -> None:
         self.on_gpu = on_gpu
         self.local_rank = local_rank
 
-    def join(self) -> None:
+    def _join(self) -> None:
         """Horovod function that indicates that the rank finished processing data.
 
         All ranks that did not call join() continue to process allreduce operations. This function blocks the Python
@@ -49,7 +49,7 @@ class HorovodCollective(Collective):
 
     def barrier(self, *args: Any, **kwargs: Any) -> None:
         if distributed_available():
-            self.join()
+            self._join()
 
     def broadcast(self, obj: object, src: int = 0) -> object:
         obj = hvd.broadcast_object(obj, src)
@@ -66,7 +66,7 @@ class HorovodCollective(Collective):
             result = result.reshape(1)
 
         # sync and gather all
-        self.join()
+        self._join()
         gathered = hvd.allgather(result)
         gathered_result = list(gathered.split(1, dim=0))
         return gathered_result
@@ -77,17 +77,6 @@ class HorovodCollective(Collective):
         group: Optional[Any] = None,
         reduce_op: Optional[Union[ReduceOp, str]] = "mean",
     ) -> Union[torch.Tensor, Any]:
-        """Reduces a tensor from several distributed processes to one aggregated tensor.
-
-        Args:
-            tensor: the tensor to sync and reduce
-            group: the process group to gather results from. Defaults to all processes (world)
-            reduce_op: the reduction operation. Defaults to 'mean'/'avg'.
-                Can also be a string 'sum' to calculate the sum during reduction.
-
-        Return:
-            reduced value, except when the input was not a tensor the output remains is unchanged
-        """
         if group is not None:
             raise ValueError("Horovod does not support allreduce using a subcommunicator at this time. Unset `group`.")
 
@@ -99,9 +88,5 @@ class HorovodCollective(Collective):
             raise ValueError(f"unrecognized `reduce_op`: {reduce_op}")
 
         # sync all processes before reduction
-        self.join()
+        self._join()
         return hvd.allreduce(tensor, op=reduce_op)
-
-    def reduce_boolean_decision(self, decision: bool) -> bool:
-        """Reduce the early stopping decision across all processes."""
-        return decision
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -56,9 +56,9 @@ class HorovodCollective(CollectivePlugin):
         return obj
 
     def all_gather(
-        self, result: Union[torch.Tensor], group: Optional[Any] = dist_group.WORLD, sync_grads: bool = False
+        self, result: Union[torch.Tensor], process_group: Optional[Any] = dist_group.WORLD, sync_grads: bool = False
     ) -> List[torch.Tensor]:
-        if group is not None and group != dist_group.WORLD:
+        if process_group is not None and process_group != dist_group.WORLD:
             raise ValueError("Horovod does not support allgather using a subcommunicator at this time. Unset `group`.")
 
         if len(result.shape) == 0:
@@ -74,11 +74,13 @@ class HorovodCollective(CollectivePlugin):
     def reduce(
         self,
         tensor: Union[torch.Tensor, Any],
-        group: Optional[Any] = None,
+        process_group: Optional[Any] = None,
         reduce_op: Optional[Union[ReduceOp, str]] = "mean",
     ) -> Union[torch.Tensor, Any]:
-        if group is not None:
-            raise ValueError("Horovod does not support allreduce using a subcommunicator at this time. Unset `group`.")
+        if process_group is not None:
+            raise ValueError(
+                "Horovod does not support allreduce using a subcommunicator at this time. Unset `process_group`."
+            )
 
         if reduce_op in (None, "avg", "mean"):
             reduce_op = hvd.Average
--- a/pytorch_lightning/plugins/collective/horovod_collective.py
+++ b/pytorch_lightning/plugins/collective/horovod_collective.py
@@ -52,8 +52,7 @@ class HorovodCollective(CollectivePlugin):
             self._join()
 
     def broadcast(self, obj: object, src: int = 0) -> object:
-        obj = hvd.broadcast_object(obj, src)
-        return obj
+        return hvd.broadcast_object(obj, src)
 
     def all_gather(
         self, result: Union[torch.Tensor], process_group: Optional[Any] = dist_group.WORLD, sync_grads: bool = False
--- a/pytorch_lightning/plugins/collective/single_device_collective.py
+++ b/pytorch_lightning/plugins/collective/single_device_collective.py
@@ -18,7 +18,7 @@ import torch
 from pytorch_lightning.plugins.collective import Collective
 
 
-class SingleNodeCollective(Collective):
+class SingleDeviceCollective(Collective):
     """Collective interface for single device training type plugins."""
 
     def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
--- a/pytorch_lightning/plugins/collective/single_device_collective.py
+++ b/pytorch_lightning/plugins/collective/single_device_collective.py
@@ -21,7 +21,7 @@ from pytorch_lightning.plugins.collective import Collective
 class SingleNodeCollective(Collective):
     """Collective interface for single device training type plugins."""
 
-    def barrier(self, name: Optional[str] = None, *args, **kwargs) -> None:
+    def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
         """Forces all possibly joined processes to wait for each other."""
         pass
 
@@ -42,3 +42,7 @@ class SingleNodeCollective(Collective):
             **kwargs: plugin-specific keyword arguments
         """
         return tensor
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
+        return decision
--- a/pytorch_lightning/plugins/collective/single_device_collective.py
+++ b/pytorch_lightning/plugins/collective/single_device_collective.py
@@ -1,0 +1,44 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, Optional, Union
+
+import torch
+
+from pytorch_lightning.plugins.collective import Collective
+
+
+class SingleNodeCollective(Collective):
+    """Collective interface for single device training type plugins."""
+
+    def barrier(self, name: Optional[str] = None, *args, **kwargs) -> None:
+        """Forces all possibly joined processes to wait for each other."""
+        pass
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        """Broadcasts an object to all processes."""
+        return obj
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        """Perform a all_gather on all processes."""
+        return tensor
+
+    def reduce(self, tensor: Union[torch.Tensor, Any], *args: Any, **kwargs: Any) -> Union[torch.Tensor, Any]:
+        """Reduces the given tensor (e.g. across GPUs/processes).
+
+        Args:
+            tensor: the tensor to sync and reduce
+            *args: plugin-specific positional arguments
+            **kwargs: plugin-specific keyword arguments
+        """
+        return tensor
--- a/pytorch_lightning/plugins/collective/single_device_collective.py
+++ b/pytorch_lightning/plugins/collective/single_device_collective.py
@@ -15,34 +15,20 @@ from typing import Any, Optional, Union
 
 import torch
 
-from pytorch_lightning.plugins.collective import Collective
+from pytorch_lightning.plugins.collective import CollectivePlugin
 
 
-class SingleDeviceCollective(Collective):
+class SingleDeviceCollective(CollectivePlugin):
     """Collective interface for single device training type plugins."""
 
     def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
-        """Forces all possibly joined processes to wait for each other."""
         pass
 
     def broadcast(self, obj: object, src: int = 0) -> object:
-        """Broadcasts an object to all processes."""
         return obj
 
     def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
-        """Perform a all_gather on all processes."""
         return tensor
 
     def reduce(self, tensor: Union[torch.Tensor, Any], *args: Any, **kwargs: Any) -> Union[torch.Tensor, Any]:
-        """Reduces the given tensor (e.g. across GPUs/processes).
-
-        Args:
-            tensor: the tensor to sync and reduce
-            *args: plugin-specific positional arguments
-            **kwargs: plugin-specific keyword arguments
-        """
         return tensor
-
-    def reduce_boolean_decision(self, decision: bool) -> bool:
-        """Reduce the early stopping decision across all processes."""
-        return decision
--- a/pytorch_lightning/plugins/collective/single_device_collective.py
+++ b/pytorch_lightning/plugins/collective/single_device_collective.py
@@ -27,7 +27,9 @@ class SingleDeviceCollective(CollectivePlugin):
     def broadcast(self, obj: object, src: int = 0) -> object:
         return obj
 
-    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+    def all_gather(
+        self, tensor: torch.Tensor, process_group: Optional[Any] = None, sync_grads: bool = False
+    ) -> torch.Tensor:
         return tensor
 
     def reduce(self, tensor: Union[torch.Tensor, Any], *args: Any, **kwargs: Any) -> Union[torch.Tensor, Any]:
--- a/pytorch_lightning/plugins/collective/torch_collective.py
+++ b/pytorch_lightning/plugins/collective/torch_collective.py
@@ -16,22 +16,20 @@ from typing import Any, Optional, Union
 import torch
 import torch.distributed
 
+from pytorch_lightning.overrides.torch_distributed import broadcast_object_list
 from pytorch_lightning.plugins.collective import Collective
 from pytorch_lightning.utilities import _TORCH_GREATER_EQUAL_1_8
 from pytorch_lightning.utilities.apply_func import apply_to_collection
-from pytorch_lightning.utilities.distributed import (
-    all_gather_ddp_if_available,
-    distributed_available,
-    ReduceOp,
-    sync_ddp_if_available,
-)
+from pytorch_lightning.utilities.distributed import all_gather_ddp_if_available, distributed_available
+from pytorch_lightning.utilities.distributed import group as _group
+from pytorch_lightning.utilities.distributed import ReduceOp, sync_ddp_if_available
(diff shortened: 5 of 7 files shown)


pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,47 @@
...
+
+        Args:
+            tensor: the tensor to sync and reduce
+            *args: plugin-specific positional arguments
+            **kwargs: plugin-specific keyword arguments
+        """
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
+        return decision
```
ananthsub, line 47: can this be automatically derived given a collective as a utility function?
```
def reduce_boolean_decision(Collective c, decision: bool) -> bool
``` 

if so, it doesn't need to be on the interface. i prefer it either being abstract or in a utils. otherwise implementors will forget to override this

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,47 @@
...
+class Collective(ABC):
+    """Base class for collective functions for training type plugins."""
+
+    @abstractmethod
+    def barrier(self, name: Optional[str] = None, *args, **kwargs) -> None:
+        """Forces all possibly joined processes to wait for each other."""
+
+    @abstractmethod
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        """Broadcasts an object to all processes."""
```
ananthsub, line 29: there's a typehint here we can carry forward here: https://github.com/PyTorchLightning/pytorch-lightning/blob/089ae9b3e82ddc31942e315294e31e48c0a899db/pytorch_lightning/plugins/training_type/training_type_plugin.py#L109-L110

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Any, Optional, Union
+
+import torch
+
+
+class Collective(ABC):
+    """Base class for collective functions for training type plugins."""
+
+    @abstractmethod
+    def barrier(self, name: Optional[str] = None, *args, **kwargs) -> None:
```
ananthsub, line 24: are the *args, **kwargs needed right now?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+from pytorch_lightning.utilities.types import _TPU_AVAILABLE
+
+if _TPU_AVAILABLE:
+    import torch_xla.core.xla_model as xm
+    from torch_xla.core.xla_model import rendezvous
+else:
+    xm, rendezvous = [None] * 4
+
+if _HOROVOD_AVAILABLE:
+    import horovod.torch as hvd
```
ananthsub, line 31: this should raise an exception if horovod isn't available

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+        if not self.is_distributed:
+            return obj
+        buffer = io.BytesIO()
+        torch.save(obj, buffer)
+        data = bytearray(buffer.getbuffer())
+        data_tensor = torch.tensor(data, device=self.root_device, dtype=torch.float)
+        data = xm.all_gather(data_tensor)
+        buffer = io.BytesIO(data.cpu().byte().numpy())
+        obj = torch.load(buffer)
+        return obj
```
ananthsub, line 70: is this copypaste? this shouldn't depend on XLA at all. https://github.com/PyTorchLightning/pytorch-lightning/blob/089ae9b3e82ddc31942e315294e31e48c0a899db/pytorch_lightning/plugins/training_type/horovod.py#L132-L134

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+        self._local_rank = local_rank
+
+    def join(self):
+        """Horovod function that indicates that the rank finished processing data.
+
+        All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread
+        until all ranks join.
+        """
+        if self.on_gpu:
+            hvd.join(self.local_rank)
```
ananthsub, line 52: self.on_gpu and self.local_rank are not initialized. they're prefixed with underscores in the constructor

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+    import horovod.torch as hvd
+
+
+class HorovodCollective(Collective):
+    """Collective interface for Horovod training type plugins."""
+
+    def __init__(
+        self,
+        on_gpu: Optional[bool] = False,
+        local_rank: Optional[int] = 0,
```
ananthsub, line 40: why optional?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+    ):
+        self._on_gpu = on_gpu
+        self._local_rank = local_rank
+
+    def join(self):
```
ananthsub, line 45: is join expected to be exposed publicly? or is it an implementation of the rest of the collective interface?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+        buffer = io.BytesIO()
+        torch.save(obj, buffer)
+        data = bytearray(buffer.getbuffer())
+        data_tensor = torch.tensor(data, device=self.root_device, dtype=torch.float)
+        data = xm.all_gather(data_tensor)
+        buffer = io.BytesIO(data.cpu().byte().numpy())
+        obj = torch.load(buffer)
+        return obj
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
```
ananthsub, line 72: same here: https://github.com/PyTorchLightning/pytorch-lightning/blob/089ae9b3e82ddc31942e315294e31e48c0a899db/pytorch_lightning/plugins/training_type/horovod.py#L174-L188

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,110 @@
...
+from pytorch_lightning.plugins.collective import Collective
+from pytorch_lightning.utilities import _HOROVOD_AVAILABLE
+from pytorch_lightning.utilities.distributed import ReduceOp
+from pytorch_lightning.utilities.types import _TPU_AVAILABLE
+
+if _TPU_AVAILABLE:
+    import torch_xla.core.xla_model as xm
+    from torch_xla.core.xla_model import rendezvous
+else:
+    xm, rendezvous = [None] * 4
```
ananthsub, line 28: there should be no dependency on xla here

pytorch_lightning/plugins/collective/single_device_collective.py
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, Optional, Union
+
+import torch
+
+from pytorch_lightning.plugins.collective import Collective
+
+
+class SingleNodeCollective(Collective):
```
ananthsub, line 21: single device?

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,95 @@
...
+
+        local_reduce set in Plugins.setup() functions
+        """
+        self.local_reduce = local_reduce
+
+    def barrier(self, *args, **kwargs) -> None:
+        if not distributed_available():
+            return
+        if _TORCH_GREATER_EQUAL_1_8 and torch.distributed.get_backend() == "nccl":
+            torch.distributed.barrier(device_ids=self.determine_ddp_device_ids())
```
ananthsub, line 49: determine ddp device ids isn't available on the class

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            torch.distributed.barrier()
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        if not distributed_available():
+            return obj
+        return self.dist.broadcast(obj)
```
ananthsub, line 56: copy/paste issue from plugin?

ananthsub (changes requested):

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: no we don't, will udpate

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: This files wasn't finished...let me resubmit

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: yeah make sense

four4fish: Another RFC : Should I have "distributed_available" as a param for this class instead of call the distributed_available util? It may make the logic in this class more clear, but we don't know distributed_available() when plugin/TorchCollective initialized, so we need update the TorchCollective.distributed_available in setup or setup_environment

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: This one is on purpose. As DP and DDP both will use this class as collective plugin, we want to cover both use cases?

DDP broadcast
```   
def broadcast(self, obj: object, src: int = 0) -> object:
        return self.dist.broadcast(obj)
```

DDPSpawn broadcast
```   
def broadcast(self, obj: object, src: int = 0) -> object:
        if not distributed_available():
            return obj
        return self.dist.broadcast(obj)
```

DP broadcast:
   ```
def broadcast(self, obj: object, src: int = 0) -> object:
        return obj

```

Comment (ananthsub) [reply pytorch_lightning/plugins/collective/torch_collective.py:56]: ah i mean there's no `self.dist` defined here

Comment (four4fish) [reply pytorch_lightning/plugins/collective/torch_collective.py:56]: Ohh, yeah!.. I think we can deprecate LightningDistributed? It's only used by ddp and ddpspawn, and we can directly include the broadcast function here?

File: pytorch_lightning/plugins/collective/collective_plugin.py (hunk 1 above)
Comment (four4fish) [reply pytorch_lightning/plugins/collective/collective_plugin.py:47]: Make sense, I think abstract probably better, as self.reduce is been called in ddp reduce_boolean_decision functions?

File: pytorch_lightning/plugins/collective/torch_collective.py (hunk 13 above)
Comment (ananthsub) [reply pytorch_lightning/plugins/collective/torch_collective.py:56]: yes we should. the deprecation for it should be in a separate PR, but it ought to be succeeded by the collectives here. its one of the sources of collectives mentioned in the issue 😄

File: pytorch_lightning/plugins/collective/horovod_collective.py (hunk 4 above)
Comment (four4fish) [reply pytorch_lightning/plugins/collective/horovod_collective.py:25]: Yeah make sense, current horovod_plugin doesn't raise exception either, how about we fix the two together.

File: pytorch_lightning/plugins/collective/collective_plugin.py [hunk 14]
```diff
@@ -0,0 +1,56 @@
...
+
+        Args:
+            tensor: the tensor to sync and reduce
+            *args: plugin-specific positional arguments
+            **kwargs: plugin-specific keyword arguments
+        """
+
+    @abstractmethod
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
```
ananthsub, line 56: - we shouldn't mention early stopping here
- the world size is also needed

ananthsub (commented):
- for getting something initially working, one PR makes sense, especially to have clear up the API requirements. but once the API requirements are clear, I think this will be easier to review in a stack, especially for the implementations. that'll make it easier so we don't miss anything during the carryover period
- from the gdoc (@yifuwang ) for torch distributed, the process group initialization could be part of the collectives. do you think it makes sense to offer `setup`/`teardown` functions on the collective interface that allow for this? this way, the trainer can know if it was the agent which initialized the global process group vs not (cc @kaushikb11 as it potentially relates to https://github.com/PyTorchLightning/pytorch-lightning/pull/9044)

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: Will add world size to be a param in torch_collective, and setup at ddp setup_distributed() and dp go with default 1

pytorch_lightning/plugins/collective/horovod_collective.py
awaelchli: we don't need to raise at import level, we can raise when the collective gets instantiated.

awaelchli (commented):
can you add these classes to the project.toml for mypy types?
we want newly added classes and files to be fully type checked

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+        local_rank: int = 0,
+    ):
+        self.on_gpu = on_gpu
+        self.local_rank = local_rank
+
+    def join(self) -> None:
+        """Horovod function that indicates that the rank finished processing data.
+
+        All ranks that did not call join() continue to process allreduce operations. This function blocks Python thread
+        until all ranks join.
```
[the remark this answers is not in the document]
awaelchli: join is not in the base class so it should probably go at the bottom?

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,112 @@
...
+        world_size: int = 1,
+    ):
+        """.. note::
+
+        DDP and DDPSpawn sync accross multiple nodes/devices, local_reduce = False
+        DP run reduce in on node, local_reduce = True
+        DDP2 behaves like DP in one node, local_reduce = True
+
+        local_reduce set in Plugins.setup() functions
+        """
```
awaelchli, line 47:
```suggestion
        """
        Note:
            DDP and DDPSpawn sync accross multiple nodes/devices, local_reduce = False
            DP run reduce in on node, local_reduce = True
            DDP2 behaves like DP in one node, local_reduce = True

        local_reduce set in Plugins.setup() functions
        """
```

awaelchli: we use Google style for these sections. 
Also can you make all these fragments full sentences?

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,112 @@
...
+    """Collective interface for DDP, DDPSpawn, DP and DDP2."""
+
+    def __init__(
+        self,
+        local_reduce: bool = False,
+        rank: Optional[int] = None,
+        device: Optional[Union[str, torch.device]] = torch.device("cpu"),
+        device_id: Optional[int] = None,
+        world_size: int = 1,
+    ):
```
awaelchli, line 39:
```suggestion
    ) -> None:
```

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,112 @@
...
+            return
+        if _TORCH_GREATER_EQUAL_1_8 and torch.distributed.get_backend() == "nccl":
+            torch.distributed.barrier(device_ids=self.device_id)
+        else:
+            torch.distributed.barrier()
+
+    def broadcast(self, obj: Any, src: int = 0) -> Any:
+        if not distributed_available():
+            return obj
+        else:
```
awaelchli, line 65: else not necesssary

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
+        if not distributed_available():
+            return
+        if _TORCH_GREATER_EQUAL_1_8 and torch.distributed.get_backend() == "nccl":
+            torch.distributed.barrier(device_ids=self.device_id)
+        else:
+            torch.distributed.barrier()
+
+    def broadcast(self, obj: Any, src: int = 0) -> Any:
```
awaelchli, line 62: type annotation inconsistency, object / Any

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,112 @@
...
+        tensor: Union[torch.Tensor, _METRIC_COLLECTION],
+        group: Optional[Any] = None,
+        reduce_op: Optional[Union[ReduceOp, str]] = "mean",
+    ) -> Union[torch.Tensor, _METRIC_COLLECTION]:
+        """Reduces the given tensor (e.g. across GPUs/processes)
+
+        If local_reduce = True (dp and ddp2), reduces tensor from all local processes.
+
+        If local_reduce = False (ddp, ddpspawning and extentions), reduces a tensor from several distributed processes
+        Args:
```
awaelchli, line 87: careful with spacing! needs empty line before sections
```suggestion
        If local_reduce = False (ddp, ddpspawning and extentions), reduces a tensor from several distributed processes
        
        Args:
```

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+        output = xm.mesh_reduce("reduce", output, sum)
+
+        if isinstance(reduce_op, str) and reduce_op.lower() in ("avg", "mean"):
+            output = output / self.world_size
+
+        return output
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
+        return decision
```
awaelchli, line 99: in several places docstrings are left out. that's fine because they get inherited from the super class. so for consistency, either duplicate all or let them be inherited :)

awaelchli (approved):
I like it very much. 

3/n will be the integration?

pytorch_lightning/plugins/collective/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from pytorch_lightning.plugins.collective.collective_plugin import Collective  # noqa: F401
+from pytorch_lightning.plugins.collective.horovod_collective import HorovodCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.single_device_collective import SingleDeviceCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.torch_collective import TorchCollective  # noqa: F401
+from pytorch_lightning.plugins.collective.tpu_collective import TPUCollective  # noqa: F401
```
ananthsub, line 18: @justusschock @awaelchli @SeanNaren is there a naming convention we should follow for plugins? the style here is different from https://github.com/PyTorchLightning/pytorch-lightning/tree/master/pytorch_lightning/plugins/io

i think we should have or establish a common pattern here to avoid issues like https://github.com/PyTorchLightning/pytorch-lightning/issues/6094

pytorch_lightning/plugins/collective/tpu_collective.py
four4fish: Got you! Will only override when docstring is different and leave the rest inherited

four4fish (commented):
> I like it very much.
> 
> 3/n will be the integration?

Yep, will integrate with ttps and accelerators.

pytorch_lightning/plugins/collective/__init__.py
justusschock: I think we don't have one (yet).

Personally, I think the base/abstract class should contain the `'Plugin'` suffix while the others should not. But that's just personal preference :)

tchaton: Yes, I believe yours is quite neat ! We could rename `torch_plugin.py` to `torch_checkpoint_plugin.py` @SeanNaren

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+
+class HorovodCollective(Collective):
+    """Collective interface for Horovod training type plugins."""
+
+    def __init__(
+        self,
+        on_gpu: Optional[bool] = False,
+        local_rank: int = 0,
+    ) -> None:
+        self.on_gpu = on_gpu
```
tchaton, line 36: Let's raise here if `_HOROVOD_AVAILABLE ` isn't available.

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+        self.local_rank = local_rank
+
+    def join(self) -> None:
```
tchaton, line 39: Should this be private ?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+        elif reduce_op in ("sum", ReduceOp.SUM):
+            reduce_op = hvd.Sum
+        else:
+            raise ValueError(f"unrecognized `reduce_op`: {reduce_op}")
+
+        # sync all processes before reduction
+        self.join()
+        return hvd.allreduce(tensor, op=reduce_op)
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
```
tchaton, line 105: Careful: 

This was implemented by the ParallelPlugin.

```
    def reduce_boolean_decision(self, decision: bool) -> bool:
        decision = torch.tensor(int(decision), device=self.lightning_module.device)
        decision = self.reduce(decision, reduce_op=ReduceOp.SUM)
        decision = bool(decision == self.world_size)
        return decision
```

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+        device: Union[str, torch.device] = torch.device("xla"),
+        root_device: torch.device = torch.device("xla"),
+        world_size: int = 1,
+    ):
+        self.device = device
+        self.root_device = root_device
+        self.world_size = world_size
+
+    @property
+    def is_distributed(self) -> bool:
```
tchaton, line 45: This should be private or exposed on all.

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+
+        output = xm.mesh_reduce("reduce", output, sum)
+
+        if isinstance(reduce_op, str) and reduce_op.lower() in ("avg", "mean"):
+            output = output / self.world_size
+
+        return output
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
```
tchaton, line 97: Same here, reduce_boolean_decision isn't implemented properly. This would result in hanging within ModelCheckpoint.

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+class TPUCollective(Collective):
+    """Collective interface for TPUSpawning training type plugins."""
+
+    def __init__(
+        self,
+        device: Union[str, torch.device] = torch.device("xla"),
+        root_device: torch.device = torch.device("xla"),
+        world_size: int = 1,
+    ):
+        self.device = device
```
tchaton, line 40: Should those properties be private ?

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: Should we raise it in HorovodPlugin then?
I think current plugins has consistent behavior that : doesn't have raise exception when module/device doesn't exist (eg: hvd, tpu, deepspeed), should we do a improvement have raise exception for all of them?

Current deepspeed
```
if _DEEPSPEED_AVAILABLE:
    import deepspeed
```


Current tpu
```
if _TPU_AVAILABLE:
    import torch_xla.core.xla_env_vars as xenv
    import torch_xla.core.xla_model as xm
    import torch_xla.distributed.xla_multiprocessing as xmp
    from torch_xla.core.xla_model import rendezvous
    from torch_xla.distributed.parallel_loader import MpDeviceLoader
else:
    xm, xmp, MpDeviceLoader, rendezvous = [None] * 4
```
If None.function will through exception anyway. Should we add check in each plugin?

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: HorovodPlugin use this function too, in start_training, eval and test

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: This is a really good catch! thanks!

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: This is the logic in used by current ddp and ddpspawn logic, originally defined here
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/distributed/dist.py

Could you guys elaborate more why we don't need it here?

pytorch_lightning/plugins/collective/tpu_collective.py
four4fish: will change to private

pytorch_lightning/plugins/collective/tpu_collective.py
four4fish: It's not set at the initialize time so we need access it outside this class, draft of integration https://github.com/PyTorchLightning/pytorch-lightning/pull/9472/files#diff-3124c14524cdfa032a95acc3f319d9293a0a84c48a52960316ccc03253baa6f5R167-R169

pytorch_lightning/plugins/collective/tpu_collective.py
four4fish: good catch!

pytorch_lightning/plugins/collective/horovod_collective.py
ananthsub: `join` is essentially acting as a `barrier` in the HorovodPlugin. I think we should be calling the `barrier` in the HorovodPlugin, and use `join` as an implementation detail inside the collective.

pytorch_lightning/plugins/collective/__init__.py
four4fish: How about this
1. File Name: base class has *_plugin.py and subclass doesn't. eg: collective_plugin.py and torch_collective.py (+1 @justusschock's suggestion)
2. Base Class Name:  cemal case of the file name WithOut 'plugin' and With first letter capitalized eg: collective_plugin.py -> class Collective; checkpoint_io_plugin.py -> class CheckpointIO
2. Class name: cemal case of the file name With first letter capitalized. eg: torch_collective.py -> class TorchCollective; torch_checkpoint_io_plugin.py -> class TorchCheckpointIO

@ananthsub @tchaton @awaelchli @justusschock @SeanNaren

pytorch_lightning/plugins/collective/horovod_collective.py
tchaton: Yes, join is an implementation detail from Horovod and could be private.

pytorch_lightning/plugins/collective/__init__.py
justusschock: I would add plugin to the name of the baseclass as well (just to make it easier to recognize that it is implementing a plugin interface when subclassing it).

four4fish: address @justusschock's comments
1. File Name: base class has *_plugin.py and subclass doesn't. eg: collective_plugin.py and torch_collective.py (+1 @justusschock's suggestion)
2. Class Name: cemal case of the file name With first letter capitalized eg: collective_plugin.py -> class CollectivePlugin; checkpoint_io_plugin.py -> class CheckpointIOPlugin; torch_collective.py -> class TorchCollective; torch_checkpoint_io_plugin.py -> class TorchCheckpointIO

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Collective(ABC):
+    """Base class for collective functions for training type plugins."""
+
+    @abstractmethod
+    def barrier(self, name: Optional[str] = None) -> None:
+        """Forces all possibly joined processes to wait for each other."""
+
+    @abstractmethod
+    def broadcast(self, obj: object, src: int = 0) -> object:
```
yifuwang, line 30: Looking at the derived classes, this really should be called `broadcast_object`, which has a different semantic from the plain `broadcast` for tensors.

For example, when NCCL `broadcast` returns (or `.wait()` on the async handle returns), it is only guaranteed that the operation is enqueued to a non-default Cuda stream. The main thread will only block when the result (or something that is dependent on the result) is used to produce side effect (e.g. printing the result).

However, since `broadcast_object_list` (or its variant for other accelerator types) involves serialization/deserialization, the main thread will always block before the collective finishes. This can lead to performance implications when `broadcast_object_list` is mis-used to transfer large tensors.

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+        else:
+            obj = [obj]
+            if self.rank != 0:
+                obj = [None] * len(obj)
+            broadcast_object_list(obj, 0, group=_group.WORLD)
+            return obj[0]
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        """Perform a all_gather on all processes."""
+        return all_gather_ddp_if_available(tensor, group=group, sync_grads=sync_grads)
```
yifuwang, line 74: Same as the other comment. We should never silently produce incorrect results.

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+from pytorch_lightning.utilities.distributed import ReduceOp, sync_ddp_if_available
+from pytorch_lightning.utilities.types import _METRIC_COLLECTION
+
+
+class TorchCollective(Collective):
+    """Collective interface for DDP, DDPSpawn, DP and DDP2."""
+
+    def __init__(
+        self,
+        local_reduce: bool = False,
```
yifuwang, line 34: I don't believe this should be here. I think `TorchCollective` should only be used for encapsulating the difference between different distributed backends. The `local_reduce` functionality is not related to `torch.distributed` and should be handled by upper layers.

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+        rank: Optional[int] = None,
+        device: Optional[Union[str, torch.device]] = torch.device("cpu"),
+        device_id: Optional[int] = None,
+        world_size: int = 1,
```
yifuwang, line 38: Since `Collective` doesn't own a pg (a pg can be passed in), it shouldn't maintain a `rank` and a `world_size`. A process can be a member of process groups of different size, and be different ranks in different groups.

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        self.local_reduce = local_reduce
+        self.rank = rank
+        self.device = device
+        self.device_id = device_id
+        self.world_size = world_size
+
+    def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
+        if not distributed_available():
+            return
```
yifuwang, line 56: IMO this should result in a failure. Silently skipping synchronization primitives can lead to issues that are tough to troubleshoot.

pytorch_lightning/plugins/collective/torch_collective.py
yifuwang: If `Collective` is meant to represent a communicator instance whose membership includes all processes that runs Lightning trainer, it should own or be assigned to a fixed reference to a process group.

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+            torch.distributed.barrier()
+
+    def broadcast(self, obj: Any, src: int = 0) -> Any:
+        if not distributed_available():
+            return obj
+        else:
+            obj = [obj]
+            if self.rank != 0:
+                obj = [None] * len(obj)
+            broadcast_object_list(obj, 0, group=_group.WORLD)
```
yifuwang, line 69: The `src` argument is not respected here.

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+import torch
+
+from pytorch_lightning.utilities.distributed import ReduceOp
+
+
+class Collective(ABC):
+    """Base class for collective functions for training type plugins."""
+
+    @abstractmethod
+    def barrier(self, name: Optional[str] = None) -> None:
```
yifuwang, line 26: Named barrier/rendezvous is not supported by all distributed backends we are encapsulating. Maybe `name` shouldn't be an argument in the interface?

four4fish:
@yifuwang Thank you for all the comments! They are all good points, can I set up a time with you and go through some details?

pytorch_lightning/plugins/collective/__init__.py
SeanNaren: Sounds good :)

pytorch_lightning/plugins/collective/collective_plugin.py
tchaton: Should we go through a depreciation process and provide a broadcast_object hook ?

pytorch_lightning/plugins/collective/single_device_collective.py
```diff
@@ -0,0 +1,48 @@
...
+
+from pytorch_lightning.plugins.collective import Collective
+
+
+class SingleDeviceCollective(Collective):
+    """Collective interface for single device training type plugins."""
+
+    def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
+        """Forces all possibly joined processes to wait for each other."""
+        pass
```
tchaton, line 26: pass is not required.

pytorch_lightning/plugins/collective/torch_collective.py
tchaton: Good catch !

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+        Args:
+            tensor: the tensor to sync and reduce
+            group: the process group to gather results from. Defaults to all processes (world)
+            reduce_op: the reduction operation. Defaults to 'mean'/'avg'.
+                Can also be a string 'sum' to calculate the sum during reduction.
+
+        Return:
+            reduced value, except when the input was not a tensor the output remains is unchanged
+        """
+        if self.local_reduce:
```
tchaton, line 97: We could move this within the `DataParellelPlugin` reduce hook

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+    @property
+    def is_distributed(self) -> bool:
+        # HOST_WORLD_SIZE is None outside the xmp.spawn process
+        return os.getenv(xenv.HOST_WORLD_SIZE, None) is not None and self.world_size != 1
+
+    def barrier(self, name: Optional[str] = None) -> None:
+        if self.is_distributed:
+            rendezvous(name)
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
```
tchaton, line 53: Ideally, it would be good to merge broadcast object logic and add support for TPU. I found this implementation to be quite faulty in the past.

pytorch_lightning/plugins/collective/collective_plugin.py
yifuwang: @tchaton yes I think distinguishing `broadcast` and `broadcast_object(_list])` in the `Collective` interface would be nice.

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: will *args: Any, **kwargs: Any be better then?

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: For DP and DDP collective have different behaviors, should we have separate collective for each type then?

yifuwang: IMHO since local reduce is not a behavior offered by the distributed backends we are trying to encapsulate, the logic should just live where it's used.

four4fish: I see, you mean pass it in to specific functions like reduce(), yeah make sense, will do!

pytorch_lightning/plugins/collective/collective_plugin.py
yifuwang: I think it's better to not expose the functionality. IMHO the value of the interface is to offer a consistent semantic across different underlying implementation, so ideally feature wise it captures the lowest common denominator.

four4fish: Yeah, the only place needs name param in barrier is TPU, TPU barrier() is calling [rendezvous]( ()https://pytorch.org/xla/release/1.5/_modules/torch_xla/core/xla_model.html). But base on [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle) we needs include it in the base collective api. Will add comments as you suggested "For backends that don't support named barrier, the argument is ignores and only serves as an annotation"

pytorch_lightning/plugins/collective/collective_plugin.py
bowangbj: nit: s/all/from src to all

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @abstractmethod
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        """Broadcasts an object to all processes."""
+
+    @abstractmethod
+    def all_gather(
+        self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False
+    ) -> Union[List[torch.Tensor], torch.Tensor]:
+        """Perform a all_gather on all processes."""
```
bowangbj, line 37: do you want some cmt to explain semantics of each arg, e.g., group, sync_grads?

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+        """Forces all possibly joined processes to wait for each other."""
+
+    @abstractmethod
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        """Broadcasts an object to all processes."""
+
+    @abstractmethod
+    def all_gather(
+        self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False
+    ) -> Union[List[torch.Tensor], torch.Tensor]:
```
bowangbj, line 36: in which situation output is Tensor vs list[Tensor] ? probably a cmt and use case

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: yeah good point!

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+        """Perform a all_gather on all processes."""
+
+    @abstractmethod
+    def reduce(
+        self,
+        tensor: Union[torch.Tensor, Any],
+        group: Optional[Any] = None,
+        reduce_op: Optional[Union[ReduceOp, str]] = "mean",
+    ) -> Union[torch.Tensor, Any]:
```
bowangbj, line 45: is this all_reduce or just reduce to one rank? 

if former, maybe rename to all_reduce
if later, maybe add dest_rank as a new arg

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+    def all_gather(
+        self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False
+    ) -> Union[List[torch.Tensor], torch.Tensor]:
+        """Perform a all_gather on all processes."""
+
+    @abstractmethod
+    def reduce(
+        self,
+        tensor: Union[torch.Tensor, Any],
+        group: Optional[Any] = None,
```
bowangbj, line 43: nit: does it make sense to make it more specific, pg or process_group ?

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+        reduce_op: Optional[Union[ReduceOp, str]] = "mean",
+    ) -> Union[torch.Tensor, Any]:
+        """Reduces the given tensor (e.g. across GPUs/processes).
+
+        Args:
+            tensor: the tensor to sync and reduce
+            *args: plugin-specific positional arguments
+            **kwargs: plugin-specific keyword arguments
```
bowangbj, line 51: miss reduce_op and group here

args and kwargs -- should they also be part of the API above?

pytorch_lightning/plugins/collective/collective_plugin.py
bowangbj: +1 the cmt looks quite specific. 

what does reduce_boolean really mean here? whats the reduce op given list of boolean? probably add a sentence to explain and add a concrete example.

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional, Union
+
+import torch
+
+from pytorch_lightning.utilities.distributed import ReduceOp
+
```
bowangbj, line 20: should we add a high level cmt description on 
1) what's collective
2) why do we have it here
3) example how to use it?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+if _HOROVOD_AVAILABLE:
+    import horovod.torch as hvd
+
+
+class HorovodCollective(Collective):
+    """Collective interface for Horovod training type plugins."""
+
+    def __init__(
+        self,
+        on_gpu: Optional[bool] = False,
```
bowangbj, line 33: just make it bool ?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+
+        All ranks that did not call join() continue to process allreduce operations. This function blocks the Python
+        thread until all ranks join.
+        """
+        if self.on_gpu:
+            hvd.join(self.local_rank)
+        else:
+            hvd.join()
+
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
```
bowangbj, line 50: should args be consistent with base class?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        if group is not None:
+            raise ValueError("Horovod does not support allreduce using a subcommunicator at this time. Unset `group`.")
+
+        if reduce_op in (None, "avg", "mean"):
+            reduce_op = hvd.Average
+        elif reduce_op in ("sum", ReduceOp.SUM):
+            reduce_op = hvd.Sum
+        else:
+            raise ValueError(f"unrecognized `reduce_op`: {reduce_op}")
```
bowangbj, line 99: nit: also mention which reduce_ops are supported

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # sync all processes before reduction
+        self.join()
+        return hvd.allreduce(tensor, op=reduce_op)
+
+    def reduce_boolean_decision(self, decision: bool) -> bool:
+        """Reduce the early stopping decision across all processes."""
+        return decision
```
bowangbj, line 107: need cmt to explain why we return it directly

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+        """Reduces a tensor from several distributed processes to one aggregated tensor.
+
+        Args:
+            tensor: the tensor to sync and reduce
+            group: the process group to gather results from. Defaults to all processes (world)
+            reduce_op: the reduction operation. Defaults to 'mean'/'avg'.
+                Can also be a string 'sum' to calculate the sum during reduction.
+
+        Return:
+            reduced value, except when the input was not a tensor the output remains is unchanged
```
bowangbj, line 89: do u need a logic to check tensor's type? and return if tensor is not of type Tensor

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+    @abstractmethod
+    def all_gather(
+        self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False
+    ) -> Union[List[torch.Tensor], torch.Tensor]:
+        """Perform a all_gather on all processes."""
+
+    @abstractmethod
+    def reduce(
+        self,
+        tensor: Union[torch.Tensor, Any],
```
bowangbj, line 42: s/tensor/tenosr_or_obj

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,56 @@
...
+        group: Optional[Any] = None,
+        reduce_op: Optional[Union[ReduceOp, str]] = "mean",
```
bowangbj, line 44: possible to restrict reduce_op to only enum ReduceOp ?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,107 @@
...
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
+        if distributed_available():
+            self.join()
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        obj = hvd.broadcast_object(obj, src)
+        return obj
+
+    def all_gather(
+        self, result: Union[torch.Tensor], group: Optional[Any] = dist_group.WORLD, sync_grads: bool = False
```
bowangbj, line 59: api signature diffs from that from base class.

pytorch_lightning/plugins/collective/single_device_collective.py
```diff
@@ -0,0 +1,48 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, Optional, Union
+
+import torch
+
+from pytorch_lightning.plugins.collective import Collective
+
+
```
bowangbj, line 20: probably you want to rename it to NoOpCollective 

and explain why and which situation we need it.

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+from pytorch_lightning.utilities import _TORCH_GREATER_EQUAL_1_8
+from pytorch_lightning.utilities.apply_func import apply_to_collection
+from pytorch_lightning.utilities.distributed import all_gather_ddp_if_available, distributed_available
+from pytorch_lightning.utilities.distributed import group as _group
+from pytorch_lightning.utilities.distributed import ReduceOp, sync_ddp_if_available
+from pytorch_lightning.utilities.types import _METRIC_COLLECTION
+
+
+class TorchCollective(Collective):
+    """Collective interface for DDP, DDPSpawn, DP and DDP2."""
```
bowangbj, line 30: nit: the interface does not have to couple with DDP, DDPSpawn, etc.

probably sth like

Collective interfaces for PyTorch. Mainly used by DDP, DDPSpawn, DP and DDP2.

pytorch_lightning/plugins/collective/torch_collective.py
bowangbj: +1 , unclear whats local_reduce :)

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+        local_reduce: bool = False,
+        rank: Optional[int] = None,
+        device: Optional[Union[str, torch.device]] = torch.device("cpu"),
+        device_id: Optional[int] = None,
+        world_size: int = 1,
+    ) -> None:
+        """
+        Note:
+            DDP and DDPSpawn sync accross multiple nodes/devices, local_reduce = False
+            DP run reduce in on node, local_reduce = True
```
bowangbj, line 43: i heard DP is not performant.  Should we move all the usage to DDP instead of DP?

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+from pytorch_lightning.utilities.types import _METRIC_COLLECTION
+
+
+class TorchCollective(Collective):
+    """Collective interface for DDP, DDPSpawn, DP and DDP2."""
+
+    def __init__(
+        self,
+        local_reduce: bool = False,
+        rank: Optional[int] = None,
```
bowangbj, line 35: probably int is enough? since it wont be null?

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: It's the logic in current dp and ddp2

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,113 @@
...
+            return obj[0]
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        """Perform a all_gather on all processes."""
+        return all_gather_ddp_if_available(tensor, group=group, sync_grads=sync_grads)
+
+    def reduce(
+        self,
+        tensor: Union[torch.Tensor, _METRIC_COLLECTION],
+        group: Optional[Any] = None,
```
bowangbj, line 79: do we need _METRIC_COLLECTION ?

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: There is use cases for DP, for example user has to use older pytorch version which break for DDP so they have to go with DP

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: for dp use case

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: Yeah in ddp2 metrics could be passed into reduce

bowangbj (commented):
Thanks Siyu putting it together so quickly! Looks great directional-wise, added a few cmts inline.

pytorch_lightning/plugins/collective/torch_collective.py
yifuwang: > I see, you mean pass it in to specific functions like reduce(), yeah make sense, will do!

No I actually meant the logic for handling local reduce should be moved out of `Collective`. Because (1) it's not a functionality provided by the underlying distributed backend (2) it's not an operation that all members of the process group participate (3) it's not supported in other implementations

pytorch_lightning/plugins/collective/torch_collective.py
ananthsub: > i heard DP is not performant. Should we move all the usage to DDP instead of DP?

This has to be initiated by torch distributed. Lightning recommends all multi-gpu training to use DDP over DP, but as long as pytorch supports DP, Lightning will too

pytorch_lightning/plugins/collective/torch_collective.py
ananthsub: @yifuwang since it previously came up around whether the trainer should teardown the global process group - do you think the collective should be responsible for the setup/teardown of this? or do you think that logic should remain in the training type plugin, while the `Collective` can be called with any pg/rank/world size passed in?

pytorch_lightning/plugins/collective/torch_collective.py
ananthsub: the if case above returns from the function. therefore we can remove the `else` here and save a level of indentation

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+from pytorch_lightning.utilities.distributed import ReduceOp
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+
+if _TPU_AVAILABLE:
+    import torch_xla.core.xla_env_vars as xenv
+    import torch_xla.core.xla_model as xm
+    from torch_xla.core.xla_model import rendezvous
+
+
+class TPUCollective(Collective):
```
ananthsub, line 31: naming nit: TPU or TorchXLA?

pytorch_lightning/plugins/collective/tpu_collective.py
```diff
@@ -0,0 +1,99 @@
...
+        world_size: int = 1,
+    ):
+        self.device = device
+        self.root_device = root_device
+        self.world_size = world_size
+
+    @property
+    def is_distributed(self) -> bool:
+        # HOST_WORLD_SIZE is None outside the xmp.spawn process
+        return os.getenv(xenv.HOST_WORLD_SIZE, None) is not None and self.world_size != 1
```
ananthsub, line 47: why do we need is_distributed here? couldn't the training type plugin use the single-device/no-op plugin instead of having this handling here?

pytorch_lightning/plugins/collective/torch_collective.py
yifuwang: > since it previously came up around whether the trainer should teardown the global process group - do you think the collective should be responsible for the setup/teardown of this?


It'll still depend on whether or `Lightning` set up the global process group. It's not something `Lightning` can control. More and more model parallel use cases need to set up the global pg before even initializing the trainer (e.g. `torchrec`). The use of ` ShardedTensor` depends on a pg as well, which likely means that it'll be the case too for the `torch.distributed` version of `FSDP`.

So we can do what we are doing today: only shutdown the global pg if we created it.

> or do you think that logic should remain in the training type plugin, while the Collective can be called with any pg/rank/world size passed in?


If somehow a caller already has access to the raw process group, I don't think it makes any sense to go through `Collective` anymore, whose sole purpose is to hide the difference between different communication backends. In that case, they should just use `torch.distributed` APIs directly.

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: This is also used by DP when distributed is not available

pytorch_lightning/plugins/collective/tpu_collective.py
tchaton: `TorchXLA` is interesting. Does XLA GPU have different collectives than pure PyTorch ?

pytorch_lightning/plugins/collective/tpu_collective.py
tchaton: @kaushikb11

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: Name is not needed here.

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: Discussed and resolved offline. Due to current lightning structure, we will keep as is and add more comments

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: Does lightning support multiple process group now?
How about we address in a different issue `support multiple process group in lightning`

pytorch_lightning/plugins/collective/tpu_collective.py
four4fish: yeah, I'm also curious why we have this in current TPUSpawning implementation. Is that possible to have TPUSpawning running on single process?

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: Will update in integration, as caller side pass group= X now

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: there is use case user will pass str in

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: There is DP use case reduce on the same node

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: same as above

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: collective intent to hide behavior of different backend and have a unified api. Different collective also have different broadcast, broadcast_object, broadcast_object_list names, I feel it's hard to choose a specific one.
For example:
1. torch collective : version < 1.8 using broadcast
                              version >= 1.8 using broadcast_object_list
2. horovod: broadcast_object  (official document mentioned it's mainly for broadcast optimizer.state_dict(), for model param has broadcast_parameters? https://horovod.readthedocs.io/en/stable/api.html#module-horovod.torch)

 Does  broadcast vs broadcast_object vs broadcast_object_list name really matter here? How about we add more detailed comments?

pytorch_lightning/plugins/collective/single_device_collective.py
four4fish: it's abstract class in base class, need define it here. Without pass python will show error

pytorch_lightning/plugins/collective/tpu_collective.py
four4fish: How about we address in a different PR. Keep this one mainly as-is implementation and no breaking change

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: Could you elaborate more how can we add this to DataParellelPlugin reduce hook?
we use this param to distringush dp vs ddp for now

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: It's in class level comments

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,108 @@
...
+            DDP and DDPSpawn sync accross multiple nodes/devices, local_reduce = False
+            DP run reduce in on node, local_reduce = True
+            DDP2 behaves like DP in one node, local_reduce = True
+
+        local_reduce set in Plugins.setup() functions
+        """
+        self.local_reduce = local_reduce
+        self.rank = rank
+        self.device = device
+        self.device_id = device_id
```
ananthsub, line 53: are these meant to be exposed outside the class? if not let's mark these as protected

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
+        if not distributed_available():
```
ananthsub, line 56: prior comments here mentioned that checking for distributed_available could lead to subtle bugs in case torch distributed isn't initialized, but the caller expects it to be

pytorch_lightning/plugins/collective/torch_collective.py
```diff
@@ -0,0 +1,108 @@
...
+            DDP2 behaves like DP in one node, local_reduce = True
+
+        local_reduce set in Plugins.setup() functions
+        """
+        self.local_reduce = local_reduce
+        self.rank = rank
+        self.device = device
+        self.device_id = device_id
+
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
```
ananthsub, line 55: typehint is inconsistent with parent class
please add these module paths here to ensure we start with this new code being strongly typed: https://github.com/PyTorchLightning/pytorch-lightning/blob/4bebf82f271fb577ed41831403551f0bff97651b/pyproject.toml#L62-L86

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,74 @@
...
+        Args:
+            name: a str pass into barrier. Only torch xla respect this param
+        """
+
+    @abstractmethod
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        """Broadcasts an object to all processes.
+
+        Args:
+            obj: the object to broadcast
```
ananthsub, line 42: use a typevar to ensure that the type of the input `obj` is the same as the return value: https://github.com/PyTorchLightning/pytorch-lightning/blob/4bebf82f271fb577ed41831403551f0bff97651b/pytorch_lightning/plugins/training_type/training_type_plugin.py#L108

pytorch_lightning/plugins/collective/collective_plugin.py
ananthsub: > Does broadcast vs broadcast_object vs broadcast_object_list name really matter here? How about we add more detailed comments?

as @yifuwang points out, there are performance considerations with broadcasting large tensors depending on the underlying library call used. we could either:
- offer dedicated `broadcast` vs `broadcast_object_list` interfaces on the collective
- special-case the implementation of `broadcast` on collective implementations based on the type of the input passed in. 

of these i prefer the first option since it's the most explicit

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,74 @@
...
+
+from pytorch_lightning.utilities.distributed import ReduceOp
+
+
+class CollectivePlugin(ABC):
+    """Interface for collective functions for training type plugins.
+
+    Lightning collective supports communications between multiple processes and multiple nodes, provides routines such
+    as barrier, broadcast, all_gather, and reduce, reduce
+    """
```
ananthsub, line 27: given this is brand new, imo we should include a comment that this API is experimental/in-beta and subject to change. once this is baked in more, we can declare it stable

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,74 @@
...
+import torch
+
+from pytorch_lightning.utilities.distributed import ReduceOp
+
+
+class CollectivePlugin(ABC):
+    """Interface for collective functions for training type plugins.
+
+    Lightning collective supports communications between multiple processes and multiple nodes, provides routines such
+    as barrier, broadcast, all_gather, and reduce, reduce
```
ananthsub, line 26: `reduce` repeated twice

pytorch_lightning/plugins/collective/collective_plugin.py
```diff
@@ -0,0 +1,74 @@
...
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional, Union
+
+import torch
+
+from pytorch_lightning.utilities.distributed import ReduceOp
+
+
+class CollectivePlugin(ABC):
+    """Interface for collective functions for training type plugins.
```
ananthsub, line 23: i don't think we ought to mention training type plugins here. it should be as self-contained as possible, though within Lightning these will be managed by the training type plugins.

A good test would be that someone could use these utilities independently of the rest of lightning if they wanted to. that means it'll be easier for us to integrate

pytorch_lightning/plugins/collective/horovod_collective.py
ananthsub: IMO we could raise it here too. its a good check in case someone uses these collectives independently of the training type (for whatever reason)

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            hvd.join()
+
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
+        if distributed_available():
+            self._join()
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        obj = hvd.broadcast_object(obj, src)
+        return obj
```
ananthsub, line 56:
```suggestion
        return hvd.broadcast_object(obj, src)
```

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,94 @@
...
+        All ranks that did not call join() continue to process allreduce operations. This function blocks the Python
+        thread until all ranks join.
+        """
+        if self.on_gpu:
+            hvd.join(self.local_rank)
+        else:
+            hvd.join()
+
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
+        if distributed_available():
```
ananthsub, line 51: do we need the distributed available check?

pytorch_lightning/plugins/collective/horovod_collective.py
```diff
@@ -0,0 +1,94 @@
...
+    ) -> List[torch.Tensor]:
+        if process_group is not None and process_group != dist_group.WORLD:
+            raise ValueError("Horovod does not support allgather using a subcommunicator at this time. Unset `group`.")
+
+        if len(result.shape) == 0:
+            # Convert scalars to single dimension tensors
+            result = result.reshape(1)
+
+        # sync and gather all
+        self._join()
```
ananthsub, line 69: should we be caling self._join or self.barrier ?

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: >offer dedicated broadcast vs broadcast_object_list interfaces on the collective

Current broadcast usage is 
1. in timer.py -> broadcast(should stop)
2. in model_checkpoint.py -> broadcast(skip_time), broadcast(ckpt_path), broadcast(exists)
3. in Trainer properties -> broadcast(dirpath)

I don't think having both broadcast and broadcast_object_list two API fit our use cases because:
1. Which one should we use in the three callers mentioned above?
2. Not all backend has broadcast and broadcast_object/broadcast_object_list. Like hovorod only have broadcast_object vs TPUspawn only have broadcast lightning defined without using any backend

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: I feel tensor is a better param name

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: Will create a issue add raise exception for xla and horovod in training type plugin too

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: distributed_available() is checking `torch.distributed.is_available() and torch.distributed.is_initialized() or tpu_distributed()`, is it better to check distributed_available() before wait?

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: yeah, it will be set in setup or setup distributed instead of initialization time.

pytorch_lightning/plugins/collective/horovod_collective.py
four4fish: I think they are the same, which one do you prefer?

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: Torch distributed will return tensor, and horovod will return a list of tensor. Is it better to have base class more generic without too much details about the subclass?

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: This is not necessary true. 
1. Horovod broadcast_object will return the object that was broadcast from the root_rank
2. for pytorch version< 1.8, torch collective using broadcast, which returns `Async work handle, if async_op is set to True. None, if not async_op or if not part of the group` 
3. for pytorch version .=1.8 torch collective using broadcast, which returns `None. If rank is part of the group, object_list will contain the broadcasted objects from src rank.`

pytorch_lightning/plugins/collective/collective_plugin.py
four4fish: sorry I was wrong. Horovod has broadcast (totally 7 different types of broadcast)

pytorch_lightning/plugins/collective/torch_collective.py
four4fish: distributed_available is checking torch.distributed.is_initialized()?
```
def distributed_available() -> bool:
    return torch.distributed.is_available() and torch.distributed.is_initialized() or tpu_distributed()
```

pytorch_lightning/plugins/collective/horovod_collective.py
ananthsub: In the current implementation one is checking for if distributed is available and one isn't

ananthsub:
@four4fish at this point, I don't think offering a new collective base class that unifies all these disparate backends is the right approach. It offers new API surface area in Lightning but can't guarantee that the behavior from the different underlying libraries won't diverge or have different semantics.

IMO this should be addressed at the torch distributed level, especially for XLA usage. In the meantime, Lightning can get by calling the existing utilities as-is inside of the corresponding strategy class.

carmocca:
@ananthsub are you advocating to drop this proposal entirely?

is there any specific reason for the change of direction?

four4fish:
> @ananthsub are you advocating to drop this proposal entirely?
> 
> is there any specific reason for the change of direction?

@carmocca @awaelchli  I have synced up with Ananth offline. Not changing direction or drop this plan, we are thinking about switch the implementation order a bit.

1. Unified collective behavior should be provided from pytorch side, now there is effort going on in pytorch to moving to this direction. We will keep you guys updated.
2. I think precision plugin refactor and accelerator refactor in #10416 is higher priority towards stable API 1.6.

How about let's have this after the step 2 and 3.  (I'm working on the PR for #7324 right now)
What do you guys think?  I can also rebase and update this first.

PR #9418, opened 2021-09-09
Type the Loop base class as generic

carmocca:
## What does this PR do?

Signal that the return type of `run` is the same as the return type of `on_run_end`.

Minimal example:

```python
from dataclasses import dataclass
from typing import TypeVar, Generic

T = TypeVar("T")


class Base(Generic[T]):
    def run(self) -> T:
        return self.end()

    def end(self) -> T:
        ...


@dataclass
class Foo:
    ...


class Child(Base[Foo]):
    def end(self) -> Foo:
        ...
```

without this, when `result = self.a_loop.run(...)`, mypy doesn't know the type of `result` even if its annotated in the `on_run_end` signature.

None

## Before submitting

## PR review

### DIFF
--- a/pytorch_lightning/loops/base.py
+++ b/pytorch_lightning/loops/base.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from abc import ABC, abstractmethod
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Generic, Optional, TypeVar
 
 from deprecate import void
 from torchmetrics import Metric
@@ -24,8 +24,10 @@ from pytorch_lightning.trainer.progress import BaseProgress, Progress
 from pytorch_lightning.utilities.apply_func import apply_to_collection
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 
+T = TypeVar("T")  # the output type of `run`
 
-class Loop(ABC):
+
+class Loop(ABC, Generic[T]):
     """Basic Loops interface. All classes derived from this must implement the following properties and methods:
 
         * :attr:`done` (property): Condition to break the loop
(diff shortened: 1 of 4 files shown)


pytorch_lightning/loops/base.py
```diff
@@ -24,8 +24,10 @@
 from pytorch_lightning.utilities.apply_func import apply_to_collection
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 
+T = TypeVar("T")  # the output type of `run`
 
-class Loop(ABC):
+
+class Loop(ABC, Generic[T]):
```
awaelchli, line 30: what if we hypothetically also wanted to express that the output of on advance is the input to on_advance_end but not the same as for on_run_end/run? Would you inherit here from another generic?

awaelchli (approved):

pytorch_lightning/loops/base.py
carmocca: This seems to work:

```python
from typing import TypeVar, Generic

A = TypeVar("A")
B = TypeVar("B")


class Base(Generic[A, B]):
    def run(self) -> A:
        return self.on_run_end()

    def advance(self) -> B:
        ...

    def on_advance_end(self, b: B) -> None:
        ...

    def on_run_end(self) -> A:
        ...


class Foo:
    ...


class Bar:
    ...


class Child(Base[Foo, Bar]):
    def end(self) -> Foo:
        ...

    def on_advance_end(self, b: Bar) -> None:
        ...
```

ananthsub (approved):

tchaton (approved):
LGTM !

kaushikb11 (approved):

PR #9442, opened 2021-09-10
insert metrics for last checkpoint

magehrig:
Insert metrics for last checkpoint. 

When a custom CHECKPOINT_NAME_LAST is defined with metrics, they will not be filled.

e.g. with epoch = 0, step =0 and val/metric = 1.00
`CHECKPOINT_NAME_LAST = 'last_epoch={epoch:03d}-step={step}-val_metric={val/metric:.2f}'`
will result in a filename
`'last_epoch=epoch=000-step=step=0-val_metric=val/metric=1.00'`
which will then create a directory
`'last_epoch=epoch=000-step=step=0-val_metric=val`
with a checkpoint name of
`metric=1.00`

This PR fixes this issue by filling the checkpoint metrics if the user desires this.

### DIFF
--- a/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/pytorch_lightning/callbacks/model_checkpoint.py
@@ -655,7 +655,7 @@ class ModelCheckpoint(Callback):
         if not self.save_last:
             return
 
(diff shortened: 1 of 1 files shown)


magehrig:
I am closing this PR because it's probably better to avoid metrics in the last checkpoint anyway

PR #8967, opened 2021-08-18
Support post-localSGD in Lightning DDP plugin

SciPioneer:
Integrate a new feature post-localSGD from PyTorch Distributed into Lightning. We have seen good speedup on our benchmark with parity on accuracy.

Proposal: https://github.com/pytorch/pytorch/issues/59699

This PR has already been reviewed by @ananthsub and his teammates in FB internally.

**Usage:**
To enable post-localSGD, expect the user to provide two args:
1. A DDP comm hook (`post_localSGD.post_localSGD_hook`), which is constructed by `_ddp_comm_hook` and `_ddp_comm_state` in DDP plugin;
2. A new optional field, `_model_averaging_period`. The default value is set as 16 if post-localSGD hook is specified but no model averaging period is provided.

**Requirements:**
1. PyTorch Version >= 1.9.0
2. not _IS_WINDOWS

PR #9446, opened 2021-09-10
Replace PostLocalSGDOptimizer with a dedicated model averaging component

wayi1:
## What does this PR do?

This is an improvement of https://github.com/PyTorchLightning/pytorch-lightning/pull/8967.

Replace `PostLocalSGDOptimizer` by a dedicated model averaging component that can run after `optimizer_step` in lightning.py.

The previous implementation can cause data races and fail the training for some cases. For example, if the data loading phase also involves allreduce (e.g., for checking if the input data stream has reached the end), and such allreduce can be kicked off during `optimizer.step()`. Therefore, the implementation in this PR can fix such issue.

Proposal: https://github.com/pytorch/pytorch/issues/59699

There are two options to enable post-local SGD by using vanilla PyTorch:

1.  Post-LocalSGD Comm Hook + Periodic Model Averager
https://github.com/pytorch/pytorch/blob/master/torch/distributed/algorithms/model_averaging/averagers.py#L48-L78

This option requires 1-line addition (such as `averager.average_parameters(model.parameters())`) in the training loop.

2. Post-LocalSGD Comm Hook + Post-LocalSGD Optimizer
https://github.com/pytorch/pytorch/blob/master/torch/distributed/optim/post_localSGD_optimizer.py#L15-L48

This option does not need any code change in the training loop. I tried this option earlier. However, one limitation is that if there is another phase such as data loading that also runs some communication like allreduce and overlaps with `optimizer_step()`, then this can cause potential data race and fail the training. Since such concurrent phase like data loading is out of the control of model averaging, I didn't choose this implementation.

None.

## Before submitting

## PR review


SciPioneer:
It's ready to review.

pytorch_lightning/core/lightning.py
```diff
@@ -1562,6 +1572,9 @@ def optimizer_step(
         """
         optimizer.step(closure=optimizer_closure)
 
+        if _TORCH_GREATER_EQUAL_1_10 and self._model_averager is not None:
+            self._model_averager.average_parameters(self.model.parameters())
+
```
ananthsub, line 1577: in this case, what happens if the LightningModule contains multiple modules for training?

SciPioneer: Wouldn't 'self.model.parameters()' here recursively contain all the child modules? Or are you suggesting that we should still read parameters from `self.optimizers()`?

awaelchli (approved):

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -321,43 +322,30 @@ def _register_ddp_hooks(self) -> None:
...
-                optimizer_class=optim_class,
-                averager=averager,
-                **optimizer.defaults,
-            )
-            optimizers[x] = post_localSGD_optimizer
-            del optimizer
-        trainer = self.lightning_module.trainer
-        trainer.optimizers = optimizers
-        trainer.convert_to_lightning_optimizers()
+        self.lightning_module._model_averager = averagers.PeriodicModelAverager(
```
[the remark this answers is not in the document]
SciPioneer: Yes, this will be released as a prototype feature in PyTorch 1.10, and we only have docstring for the documentation purpose so far.

See: https://github.com/pytorch/pytorch/search?q=PeriodicModelAverager

pytorch_lightning/core/lightning.py
SciPioneer: Updated similar to `PostLocalSGDOptimizer`, which only average parameters conditionally;
https://github.com/pytorch/pytorch/blob/master/torch/distributed/optim/post_localSGD_optimizer.py?l#L81-L85

pytorch_lightning/core/lightning.py
```diff
@@ -1562,6 +1572,16 @@ def optimizer_step(
         """
         optimizer.step(closure=optimizer_closure)
 
+        # TODO: Refactor the following code out of this file.
```
tchaton, line 1575: Personally, I am not a huge fan of the comm hooks leaking within the Plugin. Ideally, there should be their own isolated components.

tchaton: Here is an example: https://github.com/PyTorchLightning/pytorch-lightning/blob/490c40a8be339afd08dcfdcdd658db6b0be4671b/pytorch_lightning/plugins/ddp_comm_hooks/allreduce_invalid.py

SciPioneer: Agree that the current implementation is not a very elegant way, and I also think that ideally `lightning.py` file should be untouched in this PR.

I had a discussion with @ananthsub yesterday, who would later let someone do some follow-up refactoring. Sorry that I am not familiar with lightning internals, and actually I am leaving Facebook soon, so I really hope that this PR can be unblocked, and some follow-up works including refactoring and experimentations can be done concurrently.

Thanks for the understanding!

pytorch_lightning/core/lightning.py
```diff
@@ -1562,6 +1572,16 @@ def optimizer_step(
...
+        # CAVEAT: If the user overrides this method, then model averaging will be disabled by default.
+        if _TORCH_GREATER_EQUAL_1_10 and self._model_averager is not None:
+            for opt in self.optimizers(use_pl_optimizer=False):
```
awaelchli, line 1578: this hook is getting called per optimizer. we still loop over all optimizers?

SciPioneer: Yes, this hook is called per optimizer step, and we loop over all the optimizers. Certainly we can cache all the target parameters somewhere and only run such looping once. However, I don't think we should include such optimization in this PR:

1) In practice, a trainer usually only has a very small number of optimizers, so the cost of looping here should be trivial. We probably don't want to do over-optimization unless if necessary, and even if we want, it's better to do this incrementally in a separate PR.
2) More important, for some use cases the number of parameters that compute gradients and get averaged can change over time. For example, some base layers may be frozen gradually during pre-train. In this case, it's much safer to run such loop every time.

wayi1:
Close this obsolete PR to favor the replacement https://github.com/PyTorchLightning/pytorch-lightning/pull/12378 from @daniellepintz

PR #9344, opened 2021-09-06
feat: Add ModelSummary Callback

kaushikb11:
## What does this PR do?

There are two parts to #9043 

1. Introduce `ModelSummary` callback
2. Support flag `weights_summary` as Union[str, bool] = True and mark the string values as deprecated (follow up PR)

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

Issue #7037, opened 2021-04-15
Improve typing coverage


tchaton:
**Please make one PR per file!**

The list of pending files is can be found below here

https://github.com/Lightning-AI/lightning/blob/master/pyproject.toml#L39-L43

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7

Borda:
Duplicate to #5023

akihironitta:
https://github.com/PyTorchLightning/pytorch-lightning/pull/7035#issue-615794299

> Open for discussion: Should we also type the tests? I don't think we have to!

I don't think we need type annotations in the tests, but we should run mypy to check if variables/functions of the package are correctly typed. What do you think?

stancld:
After the chat with Eden, I'll tackle lightning.utilities :)

stancld:
Hello @tchaton, I noticed new information about making one PR per file was added to the first comment. As I've been dealing with some mypy issues, my PR #7218 has become quite heavy. Is it, therefore, appropriate to close this PR, though already with a few reviews given, and open new ones separately for each file? Thanks :)

otaj:
@kingjuno, you're more than welcome to help! We are closing this issue in favor of #13445  in order to promote better visibility, please feel free to comment there.

Issue #9043, opened 2021-08-23
[RFC] Deprecate `weights_summary` off the Trainer constructor


ananthsub:
## Proposed refactoring or deprecation

- Introduce a new ModelSummary callback, which calls `summarize`: https://github.com/PyTorchLightning/pytorch-lightning/blob/8a931732ae5135e3e55d9c7b7031d81837e5798a/pytorch_lightning/utilities/model_summary.py#L437-L439
- Deprecate `weights_summary` off the Trainer constructor

### Motivation

We are auditing the Lightning components and APIs to assess opportunities for improvements:
- https://github.com/PyTorchLightning/pytorch-lightning/issues/7740
- https://docs.google.com/document/d/1xHU7-iQSpp9KJTjI3As2EM0mfNHHr37WZYpDpwLkivA/edit#

This is a followup to https://github.com/PyTorchLightning/pytorch-lightning/issues/8478 and https://github.com/PyTorchLightning/pytorch-lightning/discussions/9006

Why do we want to remove this from the core trainer logic?
- We need a way for users to customize more of the inputs to the model summary over time without affecting the trainer API. Today, changes to the model summary API also require changes in the core trainer (e.g. the addition of `max_depth` ). This gives model summarization more room to grow without cascading changes elsewhere. 
- Users may want to configure this summarization for different points of execution. For instance, calling this at the start of `trainer.fit()`, `trainer.validate()`, `trainer.test()` or `trainer.predict()`. Right now, this is hardcoded to be run only during `fit()`. 
- Users may want to customize where they save the summary. Right now, it's printed to stdout, but this could also be useful to save to a file or upload to another service for tracking the run.
- The current implementation runs on global rank 0 only in order to avoid printing out multiple summary tables. However, running this on rank 0 will break for model parallel use cases that require communication across ranks. This can lead to subtle failures if `example_input_array` is set as a property on the LightningModule. For instance, a model wrapped with FSDP will break because parameters need to be all-gathered across layers across ranks.
- In case the LightningModule leverages PyTorch LazyModules, users may want to generate this summary only after the first batch is processed in order to get accurate parameter estimations. Estimates of parameter sizes with lazy modules would be misleading.

### Pitch

A callback in Lightning naturally fits this extension purpose. It generalizes well across lightning modules, has great flexibility for when it can be called, and allows users to customize the summarization logic (e.g. integrate other libraries more easily). 
- https://github.com/tyleryep/torchinfo
[...]

PR #9447, opened 2021-09-11
fix mypy typing for model summary

awaelchli:
## What does this PR do?

Part of https://github.com/PyTorchLightning/pytorch-lightning/issues/7037
Follow up to #9344 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/callbacks/model_summary.py
+++ b/pytorch_lightning/callbacks/model_summary.py
@@ -22,7 +22,7 @@ the name, type and number of parameters for each layer.
 
 """
 import logging
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Tuple
 
 import pytorch_lightning as pl
 from pytorch_lightning.callbacks.base import Callback
@@ -62,7 +62,7 @@ class ModelSummary(Callback):
 
     @staticmethod
     def summarize(
-        summary_data: List[List[Union[str, List[str]]]],
+        summary_data: List[Tuple[str, List[str]]],
         total_parameters: int,
         trainable_parameters: int,
         model_size: float,
--- a/pytorch_lightning/callbacks/model_summary.py
+++ b/pytorch_lightning/callbacks/model_summary.py
@@ -22,7 +22,7 @@ the name, type and number of parameters for each layer.
 
 """
 import logging
-from typing import List, Optional, Union, Tuple
+from typing import List, Tuple
 
 import pytorch_lightning as pl
 from pytorch_lightning.callbacks.base import Callback
--- a/pytorch_lightning/callbacks/model_summary.py
+++ b/pytorch_lightning/callbacks/model_summary.py
@@ -46,11 +46,11 @@ class ModelSummary(Callback):
         >>> trainer = Trainer(callbacks=[ModelSummary(max_depth=1)])
     """
 
-    def __init__(self, max_depth: Optional[int] = 1):
+    def __init__(self, max_depth: int = 1) -> None:
         self._max_depth: int = max_depth
 
     def on_pretrain_routine_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        if trainer.is_global_zero and self._max_depth is not None and not trainer.testing:
+        if trainer.is_global_zero and self._max_depth and not trainer.testing:
             model_summary = summarize(pl_module, max_depth=self._max_depth)
 
             summary_data = model_summary._get_summary_data()
@@ -68,5 +68,4 @@ class ModelSummary(Callback):
         model_size: float,
     ) -> None:
         summary_table = _format_summary_table(total_parameters, trainable_parameters, model_size, *summary_data)
-
         log.info("\n" + summary_table)
--- a/pytorch_lightning/utilities/model_summary.py
+++ b/pytorch_lightning/utilities/model_summary.py
@@ -297,22 +297,22 @@ class ModelSummary:
                 model(input_)
         model.train(mode)  # restore mode of module
 
-    def _get_summary_data(self) -> List[List[Union[str, List[str]]]]:
+    def _get_summary_data(self) -> List[Tuple[str, List[str]]]:
         """Makes a summary listing with:
 
         Layer Name, Layer Type, Number of Parameters, Input Sizes, Output Sizes, Model Size
         """
         arrays = [
-            [" ", list(map(str, range(len(self._layer_summary))))],
-            ["Name", self.layer_names],
-            ["Type", self.layer_types],
-            ["Params", list(map(get_human_readable_count, self.param_nums))],
+            (" ", list(map(str, range(len(self._layer_summary))))),
+            ("Name", self.layer_names),
+            ("Type", self.layer_types),
+            ("Params", list(map(get_human_readable_count, self.param_nums))),
         ]
         if self._model.example_input_array is not None:
-            arrays.append(["In sizes", self.in_sizes])
-            arrays.append(["Out sizes", self.out_sizes])
+            arrays.append(("In sizes", self.in_sizes))
+            arrays.append(("Out sizes", self.out_sizes))
 
-        return arrays  # type: ignore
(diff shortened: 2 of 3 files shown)


pytorch_lightning/utilities/model_summary.py
```diff
@@ -310,9 +312,9 @@ def _get_summary_data(self):
             arrays.append(["In sizes", self.in_sizes])
             arrays.append(["Out sizes", self.out_sizes])
 
-        return arrays
+        return arrays  # type: ignore
```
awaelchli, line 315: `pytorch_lightning/utilities/model_summary.py:315: error: Incompatible return value type (got "List[List[Sequence[str]]]", expected "List[List[Union[str, List[str]]]]")`

why???

carmocca: I'm not exactly sure about why it converts it to `Sequence`, however, the return annotation doesn't seem right to me.

`List[Union[str, List[str]]]` suggests that it can be of any length and that `str` and `List[str]` can be in any order, however, the actual format represents just two elements where the first is a `str` and the second a `List[str]`.

For that reason, I'd annotate this return type as `List[Tuple[str, List[str]]]` and update the code to this stricter requirement:

```python
        arrays = [
            (" ", [str(i) for i in range(len(self._layer_summary))]),
            ("Name", self.layer_names),
            ("Type", self.layer_types),
            ("Params", [get_human_readable_count(n) for n in self.param_nums]),
        ]
        if self._model.example_input_array is not None:
            arrays.append(("In sizes", self.in_sizes))
            arrays.append(("Out sizes", self.out_sizes))
        return arrays
```

note: list comprehensions are more pythonic than `list(map(str` :D

awaelchli: great, thanks that works. I won't change the rest of the code, I want this PR to be just about typing.

pytorch_lightning/callbacks/model_summary.py
```diff
@@ -46,11 +46,11 @@ class ModelSummary(Callback):
         >>> trainer = Trainer(callbacks=[ModelSummary(max_depth=1)])
     """
 
-    def __init__(self, max_depth: Optional[int] = 1):
+    def __init__(self, max_depth: int = 1) -> None:
         self._max_depth: int = max_depth
 
     def on_pretrain_routine_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        if trainer.is_global_zero and self._max_depth is not None and not trainer.testing:
+        if trainer.is_global_zero and self._max_depth and not trainer.testing:
```
ananthsub, line 53: this can be a separate PR, but creating the summary should not be gated by `is_global_zero` here. This will break for DeepSpeed/FSDP or any model with parameters on different devices in the forward pass if the LightningModule sets an `example_input_array`. `forward` needs to act like a collective, so it must run on all ranks at the same time. it's one of the points listed in https://github.com/PyTorchLightning/pytorch-lightning/issues/9043

It is safer to generate the summary data unconditionally on all ranks and then let the logging of the summarization be determined by the particular rank id

pytorch_lightning/callbacks/model_summary.py
```diff
@@ -62,11 +62,10 @@ def on_pretrain_routine_start(self, trainer: "pl.Trainer", pl_module: "pl.Lightn
 
     @staticmethod
     def summarize(
-        summary_data: List[List[Union[str, List[str]]]],
+        summary_data: List[Tuple[str, List[str]]],
```
ananthsub, line 65: @kaushikb11 - if this set of arguments is going to grow, should we use a dataclass instead? that way the `summarize` signature doesn't change, and we can still add more arguments over time w/o breaking backwards compatibility or making everything optional.

pytorch_lightning/callbacks/model_summary.py
```diff
@@ -46,11 +46,11 @@ class ModelSummary(Callback):
         >>> trainer = Trainer(callbacks=[ModelSummary(max_depth=1)])
     """
 
-    def __init__(self, max_depth: Optional[int] = 1):
+    def __init__(self, max_depth: int = 1) -> None:
         self._max_depth: int = max_depth
```
awaelchli, line 50: reviewer: here in the new callback we don't need to type it None, because it does not mix with the deprecated "mode" argument that the `ModelSummary` has.

pytorch_lightning/utilities/model_summary.py
```diff
@@ -85,19 +85,20 @@ def _register_hook(self) -> Optional[RemovableHandle]:
             A handle for the installed hook, or ``None`` if registering the hook is not possible.
         """
 
-        def hook(module, inp, out):
+        def hook(_: nn.Module, inp: Any, out: Any) -> None:
             if len(inp) == 1:
                 inp = inp[0]
             self._in_size = parse_batch_shape(inp)
             self._out_size = parse_batch_shape(out)
+            assert self._hook_handle is not None
```
awaelchli, line 93: reviewer: the outer function here registers the hook, so the handle must not be None. and if it was, it would mean the hook would not run. so this assert is good.

pytorch_lightning/utilities/model_summary.py
```diff
@@ -207,21 +208,22 @@ def __init__(self, model, mode: Optional[str] = None, max_depth: Optional[int] =
...
         self._layer_summary = self.summarize()
         # 1 byte -> 8 bits
-        # TODO: how do we compute precisin_megabytes in case of mixed precision?
+        # TODO: how do we compute precision_megabytes in case of mixed precision?
         precision = self._model.precision if isinstance(self._model.precision, int) else 32
         self._precision_megabytes = (precision / 8.0) * 1e-6
 
     @property
     def named_modules(self) -> List[Tuple[str, nn.Module]]:
+        mods: List[Tuple[str, nn.Module]]
```
awaelchli, line 217: mypy doesn't get it, so I had to put it here

pytorch_lightning/utilities/model_summary.py
```diff
@@ -450,14 +457,16 @@ def summarize(
 
     # temporary mapping from mode to max_depth
     if max_depth is None:
-        if mode in ModelSummaryMode.supported_types():
+        if mode is None:
+            model_summary = ModelSummary(lightning_module, max_depth=1)
+        elif mode in ModelSummaryMode.supported_types():
```
awaelchli, line 462: reviewer: had to change the if-else this structure here. Otherwise we could have ended up with a reference to the variable `model_summary` before assignment

pytorch_lightning/utilities/model_summary.py
```diff
@@ -430,8 +437,8 @@ def _is_lazy_weight_tensor(p: Tensor) -> bool:
 
 
 def summarize(
-    lightning_module: "pl.LightningModule", mode: Optional[str] = "top", max_depth: Optional[int] = None
-) -> Optional[ModelSummary]:
+    lightning_module: "pl.LightningModule", mode: Optional[str] = None, max_depth: Optional[int] = None
+) -> ModelSummary:
```
awaelchli, line 441: this should have been the correct way in the original PR. mode is deprecated, we want to set it to None and then detect if someone sets it so we can warn appropriately.

pytorch_lightning/callbacks/model_summary.py
awaelchli: yes I agree. only guard the logging instruction

awaelchli: also the condition  not trainer.testing here is always true, the hook here only runs for training :)

ananthsub (approved):

carmocca (approved):

pytorch_lightning/callbacks/model_summary.py
kaushikb11: @ananthsub The idea makes sense but I am not sure what additional arguments we will be planning to add in the future.

kaushikb11 (approved):

PR #9460, opened 2021-09-11
Changed max_steps to default to -1 (vs. `None)

### LINKED ISSUE #8818: Allow disabling automatic stopping during fitting
## 🚀 Feature
Currently if neither `max_epochs` nor `max_steps` aren't set, lightning defaults to using `max_epochs = 1000`. However, in some situations the user actually doesn't want `max_epochs` or `max_steps` to be set, and automatically stopping after `1000` epochs isn't wanted. It would be great if there were a way to specify that no epochs should be set.

### Motivation
I was running a very large training job over the weekend with a very large dataset. Because the dataset is so large, I set the number of batches per epoch to be very small (relative to the size of the dataset) so that logging would occur more frequently. I set `max_epochs` and `max_steps` to be `None` because I believed this would disable automatic stopping, on Monday when I checked on the model again it had exited early after only a couple of hours.

### Pitch
Have some way to specify that automatic stopping should be disabled. Having to hard-code a large number like `2**1000` isn't the most elegant (especially compared to how elegant the rest of Lightning is).

### Alternatives
The user could pass `float("inf")` as `max_epochs` to disable stopping if not using multiple GPUs. However, if using multiple GPUs they would need to use a very large integer instead (ex. `2**1000`).

If using a float for `max_epochs`, you will get the following error:
```
Training: 0it [00:00, ?it/s]Traceback (most recent call last):
  File "/home/eric/.../train.runfiles/train.py", line 371, in <module>
    main(sys.argv[1:])
  File "/home/eric/.../train.runfiles/train.py", line 362, in main
    trainer.fit(model, data_module)
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/trainer/trainer.py", line 458, in fit
    self._run(model)
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/trainer/trainer.py", line 756, in _run
    self.dispatch()
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/trainer/trainer.py", line 797, in dispatch
    self.accelerator.start_training(self)
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/accelerators/accelerator.py", line 96, in start_training
    self.training_type_plugin.start_training(trainer)
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/plugins/training_type/training_type_plugin.py", line 144, in start_training
    self._results = trainer.run_stage()
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/trainer/trainer.py", line 807, in run_stage
    return self.run_train()
  File "/home/eric/.../train.runfiles/pypi__pytorch_lightning_python3_deps/pytorch_lightning/trainer/trainer.py", line 861, in run_train
    epochs = range(self.current_epoch, self.max_epochs) if self.max_epochs else count(self.current_epoch)
TypeError: 'float' object cannot be interpreted as an integer
```

### Additional context
https://pytorch-lightning.slack.com/archives/CRBLFHY79/p1627425271133200

EricWiener:
## What does this PR do?
This PR addresses [this comment](https://github.com/PyTorchLightning/pytorch-lightning/issues/8818#issuecomment-907825247). It changes `max_steps` to default to `-1` and no longer allows `None` to be specified for `max_steps`.

Fixes #8818

Yes. You can no longer specify `max_steps = None`.

Not sure what needs to be done w/ regards to this.

## Before submitting

- [x ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)

## PR review

## Did you have fun?
Always


pytorch_lightning/trainer/trainer.py
```diff
@@ -382,10 +382,19 @@ def __init__(
...
+
+        elif max_epochs < -1:
+            # Allow max_epochs to be zero, since this will be handled by fit_loop.done
+            raise MisconfigurationException(
+                f"`max_epochs` must be a positive integer or -1. You passed in {max_epochs}."
+            )
+
         fit_loop = FitLoop(
             min_epochs=(1 if (min_epochs is None and min_steps is None and max_time is None) else min_epochs),
-            max_epochs=(1000 if (max_epochs is None and max_steps is None and max_time is None) else max_epochs),
```
awaelchli, line 388: at this point it is guaranteed an integer, so we can assume that in the fit loop signature (see my suggestion above)

pytorch_lightning/trainer/trainer.py
```diff
@@ -382,10 +382,19 @@ def __init__(
...
-        # max_epochs won't default to 1000 if max_steps/max_time are specified (including being set to -1).
+        if max_epochs is None:
+            # max_epochs won't default to 1000 if max_steps/max_time are non-default values.
+            max_epochs = 1000 if (max_steps == -1 and max_time is None) else -1
+
+        elif max_epochs < -1:
+            # Allow max_epochs to be zero, since this will be handled by fit_loop.done
+            raise MisconfigurationException(
+                f"`max_epochs` must be a positive integer or -1. You passed in {max_epochs}."
+            )
```
awaelchli, line 393: let's keep this in the fit loop

pytorch_lightning/trainer/trainer.py
```diff
@@ -122,7 +122,7 @@ def __init__(
         accumulate_grad_batches: Union[int, Dict[int, int]] = 1,
         max_epochs: Optional[int] = None,
         min_epochs: Optional[int] = None,
-        max_steps: Optional[int] = None,
+        max_steps: int = -1,
```
ananthsub, line 125: this is a breaking change in case, for whatever reason, someone is passing `max_steps=None` explicitly into the trainer constructor. 

we could leave this as Optional[int] = -1
and if someone is passing this as `None`, we log a deprecation warning, and default to using -1 instead.

pytorch_lightning/trainer/trainer.py
carmocca: this was marked as resolved but doesn't look like it - this check is already in the `FitLoop.__init__`

pytorch_lightning/trainer/trainer.py
```diff
@@ -382,10 +382,26 @@ def __init__(
         self.slurm_connector = SLURMConnector(self)
         self.tuner = Tuner(self)
 
-        # max_epochs won't default to 1000 if max_steps/max_time are specified (including being set to -1).
+        if max_steps is None:
+            rank_zero_deprecation(
+                "Setting `max_steps = None` is deprecated in v1.5 and will be removed in v1.7."
+                " Use `max_steps = -1` instead."
+            )
+            max_steps = -1
```
carmocca, line 390: Do this in the `TrainingBatchLoop.__init__`?

pytorch_lightning/trainer/trainer.py
```diff
@@ -122,7 +122,7 @@ def __init__(
         accumulate_grad_batches: Union[int, Dict[int, int]] = 1,
         max_epochs: Optional[int] = None,
         min_epochs: Optional[int] = None,
-        max_steps: Optional[int] = None,
+        max_steps: Optional[int] = -1,
```
carmocca, line 125: I think it's better for us to remove the `Optional` now, so it's easier for users to know they should change the value

```suggestion
        max_steps: int = -1,
```

pytorch_lightning/loops/fit_loop.py
```diff
@@ -156,8 +137,8 @@ def done(self) -> bool:
         is reached.
         """
         # TODO(@awaelchli): Move track steps inside training loop and move part of these condition inside training loop
-        stop_steps = FitLoop._is_max_limit_enabled(self.max_steps) and self.global_step >= self.max_steps
-        stop_epochs = FitLoop._is_max_limit_enabled(self.max_epochs) and self.current_epoch >= self.max_epochs
+        stop_steps = self.max_steps != -1 and self.global_step >= self.max_steps
```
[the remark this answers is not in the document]
carmocca: @awaelchli do you think this indirection is justified for such a simple check?

what if the logic changes slightly? or maybe a property now checks `>` instead of `>=`. The utility becomes useless and it's something else to deprecate.

For example, in `checkpoint_connector` we cannot use it.

carmocca (commented):
Looking great! Thanks for your work ❤️

pytorch_lightning/loops/fit_loop.py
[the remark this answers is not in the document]
awaelchli: I don't care, you can choose :) 

But I disagree that we need to deprecate these utilities. First of all they are all protected, second they are under `loops.utilities` so clearly not meant for any other purpose than loops.

pytorch_lightning/trainer/trainer.py
EricWiener: Whoops sorry I thought it got added to my last commit. Looked like its still staged

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -18,7 +18,7 @@
 from pytorch_lightning import loops  # import as loops to avoid circular imports
 from pytorch_lightning.loops.batch import TrainingBatchLoop
 from pytorch_lightning.loops.optimization.closure import ClosureResult
-from pytorch_lightning.loops.utilities import _prepare_dataloader_iter
+from pytorch_lightning.loops.utilities import _prepare_dataloader_iter, is_max_limit_reached
```
tchaton, line 21: Let's make is_max_limit_reached private.

pytorch_lightning/loops/utilities.py
```diff
@@ -164,3 +164,16 @@ def _block_parallel_sync_behavior(trainer: "pl.Trainer", block: bool = True) ->
...
+    """Check if the limit has been reached (if enabled).
+
+    Args:
+        current (int): the current value
+        maximum (int): the maximum value (or -1 to disable limit)
+
+    Returns:
+        bool: whether the limit has been reached
+    """
+    return maximum != -1 and current >= maximum
```
tchaton, line 179: Neat !

pytorch_lightning/trainer/trainer.py
```diff
@@ -273,9 +273,9 @@ def __init__(
             min_epochs: Force training for at least these many epochs. Disabled by default (None).
                 If both min_epochs and min_steps are not specified, defaults to ``min_epochs = 1``.
 
-            max_steps: Stop training after this number of steps. Disabled by default (None). If ``max_steps = None``
+            max_steps: Stop training after this number of steps. Disabled by default (-1). If ``max_steps = -1``
                 and ``max_epochs = None``, will default to ``max_epochs = 1000``. To disable this default, set
```
tchaton, line 277:
```suggestion
                and ``max_epochs = None``, will default to ``max_epochs = 1000``. To enable infinite training, set
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -382,10 +382,20 @@ def __init__(
...
 
-        # max_epochs won't default to 1000 if max_steps/max_time are specified (including being set to -1).
+        if max_steps is None:
+            rank_zero_deprecation(
+                "Setting `max_steps = None` is deprecated in v1.5 and will be removed in v1.7."
+                " Use `max_steps = -1` instead."
+            )
+            max_steps = -1
+
+        if max_epochs is None:
```
tchaton, line 392: same here. The Trainer init should be the cleanest part of the codebase ;)

tchaton (commented):
Great work ! Almost there.

pytorch_lightning/trainer/trainer.py
EricWiener: This is needed in order to initialize the `FitLoop` (it isn't error checking). Is it okay to leave it?

pytorch_lightning/loops/fit_loop.py
```diff
@@ -30,17 +33,27 @@ class FitLoop(Loop):
...
+        self,
+        min_epochs: int = 0,
+        max_epochs: Optional[int] = None,
+        max_steps: Optional[int] = -1,
+        max_time: Optional[Union[str, timedelta, Dict[str, int]]] = None
+    ):
         super().__init__()
-        # Allow max_epochs or max_steps to be zero, since this will be handled by fit_loop.done
-        if max_epochs and max_epochs < -1:
+        if max_epochs < -1:
```
ananthsub, line 47: this will raise an error if `max_epochs` is None. 
```suggestion
        if max_epochs is not None and max_epochs < -1:
```

carmocca: Or:

```python
if max_epochs is None:
    max_epochs = 1k
elif max_epochs < -1:
    fail() 
```

which is the same pattern we have for `max_steps` below

pytorch_lightning/loops/fit_loop.py
```diff
@@ -104,7 +117,13 @@ def max_steps(self) -> int:
     def max_steps(self, value: int) -> None:
         """Sets the maximum number of steps (forwards to epoch_loop)"""
         # TODO(@awaelchli): This setter is required by debugging connector (fast dev run), should be avoided
-        if value and value < -1:
+        if value is None:
+            rank_zero_deprecation(
```
carmocca, line 121: A deprecation test is missing

Check the tests in https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/deprecated_api/test_remove_1-7.py

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -38,7 +39,13 @@ def __init__(self, min_steps: int, max_steps: int):
...
         self.min_steps: int = min_steps
 
-        if max_steps and max_steps < -1:
+        if max_steps is None:
+            rank_zero_deprecation(
+                "Setting `max_steps = None` is deprecated in v1.5 and will be removed in v1.7."
+                " Use `max_steps = -1` instead."
+            )
+            max_steps = -1
+        elif max_steps < -1:
```
ananthsub, line 48: is the error message below wrong? what if `max_steps=0` ?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -197,15 +197,12 @@ def restore_loops(self) -> None:
...
-            FitLoop._is_max_limit_enabled(self.trainer.max_epochs)
-            and self.trainer.max_epochs is not None
-            and self.trainer.current_epoch > self.trainer.max_epochs
-        ):
-            raise MisconfigurationException(
-                f"You restored a checkpoint with current_epoch={self.trainer.current_epoch},"
-                f" but you have set Trainer(max_epochs={self.trainer.max_epochs})."
-            )
+        if not (self.trainer.max_epochs is None or self.trainer.max_epochs == -1):
+            if self.trainer.current_epoch > self.trainer.max_epochs:
```
ananthsub, line 201: nit: merge these ifs into one conditional to avoid a level of nesting

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -17,13 +17,14 @@
 
 from pytorch_lightning import loops  # import as loops to avoid circular imports
 from pytorch_lightning.loops.batch import TrainingBatchLoop
-from pytorch_lightning.loops.optimization.closure import OutputResult
-from pytorch_lightning.loops.utilities import _prepare_dataloader_iter
+from pytorch_lightning.loops.optimization.closure import ClosureResult, OutputResult
```
ananthsub, line 20: this looks like a bad rebase which is causing the tests to ail

pytorch_lightning/loops/epoch/training_epoch_loop.py
EricWiener: `max_steps = 0` is fine and just results in training immediately ending

pytorch_lightning/trainer/connectors/checkpoint_connector.py
EricWiener: Yeah I didn't like this either and ended up doing this to make mypy happy + the auto-formatting with casting was really ugly. I just tried again and it's working (not sure what I was thinking earlier) and will push that

pytorch_lightning/loops/epoch/training_epoch_loop.py
EricWiener: yep I made a mistake when merging master in

pytorch_lightning/loops/epoch/training_epoch_loop.py
awaelchli: what Ananth meant is that the formulation "positive integer or -1" is excluding the zero, but the 0 may be allowed. To be exact, we could say "must be a non-negative integer or -1".

Issue #8731, opened 2021-08-05
[RFC] Deprecate the _epoch_end hooks


ananthsub:
Lightning has had some recent issues filed around these hooks:
- `training_epoch_end`
- `validation_epoch_end`
- `test_epoch_end`
- `predict_epoch_end`

Examples:
- https://github.com/PyTorchLightning/pytorch-lightning/issues/8603
- https://github.com/PyTorchLightning/pytorch-lightning/issues/8657
- https://github.com/PyTorchLightning/pytorch-lightning/issues/8670 
- https://github.com/PyTorchLightning/pytorch-lightning/issues/8479

Do we really need these on the LightningModule? Why don't users track these outputs directly inside the LightningModule?

Asking users to do this can offer some benefits:
- less API surface area for the trainer reduces risk of bugs like this. these bugs hurt user trust because the control flow isn't visible to end users, so removing this class of bugs can have a disproportionate benefit to user trust
- the trainer doesn't take responsibility of data it doesn't directly use. this simplifies the loop internals
- less "framework magic" means more readable user code because this tracking is explicit
- because the tracking is explicit, the responsibility of testing also falls to users
- we avoid API confusion: for instance, when should users implement something in `training_epoch_end` vs `on_train_epoch_end` ? this can improve the onboarding experience (one less class of hooks to learn about, only 1 way to do things). 
- it can also improve performance: if users implement something in `training_epoch_end` and don't use `outputs`, the trainer accumulates results needlessly which wastes memory. This is not clearly visible to the user either.

Cons:
- (marginally) more boilerplate code in LightningModules. for instance, users would need to pay attention to resetting the results in `on_{train/val}_epoch_start` unless they want to accumulate results across epochs. 

@PyTorchLightning/core-contributors

_Originally posted by @ananthsub in https://github.com/PyTorchLightning/pytorch-lightning/discussions/8690_

ananthsub:
one thing that will emerge is the hook call order between callbacks and the lightning module. logging something in the module's `on_train_epoch_end`  needs to happen first for it to be used in the callback's `on_train_epoch_end` hook. this would be a change compared to today's behavior. this is required for items like logging a value at the end of the epoch in the module, and using that key/value as a monitor for callbacks like model checkpointing or early stopping.

so this will be related to https://github.com/PyTorchLightning/pytorch-lightning/issues/8506

carmocca:
[...]

PR #9490, opened 2021-09-13
reduce loop structure leakage into the `TrainingEpochLoop`

carmocca:
## What does this PR do?

This refactor aims at simplifying the output processing of our nested loop structure. The processing mainly revolves around the nested list structure that the `training_epoch_end(self, outputs)` hook would receive. Until now, the processing was scattered in different places and loops assumed knowledge of each other's output format. We refactor the following components:

- each loop obtains strict typing of its output list
- `OptimizerLoop`: Now returns a dict, indexed by the optimizer_idx that was active
- `TrainingBatchLoop`: mostly unchanged, only typing updates
- `TrainingEpochLoop`: No longer assumes knowledge of how many optimizers were running in OptimizerLoop. It only maintains a list of batch outputs and accumulates them in a list (if required).

As an intermediate step, we modify the processing in the `TrainingEpochLoop` such that the old format that the `training_epoch_end` hook accepts is still respected. No breaking changes intended. 
However, the "old" format of the output structure does not correspond to the actual loop structure. The choice of how the array list is layed out to the user is an arbitrary one. This unintuitive output format only matters when we have multiple optimizers or `truncated_bptt_steps > 1` and in the future we would like to standardize this such that the majority of our output processing can be completely removed. 

co-author: @awaelchli

None user-facing intended

## Before submitting

## PR review


pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -266,76 +263,66 @@ def _should_accumulate(self) -> bool:
...
 
-    def _track_epoch_end_reduce_metrics(
-        self, epoch_output: List[List[STEP_OUTPUT]], batch_end_outputs: STEP_OUTPUT
-    ) -> None:
-        """Adds the batch outputs to the epoch outputs and prepares reduction."""
-        hook_overridden = is_overridden("training_epoch_end", self.trainer.lightning_module)
-        if not hook_overridden:
-            return
+    @staticmethod
+    def _swap_optimizers_and_tbptt(tbptt_outputs: _BATCH_OUTPUTS_TYPE, n_opt: int) -> List[List[Dict[str, Any]]]:
```
tchaton, line 267: Do we want to move all tbptt utilties to their own folder, same for Lighting hooks ?

carmocca: > Do we want to move all tbptt utilties to their own folder

No because this case is a particularity of the `training_epoch_end(outputs)` format, not of `tbptt` itself. For that reason, this PR keeps this processing logic encapsulated in these `staticmethods` which ideally would be removed in the future.

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -146,18 +145,16 @@ def advance(self, *args: Any, **kwargs: Any) -> None:
...
             self.update_lr_schedulers("epoch", update_plateau_schedulers=False)
 
-        batch_end_outputs = [opt_idx_out for opt_idx_out in batch_output.training_step_output if len(opt_idx_out)]
-        processed_batch_end_outputs = self._prepare_outputs(batch_end_outputs, batch_mode=True)
-
-        # hook
-        self.trainer.call_hook("on_train_batch_end", processed_batch_end_outputs, batch, self.batch_idx, 0)
+        if is_overridden("on_train_batch_end", self.trainer.lightning_module):
+            batch_end_outputs = self._prepare_outputs_training_batch_end(batch_output.outputs)
+            self.trainer.call_hook("on_train_batch_end", batch_end_outputs, batch, self.batch_idx, 0)
```
awaelchli, line 150: I think we need to call this without the overridden check, because the checkpoint callback for example relies on this. do you remember why we did this?

carmocca: Ahh good eye.

We did this to avoid the computation when it's not overridden

But forgot that `call_hook` calls everything.

I can try and address this with #9029 - we can undo the check here.

carmocca: It's unfortunate that `ModelCheckpoint` uses `on_train_batch_end` since it doesn't use the outputs at all. This would be one pro for #8731

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -268,65 +265,76 @@ def _should_accumulate(self) -> bool:
...
+        This processing is necessary because the format of the inputs to the ``training_epoch_end`` hook does not
+        match the loop structure and because empty dimensions are squeezed. This could break with loop customization.
         """
-        processed_outputs = []
-        for opt_outputs in outputs:
-            # handle an edge case where an optimizer output is the empty list
-            if len(opt_outputs) == 0:
+        # `batch_outputs` (plural) is the same as `epoch_end_output` (singular)
+        if not batch_outputs:
+            return {}
```
awaelchli, line 307: we return a dict here but I don't remember why. the type hint says we can but I don't think that's correct.

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -268,65 +275,83 @@ def _should_accumulate(self) -> bool:
...
-        outputs: List[List[List[OutputResult]]], batch_mode: bool
-    ) -> Union[List[List[List[Dict]]], List[List[Dict]], List[Dict], Dict]:
-        """Extract required information from batch or epoch end results.
-
-        Args:
-            outputs: A 3-dimensional list of ``OutputResult`` objects with dimensions:
-                ``[optimizer outs][batch outs][tbptt steps]``.
-
-            batch_mode: If True, ignore the batch output dimension.
+    def _prepare_outputs_training_batch_end(
```
awaelchli, line 279: @carmocca this code here is now very similar to the one for epoch_end. The only difference is that we have one dimension less and that the final sanitation is different. 

What do you prefer? Merge the functions into one or keep as is?

carmocca: I'd keep them separate, in case the format for a hook changes

awaelchli: sounds good

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -211,24 +216,28 @@ def on_run_end(self) -> None:
...
+        if is_overridden("training_epoch_end", model) and self._outputs:
+            epoch_end_outputs = self._prepare_outputs_training_epoch_end(
+                self._outputs,
+                automatic=model.automatic_optimization,
+                num_optimizers=len(self.trainer.optimizers),
+            )
             # check that the dataloader/iterator produced a batch
-            if processed_outputs:
+            # FIXME: still necessary?
+            if epoch_end_outputs:
```
awaelchli, line 227: @carmocca should we make a separate PR for this? otherwise the diff here gets destroyed :)

carmocca: For what exactly? The variable rename? The diff looks okay to me

awaelchli: the fixme, which I interpret as "is this if statement still necessary".

carmocca: ah sure, we can check that in a separate PR

carmocca:
```suggestion
            if epoch_end_outputs:
```

awaelchli (approved):

pytorch_lightning/loops/batch/training_batch_loop.py
```diff
@@ -112,18 +115,17 @@ def advance(self, batch, batch_idx):
         # let logger connector extract current batch size
         self.trainer.logger_connector.on_train_split_start(batch_idx, split_idx, split_batch)
 
+        # choose which loop will run the optimization
         if self.trainer.lightning_module.automatic_optimization:
-            # in automatic optimization, hand over execution to the OptimizerLoop
             optimizers = _get_active_optimizers(self.trainer.optimizers, self.trainer.optimizer_frequencies, batch_idx)
```
tchaton, line 120: We can make get_active_optimizers a static_method ?

tchaton: Users might want to override to implement custom scheduling strategy.

tchaton (approved):
LGTM ! Great work !

pytorch_lightning/loops/batch/training_batch_loop.py
tchaton: Ignore it if you want as not fully relevant to this PR.

carmocca: It could be a good idea. I guess it was put in utilities to avoid traversing the loops for it because the `EpochLoop`, `BatchLoop` and `FinetuniningCallback` all use it.

cc @awaelchli 

But definitely not in this PR

awaelchli: Yes, as @carmocca said I moved `_get_active_optimizers` out of the loops because it was used in different loops and defined on the TrainingEpochLoop. But if it becomes a method, IMO it should be in the class that owns the optimizers

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -381,6 +394,71 @@ def _save_loggers_on_train_batch_end(self) -> None:
...
+    """Calculates the max size in each dimension of a jagged (non-rectangular) nested list of lists.
+
+    Example::
+        >>> _get_max_shape([[], [[1], [2]], []])
+        [3, 2, 1]
+    """
+    dimensions = defaultdict(int)
+    for level, length in _get_dimensions(array):
+        dimensions[level] = max(dimensions[level], length)
+    return [value for _, value in sorted(dimensions.items())]
```
rohitgr7, line 456:
```suggestion
    return sorted(dimensions.values())
```

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -381,6 +394,71 @@ def _save_loggers_on_train_batch_end(self) -> None:
...
+    """Converts an optimizer dict to a list in which the key of the dict determines the position of the element.
+
+    Example::
+        >>> _convert_optim_dict({0: {"loss": 0.0}, 2: {"loss": 0.2}}, num_optimizers=3)
+        [{'loss': 0.0}, None, {'loss': 0.2}]
+    """
+    return [outs[opt_idx] if opt_idx in outs else None for opt_idx in range(num_optimizers)]
+
+
+def _recursive_unpad(nested: List[Any], value: Optional[Any] = None) -> List:
```
rohitgr7, line 408: should nested be marked as Any since in recursive call there is a possibility of not getting a list??

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -381,6 +394,71 @@ def _save_loggers_on_train_batch_end(self) -> None:
...
+        >>> _get_max_shape([[], [[1], [2]], []])
+        [3, 2, 1]
+    """
+    dimensions = defaultdict(int)
+    for level, length in _get_dimensions(array):
+        dimensions[level] = max(dimensions[level], length)
+    return [value for _, value in sorted(dimensions.items())]
+
+
+def _iterate_nested_array(array: List[Any], index: Tuple = ()) -> Generator:
```
rohitgr7, line 459: same ques here

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: This isn't exactly equivalent. We want to sort the keys but keep the values

Whereas your suggestions sorts and keeps the values

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: This calls `len(array)` so I think it's expected to be a List

pytorch_lightning/loops/epoch/training_epoch_loop.py
rohitgr7: oh! okay!

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: Technically yes, done with `overload`

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -381,6 +394,81 @@ def _save_loggers_on_train_batch_end(self) -> None:
...
+
+
+@overload
+def _recursive_unpad(nested: Any, value: Optional[Any] = None) -> Any:
+    ...
+
+
+@overload
+def _recursive_unpad(nested: List[Any], value: Optional[Any] = None) -> List[Any]:
+    ...
```
rohitgr7, line 415: sorry for nerding out 😅 , but what does it do? never seen this before.

rohitgr7 (approved):
LG!

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: https://adamj.eu/tech/2021/05/29/python-type-hints-how-to-use-overload/

rohitgr7: this looks like a good blog to follow .. thanks 👍

Issue #9130, opened 2021-08-26
Fault-tolerant checklist


awaelchli:
## 🚀 Fault-tolerant training progress tracker

Fault-tolerant training in PL can be activated by setting the environment variable. 

```
PL_FAULT_TOLERANT_TRAINING=1
```

[Description of the architecture around fault-tolerant training
](https://docs.google.com/document/d/1pvhyNg17pm7-ECMTpgpZiJf6sbypeuaB8QsSP79n6k0/edit?usp=sharing)

- [x] progress tracking in loops
- [x] progress restart of loops
- [ ] capturing state of iterable datasets
   - [x] non-random state, multiple workers
   - [ ] random state
   - [x] sampler inside of iterable dataset
   - [ ] batch sampler
   - [ ] testing of items above
   
- [ ] capturing state of map-style dataset
   - [x] single worker global random state (#8950)
   - [x] multiple workers global random state (#8950)
   - [ ] local generators
   - [ ] random sampler state, after updates in [pytorch 1.10](https://github.com/pytorch/pytorch/pull/63026)
   - [ ] testing of items above (partially done in #8950)

- [ ] mid epoch restart
   - [ ] failure on mid training epoch loop
      - [x] failure on mid training epoch loop with 1 optimizer
      - [x] failure on mid training epoch loop with multiple optimizers  #9537 
      - [ ] failure between optimizer step and scheduler step

   - [x] failure on mid validation epoch loop: #9563, #9681
      - [x] reset on epoch end validation with 1 dataloaders
      - [x] reset on epoch end validation with multiple dataloaders
      - [ ] reset on validation using val_check_interval: https://github.com/PyTorchLightning/pytorch-lightning/pull/9563/files#diff-c2e3dba8e80f865d593f0310cdf80eaeccdb80c9a250f7eda9edd466b7710874R989-R995

- [ ] epoch end restart
   - [ ] failure on training epoch loop end
      - [ ] failure between optimizer step and scheduler step
       - [ ] failure before logging metrics
   - [ ] no failure on training epoch end

- [ ] reloading multiple train dataloaders with individual random states (see #8950 for todos in added tests)
- [ ] Fault-tolerant training with multiple dataloaders (#11349)
- [x] parametrize the important tests in `test_auto_restart.py` with num_workers > 0 and run as part of slow tests #9086  
- [x] end-user guide - how to use and description of limitations  #9256 
[...]

PR #9484, opened 2021-09-13
move state extraction for CaptureMapDataset

awaelchli:
## What does this PR do?

Simplifies the `_capture_metadata_collate` custom collate function when  capturing of state dict in the CaptureMapDataset.
This change was proposed by @tchaton in offline discussion.

Part of #9130 

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [ ] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

Issue #8835, opened 2021-08-10
resume_from_checkpoint broken when fault-tolerant feature enabled


awaelchli:
## 🐛 Bug

A trainer with the resume from checkpoint option does not continue training and stops immediately, despite increased max_epoch settings.

### To Reproduce
```python

import os
import unittest.mock

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        print(batch.sum())
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def validation_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("valid_loss", loss)

    def test_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("test_loss", loss)

    def on_train_epoch_end(self):
        print("epoch ended")

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)

    def on_load_checkpoint(self, checkpoint):
        pass


@unittest.mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": "1"})
def run():
    train_data = DataLoader(RandomDataset(32, 64), batch_size=2)

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=3,
        limit_val_batches=0,
        num_sanity_val_steps=0,
        max_epochs=1,
        weights_summary=None,
    )
    trainer.fit(model, train_dataloader=train_data)

    trainer.save_checkpoint("lightning_logs/auto.pt")

    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=3,
        limit_val_batches=0,
        num_sanity_val_steps=0,
        max_epochs=3,
        weights_summary=None,
        resume_from_checkpoint="lightning_logs/auto.pt",
    )
    trainer.fit(model, train_dataloader=train_data)


if __name__ == "__main__":
    run()


```

Output: 

[...]

PR #9491, opened 2021-09-13
[Feat] Add Fault Tolerant Training for ValidationLoop

tchaton:
## What does this PR do?

Part of https://github.com/PyTorchLightning/pytorch-lightning/issues/9130

This PR would be likely splitted in smaller pieces. First pass reviews on the logic structure.

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -98,7 +105,7 @@ def reset(self) -> None:
         # track epoch output
         self._epoch_output = [[] for _ in range(self.batch_loop.num_active_optimizers(self.total_batch_idx))]
 
-        if not self.restarting or self._num_training_batches_reached():
```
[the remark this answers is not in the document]
awaelchli: I ran your branch and I see the same issue here again #8835

pytorch_lightning/trainer/progress.py
```diff
@@ -153,6 +153,10 @@ def load_state_dict(self, state_dict: dict) -> None:
         self.total.load_state_dict(state_dict["total"])
         self.current.load_state_dict(state_dict["current"])
 
+    def reset_on_restart(self):
+        self.total.reset_on_restart()
```
awaelchli, line 157: what does it mean to reset the total counter? I don't understand this

pytorch_lightning/loops/optimization/closure.py
```diff
@@ -188,3 +189,25 @@ def closure(self, *args: Any, **kwargs: Any) -> ClosureResult:
...
+    def consume_result(self) -> ClosureResult:
+        return self._closure.consume_result()
+
+    def __call__(self):
+        # enable fault tolerant during closure execution
+        with self._trainer._fault_tolerant_supported(enable=True):
+            loss = self._closure()
+        self._fault_tolerant_possible = self._trainer._fault_tolerant_possible
+        # prevent fault tolerant during parameters update.
+        self._trainer._fault_tolerant_possible = False
```
awaelchli: do we expect a failure during the parameter update in the optimizer?

awaelchli: don't we want the lightning optimizer to take care of this?

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -464,18 +464,10 @@ def _capture_metadata_collate(samples: List, dataset: Dataset, default_collate:
...
-
-    elif isinstance(dataset, CaptureMapDataset):
-        samples, states = zip(*samples)
-        data = default_collate(samples)
-        metadata = states[-1]
-    else:
-        return default_collate(samples)
-
-    return {"data": data, AutoRestartBatchKeys.PL_RESTART_META: metadata}
+    data = default_collate(samples)
```
awaelchli, line 467: I already made these changes based on your recommendation in a separate PR: #9484 
this is not relevant to this PR as far as I can tell.

pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -278,6 +279,8 @@ def _run_optimization(
         # ------------------------------
         # gradient update with accumulated gradients
         else:
+            if _fault_tolerant_training():
+                closure = _FaultToleranceClosureExecutor(closure, self.trainer)
```
carmocca, line 283: Can you explain this? Is it to avoid saving on exception if `optimizer.step` hasn't been called?

Why do we need it?
How is it related to the validation loop?

pytorch_lightning/loops/fit_loop.py
```diff
@@ -263,8 +263,9 @@ def teardown(self) -> None:
 
     def on_save_checkpoint(self) -> Dict:
         state_dict = super().on_save_checkpoint()
-        # TODO: update has_completed to its proper value
-        state_dict["dataloader_state_dict"] = self.trainer.train_dataloader.state_dict(has_completed=False)
+        # FIXME(@tchaton) Should pass has_completed=True when iterator is exhausted ?
```
carmocca, line 266: I think @awaelchli recently changed this

awaelchli (commented):
are we going to partition every loop into 
`fault_tolerant_supported(true)`
and
`fault_tolerant_supported(false)`
? 
I don't have the full picture on this.

pytorch_lightning/core/optimizer.py
```diff
@@ -121,6 +121,16 @@ def toggle_model(self, sync_grad: bool = True):
             yield
             self._untoggle_model()
 
+    @contextmanager
+    def __fault_tolerant_supported(self) -> Generator:
+        """The closure should prevent execution fault tolerance within the parameters update.
+
+        Restore the fault tolerant state accordingly
```
carmocca, line 128: Is this asuming that the inner code (the one that yields) will change the value of `self._trainer._fault_tolerant_possible`?

Is the purpose here to set the previous value in case that happens?

This is very ugly in terms of side effects and misdirection.

pytorch_lightning/loops/base.py
```diff
@@ -217,7 +217,7 @@ def _load_from_state_dict(
             if isinstance(v, BaseProgress):
                 v.load_state_dict(state_dict[key])
                 if restart_progress:
-                    apply_to_collection(v, Progress, lambda p: p.current.reset_on_restart())
+                    apply_to_collection(v, BaseProgress, lambda p: p.reset_on_restart())
```
carmocca, line 220: Why is this necessary?

Not reseting `p.total` is done on purpose.

Why do we need this `restart_progress` flag and this code here instead of relying on

```python
def reset():
    if self.restarting:
         reset_something()
```

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -76,7 +76,14 @@ def done(self) -> bool:
         signals to stop (e.g. by early stopping).
         """
         max_steps_reached = self.max_steps is not None and self.global_step >= self.max_steps
-        return max_steps_reached or self.trainer.should_stop or self._num_training_batches_reached()
+        # ßdone should be True only when `batch_loop` and `val_loop` are done.
```
carmocca, line 79:
```suggestion
        # done should be True only when `batch_loop` and `val_loop` are done.
```

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -76,7 +76,14 @@ def done(self) -> bool:
...
+        should_check_val = self.restarting and self._should_check_val_fx(self.batch_idx, self.is_last_batch)
+        return (
+            max_steps_reached
+            or self.trainer.should_stop
+            or self._num_training_batches_reached()
+            and not should_check_val
```
carmocca, line 85: why is adding this necessary?

running validation should have happened if it had to in the previous iteration

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca, line 101: bad merge? this was recently added by @awaelchli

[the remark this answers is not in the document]
tchaton: Interesting. This might be problematic as it would reset the progress even if validation hasn't been run. I need to find about the right solution there.

pytorch_lightning/trainer/progress.py
tchaton: Good catch ! This is a bug.

pytorch_lightning/loops/optimization/closure.py
tchaton: The LightningOptimizer can't handle this logic as the closure is being passed the accelerators. The only solution I found was to wrap the closure on optimizer.step call.

pytorch_lightning/loops/epoch/training_epoch_loop.py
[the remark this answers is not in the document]
awaelchli: we could save to the state_dict if validation has run or not

pytorch_lightning/loops/fit_loop.py
tchaton: Reverted back.

pytorch_lightning/loops/base.py
tchaton: `OptimizationProgress` was not properly resetted. It took me some to find this. I think it is safer to always call the BaseProgress and rely on the reset_on_restart to not reset progress.

pytorch_lightning/loops/epoch/training_epoch_loop.py
tchaton: `done` shouldn't be `True` if validation has run yet in advance_end. Therefore, `should_check_val` needs to be added to enable restarting within the validation loop.

pytorch_lightning/loops/epoch/training_epoch_loop.py
tchaton: No, it wasn't a bad merge. Without this, restart on validation loop won't work.
But I would have to find a way to make both work together.

pytorch_lightning/loops/optimization/optimizer_loop.py
tchaton: This is to avoid saving a checkpoint on failure if the weights were being updated.

pytorch_lightning/core/optimizer.py
tchaton: Yes, this was a solution to complete the `self._trainer._fault_tolerant_possible=False` applied after the closure is being run.

tchaton:
Hey @awaelchli @carmocca,

From your comments, you aren't too fan of the `fault_tolerant_supported`, therefore I will drop it for now.

And I will try to find a way to make @awaelchli fix works with restart in validation loop.

Best,
T.C

pytorch_lightning/loops/epoch/training_epoch_loop.py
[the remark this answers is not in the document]
tchaton: Yes, this is what I ended up doing.

pytorch_lightning/loops/base.py
tchaton: IMO, I prefer the loop to take care of sanitising progress.

Issue #9499, opened 2021-09-13
Remove ABC from Lightning Module


daniellepintz:
## Proposed refactoring or deprecation

No longer extend ABC in LightningModule:
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/core/lightning.py#L57

### Motivation

According to @carmocca in https://docs.google.com/document/d/1xHU7-iQSpp9KJTjI3As2EM0mfNHHr37WZYpDpwLkivA/edit?disco=AAAANOyYJFs this is legacy and can be removed. Originally added in https://github.com/PyTorchLightning/pytorch-lightning/pull/644 but we are no longer using the `@abstractmethod` decorator.

PuneetDabral:
mam can i do?

daniellepintz:
sure go for it!

PuneetDabral:
mam i am new in open source also in github

krkode:
hey @PuneetDabral , you can find this project's guide on how to contribute via their blog https://devblog.pytorchlightning.ai/quick-contribution-guide-86d977171b3a
(ps. im new to contributing as well so i tend to check a project's README.md file first because it typically has instructions on how to contribute)

PuneetDabral:
@krkode hey bro can you please explain me in a simple language....

PR #9529, opened 2021-09-14
do not mark lightningmodule methods as abstract

### LINKED ISSUE #8752: Make default implementation for train_dataloader in DataHooks raise NotImplementedError
### Discussed in https://github.com/PyTorchLightning/pytorch-lightning/discussions/8734

The default implementation for these is logging a warning that nothing is implemented. https://github.com/PyTorchLightning/pytorch-lightning/blob/963c26764682fa4cf64c93c5a7572ae0040e9c32/pytorch_lightning/core/hooks.py#L529

The current default implementation logs a warning, which can be lost, and which forces the framework to perform is_overridden checks. Instead, having the default implementation raise a NotImplementedError will make errors clear to users in case users forget to implement these hooks.

This applies for all the dataloader hooks in DataHooks, not just train_dataloader

awaelchli:
## What does this PR do?

Fixes #8752
Follow up to #9499 (not blocking each other)

Recently, #8752 added NotImplementedError being raised in LightningModule data hooks, but this marks the methods as abstract and linters complain that user did not implement all abstract methods. 
This is confusing and a bad user experience, since our docs never state that you need to implement these methods. In fact, we always recommend that a datamodule should be considered first!

image

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/core/hooks.py
+++ b/pytorch_lightning/core/hooks.py
@@ -19,6 +19,7 @@ import torch
 from torch.optim.optimizer import Optimizer
 
 from pytorch_lightning.utilities import move_data_to_device
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.types import EVAL_DATALOADERS, STEP_OUTPUT, TRAIN_DATALOADERS
 
 
@@ -471,7 +472,7 @@ class DataHooks:
                 # each batch will be a dict of tensors: {'mnist': batch_mnist, 'cifar': batch_cifar}
                 return {'mnist': mnist_loader, 'cifar': cifar_loader}
         """
-        raise NotImplementedError("`train_dataloader` must be implemented to be used with the Lightning Trainer")
+        raise MisconfigurationException("`train_dataloader` must be implemented to be used with the Lightning Trainer")
 
     def test_dataloader(self) -> EVAL_DATALOADERS:
         r"""
@@ -533,7 +534,7 @@ class DataHooks:
             In the case where you return multiple test dataloaders, the :meth:`test_step`
             will have an argument ``dataloader_idx`` which matches the order here.
         """
-        raise NotImplementedError("`test_dataloader` must be implemented to be used with the Lightning Trainer")
+        raise MisconfigurationException("`test_dataloader` must be implemented to be used with the Lightning Trainer")
 
     def val_dataloader(self) -> EVAL_DATALOADERS:
         r"""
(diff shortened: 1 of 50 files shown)


SeanNaren (approved):

ethanwharris (approved):

pytorch_lightning/core/hooks.py
```diff
@@ -471,7 +472,7 @@ def train_dataloader(self):
                 # each batch will be a dict of tensors: {'mnist': batch_mnist, 'cifar': batch_cifar}
                 return {'mnist': mnist_loader, 'cifar': cifar_loader}
         """
-        raise NotImplementedError("`train_dataloader` must be implemented to be used with the Lightning Trainer")
```
carmocca, line 474: I don't get why PyCharm shows that warning. Other classes in our codebase raise `NotImplementedError` but they don't have that warning. For example:

https://github.com/PyTorchLightning/pytorch-lightning/blob/23450e29055addcf8f1100b4fc3a4ed4fa9c0e2d/pytorch_lightning/overrides/base.py#L39

https://github.com/PyTorchLightning/pytorch-lightning/blob/23450e29055addcf8f1100b4fc3a4ed4fa9c0e2d/pytorch_lightning/callbacks/progress/base.py#L140

(among others)

carmocca (approved):

pytorch_lightning/core/hooks.py
awaelchli: Yes, I also find it rude. If this changes in the future, we can change it back too.

carmocca: minimal repro:

```python
class Foo:
    def fn(self):
        raise NotImplementedError


class HasWarning(Foo):
    ...


class NoWarning:
    def fn(self):
        raise NotImplementedError
```

This looks like a false positive. Already reported in https://youtrack.jetbrains.com/issue/PY-38680

We can just disable the warning

awaelchli: do you recommend to close the PR? is it important that we raise NotImplemented?
I read the [docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError) and interpret it differently

carmocca: We can close this or add `# noqa`

Unfortunately, there's no mechanism to silence that warning in particular so any other warnings would be hidden.

I personally lean towards closing this but no strong opinion.

awaelchli: I tried adding noqa to the lines where we raise NotImiplementedError and the signature of the method but both did not work

carmocca: `class LightningModule(  # noqa` works for me

No need to add it to the `NotImplementedError`

ananthsub: If adding the noqa marker, could we link to the JetBrains issue in the code to know why it's there? 

Once that issue is fixed I'd prefer using NotImplementedError as its canonical Python while Misconfiguration is Lightning specific. It's very minor though, as long as we raise an exception compared to before of logging a warning message.

awaelchli: @carmocca any adding noqa to the base class gets rid of the warning there but not in any of the derived classes.

PR #9371, opened 2021-09-08
fix resuming from checkpoint for fault-tolerant in case of no failure

awaelchli:
## What does this PR do?

Fixes #8835

When saving the loop state dict on a failure (breaking in the middle of an epoch), we want to save the progress counters exactly as they are in order to resume on the failed batch. The same applies to runs that finish successfully and with the full epoch completed. If we resume training in such a case, we don't want to resume on the last batch that we saved, but instead on the next one which is the batch 0 from the next epoch. So the counters need to reset every start of an epoch if:

- we are not restarting (regular case)
- if we are restarting, the current counters are at the limit of the loop AND we did not fail

**Note:** 
- This fix applies to master branch only (fault-tolerant training).
- Test fails on master

TODO: 
- [x] test

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

PR #9537, opened 2021-09-15
multiple optimizer restart with fault-tolerant training

awaelchli:
## What does this PR do?

Adds necessary changes and tests for restarting a run that failed at a training_step when multiple optimizers are involved. Note: if the user has interactions between the different training_step calls during the loop over the optimizers, this will not work or silently lead to different results, as it is the case with all other places were fault-tolerant takes place.

Part of #9130 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/loops/optimization/optimizer_loop.py
+++ b/pytorch_lightning/loops/optimization/optimizer_loop.py
@@ -204,7 +204,7 @@ class OptimizerLoop(Loop):
         raise NotImplementedError(f"{self.__class__.__name__} does not connect any child loops.")
 
     def reset(self) -> None:
-        if not self.restarting or self.done:
+        if not self.restarting:
             self.optim_progress.optimizer_position = 0
         self.outputs = [[] for _ in range(len(self.trainer.optimizers))]
 
@@ -226,6 +226,7 @@ class OptimizerLoop(Loop):
         self.optim_progress.optimizer_position += 1
 
     def on_run_end(self) -> _OUTPUTS_TYPE:
+        self.optim_progress.optimizer_position = 0
         outputs, self.outputs = self.outputs, []  # free memory
         return outputs
 
--- a/pytorch_lightning/loops/optimization/optimizer_loop.py
+++ b/pytorch_lightning/loops/optimization/optimizer_loop.py
@@ -205,6 +205,7 @@ class OptimizerLoop(Loop):
 
     def reset(self) -> None:
         if not self.restarting:
+            # when reset() is called from outside (manually), we reset the loop progress
             self.optim_progress.optimizer_position = 0
         self.outputs = [[] for _ in range(len(self.trainer.optimizers))]
 
@@ -213,6 +214,8 @@ class OptimizerLoop(Loop):
     ) -> None:
         self._batch_idx = batch_idx
         self._indices, self._optimizers = zip(*optimizers)
+        if self.done:
+            self.optim_progress.optimizer_position = 0
 
     def advance(self, batch: Any, *args: Any, **kwargs: Any) -> None:  # type: ignore[override]
         result = self._run_optimization(
@@ -226,7 +229,6 @@ class OptimizerLoop(Loop):
         self.optim_progress.optimizer_position += 1
 
     def on_run_end(self) -> _OUTPUTS_TYPE:
-        self.optim_progress.optimizer_position = 0
         outputs, self.outputs = self.outputs, []  # free memory
         return outputs
 
--- a/tests/loops/optimization/test_optimizer_loop.py
+++ b/tests/loops/optimization/test_optimizer_loop.py
@@ -118,6 +118,10 @@ def test_optimizer_frequencies(tmpdir, frequencies, expected):
     assert list(zip(opt_idx_sequence, optimizer_sequence)) == expected
 
 
+class CustomException(Exception):
+    pass
+
+
 @RunIf(min_torch="1.7.0")
 @mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": "1"})
 @pytest.mark.parametrize("stop_epoch", (0, 1))
--- a/tests/loops/optimization/test_optimizer_loop.py
+++ b/tests/loops/optimization/test_optimizer_loop.py
@@ -13,21 +13,18 @@
 # limitations under the License.
 
 
-from torch.optim import Adam, SGD
-
-from pytorch_lightning.core.optimizer import LightningOptimizer
-from pytorch_lightning.loops.optimization.optimizer_loop import ClosureResult
-
 import os
 from unittest import mock
 from unittest.mock import Mock
 
 import pytest
 import torch
+from torch.optim import Adam, SGD
 
 from pytorch_lightning import seed_everything, Trainer
+from pytorch_lightning.core.optimizer import LightningOptimizer
+from pytorch_lightning.loops.optimization.optimizer_loop import ClosureResult
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
-
 from tests.helpers import BoringModel
 from tests.helpers.runif import RunIf
 
--- a/tests/loops/optimization/test_optimizer_loop.py
+++ b/tests/loops/optimization/test_optimizer_loop.py
@@ -145,8 +145,10 @@ class CustomException(Exception):
 @pytest.mark.parametrize("n_optimizers,stop_optimizer", [(2, 0), (2, 1), (3, 2)])
 def test_loop_restart_progress_multiple_optimizers(tmpdir, n_optimizers, stop_optimizer, stop_epoch, stop_batch):
     """Test that Lightning can resume from a point where a training_step failed while in the middle of processing
-    several optimizer steps for one batch. The test asserts that we end up with the same trained weights as if no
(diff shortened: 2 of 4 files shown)


pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -226,6 +226,7 @@ def advance(self, batch: Any, *args: Any, **kwargs: Any) -> None:  # type: ignor
         self.optim_progress.optimizer_position += 1
 
     def on_run_end(self) -> _OUTPUTS_TYPE:
+        self.optim_progress.optimizer_position = 0
```
awaelchli, line 229: @carmocca you commented on a PR here #9371 that you prefer the reset logic for the optim progress to live inside the `reset()` method entirely. But this doesn't work anymore, because the above `self.done` call inside reset is always True due to `self._indices` being an empty list. The list cannot be populated until `on_run_start`. 

The alternative is to override run:

```py
    def run(self, batch: Any, optimizers: List[Tuple[int, Optimizer]], batch_idx: int):
        self._indices, self._optimizers = zip(*optimizers)
        return super().run(batch, optimizers, batch_idx)
```
to populate the list as early as possible so `reset()` can run `self.done`. Would you prefer this or the current PR content?

tests/loops/test_loops.py
```diff
@@ -334,13 +332,151 @@ def val_dataloader(self):
     assert trainer.fit_loop.epoch_loop.val_loop.epoch_loop.batch_progress.state_dict() == expected
 
 
+@RunIf(min_torch="1.7.0")
+@mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": "1"})
+@pytest.mark.parametrize("stop_epoch", (0, 1))
+@pytest.mark.parametrize("stop_batch", (0, 1, 2))
+@pytest.mark.parametrize("n_optimizers,stop_optimizer", [(2, 0), (2, 1), (3, 2)])
+def test_loop_restart_progress_multiple_optimizers(tmpdir, n_optimizers, stop_optimizer, stop_epoch, stop_batch):
```
awaelchli, line 340: recommendations where such a test should live? here? in `tests/utilities/test_auto_restart.py`

awaelchli: also, the test runs for about 3-4 seconds on my CPU. not sure how to cut it down. if we want to compare the actual trained weights, we need to run a training. 
perhaps cut down on the parametrizations? (though it is important to cover the boundary cases)

tests/loops/test_loops.py
```diff
@@ -334,13 +332,112 @@ def val_dataloader(self):
...
+
+@RunIf(min_torch="1.7.0")
 @mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": "1"})
 @pytest.mark.parametrize("accumulate_grad_batches", (1, 2, 3))
 @pytest.mark.parametrize("n_optimizers", (1, 3, 5))
 @pytest.mark.parametrize("stop_epoch", (1, 2))
 @pytest.mark.parametrize("stop_batch", (1, 2))
 @pytest.mark.parametrize("stop_optimizer", (1, 2))
-@RunIf(min_torch="1.7.0")
 def test_loop_state_on_exception(accumulate_grad_batches, stop_epoch, stop_batch, stop_optimizer, n_optimizers, tmpdir):
```
awaelchli, line 441: reviewer: no changes to the test, just adding more assertions

tests/loops/test_loops.py
```diff
@@ -506,19 +601,35 @@ def configure_optimizers_multiple(self):
     # need to remove these elements for comparison; comparing with `fit_loop.state_dict()` would require the
     # fit loop to have an iterator, which is only available during training
     checkpoint["loops"]["fit_loop"]["state_dict"]["dataloader_state_dict"] = ANY
-
     assert state_dict == checkpoint["loops"]["fit_loop"]
 
-    trainer.fit_loop.load_state_dict(checkpoint["loops"]["fit_loop"])
+    # with `restart_progress=True`, we expect all `ready` counters to be reset to `completed`
```
carmocca, line 606: side note: I discussed with @tchaton this morning that `restart_progress` shouldn't be on `load_state_dict`.

And instead we load exactly what was passed, and then each loop is responsible of resetting its state.

tests/loops/test_loops.py
```diff
@@ -506,19 +601,35 @@ def configure_optimizers_multiple(self):
...
+    assert batch_progress.current.ready == be_batches_completed
+    assert batch_progress.current.completed == be_batches_completed
+
+    optim_progress = trainer.fit_loop.epoch_loop.batch_loop.optimizer_loop.optim_progress
+    assert optim_progress.optimizer.step.current.ready == be_total_opt_steps
+    assert optim_progress.optimizer.step.current.completed == be_total_opt_steps
+    assert optim_progress.optimizer.zero_grad.current.ready == be_total_zero_grad
+    assert optim_progress.optimizer.zero_grad.current.completed == be_total_zero_grad
+
+    # not sure what the point of these assertions is
```
carmocca, line 623: I think these 3 asserts were added by @tchaton in the original test - perhaps we should remove them

tests/loops/test_loops.py
carmocca: `tests/loops/optimization/test_optimizer_loop.py`?

tests/loops/test_loops.py
```diff
@@ -334,13 +332,112 @@ def val_dataloader(self):
...
+    num_optimizers_incomplete = stop_epoch * n_batches * n_optimizers + stop_batch * n_optimizers + stop_optimizer
+
+    opt_idx_sequence_complete = list(range(n_optimizers)) * n_epochs * n_batches  # [0, 1, 2, 0, 1, 2, 0, 1, ...]
+    # +1 because we fail inside the closure inside optimizer_step()
+    opt_idx_sequence_incomplete = opt_idx_sequence_complete[: (num_optimizers_incomplete + 1)]
+    opt_idx_sequence_resumed = opt_idx_sequence_complete[num_optimizers_incomplete:]
+
+    class MultipleOptimizerModel(BoringModel):
+        def __init__(self):
+            super().__init__()
```
carmocca, line 359: is this here for a reason?

tests/loops/test_loops.py
```diff
@@ -334,13 +332,112 @@ def val_dataloader(self):
...
+                and optimizer_idx == stop_optimizer
+            ):
+                raise CustomException
+            return super().training_step(batch, batch_idx)
+
+        def configure_optimizers(self):
+            return [torch.optim.SGD(self.parameters(), lr=0.1) for _ in range(n_optimizers)]
+
+    # run without a failure, collect weights
+    fail = False
```
carmocca, line 375: this is defined above, remove one

pytorch_lightning/loops/optimization/optimizer_loop.py
carmocca: The current solution looks simpler, but I see a third option which is to save the `self._indices` as part of the tracking so that we can still call `self.done`

tchaton: I think overriding run seems the best option no ? It means we could move BaseProgress reset_on_restart within the reset function.

tests/loops/test_loops.py
tchaton: Can you skip validation by setting limit_val_batches=0 and logger=None and checkpointing ?

tests/loops/test_loops.py
tchaton: Yes, I will make a PR for this.

tests/loops/test_loops.py
awaelchli: yes, it is there to show the beauty of python constructors

pytorch_lightning/loops/optimization/optimizer_loop.py
awaelchli: @carmocca and I found a shorter solution that does not require overriding `run()` or move the existing code for indices.

tests/loops/test_loops.py
awaelchli: done, both

tests/loops/test_loops.py
awaelchli: they don't hurt, we can keep them

tests/loops/optimization/test_optimizer_loop.py
```diff
@@ -11,17 +11,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+
```
carmocca, line 15: nit: empty newlines

```suggestion
```

carmocca (approved):
🚀

tests/loops/optimization/test_optimizer_loop.py
awaelchli: hmm, sometimes we can't rely on black?

carmocca: I don't think black or isort have any rules about that

tchaton (approved):
LGTM !

justusschock (approved):

Issue #8088, opened 2021-06-22
Limited types of data can be logged


borisdayma:
## 🐛 Bug

Function `__check_allowed` limits the type of parameters that can be logged, even this data would be supported by loggers such as wandb.

### To Reproduce

See following [**notebook**](https://colab.research.google.com/drive/1LLIPqS2xgSFtGGg4dnoMBeQOkgPW9bnp?usp=sharing).

Error stack:

```
ValueError                                Traceback (most recent call last)
<ipython-input-12-6d5bedbc661e> in <module>()
----> 1 trainer.fit(model, datamodule=mnist)

18 frames
/usr/local/lib/python3.7/dist-packages/pytorch_lightning/trainer/trainer.py in fit(self, model, train_dataloaders, val_dataloaders, datamodule, train_dataloader)
    509         self.checkpoint_connector.resume_start()
    510 
--> 511         self._run(model)
    512 
    513         assert self.state.stopped

/usr/local/lib/python3.7/dist-packages/pytorch_lightning/trainer/trainer.py in _run(self, model)
    870 
    871         # dispatch `start_training` or `start_evaluating` or `start_predicting`
--> 872         self._dispatch()
    873 
    874         # plugin will finalized fitting (e.g. ddp_spawn will load trained model)

/usr/local/lib/python3.7/dist-packages/pytorch_lightning/trainer/trainer.py in _dispatch(self)
    912             self.accelerator.start_predicting(self)
    913         else:
--> 914             self.accelerator.start_training(self)
    915 
    916     def run_stage(self):

/usr/local/lib/python3.7/dist-packages/pytorch_lightning/accelerators/accelerator.py in start_training(self, trainer)
     94 
     95     def start_training(self, trainer: 'pl.Trainer') -> None:
---> 96         self.training_type_plugin.start_training(trainer)
     97 
     98     def start_evaluating(self, trainer: 'pl.Trainer') -> None:

/usr/local/lib/python3.7/dist-packages/pytorch_lightning/plugins/training_type/training_type_plugin.py in start_training(self, trainer)
    163     def start_training(self, trainer: 'pl.Trainer') -> None:
    164         # double dispatch to initiate the training loop
--> 165         self._results = trainer.run_stage()
    166 
    167     def start_evaluating(self, trainer: 'pl.Trainer') -> None:

/usr/local/lib/python3.7/dist-packages/pytorch_lightning/trainer/trainer.py in run_stage(self)
[...]

PR #6227, opened 2021-02-26
Abstract log_figure API that implements reasonable default for all lo…

Haydnspass:
Transfer of https://github.com/PyTorchLightning/pytorch-lightning/pull/4975
Sorry for the mess with choosing the wrong branch. It was faster for me to manually include the changes, so I am opening a new PR.

## What does this PR do?
Make abstraction layer for figure logging, i.e. add default for logging figures for all implemented loggers.

My suggestion to https://github.com/PyTorchLightning/pytorch-lightning/issues/4973

ToDo:
- [x] Base implementation (can close figure)
- [x] Comet.ML
- [x] MLFlow.
- [x] Neptune
- [x] Tensorboard
- [x] ~~TestTube~~ (defaults to base)
- [x] ~~CSV~~ (defaults to base)

## Before submitting
- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together? Otherwise, we ask you to create a separate PR for every change.
- [x] Did you make sure to update the documentation with your changes?
- [x] Did you write any new necessary tests? 
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)?

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:  

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [ ] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified; _Bugfixes should be including in bug-fix release milestones (m.f.X) and features should be included in (m.X.b) releases._
 

## Did you have fun?
Make sure you had fun coding 🙃

PR #9545, opened 2021-09-15
[WIP] feat(wandb): support media logging

borisdayma:
## What does this PR do?

Support media logging with W&B.

TODO:
* [x] implement `log_text`
* [x] implement `log_image`
* [ ] evaluate other media logging
* [ ] add documentation
* [ ] add tests

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/loggers/wandb.py
+++ b/pytorch_lightning/loggers/wandb.py
@@ -19,8 +19,11 @@ import operator
 import os
 from argparse import Namespace
 from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 from weakref import ReferenceType
+
+if TYPE_CHECKING:
+    import pandas
 
 import torch.nn as nn
 
@@ -216,6 +219,25 @@ class WandbLogger(LightningLoggerBase):
             self.experiment.log({**metrics, "trainer/global_step": step})
         else:
             self.experiment.log(metrics)
+
+    @rank_zero_only
+    def log_text(
+        self,
+        key: str,
+        columns: List[str] = None,
+        data: List[List[str]] = None,
+        dataframe: "pandas.Dataframe" = None,
+        step: Optional[int] = None,
+    ) -> None:
+
+        metrics = {key: wandb.Table(columns=columns, data=data, dataframe=dataframe)}
+        self.log_metrics(metrics, step)
+
+    @rank_zero_only
+    def log_image(self, key: str, **kwargs: str) -> None:
+        step = kwargs.pop("step", None)
+        metrics = {key: wandb.Image(kwargs)}
+        self.log_metrics(metrics, step)
 
     @property
     def save_dir(self) -> Optional[str]:


borisdayma:
Not really sure how to fix the error building the docs.
I think it's related to `pandas` not being available. Should I change the type hint to `Any`?

justusschock (changes requested):
Hi @borisdayma ,

Thanks for working on this. However, I am afraid that we cannot merge it as is, since at least we require the API to be built into the base class. We cannot have the logger API to diverge across the different logger implementations.

Something similar (but apparently no longer actively worked on) was started in https://github.com/PyTorchLightning/pytorch-lightning/pull/6227

pytorch_lightning/loggers/wandb.py
```diff
@@ -217,6 +374,53 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+    ) -> None:
+        """Log text as a Table.
+
+        Can be defined either with `columns` and `data` or with `dataframe`.
+        """
+
+        self.log_table(key, columns, data, dataframe, step)
+
+    @rank_zero_only
+    def log_images(self, key: str, images: List[Any], **kwargs: str) -> None:
```
tchaton, line 411: For consistency with other loggers, `log_image` should be used there.

borisdayma: The issue here is I wanted to make it explicit that it's a list of images (vs a single image).
In W&B, images need to be logged in batch to have a meaningful representation in the UI (and for better efficiency).
We just need to make sure that users won't loop through their images calling this method on each individual sample.
What do you think?

borisdayma:
Hi @justusschock 

> We cannot have the logger API to diverge across the different logger implementations.

I understand the interest to have a unified logger API but I see a few drawbacks:

* Loggers already have their differences
  * the input arguments are different between each logger due to their unique functionality
  * `WandbLogger` supports a [`watch`](https://github.com/PyTorchLightning/pytorch-lightning/blob/982a9560a5066c096a7a28691bb109058fd92ed0/pytorch_lightning/loggers/wandb.py#L200-L201) method used to track gradients, parameters, etc
  * `WandbLogger` has a unique way of logging models as artifacts with aliases
  * `NeptuneLogger` also has its own unique methods such as `log_text`, `log_image` and `log_artifact`
* It would be limiting and would prevent from offering features that are unique to each logger
  * Neptune seems to log images one at a time while in wandb, they must be logged in batches for a single step
  * wandb supports unique medias (segmentation masks, bounded boxes, video, audio, 3d point cloud, molecule, histograms, scatter plots, ROC curves, confusion matrix, interactive charts through vega spec, etc). It would be a pain to prevent users from using these media types just because they are not supported by the other loggers
  * wandb now also supports [W&B Tables](https://docs.wandb.ai/guides/data-vis) that can be queried to visualize, analyze and compare any form of data. It would have been nice to add this feature in Pytorch Lightning (through `log_table` as proposed in this PR)
* Limiting logging to a list of supported types is actually a breaking change as reported in #8088. I would suggest letting people pass any type of object to their loggers to ensure we don't limit any logger functionality. Typically only one logger is used and it is probably expected that users would have to adapt their logging code if using very specific features and switching between loggers (wandb, mlflow, neptune, tensorboard, etc).

borisdayma:
As an example this is some data logged directly using `log_images` and `log_text` from this PR: [W&B run](https://wandb.ai/borisd13/MNIST/runs/34gw93ht?workspace=user-borisd13)

justusschock:
Hi @borisdayma 

I know that wandb already has different methods such as `watch` but these are things that are simply not supported by other frameworks like that.

With other more common features like logging media though, I think we should generalize the API there. 

I'm not saying that we should already have implementations for all loggers right away. More that we need to provide abstract functions as interfaces to the base logger class and then add the implementations for wandb. And all loggers not implementing these methods will just `raise NotImplementedError`

borisdayma:
Are you ok using `*args, *kwargs` in the generic implementation so loggers have more flexibility with the methods?
Wandb accepts for example many different arguments and multiple types for images so it would be better to control it at the logger level.

justusschock:
To some extend yes. E.g. when logging an image, the name and the image should not be in args or kwargs :) All the others can probably go in there.

pytorch_lightning/loggers/wandb.py
```diff
@@ -46,12 +49,180 @@ class WandbLogger(LightningLoggerBase):
...
-    Install it with pip:
+    **Installation and set-up**
+
+    Install with pip:
 
     .. code-block:: bash
 
         pip install wandb
 
+    Create a `WandbLogger` instance:
```
awaelchli, line 60: Would you be ok if we put this in a separate documentation page? IMO user guide does not belong into the API docs here. What is your opinion?

pytorch_lightning/loggers/wandb.py
awaelchli: @borisdayma that sounds reasonable. IMO it could be named log_image (singular) and still log either a single image or a list/batch of them.

borisdayma: yeah, need to give it more thoughts. I want at least to force users to pass list of images so they don't try to loop over images which would give a sub-optimal experience in the UI. Calling it `log_images` (and not supporting `log_image`) would make it obvious.

pytorch_lightning/loggers/wandb.py
borisdayma: So it was actually inspired by [NeptuneLogger](https://pytorch-lightning.readthedocs.io/en/latest/extensions/generated/pytorch_lightning.loggers.NeptuneLogger.html#neptunelogger).

The reason is that W&B gets a lot of questions on this integration where the answers are in wandb doc but not in PL doc and people always refer to one of these pages when asking their questions:
* [WandbLogger](https://pytorch-lightning.readthedocs.io/en/latest/extensions/generated/pytorch_lightning.loggers.WandbLogger.html#pytorch_lightning.loggers.WandbLogger) where I added basic info in this PR (and referred to W&B colab and doc for more details)
* [Loggers](https://pytorch-lightning.readthedocs.io/en/latest/common/loggers.html#weights-and-biases) where I actually was going to duplicate this same info

borisdayma:
@justusschock What do you think of making them generic [like this](https://github.com/PyTorchLightning/pytorch-lightning/pull/9545/files#diff-2b5424d2a2d21c741211768d5fd6cc0178c402a52e99535b0084649e188cef60R300-R305).

The reason is that loggers may have different implementations:
  * wandb is more adapted to lists vs neptune seems to use individual items
  * for text the item is not required as a dataframe could be passed to wandb (so we can pass "data + columns" or "dataframe"
  * for images, it would be hard to do type checking as each logger may accept different types

justusschock:
But wouldn't you agree that each logger takes at least a tag and the object to log? This is why I'd like to have them explicit and use *args and **kwargs only for the other arguments.

Thoughts @awaelchli @borisdayma ?

borisdayma:
> But wouldn't you agree that each logger takes at least a tag and the object to log?

Giving a tag/key seems to make sense. At least I can't think of an edge case against it.

For `log_text`, within `LigthningLoggerBase`, you would have:

```
def log_text(self, key:str, value, *args, **kwargs):
    pass
```

Then if we want to a type checking on value, we already have `NeptuneLogger`, which accepts a `str` and `WandbLogger` which would accept `Union(List[List[str]], pandas.Dataframe)`. If we try to unify the type of the argument, we limit one of the loggers.

In the case of Images it becomes even complex as for wandb it would be something like `Union[str, wandb.Image, matplotlib.artist.Artist, PIL.Image, TorchTensorType, np.ndarray]`.

Also I would not expect users to think anything is gonna get logged if they only pass a key with no other args and an error should be raised by the loggers themselves (if they have the method implemented). It would fail in wandb for example.

> In case of your wandb example I'd then suggest to internally do a typecheck and if text is a dataframe or a list parse for the respective other arguments

It could be ok here but I've been trying to simplify as much as possible this integration and use it as much as possible as a wrapper to wandb client for easier maintenance, which is why I even try to use consistent args whenever I can so wandb users have an easier time with the integrations.

`wandb` already does a lot of checking, debugging, tries to provide helpful error messages so it would be easier to delegate the entire task of parsing/checking/processing the arguments directly to that library.

pytorch_lightning/loggers/wandb.py
```diff
@@ -408,11 +408,12 @@ def log_text(
         self.log_table(key, columns, data, dataframe, step)
 
     @rank_zero_only
-    def log_images(self, key: str, images: List[Any], **kwargs: str) -> None:
+    def log_image(self, key: str, images: List[Any], **kwargs: str) -> None:
         """Log images (tensors, numpy arrays, PIL Images or file paths).
 
         Optional kwargs are lists passed to each image (ex: caption, masks, boxes).
         """
+        assert isinstance(images, list), f'Expected a list as "images", found {type(images)}'
```
justusschock, line 416: can you raise a TypeError instead? In an ideal world asserts only live in tests :)

pytorch_lightning/loggers/wandb.py
borisdayma: I changed it to `log_image`

pytorch_lightning/loggers/wandb.py
borisdayma: Done!

borisdayma:
So on my side my last questions are:

* How to solve the error building the docs? I think it's due to pandas not being available and it's only for type checking of `log_text`. Should I change the type checking to `Any`?
* Can I copy the same doc in the [loggers page](https://pytorch-lightning.readthedocs.io/en/latest/common/loggers.html)? The reason is that W&B gets a lot of questions on this integration where the answers are in wandb doc but not in PL doc and people always refer to the loggers page or the WandbLogger page when asking their questions.

pytorch_lightning/loggers/base.py
```diff
@@ -297,6 +297,20 @@ def log_graph(self, model: "pl.LightningModule", input_array=None) -> None:
         """
         pass
 
+    def log_text(self, *args, **kwargs) -> None:
+        """Log text.
+
+        Arguments are directly passed to the logger.
```
ananthsub, line 303: the docstring is somewhat confusing. won't people be calling `SomeLoggerImplementation.log_text` ?

pytorch_lightning/loggers/base.py
```diff
@@ -297,6 +297,20 @@ def log_graph(self, model: "pl.LightningModule", input_array=None) -> None:
         """
         pass
 
+    def log_text(self, *args, **kwargs) -> None:
```
ananthsub, line 300: is it misleading for users to need to figure out which loggers actually implement this vs which don't? the expectation for those users will be that `NotWandBlogger.log_text` actually logs text instead of the no-op here

pytorch_lightning/loggers/base.py
borisdayma: It is actually working like the other methods so we can call the method through the logger or through the lightning module.

pytorch_lightning/loggers/base.py
borisdayma: Not sure, some of the other methods are not implemented in all loggers either.
We could instead add a warning that appears once if the method is not implemented in one of the loggers used.

justusschock: I agree with @ananthsub . Since users will call these methods explicitly, we should also raise an Error if not implemented

borisdayma: Gread idea! I'm now raising `NotImplementedError`

borisdayma:
There seems to be a mismatch version between pandas & numpy in the test environment.
Should I remove my pandas test?

```
ImportError: this version of pandas is incompatible with numpy < 1.17.3
   your numpy version is 1.17.2.
   Please upgrade numpy to >= 1.17.3 to use this pandas version
```

borisdayma:
The PR is now ready for final review (failed test seem unrelated).
Let me know if any change is needed.

pytorch_lightning/loggers/wandb.py
awaelchli: ok, your point about people not finding the answers in the docs is valid. 
But still I want to push on the issue off having extensive user guide mixed up with api docs.
I would like to make a follow up to this PR:
- give each logger a dedicated page
- on the current logger page, only summarize the third-party capabilities in a quick teaser section, redirect to individual pages
- the dedicated logger pages will contain more detailed usage examples like you show here, separated from the api documentation

However, for the individual page we also want to keep it concise and avoid duplicating the whole wandb documentation. 

What do you think of my thoughts here.

awaelchli (commented):
nice work, I like it

pytorch_lightning/loggers/wandb.py
```diff
@@ -217,6 +371,56 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        self.log_table(key, columns, data, dataframe, step)
+
+    @rank_zero_only
+    def log_image(self, key: str, images: List[Any], **kwargs: str) -> None:
+        """Log images (tensors, numpy arrays, PIL Images or file paths).
+
+        Optional kwargs are lists passed to each image (ex: caption, masks, boxes).
+        """
+        if not isinstance(images, list):
+            raise TypeError(f'Expected a list as "images", found {type(images)}')
```
awaelchli, line 414: would we be ok here if we just converted it to a list of a single element?

awaelchli (approved):

pytorch_lightning/loggers/wandb.py
borisdayma: That's a great idea.
Let's move this content to a dedicated page in a follow-up PR then. Feel free to set up how you want it to be organized and ping me to create/move specific doc.

pytorch_lightning/loggers/wandb.py
borisdayma: Sure, should I add a warning as well?

borisdayma (commented):
Thanks for the comments

pytorch_lightning/loggers/wandb.py
tchaton: Great idea @awaelchli. Definitely would bring more visibility to all loggers :)

pytorch_lightning/loggers/wandb.py
```diff
@@ -46,12 +46,180 @@ class WandbLogger(LightningLoggerBase):
...
+    Log from :class:`~pytorch_lightning.core.lightning.LightningModule`:
+
+    .. code-block:: python
+
+        class LitModule(LightningModule):
+            def training_step(self, batch, batch_idx):
+                self.log("train/loss", loss)
+
+    Use directly wandb module:
+
```
tchaton, line 84: I won't recommend to present the wandb client directly. This would be confusing the users. They can get this information from wandb doc directly.

justusschock (approved):

pytorch_lightning/loggers/wandb.py
borisdayma: Sure I could remove it but the main idea here was:
* for custom objects not supported yet (I will add more in a follow-up PR).
* logging metrics before the start of the run

borisdayma:
Hi guys, just checking if I need to do anything else.

Issue #9407, opened 2021-09-09
Feature incompatibilities with HPC/Slurm saving & loading


ananthsub:
## Proposed refactoring or deprecation

Proposal: Deprecate dedicated HPC saving & loading
Part of #7740 

### Motivation
Feature compatibility is quickly dropping for HPC checkpointing in Lightning. 
- This is enabled only for SLURM schedulers through the SLURM connector. It has not generalized to any other scheduler: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/slurm_connector.py
- Checkpoint saving is incompatible with plugins like DeepSpeed
- Customizing IO is incompatible with the newly introduced CheckpointIO plugin: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L271-L302
- Users are asking for ways to disable the automatic resubmission of jobs because they handle this logic in their own scripts outside of the Trainer

Moving forward, I am concerned with supporting 4 distinct codepaths for saving checkpoints given what's happened with HPC. Paths that exist or are being worked on to trigger saving:
- Using a checkpoint callback during fitting
- Calling trainer.save_checkpoint directly
- Checking for HPC/SLURM preemption through signal handlers
- Checkpointing on exception if fault tolerance is enabled

All of these have differences around when & where checkpoints are saved, which in turn impact how the trainer resumes from these checkpoints: https://github.com/PyTorchLightning/pytorch-lightning/issues/9405

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

[...]

PR #9566, opened 2021-09-16
[Feat] Add graceful detection of signal to exit + SignalConnector and merge SlurmConnector.

tchaton:
## What does this PR do?

Right now, when the script is being killed, a fault tolerant checkpoint would be saved.
However, Lightning can't always ensure it is always fully happening in reproducible part of the codebase.

When running in the cloud, there is 2-3 min to kill the script.

The proposal is to add a mechanism to detect a killing signal as been sent and Lightning will terminate on the next reproducible part in this time windown.

Parts #9567

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/fault_tolerant_connector.py
```diff
@@ -0,0 +1,23 @@
+import logging
+import signal
+
+from pytorch_lightning.utilities.imports import _fault_tolerant_training
+
+log = logging.getLogger(__name__)
+
+
+class FaultTolerantConnector:
```
awaelchli, line 9: what if we rethink our signal handling together with the remains that are in the SLURM connector? 
also we agreed we don't want any connectors modifying the trainer anymore :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -1099,6 +1101,9 @@ def _pre_training_routine(self):
         # register auto-resubmit when on SLURM
         self.slurm_connector.register_slurm_signal_handlers()
 
+        # used to register gracefully detection of signal to ensure clean exit mechanism
+        self.fault_tolerant_connector.register_fault_tolerant_signal_handlers()
```
awaelchli, line 1105: note it's probably also equally important to de-register the handlers on teardown. or to at least provide one so people can unregister them. what do you think?

pytorch_lightning/trainer/connectors/fault_tolerant_connector.py
tchaton: Done ! Changed for SignalConnector and merged it with Slurm.

pytorch_lightning/trainer/trainer.py
tchaton: Yes, I will do this in another PR.

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -3,12 +3,27 @@
...
 
 
-class SLURMConnector:
+class SignalConnector:
     def __init__(self, trainer):
         self.trainer = trainer
+        self.trainer._should_gracefully_terminate = False
+
+    def register_signal_handlers(self):
+        cluster_env = getattr(self.trainer.training_type_plugin, "cluster_environment", None)
```
carmocca, line 18: noob q: is this not defined for all TTPs? why?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -3,12 +3,27 @@
...
+        if isinstance(cluster_env, SLURMEnvironment):
+            self.register_slurm_signal_handlers()
+        elif _fault_tolerant_training():
```
carmocca, line 21: This means there won't be fault tolerance with SLURM?

Having fault-tolerance, is there a point to the slurm resubmitting?

carmocca (approved):

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: It is not decided for SingleDevice ahah

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: There would be, but less safer.

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -23,10 +38,10 @@ def register_slurm_signal_handlers(self):
 
         if on_slurm:
             log.info("Set SLURM handle signals.")
-            signal.signal(signal.SIGUSR1, self.sig_handler)
+            signal.signal(signal.SIGUSR1, self.sig_slurm_handler)
             signal.signal(signal.SIGTERM, self.term_handler)
 
-    def sig_handler(self, signum, frame):  # pragma: no-cover
+    def sig_slurm_handler(self, signum, frame):  # pragma: no-cover
```
ananthsub, line 45: I would instead much prefer:
- The checkpoint saving for fault tolerance and the slurm signal handler use the same code
- After the fault-tolerant checkpoint is saved, allow the SLURM cluster environment to dictate whether job resubmission should happen.

otherwise there are no guarantees `hpc_save` keeps up : https://github.com/PyTorchLightning/pytorch-lightning/issues/9407

ananthsub (commented):
does signal handling generalize across all clusters/schedulers?

our scheduler offers a dedicated mechanism to check for preemption, which we've integrated in the model checkpoint callback on batch end for whether we ought to save a checkpoint that batch. but i also thought this was atypical in cloud providers workloads where the job scheduler could kill the job at any point

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -1,14 +1,30 @@
...
+    def register_signal_handlers(self):
+        cluster_env = getattr(self.trainer.training_type_plugin, "cluster_environment", None)
+        if isinstance(cluster_env, SLURMEnvironment):
+            self.register_slurm_signal_handlers()
+        elif _fault_tolerant_training() and not sys.platform == "win32":
+            self.register_fault_tolerant_handlers()
+
+    def register_fault_tolerant_handlers(self):
+        signal.signal(signal.SIGUSR1, self.sig_fault_tolerant_handler)
+        signal.signal(signal.SIGTERM, self.term_handler)
```
yifuwang, line 27: It's very typical for the user script or the runtime environment to register these signal handlers. It may lead to issues if a library overrides already registered custom signal handlers without the user knowing. For example, one of FB's production environments relies on SIGTERM handler to collect diagnostic information about the job, which this change will break.

Some alternatives I can think of:
- Instead of having the trainer register the signal handlers automatically, offer a utility for the user to register it themselves. This gives the user a chance to merge/reconcile multiple signal handlers.
- Save the previously registered signal handlers and invoke them when the trainer is ready to terminate gracefully. A catch though is that invoking a signal handler at a different place/time may not always make sense.

awaelchli (approved):

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Hey @ananthsub. We don't have access to a SLURM cluster, therefore I don't want to touch this feature as it is quite simple to break and lot of users are relying on this.

And yes, hpc_save would be to be refactored in further PRs.

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -1,14 +1,30 @@
...
+class SignalConnector:
     def __init__(self, trainer):
         self.trainer = trainer
+        self.trainer._should_gracefully_terminate = False
+
+    def register_signal_handlers(self):
+        cluster_env = getattr(self.trainer.training_type_plugin, "cluster_environment", None)
+        if isinstance(cluster_env, SLURMEnvironment):
+            self.register_slurm_signal_handlers()
+        elif _fault_tolerant_training() and not sys.platform == "win32":
```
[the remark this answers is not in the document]
tchaton: Removed !

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Good point. I added a mechanism to bypass it.

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,121 @@
...
+            if result == 0:
+                log.info(f"requeued exp {job_id}")
+            else:
+                log.warning("requeue failed...")
+
+            # close experiment to avoid issues
+            self.trainer.logger.close()
+
+    def fault_tolerant_sigusr1_handler_fn(self, signum, frame):  # pragma: no-cover
+        self.trainer._terminate_gracefully = True
```
carmocca, line 112: When will this get set back to `False`?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,121 @@
+import logging
```
carmocca, line 1: Add this to https://github.com/PyTorchLightning/pytorch-lightning/blob/2ec39e275ec10014da6b72af3cab4efad08a096e/pyproject.toml#L63

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,121 @@
...
+    def sigusr1_handler(self, sigusr1_handler: Callable) -> None:
+        self._sigusr1_handler = sigusr1_handler
+
+    @property
+    def sigterm_handler(self) -> Optional[Callable]:
+        return self._sigterm_handler
+
+    @sigterm_handler.setter
+    def sigterm_handler(self, sigterm_handler: Callable) -> None:
+        self._sigterm_handler = sigterm_handler
```
carmocca, line 47: What's the point of these getter/setters?

You can instead make the property public in the init.

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,121 @@
...
+        self.trainer._terminate_gracefully = True
+
+    def sigterm_handler_fn(self, signum, frame):  # pragma: no-cover
+        log.info("bypassing sigterm")
+
+    def _has_already_handler(self, signal: Signals) -> bool:
+        try:
+            return isinstance(_signal.getsignal(signal), Callable)
+        except AttributeError:
+            return False
```
carmocca, line 121: Why do we need to check this and use the protected `_signal`?

You can just overwrite the existing one, and maybe save it to set it back later

See the snippets in https://stackoverflow.com/a/10972804

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: It shoudln't be set back to False. The value is being set to True and as soon as possible, the script will exit.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: This isn't connected yet to the Accelerator. Advanced users might want to set them using a callback instead.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Done !

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: But if the user set them outside of Lightning, they won't be overridden, so it would be fine.

tchaton: You are right, with the detection, it won't be required anymore.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Done !

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,108 @@
...
+        if _fault_tolerant_training():
+            sigusr1_handlers.append(self.fault_tolerant_sigusr1_handler_fn)
+
+        if self._is_on_slurm():
+            log.info("Set SLURM handle signals.")
+            sigusr1_handlers.append(self.slurm_sigusr1_handler_fn)
+
+        sigterm_handlers.append(self.sigterm_handler_fn)
+
+        if not self._is_on_windows():
```
justusschock, line 45: why won't this work on windows?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self, trainer: "pl.Trainer"):
+        self.trainer = trainer
+        self.trainer._terminate_gracefully = False
+
+    def register_signal_handlers(self) -> None:
+        sigusr1_handlers: List[Callable] = []
+        sigterm_handlers: List[Callable] = []
+
+        if _fault_tolerant_training():
+            sigusr1_handlers.append(self.fault_tolerant_sigusr1_handler_fn)
```
justusschock, line 37: why do we need them only for sigurs1? shouldn't they apply to all signals?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,108 @@
...
+                log.warning("requeue failed...")
+
+            # close experiment to avoid issues
+            self.trainer.logger.close()
+
+    def fault_tolerant_sigusr1_handler_fn(self, signum: Signals, frame: FrameType) -> None:
+        self.trainer._terminate_gracefully = True
+
+    def sigterm_handler_fn(self, signum: Signals, frame: FrameType) -> None:
+        log.info("bypassing sigterm")
```
justusschock, line 86: How are you bypassing this here? Does it mean that in the end python won't know about sigterm any more?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _is_on_slurm(self) -> bool:
+        # see if we're using slurm (not interactive)
+        on_slurm = False
+        try:
+            job_name = os.environ["SLURM_JOB_NAME"]
+            if job_name != "bash":
+                on_slurm = True
```
justusschock, line 94: This would fail for all other shells right?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -0,0 +1,108 @@
...
+                signal.signal(signal.SIGUSR1, HandlersCompose(sigusr1_handlers))
+
+            if not self._has_already_handler(signal.SIGTERM):
+                signal.signal(signal.SIGTERM, HandlersCompose(sigterm_handlers))
+
+    def slurm_sigusr1_handler_fn(self, signum: Signals, frame: FrameType) -> None:
+        if self.trainer.is_global_zero:
+            # save weights
+            log.info("handling SIGUSR1")
+            self.trainer.checkpoint_connector.hpc_save(self.trainer.weights_save_path, self.trainer.logger)
```
justusschock, line 56: Will this also handle weights properly that may be different on each rank? e.g. DeepSpeed, Sharded or Model Parallel?

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: SIGUSR1 isn't defined.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: This is the current registered signal. We can explore why one we should support, but in another PR.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: This is legacy code. I am just conserving the behaviour.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Same as before, this code is just being moved around. This conserves the previous behaviour.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: This is on Slurm. I guess we should explore this if we get access to it.

CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     * Added partial support for global random state fault-tolerance in map-style datasets ([#8950](https://github.com/PyTorchLightning/pytorch-lightning/pull/8950))
     * Converted state to tuple explicitly when setting Python random state ([#9401](https://github.com/PyTorchLightning/pytorch-lightning/pull/9401))
     * Added support for restarting an optimizer loop (multiple optimizers) ([#9537](https://github.com/PyTorchLightning/pytorch-lightning/pull/9537))
+    * Added a mechanism to detect a signal as been sent so the Trainer can gracefully exit ([#9566](https://github.com/PyTorchLightning/pytorch-lightning/pull/9566))
```
SeanNaren, line 68:
```suggestion
    * Added mechanism to detect a signal has been sent so the Trainer can gracefully exit ([#9566](https://github.com/PyTorchLightning/pytorch-lightning/pull/9566))
```

SeanNaren (approved):

pytorch_lightning/trainer/connectors/signal_connector.py
SeanNaren: might be worth a comment :)

tchaton: Done !

justusschock (approved):

yifuwang:
Think looks great! Thanks for adding the mechanism @tchaton!

yifuwang (approved):

ananthsub (approved):

PR #9572, opened 2021-09-16
guard against None in pytorch get_xla_supported_devices

### LINKED ISSUE #9552: XLA error on CPU
https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/utilities/xla_device.py#L75

A somewhat unusual case, but this line causes a TypeError if you are using an XLA-enabled version of pytorch without any TPUs attached. (Google deprecated all of its non-XLA pytorch deep learning images so maybe this will become less unusual over time.)

This is because pytorch/xla returns a `None` and not an empty list if it can't find any devices of the desired type. Could we guard this with an `is None or`?

ckchow:
## What does this PR do?
Guard against `None` when checking if TPU's are available. This happens (e.g.) if you use google's standard deep learning containers on CPU only machines.

Fixes #9552 

None

## Before submitting

- [x ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/utilities/xla_device.py
+++ b/pytorch_lightning/utilities/xla_device.py
@@ -72,7 +72,8 @@ class XLADeviceUtils:
         # `xm.get_xla_supported_devices("TPU")` won't be possible.
         if xm.xrt_world_size() > 1:
             return True
-        return len(xm.get_xla_supported_devices("TPU")) > 0
+        tpus = xm.get_xla_supported_devices("TPU")
+        return False if tpus is None else len(tpus) > 0
 
     @staticmethod
     def xla_available() -> bool:


ckchow:
I'm actually not sure how to write a proper test for this. Seems like it would have to pull an XLA container.

kaushikb11:
> I'm actually not sure how to write a proper test for this. Seems like it would have to pull an XLA container.

We have TPU CI tests in place, But for this, we would require to pull an XLA container with no TPUs attached. Could you please add a comment for this in the code?

Do you think "the guard against None and return an empty list" should be added to the XLA's API itself? cc: @JackCaoG

awaelchli (approved):

SeanNaren (approved):
Thanks for this! At the least I think add a `todo` saying that this needs to be tested in a XLA supported container without TPUs!

kaushikb11 (changes requested):
Blocking this merge for now. I have asked the XLA team if this change could instead be introduced on the pt/xla's side

JackCaoG:
@kaushikb11 so `get_xla_supported_devices` will never return a empty list, so maybe we can just check if `get_xla_supported_devices('TPU') != None`

ckchow:
I feel like somebody might change `get_xla_supported_devices` to return empty lists in the future (i.e. if I were a contributor to that codebase I would feel tempted to do so), so I think checking for `None` and `len() > 0` explicitly is safer. I guess it's also correct to check `tpus != None and tpus`, but I always forget that empty lists are false in python so that feels a little unexpected.

JackCaoG:
I could try to change the pt/xla code to return an empty list, but 
1. That will also require me to change a couple other places in pt/xla
2. This is BC breaking, if there are users/frameworks that explicitly checking for `!=None` that will also have to change.

so I will prefer to leave the `return None if no device found` logic

pytorch_lightning/utilities/xla_device.py
```diff
@@ -72,7 +72,8 @@ def _is_device_tpu() -> bool:
         # `xm.get_xla_supported_devices("TPU")` won't be possible.
         if xm.xrt_world_size() > 1:
             return True
-        return len(xm.get_xla_supported_devices("TPU")) > 0
+        tpus = xm.get_xla_supported_devices("TPU")
+        return False if tpus is None else len(tpus) > 0
```
carmocca, line 76: This should work for both `None` and `[]`

```suggestion
        return bool(xm.get_xla_supported_devices("TPU"))
```

carmocca: or

```python
return bool(xm.xrt_world_size()) or bool(xm.get_xla_supported_devices("TPU"))
```

kaushikb11:
Thanks @ckchow for the contribution! We always look forward to improving our compatibility with TPUs! :)

Issue #7938, opened 2021-06-11
TODO tracker for ongoing training loop refactors


awaelchli:
## 📚 TODOs around the loop refactor

This issue tracks the work planned for and after the loop refactor #7700. 
Core contributors feel free to update the description here directly.

### Hygiene and code quality

- [ ] Typing for full mypy compatibility (see ignored files in setup.cfg)
- [x] Documentation for the public interface + address https://github.com/PyTorchLightning/pytorch-lightning/pull/8065#discussion_r656160991
- [x] Evaluate which methods should be public and which should be protected
- [x] Rename Trainer.train_loop to Trainer.fit_loop and deprecate the former access. 
- [x] Formalize how arguments are forwarded to lower loops (especially in `connect`

### Feature integration and follow-up refactors
- [x] Extract a OptimizerLoop, potentially a loop for TBPTT  #9191
    - [x] extract utility for `_run_optimization` 
    - [x] for manual optimization, replace closure call with direct call to `_training_step()`
- [x] Integrate progress tracking #6429, this will also increase clarity and minimize attack surface for bugs!
- [x] Determine who should own the dataloaders and responsibility of resetting them https://github.com/PyTorchLightning/pytorch-lightning/pull/7871#discussion_r649680923. Can references fall out of sync? https://github.com/PyTorchLightning/pytorch-lightning/pull/7871#discussion_r649680539
- [x] Add a teardown, let loops move their results to CPU instead of LoggerConnector: https://github.com/PyTorchLightning/pytorch-lightning/pull/8245
- [x] The `EvaluationDataLoaderLoop.predictions` are currently accumulated inside the inner evaluation loop instead of per epoch and then combined over all dataloaders. #8850 
- [x] Rename inner loops to epoch_loop, batch_loope etc to distinguish top level loop access. Also propose file structure organizing these. Resolved by #8077

### Customization
- [x] Formalize how loops get instantiated, set on the trainer, and connected to each other. #8226 

### Miscellaneous
- Several todo's are annotated as TODO(@awaelchli) 
- Some todo's are annotated as TODO(@justusschock) as well. 

cc: @justusschock @carmocca @ananthsub @tchaton

carmocca:
Closing as everything looks done here

PR #9191, opened 2021-08-29
extract optimizer loop

awaelchli:
## What does this PR do?

Part of #7938

Extract an OptimizerLoop from the existing TrainingBatchLoop. This cleanly separates the code path for automatic and manual optimization. Automatic optimization now completely lives inside the OptimizerLoop.

### Previously:
(details left out)

```python
if lightning_module.automatic_optimization:
    for opt_idx, opt in enumerate(trainer.optimizers):
        self._run_optimization(batch, opt_idx, opt)
else:
     self._run_optimization(batch, 0, None)

```
The `run_optimization` had to handle both code paths for automatic and manual optimization, making it hard to follow, debug, and extend.

### Now:
(details left out)

```python
if lightning_module.automatic_optimization:
    self.optimizer.run(batch, optimizers)
else:
     self._run_manual_optimization(batch)
```
The code path branches here cleanly. The new OptimizerLoop handles automatic optimization completely and the manual optimization path is kept isolated. This greatly improves readability, debugging, and extensibility for loop customization.

A follow-up PR will remove some remaining traces of mixed interfaces, we are holding it out from this PR to make it more manageable to review.

Blocked by:
- #9188
- #9192
- #9198

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
[...]

PR #9581, opened 2021-09-17
move get_active_optimizers to utilities

awaelchli:
## What does this PR do?

Part of #7938 

Moves the `get_active_optimizers()` method to the utilities. It is currently used by multiple loops and after #9490 only in one loop, but there is no reason why it should be a method on any of the loops as loops don't own optimizers, nor should loop customization make any assumptions which loop has this method. 

## Before submitting
- [ ] Did you verify new and existing tests pass locally with your changes?

## PR review

## Did you have fun?


pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -374,3 +377,7 @@ def _save_loggers_on_train_batch_end(self) -> None:
         should_flush_logs = self.trainer.logger_connector.should_flush_logs
         if should_flush_logs and self.trainer.is_global_zero and self.trainer.logger is not None:
             self.trainer.logger.save()
+
+    def _num_active_optimizers(self, batch_idx: Optional[int] = None) -> int:
```
awaelchli, line 381: this can be removed once unused after #9490

pytorch_lightning/loops/batch/training_batch_loop.py
```diff
@@ -41,21 +40,13 @@ def __init__(self) -> None:
...
     @property
     def done(self) -> bool:
         """Returns if all batch splits have been processed already."""
         return len(self._remaining_splits) == 0
 
-    @property
-    def optimizer_freq_cumsum(self) -> int:
-        """Returns the cumulated sum of optimizer frequencies."""
-        if self._optimizer_freq_cumsum is None:
-            self._optimizer_freq_cumsum = np.cumsum(self.trainer.optimizer_frequencies)
```
awaelchli, line 56: this value no longer gets cached. we recompute it every epoch because now the function is a utility. is this a problem?

carmocca: If we still want to cache it, we could add an `@lru_cache` decorator

pytorch_lightning/loops/utilities.py
```diff
@@ -139,3 +141,25 @@ def _block_parallel_sync_behavior(trainer: "pl.Trainer", block: bool = True) ->
...
+    if not frequencies:
+        # call training_step once per optimizer
+        return list(enumerate(optimizers))
+
+    freq_cumsum = np.cumsum(frequencies)
+    optimizers_loop_length = freq_cumsum[-1]
+    current_place_in_loop = batch_idx % optimizers_loop_length
+
+    # find optimizer index by looking for the first {item > current_place} in the cumsum list
+    opt_idx = int(np.argmax(freq_cumsum > current_place_in_loop))
```
carmocca, line 164: since `freq_cumsum` will be sorted, this would be more efficient

```suggestion
    opt_idx = np.searchsorted(freq_cumsum, current_place_in_loop)
```

ref: https://stackoverflow.com/a/48360950

pytorch_lightning/loops/batch/training_batch_loop.py
awaelchli: I tried but I run into the issue that list inputs can't be cached because they are not hashable

carmocca: yeah... would need to use a tuple, up to you.

pytorch_lightning/loops/utilities.py
awaelchli: this gave me a failing test: `tests/loops/optimization/test_optimizer_loop.py::test_optimizer_frequencies`
could potentially be added in a separate PR, I want to keep this a refactor

pytorch_lightning/loops/batch/training_batch_loop.py
```diff
@@ -123,7 +114,9 @@ def advance(self, batch, batch_idx):
 
         if self.trainer.lightning_module.automatic_optimization:
             # in automatic optimization, hand over execution to the OptimizerLoop
-            batch_outputs = self.optimizer_loop.run(split_batch, self.get_active_optimizers(batch_idx), batch_idx)
+            # TODO: total_batch_idx here??
```
carmocca, line 117: Any thoughts about this TODO? Isn't batch idx safer?

carmocca (approved):

pytorch_lightning/loops/batch/training_batch_loop.py
awaelchli: forgot to remove this todo. 
yes, need to leave it here as it was like this before. but in the other location where get_active_optimizers is used we are passing in total_batch_idx. that's why I was wondering. 

the choice here depends whether we expect the optimizer frequencies to apply globally across epoch boundaries or be locally defined/rest. 

I tried total batch idx but we don't have any tests that would check this or fail due to a mismatch.

tchaton (approved):
LGTM !

pytorch_lightning/loops/batch/training_batch_loop.py
```diff
@@ -175,21 +163,3 @@ def _update_running_loss(self, current_loss: Tensor) -> None:
 
         # reset for next set of accumulated grads
         self.accumulated_loss.reset()
-
-    def get_active_optimizers(self, batch_idx: Optional[int] = None) -> List[Tuple[int, Optimizer]]:
```
Borda, line 179: Is any deprecation needed here? till now it was a public API, since now it is a protected function...

awaelchli: I can do that if it is desired. However, at this point we changed already so much in the loops from 1.4 to 1.5 that this one is not going to matter. For example, we moved a lot of public API from one loop to the other in #9191 without deprecations, because it's not really possible to even keep backward compatibility there if we want to move fast.  

@carmocca @justusschock opinions on this?

carmocca: i'm on the yolo side

ananthsub (approved):

PR #9584, opened 2021-09-17
Add `clip_gradients` hook in `LightningModule

### LINKED ISSUE #6346: [RFC] Gradient clipping hooks in the LightningModule
## 🚀 Feature

Add clipping hooks to the LightningModule

### Motivation

It's currently very difficult to change the clipping logic

### Pitch

```python
class LightningModule:
    def on_before_clip_gradients():
        ...

    def clip_gradients(self, optimizer):
        ...
    
    def on_after_clip_gradients():
        ...
```

The default implementation would be the same as we currently provide, where the trainer clipping parameters are used.

Maybe those would be deprecated in favor of LightningModule properties.

```python
class LightningOptimizer
    def step(closure=closure)
        if closure is None:
            closure = do_nothing_closure
        def wrapper_closure()
            closure()
            self._trainer.call_hook("on_before_clip_gradients")
            self._trainer.call_hook("clip_gadients")
            self._trainer.call_hook("on_after_clip_gadients")
        self.optimizer.step(closure=wrapper_closure)
```

Need to evaluate the limitations since clipping is currently tied to plugins

### Additional context

This would fix #5096, https://github.com/PyTorchLightning/pytorch-lightning/pull/6123#discussion_r587415267, https://github.com/PyTorchLightning/pytorch-lightning/issues/5671, https://github.com/PyTorchLightning/pytorch-lightning/issues/5982, and allow easily implementing new clipping techniques without having to merge them into Lightning

cc: @rohitgr7 who has been pushing for this for a while

rohitgr7:
## What does this PR do?

Fixes #6346

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/core/lightning.py
+++ b/pytorch_lightning/core/lightning.py
@@ -36,7 +36,12 @@ from pytorch_lightning.core.mixins import DeviceDtypeModuleMixin, Hyperparameter
 from pytorch_lightning.core.optimizer import LightningOptimizer
 from pytorch_lightning.core.saving import ModelIO
 from pytorch_lightning.trainer.connectors.logger_connector.fx_validator import _FxValidator
-from pytorch_lightning.utilities import _TORCH_SHARDED_TENSOR_AVAILABLE, rank_zero_deprecation, rank_zero_warn
+from pytorch_lightning.utilities import (
+    _TORCH_SHARDED_TENSOR_AVAILABLE,
+    GradClipAlgorithmType,
+    rank_zero_deprecation,
+    rank_zero_warn,
+)
 from pytorch_lightning.utilities.apply_func import apply_to_collection, convert_to_tensors
 from pytorch_lightning.utilities.cloud_io import get_filesystem
 from pytorch_lightning.utilities.distributed import distributed_available, sync_ddp
@@ -1472,6 +1477,24 @@ class LightningModule(
                             param.requires_grad = self._param_requires_grad_state[param]
         # save memory
         self._param_requires_grad_state = {}
+
+    def clip_gradients(
+        self,
+        optimizer: Optimizer,
+        optimizer_idx: int,
+        gradient_clip_val: Union[int, float] = 0.0,
+        gradient_clip_algorithm: str = "norm",
+    ):
+        gradient_clip_val = gradient_clip_val or self.trainer.gradient_clip_val
+        gradient_clip_algorithm = gradient_clip_algorithm or self.trainer.gradient_clip_algorithm
+
+        if not isinstance(gradient_clip_val, (int, float)):
+            raise MisconfigurationException("`gradient_clip_val` should be either an int or a float")
+
+        if gradient_clip_algorithm not in list(GradClipAlgorithmType):
+            raise MisconfigurationException(f"`gradient_clip_algorithm` should be in {list(GradClipAlgorithmType)}")
+
+        self.trainer.accelerator.clip_gradients(optimizer, gradient_clip_val, gradient_clip_algorithm)
 
     def optimizer_step(
         self,
--- a/pytorch_lightning/core/lightning.py
+++ b/pytorch_lightning/core/lightning.py
@@ -1485,6 +1485,23 @@ class LightningModule(
         gradient_clip_val: Union[int, float] = 0.0,
         gradient_clip_algorithm: str = "norm",
     ):
+        """Perform Gradient Clipping for the optimizer parameters. Called before :meth:`optimizer_step`.
+
+        Args:
+            optimizer: Current optimizer being used. ``None`` if using manual optimization.
+            optimizer_idx: Index of the current optimizer being used. ``None`` if using manual optimization.
+            gradient_clip_val: The value at which to clip gradients. Passing ``gradient_clip_val=0`` disables gradient
+                clipping.
+            gradient_clip_algorithm: The gradient clipping algorithm to use. Pass ``gradient_clip_algorithm="value"``
+                for clip_by_value, and ``gradient_clip_algorithm="norm"`` for clip_by_norm.
+
+        Example::
+
+            # Perform gradient clipping on discriminator (optimizer_idx=1)
+            def clip_gradients(self, optimizer, optimizer_idx, gradient_clip_val, gradient_clip_algorithm):
+                if optimizer_idx == 1:
+                    super().clip_gradients(optimizer, optimizer_idx, gradient_clip_val, gradient_clip_algorithm)
+        """
         gradient_clip_val = gradient_clip_val or self.trainer.gradient_clip_val
         gradient_clip_algorithm = gradient_clip_algorithm or self.trainer.gradient_clip_algorithm
 
--- a/pytorch_lightning/loops/optimization/optimizer_loop.py
+++ b/pytorch_lightning/loops/optimization/optimizer_loop.py
@@ -240,7 +240,7 @@ class OptimizerLoop(Loop):
 
         if not self.trainer.fit_loop._should_accumulate():
             # track gradients
-            grad_norm_dict = self._track_and_norm_grad(optimizer=optimizer)
+            grad_norm_dict = self._track_and_norm_grad(optimizer=optimizer, opt_idx=opt_idx)
             if grad_norm_dict:
                 self.trainer.lightning_module._current_fx_name = "on_after_backward"
                 self.trainer.lightning_module.log_grad_norm(grad_norm_dict)
@@ -478,7 +478,7 @@ class OptimizerLoop(Loop):
 
         return result
 
-    def _track_and_norm_grad(self, optimizer: torch.optim.Optimizer) -> Dict[str, float]:
+    def _track_and_norm_grad(self, optimizer: torch.optim.Optimizer, opt_idx: int) -> Dict[str, float]:
         """Tracks gradient norms and clips the gradients of all parameters optimized by the current optimizer.
 
         Args:
@@ -492,7 +492,5 @@ class OptimizerLoop(Loop):
             grad_norm_dict = grad_norm(self.trainer.lightning_module, self.trainer.track_grad_norm)
 
         # clip gradients
-        self.trainer.accelerator.clip_gradients(
-            optimizer, self.trainer.gradient_clip_val, gradient_clip_algorithm=self.trainer.gradient_clip_algorithm
-        )
+        self.trainer.lightning_module.clip_gradients(optimizer, opt_idx)
         return grad_norm_dict


rohitgr7:
@carmocca should we deprecate `gradient_clip_val` and `gradient_clip_hook` arguments from Trainer? Also initial thoughts on this please :)

ananthsub:
@rohitgr7 imo we should discuss this more on the issue. I think there are some design aspects that have to be worked out first

pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -478,7 +478,7 @@ def _training_step(self, split_batch: Any, batch_idx: int, opt_idx: int) -> Clos
 
         return result
 
-    def _track_and_norm_grad(self, optimizer: torch.optim.Optimizer) -> Dict[str, float]:
+    def _track_and_norm_grad(self, optimizer: torch.optim.Optimizer, opt_idx: int) -> Dict[str, float]:
```
rohitgr7, line 481: off topic.. but there is a little inconsistency with argument name `opt_idx` here. In the hooks open to the users we have `optimizer_idx` but internally at some places, we have `opt_idx`.

pytorch_lightning/core/lightning.py
```diff
@@ -1473,6 +1478,43 @@ def untoggle_optimizer(self, optimizer_idx: int):
...
+    ):
+        """Perform Gradient Clipping for the optimizer parameters. Called before :meth:`optimizer_step`.
+
+        Args:
+            optimizer: Current optimizer being used. ``None`` if using manual optimization.
+            optimizer_idx: Index of the current optimizer being used. ``None`` if using manual optimization.
+            gradient_clip_val: The value at which to clip gradients. Passing ``gradient_clip_val=0`` disables gradient
+                clipping.
+            gradient_clip_algorithm: The gradient clipping algorithm to use. Pass ``gradient_clip_algorithm="value"``
+                for clip_by_value, and ``gradient_clip_algorithm="norm"`` for clip_by_norm.
```
awaelchli, line 1496: If it's a hook, the user doesn't call it, yet the docs here say that "passing =0 disables gradient clipping"? 
Meanwhile, this PR also deprecates the trainer arguments, so how is this feature supposed to be used now?

awaelchli (commented):
I don't understand this PR. It seems to be incomplete. 
Today I am doing this: 

```
Trainer(gradient_clip_val=15.0)
```
or 
```
python train.py --gradient_clip_val 15.0
```

This PR aims at deprecating this option. How do I achieve the above now? There seems to be a hook attempting to replace this functionality, but this is missing the point why the Trainer argument exists in the first place. 99.9% of the time the user does not need to implement gradient clipping because Lightning can do it. But by removing this now, you force the user to  100% of the time have to implement a hook? IMO you are removing a perfectly fine feature and I vote against going in this direction. If configurability is desired, it should be entirely optional.

rohitgr7:
so for now I have implemented option 2 from here: https://github.com/PyTorchLightning/pytorch-lightning/issues/6346#issuecomment-925914953

feel free to put your suggestions/recommendations here or on the issue :)

tchaton (approved):
LGTM ! Nice work :)

pytorch_lightning/core/lightning.py
```diff
@@ -1473,6 +1478,64 @@ def untoggle_optimizer(self, optimizer_idx: int):
         # save memory
         self._param_requires_grad_state = {}
 
+    def clip_gradients(self, optimizer: Optimizer, gradient_clip_val: Union[int, float], gradient_clip_algorithm: str):
```
carmocca, line 1481: Can you scan the docs for a place where to advertise this hook?

Also add it to https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_module.html#hooks

pytorch_lightning/core/lightning.py
```diff
@@ -1473,6 +1478,64 @@ def untoggle_optimizer(self, optimizer_idx: int):
...
+        """Handles Gradient Clipping internally.
+
+        Args:
+            optimizer: Current optimizer being used.
+            gradient_clip_val: The value at which to clip gradients.
+            gradient_clip_algorithm: The gradient clipping algorithm to use. Pass ``gradient_clip_algorithm="value"``
+                for clip_by_value, and ``gradient_clip_algorithm="norm"`` for clip_by_norm.
```
carmocca, line 1488:
```suggestion
                to clip by value, and ``gradient_clip_algorithm="norm"`` to clip by norm.
```

pytorch_lightning/core/lightning.py
```diff
@@ -1473,6 +1478,64 @@ def untoggle_optimizer(self, optimizer_idx: int):
...
+
+        Note:
```
carmocca, line 1490: Mind removing the "Note" and having it before `Args:`? It's an important detail

```python
        """Handles Gradient Clipping internally.

        <Move here>

        Args:
            ...
```

carmocca (approved):

pytorch_lightning/core/lightning.py
```diff
@@ -1473,6 +1478,64 @@ def untoggle_optimizer(self, optimizer_idx: int):
...
+                will be available here.
+            gradient_clip_algorithm: The gradient clipping algorithm to use. By default value
+                passed in Trainer will be available here.
+
+        Example::
+
+            # Perform gradient clipping on discriminator (optimizer_idx=1) in GAN
+            def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_val, gradient_clip_algorithm):
+                if optimizer_idx == 1:
+                    # lightning will handle the gradient clipping
```
awaelchli, line 1528:
```suggestion
                    # Lightning will handle the gradient clipping
```

pytorch_lightning/core/lightning.py
```diff
@@ -1473,6 +1478,64 @@ def untoggle_optimizer(self, optimizer_idx: int):
         # save memory
         self._param_requires_grad_state = {}
 
+    def clip_gradients(self, optimizer: Optimizer, gradient_clip_val: Union[int, float], gradient_clip_algorithm: str):
+        """Handles Gradient Clipping internally.
```
awaelchli, line 1482:
```suggestion
        """Handles gradient clipping internally.
```
the capitalization seems arbitrary here, or is it referred to as a name somewhere else too?

awaelchli (approved):
while this addition is probably needed and desired from a practical point of view, it also shows where Lightning's automatic optimization breaks down imo, because the desire of configuration here means ignoring settings that were provided through the trainer, which then in turn also challenges reproducibility.

rohitgr7:
> while this addition is probably needed and desired from a practical point of view, it also shows where Lightning's automatic optimization breaks down imo, because the desire of configuration here means ignoring settings that were provided through the trainer, which then in turn also challenges reproducibility.

settings provided in the trainer will still work and one doesn't need to change their code at all since those trainer args are passed here by default. This is completely optional for those who want to configure it as per their requirement. Doesn't break automatic optimization.

pytorch_lightning/core/lightning.py
rohitgr7: na.. just looks better in docs.

awaelchli:
Yes it does! Let me make it explicit with an example:

```
def configure_gradient_clipping(self,optimizer, optimizer_idx, gradient_clip_val, gradient_clip_algorithm)
    if optimizer_idx == 0:
        self.clip_gradients(optimizer, optimizer_idx, 10, "norm")
    else:
        self.clip_gradients(optimizer, optimizer_idx, 20, "norm")
```

Now:

```
trainer = Trainer(gradient_clip_val=30)  # has no effect!!!
```

Same issue when configuration is done externally:

```
python train.py --gradient_clip_val=30 
```

This is where Lightning reaches the limit and essentially breaks. This is an issue we will face more in the future.

rohitgr7:
> n issue we will face more in th

okay.. so for this what we can do is make these two trainer arguments None by default. Now inside clip_gradients we can check:
1. if they are not None and different value is passed, then we can raise a warning/misconfigexception.
2. if they are not None and same value is passed, then it's fine.
3. if they are None and different value is passed, use the new value
4. if they are None, and no value is passed, then use the current default values that we have in Trainer (0.0 and 'norm')

wdyt @awaelchli ?

docs/source/common/lightning_module.rst
```diff
@@ -1195,6 +1195,7 @@ for more information.
             on_after_backward()
 
             on_before_optimizer_step()
+            configure_gradient_clipping()
```
awaelchli, line 1198: good that you added it here. because we discover here something not so nice. 

the `on_before_optimizer_step()` is now not running anymore directly before the optimizer step. do we want honor that and switch it up?

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
     ):
         optimizer = optimizer.optimizer
         optimizer.step(closure=optimizer_closure)
+
+-----
+
+Configure Gradient Clipping
```
awaelchli, line 522: the titles in this file are not capitalized

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
...
+---------------------------
+To configure custom gradient clipping, consider overriding
+the :meth:`~pytorch_lightning.core.lightning.LightningModule.configure_gradient_clipping` function.
```
awaelchli, line 525: it's a method!

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
...
+By default it will get arguments :attr:`~pytorch_lightning.trainer.Trainer.gradient_clip_val` and
+:attr:`~pytorch_lightning.trainer.Trainer.gradient_clip_algorithm` from ``Trainer`` in their respective
```
awaelchli, line 527:
```suggestion
By default it will get arguments :attr:`~pytorch_lightning.trainer.trainer.Trainer.gradient_clip_val` and
:attr:`~pytorch_lightning.trainer.trainer.Trainer.gradient_clip_algorithm` from ``Trainer`` in their respective
```

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
...
         optimizer = optimizer.optimizer
         optimizer.step(closure=optimizer_closure)
+
+-----
+
+Configure Gradient Clipping
+---------------------------
+To configure custom gradient clipping, consider overriding
+the :meth:`~pytorch_lightning.core.lightning.LightningModule.configure_gradient_clipping` function.
+By default it will get arguments :attr:`~pytorch_lightning.trainer.Trainer.gradient_clip_val` and
```
awaelchli, line 526: it will always get that, not just "by default".

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
...
+:attr:`~pytorch_lightning.trainer.Trainer.gradient_clip_algorithm` from ``Trainer`` in their respective
+fields and lightning will handle gradient clipping on its own. In case you want to set different values
```
awaelchli, line 528:
```suggestion
:attr:`~pytorch_lightning.trainer.Trainer.gradient_clip_algorithm` from ``Trainer`` in their respective
fields and Lightning will handle gradient clipping on its own. In case you want to set different values
```
fields?
You mean it will just get the values of the Trainer arguments as input. Can it be formulated without mentioning any "fields"?

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
...
+for your arguments of your choice and let Lightning handle the gradient clipping, you can use the inbuilt
+:meth:`~pytorch_lightning.core.lightning.LightningModule.clip_gradients` function and pass the arguments
+along with your optimizer.
+
+.. note::
+    Make sure to not override :meth:`~pytorch_lightning.core.lightning.LightningModule.clip_gradients`
+    function. If you want to customize gradient clipping, consider using
+    :meth:`~pytorch_lightning.core.lightning.LightningModule.configure_gradient_clipping` function.
+
+For example, here we will apply gradient clipping only to optimizer A.
```
awaelchli, line 538:
```suggestion
For example, here we will apply gradient clipping only to the gradients associated with optimizer A.
```

docs/source/common/optimizers.rst
```diff
@@ -516,3 +516,49 @@ to perform a step, Lightning won't be able to support accelerators and precision
...
+            # Lightning will handle the gradient clipping
+            self.clip_gradients(
+                optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
+            )
+        elif optimizer_idx == 1:
+            custom_gradient_clip_val = 1e-2
+            for pg in optimizer.param_groups:
+                for p in pg["params"]:
+                    p.grad[p.grad > custom_gradient_clip_val] = custom_gradient_clip_val
+                    p.grad[p.grad <= 0] = 0
```
awaelchli, line 564: that is a very funny way of clipping gradients haha
it will actually change the direction of the gradient
SGD will not be happy about this lol

awaelchli: In all seriousness, I wouldn't put such an example.
maybe something like this could be more useful:

N = 2
# less clipping for parameters of optimizer 1
elif optimizer_idx == 1:
       self.clip_gradients(
                optimizer, gradient_clip_val=(N * gradient_clip_val), gradient_clip_algorithm=gradient_clip_algorithm
            )

or choose a different algorithm

docs/source/common/lightning_module.rst
awaelchli: EDIT: I see it may not be so easy to do it here. 
But still, I think it's slightly problematic

rohitgr7: yes.. i noticed.. so didn't change the sequence here.. maybe in another PR.

PR #9585, opened 2021-09-17
[wip] Support serialized checkpoint loading

### LINKED ISSUE #9406: Support serialized checkpoint loading
### Motivation

Currently, all processes load the checkpoint at the same time. This can quickly OOM for large models. These use cases, especially with things like mixture of experts, might require serialized loading of checkpoint dicts across ranks (ie load the checkpoint one rank at a time per node). Could we enable this for DDP? 

Prior work: https://github.com/PyTorchLightning/pytorch-lightning/pull/8515

### Pitch

This would be controlled per training type plugin. Example pseudocode: https://gist.github.com/ananthsub/4ceedff56b2049a63bbb05ccd283b919

To work through: 
1. Should the TrainingTypePlugin have responsibility of calling `on_load_checkpoint` instead of the Trainer/connector? This would make sense as the TTP "owns" the LightningModule inside of the trainer, and since it already offers `load_model_state_dict`: https://github.com/PyTorchLightning/pytorch-lightning/blob/41ba639859cf6c6bf319eb33e5b3394504315962/pytorch_lightning/plugins/training_type/training_type_plugin.py#L159-L160

DeepSpeed already eschews most of the checkpoint connector logic when it comes to loading the lightning module state. This could be a gap for metrics, and this means we could be calling `on_load_checkpoint` multiple times with certain plugins. In my opinion, this points to all LightningModule state load/save/alterations should sit inside of the training type plugin. 

2. Does loading one process per node run into other issues with collectives potentially being used in `LightningModule.on_load_checkpoint` ?

### Additional context

<sub>

</sub>

jjenniferdai:
## What does this PR do?

Fixes #9406 

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -153,11 +153,17 @@ def results(self) -> Optional[Union[_EVALUATE_OUTPUT, _PREDICT_OUTPUT]]:
...
+        self.on_load_checkpoint(checkpoint)
+        self.load_model_state_dict(checkpoint)
+        return checkpoint
 
     def load_model_state_dict(self, checkpoint: Mapping[str, Any]) -> None:
         self.lightning_module.load_state_dict(checkpoint["state_dict"])
 
+    def on_load_checkpoint(self, checkpoint: Mapping[str, Any]) -> None:
+        self.lightning_module.on_load_checkpoint(checkpoint)
+
```
ananthsub, line 166: I don't think we should expose an `on_load_checkpoint` method separately here. instead, do you think the lightning module hook could be called directly in `TrainingTypePlugin.load_checkpoint` ?

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -535,3 +536,13 @@ def teardown(self) -> None:
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+    def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
+        rank_zero_info(f"DistributedDataParallel has {self.num_processes} processes. Serializing to avoid CPU OOMs.")
+        for current_worker in range(self.num_processes):
+            if self.local_rank == current_worker and torch.distributed.is_initialized():
```
ananthsub, line 543:
```suggestion
            if self.local_rank == current_worker:
```
we ought to fail loudly if torch distributed isn't initialized. otherwise nothing will be loaded

pytorch_lightning/trainer/trainer.py
```diff
@@ -957,6 +954,8 @@ def _run(self, model: "pl.LightningModule") -> Optional[Union[_EVALUATE_OUTPUT,
 
         # check if we should delay restoring checkpoint till later
         if not self.accelerator.restore_checkpoint_after_pre_dispatch:
+            if self._ckpt_path:
+                self._load_checkpoint_weights()
```
ananthsub, line 958: note to reviewer: this must be moved after the `setup_environment` is called to ensure that torch distributed is initialized. it also matches the symmetry in later on, and makes it easier for us to reason about merging these checkpoint loading paths altogether.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -154,12 +148,8 @@ def restore_model(self) -> None:
 
     def restore_model_weights(self, checkpoint_path: Optional[_PATH]) -> None:
         """Restore only the model weights."""
-        checkpoint = self._loaded_checkpoint
         if checkpoint_path is not None:
-            checkpoint = self._load_and_validate_checkpoint(checkpoint_path)
-
-        self.trainer.lightning_module.on_load_checkpoint(checkpoint)
-        self.trainer.training_type_plugin.load_model_state_dict(checkpoint)
```
ananthsub, line 162: this means the state dict isn't loaded anymore. why this change?

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -535,3 +536,13 @@ def teardown(self) -> None:
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+    def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
+        rank_zero_info(f"DistributedDataParallel has {self.num_processes} processes. Serializing to avoid CPU OOMs.")
+        for current_worker in range(self.num_processes):
+            if self.local_rank == current_worker:
+                checkpoint = super().load_checkpoint(checkpoint_path)
+                del checkpoint["state_dict"]
```
ananthsub, line 545: load_model_state_dict should check if `"state_dict"` as a key is in the checkpoint dict. otherwise it'll fail on lookup

pytorch_lightning/trainer/connectors/checkpoint_connector.py
jjenniferdai: it gets called in new line 152 `_load_and_validate_checkpoint` --> `training_type_plugin.load_checkpoint` ?

pytorch_lightning/plugins/training_type/ddp.py
jjenniferdai: what's the expected behavior if `"state_dict"` key does not exist?

ananthsub: got it!

ananthsub: - the plugin can track whether it was the entity that deleted the key from the checkpoint in `load_checkpoint` . if it did, and there's no `"state_dict"` found when calling `load_model_state_dict` the plugin can simply return early
- if it didn't, then the plugin could raise an exception about the key not found as this is unexpected

jjenniferdai: hm do you mean something like the below edit?

ananthsub: yup, but this doesn't need to be on the base training type plugin. it can be initialized on the plugins which actually serialize checkpoints

jjenniferdai:
messed up git, moving to #9605

PR #9586, opened 2021-09-17
[1/4] Add get_device_stats to accelerator interface

### LINKED ISSUE #9032: Revamp Device Stats Logging

Deprecate `log_gpu_memory` from the Trainer constructor

### Motivation

Carrying forward some of the discussion from here: https://github.com/PyTorchLightning/pytorch-lightning/discussions/9006

By default, logging gpu stats is disabled. Users must opt into this by setting. 

Issues today with this argument:
- Decoupling this from the trainer constructor gives us more flexibility to evolve the component which is actually responsible for gathering these stats without worrying about backwards compatibility guarantees for the trainer constructor argument as well. For example, the currently supported values for this flag are nvidia-smi specific. If `nvidia-smi` isn't available, this will fail. But PyTorch now supports getting CUDA memory stats directly: https://github.com/PyTorchLightning/pytorch-lightning/issues/8780 which we could take advantage of.

- Lack of consistency across different devices: We do not offer corresponding flags for TPUs nor IPUs. Creating one flag per device type would lead to further bloat in the trainer.

- There are multiple codepaths Lightning offers to log GPU memory. We have logging both [internal to the trainer](https://github.com/PyTorchLightning/pytorch-lightning/blob/13e64e6a80bc19308e99ff6aedb95ad68968b5b8/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py#L225-L229), as well as a [GPU stats monitor callback](https://github.com/PyTorchLightning/pytorch-lightning/blob/13e64e6a80bc19308e99ff6aedb95ad68968b5b8/pytorch_lightning/callbacks/gpu_stats_monitor.py#L38
). Having both is confusing and can lead to bugs like https://github.com/PyTorchLightning/pytorch-lightning/issues/9010 . 

### Pitch

Instead of offering separate flags & callbacks for different device types, why don't we consolidate this on the Accelerator interface, as this is where all the device-specific information already should be.

1. Offer a method like `get_device_stats` as an abstractmethod on the Accelerator interface. CPU, GPU, TPU, and IPU classes would implement this to return whatever stats are needed. This payload should be a black box to the trainer as long as it conforms to the expected return type. The implementations of this function could rely on the existing utilities we already have. This would ensure that all the device-specific code lives roughly in the same place. The stats returned should be per-device: the calling entities fetching these stats should determine whether or how to aggregate these stats (e.g. log stats from rank 0 only, reduce across ranks and log from rank 0, log the min/max across ranks, or log data from all ranks). 

2. In terms of what actually fetches these stats, we could either:
- offer a generic callback like `DeviceStatsMonitor` to periodically gather and log this data. This callback would consolidate the GPU and XLA ones and provide a general extension point for any new device types Lightning supports via the Accelerator interface. 
- Or, the trainer/loops themselves could periodically gather & log this information.

My preference is to go with the callback-based approach first. It remains the most extensible to customize options like logging frequency (e.g. step based or time-based) .

What do you think? @kaushikb11 @tchaton @awaelchli @carmocca @four4fish @edward-io @yifuwang 

### Additional context

https://github.com/PyTorchLightning/pytorch-lightning/pull/8174#discussion_r660668285
https://github.com/PyTorchLightning/pytorch-lightning/pull/9013#issuecomment-902680473

<sub>

</sub>

daniellepintz:
## What does this PR do?
Adds `get_device_stats` function to accelerator class, and implements it for GPU and TPU. To be called in later PRs
Tested with https://github.com/daniellepintz/pytorch-lightning/pull/2

Part of #9032

Please also see subsequent PRs: 
https://github.com/daniellepintz/pytorch-lightning/pull/2
https://github.com/daniellepintz/pytorch-lightning/pull/3
https://github.com/daniellepintz/pytorch-lightning/pull/4

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/accelerators/accelerator.py
```diff
@@ -436,6 +437,10 @@ def restore_checkpoint_after_pre_dispatch(self) -> bool:
         """
         return self.training_type_plugin.restore_checkpoint_after_pre_dispatch
 
+    def get_device_stats(self, device: Optional[torch.device] = None) -> Dict[str, Any]:
+        """Gets stats for a given device."""
+        pass
```
ananthsub, line 442: this could be an abstractmethod

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +64,58 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+        return stats
+
+    @staticmethod
+    def _get_gpu_ids(device_ids: List[int]) -> List[str]:
+        """Get the unmasked real GPU IDs."""
+        # All devices if `CUDA_VISIBLE_DEVICES` unset
+        default = ",".join(str(i) for i in range(torch.cuda.device_count()))
+        cuda_visible_devices: List[str] = os.getenv("CUDA_VISIBLE_DEVICES", default=default).split(",")
+        return [cuda_visible_devices[device_id].strip() for device_id in device_ids]
+
```
ananthsub, line 118: let's call the memory utils here: https://github.com/PyTorchLightning/pytorch-lightning/blob/c7451b3ccf742b0e8971332caf2e041ceabd9fe8/pytorch_lightning/utilities/memory.py#L126-L152

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +64,58 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+                ("memory.used", "MB"),
+                ("memory.free", "MB"),
+                ("utilization.memory", "%"),
+                ("fan.speed", "%"),
+                ("temperature.gpu", "°C"),
+                ("temperature.memory", "°C"),
+            ]
+            gpu_stats = self._get_gpu_stats([k for k, _ in gpu_stat_keys])
+            device_stats = self._parse_gpu_stats(self._device_ids, gpu_stats, gpu_stat_keys)
+            return device_stats
```
ananthsub, line 83: this doesn't use the `device` passed in anywhere?

pytorch_lightning/accelerators/tpu.py
```diff
@@ -59,3 +59,18 @@ def _move_optimizer_state(self, device: Optional[torch.device] = None) -> None:
         for opt in self.optimizers:
             for p, v in opt.state.items():
                 opt.state[p] = apply_to_collection(v, torch.Tensor, move_data_to_device, self.root_device)
+
+    def get_device_stats(self, device: Optional[torch.device] = None) -> Dict[str, Any]:
+        """Gets stats for the given TPU device."""
+        device_stats = {}
+        memory_info = xm.get_memory_info(device)
```
ananthsub, line 66: memory_info accepts a string: https://pytorch.org/xla/release/1.9/index.html#torch_xla.core.xla_model.get_memory_info

which is inconsistent with the typehint or `device` above

pytorch_lightning/accelerators/tpu.py
```diff
@@ -59,3 +59,18 @@ def _move_optimizer_state(self, device: Optional[torch.device] = None) -> None:
...
+
+        free_memory = memory_info["kb_free"]
+        peak_memory = memory_info["kb_total"] - free_memory
+
+        free_memory = self.training_type_plugin.reduce(free_memory) * 0.001
+        peak_memory = self.training_type_plugin.reduce(peak_memory) * 0.001
```
ananthsub, line 72: @kaushikb11 n00b question: why do we need the reduce here?

pytorch_lightning/accelerators/gpu.py
daniellepintz: it does in line 70: `return torch.cuda.memory_stats(device=device)`

pytorch_lightning/accelerators/gpu.py
daniellepintz: `get_gpu_memory_map` only queries for `memory.used`, not the other fields like:
```
("utilization.gpu", "%"),
                ("memory.used", "MB"),
                ("memory.free", "MB"),
                ("utilization.memory", "%"),
                ("fan.speed", "%"),
                ("temperature.gpu", "°C"),
                ("temperature.memory", "°C"),
```

are you saying we only need `memory.used`?

pytorch_lightning/accelerators/gpu.py
ananthsub: but afterward, even if someone isn't using PyTorch 1.8+, they'd expect the `device` argument to still be respected and for the method to get stats only for that particular device

pytorch_lightning/accelerators/gpu.py
daniellepintz: The function `_get_gpu_stats` was taken from gpu_stats_monitor. I was actually thinking we could deprecate `get_gpu_memory_map` since it is almost the same as `_get_gpu_stats`

ananthsub: I see, I do think we ought to merge the utilities somewhere and not have them spread across the codebase like this. @awaelchli @carmocca do you think having a `utilities/gpu.py` makes sense to capture this? then the GPU accelerator simply calls the utils function there.

daniellepintz: I updated the PR with a utility function `_get_nvidia_gpu_stats`. Personally I think it makes sense to just keep it at the bottom of gpu.py, since it's a short file and I don't think the func will be used in other places, but can also move it elsewhere.

My plan is to mark `get_gpu_memory_map` and `get_memory_profile` as deprecated at the same as marking `log_gpu_memory` as deprecated, and then remove them all in 1.7.

pytorch_lightning/accelerators/gpu.py
daniellepintz: good point, I was a bit confused. I updated the PR so that it does that, and tested with https://github.com/daniellepintz/pytorch-lightning/pull/2

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -436,6 +438,11 @@ def restore_checkpoint_after_pre_dispatch(self) -> bool:
         """
         return self.training_type_plugin.restore_checkpoint_after_pre_dispatch
 
+    @abstractmethod
+    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
+        """Gets stats for a given device."""
+        pass
```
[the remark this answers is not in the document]
daniellepintz: thanks - was going to raise NotImplementedError, is that not what we want?

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,57 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+        gpu_stat_keys = [
+            "utilization.gpu",
+            "memory.used",
+            "memory.free",
+            "utilization.memory",
+            "fan.speed",
+            "temperature.gpu",
+            "temperature.memoy",
+        ]
+        gpu_ids = self._get_gpu_id(device.index)
```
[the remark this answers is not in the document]
daniellepintz: my understanding from 

https://github.com/PyTorchLightning/pytorch-lightning/blob/f5608e90d6296fb448b55821b358b15e6cabaf4f/pytorch_lightning/core/mixins/device_dtype_mixin.py

was if device is GPU it will always be of type torch.device not str. But I am a bit unclear on this part - under what conditions would this be a str?

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,78 @@ def set_nvidia_flags(local_rank: int) -> None:
         devices = os.getenv("CUDA_VISIBLE_DEVICES", all_gpu_ids)
         _log.info(f"LOCAL_RANK: {local_rank} - CUDA_VISIBLE_DEVICES: [{devices}]")
 
+    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
+        """Gets stats for the given GPU device.
+
+        Returns:
+        A dictionary mapping the metrics to their values.
```
awaelchli, line 64:
```suggestion
            A dictionary mapping the metrics to their values.
```
needs one tab indentation

awaelchli (approved):

pytorch_lightning/accelerators/tpu.py
tchaton: I guess this is used to compute the mean across the different tpu_cores.

CHANGELOG.md
```diff
@@ -142,6 +142,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `PL_RECONCILE_PROCESS` environment variable to enable process reconciliation regardless of cluster environment settings ([#9389](https://github.com/PyTorchLightning/pytorch-lightning/pull/9389))
 
 
+- Added `get_device_stats` to Accelerator interface and implement it for GPU and TPU ([#9586](https://github.com/PyTorchLightning/pytorch-lightning/pull/9586))
```
tchaton, line 145:
```suggestion
- Added `get_device_stats` to the Accelerator Interface and add its implementation for GPU and TPU ([#9586](https://github.com/PyTorchLightning/pytorch-lightning/pull/9586))
```

tchaton (approved):
LGMT !

tchaton:
Hey @daniellepintz, thanks for linking all the following PRs :)

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -436,6 +438,10 @@ def restore_checkpoint_after_pre_dispatch(self) -> bool:
         """
         return self.training_type_plugin.restore_checkpoint_after_pre_dispatch
 
+    @abstractmethod
+    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
```
justusschock, line 442: What's the point of having this abstract of at least one subclass raises NotImplemented here? Can't we just raise it directly here and only overwrite it where we actually have an implementation?

I get that ideally we would have purely abstract base classes for interface, but personally I'm not a fan of mixing the concepts within a subpackage.

justusschock (approved):

ananthsub (commented):
@daniellepintz it'd be good to also add tests for the GPU and TPU accelerator to confirm that get_device_stats returns something

pytorch_lightning/accelerators/accelerator.py
daniellepintz: That makes sense, I am fine with doing that. cc @ananthsub

daniellepintz:
> noob question for PR 2/4. Should we differentiate `on_train_batch_start` and `on_train_batch_end` ?
@tchaton this was modeled after GPUStatsMonitor where we log in both places. If people think it makes sense to only log in one place that is fine as well, as it can always be customized by the user

daniellepintz:
@ananthsub I believe any tests I would write would be similar to the ones in https://github.com/daniellepintz/pytorch-lightning/pull/2/files, or did you have something else in mind?

ananthsub:
> @ananthsub I believe any tests I would write would be similar to the ones in https://github.com/daniellepintz/pytorch-lightning/pull/2/files, or did you have something else in mind?

the tests there are end to end tests relying on multiple components: the accelerator, callback, and trainer. 

the accelerator tests would test similar functionality but should be more self-contained as a unit test specifically for `get_device_stats`. for example, we could instantiate some large tensors on gpu, and ensure that we get corresponding keys back from the GPU Accelerator as a sanity check. 

having the tests here & in the callback will be helpful to detect if something breaks in one place vs another

pytorch_lightning/accelerators/accelerator.py
ananthsub: sounds good to me

pytorch_lightning/accelerators/tpu.py
ananthsub: I agree with @tchaton - the accelerator ought to report per-device stats here (this is what the docstring says too). So I don't think we should be doing any reductions using the training type. That can be let to the calback/some other agent for how they want to aggregate these stats

@kaushikb11 what do you think?

pytorch_lightning/accelerators/cpu.py
```diff
@@ -11,6 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from typing import Any, Dict, Union
```
ananthsub, line 14: I would add a dummy implementtation of `get_device_stats` just so we don't raise a not implemented error. in the future we could use something like psutil to get per-process stats https://psutil.readthedocs.io/

ananthsub (approved):

daniellepintz:
Does anyone have any suggestions for how to test the TPU-related changes here? It seems that no TPU tests are run in the CI?

pytorch_lightning/accelerators/tpu.py
daniellepintz: sg I removed the reduce

daniellepintz:
Looking at the [output](https://circleci.com/api/v1.1/project/github/PyTorchLightning/pytorch-lightning/130708/output/119/0?file=true&allocation-id=614bd26a5d3161001504dd6d-0-build%2F4359DECC) of the [CI TPU tests]( https://app.circleci.com/pipelines/github/PyTorchLightning/pytorch-lightning/43914/workflows/a7b45a79-2161-4412-b60e-328ed02093e9/jobs/130708) I see two odd things:
1. There are a bunch of RuntimeErrors but it says at the bottom that all the tests passed (this is the case on other PRs as well)
2. I can't find `test_device_stats_tpu` in the output (added in this PR), and it doesn't seem to be running it.

Does anyone have any insight into these issues?

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,78 @@ def set_nvidia_flags(local_rank: int) -> None:
         devices = os.getenv("CUDA_VISIBLE_DEVICES", all_gpu_ids)
         _log.info(f"LOCAL_RANK: {local_rank} - CUDA_VISIBLE_DEVICES: [{devices}]")
 
+    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
+        """Gets stats for the given GPU device.
+
```
Borda, line 62: missing Args in docs, the same in the top-level sinature

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,78 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+            A dictionary mapping the metrics to their values.
+
+        Raises:
+            FileNotFoundError:
+                If nvidia-smi installation not found
+        """
+        if _TORCH_GREATER_EQUAL_1_8:
+            return torch.cuda.memory_stats(device)
+        else:
+            return _get_nvidia_gpu_stats(device)
```
Borda, line 73:
```suggestion
        return _get_nvidia_gpu_stats(device)
```

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,78 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+
     def teardown(self) -> None:
         super().teardown()
         self._move_optimizer_state(torch.device("cpu"))
+
+
+def _get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
```
Borda, line 80: how about passing the device as a string?

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,78 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+        ("temperature.gpu", "°C"),
+        ("temperature.memory", "°C"),
+    ]
+    gpu_stat_keys = [k for k, _ in gpu_stat_metrics]
+    gpu_query = ",".join(gpu_stat_keys)
+
+    gpu_id = _get_gpu_id(device.index)
+    nvidia_smi_path = shutil.which("nvidia-smi")
+    if nvidia_smi_path is None:
+        raise FileNotFoundError("nvidia-smi: command not found")
```
Borda, line 105: just curious what is the case of having GPU support and not SMI? is it with Conda and `cudatoolkit`?
eventually, do we need some min SMI version?

pytorch_lightning/accelerators/tpu.py
```diff
@@ -61,3 +61,19 @@ def _move_optimizer_state(self, device: Optional[torch.device] = None) -> None:
...
+        A dictionary mapping the metrics (free memory and peak memory) to their values.
+        """
+        device_stats = {}
+        memory_info = xm.get_memory_info(device)
+
+        free_memory = memory_info["kb_free"]
+        peak_memory = memory_info["kb_total"] - free_memory
+
+        device_stats["avg. free memory (MB)"] = free_memory
+        device_stats["avg. peak memory (MB)"] = peak_memory
```
Borda, line 78:
```suggestion
        memory_info = xm.get_memory_info(device)
        free_memory = memory_info["kb_free"]
        device_stats = {
            "avg. free memory (MB)": free_memory,
            "avg. peak memory (MB)": memory_info["kb_total"] - free_memory
        }
```

tchaton:
Hey @kaushikb11 Mind looking into @daniellepintz TPU testing ?

pytorch_lightning/accelerators/gpu.py
daniellepintz: hmm, why is that ideal? `_get_gpu_id` takes an int, device.index

pytorch_lightning/accelerators/tpu.py
daniellepintz: nice, thanks!

pytorch_lightning/accelerators/gpu.py
daniellepintz: Hmm not sure about the answer to this. maybe @rohitgr7 knows since he worked on this in the past?

ananthsub: @Borda it is possible for nvidia-smi to not be available in the image. this is the case for one of our clusters. sharing some motivations for why it was removed:

- it has undesired side-effects that we had to patch (like loading the module, creating all the devices because by default it's setuid)
- it causes slowness and potentially deadlocks as we have more than one process doing NVML calls
- it can be confusing, like the version of CUDA shown in its output is the latest supported version as per the nvidia-smi binary, NOT what is actually installed / available on the host
- it requires being in sync'ed between the host and the image, which is a really hard thing to maintain. nvidia-smi requires a lot of NVIDIA packages to be installed, which would be really hard to expose through the container.

rohitgr7: back then I was just testing the `GPUStatsMonitor` on Kaggle/Colab GPU (can't remember exactly) and faced this issue sometimes in their sessions. so added this check as well.

pytorch_lightning/accelerators/cpu.py
```diff
@@ -29,3 +33,7 @@ def setup(self, trainer: "pl.Trainer") -> None:
             raise MisconfigurationException(f"Device should be CPU, got {self.root_device} instead.")
 
         return super().setup(trainer)
+
+    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
+        """Returns dummy implementation for now."""
```
kaushikb11, line 38: > for now?

Are we planning to return any stats in the future?

pytorch_lightning/accelerators/gpu.py
```diff
@@ -53,6 +57,83 @@ def set_nvidia_flags(local_rank: int) -> None:
...
+        if _TORCH_GREATER_EQUAL_1_8:
+            return torch.cuda.memory_stats(device)
+        return _get_nvidia_gpu_stats(device)
+
     def teardown(self) -> None:
         super().teardown()
         self._move_optimizer_state(torch.device("cpu"))
+
+
+def _get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
```
kaushikb11, line 82: Should we move this to `utils`?

tests/accelerators/test_tpu.py
```diff
@@ -0,0 +1,16 @@
+from pytorch_lightning.accelerators import TPUAccelerator
+from pytorch_lightning.plugins import SingleTPUPlugin
+from pytorch_lightning.plugins.training_type import TPUSpawnPlugin
+from tests.helpers.runif import RunIf
+
+
+@RunIf(tpu=True)
+def test_device_stats_tpu(tmpdir):
```
kaushikb11, line 8: This TPU test won't run as this file needs to be mentioned in the test suite for `tpus`. Reference https://github.com/PyTorchLightning/pytorch-lightning/blob/master/dockers/tpu-tests/tpu_test_cases.jsonnet#L38

pytorch_lightning/accelerators/cpu.py
daniellepintz: @ananthsub mentioned "in the future we could use something like psutil to get per-process stats https://psutil.readthedocs.io/" so potentially, but I don't think we have concrete plans. Should I remove that from the comment?

pytorch_lightning/accelerators/gpu.py
daniellepintz: I personally don't think so, since the func is only used in this file, similarly to how this util function is not in `utils`:
https://github.com/PyTorchLightning/pytorch-lightning/blob/64bbebc8694b90df2343185e702ca4d873aa8669/pytorch_lightning/callbacks/progress/base.py#L217-L250

tests/accelerators/test_tpu.py
daniellepintz: Ah, thank you I was confused about that!

pytorch_lightning/accelerators/cpu.py
carmocca: Yes

ananthsub: its coming up now: https://github.com/PyTorchLightning/pytorch-lightning/issues/11253
fyi @EricWiener

PR #8352, opened 2021-07-09
Load ckpt path when model provided in validate/test/predict

SeanNaren:
## What does this PR do?

Closes #8347 

Behaviour now is for test, validate and predict:

```
trainer.test(model, ckpt_path=None) # use provided model
trainer.test(model, ckpt_path='best') # load best model
trainer.test(model, ckpt_path='my_path') # load path
```

```
trainer.fit(model)
# then
trainer.test(ckpt_path=None) # load best model
trainer.test(ckpt_path='my_path') # load path
```

The biggest change is that if anyone did `trainer.test(ckpt_path=None)` to load the latest model, they would now need to change this to `trainer.test(model)` and pass the model reference directly (same for `validate/predict`).

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)
- [x] Did you list all the breaking changes introduced by this pull request?

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #9605, opened 2021-09-20
[wip][redo 9585] Support serialized checkpoint loading

### LINKED ISSUE #9406: Support serialized checkpoint loading

jjenniferdai:
## What does this PR do?

Fixes #9406 (redo of #9585, messed up git)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/plugins/training_type/ddp.py
+++ b/pytorch_lightning/plugins/training_type/ddp.py
@@ -21,7 +21,7 @@ import tempfile
 import time
 from pathlib import Path
 from time import sleep
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Mapping, Optional, Union
 
 import __main__
 import numpy as np
@@ -51,13 +51,14 @@ from pytorch_lightning.utilities import (
 from pytorch_lightning.utilities.distributed import (
     distributed_available,
     init_ddp_connection,
+    rank_zero_info,
     rank_zero_only,
     ReduceOp,
     sync_ddp_if_available,
 )
 from pytorch_lightning.utilities.exceptions import DeadlockDetectedException, MisconfigurationException
 from pytorch_lightning.utilities.seed import reset_seed
-from pytorch_lightning.utilities.types import STEP_OUTPUT
+from pytorch_lightning.utilities.types import _PATH, STEP_OUTPUT
 
 if _TORCH_GREATER_EQUAL_1_10:
     from torch.distributed.optim import DistributedOptimizer, PostLocalSGDOptimizer, ZeroRedundancyOptimizer
@@ -127,6 +128,7 @@ class DDPPlugin(ParallelPlugin):
         self._pids: Optional[List[int]] = None
         self._sync_dir: Optional[str] = None
         self._rank_0_has_called_call_children_scripts: bool = False
+        self._self_deleted_checkpoint_state_dict: bool = False
         self.set_world_ranks()
 
     @property
@@ -535,3 +537,19 @@ class DDPPlugin(ParallelPlugin):
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+    def load_model_state_dict(self, checkpoint: Mapping[str, Any]) -> None:
+        if "state_dict" not in checkpoint and self._self_deleted_checkpoint_state_dict:
+            return
+        self.lightning_module.load_state_dict(checkpoint["state_dict"])
+
+    def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
+        rank_zero_info(f"DistributedDataParallel has {self.num_processes} processes. Serializing to avoid CPU OOMs.")
+        for current_worker in range(self.num_processes):
+            if self.local_rank == current_worker:
+                checkpoint = super().load_checkpoint(checkpoint_path)
+                del checkpoint["state_dict"]
+                self._self_deleted_checkpoint_state_dict = True
+                log.info(f"Rank {self.global_rank}: done loading model states from {checkpoint_path}.")
+            self.barrier()
+        return checkpoint
--- a/pytorch_lightning/plugins/training_type/fully_sharded.py
+++ b/pytorch_lightning/plugins/training_type/fully_sharded.py
@@ -141,16 +141,16 @@ class DDPFullyShardedPlugin(DDPPlugin):
         ):
             yield
 
-    def setup_environment(self) -> None:
-        super().setup_environment()
-        model_call_configure_sharded_model_hook = getattr(
-            self.lightning_module, "call_configure_sharded_model_hook", False
-        )
-        if not model_call_configure_sharded_model_hook:
-            # if model has not called configure sharded model, we reset
-            # the training type plugin's call_configure_sharded_model_hook
-            # to give trainer a chance to configure.
-            self.call_configure_sharded_model_hook = True
+    # def setup_environment(self) -> None:
+    #     super().setup_environment()
+        # model_call_configure_sharded_model_hook = getattr(
+        #     self.lightning_module, "call_configure_sharded_model_hook", False
+        # )
+        # if not model_call_configure_sharded_model_hook:
+        #     # if model has not called configure sharded model, we reset
+        #     # the training type plugin's call_configure_sharded_model_hook
+        #     # to give trainer a chance to configure.
+        #     self.call_configure_sharded_model_hook = True
 
     def configure_ddp(self) -> None:
         if not self.cpu_offload:
--- a/pytorch_lightning/plugins/training_type/training_type_plugin.py
+++ b/pytorch_lightning/plugins/training_type/training_type_plugin.py
@@ -153,7 +153,10 @@ class TrainingTypePlugin(ABC):
 
     def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
         torch.cuda.empty_cache()
-        return self.checkpoint_io.load_checkpoint(checkpoint_path)
+        checkpoint = self.checkpoint_io.load_checkpoint(checkpoint_path)
+        self.lightning_module.on_load_checkpoint(checkpoint)
+        self.load_model_state_dict(checkpoint)
+        return checkpoint
 
     def load_model_state_dict(self, checkpoint: Mapping[str, Any]) -> None:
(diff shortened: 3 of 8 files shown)


pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -535,3 +537,19 @@ def teardown(self) -> None:
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+    def load_model_state_dict(self, checkpoint: Mapping[str, Any]) -> None:
+        if "state_dict" not in checkpoint and self._self_deleted_checkpoint_state_dict:
```
tchaton, line 542: Why do you need `self._self_deleted_checkpoint_state_dict` ? note: The variable name isn't great :)

jjenniferdai: original PR [comment thread](https://github.com/PyTorchLightning/pytorch-lightning/pull/9585#discussion_r711380142):  I think in case `load_model_state_dict` gets called after `load_checkpoint`, @ananthsub was this the intention?

ananthsub: @tchaton it's to distinguish if the 

@jjenniferdai - we should also ensure we reset the property in the `teardown` function

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -154,12 +148,8 @@ def restore_model(self) -> None:
 
     def restore_model_weights(self, checkpoint_path: Optional[_PATH]) -> None:
         """Restore only the model weights."""
-        checkpoint = self._loaded_checkpoint
```
ananthsub, line 157: - when is checkpoint_path `None` ? 
- i haven't traced through the paths fully here. is there a concern we're now loading a different checkpoint dict somewhere?

pytorch_lightning/plugins/training_type/fully_sharded.py
```diff
@@ -141,17 +141,6 @@ def wrap_policy(*args, **kwargs):
...
-    def setup_environment(self) -> None:
-        super().setup_environment()
-        model_call_configure_sharded_model_hook = getattr(
-            self.lightning_module, "call_configure_sharded_model_hook", False
-        )
-        if not model_call_configure_sharded_model_hook:
-            # if model has not called configure sharded model, we reset
-            # the training type plugin's call_configure_sharded_model_hook
-            # to give trainer a chance to configure.
-            self.call_configure_sharded_model_hook = True
```
ananthsub, line 153: on rebase this should be gone now

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -153,7 +153,10 @@ def results(self) -> Optional[Union[_EVALUATE_OUTPUT, _PREDICT_OUTPUT]]:
 
     def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
         torch.cuda.empty_cache()
-        return self.checkpoint_io.load_checkpoint(checkpoint_path)
+        checkpoint = self.checkpoint_io.load_checkpoint(checkpoint_path)
+        self.lightning_module.on_load_checkpoint(checkpoint)
+        self.load_model_state_dict(checkpoint)
```
ananthsub, line 158: n00b question: do you think this ought to be on the base class? vs calling `self.lightning_module.on_load_checkpoint(checkpoint)` in `load_model_state_dict` ?

jjenniferdai: do you mean if `on_load_checkpoint` should be in `load_model_state_dict` v.s. `load_checkpoint` here in ttp base class? if so I was thinking to maybe keep `load_model_state_dict` only performing `load_state_dict` as is, wdyt?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
jjenniferdai: its never None, `restore_model_weights` is only called with `trainer._ckpt_path` if `trainer._ckpt_path` is not None. should we remove the None check for clarity?

ananthsub: in that case i believe we should. @awaelchli what do you think?

pytorch_lightning/plugins/training_type/training_type_plugin.py
ananthsub: `load_checkpoint` is materializing the file contents in memory whereas `load_model_state_dict` is specifically on reloading the model's state

in my head, i think `on_load_checkpoint` is in spirit closer to `load_model_state_dict` than it is to loading the file...but i'm not 100% certain if my assumption is right. that's based on the current codebase where `model.on_load_checkpoint` is called right before the state dict.

if this was redone, maybe the name for this hook would've been `on_load_state_dict` ..

@tchaton @awaelchli does my understanding sqaure up with yours?

pytorch_lightning/plugins/training_type/ddp.py
jjenniferdai: "but later, the checkpoint connector will call `load_model_state_dict`" - this doesn't happen anymore after these checkpoint connector changes right?

pytorch_lightning/plugins/training_type/training_type_plugin.py
jjenniferdai: hm agree `on_load_checkpoint` is closer to `load_model_state_dict` than `load_checkpoint (file)` - maybe some renaming?
- old `load_checkpoint` --> `load_checkpoint_file`
- new method `load_checkpoint_file_and_model_state` (a better name than this?) = `load_checkpoint_file`, `on_load_checkpoint`, `load_model_state_dict`

jjenniferdai: just to confirm - like the below? (so `load_model_state_dict`, `load_checkpoint` will have different functionalities across different TTPs)

base TTP:
- `load_model_state_dict` = `on_load_checkpoint`, `model.load_state_dict`
- `load_checkpoint` = load checkpoint file

serialized loading TTPs:
- `load_model_state_dict` = `model.load_state_dict` only
- `load_checkpoint` = for rank in ranks, `super().load_checkpoint`, `on_load_checkpoint`, `load_model_state_dict`
- (`load_model_state_dict` will be called again in checkpoint connector, but early exits with the `_self_deleted_checkpoint_state_dict` boolean)

ananthsub: yes that's correct, at least this is my preference. what do you think @jjenniferdai @awaelchli @carmocca ?

jjenniferdai: I slightly prefer the second option because it's more consistent across TTPs. but defer to others more experienced :)

pytorch_lightning/plugins/training_type/ddp.py
tchaton: Sounds good to me, but let's find a better variable name. Something like `self._has_loaded_state_dict`

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: was pinged here but sorry, I'm not able to follow the conversation.
- you can rename things in the io plugin because its not released yet
- we are not allowed to change the order of hook calls

pytorch_lightning/trainer/trainer.py
```diff
@@ -984,9 +984,6 @@ def _run(self, model: "pl.LightningModule") -> Optional[Union[_EVALUATE_OUTPUT,
         self.data_connector.prepare_data()
         self.callback_connector._attach_model_callbacks()
 
-        if self._ckpt_path and not self.accelerator.restore_checkpoint_after_pre_dispatch:
```
awaelchli, line 987: this was here for a reason. check deepspeed vs. training type base:

deepspeed: restore_checkpoint_after_pre_dispatch = True
all others: restore_checkpoint_after_pre_dispatch = False

pytorch_lightning/plugins/training_type/training_type_plugin.py
jjenniferdai: no change to order of hook calls or io plugin, but moving calling responsibility of `LightningModule.on_load_checkpoint` from checkpoint connector to TTP ("to work through" portion of issue #9406)

2 different options in ([comment](https://github.com/PyTorchLightning/pytorch-lightning/pull/9605#discussion_r717806551)) on how to organize this:
- first option = this [comment](https://github.com/PyTorchLightning/pytorch-lightning/pull/9605#discussion_r717165512)
- second option = this PR currently

pytorch_lightning/trainer/trainer.py
jjenniferdai: hm requiring validate/test/predict to restore model weights before setup while fit restores after seems like an odd reliance, does anyone have context/insights on use cases that would rely on this gap? originally introduced in PR #8352

Issue #9330, opened 2021-09-05
Gradient clip norm is called before AMP's unscale leading to wrong gradients


phizaz:
## 🐛 Bug

Gradient clip norm is called before AMP's unscale leading to wrong gradients. 

### To Reproduce
This happens with:

```
trainer = pl.Trainer(
        gpus=1,
        precision=16,
        gradient_clip_val=1,
)
```

The gradient norm at ` training_step_end` is super small! My interpretation is that the gradient clipping happens before AMP's unscale so that unscale makes the gradient much smaller than the clip norm.

https://colab.research.google.com/drive/1CXMo2JP_JmwG_YNrTwTG5S0pAsr41pGC?usp=sharing

### Expected behavior

Gradient norms should be close to 1 (clip value). This is verifiable by setting `precision=32` and rerun. 

### Environment

```
* CUDA:
	- GPU:
		- Tesla K80
	- available:         True
	- version:           10.2
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     False
	- pyTorch_version:   1.9.0+cu102
	- pytorch-lightning: 1.4.5
	- tqdm:              4.62.0
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.7.11
	- version:           #1 SMP Sat Jun 5 09:50:34 PDT 2021
```

cowwoc:
I believe this is fixed in release 1.4.9: https://github.com/PyTorchLightning/pytorch-lightning/releases/tag/1.4.9

Can this issue be closed?

carmocca:
Keeping it alive as the fix is not in master

bergen:
Has this issue been fixed for TPUs? I am getting a discrepancy between single core and multi-core TPU training. Using version 1.4.9.

PR #9287, opened 2021-09-02
[WIP] Fix gradient norm tracking and gradient clipping

carmocca:
## What does this PR do?

THIS PR WILL BE SPLIT UP

Reported in Slack

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

Issue #9205, opened 2021-08-30
gradient value should be unscaled when precision=16


AndersJing:
## 🐛 Bug

On 1.4.x, when track_grad_norm=2 and precision=16, the logged gradient value is the scaled value.

### To Reproduce

https://colab.research.google.com/drive/1kV3wP8HmprPVflKXfY5eMdqIOar8reKk?usp=sharing

### Expected behavior

The gradient value should be unscaled when preciosn=16

### Environment

* CUDA:
	- GPU:
		- Tesla K80
	- available:         True
	- version:           10.2
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     False
	- pyTorch_version:   1.9.0+cu102
	- pytorch-lightning: 1.4.4
	- tqdm:              4.62.0
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.7.11
	- version:            Sat Jun 5 09:50:34 PDT 2021

### Additional context

On v1.3.8, the behavior is normal, see https://colab.research.google.com/drive/1AXMhr1gPll5ZPDTphQqAOJlI_9iEMISU?usp=sharing

@carmocca

cowwoc:
I believe this is fixed in release 1.4.9: https://github.com/PyTorchLightning/pytorch-lightning/releases/tag/1.4.9

Can this issue be closed?

AndersJing:
I run the code again on 1.4.9, the issue is fixed.

PR #9606, opened 2021-09-20
[Bugfix] Fix location of `unscale` in mixed precision plugin

### LINKED ISSUE #9599: Clip norm after scaler.unscale_ in native fp16 training
## 🐛 Bug
I trained a large model using native amp, but the loss converged very slow. After a careful check of the backward and optimization code, I found the `clip_gradients` is executed right after backward, but `scaler.unscale_` is conducted in pre_optimization_step.
According to the [instruction of Pytorch](https://pytorch.org/docs/stable/notes/amp_examples.html#gradient-clipping), the order of clip and unscale should be exchanged. Currently `gradient_clip_val` may lead to a very flat learning curve if used together with native amp.
Hope to be fixed.

justusschock:
## What does this PR do?

Fixes #9599, #9330, #9205, #9694

Unscale was called after gradnorm tracking and gradient clipping. Moving it to a different hook in the plugins should fix that.

Alternative solution would be to move gradient tracking and clipping. Finding a solution that works for all plugin combinations might be difficult though.

Probably same purpose as #9287

`None`

## Before submitting

## PR review

## Did you have fun?

**Please don't merge before review from @carmocca**

### DIFF
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -264,7 +264,7 @@ class Accelerator:
         """
         return self.training_type_plugin.validation_step_end(output)
 
-    def backward(self, closure_loss: Tensor, *args: Any, **kwargs: Any) -> Tensor:
+    def backward(self, closure_loss: Tensor, optimizer: torch.optim.Optimizer, *args: Any, **kwargs: Any) -> Tensor:
         """Forwards backward-calls to the precision plugin.
 
         Args:
@@ -273,9 +273,9 @@ class Accelerator:
         self.training_type_plugin.pre_backward(closure_loss)
         closure_loss = self.precision_plugin.pre_backward(self.lightning_module, closure_loss)
 
-        self.precision_plugin.backward(self.lightning_module, closure_loss, *args, **kwargs)
+        self.precision_plugin.backward(self.lightning_module, closure_loss, optimizer, *args, **kwargs)
 
-        closure_loss = self.precision_plugin.post_backward(self.lightning_module, closure_loss)
+        closure_loss = self.precision_plugin.post_backward(self.lightning_module, closure_loss, optimizer)
         self.training_type_plugin.post_backward(closure_loss)
 
         return closure_loss
--- a/pytorch_lightning/plugins/precision/native_amp.py
+++ b/pytorch_lightning/plugins/precision/native_amp.py
@@ -55,7 +55,8 @@ class NativeMixedPrecisionPlugin(MixedPrecisionPlugin):
                 " To request, please file a Github issue in PyTorch and tag @mcarilli"
             )
         result = lambda_closure()  # native amp does not support closures
-        self.scaler.unscale_(optimizer)
+        if not model.automatic_optimization:
+            self.scaler.unscale_(optimizer)
         super().pre_optimizer_step(model, optimizer, optimizer_idx, lambda_closure, **kwargs)
         skipped_backward = result is None
         # in manual optimization, the closure does not return a value
@@ -64,6 +65,16 @@ class NativeMixedPrecisionPlugin(MixedPrecisionPlugin):
             self.scaler.step(optimizer)
             self.scaler.update()
         return False
+
+    def post_backward(
+        self, model: "pl.LightningModule", closure_loss: torch.Tensor, optimizer: Optimizer
+    ) -> torch.Tensor:
+        ret_val = super().post_backward(model, closure_loss, optimizer)
+        # unscale here to have it inside the closure before the grad tracking and clipping
+        if model.automatic_optimization and not model.trainer.fit_loop._should_accumulate():
+            self.scaler.unscale_(optimizer)
+        return ret_val
+
 
     @contextmanager
     def train_step_context(self) -> Generator[None, None, None]:
--- a/pytorch_lightning/plugins/precision/native_amp.py
+++ b/pytorch_lightning/plugins/precision/native_amp.py
@@ -71,10 +71,9 @@ class NativeMixedPrecisionPlugin(MixedPrecisionPlugin):
     ) -> torch.Tensor:
         ret_val = super().post_backward(model, closure_loss, optimizer)
         # unscale here to have it inside the closure before the grad tracking and clipping
-        if model.automatic_optimization and not model.trainer.fit_loop._should_accumulate():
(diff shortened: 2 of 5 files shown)


pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -106,6 +105,15 @@ def pre_optimizer_step(
             self.scaler.update()
         return False
 
+    def post_backward(
+        self, model: "pl.LightningModule", closure_loss: torch.Tensor, optimizer: torch.optim.Optimizer
+    ) -> torch.Tensor:
+        ret_val = super().post_backward(model, closure_loss, optimizer)
+        # unscale here to have it inside the closure before the grad tracking and clipping
+        if not model.trainer.fit_loop._should_accumulate():
+            self.scaler.unscale_(optimizer)
```
tchaton, line 114: Can you check this isn't the LightningOptimizer ?

tests/plugins/test_amp_plugins.py
```diff
@@ -232,3 +232,39 @@ def test_cpu_amp_precision_16_throws_error(tmpdir):
...
+    def post_backward(
+        self, model: "pl.LightningModule", closure_loss: torch.Tensor, optimizer: torch.optim.Optimizer
+    ) -> torch.Tensor:
+        norm_before = torch.nn.utils.clip_grad_norm_(model.parameters(), 2)
+        ret_val = super().post_backward(model, closure_loss, optimizer)
+        norm_after = torch.nn.utils.clip_grad_norm_(model.parameters(), 2)
+
+        # norm_after unscale should be smaller by scaling factor greater than 1
+        if not (torch.isinf(norm_before) or torch.isnan(norm_before)):
+            assert norm_after < norm_before
```
tchaton, line 249: The scale should be much larger than 1. 

```suggestion
            assert norm_after < norm_before * 10
```

awaelchli:
@justusschock I think this is ready for review now

awaelchli (approved):

pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -96,7 +96,8 @@ def pre_optimizer_step(
                 " To request, please file a Github issue in PyTorch and tag @mcarilli"
             )
         result = lambda_closure()  # native amp does not support closures
-        self.scaler.unscale_(optimizer)
+        if not model.automatic_optimization:
```
carmocca, line 99: Why is this if added? `step` below will not unscale again if it has been done manually already

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -228,7 +228,7 @@ def validation_step_end(self, output: Optional[STEP_OUTPUT]) -> Optional[STEP_OU
         """
         return self.training_type_plugin.validation_step_end(output)
 
-    def backward(self, closure_loss: Tensor, *args: Any, **kwargs: Any) -> Tensor:
+    def backward(self, closure_loss: Tensor, optimizer: torch.optim.Optimizer, *args: Any, **kwargs: Any) -> Tensor:
```
carmocca, line 231: as a side note: this new parameter would be removed again with the long-term solution of moving the `step` into the precision plugin

four4fish, line 231: Does Optimizer vs torch.optim.Optimizer the same here?

pytorch_lightning/plugins/precision/native_amp.py
four4fish, line 99: Do we need check isinstance(model, pl.LightningModule) here? Is it possible that model is wrapped ?

awaelchli: @carmocca I don't like it either. This makes the hacks that exist here even more hacky. 
The PR aims at solving the bug that unscaling is happening after gradient clipping. So to fix this, one can move the unscaling to an earlier point, which we choose to do directly after backward (post_backward). However, we ONLY want to do that in the case of automatic optimization, because gradient clipping does not exist for manual backward and we need to keep the unscaling in the optimizer_step here because the user of manual optimization will call the LightningOptimizer.step themselves. 

Bottom line: we have unscale in two locations, but only one is ever active depending on whether we are in automatic or manual optimization.

pytorch_lightning/accelerators/accelerator.py
awaelchli: yes, we should choose just "Optimizer" here for consistency with the other signatures. It does not make a functional difference :)

pytorch_lightning/plugins/precision/native_amp.py
awaelchli: No, not needed. 1) the methods here are type annotated with LightningModule. 2) we are explicitly passing in `self.lighting_module` where we call the method, and `self.lightning_module` is guaranteed to be the unwrapped pure LM.

pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -96,7 +96,8 @@ def pre_optimizer_step(
...
+            self.scaler.unscale_(optimizer)
```
SeanNaren, line 100:
```suggestion
            # unscale in manual optimization as user does not rely on lightning
            # to call backward, but does call LightningOptimizer.step
            self.scaler.unscale_(optimizer)
```

SeanNaren (approved):

carmocca (approved):

awaelchli:
As discussed with @carmocca I rebased this onto the 1.4.x bugfix branch as we desire a better, less hacky long-term solution for master #9287 . So this is considered a hotfix.

There were basically no merge conflicts, the changes stay the same.

tests/plugins/test_amp_plugins.py
awaelchli: I don't fully understand, adding the `* 10` on the right hand side of the inequality is not going to add any more value to the assertion

pytorch_lightning/plugins/precision/native_amp.py
awaelchli: done, added assertion

tests/plugins/test_amp_plugins.py
tchaton: The amp scaling value is quite large. Could it be possible the norm after is smaller due to un-precision `norm_after < norm_before`.

The *10 would just prevent this type of regression. This is optional.

tchaton (approved):
LGTM !

Issue #8722, opened 2021-08-04
Cleanup FSDP integration to not require boilerplate logic


SeanNaren:
## 🚀 Feature

Motivated by debugging FSDP in a recent PR made by @carmocca, I think we should try clean out the interface for FSDP. 

Currently FSDP supports a case where we wrap layers inside the `configure_sharded_model` hook, with an assumption that these layers are defined outside the hook. This is probably because in most cases the model has been defined in setup, or `__init__`.

This can be seen here: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py#L58-L74

```python

class TestFSDPModel(BoringModel):
    def setup(self, stage: str) -> None:
        if stage != "fit":
            # when running stages like test, validate, and predict, we will skip setting up,
            # will directly use the module itself unless we load from checkpoint
            return
        # resetting call_configure_sharded_model_hook attribute so that we could call
        # configure sharded model
        self.call_configure_sharded_model_hook = False
        # for loading full state dict, we first need to create a new unwrapped model
        # to load state dict and then wrapping
        self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))

    def configure_sharded_model(self) -> None:
        for i, layer in enumerate(self.layer):
            if i % 2 == 0:
                self.layer[i] = wrap(layer)
        self.layer = wrap(self.layer)

    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
        # when loading full state dict, we first need to create a new unwrapped model
        self.setup("fit")
```

Also included is a lot of boilerplate logic to handle the case where a user wants to load weights back into model, and we re-create the model -> load weights -> call configure sharded model again. 

This is a bit unclean as we see an internal variable needing to be reset (`call_configure_sharded_model_hook`) and more importantly, assume that the model state hasn't been altered (which it has by FSDP which permanently flattens the parameters).

Imo we should move towards this API:

```python
class TestFSDPModel(BoringModel):
    def __init__(self):
        self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))

    def configure_sharded_model(self) -> None:
        for i, layer in enumerate(self.layer):
            if i % 2 == 0:
                self.layer[i] = wrap(layer)
        self.layer = wrap(self.layer)
[...]

Issue #8593, opened 2021-07-28
Avoid rewrapping LightningModules in plugins


ananthsub:
## 🚀 Feature

Background:

We are auditing the Lightning components and APIs to assess opportunities for improvements:
- https://github.com/PyTorchLightning/pytorch-lightning/issues/7740
- https://docs.google.com/document/d/1xHU7-iQSpp9KJTjI3As2EM0mfNHHr37WZYpDpwLkivA/edit#

### Motivation

Lightning has the concept of a Training Type Plugin, which functions as the distribution strategy to be used during execution. For this, Lightning offers an API and out of the box implementations for data parallel-based approaches (DataParallel, DistributedDataParallel, ShardedDataParallel, DeepSpeed, etc). The trainer wraps the whole LightningModule with a module-wrapper in order to facilitate the gradient synchronization.

This allows users to easily specify distributed training options like:
```
trainer = Trainer(num_nodes=..., gpus=..., plugins=["..."]) 
module = MyLightningModule(...)
trainer.fit(module)
```

The situation we'd like to explicitly avoid is if the user makes successive calls to Trainer APIs like this:
```
trainer = Trainer(num_nodes=..., gpus=..., plugins=["..."]) 
module = MyLightningModule(...)
trainer.fit(module)
trainer.test(module)
```
potentially causing the wrapper to be applied multiple times to the LightningModule.  (e.g. DistributedDataParallel(DistributedDataParallel(LightningModule)) )

### Pitch

Inside of the following plugins, avoid wrapping the plugin's `self._model` if the model is already an instance of the wrapped type.

Plugins to update
DDP: https://github.com/PyTorchLightning/pytorch-lightning/blob/80c529351439a0f8d3d6e9449cd47d16ba3abbec/pytorch_lightning/plugins/training_type/ddp.py#L249-L256
DDP Spawn: https://github.com/PyTorchLightning/pytorch-lightning/blob/6b47cbe3ca8aa3fd82211bc9fa32e734753a6950/pytorch_lightning/plugins/training_type/ddp_spawn.py#L247-L252
Sharded Data Parallel: https://github.com/PyTorchLightning/pytorch-lightning/blob/6b47cbe3ca8aa3fd82211bc9fa32e734753a6950/pytorch_lightning/plugins/training_type/sharded.py#L37-L45
[...]

PR #8080, opened 2021-06-22
Destroy process group in DDP destructor

carmocca:
## What does this PR do?

Discovered by running a DeepSpeed test without `special=True` which would produce the error:

`RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.`

This is because a previous test initialized the process group with the `gloo` backend and we never destroyed it because it raised a `SystemExit` on setup. And since special tests are run separately this wasn't caught before.

https://github.com/PyTorchLightning/pytorch-lightning/blob/f79f0f9de112bc0a810587147b2ff5d1f3bfb256/tests/accelerators/test_ddp.py#L98-L122

So the fix is:

- Properly release the group in ~~teardown~~ (edit: destructor, we want to reuse the process group across `fit/validate/test` calls) 
- ~~Do not raise a SystemExit in setup as teardown won't get called in that case. We should eventually also clean up after ourselves in cases like that.~~

The changes in #8070 need this to work so that will act as a test

Closes #8115 

## Before submitting
- [ ] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [ ] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

PR #9612, opened 2021-09-20
Remove `call_configure_sharded_model` lifecycle property

ananthsub:
## What does this PR do?

Part of #8722

Changes:
- Remove the property guarding the hook: The Trainer will always call `configure_sharded_model` in each of `fit` `validate` `test` etc. This avoids subtle side-effects from prior runs and makes the call order consistent
- In turn, we strongly recommend users to implement `configure_sharded_model` idempotently. The update to `TestFSDPModel` demonstrates how one can check if the layers are already wrapped with FSDP and return early if so. This is the same spirit of https://github.com/PyTorchLightning/pytorch-lightning/issues/8593 but in user-land

@SeanNaren 

- Calls `configure_sharded_model` unconditionally
- This removes the property on the accelerator/training type plugin
- Removes support for `call_configure_sharded_model_hook` on the LightningModule (which is not officially part of the LightningModule API) 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -402,20 +402,6 @@ class Accelerator:
         self.training_type_plugin.save_checkpoint(checkpoint, filepath)
 
     @property
-    def call_configure_sharded_model_hook(self) -> bool:
-        """Allow model parallel hook to be called in suitable environments determined by the training type plugin.
-        This is useful for when we want to shard the model once within fit.
-
-        Returns:
-            True if we want to call the model parallel setup hook.
-        """
-        return self.training_type_plugin.call_configure_sharded_model_hook
-
-    @call_configure_sharded_model_hook.setter
-    def call_configure_sharded_model_hook(self, mode: bool) -> None:
-        self.training_type_plugin.call_configure_sharded_model_hook = mode
-
-    @property
     def setup_optimizers_in_pre_dispatch(self) -> bool:
         """Override to delay setting optimizers and schedulers till after dispatch. This is useful when the
         `TrainingTypePlugin` requires operating on the wrapped accelerator model. However this may break certain
--- a/pytorch_lightning/core/hooks.py
+++ b/pytorch_lightning/core/hooks.py
@@ -297,12 +297,8 @@ class ModelHooks:
         where we'd like to shard the model instantly, which is useful for extremely large models which can save
         memory and initialization time.
 
-        The accelerator manages whether to call this hook at every given stage.
-        For sharded plugins where model parallelism is required, the hook is usually on called once
-        to initialize the sharded parameters, and not called again in the same process.
-
-        By default for accelerators/plugins that do not use model sharding techniques,
-        this hook is called during each fit/val/test/predict stages.
+        This hook is called during each of fit/val/test/predict stages in the same process, so ensure that
+        implementation of this hook is idempotent.
         """
 
 
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -1276,18 +1276,9 @@ class Trainer(
         self.accelerator.barrier("post_setup")
 
     def _call_configure_sharded_model(self) -> None:
-        # Call configure sharded model hook if accelerator requests. In some cases
-        # we will not call the hook; the hook has initialized the sharded model for example.
-
-        # used on the model if the user re-create a trainer with resume_from_checkpoint
-        model = self.lightning_module
-        model_call_configure_sharded_model_hook = getattr(model, "call_configure_sharded_model_hook", False)
-        if self.accelerator.call_configure_sharded_model_hook and not model_call_configure_sharded_model_hook:
-            with self.accelerator.model_sharded_context():
-                self.call_hook("configure_sharded_model")
-                self.call_hook("on_configure_sharded_model")
-            model.call_configure_sharded_model_hook = True
-            self.accelerator.call_configure_sharded_model_hook = False
+        with self.accelerator.model_sharded_context():
+            self.call_hook("configure_sharded_model")
+            self.call_hook("on_configure_sharded_model")
 
     def _call_teardown_hook(self) -> None:
         fn = self.state.fn._setup_fn
--- a/tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py
+++ b/tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py
@@ -71,7 +71,8 @@ class TestFSDPModel(BoringModel):
 
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
         # when loading full state dict, we first need to create a new unwrapped model
-        self._init_model()
+        if self.layer is None or isinstance(self.layer, FullyShardedDataParallel):
+            self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))
 
     def configure_optimizers(self):
         return torch.optim.SGD(self.layer.parameters(), lr=0.1)
--- a/tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py
+++ b/tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py
@@ -54,8 +54,7 @@ class TestFSDPModel(BoringModel):
         self.layer: Optional[torch.nn.Module] = None
 
     def _init_model(self) -> None:
-        if self.layer is None:
-            self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))
+        self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))
 
     def setup(self, stage: str) -> None:
         self._init_model()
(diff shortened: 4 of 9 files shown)


tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py
```diff
@@ -49,27 +49,29 @@ def test_fsdp_with_sharded_amp(device_count_mock, mock_cuda_available, tmpdir):
...
-        # configure sharded model
-        self.call_configure_sharded_model_hook = False
-        # for loading full state dict, we first need to create a new unwrapped model
-        # to load state dict and then wrapping
-        self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))
+        self._init_model()
 
     def configure_sharded_model(self) -> None:
+        # the model is already wrapped with FSDP: no need to wrap again!
+        if isinstance(self.layer, FullyShardedDataParallel):
```
four4fish, line 65: Will thi always return False? self.layer is a dict right?

ananthsub: no, it's an nn.Sequential module
but L70 `self.layer = wrap(self.layer)` will make this an instance of `FullyShardedDataParallel`

tests/plugins/test_ddp_fully_sharded_with_full_state_dict.py
```diff
@@ -49,27 +49,29 @@ def test_fsdp_with_sharded_amp(device_count_mock, mock_cuda_available, tmpdir):
...
-            return
-        # resetting call_configure_sharded_model_hook attribute so that we could call
-        # configure sharded model
-        self.call_configure_sharded_model_hook = False
-        # for loading full state dict, we first need to create a new unwrapped model
-        # to load state dict and then wrapping
-        self.layer = torch.nn.Sequential(torch.nn.Linear(32, 32), torch.nn.ReLU(), torch.nn.Linear(32, 2))
+        self._init_model()
 
     def configure_sharded_model(self) -> None:
```
carmocca, line 63: Can you update the docstring for this function to mention the idempotency requirement?

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/core/hooks.py#L295-L306

carmocca (approved):

pytorch_lightning/core/hooks.py
```diff
@@ -297,12 +297,8 @@ def configure_sharded_model(self) -> None:
...
         memory and initialization time.
 
-        The accelerator manages whether to call this hook at every given stage.
-        For sharded plugins where model parallelism is required, the hook is usually on called once
-        to initialize the sharded parameters, and not called again in the same process.
-
-        By default for accelerators/plugins that do not use model sharding techniques,
-        this hook is called during each fit/val/test/predict stages.
+        This hook is called during each of fit/val/test/predict stages in the same process, so ensure that
+        implementation of this hook is idempotent.
```
ananthsub, line 301: @carmocca do you think it'd be helpful to provide an example in case people are unfamiliar with this. such as, "For example, return early if your model is already sharded" ?

carmocca: sure!

SeanNaren (approved):
Multiple calls to `configure_sharded_model` should be fine for DeepSpeed as well as there is a guard in place to not re-partition parameters that have already been partitioned: https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/runtime/zero/partition_parameters.py#L499

@ananthsub is it important to update the docs now? or should we wait till further changes are made to support a persistent model state across Trainer stages?

awaelchli (approved):

tchaton (approved):
LGTM !

tchaton:
Hey @ananthsub,

Any progress on the failing Azure tests ?

Best,
T.C

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -401,20 +401,6 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
         """
         self.training_type_plugin.save_checkpoint(checkpoint, filepath)
 
-    @property
-    def call_configure_sharded_model_hook(self) -> bool:
```
Borda, line 405: consider deprecation as it used to be public API
on the other hand, Accelerators are meant to be for internal use only...
cc: @awaelchli

ananthsub:
> Hey @ananthsub,
> 
> Any progress on the failing Azure tests ?
> 
> Best,
> T.C

@tchaton I'm unable to reproduce the failures locally. is there any interleaving of tests that could cause this to fail in CI but not locally?

awaelchli:
@ananthsub I can reproduce locally by running `py.test -v tests/plugins`. 

the first test that fails seems to be `test_deepspeed_skip_backward_raises`. 

When calling the tests individually, they pass. This is not unfamiliar as we had this a few times before. Since distributed logic is all globally shared in the e.g. in the torch package, things can leak from one test to the other (is my interpretation). 

```bash
tests/plugins/test_sharded_plugin.py::test_configure_ddp FAILED                                                                                                                                                                                                                                                                       [ 77%]
tests/plugins/test_sharded_plugin.py::test_custom_kwargs_sharded[DDPShardedPlugin] FAILED                                                                                                                                                                                                                                             [ 77%]
tests/plugins/test_sharded_plugin.py::test_custom_kwargs_sharded[DDPSpawnShardedPlugin] FAILED                                                                                                                                                                                                                                        [ 78%]
tests/plugins/test_sharded_plugin.py::test_custom_kwargs_sharded_reduce_buffer_size[1-params0-0] FAILED                                                                                                                                                                                                                               [ 78%]
tests/plugins/test_sharded_plugin.py::test_custom_kwargs_sharded_reduce_buffer_size[1-params1-128] FAILED                                                                                                                                                                                                                             [ 79%]
tests/plugins/test_sharded_plugin.py::test_custom_kwargs_sharded_reduce_buffer_size[2-params0-0] FAILED                                                                                                                                                                                                                               [ 80%]
tests/plugins/test_sharded_plugin.py::test_custom_kwargs_sharded_reduce_buffer_size[2-params1-128] FAILED                                                                                                                                                                                                                             [ 80%]
tests/plugins/test_sharded_plugin.py::test_block_backward_sync PASSED                                                                                                                                                                                                                                                                 [ 81%]
tests/plugins/test_single_device_plugin.py::test_single_cpu PASSED                                                                                                                                                                                                                                                                    [ 81%]
tests/plugins/test_single_device_plugin.py::test_single_gpu FAILED
```

Looking at the error messages, I see something disturbing:

```
tests/plugins/test_sharded_plugin.py:303:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pytorch_lightning/plugins/training_type/sharded.py:46: in configure_ddp
    LightningShardedDataParallel(self.model),
../../anaconda3/envs/pl/lib/python3.8/site-packages/deepspeed/runtime/zero/partition_parameters.py:240: in wrapper
    print_rank_0(f'Before initializing {module.__class__.__name__}',
```

The error message is originating from deepspeed 🤣 No clue how that can be. Need to investigate the code path.

ananthsub:
> This is not unfamiliar as we had this a few times before. Since distributed logic is all globally shared in the e.g. in the torch package, things can leak from one test to the other (is my interpretation).

Do you remember how this was solved before?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1276,18 +1276,9 @@ def _call_setup_hook(self) -> None:
...
-        # used on the model if the user re-create a trainer with resume_from_checkpoint
-        model = self.lightning_module
-        model_call_configure_sharded_model_hook = getattr(model, "call_configure_sharded_model_hook", False)
-        if self.accelerator.call_configure_sharded_model_hook and not model_call_configure_sharded_model_hook:
-            with self.accelerator.model_sharded_context():
-                self.call_hook("configure_sharded_model")
-                self.call_hook("on_configure_sharded_model")
-            model.call_configure_sharded_model_hook = True
-            self.accelerator.call_configure_sharded_model_hook = False
+        with self.accelerator.model_sharded_context():
```
awaelchli, line 1279: what I found so far is that this call to the model_sharded_context is causing the issue if it gets called for the deepspeed plugin. If I add 

```python
        if isinstance(self.training_type_plugin, DeepSpeedPlugin):
            return
```
before the call here, I don't see unrelated tests failing. 
So, investigating `DeepSpeedPlugin.model_sharded_context()`

awaelchli: @ananthsub possible that I found the test that causes this. Can you try to got to the test
`test_deepspeed_multigpu_stage_3_warns_resume_training` in` tests/plugins/test_deepspeed_plugin`
and flip the flag `special=False` to `special=True`
?

carmocca:
> Do you remember how this was solved before?

There is no solution without something like #8080. The workaround is to run the test per-process (what `special_tests.sh` does) and/or unrolling parametrizations

pytorch_lightning/accelerators/accelerator.py
awaelchli: that's ok imo. the hook is not really valuable as a way to extend the plugins and it really only was a hack for the sharded integration :)

Issue #9074, opened 2021-08-24
[RFC] Deprecate `should_rank_save_checkpoint`


ananthsub:
## Proposed refactoring or deprecation

Now that the checkpoint is better consolidated in the training type plugin, we no longer need this property, as this becomes an internal implementation detail of the training type.

Users (via the checkpoint callback) only need to call `trainer.save_checkpoint` and assume the plugin will handle checks surrounding this for them

### Motivation

- Ensure consolidation of saving logic in one place (e.g. all fsspec code for checkpoint paths shoud sit in one place vs. being scattered around the codebase)
- API simpliication: fewer properties exposed on the Trainer

### Pitch

Deprecate this property: https://github.com/PyTorchLightning/pytorch-lightning/blob/538e743f17c7da4624c902f762922e2837661818/pytorch_lightning/trainer/properties.py#L117-L120

Deprecate this property: https://github.com/PyTorchLightning/pytorch-lightning/blob/538e743f17c7da4624c902f762922e2837661818/pytorch_lightning/plugins/training_type/training_type_plugin.py#L313-L316

Move the directory creation from here: https://github.com/PyTorchLightning/pytorch-lightning/blob/538e743f17c7da4624c902f762922e2837661818/pytorch_lightning/callbacks/model_checkpoint.py#L511-L514

Into here: https://github.com/PyTorchLightning/pytorch-lightning/blob/538e743f17c7da4624c902f762922e2837661818/pytorch_lightning/plugins/io/torch_plugin.py#L30-L41

@SeanNaren - this is where we should also have a `rm_checkpoint` on the Checkpoint IO plugin such that we can deprecate this from the model checkpoint: https://github.com/PyTorchLightning/pytorch-lightning/blob/538e743f17c7da4624c902f762922e2837661818/pytorch_lightning/callbacks/model_checkpoint.py#L502-L505

One thing I'm not sure about: because the `should_rank_save_checkpoint` is exposed from the accelerator to the trainer as part of the public trainer API, does this need to go through a full deprecation cycle? Or is a breaking change as part of the plugins API permissible?

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

[...]

PR #9433, opened 2021-09-10
Remove `should_rank_save_checkpoint` property from Trainer

kaushikb11:
## What does this PR do?

Fixes #9074 

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #9615, opened 2021-09-20
Optimize non-empty directory warning check in model checkpoint callback

jjenniferdai:
## What does this PR do?

Along the lines of #9074, consolidate more fs calls from checkpoint callback. For non-empty directory warning:
- move to after `__resolve_ckpt_dir` 
- rank zero only

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/pytorch_lightning/callbacks/model_checkpoint.py
@@ -24,8 +24,7 @@ import re
 import time
 from copy import deepcopy
 from datetime import timedelta
-from pathlib import Path
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional
 from weakref import proxy
 
 import numpy as np
--- a/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/pytorch_lightning/callbacks/model_checkpoint.py
@@ -617,7 +617,12 @@ class ModelCheckpoint(Callback):
             self._fs.makedirs(self.dirpath, exist_ok=True)
 
     def __check_if_dir_not_empty(self, dirpath: _PATH, trainer: "pl.Trainer") -> None:
-        if trainer.is_global_zero and self.save_top_k != 0 and self._fs.isdir(dirpath) and len(self._fs.ls(dirpath)) > 0:
+        if (
+            trainer.is_global_zero
+            and self.save_top_k != 0
+            and self._fs.isdir(dirpath)
+            and len(self._fs.ls(dirpath)) > 0
+        ):
             rank_zero_warn(f"Checkpoint directory {dirpath} exists and is not empty.")
 
     def _validate_monitor_key(self, trainer: "pl.Trainer") -> None:
--- a/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/pytorch_lightning/callbacks/model_checkpoint.py
@@ -266,6 +266,7 @@ class ModelCheckpoint(Callback):
     def on_pretrain_routine_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         """When pretrain routine starts we build the ckpt dir on the fly."""
         self.__resolve_ckpt_dir(trainer)
+        self.__check_if_dir_not_empty(self.dirpath, trainer)
 
     def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         self._last_time_checked = time.monotonic()
@@ -614,6 +615,10 @@ class ModelCheckpoint(Callback):
 
         if not trainer.fast_dev_run and trainer.should_rank_save_checkpoint:
             self._fs.makedirs(self.dirpath, exist_ok=True)
+
+    def __check_if_dir_not_empty(self, dirpath: _PATH, trainer: "pl.Trainer") -> None:
+        if trainer.is_global_zero and self.save_top_k != 0 and self._fs.isdir(dirpath) and len(self._fs.ls(dirpath)) > 0:
+            rank_zero_warn(f"Checkpoint directory {dirpath} exists and is not empty.")
 
     def _validate_monitor_key(self, trainer: "pl.Trainer") -> None:
         metrics = trainer.callback_metrics
--- a/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/pytorch_lightning/callbacks/model_checkpoint.py
@@ -264,7 +264,7 @@ class ModelCheckpoint(Callback):
             self._save_on_train_epoch_end = trainer.val_check_interval == 1.0 and trainer.check_val_every_n_epoch == 1
 
     def on_pretrain_routine_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        """When pretrain routine starts we determine the ckpt dir on the fly."""
+        """When pretrain routine starts we build the ckpt dir on the fly."""
         self.__resolve_ckpt_dir(trainer)
 
     def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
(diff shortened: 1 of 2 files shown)


pytorch_lightning/plugins/io/torch_plugin.py
```diff
@@ -31,7 +31,10 @@ class TorchCheckpointIO(CheckpointIO):
 
     def save_checkpoint(self, checkpoint: Dict[str, Any], path: _PATH, storage_options: Optional[Any] = None) -> None:
         fs = get_filesystem(path)
-        fs.makedirs(os.path.dirname(path), exist_ok=True)
+        dirname = os.path.dirname(path)
+        if fs.isdir(dirname) and len(fs.ls(dirname)) > 0:
+            rank_zero_warn(f"Checkpoint directory {dirname} exists and is not empty.")
```
ananthsub, line 36: this will be noisy in the middle of the run. every time we go to save a new checkpoint this warning will show up even though it's expected. perhaps this warning can sit either inside of `__resolve_ckpt_dir`  or in `on_pretraine_routine_start` to indicate that no checkpoints have been saved yet but the path is not empty.

jjenniferdai: oh duh sorry for missing this - maybe move the warning inside this `if` in `__resolve_ckpt_dir`? https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/callbacks/model_checkpoint.py#L619-L620

(commented on maybe keeping this [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/9433#discussion_r712715429))

tchaton: We could use a warning cache to prevent over spamming the users.

SeanNaren (approved):

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -440,11 +439,8 @@ def __validate_init_configuration(self) -> None:
                     " will duplicate the last checkpoint saved."
                 )
 
-    def __init_ckpt_dir(self, dirpath: Optional[Union[str, Path]], filename: Optional[str]) -> None:
-        self._fs = get_filesystem(str(dirpath) if dirpath else "")
-
-        if self.save_top_k != 0 and dirpath is not None and self._fs.isdir(dirpath) and len(self._fs.ls(dirpath)) > 0:
-            rank_zero_warn(f"Checkpoint directory {dirpath} exists and is not empty.")
```
ananthsub, line 447: note for other reviewers:

This PR is an optimization we'd like to make when using remote filesystems with lightning. this code is currently called on every rank, which creates duplicate, bursty outbound network calls. when training with larger configurations, we ran into quota limitations for jobs our jobs. this will likely benefit anyone saving checkpoints using S3/GCS/azure storage

in reality, we only needed to check this on rank zero (assuming we have a common `dirpath` across nodes)

pytorch_lightning/plugins/io/torch_plugin.py
```diff
@@ -31,7 +33,10 @@ class TorchCheckpointIO(CheckpointIO):
 
     def save_checkpoint(self, checkpoint: Dict[str, Any], path: _PATH, storage_options: Optional[Any] = None) -> None:
         fs = get_filesystem(path)
-        fs.makedirs(os.path.dirname(path), exist_ok=True)
+        dirname = os.path.dirname(path)
+        if fs.isdir(dirname) and len(fs.ls(dirname)) > 0:
+            warning_cache.warn(f"Checkpoint directory {dirname} exists and is not empty.")
+        fs.makedirs(dirname, exist_ok=True)
```
ananthsub, line 39: i don't prefer this check here as much becaus if we're training with a plugin that supports saving sharded state, we still run the isdir check on all ranks. this will run into the network issues we saw earlier

@jjenniferdai what do you think about moving this check into `on_pretrain_routine_start` in the trainer, after `self.__resolve_ckpt_dir` finishes?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -744,7 +740,7 @@ def to_yaml(self, filepath: Optional[Union[str, Path]] = None) -> None:
         with self._fs.open(filepath, "w") as fp:
             yaml.dump(best_k, fp)
 
-    def file_exists(self, filepath: Union[str, Path], trainer: "pl.Trainer") -> bool:
+    def file_exists(self, filepath: _PATH, trainer: "pl.Trainer") -> bool:
```
ananthsub, line 743: thanks for updating these!

ananthsub (commented):
note for other reviewers:

This PR is an optimization we'd like to make when using remote filesystems with lightning. this code is currently called on every rank, which creates duplicate, bursty outbound network calls. when training with larger configurations, we ran into quota limitations for jobs our jobs. this will likely benefit anyone saving checkpoints using S3/GCS/azure storage

in reality, we only needed to check this on rank zero (assuming we have a common dirpath across nodes)

pytorch_lightning/plugins/io/torch_plugin.py
jjenniferdai: good point I wasn't sure about trading off point 1 with fs code io plugin consolidation, but for point 2 this shouldn't live here.

what about here https://github.com/PyTorchLightning/pytorch-lightning/pull/9615#discussion_r713353601 - at the end of `__resolve_ckpt_dir`, within existing rank if case and before makedirs

ananthsub: there's potential duplication because we need to handle the early return here: https://github.com/PyTorchLightning/pytorch-lightning/blob/8fcdcb598bbf0fbe6cda92415d1d69eb1e2d0aa0/pytorch_lightning/callbacks/model_checkpoint.py#L595-L598

which is why i prefer this sequence:
```
__resolve_ckpt_dir -> only initializes attributes on the checkpoint callback. does no IO
__check_if_dir_not_empty -> run this only on global rank 0 for the warning
```
and leave the `mkdirs` creation to the IO plugin

what do you think?

jjenniferdai: op forgot about the early return, yep this makes more sense!

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -267,6 +266,7 @@ def on_init_end(self, trainer: "pl.Trainer") -> None:
     def on_pretrain_routine_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         """When pretrain routine starts we build the ckpt dir on the fly."""
         self.__resolve_ckpt_dir(trainer)
+        self.__check_if_dir_not_empty(self.dirpath, trainer)
```
ananthsub, line 269:
```suggestion
        if trainer.is_global_zero:
            self.__warn_if_dir_not_empty(self.dirpath)
```
nit to keep the warning function more self-contained

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -619,6 +616,15 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
             self._fs.makedirs(self.dirpath, exist_ok=True)
 
+    def __check_if_dir_not_empty(self, dirpath: _PATH, trainer: "pl.Trainer") -> None:
+        if (
+            trainer.is_global_zero
+            and self.save_top_k != 0
+            and self._fs.isdir(dirpath)
+            and len(self._fs.ls(dirpath)) > 0
+        ):
+            rank_zero_warn(f"Checkpoint directory {dirpath} exists and is not empty.")
```
ananthsub, line 626:
```suggestion
    def __warn_if_dir_not_empty(self, dirpath: _PATH) -> None:
        if (
            self.save_top_k != 0
            and self._fs.isdir(dirpath)
            and len(self._fs.ls(dirpath)) > 0
        ):
            rank_zero_warn(f"Checkpoint directory {dirpath} exists and is not empty.")
```

ananthsub (approved):
LGTM!

carmocca (approved):

PR #9616, opened 2021-09-20
Deprecate `progress_bar_refresh_rate` from Trainer constructor

### LINKED ISSUE #9500: Deprecate `progress_bar_refresh_rate` from Trainer constructor

Deprecate `progress_bar_refresh_rate` from Trainer constructor and use default refresh rate of 1 if no progress bar callback is passed

### Motivation

We are trying to minimize the number of arguments on Trainer constructor: #9006
`progress_bar_refresh_rate` makes more sense to be part of the Progress Bar callback (in fact it already is). If someone wants to customize the progress bar refresh rate, they must create a Progress Bar callback, just like if they wanted to customize any other part of the the progress bar.

### Pitch
Deprecate `progress_bar_refresh_rate` from Trainer constructor:
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/trainer.py#L117

Don't send `progress_bar_refresh_rate` here: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/trainer.py#L415
and update the relevant functions in callback_connector.py

### Considerations
One implication of this is there will no longer be a way to specify no progress bar callback - one will always be created by default (currently if you dont want a progress bar you can set `progress_bar_refresh_rate` equal to 0, which is a bit hacky IMO). Do people see this as a problem?

daniellepintz:
## What does this PR do?
Fixes #9500

## Before submitting

## PR review

## Did you have fun?


ananthsub (commented):
One of the roles that progress bar refresh rate is serving is to disable the progress bar by setting the rate to 0. 

IMO, with this flag deprecated, we should have an `enable_progress_bar: bool = True` on the trainer constructor to make this easy to toggle.

This would match what we've done for checkpointing and what we will do for the model summary. 

@awaelchli what do you think?

ananthsub:
> In what cases would a user want to disable progress bar? Also, what do you think about [#9500 (comment)](https://github.com/PyTorchLightning/pytorch-lightning/issues/9500#issuecomment-918628291)

1. user preference: in case people find it unhelpful or visually unappealing or spamming outputs for whatever reason, they should be able to disable it
2. unit tests: its never looked at there, so we could disable it for marginally faster tests and reduced log output in the tests

tchaton:
> One of the roles that progress bar refresh rate is serving is to disable the progress bar by setting the rate to 0.
> 
> With this flag deprecated, we still need a way for users to disable this setting.
> 
> The pattern for features that are by default enabled has been to have a boolean flag for this. For example `enable_progress_bar: bool = True` on the trainer constructor makes this easy to toggle on/off.
> 
> This would match what we've done for checkpointing and what we will do for the model summary.
> 
> @awaelchli what do you think?

I would be fine with this option, but to be consistent, let call it `progress_bar_callback: bool = True/False,`

daniellepintz:
sounds good I will add a boolean flag to enable the progress_bar. @tchaton I am not a big fan of calling it `progress_bar_callback` because then it sounds like the argument is literally a progress bar callback, and it is not clear that it is a boolean flag. I believe we have the same issue with `checkpoint_callback`, and I think it would make sense to change that flag to something like `enable_checkpoint_callback`

tchaton:
> sounds good I will add a boolean flag to enable the progress_bar. @tchaton I am not a big fan of calling it `progress_bar_callback` because then it sounds like the argument is literally a progress bar callback, and it is not clear that it is a boolean flag. I believe we have the same issue with `checkpoint_callback`, and I think it would make sense to change that flag to something like `enable_checkpoint_callback`

I agree. Ideally, we would have `enable_checkpointing` and `enable_progress_bar`

CHANGELOG.md
```diff
@@ -249,8 +249,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated passing `flush_logs_every_n_steps` as a Trainer argument, instead pass it to the logger init if supported ([#9366](https://github.com/PyTorchLightning/pytorch-lightning/pull/9366))
 
 
-- Deprecated `LightningLoggerBase.close`, `LoggerCollection.close` in favor of `LightningLoggerBase.finalize`, `LoggerCollection.finalize` ([#9422](https://github.com/PyTorchLightning/pytorch-lightning/pull/9422))
+- Deprecated passing `progress_bar_refresh_rate` to the `Trainer` constructor in favor of adding the `ProgressBar` callback with `refresh_rate` directly to the list of callbacks ([#9616] (https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
```
awaelchli, line 252:
```suggestion
- Deprecated passing `progress_bar_refresh_rate` to the `Trainer` constructor in favor of adding the `ProgressBar` callback with `refresh_rate` directly to the list of callbacks ([#9616](https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
```
there is a problem here, one empty space too much

pytorch_lightning/trainer/trainer.py
```diff
@@ -137,7 +137,7 @@ def __init__(
         tpu_cores: Optional[Union[List[int], str, int]] = None,
         ipus: Optional[int] = None,
         log_gpu_memory: Optional[str] = None,
-        progress_bar_refresh_rate: Optional[int] = None,
+        progress_bar_refresh_rate: Optional[int] = None,  # TODO: remove in v1.7
```
awaelchli, line 140: will we do a follow up PR for changing all our tests that use this argument, to simplify the removal in 1.7?

awaelchli (approved):

CHANGELOG.md
daniellepintz: good eye!

pytorch_lightning/trainer/trainer.py
daniellepintz: ah yes I will do that when I add the `enable_progress_bar` flag

ananthsub (approved):

Borda (dismissed):
I am not much a fan of this deprecation I feel this shall stay in Trainer...
Progressbar is a basic feature, this pushed all users to learn and define special progress bar callback just for trivial adjustment... :face_with_thermometer: 
cc: @PyTorchLightning/core-contributors 

temp block for a day till I hear some more thoughts :rabbit:

Borda:
> I agree. Ideally, we would have `enable_checkpointing` and `enable_progress_bar`

what would be the logic then, how many progressBar callbacks you would have to define, assuming you have a bar per step in validation, training?

daniellepintz:
@Borda please see my comment here for a summary of why I think this makes sense: https://github.com/PyTorchLightning/pytorch-lightning/issues/9500#issuecomment-923411053

>this pushed all users to learn and define special progress bar callback just for trivial adjustment

whether trivial or not, I think all adjustments should go in callbacks, otherwise the Trainer constructor gets cluttered with lots of different "trivial adjustments"

daniellepintz:
> what would be the logic then, how many progressBar callbacks you would have to define, assuming you have a bar per step in validation, training?

Not sure what you mean by multiple progress bar callbacks... I believe we only allow one.  Once this deprecation is complete and we have `enable_progress_bar` (#9664), the logic will be if `enable_progress_bar` is True, create a default ProgressBar with a refresh_rate of 1 (20 if in Colab notebook) upon Trainer initialization.

SeanNaren (approved):

SeanNaren:
Great work @daniellepintz!

PR #9664, opened 2021-09-23
Add `enable_progress_bar` to Trainer constructor

### LINKED ISSUE #9500: Deprecate `progress_bar_refresh_rate` from Trainer constructor

daniellepintz:
## What does this PR do?
Follow up from #9616 
Fixes #9500

Creates a new flag `enable_progress_bar` which allows users to disable the progress bar, since the previous method, setting `progress_bar_refresh_rate` to 0, was deprecated in #9616 
Also, replace `progress_bar_refresh_rate` with `enable_progress_bar` in the tests.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -74,7 +75,20 @@ def on_trainer_init(
...
+            rank_zero_deprecation(
+                f"Setting `Trainer(progress_bar_refresh_rate={progress_bar_refresh_rate})` is deprecated in v1.5 and"
+                " will be removed in v1.7. Please pass `pytorch_lightning.callbacks.progress.ProgressBar` with"
+                " `refresh_rate` directly to the Trainer's `callbacks` argument instead."
+            )
+
+        if enable_progress_bar:
+            self.trainer._progress_bar_callback = self.configure_progress_bar(
+                progress_bar_refresh_rate, process_position
+            )
```
ananthsub, line 89: what happens if progress_bar_refresh_rate=0 and enable_progress_bar=True? 
we still have to honor backwards compat for users setting progress_bar_refresh_rate=0 and expecting the progress bar to be disabled until v1.7

daniellepintz: > what happens if progress_bar_refresh_rate=0 and enable_progress_bar=True?
same as before, we don't create a progress bar. the logic in configure_progress_bar isn't changing

awaelchli (approved):
we would need one more follow up to also update the occurrences in the docs

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -82,7 +83,12 @@ def on_trainer_init(
                 " `refresh_rate` directly to the Trainer's `callbacks` argument instead."
             )
 
-        self.trainer._progress_bar_callback = self.configure_progress_bar(progress_bar_refresh_rate, process_position)
+        if enable_progress_bar:
```
carmocca, line 86: Can you mention the newly added flag in the deprecation message above?

daniellepintz:
@awaelchli added it to this PR!

CHANGELOG.md
```diff
@@ -150,6 +150,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `RichModelSummary` callback ([#9546](https://github.com/PyTorchLightning/pytorch-lightning/pull/9546))
 
 
+- Added `enable_progress_bar` to Trainer constructor ([#9664](https://github.com/PyTorchLightning/pytorch-lightning/pull/9664))
```
ananthsub, line 153: you could update the deprecation entry in the changelog for progress bar refresh rate to mention this new flag now

ananthsub (approved):

Borda (changes requested):
missing test for raising deprecation warning

ananthsub:
> missing test for raising deprecation warning

@Borda a test was added here: https://github.com/PyTorchLightning/pytorch-lightning/blob/9148a13de0b121985bdf03e2832daf424f450ae5/tests/deprecated_api/test_remove_1-7.py#L230-L233 as part of #9616

carmocca (approved):

Borda (approved):

awaelchli:
I'm asking myself now why did we prefix this with "enable_"? Just `progress_bar=True/False` would be shorter and the type makes it clear that a bool is expected.

ananthsub:
> I'm asking myself now why did we prefix this with "enable_"? Just `progress_bar=True/False` would be shorter and the type makes it clear that a bool is expected.

Likely to avoid any confusion with ProgressBar as a callback instance. Similar to how checkpoint_callback is now typed as a bool, but sometimes people still pass a callback instance there

daniellepintz:
@awaelchli Yeah it's exactly what @ananthsub said. See discussion here https://github.com/PyTorchLightning/pytorch-lightning/pull/9616#issuecomment-924387620

PR #9666, opened 2021-09-23
Add warnings regarding unsupported keys in optim config and OneCycleLR

### LINKED ISSUE #9630: Improve LR Schedulers doc
## 📚 Documentation

The documentation on [configure_optimizer](https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html#configure-optimizers) seems to be confusing to newcomers, regarding the proper configuration of LR Schedulers through the use of the `lr_scheduling_config`. This leading to multiple issues that could be easily avoided, see #4929 #6948 #9475 #3869 #9300

In my particular case I was using OneCycleLR and unknowingly was doing steps on each epoch instead of on each step, see pytorch/pytorch#65417. As stated by @cowwoc this is an issue a lot of people stumble upon.

Even though the information is available on the documentation it may not be very user-friendly. The solution I incline for is adding an example as it is done for `ReduceLROnPlateau` but with the default configuration necessary for `OneCycleLR`.

kingyiusuen:
## What does this PR do?

Fixes #9630 

- Added a warning when unsupported keys are found in a optimizer config dict.
- Added a warning when a OneCycleLR scheduler is used and "interval" is "epoch".

## Before submitting

- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [X] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [X] Is this pull request ready for review? (if not, please submit in draft mode)
- [X] Check that all items from **Before submitting** are resolved
- [X] Make sure the title is self-explanatory and the description concisely explains the PR

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/optimizers.py
+++ b/pytorch_lightning/trainer/optimizers.py
@@ -222,9 +222,9 @@ class _MockOptimizer(Optimizer):
 
 def _validate_optim_conf(optim_conf: Dict[str, Any]) -> None:
     valid_keys = {"optimizer", "lr_scheduler", "frequency", "monitor"}
-    extra_keys = [k for k in optim_conf.keys() if k not in valid_keys]
+    extra_keys = optim_conf.keys() - valid_keys
     if extra_keys:
-        rank_zero_warn(f"Found unsupported keys in the optimizer configuration: {extra_keys}", RuntimeWarning)
+        rank_zero_warn(f"Found unsupported keys in the optimizer configuration: {set(extra_keys)}", RuntimeWarning)
 
 
 def _validate_scheduler_optimizer(optimizers, lr_schedulers):
--- a/pytorch_lightning/trainer/optimizers.py
+++ b/pytorch_lightning/trainer/optimizers.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from abc import ABC
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import torch
 from torch import optim
@@ -40,6 +40,15 @@ class TrainerOptimizersMixin(ABC):
             )
             optim_conf = _MockOptimizer()
 
+        optimizers, lr_schedulers, optimizer_frequencies, monitor = self._configure_optimizers(optim_conf)
+        lr_schedulers = self._configure_schedulers(lr_schedulers, monitor, not pl_module.automatic_optimization)
+        _validate_scheduler_optimizer(optimizers, lr_schedulers)
+        return optimizers, lr_schedulers, optimizer_frequencies
+
+    @staticmethod
+    def _configure_optimizers(
+        optim_conf: Union[Dict[str, Any], List, Optimizer, Tuple]
+    ) -> Tuple[List, List, List, Optional[str]]:
         optimizers, lr_schedulers, optimizer_frequencies = [], [], []
         monitor = None
 
@@ -98,12 +107,7 @@ class TrainerOptimizersMixin(ABC):
                 ' * {"optimizer": `torch.optim.Optimizer`, (optional) "lr_scheduler": `torch.optim.lr_scheduler`}\n'
                 ' * A list of the previously described dict format, with an optional "frequency" key (int)'
             )
-
-        is_manual_optimization = not pl_module.automatic_optimization
-        lr_schedulers = self.configure_schedulers(lr_schedulers, monitor, is_manual_optimization)
-        _validate_scheduler_optimizer(optimizers, lr_schedulers)
-
-        return optimizers, lr_schedulers, optimizer_frequencies
+        return optimizers, lr_schedulers, optimizer_frequencies, monitor
 
     def convert_to_lightning_optimizers(self):
         def _convert_to_lightning_optimizer(trainer, optimizer):
@@ -116,8 +120,9 @@ class TrainerOptimizersMixin(ABC):
             opt_idx: _convert_to_lightning_optimizer(self, opt) for opt_idx, opt in enumerate(self.optimizers)
         }
 
-    def configure_schedulers(
-        self, schedulers: list, monitor: Optional[str], is_manual_optimization: bool
+    @staticmethod
+    def _configure_schedulers(
+        schedulers: list, monitor: Optional[str], is_manual_optimization: bool
     ) -> List[Dict[str, Any]]:
         """Convert each scheduler into dict structure with relevant information."""
         lr_schedulers = []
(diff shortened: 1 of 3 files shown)


pytorch_lightning/trainer/optimizers.py
```diff
@@ -205,6 +215,13 @@ def __repr__(self):
         return "No Optimizer"
 
 
+def _validate_optim_conf(optim_conf):
```
carmocca, line 218: Add typing

tests/trainer/optimization/test_optimizers.py
```diff
@@ -99,6 +99,19 @@ def test_reducelronplateau_with_no_monitor_in_lr_scheduler_dict_raises(tmpdir):
...
 
+def test_onecyclelr_with_epoch_interval_dict_warns(tmpdir):
+    """Test warning when a OneCycleLR is used and interval is epoch."""
+    model = BoringModel()
+    optimizer = optim.Adam(model.parameters())
+    model.configure_optimizers = lambda: {
+        "optimizer": optimizer,
+        "lr_scheduler": {"scheduler": optim.lr_scheduler.OneCycleLR(optimizer, max_lr=0.01, total_steps=3)},
+    }
+    trainer = Trainer(default_root_dir=tmpdir, fast_dev_run=True)
```
carmocca, line 110: I think it would be great to split `init_optimizers` into two parts, one static so that this can be tested without instantiating a `Trainer`.

Something like this:

```python
def init_optimizers(self, model: Optional["pl.LightningModule"]) -> Tuple[List, List, List]:
        pl_module = self.lightning_module or model
        self._lightning_optimizers = None
        optim_conf = self.call_hook("configure_optimizers", pl_module=pl_module)
        if optim_conf is None:
            rank_zero_warn(
                "`LightningModule.configure_optimizers` returned `None`, this fit will run with no optimizer",
                UserWarning,
            )
            optim_conf = _MockOptimizer()
        
        optimizers, lr_schedulers, optimizer_frequencies = self._configure_optimizers(optim_conf)
        lr_schedulers = self._configure_schedulers(lr_schedulers, monitor, not pl_module.automatic_optimization)
        _validate_scheduler_optimizer(optimizers, lr_schedulers)
        return optimizers, lr_schedulers, optimizer_frequencies

@staticmethod
def _configure_optimizers(...)

@staticmethod
def _configure_schedulers(...)
```

pytorch_lightning/trainer/optimizers.py
```diff
@@ -205,6 +220,13 @@ def __repr__(self):
         return "No Optimizer"
 
 
+def _validate_optim_conf(optim_conf: Dict[str, Any]) -> None:
+    valid_keys = {"optimizer", "lr_scheduler", "frequency", "monitor"}
+    extra_keys = [k for k in optim_conf.keys() if k not in valid_keys]
+    if extra_keys:
+        rank_zero_warn(f"Found unsupported keys in the optimizer configuration: {extra_keys}", RuntimeWarning)
```
rohitgr7, line 227: should it raise an exception here? I have seen users passing scheduler config like `interval` and `scheduler` in optimizer config and then expects it to work.. warnings are ignored sometimes but atleast it should stop at the beginning rather than releasing the mistake after an epoch/some epochs while logging at the logs.

carmocca: I don't think so considering they will just get ignored and shouldn't impact the flow.

carmocca (approved):

pytorch_lightning/trainer/optimizers.py
rohitgr7: Exactly my point.. they will be ignored with a warning and it will continue to run with possibly due to the incorrect config.. and possibly user might not know about this if they ignore the warning. Since optimizers and schedulers highly impact the training, we should ensure they are correct from users side and tell them at the beginning only.

awaelchli: we could consider typing the dict for users who want stricter checking with mypy. something for the future maybe.

ananthsub: > I don't think so considering they will just get ignored and shouldn't impact the flow.

is it a misconfiguration from the user POV? as in, they input these keys expecting them to take effect but they don't? If so, i'd be in favor of raising an exception to warn earlier. 

i am also in favor of a typed dict/dataclass that users can return which makes the schema explicit and makes it harder to input bad configs

kingyiusuen: Please let me know how I should proceed with this PR. Should I raise an exception and create a typed dict?

carmocca: I'd say create a `TypedDict`.

This has the advantage of easily seeing all fields and expected types. With an exception you have to convey the information through text which is not as useful.

Additionally, raising an exception can be a breaking change in case people use `configure_optimizers` for other things.

The `TypedDict` improvement should be done in a follow-up PR

awaelchli (approved):
LGTM
users may not know how to turn off the warning

rohitgr7 (approved):

pytorch_lightning/trainer/optimizers.py
```diff
@@ -205,6 +220,13 @@ def __repr__(self):
         return "No Optimizer"
 
 
+def _validate_optim_conf(optim_conf: Dict[str, Any]) -> None:
+    valid_keys = {"optimizer", "lr_scheduler", "frequency", "monitor"}
```
cowwoc, line 224: Something doesn't look right here. Based on https://pytorch-lightning.readthedocs.io/en/latest/api/pytorch_lightning.core.lightning.html#pytorch_lightning.core.lightning.LightningModule.configure_optimizers we're expecting something like this:

```
"optimizer": optimizer,
        "lr_scheduler": {
            "scheduler": ReduceLROnPlateau(optimizer, ...),
            "monitor": "metric_to_track",
            "frequency": 5,
        }
```

This function seems to assume all keys will show up on the same level.

rohitgr7: yep, this function just checks the optimizer level keys.. doesn't go inside the scheduler config

cowwoc: Are frequency and monitor even legal on this level? I only ever saw them used inside the scheduler config.

And can we also validate the scheduler config?

rohitgr7: yes monitor is legal at this level, although not a big fan.. makes it a bit confusing IMO for some..
frequency here belongs to the optimizer.. usually used in case of multiple optimizers.
yes we have validation of scheduler config as well here: https://github.com/PyTorchLightning/pytorch-lightning/blob/3392215ef69b483bff82bf7b0270c2349b33a73c/pytorch_lightning/trainer/optimizers.py#L150-L152

cowwoc: Okay, that works. Thanks for the clarification!

PR #9669, opened 2021-09-23
[Refactor] Cleanout the logic associated to replacing the ddp sampler 

### LINKED ISSUE #9655: Sampler is being replaced even with trainer.replace_sampler_ddp=False
## 🐛 Bug

In the following line:

https://github.com/PyTorchLightning/pytorch-lightning/blob/4bebf82f271fb577ed41831403551f0bff97651b/pytorch_lightning/trainer/data_loading.py#L152

PTL tries to replace the sampler object even if `trainer.replace_sampler_ddp` is `False`:

https://github.com/PyTorchLightning/pytorch-lightning/blob/4bebf82f271fb577ed41831403551f0bff97651b/pytorch_lightning/trainer/data_loading.py#L135

### To Reproduce

Should be easy to reproduce with any custom Sampler class that has different init args from the PyTorch default.

### Expected behavior

Function in line 152 should not be called and moved just below 147. This is how it was in version v1.4.8:

https://github.com/PyTorchLightning/pytorch-lightning/blob/1.4.8/pytorch_lightning/trainer/data_loading.py#L143

### Environment

- PyTorch Lightning Version (main):
- PyTorch Version (e.g., 1.10)
- Python version: 3.8.10
- OS (e.g., Linux): Linux
- CUDA/cuDNN version: N/A
- How you installed PyTorch (`conda`, `pip`, source): source

tchaton:
## What does this PR do?

Fixes #9655

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/data_loading.py
```diff
@@ -119,172 +112,59 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
+            and self.accelerator_connector.is_distributed
+            and not isinstance(dataloader.sampler, DistributedSampler)
+            and not has_iterable_dataset(dataloader)
+        )
+
+    def prepare_dataloader(self, dataloader: Any, shuffle: bool, mode: Optional[RunningStage] = None) -> Any:
+        """This function handles to following functionalities:
+
+        - Injecting a `DistributedDataSampler` within the `DataLoader if accelertor is is_distributed
+        - Wrapping the datasets and samplers into Fault Tolerant Components
```
carmocca, line 127:
```suggestion
        - Wrapping the datasets and samplers into fault-tolerant components
```

pytorch_lightning/trainer/data_loading.py
```diff
@@ -119,172 +112,59 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
+                dataloader, shuffle, mode=mode, overfit_batches=self.overfit_batches, **self.distributed_sampler_kwargs
+            )
         else:
-            # use current sampler
             sampler = dataloader.sampler
 
-        dataloader = self.replace_sampler(dataloader, sampler, mode=mode)
+        # the DataLoader should be re-created only if we need to inject
+        # the fault tolerant components or the distributed sampler.
+        if _fault_tolerant_training() or isinstance(sampler, DistributedSampler):
```
carmocca, line 156: Isn't the condition: `or isinstance(sampler, DistributedSampler)` wrong when the above if resolved to the `else` clause and the user already set a `DistributedSampler`?

We wouldn't need to call this in that case, right?

pytorch_lightning/utilities/dataloader_preparation.py
```diff
@@ -0,0 +1,163 @@
+# Copyright The PyTorch Lightning team.
```
carmocca, line 1: Why does this code need to be moved into an utility?

IMO it's better to have these things together and we don't want to bloat the utilities.

Additionally, moving it messes up the git history so unless there's is a good reason, I wouldn't do it

pytorch_lightning/trainer/data_loading.py
```diff
@@ -119,172 +112,59 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
-                dl_kwargs["dataset"] = CaptureMapDataset(dataset=dl_kwargs["dataset"])
-            else:
-                raise MisconfigurationException(
-                    "This shouldn't happen, please open an issue on Lightning Github repository."
-                )
-
-        return dl_kwargs
-
-    @staticmethod
-    def replace_sampler(dataloader: DataLoader, sampler, mode: Optional[RunningStage] = None) -> DataLoader:
```
carmocca, line 272: Somebody might be using this, just pointing it out.

pytorch_lightning/utilities/dataloader_preparation.py
tchaton: data loading is quite complex and I think grouping the utilities dataloader would help to understand the code better.

carmocca: I'd argue that grouping the utilities in the same file `data_loading.py` is more readable since you avoid the file jump. It also makes patching and subclassing easier.

tests/trainer/test_supporters.py
```diff
@@ -325,22 +332,46 @@ def __len__(self):
...
+        dataloader = trainer.prepare_dataloader(dataloader, shuffle=True)
+        _count = 0
 
     def _assert_distributed_sampler(v):
         nonlocal _count
         _count += 1
-        assert isinstance(v, DistributedSampler)
+        if use_fault_tolerant:
+            assert isinstance(v, FastForwardSampler)
+            v = v._sampler
```
awaelchli, line 361: does this do anything?

awaelchli (approved):

tests/trainer/test_supporters.py
tchaton: Yes, it extracts the wrapped sampler to make sure it is a distributed sampler. Added a check to make sure we pass within the part of the code.

pytorch_lightning/utilities/dataloader_preparation.py
tchaton: Fine ! Done !

pytorch_lightning/trainer/data_loading.py
tchaton: Done !

pytorch_lightning/trainer/data_loading.py
```diff
@@ -67,224 +67,6 @@ class TrainerDataLoadingMixin(ABC):
     dev_debugger: InternalDebugger
     call_hook: Callable
 
-    def _worker_check(self, dataloader: DataLoader, name: str) -> None:
```
carmocca, line 70: Can you put the code where it was? This diff is very hard to review.

If moving the position of the code was intended, then it should be in 2 PRs. 1 modifying the code and another one moving it

PR #9614, opened 2021-09-20
add `_is_fresh_start_epoch` attribute to fit_loop to avoid reload dl twice when training starts

ninginthecloud:
## What does this PR do?

Fixes #9502 

This PR aims to fix the bug mentioned in issue #9502. 
During fit loop, train_dataloader is loaded twice when resuming from checkpoint. This underlying behavior does not meet user's expectation. 
There are two options to fix this issue:
- **Option 1** (implemented in this PR): introduce an attribute `_is_fresh_start_epoch` in fit_loop to track if a new fit_loop is called. `_is_fresh_start_epoch` can be `True` when 
1. Trainer. fit(model), current_epoch = 0
2. Trainer.fit(model, dm), current_epoch != 0 but new dataloader/datamoudle is provided
3. trainer=Trainer(resume_from_checkpoint=ckpt), Trainer.fit(), current_epoch!=0 but Trainer.fit() resumes from checkpoint

- **Option 2**: remove `self.reset_train_val_dataloaders(model)` in `Trainer._run_train()` and set the reload logics in corresponding loops. When I worked on this option, I found a lot of other underlying assumptions:
1. `self.num_training_batches = 0` in `trainer._setup_on_init()`. This value should be initialized as `self.num_training_batches = float('inf')`, otherwise, if we remove dataloader reloading in `Trainer._run_train()` before calling fit_loop.run(), the `skip` flag from `fit_loop` will be true. which makes `fit_loop` falsely stops. 
2. we allow users to call `trainer.fit()` multiple times with different dataloaders. removing dataloader reloading in `Trainer._run_train()`, it marks `fit_loop` unable to continue fitting with new data. In this sense, we still need  attribute `_is_fresh_start_epoch` mentioned in Option 1. 
3. `reset_val_dataloader()` is called before fit_loop officially starts. Sometimes it's called if sanity_check is enabled. naively removing `self.reset_train_val_dataloaders(model)` in `Trainer._run_train()` breaks a lot of tests since our usage and assumption of loading `val_dataloder()` changed. As mentioned in issue #8738, it's better to provide more gradually control of reloading train and val dataloader seperately. 

In summary, this PR adopts the **option 1** to quickly fix this bug. In the next step, let's complete issue #8738 and then refactor dataloader reloading. 
 
## Before submitting

- [X] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
[...]

PR #9671, opened 2021-09-23
remove reset_train_val_dataloaders(model) from Trainer and move data reloading logic to loop

### LINKED ISSUE #9502: Dataloader is reloaded twice after resuming from checkpoint
### Motivation

PyTorch Lightning reloads dataloader twice after resuming from checkpoint:
the first time reloads before train loop start [[link](https://github.com/PyTorchLightning/pytorch-lightning/blob/d2def369ec0ecdcf65acbe834c75cd2ff778d491/pytorch_lightning/trainer/trainer.py#L1122)];
The second reloads of default value for `reload_dataloaders_every_n_epochs` [[link](https://github.com/PyTorchLightning/pytorch-lightning/blob/6ff43cbff7a7b4ea75cd3d17b27aa7addf46421c/pytorch_lightning/loops/fit_loop.py#L202)]
This behavior could be problematic:
- This behavior could bring user confusion. They expect `reload_dataloaders_every_n_epochs = 1` forces the dataloaders to reload once every n epochs, but it loads multiple times sometimes. 
- The dataloader loading behavior is inconsistent between resuming from checkpoint and fresh training start. This could break the internal states of dataloader, which requires users to keep track of global states. 
- It may impact Fault tolerance training correctly captures the state of dataloader

### Pitch
Let's remove  `self.reset_train_val_dataloaders(model)` in `_run_train()` [[link](https://github.com/PyTorchLightning/pytorch-lightning/blob/d2def369ec0ecdcf65acbe834c75cd2ff778d491/pytorch_lightning/trainer/trainer.py#L1122)]
Because reset_*_dataloader() will be called in fit loop, there's no need to call it in trainer. 

### Additional context

<sub>

</sub>

ninginthecloud:
## What does this PR do?

Fixes #9502 

This PR aims to fix the bug mentioned in issue #9502. 
During fit loop, train_dataloader is loaded twice when resuming from checkpoint. This underlying behavior does not meet user's expectation. 
As discussed in an closed PR #9614, we will follow the second option to fix this issue: 
- **Option 2**: remove `self.reset_train_val_dataloaders(model)` in `Trainer._run_train()` and set the reload logics in corresponding loops. Additionally, 
1. `self.num_training_batches = 0` in `trainer._setup_on_init()`. This value is initialized as `self.num_training_batches = float('inf')`, 
2. `fit_loop.done` and `fit_loop.skip` has been updated. 
```python
@property
    def done(self) -> bool:
        """Evaluates when to leave the loop.

        Returns True if trainer.should_stop was set (e.g. by early stopping) or if the maximum number of steps or epochs
        is reached.
        """
        return stop_steps or should_stop or stop_epochs or self.trainer.num_training_batches == 0

@property
    def skip(self) -> bool:
        """Whether we should skip the training and immediately return from the call to :meth:`run`."""
        return self.done or self.trainer.limit_train_batches == 0
```
 
## Before submitting

- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?

## PR review

## Did you have fun?


pytorch_lightning/loops/base.py
```diff
@@ -49,6 +49,7 @@ class Loop(ABC, Generic[T]):
 
     def __init__(self) -> None:
         self.restarting = False
+        self.is_fresh_start_epoch: bool = True
```
ananthsub, line 52: i do not think we should be introducing this attribute to the base loop interface. not all loops implement epoch-based semantics.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -161,12 +161,12 @@ def done(self) -> bool:
...
         self.trainer.should_stop = should_stop
 
-        return stop_steps or should_stop or stop_epochs
+        return stop_steps or should_stop or stop_epochs or self.trainer.num_training_batches == 0
 
     @property
     def skip(self) -> bool:
         """Whether we should skip the training and immediately return from the call to :meth:`run`."""
-        return self.done or self.trainer.num_training_batches == 0
+        return self.done or self.trainer.limit_train_batches == 0
```
ananthsub, line 169: @awaelchli @tchaton n00b question: why do we need `skip` at all vs directly calling `run` and having it exit?

pytorch_lightning/trainer/trainer.py
```diff
@@ -511,7 +511,7 @@ def _setup_on_init(self, num_sanity_val_steps: int) -> None:
 
         self.should_stop = False
         self.state = TrainerState()
-        self.num_training_batches = 0
+        self.num_training_batches = float("inf")
```
ananthsub, line 514: is this inconsistent if we don't do the same for validation/test/predict batches too?

tests/callbacks/test_progress_bar.py
```diff
@@ -146,9 +146,10 @@ def test_progress_bar_fast_dev_run(tmpdir):
     assert 1 == progress_bar.val_batch_idx
     assert 0 == progress_bar.test_batch_idx
 
-    # the main progress bar should display 2 batches (1 train, 1 val)
-    assert 2 == progress_bar.main_progress_bar.total
-    assert 2 == progress_bar.main_progress_bar.n
+    # the main progress bar should display 1 batch, as max_steps = 1
+    # when fast_dev_run=True
+    assert 1 == progress_bar.main_progress_bar.total
+    assert 1 == progress_bar.main_progress_bar.n
```
ananthsub, line 152: this is indicating that this is a breaking change in the API

tests/trainer/test_data_loading.py
```diff
@@ -115,8 +116,8 @@ def on_train_start(self, *_) -> None:
     else:
         warn_str = "Consider setting accelerator=ddp"
 
-    trainer = Trainer(accelerator="ddp_spawn")
     with pytest.warns(UserWarning, match=warn_str), pytest.raises(SystemExit):
+        trainer = Trainer(accelerator="ddp_spawn")
```
ananthsub, line 120: nit: what was wrong with initializing the trainer outside this context?

pytorch_lightning/loops/base.py
tchaton: Agree, this look like a hack and it won't generalize to all loops.

pytorch_lightning/loops/fit_loop.py
tchaton: We believe it was cleaner to have a skip function if needed, so the code is slightly more organized. But it could be removed.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -188,7 +188,9 @@ def on_advance_start(self) -> None:
         model = self.trainer.lightning_module
 
         # reset train dataloader
-        if self.current_epoch != 0 and self.trainer._should_reload_dl_epoch:
```
tchaton, line 191: Hey @ninginthecloud. I might be wrong, but `self.current_epoch != 0` is just a small optimization to prevent re-creating the dataloader right ?

tests/callbacks/test_progress_bar.py
tchaton: I believe this is wrong and the previous behaviour seemed accurate. the main_progress should get 2 batches: (1 train, 1 val)

tests/trainer/test_data_loading.py
tchaton: I won't fail as the SystemExit is triggered within the `on_train_epoch_start ` of the Callback.

carmocca:
Marking this as a draft as a merge seems to have gone wrong, mark it as ready when it's ready again!

ninginthecloud:
> Marking this as a draft as a merge seems to have gone wrong, mark it as ready when it's ready again!

I've got rebase and made this PR in a clean stage. Looking for another round of review, thank you~ 😃 cc: @carmocca, @tchaton

PR #9691, opened 2021-09-24
Deprecate LightningDistributed and keep logic in ddp/ddpSpawn directly

### LINKED ISSUE #9692: Deprecate LightningDistributed and keep broadcast in ddp/ddpSpawn directly
1. 
LightningDistributed() class only used by ddp and ddpSpawn, and only have one broadcast function for torch collectives. It's unnecessary to have.
 Also, we have to set rank and device in set up steps. If subclass extend DDP or DDPSpawn, and overridden function where LightningDistributed.rank and device setted, could cause silent failures.

2. Now the src is not respected in torch broadcast

### Motivation
Simplify the code structure and reduce the possibilities for silent failure

### Pitch
Deprecate LightningDistributed. There is only one function
```
    def broadcast(self, obj: Any, group=_group.WORLD):
        # always wrap into a list so it can be broadcasted.
        obj = [obj]

        if self.rank != 0:
            obj = [None] * len(obj)

        broadcast_object_list(obj, 0, group=group or _group.WORLD)

        return obj[0]

```
Move to ddp and ddpSpawn
```
   def broadcast(self, obj: object, src: int = 0) -> object:
        if not distributed_available():
            raise RuntimeError("DDP is not initialized and torch.distributed is not available, can not broadcast object")
        obj = [obj]
        if self.global_rank != 0:
            obj = [None] * len(obj)
        broadcast_object_list(obj, src, group=_group.WORLD)
        return obj[0]
```

### Additional context

Related to https://github.com/PyTorchLightning/pytorch-lightning/issues/7534

<sub>

</sub>

four4fish:
## What does this PR do?

LightningDistributed() class only used by ddp and ddpSpawn, and only have one broadcast function for torch collectives. It's unnecessary to have.
Also, we have to set rank and device in set up steps. If subclass extend DDP or DDPSpawn, and overridden function where LightningDistributed.rank and device setted, could cause silent failures.

Now the src is not respected in torch broadcast, fix the issue

Fixes #9692

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -398,6 +398,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `trainer.accumulate_grad_batches` to be an int on init. Default value for it is now `None` inside Trainer ([#9652](https://github.com/PyTorchLightning/pytorch-lightning/pull/9652))
 
 
+- Fixed `broadcast` in `DDPPlugin` and ``DDPSpawnPlugin` to respects the src input  ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
+
+
 ## [1.4.8] - 2021-09-22
 
 - Fixed error reporting in DDP process reconciliation when processes are launched by an external agent ([#9389](https://github.com/PyTorchLightning/pytorch-lightning/pull/9389))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -267,6 +267,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated passing `progress_bar_refresh_rate` to the `Trainer` constructor in favor of adding the `ProgressBar` callback with `refresh_rate` directly to the list of callbacks ([#9616](https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
 
 
+- Deprecate `LightningDistributed` and move the broadcast logic to ddp/ddpSpawn directly ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
+
+
 ### Removed
 
 - Removed deprecated `metrics` ([#8586](https://github.com/PyTorchLightning/pytorch-lightning/pull/8586/))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -398,7 +398,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `trainer.accumulate_grad_batches` to be an int on init. Default value for it is now `None` inside Trainer ([#9652](https://github.com/PyTorchLightning/pytorch-lightning/pull/9652))
 
 
-- Fixed `broadcast` in `DDPPlugin` and ``DDPSpawnPlugin` to respects the src input  ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
+- Fixed `broadcast` in `DDPPlugin` and ``DDPSpawnPlugin` to respect the `src` input  ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
 
 
 ## [1.4.8] - 2021-09-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -267,7 +267,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated passing `progress_bar_refresh_rate` to the `Trainer` constructor in favor of adding the `ProgressBar` callback with `refresh_rate` directly to the list of callbacks ([#9616](https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
 
 
-- Deprecate `LightningDistributed` and move the broadcast logic to ddp/ddpSpawn directly ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
+- Deprecate `LightningDistributed` and move the broadcast logic to `DDPPlugin` and `DDPSpawnPlugin` directly ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
 
 
 ### Removed
--- a/pytorch_lightning/distributed/dist.py
+++ b/pytorch_lightning/distributed/dist.py
@@ -20,8 +20,8 @@ from pytorch_lightning.utilities.distributed import group as _group
 class LightningDistributed:
     """
     .. deprecated:: v1.5
-        This method is deprecated in v1.5and will be removed in v1.7.
-        Logic moved to ddp and ddpspawning
+        This class is deprecated in v1.5 and will be removed in v1.7.
+        The broadcast logic will be moved to the :class:`DDPPlugin` and  :class`DDPSpawnPlugin` classes.
     """
 
     def __init__(self, rank=None, device=None):
--- a/pytorch_lightning/distributed/dist.py
+++ b/pytorch_lightning/distributed/dist.py
@@ -28,7 +28,7 @@ class LightningDistributed:
     def __init__(self, rank=None, device=None):
         rank_zero_deprecation(
             "This class is deprecated in v1.5 and will be removed in v1.7."
-            "he broadcast logic will be moved to the :class:`DDPPlugin` and :class`DDPSpawnPlugin` classes."
+            "Broadcast logic is implemented directly in the :class:`TrainingTypePlugin` implementations."
         )
         self.rank = rank
         self.device = device
--- a/pytorch_lightning/distributed/dist.py
+++ b/pytorch_lightning/distributed/dist.py
@@ -27,7 +27,7 @@ class LightningDistributed:
 
     def __init__(self, rank=None, device=None):
         rank_zero_deprecation(
-            "This class is deprecated in v1.5 and will be removed in v1.7."
+            "LightningDistributed is deprecated in v1.5 and will be removed in v1.7."
             "Broadcast logic is implemented directly in the :class:`TrainingTypePlugin` implementations."
         )
         self.rank = rank
--- a/pytorch_lightning/distributed/dist.py
+++ b/pytorch_lightning/distributed/dist.py
@@ -18,6 +18,12 @@ from pytorch_lightning.utilities.distributed import group as _group
 
 
 class LightningDistributed:
+    """
+    .. deprecated:: v1.5
+        This method is deprecated in v1.5and will be removed in v1.7.
+        Logic moved to ddp and ddpspawning
+    """
+
     def __init__(self, rank=None, device=None):
         self.rank = rank
         self.device = device
--- a/pytorch_lightning/distributed/dist.py
+++ b/pytorch_lightning/distributed/dist.py
@@ -14,6 +14,7 @@
 from typing import Any
 
 from pytorch_lightning.overrides.torch_distributed import broadcast_object_list
+from pytorch_lightning.utilities import rank_zero_deprecation
 from pytorch_lightning.utilities.distributed import group as _group
 
 
@@ -21,10 +22,14 @@ class LightningDistributed:
     """
     .. deprecated:: v1.5
         This class is deprecated in v1.5 and will be removed in v1.7.
-        The broadcast logic will be moved to the :class:`DDPPlugin` and  :class`DDPSpawnPlugin` classes.
+        The broadcast logic will be moved to the :class:`DDPPlugin` and :class`DDPSpawnPlugin` classes.
     """
 
     def __init__(self, rank=None, device=None):
+        rank_zero_deprecation(
+            "This class is deprecated in v1.5 and will be removed in v1.7."
+            "he broadcast logic will be moved to the :class:`DDPPlugin` and :class`DDPSpawnPlugin` classes."
+        )
         self.rank = rank
         self.device = device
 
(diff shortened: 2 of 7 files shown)


pytorch_lightning/distributed/dist.py
```diff
@@ -18,6 +18,12 @@
 
 
 class LightningDistributed:
+    """
+    .. deprecated:: v1.5
+        This method is deprecated in v1.5and will be removed in v1.7.
+        Logic moved to ddp and ddpspawning
```
awaelchli, line 24:
```suggestion
        This class is deprecated in v1.5 and will be removed in v1.7.
        The broadcast logic will be moved to the :class:`DDPPlugin` and  :class`DDPSpawnPlugin` classes.
```

awaelchli: @four4fish we need to print a rank_zero_deprection message here in the init of the class and/or in the broadcast message below

CHANGELOG.md
```diff
@@ -267,6 +267,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated passing `progress_bar_refresh_rate` to the `Trainer` constructor in favor of adding the `ProgressBar` callback with `refresh_rate` directly to the list of callbacks ([#9616](https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
 
 
+- Deprecate `LightningDistributed` and move the broadcast logic to ddp/ddpSpawn directly ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
+
```
ananthsub, line 271:
```suggestion
- Deprecate `LightningDistributed` and move the broadcast logic to `DDPPlugin` and `DDPSpawnPlugin` directly ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))

```

awaelchli (approved):
Have a look at the `tests/deprecated_api` where we should add a quick test for the 1.7 removal :)

pytorch_lightning/distributed/dist.py
```diff
@@ -18,6 +18,12 @@
...
+    """
+
     def __init__(self, rank=None, device=None):
```
ananthsub, line 27: 1. add a deprecation message in the init so a warning shows up at runtime
2. Add a test for the deprecation here: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/deprecated_api/test_remove_1-7.py

so we know to delete this class in v1.7

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -403,7 +396,11 @@ def barrier(self, *args, **kwargs) -> None:
             torch.distributed.barrier()
 
     def broadcast(self, obj: object, src: int = 0) -> object:
-        return self.dist.broadcast(obj)
+        obj = [obj]
```
ananthsub, line 399: - please address the bug with `src` not being respected
- as discussed elsewhere, we can special case Tensor objects for a potential speedup as @yifuwang noted here: https://github.com/PyTorchLightning/pytorch-lightning/pull/9414#discussion_r707788699

four4fish: >please address the bug with src not being respected

It's address here, as src will be pass into `broadcast_object_list`, before the src was 0

>as discussed elsewhere, we can special case Tensor objects for a potential speedup as @yifuwang noted here: 2/n Consolidate collective functions - collective base and subclasses #9414

Yeah, good point. Will create a separate new feature issue address this

pytorch_lightning/distributed/dist.py
four4fish: @awaelchli @ananthsub  This is a internal use only class, do you guys think we need wait until 1.7 to remove this or it is safe to be directly removed?

awaelchli: yes I think we need to unfortunately. it's been around for a while and is not marked as experimental.

ananthsub: we can remove its usage from within the plugins, but we cannot delete the class outright. it will have to sit unused in the codebase until v1.7

awaelchli: @ananthsub's comment reminds me that we should then exclude this file also from coverage. we can add it in `setup.cfg ` and add it to the list of omitted files

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -403,7 +396,11 @@ def barrier(self, *args, **kwargs) -> None:
...
+        if self.global_rank != 0:
```
ananthsub, line 400:
```suggestion
        if self.global_rank != src:
```
does this need to be the src instead of hardcoded to rank 0?

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -403,7 +396,11 @@ def barrier(self, *args, **kwargs) -> None:
...
+            obj = [None] * len(obj)
```
ananthsub, line 401: this is not correct: obj is not guaranteed to be a list. and this will cause surprised with strings

four4fish: it been converted to a string at line 399

obj = [obj]

pytorch_lightning/plugins/training_type/ddp.py
four4fish: I think the old assumption is that torch distributed broadcast only works with broadcast from global rank 0. WIth respect the change respect src, i think here should be change to src as well. Good catch

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -396,14 +389,20 @@ def post_dispatch(self, trainer: "pl.Trainer") -> None:
...
+            raise RuntimeError("DDP is not initialized and torch.distributed is not avalible, can not broadcast object")
         if _TORCH_GREATER_EQUAL_1_8 and torch.distributed.get_backend() == "nccl":
             torch.distributed.barrier(device_ids=self.determine_ddp_device_ids())
         else:
             torch.distributed.barrier()
 
     def broadcast(self, obj: object, src: int = 0) -> object:
-        return self.dist.broadcast(obj)
+        if not distributed_available():
+            raise RuntimeError("DDP is not initialized and torch.distributed is not avalible, can not broadcast object")
```
ananthsub: this should throw automatically as part of  torch distributed's broadcast implementation

pytorch_lightning/plugins/training_type/ddp.py
ananthsub: it's possible we can save an extra list serialization by checking if `isinstance(obj, list)` and calling `broadcast_object_list` directly in that case w/o needing to convert it

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -324,7 +315,11 @@ def barrier(self, *args, **kwargs) -> None:
     def broadcast(self, obj: object, src: int = 0) -> object:
         if not distributed_available():
             return obj
-        return self.dist.broadcast(obj)
+        obj = [obj]
+        if self.global_rank != 0:
```
ananthsub, line 319: same here

pytorch_lightning/plugins/training_type/ddp.py
four4fish: Current logic seems works really wired with list obj. I'm not sure whether check list before convert is a breaking change or not
```
if obj = [1,2,3]
obj= [obj] -> obj = [[1,2,3]]
```

pytorch_lightning/distributed/dist.py
```diff
@@ -14,11 +14,22 @@
...
+    """
+    .. deprecated:: v1.5
+        This class is deprecated in v1.5 and will be removed in v1.7.
+        The broadcast logic will be moved to the :class:`DDPPlugin` and :class`DDPSpawnPlugin` classes.
+    """
+
     def __init__(self, rank=None, device=None):
+        rank_zero_deprecation(
+            "This class is deprecated in v1.5 and will be removed in v1.7."
+            "he broadcast logic will be moved to the :class:`DDPPlugin` and :class`DDPSpawnPlugin` classes."
```
ananthsub, line 31:
```suggestion
            "Broadcast logic is implemented directly in the :class:`TrainingTypePlugin` implementations."
```
?

ananthsub (approved):
i think there's an optimization opportunity for broadcast but we can pick it up in a follow up PR along with tensor specializations

pytorch_lightning/plugins/training_type/ddp.py
ananthsub: why would it be a breaking change? since we can post-process the return of `broadcast_object_list` before we return, there should be no leakage to the caller. its a minor efficiency point is all. but we can discuss this further with better support for broadcasting tensors in a separate PR

CHANGELOG.md
```diff
@@ -267,6 +267,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated passing `progress_bar_refresh_rate` to the `Trainer` constructor in favor of adding the `ProgressBar` callback with `refresh_rate` directly to the list of callbacks ([#9616](https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
 
 
+- Deprecate `LightningDistributed` and move the broadcast logic to `DDPPlugin` and `DDPSpawnPlugin` directly ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
```
ananthsub, line 270: also add a changelog entry that broadcast in DDP/DDP spawn plugins now respect the `src` input

CHANGELOG.md
```diff
@@ -392,6 +395,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed wrapping issue: avoid wrapping LightningModule with data-parallel modules when not fitting in `DDPPlugin`, `DDPSpawnPlugin`, `DDPShardedPlugin`, `DDPSpawnShardedPlugin` ([#9096](https://github.com/PyTorchLightning/pytorch-lightning/pull/9096))
 
 
+- Fixed `broadcast` in `DDPPlugin` and ``DDPSpawnPlugin` to respects the src input  ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
```
ananthsub, line 398:
```suggestion
- Fixed `broadcast` in `DDPPlugin` and ``DDPSpawnPlugin` to respect the `src` input  ([#9691](https://github.com/PyTorchLightning/pytorch-lightning/pull/9691))
```

pytorch_lightning/plugins/training_type/ddp.py
ananthsub: nit: won't len(obj) always be 1?

four4fish: It's not nit...I feel we can revisit this part of the logic in a separate issue/PR. Do we need set a None list on the ranks broadcast too?

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -270,8 +265,6 @@ def setup_distributed(self):
         init_ddp_connection(self.cluster_environment, self.torch_distributed_backend)
 
         # set the ranks and devices
```
carmocca, line 267: leftover comment

carmocca (approved):

pytorch_lightning/plugins/training_type/ddp.py
ananthsub: > Do we need set a None list on the ranks broadcast too?

yes, each rank must provide lists of equal sizes: https://pytorch.org/docs/stable/_modules/torch/distributed/distributed_c10d.html#broadcast_object_list

but we don't need to multiply by len(obj) here.

pytorch_lightning/distributed/dist.py
```diff
@@ -14,11 +14,22 @@
...
 class LightningDistributed:
+    """
+    .. deprecated:: v1.5
+        This class is deprecated in v1.5 and will be removed in v1.7.
+        The broadcast logic will be moved to the :class:`DDPPlugin` and :class`DDPSpawnPlugin` classes.
+    """
+
     def __init__(self, rank=None, device=None):
+        rank_zero_deprecation(
+            "This class is deprecated in v1.5 and will be removed in v1.7."
```
ananthsub, line 30:
```suggestion
            "LightningDistributed is deprecated in v1.5 and will be removed in v1.7."
```

tests/deprecated_api/test_remove_1-7.py
```diff
@@ -243,3 +243,10 @@ def test_v1_7_0_lightning_logger_base_close(tmpdir):
     ):
         logger = LoggerCollection([logger])
         logger.close()
+
+
+def test_v1_7_0_deprecate_lightning_distributed(tmpdir):
+    with pytest.deprecated_call(match="This class is deprecated in v1.5 and will be removed in v1.7."):
```
ananthsub, line 249:
```suggestion
    with pytest.deprecated_call(match="LightningDistributed is deprecated in v1.5 and will be removed in v1.7."):
```

pytorch_lightning/plugins/training_type/ddp.py
ananthsub:
```suggestion
            obj = [None]
```

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -324,7 +315,11 @@ def barrier(self, *args, **kwargs) -> None:
     def broadcast(self, obj: object, src: int = 0) -> object:
         if not distributed_available():
             return obj
-        return self.dist.broadcast(obj)
+        obj = [obj]
+        if self.global_rank != src:
+            obj = [None] * len(obj)
```
ananthsub, line 320:
```suggestion
            obj = [None]
```

Issue #5542, opened 2021-01-16
Load callback states while testing.


rohitgr7:
## 🚀 Feature
Load callback states while testing.

### Motivation
https://github.com/PyTorchLightning/pytorch-lightning/pull/5161#issuecomment-759083492

### Pitch
Two possible API changes:

with an additional argument `restore_states`:
```python
test(ckpt_path, restore_states=True/False)  # give an option whether to load states or not
test(model, ckpt_path, restore_states=True/False)  # same as above but will just load checkpoint states and not the model

# raise an error
test(ckpt_path=None, restore_states=True)
```
or without any additional argument:
```python
test(ckpt_path)  # always load states
test(ckpt_path=None)  # don't load any states.
test(model, ckpt_path)  # reload checkpoint states only from ckpt_path
```

### Alternatives
Alternatively, one can just reload checkpoints manually, call `on_load_checkpoint` for all the callbacks manually, and test.

PS: There may be a better solution. Open to suggestions :)
cc: @ananthsub

awaelchli:
@rohitgr7 what's the difference between your proposal and the existing

```python
trainer = Trainer(resume_from_checkpoint=x)
trainer.test()
```
?

rohitgr7:
```python
trainer = Trainer(resume_from_checkpoint=x)
trainer.test()
```
in the recent PR the reload from `resume_from_checkpoint` was disabled completely since it was reloading the model state too also `resume_from_checkpoint` is meant to resume the training I guess. If we use this then while testing with different checkpoints we need to pass that checkpoint path in both `Trainer(resume_from_checkpoint=ckpt)` and `trainer.test(ckpt_path=ckpt)`. I think with `resume_from_checkpoint` it will be easier to handle, it's just that this argument name sounds a bit misleading to me in case of testing.

ananthsub:
> in the recent PR the reload from resume_from_checkpoint was disabled completely since it was reloading the model state too

Isn't this a breaking API change? I commented on #5388 about this. What's the proposed workaround with the existing APIs?

ananthsub:
@rohitgr7 could we still call the checkpoint connector from `setup_training` inside `run_evaluation`, but add logic inside of `restore` to check for `trainer.testing` - if it's testing, we can load the model and callback states only and ignore the trainer states. this way we don't have to expose anything new in the trainer API. what do you think?

rohitgr7:
[...]

PR #9693, opened 2021-09-24
[wip] Unify checkpoint load paths

### LINKED ISSUE #9405: [checkpoint] Resolve 2 different checkpoint loading paths across `fit` vs `validate`/`test`/`predict`

Consolidate checkpoint loading code across `fit` and `validate`/`test`/`predict`

### Motivation

We have 2 different code paths for checkpoint loading
1. `Trainer.fit` and the constructor argument `resume_from_checkpoint`
2. `ckpt_path` passed to `Trainer.validate/test/predict`

Offering multiple code paths here risks divergence. Lightning must ensure a consistent experience for checkpoint loading across these different entry points.

### Background
These are the paths today.

Trainer.fit:
- Trainer constructor is initialized with `resume_from_checkpoint`: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L143
- Which initializes the checkpoint connector with `resume_from_checkpoint`: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L383
- Which sets this field in the connector: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L38
- In `_run` we call `Trainer._restore_modules_and_callbacks`: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L960-L962
- Which explicitly checks if we're fitting, and if so, first loads the checkpoint, before restoring the model 

Trainer.validate/test/predict:
- `ckpt_path` is an argument to the function
- There's a common helper to validate the checkpoint path: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L1200
- A property is set on the trainer for the corresponding path: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L671-L673
- We use an umbrella property to work across validate/test/predict: https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/trainer/properties.py#L626-L633
- There are checks for loading based on if this property is set inside `Trainer._run`
-- https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L950-L951
-- https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L1006-L1007

Notably:
- trainer.fit calls `restore_model` whereas `trainer.validate/test/predict` calls `restore_model_weights` in the checkpoint connector. https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L116-L150
- Load checkpoint weights is called before the accelerator setup while restore model is called after: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L950-L951r: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L960-L962

### Pitch

1. Unify in CheckpointConnector
`CheckpointConnector.restore_model` is a superset of `CheckpointConnector.restore_model_weights` which suggests we don't need both. 

2. Unify in trainer.py

The overall sequence in `_run` can be shared:
- Load & validate the checkpoint dict through the training type plugin
- Load the datamodule state
- Load the LightningModule state through the training type plugin
- Load the callback states
- Load the loop states
- If fitting, load the precision, optimizer, and LR scheduler states

We have 5 different properties exposed for the checkpoint path to resume from (excluding HPC stuff):
- `trainer.resume_from_checkpoint`: 
- `trainer.validated_ckpt_path`: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L671-L674
- `trainer.tested_ckpt_path`: https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L760-L762
- `trainer.predicted_ckpt_path` : https://github.com/PyTorchLightning/pytorch-lightning/blob/a079d7fccc0a9be25b40296f2a348c4b4f40c8cf/pytorch_lightning/trainer/trainer.py#L843-L845
- An umbrella property across these 3: `trainer._ckpt_path`: https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/trainer/properties.py#L626-L633
-- This is also inconsistent: the attributes initialized in validate/test/predict are public, while `_ckpt_path` is private. Why?

It's unclear what the lifecycle of these properties should be. Do successive calls to validate/test/predict end up relying on this?

Proposal: 
- Pass the checkpoint path to resume from as an argument to `Trainer._run` to avoid our dependency on these properties
- Deprecate these properties

3. Unify in the Trainer API
Proposal: deprecate `resume_from_checkpoint` from the Trainer constructor, and add a new argument `ckpt_path` to `Trainer.fit` . This provides API consistency with `validate`/`test`/`predict`

### Additional context

<sub>

</sub>

jjenniferdai:
## What does this PR do?

Fixes #9405 
## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -55,12 +55,8 @@ class CheckpointConnector:
         """Attempts to pre-load the checkpoint file to memory, with the source path determined in this priority:
 
         1. from HPC weights if found
-        2. from `resume_from_checkpoint` file if provided
-            .. deprecated:: v1.5
-            `Trainer(resume_from_checkpoint=)` is deprecated in v1.5 and will be removed in v1.7.
-            Please use `Trainer.fit(ckpt_path=)` instead.
-        3. from `checkpoint_path` file if provided
-        4. don't restore
+        2. from `checkpoint_path` file if provided
+        3. don't restore
 
         Raises:
             FileNotFoundError: If the path to the checkpoint file is provided but the file does not exist.
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -37,6 +37,11 @@ class CheckpointConnector:
     def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH] = None) -> None:
         self.trainer = trainer
         self.resume_checkpoint_path = resume_from_checkpoint
+        if resume_from_checkpoint is not None:
+            rank_zero_deprecation(
+                "Setting `Trainer(resume_from_checkpoint=)` is deprecated in v1.5 and"
+                " will be removed in v1.7. Please pass `Trainer.fit(ckpt_path=)` directly instead."
+            )
         self._loaded_checkpoint: Dict[str, Any] = {}
 
     @property
@@ -50,7 +55,7 @@ class CheckpointConnector:
         """Attempts to pre-load the checkpoint file to memory, with the source path determined in this priority:
 
         1. from HPC weights if found
-        2. from `resume_from_checkpoint` file if provided 
+        2. from `resume_from_checkpoint` file if provided
             .. deprecated:: v1.5
             `Trainer(resume_from_checkpoint=)` is deprecated in v1.5 and will be removed in v1.7.
             Please use `Trainer.fit(ckpt_path=)` instead.
@@ -155,20 +160,6 @@ class CheckpointConnector:
             for module in self.trainer.lightning_module.modules():
                 if isinstance(module, Metric):
                     module.reset()
-
-    def restore_model_weights(self, checkpoint_path: Optional[_PATH]) -> None:
-        """Restore only the model weights.
-
-        .. deprecated:: v1.5
-            This method is deprecated in v1.5 and will be removed in v1.7.
-            Please use `CheckpointConnector.restore_model` instead.
-        """
-        checkpoint = self._loaded_checkpoint
-        if checkpoint_path is not None:
-            checkpoint = self._load_and_validate_checkpoint(checkpoint_path)
-
-        self.trainer.lightning_module.on_load_checkpoint(checkpoint)
-        self.trainer.training_type_plugin.load_model_state_dict(checkpoint)
 
     def restore_training_state(self) -> None:
         """Restore the trainer state from the pre-loaded checkpoint.
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -36,7 +36,7 @@ if _OMEGACONF_AVAILABLE:
 class CheckpointConnector:
     def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH] = None) -> None:
         self.trainer = trainer
-        self.resume_checkpoint_path = resume_from_checkpoint
+        self.resume_checkpoint_path = None
         if resume_from_checkpoint is not None:
             rank_zero_deprecation(
                 "Setting `Trainer(resume_from_checkpoint=)` is deprecated in v1.5 and"
@@ -65,7 +65,7 @@ class CheckpointConnector:
         Raises:
             FileNotFoundError: If the path to the checkpoint file is provided but the file does not exist.
         """
-        self.resume_checkpoint_path = self.hpc_resume_path or self.resume_checkpoint_path or checkpoint_path
+        self.resume_checkpoint_path = self.hpc_resume_path or checkpoint_path
         checkpoint_path = self.resume_checkpoint_path
         if not checkpoint_path:
             return
@@ -111,7 +111,6 @@ class CheckpointConnector:
         Args:
             checkpoint_path: Path to a PyTorch Lightning checkpoint file.
         """
-        self.resume_checkpoint_path = checkpoint_path
         self.resume_start(checkpoint_path)
 
         # restore module states
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -46,17 +46,21 @@ class CheckpointConnector:
         if max_version is not None:
             return os.path.join(dir_path_hpc, f"hpc_ckpt_{max_version}.ckpt")
 
-    def resume_start(self) -> None:
+    def resume_start(self, checkpoint_path: Optional[_PATH] = None) -> None:
         """Attempts to pre-load the checkpoint file to memory, with the source path determined in this priority:
 
         1. from HPC weights if found
-        2. from `resume_from_checkpoint` file if provided
-        3. don't restore
+        2. from `resume_from_checkpoint` file if provided 
+            .. deprecated:: v1.5
+            `Trainer(resume_from_checkpoint=)` is deprecated in v1.5 and will be removed in v1.7.
+            Please use `Trainer.fit(ckpt_path=)` instead.
+        3. from `checkpoint_path` file if provided
+        4. don't restore
 
         Raises:
             FileNotFoundError: If the path to the checkpoint file is provided but the file does not exist.
         """
-        self.resume_checkpoint_path = self.hpc_resume_path or self.resume_checkpoint_path
+        self.resume_checkpoint_path = self.hpc_resume_path or self.resume_checkpoint_path or checkpoint_path
         checkpoint_path = self.resume_checkpoint_path
         if not checkpoint_path:
             return
@@ -94,7 +98,7 @@ class CheckpointConnector:
         state-restore, in this priority:
 
         1. from HPC weights if found
-        2. from `resume_from_checkpoint` file if provided
+        2. from `checkpoint_path` file if provided
         3. don't restore
 
         All restored states are listed in return value description of `dump_checkpoint`.
@@ -103,7 +107,7 @@ class CheckpointConnector:
             checkpoint_path: Path to a PyTorch Lightning checkpoint file.
         """
         self.resume_checkpoint_path = checkpoint_path
-        self.resume_start()
+        self.resume_start(checkpoint_path)
 
         # restore module states
         self.restore_datamodule()
@@ -153,7 +157,12 @@ class CheckpointConnector:
                     module.reset()
 
     def restore_model_weights(self, checkpoint_path: Optional[_PATH]) -> None:
-        """Restore only the model weights."""
+        """Restore only the model weights.
+
+        .. deprecated:: v1.5
+            This method is deprecated in v1.5 and will be removed in v1.7.
+            Please use `CheckpointConnector.restore_model` instead.
+        """
         checkpoint = self._loaded_checkpoint
         if checkpoint_path is not None:
             checkpoint = self._load_and_validate_checkpoint(checkpoint_path)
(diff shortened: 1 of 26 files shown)


pytorch_lightning/trainer/trainer.py
```diff
@@ -943,24 +955,18 @@ def tune(
...
-        if not self._device_type == DeviceType.TPU:
-            self.accelerator.barrier()
-        rank_zero_info(f"Loading model weights from checkpoint at {self._ckpt_path}")
-        self.checkpoint_connector.restore_model_weights(self._ckpt_path)
-
-    def _run(self, model: "pl.LightningModule") -> Optional[Union[_EVALUATE_OUTPUT, _PREDICT_OUTPUT]]:
+        if self.state.fn == TrainerFn.FITTING:
+            self.checkpoint_connector.restore_datamodule()
+            # restore callback states
+            self.checkpoint_connector.restore_callbacks()
```
ananthsub, line 965: we should restore datamodule & callback states even if we're not fitting right?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1038,12 +1041,11 @@ def _run(self, model: "pl.LightningModule") -> Optional[Union[_EVALUATE_OUTPUT,
...
         if self.accelerator.restore_checkpoint_after_pre_dispatch:
-            if self._ckpt_path:
-                self._load_checkpoint_weights()
-            self._restore_modules_and_callbacks()
+            self._restore_modules_and_callbacks(ckpt_path)
 
         # restore optimizers, etc.
-        self.checkpoint_connector.restore_training_state()
+        if self.state.fn == TrainerFn.FITTING:
+            self.checkpoint_connector.restore_training_state()
```
ananthsub, line 1048: restore training state includes things which can be resumed even if not fitting, such as the loop state.

imo we shouldn't add the check for fitting here, but rather inside the select parts inside of `restore_training_state`

pytorch_lightning/trainer/trainer.py
```diff
@@ -1733,7 +1735,7 @@ def checkpoint_callbacks(self) -> List[ModelCheckpoint]:
         return [c for c in self.callbacks if isinstance(c, ModelCheckpoint)]
 
     @property
-    def resume_from_checkpoint(self) -> Optional[Union[str, Path]]:
+    def resume_checkpoint_path(self) -> Optional[Union[str, Path]]:
```
ananthsub, line 1738: - let's use the typehint _PATH here too? 
- should this property be deprecated as well given it's entirely user input?

ananthsub (commented):
this is looking really promising @jjenniferdai ! unifying these paths will make the API consistent and help us simplify the trainer internals

fyi @carmocca who was interested in this too from #9405

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -37,7 +37,12 @@
 class CheckpointConnector:
     def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH] = None) -> None:
         self.trainer = trainer
-        self.resume_checkpoint_path = resume_from_checkpoint
+        self.resume_checkpoint_path: Optional[_PATH] = None
```
carmocca, line 40: Do we need to keep this attribute?

pytorch_lightning/trainer/trainer.py
jjenniferdai: i was wondering this too. currently i think only fitting restores `restore_datamodule`, `restore_callbacks`, and `restore_training_state` as these fns all return early if `checkpoint_connector._loaded_checkpoint` is not set, `checkpoint_connector._loaded_checkpoint` is only set in `resume_start` which is only called for fitting?

pytorch_lightning/trainer/trainer.py
```diff
@@ -438,11 +445,6 @@ def __init__(
         # Needed because of LightningOptimizer
         self._lightning_optimizers = None
 
-        # .validate() and .test() set this when they load a checkpoint
-        self.validated_ckpt_path: Optional[str] = None
-        self.tested_ckpt_path: Optional[str] = None
-        self.predicted_ckpt_path: Optional[str] = None
-
```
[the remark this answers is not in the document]
jjenniferdai: detailed in issue #9405, can directly pass the path to `_run` instead of relying on these properties

pytorch_lightning/trainer/trainer.py
jjenniferdai: yes on second thought i think this can be deprecated entirely. user passes`ckpt_path`, `resume_start`/`resume_end` will be called to set and forget this in checkpoint_connector

pytorch_lightning/trainer/connectors/checkpoint_connector.py
jjenniferdai: I think so, it helps CheckpointConnector remember which path `_loaded_checkpoint` is from, instead of passing a path to every function and loading path file contents each time. wdyt?

pytorch_lightning/trainer/trainer.py
jjenniferdai: (above comment)

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
carmocca: The intended use for these is to be able to get the actual `ckpt_path` used when `ckpt_path="best"` is passed.

Unifying the load paths shouldn't have anything to do with them.

These are ancient attributes. We can discuss removing them but IMO it should be done in a separate PR.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
carmocca: If we deprecate the following: https://github.com/PyTorchLightning/pytorch-lightning/pull/9693/files#r716905724
then I think we can remove it. The only thing that's not issue is keeping `rank_zero_info` message in `resume_end`

jjenniferdai: aside from losing the message, would it be odd keeping `_loaded_checkpoint` without `resume_checkpoint_path` (where is `_loaded_checkpoint` from?)

carmocca: I wouldn't say so

Any preference @ananthsub or @awaelchli?

awaelchli: I recommend not keeping the loaded checkpoint in memory beyond the resuming procedure. Please keep the log message acknowledging loading of the checkpoint and the path to it.

pytorch_lightning/trainer/trainer.py
ananthsub: @awaelchli @carmocca @rohitgr7 do you know if this was intentional?

if not fitting is it expected to only load the weights, and not data/callback states? @rohitgr7 is fixing the trainer state when not fitting here: https://github.com/PyTorchLightning/pytorch-lightning/issues/9404

pytorch_lightning/trainer/trainer.py
```diff
@@ -1740,10 +1742,6 @@ def checkpoint_callbacks(self) -> List[ModelCheckpoint]:
         in the Trainer.callbacks list."""
         return [c for c in self.callbacks if isinstance(c, ModelCheckpoint)]
 
-    @property
-    def resume_from_checkpoint(self) -> Optional[Union[str, Path]]:
-        return self.checkpoint_connector.resume_checkpoint_path
```
ananthsub, line 1745: we can't delete this outright. this is a public property, part of the trainer API, which needs to go through the deprecation cycle before its removal

pytorch_lightning/trainer/trainer.py
rohitgr7: in #9413 we are now loading the checkpoint to retrieve the `global_step` and `current_epoch` if not fitting and nothing else. Not sure about datamodule but for callbacks we never loaded any states if not fitting. we have an old issue to discuss whether to load the callback states or not here: https://github.com/PyTorchLightning/pytorch-lightning/issues/5542

pytorch_lightning/trainer/trainer.py
rohitgr7: yeah, we do restore loops there and now some other attributes as well. I'd suggest waiting for this one to get merged: https://github.com/PyTorchLightning/pytorch-lightning/pull/9413

pytorch_lightning/trainer/trainer.py
```diff
@@ -595,14 +604,17 @@ def fit(
...
+            self._fit_impl, model, train_dataloaders, val_dataloaders, datamodule, ckpt_path
+        )
 
     def _fit_impl(
         self,
         model: "pl.LightningModule",
         train_dataloaders: Optional[Union[TRAIN_DATALOADERS, LightningDataModule]] = None,
         val_dataloaders: Optional[EVAL_DATALOADERS] = None,
         datamodule: Optional[LightningDataModule] = None,
+        ckpt_path: Optional[str] = None,
```
ananthsub, line 617: should this also be typed as `_PATH` ?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1046,12 +1051,13 @@ def _run(self, model: "pl.LightningModule") -> Optional[Union[_EVALUATE_OUTPUT,
...
-                self._load_checkpoint_weights()
-            self._restore_modules_and_callbacks()
+            self._restore_modules_and_callbacks(ckpt_path)
 
         # restore optimizers, etc.
-        self.checkpoint_connector.restore_training_state()
+        if self.state.fn == TrainerFn.FITTING:
+            self.checkpoint_connector.restore_training_state()
+
+        self.checkpoint_connector.resume_end()
```
ananthsub, line 1060: n00b question: why is this bumped up to here vs in `_pre_training_routine` ?

jjenniferdai: now that this calls `resume_start` for all (not only fitting), similarly `resume_end` for all as well

tchaton:
Dear @@jjenniferdai,

Any updates ?

Best,
T.C

jjenniferdai:
sorry all for the git issues :( please see #10061 instead

PR #9699, opened 2021-09-25
Add `enable_model_summary` flag

### LINKED ISSUE #9043: [RFC] Deprecate `weights_summary` off the Trainer constructor

ananthsub:
## What does this PR do?

Fixes #9043 
Adds `enable_model_summary` flag to match `enable_progress_bar` flag being added here: https://github.com/PyTorchLightning/pytorch-lightning/pull/9664/

This PR also deprecates the Trainer `weights_summary` flag and property which dictated the mode for display. Users are expected to customize this with the callback. This deprecation was required because we needed to support arbitrary module depths for summarization (int) while the current mode only supported a few fixed modes ("top" vs "full"). This customization is much easier with the ModelSummary callbacks @kaushikb11  introduced.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -145,24 +146,43 @@ def _configure_checkpoint_callbacks(self, checkpoint_callback: bool) -> None:
...
+                " in v1.7. Please set `Trainer(enable_model_summary=False) instead."
+            )
+            return
+        # Prior default in the Trainer for `weights_summary` which we explicitly check here
+        # to preserve backwards compatibility
+        if weights_summary != "top":
+            rank_zero_deprecation(
+                f"Setting `Trainer(weights_summary={weights_summary})` is deprecated in v1.5 and will be removed"
+                " in v1.7. Please pass `pytorch_lightning.callbacks.model_summary.ModelSummary` with"
+                " `max_depth` directly to the Trainer's `callbacks` argument instead."
```
rohitgr7, line 168:
```suggestion
                f"Setting `Trainer(weights_summary={weights_summary})` is deprecated in v1.5 and will be removed "
                "in v1.7. Please pass `pytorch_lightning.callbacks.model_summary.ModelSummary` with "
                "`max_depth` directly to the Trainer's `callbacks` argument instead."
```
same here.

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -145,24 +146,43 @@ def _configure_checkpoint_callbacks(self, checkpoint_callback: bool) -> None:
...
+            return
         if any(isinstance(cb, ModelSummary) for cb in self.trainer.callbacks):
             return
-        if weights_summary is not None:
+        if weights_summary is None:
+            rank_zero_deprecation(
+                "Setting `Trainer(weights_summary=None)` is deprecated in v1.5 and will be removed"
+                " in v1.7. Please set `Trainer(enable_model_summary=False) instead."
+            )
+            return
```
rohitgr7, line 161: isn't the warning below enough? you can do a check below if it's None and return from there itself.

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -145,24 +146,43 @@ def _configure_checkpoint_callbacks(self, checkpoint_callback: bool) -> None:
         if not self._trainer_has_checkpoint_callbacks() and checkpoint_callback is True:
             self.trainer.callbacks.append(ModelCheckpoint())
 
-    def _configure_model_summary_callback(self, weights_summary: Optional[str] = None) -> None:
+    def _configure_model_summary_callback(
+        self, enable_model_summary: bool, weights_summary: Optional[str] = None
+    ) -> None:
+        if not enable_model_summary:
+            return
         if any(isinstance(cb, ModelSummary) for cb in self.trainer.callbacks):
```
rohitgr7, line 154: what if there is a callback and weights_summary is not None? shouldn't we raise an exception for that case?

ananthsub: I don't see a way we can do this now, as the existing default is `weights_summary="top"`
If we change the default value to `weights_summary=None`, then it's a breaking change, as users today use `weights_summary=None` to configure whether this is displayed or not. After this PR, they'd be forced to use `enable_model_summary=False` since there'd be no way to distinguish precedence/user input vs defaults between `enable_model_summary=True` and `weights_summary=None` 

So if we want to preserve the existing disable path, we can't change the default for `weights_summary`. 
On the other hand, if someone is using a custom model summary callback, they need all of `enable_model_summary=True`, `weights_summary != None` and the custom callback extending `ModelSummary`. and raising an exception here would block that use case entirely.

After `weights_summary` is deprecated, we can definitely strengthen this check

pytorch_lightning/trainer/connectors/callback_connector.py
ananthsub: I wanted to use a different warning message as `weights_summary=None` -> `enable_model_summary=False` whereas `weights_summary is not None` -> `callbacks=[ModelSummary(...)]`

pytorch_lightning/trainer/connectors/callback_connector.py
ananthsub: in this file it looks like the spaces are all prepended to the following line besides the gradient accumulation error message

rohitgr7: all the new ones are actually following this one (spaces appended at the end)

docs/source/common/trainer.rst
```diff
@@ -1611,6 +1611,23 @@ Options: 'full', 'top', None.
     # don't print a summary
     trainer = Trainer(weights_summary=None)
 
+.. warning:: `weights_summary` is deprecated in v1.5 and will be removed in v1.7. Please pass :class:`~pytorch_lightning.callbacks.model_summary.ModelSummary`
```
rohitgr7, line 1614: should we put this at the top of this section?

docs/source/common/trainer.rst
```diff
@@ -1611,6 +1611,23 @@ Options: 'full', 'top', None.
...
+
+Whether to enable or disable the model summarization. Defaults to True.
+
+.. testcode::
+
+    # default used by the Trainer
+    trainer = Trainer(enable_model_summary=True)
+
+    # disable summarization
+    trainer = Trainer(enable_model_summary=False)
```
rohitgr7, line 1630: shall we have a small example with `ModelSummary` callback here?

rohitgr7 (approved):
LG!. Just some minor comments.

CHANGELOG.md
```diff
@@ -296,6 +296,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated Accelerator collective API `barrier`, `broadcast`, and `all_gather`, call `TrainingTypePlugin` collective API directly ([#9677](https://github.com/PyTorchLightning/pytorch-lightning/pull/9677))
 
 
+- Deprecated passing `weights_summary` to the `Trainer` constructor in favor of adding the `ModelSummary` callback with `max_depth` directly to the list of callbacks ([#9616](https://github.com/PyTorchLightning/pytorch-lightning/pull/9616))
+
+
```
kaushikb11, line 301: Would need to update the Changelog about adding the `enable_model_summary` flag

ananthsub:
> Do we need the `enable_` prefix? I think `model_summary` is straightforward as it is a boolean flag.

This also came up in these places:
https://github.com/PyTorchLightning/pytorch-lightning/pull/9616#issuecomment-924387620
https://github.com/PyTorchLightning/pytorch-lightning/pull/9664#issuecomment-927314218
https://github.com/PyTorchLightning/pytorch-lightning/pull/9754#issuecomment-930257778

Options so far:

1. Use enable_ as a prefix
- Pros:
  - consistent naming sceheme: users can scan through the trainer flags and see what features are enabled by default by the Trainer
  - `enable_` as a a prefix indicates these are booleans
- Cons:
  - More verbose
2. Directly use the name. e.g. `model_summary`, `progress_bar`, `checkpointing`
- Pros:
  - Less verbose
- Cons:
  - May have confusion with the callback implementation (e.g. `model_summary` as a flag: does it take the `ModelSummary` callback? or a boolean? would `trainer.model_summary` be the callback instance, or the boolean flag passed into the trainer?
3. Other options?

@PyTorchLightning/core-contributors

daniellepintz:
I am strongly in favor of option 1. Another reason is for consistency, i.e. consider these Trainer arguments:
* `profiler` of type `BaseProfiler` 
* `accelerator` of type `Accelerator`
* `logger` of type `LightningLoggerBase`

if I were reading these arguments and then saw one named `model_summary` I would assume it to be an instance of a class as well. 

IMO boolean arguments should be named with a verb as a prefix, such as `enable_progress_bar`, `reload_dataloaders_every_epoch`, `terminate_on_nan`, `move_metrics_to_cpu`

daniellepintz:
Question: under what circumstances would users want to disable the model summary? Is it critical we have this flag?

docs/source/common/debugging.rst
```diff
@@ -95,11 +95,14 @@ Print a summary of your LightningModule
 ---------------------------------------
 Whenever the ``.fit()`` function gets called, the Trainer will print the weights summary for the LightningModule.
 By default it only prints the top-level modules. If you want to show all submodules in your network, use the
-`'full'` option:
+`'max_depth'` option:
```
carmocca, line 98:
```suggestion
``max_depth`` option:
```

carmocca (approved):

carmocca:
> under what circumstances would users want to disable the model summary? Is it critical we have this flag?

Maybe the summary is not correct for your model, maybe you are calling `fit` multiple times and you don't want it to appear repeatedly...

daniellepintz:
> Maybe the summary is not correct for your model, maybe you are calling fit multiple times and you don't want it to appear repeatedly...

Hmm, if the summary is not correct that sounds like a bug, and we should operate under the assumption that we have no bugs.  The other case seems pretty rare, but I don't have a lot of background knowledge here so don't feel strongly

williamFalcon:
@daniellepintz every trainer argument was added because there was overwhelming community need/request for it. our bar for adding any args for the lightning trainer is very high and arguments are added only in extreme circumstances where the community overwhelmingly has an issue or needs something.

daniellepintz:
@williamFalcon makes sense, I agree user experience is very important and didn’t mean to suggest otherwise!

Issue #9208, opened 2021-08-30
Dataclass ClassVars are not correctly handled in apply_to_collection


a-gardner1:
## 🐛 Bug

When including a dataclass in part of a batch that contains a `ClassVar` field, the `ClassVar` field is passed as a keyword argument to the dataclass instance's `__init__` function, resulting in a `TypeError`.
This appears to be due to the use of `__dataclass_fields__` versus `dataclasses.fields` introduced in a [relatively recent PR that added support for dataclasses to `apply_to_collection`](https://github.com/PyTorchLightning/pytorch-lightning/pull/7935/files#diff-04e358a842de2e45c9ccde33d2feb59c37c023234906008e156ae7c741b671d4R121).
The `__dataclass_fields__` attribute does not distinguish between pseudo-fields `ClassVar`/`InitVar` and actual fields, whereas [`dataclasses.fields`](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields) does.

### To Reproduce
https://colab.research.google.com/drive/1Ha5BsgV29buPZwn04mohfkOC4bycm7jN?usp=sharing

### Expected behavior

The model should load data and train without error.

### Environment

* CUDA:
	- GPU:
		- Tesla T4
	- available:         True
	- version:           10.2
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     False
	- pyTorch_version:   1.9.0+cu102
	- pytorch-lightning: 1.4.4
	- tqdm:              4.62.0
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.7.11
	- version:           #1 SMP Sat Jun 5 09:50:34 PDT 2021

carmocca:
Since you are already familiar with what's the problem, do you want to open a PR with the fix?

Remember to include a case for it in `tests/utilities/test_apply_func.py`

a-gardner1:
Sure, I'll try to get to it in the next week or so.

jinyoung-lim:
I could pick this up if @a-gardner1 does not.

PR #9702, opened 2021-09-25
Fix support for dataclasses with ClassVar/InitVar in `apply_to_collection`

a-gardner1:
## What does this PR do?

Adds/fixes support for ClassVar and InitVar fields in dataclasses to apply_to_collection

Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/9208

None

## Before submitting

## PR review

## Did you have fun?


tests/utilities/test_apply_func.py
```diff
@@ -115,24 +150,54 @@ class ModelExample:
...
+        assert dataclasses.is_dataclass(actual) and not isinstance(
+            actual, type
+        ), "Reduction of a dataclass should result in a dataclass"
+        assert dataclasses.is_dataclass(actual.feature) and not isinstance(
+            actual.feature, type
+        ), "Reduction of a nested dataclass should result in a nested dataclass"
+        assert (
+            actual.example_ids == expected.example_ids
+        ), "Reduction of a nested dataclass did not yield the desired result"
+        assert torch.allclose(
```
awaelchli, line 164: these should equal, not just "close" :)

tests/utilities/test_apply_func.py
```diff
@@ -115,24 +150,54 @@ class ModelExample:
...
+    assert dataclasses.is_dataclass(reduced["l"]) and not isinstance(
+        reduced["l"], type
+    ), "Reduction of a dataclass with class and init-only vars should result in a dataclass"
+    assert torch.equal(
+        WithClassAndInitVar.class_var, torch.tensor(0)
+    ), "Reduction of a dataclass with class and init-only vars should not change the class var"
+    try:
+        _assert_nested_dataclass_reduction(reduced["l"].dummy, expected_result["l"].dummy)
+    except AssertionError:
+        raise AssertionError("Reduction of a dataclass with class and init-only vars did not yield the desired result")
```
awaelchli, line 200:
```suggestion
    _assert_nested_dataclass_reduction(reduced["l"].dummy, expected_result["l"].dummy)
```
what's the point of that? all the messages there get discarded for this?

tests/utilities/test_apply_func.py
```diff
@@ -115,24 +150,54 @@ class ModelExample:
...
+        ), "Reduction of a nested dataclass did not yield the desired result"
+        assert torch.allclose(
+            actual.label, expected.label
+        ), "Reduction of a nested dataclass did not yield the desired result"
+        assert torch.allclose(
+            actual.feature.input_ids, expected.feature.input_ids
+        ), "Reduction of a nested dataclass did not yield the desired result"
+        assert np.allclose(
+            actual.feature.segment_ids, expected.feature.segment_ids
+        ), "Reduction of a nested dataclass did not yield the desired result"
```
awaelchli, line 172: what is the point of these messages? isn't it obvious that something did not yield the desired result when this assertion fails? the function is called `_assert_nested_dataclass_reduction` which should already give a clear idea that we are dealing with nested dataclasses here, and the message is just repeated. it does not add to the readability of the test.

awaelchli (approved):

tests/utilities/test_apply_func.py
a-gardner1: Since these tensors were float like all of the rest, I interpreted this comment to mean that all applications of `allclose` should be replaced with `equal` and have done so throughout the test.

tests/utilities/test_apply_func.py
a-gardner1: The main intention was to highlight that the test failed for a dataclass with class and init vars, presuming that if it had gotten this far it must have succeeded for those dataclasses without.
I introduced the `_assert_nested_dataclass_reduction` to avoid duplicating the preexisting tests.

I agree it was a bit awkward and have refactored the dataclass testing mechanisms to minimize duplication and encapsulate dataclass comparison in `__eq__` methods.

tests/utilities/test_apply_func.py
a-gardner1: Agreed. These assertions preexisted this PR; I just wrapped them in `_assert_nested_dataclass_reduction`. The latest version cuts down on the message duplication.

pytorch_lightning/utilities/apply_func.py
```diff
@@ -117,10 +117,10 @@ def apply_to_collection(
         return elem_type(*out) if is_namedtuple else elem_type(out)
 
     if _is_dataclass_instance(data):
-        out_dict = {}
-        for field in data.__dataclass_fields__:
+        result = copy(data)
```
carmocca, line 120: Could the copy be avoided?

tests/utilities/test_apply_func.py
```diff
@@ -31,12 +32,80 @@ class Feature:
         input_ids: torch.Tensor
         segment_ids: np.ndarray
 
+        def __eq__(self, o: object) -> bool:
```
carmocca, line 35: noob q: why did you need to implement all the `__eq__` methods?

a-gardner1: They are not strictly necessary, but they avoid repeatedly writing the same or slightly varied checks for each occurrence of one of the dataclasses in the test. A single `==` works for each assertion.

I'll also note that dataclasses are automatically provided a default `__eq__`, but the defaults won't work here since `torch.Tensor` overrides `__eq__` to return tensors instead of bool.

pytorch_lightning/utilities/apply_func.py
a-gardner1: Probably, although it might reduce the readability of the code by using lower-level operations. I think `elem_type.__new__()` could be used to instantiate an instance of the same dataclass, but I'd have to test that. 

This is at least a shallow copy, so it is pretty lightweight. Let me know what your preference is.

pytorch_lightning/utilities/apply_func.py
```diff
@@ -129,8 +129,9 @@ def apply_to_collection(
                 **kwargs,
             )
             if include_none or v is not None:
-                out_dict[field] = v
-        return elem_type(**out_dict)
+                setattr(result, field.name, v)
+            # else retain old field value
```
a-gardner1, line 133: What should the expected behavior be here?

In other branches of `apply_to_collection`, `include_none=False` leads to the `None`-mapped item being dropped from the collection. I think it would be bad to drop a field from a dataclass, but setting it to `None` might be acceptable (letting it be user error if this violates some invariant for their dataclass).

Alternatively, the original value is retained as it is currently (as though the function was not applied to it).
At the moment, the function will retain the old field value such that `id(getattr(data, field.name)) == id(getattr(result, field.name))`, which seems problematic to me and could lead to unexpected mutability.
I think the proper fix would be to `deepcopy` the field value such that `id(getattr(data, field.name)) != id(getattr(result, field.name))` but `getattr(data, field.name) == getattr(result, field.name)`.

Let me know your preference.

tests/utilities/test_apply_func.py
awaelchli: I think your additions to the test should assert equality. There is no reason to assume the value is changed by any operations in `apply_to_collection`.

awaelchli (approved):

CHANGELOG.md
```diff
@@ -368,6 +368,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed an issue where class or init-only variables of dataclasses were passed to the dataclass constructor in `utilities.apply_to_collection` ([#9208](https://github.com/PyTorchLightning/pytorch-lightning/issues/9208))
```
awaelchli, line 371:
```suggestion
- Fixed an issue where class or init-only variables of dataclasses were passed to the dataclass constructor in `utilities.apply_to_collection` ([#9702](https://github.com/PyTorchLightning/pytorch-lightning/issues/9702))
```

tchaton:
Dear @a-gardner1,

Would you mind resolving the conflicts ?

Best,
T.C

pytorch_lightning/utilities/apply_func.py
a-gardner1: Cool. I didn't realize a `deepcopy` was acceptable. Not all tests run locally for me, but I think the latest version resolves the problem.

carmocca:
Approved the CI to run - you can ping me on Slack if you need it again. It's no longer necessary after the first contribution.

a-gardner1:
I cannot reproduce the failed test. To the best of my knowledge, it does not use anything that this PR has modified.

Local environment vs failed CI environment:
* CentOS 8.2.2004 vs Ubuntu 18.0.4, 
* Python 3.9.8 vs Python 3.9.7, 
* torch 1.10.0 vs torch 1.10.0

pytorch_lightning/utilities/apply_func.py
```diff
@@ -119,21 +119,33 @@ def apply_to_collection(
...
-                    wrong_dtype=wrong_dtype,
-                    include_none=include_none,
-                    **kwargs,
-                )
-                if include_none or v is not None:
-                    out_dict[field.name] = v
-        return elem_type(**out_dict)
+            field_value = getattr(data, field.name)
+            fields[field.name] = (field_value, field.init)
+            memo[id(field_value)] = field_value
```
carmocca, line 130: nit: everything passes without the memo. do you think you could write a test to ensure it works properly?

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -74,6 +74,11 @@ def _generate_sync_fn(self) -> None:
         # ignore typing, bad support for `partial`: mypy/issues/1484
         self._fn: Callable = partial(fn, reduce_op=self.op, group=self.group)  # type: ignore [arg-type]
 
+    def __deepcopy__(self, memo: dict) -> "_Sync":
```
carmocca, line 77: Can you add a comment about why this is necessary?

carmocca (approved):

justusschock (approved):

pytorch_lightning/trainer/connectors/logger_connector/result.py
a-gardner1: In trying to find a satisfactory explanation, I discovered this change was just patching a symptom. The setter for `_Metadata.sync` alters the reduce operation for the `_Sync` object but does not regenerate its `_fn` attribute ([the problematic setter invocation](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/logger_connector/result.py#L429)). This (presumed) bug was masked in previous versions by [the call to `self.register_key`](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/logger_connector/result.py#L433), which implicitly created a new `_Sync` object and regenerated `_fn` due to the way dataclasses were handled in `apply_to_collection`. Changing to a (deep)copy (forced by the inability to infer default values for `InitVar` constructor arguments) meant that this regeneration no longer occurred and that the `partial`ly applied `_fn` retained a `None` reduce op. 

I will be pushing a more targeted fix that eschews the custom `__deepcopy__` shortly.

pytorch_lightning/utilities/apply_func.py
a-gardner1: The memo is meant to be an internal optimization that has no effect on the external API or observable functionality. However, there was one way in which this internal detail leaked: through non-init fields, which were deepcopied if the `memo` was omitted and passed through unchanged otherwise. The latest version rectifies this by ensuring that original non-init field values are explicitly set in the result.

With this change, I do not believe there is any reasonable test that could expose the usage of the memo.

PR #9704, opened 2021-09-25
Fix `lr_find` to generate same results on multiple calls

### LINKED ISSUE #9672: lr_find not resetting Trainer model
## 🐛 Bug

If `tuner.lr_find` is called multiple times on the same trainer the observed behavior is as if the tuner didn't restore the parameters upon finishing, suggesting that the trainer is maintaining a reference to the modified parameters instead of to the restored ones. 

This may have been the cause of issue #6675. 

This is pure speculation but this may as well be an issue with `tuner.scale_batch_size` leading to an apparent memory leak (see #9625)

### To Reproduce

Simple example:
```python
#init model
#init optimizer
#init trainer
for n in [100, 200, 300]:
    #trainer should be instantiated here for correct behaviour
    lr_finder = trainer.tuner.lr_find(model, ... , num_training=n)
    # do something with lr_finder
```
[BoringModel](https://gist.github.com/grudloff/a302e5778d246926d84a1a5b333d84c9)

Here 2 runs of lr_find are run, once with no reintantiation of the trainer in test_bad() and once with reinstantiation of the trainer in test_good(). Here are the curves for both cases:
image
image
As can be seen, when there is no reinstantiation it seems the second run starts from the previous run's end, while on the second case both curves are the same.

@tchaton

rohitgr7:
## What does this PR do?

Fixes #9672

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/tests/tuner/test_lr_finder.py
+++ b/tests/tuner/test_lr_finder.py
@@ -293,5 +293,8 @@ def test_multiple_lr_find_calls_gives_same_results(tmpdir):
     trainer = Trainer(default_root_dir=tmpdir, max_epochs=2)
     all_res = [trainer.tuner.lr_find(model).results for _ in range(3)]
 
-    assert all(len(res["lr"]) > 10 for res in all_res)
-    assert all(all_res[0][k] == curr_lr_finder[k] for curr_lr_finder in all_res[1:] for k in all_res[0].keys())
+    assert all(
+        all_res[0][k] == curr_lr_finder[k] and len(curr_lr_finder[k]) > 10
+        for curr_lr_finder in all_res[1:]
+        for k in all_res[0].keys()
+    )
--- a/tests/tuner/test_lr_finder.py
+++ b/tests/tuner/test_lr_finder.py
@@ -288,11 +288,10 @@ def test_lr_finder_ends_before_num_training(tmpdir):
 
 def test_multiple_lr_find_calls_gives_same_results(tmpdir):
     """Tests that lr_finder gives same results if called multiple times."""
-    dm = ClassifDataModule()
-    model = ClassificationModel()
+    model = BoringModel()
 
     trainer = Trainer(default_root_dir=tmpdir, max_epochs=2)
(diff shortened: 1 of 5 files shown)


tchaton (approved):

SkafteNicki (approved):

tests/tuner/test_lr_finder.py
```diff
@@ -283,3 +284,15 @@ def training_step_end(self, outputs):
...
+def test_multiple_lr_find_calls_gives_same_results(tmpdir):
+    """Tests that lr_finder gives same results if called multiple times."""
+    dm = ClassifDataModule()
+    model = ClassificationModel()
+
+    trainer = Trainer(default_root_dir=tmpdir, max_epochs=2)
+    all_res = [trainer.tuner.lr_find(model, datamodule=dm).results for _ in range(3)]
+
+    for k in all_res[0].keys():
+        assert all(all_res[0][k] == curr_lr_finder[k] for curr_lr_finder in all_res[1:])
```
awaelchli, line 298:
```suggestion
    model = BoringModel()

    trainer = Trainer(default_root_dir=tmpdir, max_epochs=2)
    all_res = [trainer.tuner.lr_find(model).results for _ in range(3)]

    for k in all_res[0].keys():
        assert all(all_res[0][k] == curr_lr_finder[k] for curr_lr_finder in all_res[1:])
```

there is no need to use a complex model, we can use the boring model. 
test fails on master.

awaelchli (approved):

tests/tuner/test_lr_finder.py
rohitgr7: sometimes `lr_find` fails when using `BoringModel` since it doesn't converge and it just logs the exception. So I prefer to use a model with lr_finder tests that at least converge to confirm that it works as expected.

awaelchli: don't we want to set a seed then? I  mean just initially for the test

PR #9721, opened 2021-09-27
[Refactor] Simplify data loading logic around replacing sampler to prevent confusion

### LINKED ISSUE #9655: Sampler is being replaced even with trainer.replace_sampler_ddp=False

tchaton:
## What does this PR do?

Fixes #9655

This PR remove `trainer.replace_sampler` as the function leaked to do more with samplers, datasets wrapping which might have un-wanted effects on the user side.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/data_loading.py
+++ b/pytorch_lightning/trainer/data_loading.py
@@ -304,10 +304,9 @@ class TrainerDataLoadingMixin(ABC):
         shuffle: bool,
         overfit_batches: Union[int, float],
         mode: Optional[RunningStage] = None,
-        **distributed_sampler_kwargs,
+        **kwargs: Any,
     ) -> DistributedSampler:
         """This function is used to created the distributed sampler injected within the user DataLoader."""
-        kwargs = distributed_sampler_kwargs
         kwargs["shuffle"] = shuffle and not overfit_batches
         kwargs.setdefault("seed", int(os.getenv("PL_GLOBAL_SEED", 0)))
         cls = UnrepeatedDistributedSampler if mode == RunningStage.PREDICTING else DistributedSampler
--- a/pytorch_lightning/trainer/data_loading.py
+++ b/pytorch_lightning/trainer/data_loading.py
@@ -166,9 +166,9 @@ class TrainerDataLoadingMixin(ABC):
             if not isinstance(dataloader.sampler, (SequentialSampler, RandomSampler)):
                 raise MisconfigurationException(
                     "You seem to have configured a sampler in your DataLoader. This will be replaced "
-                    " by `DistributedSampler` since `_prepare_dataloader_ddp` is True and you are using"
+                    " by `DistributedSampler` since `replace_sampler_ddp` is True and you are using"
                     " distributed training. Either remove the sampler from your DataLoader or set"
-                    " `_prepare_dataloader_ddp`=False if you want to use your custom sampler."
+                    " `replace_sampler_ddp=False` if you want to use your custom sampler."
                 )
             return self._get_distributed_sampler(
                 dataloader, shuffle, mode=mode, overfit_batches=self.overfit_batches, **self.distributed_sampler_kwargs
--- a/pytorch_lightning/trainer/data_loading.py
+++ b/pytorch_lightning/trainer/data_loading.py
@@ -117,11 +117,24 @@ class TrainerDataLoadingMixin(ABC):
         if int(os.environ.get("PL_SEED_WORKERS", 0)) and dataloader.worker_init_fn is None:
             dataloader.worker_init_fn = partial(pl_worker_init_function, rank=self.global_rank)
 
-    def auto_add_sampler(self, dataloader: Any, shuffle: bool, mode: Optional[RunningStage] = None) -> Any:
+    def _requires_distributed_sampler(self, dataloader) -> bool:
+        return (
+            self.accelerator_connector.replace_sampler_ddp
+            and self.accelerator_connector.is_distributed
+            and not isinstance(dataloader.sampler, DistributedSampler)
+            and not has_iterable_dataset(dataloader)
+        )
+
+    def prepare_dataloader(self, dataloader: Any, shuffle: bool, mode: Optional[RunningStage] = None) -> Any:
+        """This function handles to following functionalities:
+
+        - Injecting a `DistributedDataSampler` into the `DataLoader` if on a distributed environment
+        - Wrapping the datasets and samplers into fault-tolerant components
+        """
         if isinstance(dataloader, CombinedLoader):
-            # apply `auto_add_sampler` on all the collection of loaders
+            # apply `prepare_dataloader` on all the collection of loaders
             dataloader.loaders = apply_to_collection(
-                dataloader.loaders, DataLoader, self.auto_add_sampler, shuffle, mode=mode
+                dataloader.loaders, DataLoader, self.prepare_dataloader, shuffle, mode=mode
             )
             return dataloader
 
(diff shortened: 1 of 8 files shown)


pytorch_lightning/trainer/data_loading.py
```diff
@@ -117,39 +117,121 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
+
+    @staticmethod
+    def _prepare_dataloader(dataloader: DataLoader, sampler, mode: Optional[RunningStage] = None) -> DataLoader:
+        dl_kwargs = TrainerDataLoadingMixin._get_dataloader_init_kwargs(dataloader, sampler, mode=mode)
+        dl_cls = type(dataloader)
+        dataloader = dl_cls(**dl_kwargs)
+        return dataloader
+
+    def _determine_sampler(self, dataloader: DataLoader, shuffle: bool, mode: Optional[RunningStage] = None) -> Sampler:
+        if self._requires_distributed_sampler(dataloader):
```
ananthsub, line 165: why is this check repeated twice?

ananthsub (commented):
@tchaton is there a description of what the dataloader processing pipeline should look like, how it varies for across the different trainer functions (given prediction has a special case here). That could be where some of the confusion around this logic creeps in now.

Is there any expected relationship to the [training type plugin's process_dataloader ](https://github.com/PyTorchLightning/pytorch-lightning/blob/131176b9f5e544a1133240a46dca1960308b7849/pytorch_lightning/plugins/training_type/training_type_plugin.py#L221-L227)? Given the two have similar names, and there's `trainer.prepare_dataloader` and `trainer.training_type_plugin.process_dataloader` one might expect the two to go hand in hand (but they don't)


Another risk with the trainer mixin design is that these are all technically part of the trainer's public API. so removing `auto_add_sampler` is a breaking API change. the odds of someone depending on this outside of the trainer entry functions is low, but how else can we mitigate this? Some ideas:
- Short term: Move all staticmethods out of the mixin into utilities files, and ensure there's testing for them
- Short term: make all methods in the dataloading mixin protected so nothing is publicly exposed going forward

cc @ninginthecloud

pytorch_lightning/trainer/data_loading.py
```diff
@@ -117,39 +117,121 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
             return dataloader
 
+        # the DataLoader should be re-created only if we need to inject
+        # the fault tolerant components or the distributed sampler or we need to track indices for predictions.
         if (
-            self.accelerator_connector.replace_sampler_ddp
-            and self.accelerator_connector.is_distributed
-            and not isinstance(dataloader.sampler, DistributedSampler)
-            and not has_iterable_dataset(dataloader)
+            _fault_tolerant_training()
```
awaelchli, line 148: did we have this check before? id does not look like it?

awaelchli (approved):
In a refactor, we are normally not allowed to make changes to tests. Do the tests pass without the modifications/additions?

carmocca (approved):
LGTM other than open comments 👍

pytorch_lightning/trainer/data_loading.py
```diff
@@ -117,44 +117,69 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
-            # use current sampler
-            sampler = dataloader.sampler
+            return self._get_distributed_sampler(
+                dataloader, shuffle, mode=mode, overfit_batches=self.overfit_batches, **self.distributed_sampler_kwargs
+            )
 
-        dataloader = self.replace_sampler(dataloader, sampler, mode=mode)
-
-        return dataloader
+        return dataloader.sampler
```
ninginthecloud, line 170: `_determine_sampler` basically helps to generate distributed_sampler, so there's no need to return `dataloader.sampler`. If the dataloader does not require distributed_sampler, we should not update dataloader again.

pytorch_lightning/trainer/data_loading.py
```diff
@@ -298,13 +328,13 @@ def reset_train_dataloader(self, model: Optional["pl.LightningModule"] = None) -
                     "You requested to overfit but enabled training dataloader shuffling."
                     " We are turning off the training dataloader shuffling for you."
                 )
-                self.train_dataloader = self.replace_sampler(
+                self.train_dataloader = self._prepare_dataloader(
```
ninginthecloud, line 331: what if `self.train_dataloader` is not strict Dataloader, but a list of Dataloader? see the following [line351](https://github.com/PyTorchLightning/pytorch-lightning/pull/9721/files#diff-15fd4659acd10d0716002ba2dbfd67aca83dc101d44531c2f234f65140e1e39dR351). 
This is unrelated to this PR, but `reset_train_dataloader` has a lot of operations, the orders of them matter. We need to revisit them to see what is the reasonable order.

ninginthecloud (commented):
Thanks for fixing the issue #9655 and refactoring the code, @tchaton ! Overall looks good to me. I have a few minor changes. Additionally, could we also add some docstring for each updated functions, as some function names are not self-explanatory. 
As I'm learning this component, I find there's a lot of logic happening in data_loading.py, I'm curious about how we can sustainably update and maintain this file, as there's a lot of utils functions and they are depending each other, it will be hard to trace. Love to hear your thoughts.

pytorch_lightning/trainer/data_loading.py
tchaton: To improve readability of `prepare_dataloader` which is the exposed utility.

pytorch_lightning/trainer/data_loading.py
tchaton: We didn't have any check. This was done internally. Now it is exposed to the top level for readers to understand when this function is taking effects.

pytorch_lightning/trainer/data_loading.py
```diff
@@ -117,44 +117,69 @@ def auto_add_worker_init_fn(self, dataloader: DataLoader) -> None:
...
+        # the DataLoader should be re-created only if we need to inject
+        # the fault tolerant components or the distributed sampler or we need to track indices for predictions.
         if (
-            self.accelerator_connector.replace_sampler_ddp
-            and self.accelerator_connector.is_distributed
-            and not isinstance(dataloader.sampler, DistributedSampler)
-            and not has_iterable_dataset(dataloader)
+            _fault_tolerant_training()
+            or self._requires_distributed_sampler(dataloader)
+            or mode == RunningStage.PREDICTING
```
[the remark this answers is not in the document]
tchaton: Those conditions are used internally with `_get_dataloader_init_kwargs`. It doesn't change the behaviour, just make it apparent within the `prepare_dataloader` when modification would happen.

pytorch_lightning/trainer/data_loading.py
```diff
@@ -267,17 +292,22 @@ def _get_dataloader_init_kwargs(
         return dl_kwargs
 
     @staticmethod
-    def replace_sampler(dataloader: DataLoader, sampler, mode: Optional[RunningStage] = None) -> DataLoader:
+    def _prepare_dataloader(dataloader: DataLoader, sampler, mode: Optional[RunningStage] = None) -> DataLoader:
```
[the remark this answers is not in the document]
tchaton: Done !

pytorch_lightning/trainer/data_loading.py
tchaton: `_determine_sampler` doesn't generate the distributed sampler. It resolve the right sampler. Changed it to `_resolve_sampler`.

pytorch_lightning/trainer/data_loading.py
tchaton: Good point. Replace to apply_to_collection.

tests/conftest.py
```diff
@@ -66,6 +66,7 @@ def restore_env_variables():
         "RANK",  # set by DeepSpeed
         "POPLAR_ENGINE_OPTIONS",  # set by IPUPlugin
         # set by XLA
+        "TF2_BEHAVIOR",
```
carmocca, line 69: Same comment as https://github.com/PyTorchLightning/pytorch-lightning/pull/9722#discussion_r717529207

It must be coming from `TensorBoard`, a better solution would be to disable the logger for that test

tchaton: It seems coming from multiple tests. Let's address this within another PR.

kaushikb11 (approved):
LGTM!

Issue #9462, opened 2021-09-12
fix flaky `test_simple_profiler_overhead`


awaelchli:
## 🐛 Bug

A flaky profiler test blocks our auto-merge. Bad!

### To Reproduce

Test run `tests/profiler/test_profiler/test_simple_profiler_overhead`

```bash
________________________ test_simple_profiler_overhead _________________________

simple_profiler = <pytorch_lightning.profiler.simple.SimpleProfiler object at 0x7f30f86b8eb8>
n_iter = 5

    def test_simple_profiler_overhead(simple_profiler, n_iter=5):
        """Ensure that the profiler doesn't introduce too much overhead during training."""
        for _ in range(n_iter):
            with simple_profiler.profile("no-op"):
                pass
    
        durations = np.array(simple_profiler.recorded_durations["no-op"])
>       assert all(durations < PROFILER_OVERHEAD_MAX_TOLERANCE)
E       assert False
E        +  where False = all(array([3.19999992e-06, 1.90000003e-06, 1.60000002e-06, 5.33505000e-04,\n       1.90000003e-06]) < 0.0005)

/home/runner/work/pytorch-lightning/pytorch-lightning/tests/profiler/test_profiler.py:91: AssertionError

```
### Expected behavior

Obvious.

### Environment

Our own CI:
pytest (ubuntu-18.04, 3.6, minimal, stable)

carmocca:
I had a look but this is really tricky because testing runtime is inherently flaky.

The only options I see are:
- Remove these tests.
- Mark them to be re-tried on failure (just reduces the flakiness but it's still there)

awaelchli:
Thanks for taking a look. Let's try the second option and see if that helps. If not we can later consider the first option. 
Was reading here about handling flakiness with pytest: https://docs.pytest.org/en/latest/explanation/flaky.html
So looks like we would need to use one of the plugins listed there. 

Also very interesting:

> Plugins to deliberately randomize tests can help expose tests with state problems:
> 
> pytest-random-order
> pytest-randomly

Could be something to try in the future.

carmocca:
Okay. Should be easy enough. I'd use `https://github.com/pytest-dev/pytest-rerunfailures` which seems to be the most popular.

We can also mark this as a "good first issue" if you want.

awaelchli:
yup!

PR #9722, opened 2021-09-27
[Feat] Add auto_restart for fault tolerant training

tchaton:
## What does this PR do?

Once a `signal.SIGUSR1` is being sent to Lightning, the Trainer should exit to the next most reproducible place. The Trainer would be aware of this as `_terminate_gracefully` would be equal to True.

This PR adds the following:
* Logic to the exit the loops using an `ExitGracefullyException` exception once safe places are reached
*  Save a fault tolerant checkpoint
* Automatically reload the fault tolerant checkpoint if `.pl_auto_save.ckpt` files exists.

Part of #9130

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -41,11 +41,18 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
...
     @property
-    def hpc_resume_path(self) -> Optional[str]:
+    def auto_restart_path(self) -> Optional[str]:
         dir_path_hpc = str(self.trainer.weights_save_path)
         max_version = self.max_ckpt_version_in_folder(dir_path_hpc, "hpc_ckpt_")
         if max_version is not None:
             return os.path.join(dir_path_hpc, f"hpc_ckpt_{max_version}.ckpt")
+        if (
+            dir_path_hpc is not None
+            and os.path.isdir(dir_path_hpc)
```
carmocca, line 51: These two conditions should be either removed or moved up. If they resolve to False, the `join` call above will also fail.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -41,11 +41,18 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
...
+            and ".pl_auto_save.ckpt" in os.listdir(dir_path_hpc)
```
carmocca, line 52: side note: we should consider making this a class attribute and perhaps removing the leading period.

tests/conftest.py
```diff
@@ -68,6 +68,7 @@ def restore_env_variables():
         # set by XLA
         "XRT_MESH_SERVICE_ADDRESS",
         "XRT_TORCH_DIST_ROOT",
+        "TF2_BEHAVIOR",
```
carmocca, line 71: How is this related?

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -197,6 +197,9 @@ def on_advance_end(self):
             # progress global step according to grads progress
             self.global_step += 1
 
+        # if fault tolerant is enabled and process has been notified, exit.
+        self.trainer.should_exit_gracefully()
```
carmocca, line 201: So the plan is to add this everywhere we can, right?

Should we do one PR for each occurrence with its test?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: Would do, but in another PR.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: Done !

tests/conftest.py
tchaton: I run one of our test and found this variable variable was leaking. 

If you run `pytest tests/trainer/connectors/test_checkpoint_connector.py::test_hpc_max_ckpt_version`, 

you would get 
```py
ERROR tests/trainer/connectors/test_checkpoint_connector.py::test_hpc_max_ckpt_version - AssertionError: test is leaking environment variable(s): {'TF2_BEHAVIOR'}
```

I guess it gets cleaned out by another test, but fails alone. So I added there. 

@awaelchli What is the intended purpose of this test ?

pytorch_lightning/loops/epoch/training_epoch_loop.py
tchaton: Right now, I am adding it only for batch level (train, val loops), so let me add 2 tests.

tchaton: We would do another PR for epoch end if required.

tchaton: Let's use this PR to PoC it and split it afterward.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -40,9 +40,17 @@ def __init__(self) -> None:
...
+        self.batch_progress = BatchProgress()
         self.dataloader_iter: Optional[Iterator] = None
 
+    @property
+    def has_completed_validation_batch(self) -> bool:
+        return self.batch_progress.current.ready == self.batch_progress.current.completed
+
+    @property
+    def has_completed_dataloader(self) -> bool:
+        return self.has_completed_validation_batch and self.batch_progress.is_last_batch
```
carmocca, line 52: For consistency with the `TrainingEpochLoop`, maybe we should have

```suggestion
    @property
    def _num_val_batches_reached(self) -> bool:
        return self.batch_progress.current.ready == self.trainer.num_val_batches self.batch_progress.is_last_batch
```

pytorch_lightning/loops/fit_loop.py
```diff
@@ -248,9 +248,17 @@ def teardown(self) -> None:
...
+        if not epoch_loop:
+            return state_dict
+
         # TODO: update has_completed to its proper value
-        if self.trainer.train_dataloader is not None:
-            state_dict["dataloader_state_dict"] = self.trainer.train_dataloader.state_dict(has_completed=False)
+        should_skip = epoch_loop.batch_progress.is_last_batch and epoch_loop.has_completed_training_batch
+        if self.trainer.train_dataloader is not None and not should_skip:
+            state_dict["dataloader_state_dict"] = self.trainer.train_dataloader.state_dict(
+                has_completed=epoch_loop.has_completed_training_batch
```
carmocca, line 260: shouldn't this pass `not should_skip`?

tests/loops/test_loops.py
```diff
@@ -492,10 +493,10 @@ def configure_optimizers_multiple(self):
 
     # need to remove these elements for comparison; comparing with `fit_loop.state_dict()` would require the
     # fit loop to have an iterator, which is only available during training
-    checkpoint["loops"]["fit_loop"]["state_dict"]["dataloader_state_dict"] = ANY
-    assert state_dict == checkpoint["loops"]["fit_loop"]
-
```
carmocca, line 497: Why is this moved?

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -52,6 +52,10 @@ def __init__(self, min_steps: int, max_steps: int):
         self._results = ResultCollection(training=True)
         self._epoch_output: Optional[List[List[STEP_OUTPUT]]] = None
 
+    @property
+    def has_completed_training_batch(self) -> bool:
```
carmocca, line 56:
```suggestion
    def _has_completed_training_batch(self) -> bool:
```

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -40,9 +40,17 @@ def __init__(self) -> None:
         self._dl_max_batches: Optional[int] = None
         self._num_dataloaders: Optional[int] = None
         self.outputs: EPOCH_OUTPUT = []
-        self.batch_progress = Progress()
+        self.batch_progress = BatchProgress()
         self.dataloader_iter: Optional[Iterator] = None
 
+    @property
+    def has_completed_validation_batch(self) -> bool:
```
carmocca, line 47:
```suggestion
    def _has_completed_validation_batch(self) -> bool:
```

tests/conftest.py
carmocca: It does pass for me, likely an issue in your environment.

The purpose is to track which env variables are leaking. With the intent to reduce the number and avoiding introducing new ones

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
tchaton: Done !

pytorch_lightning/loops/fit_loop.py
tchaton: replace by `should_save_dataloader_dict`

tests/loops/test_loops.py
tchaton: Because `dataloader_state_dict` is optional. And as you override it and reload it, it would break as  dataloader_state_dict isn't None by ANY.

pytorch_lightning/loops/epoch/training_epoch_loop.py
tchaton: Done !

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
tchaton: Done 1

tests/conftest.py
tchaton: Yes, let me open an issue to track this later.

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -70,31 +70,37 @@ def worker_id(self) -> int:
 
     def __iter__(self) -> Iterator[Any]:
         self._current_iteration = 0
+        self.i = 0
+        self.sampler_iter = iter(self._sampler)
+        return self
+
+    def __next__(self):
```
tchaton, line 77: Note for the reviewer: This is used to simplify state management and don't change the FastForwardSampler.

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -70,31 +70,37 @@ def worker_id(self) -> int:
 
     def __iter__(self) -> Iterator[Any]:
         self._current_iteration = 0
+        self.i = 0
+        self.sampler_iter = iter(self._sampler)
```
awaelchli, line 74: these attributes are undefined in the constructor. `i` is not an expressive name for a public variable

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -70,31 +70,37 @@ def worker_id(self) -> int:
...
+        self.i += 1
+        has_raised = False
+        try:
+            return next(self.sampler_iter)
+        except StopIteration:
+            has_raised = True
 
         self._current_iteration = 0
         self._cached_state_dict = None
         self.restarting = False
```
awaelchli, line 101: this code is meant to run in any case, so you could simply put it in the finally block and don't need the bool to re-raise

pytorch_lightning/utilities/exceptions.py
```diff
@@ -19,3 +19,7 @@ class MisconfigurationException(Exception):
 
 class DeadlockDetectedException(Exception):
     """Exception used when a deadlock has been detected and processes are being killed."""
+
+
+class ExitGracefullyException(Exception):
+    """Exception used when a SIGUSR1 was sent the processes and received by Lightning."""
```
awaelchli, line 25:
```suggestion
    """Exception used when a SIGUSR1 was sent to the processes and received by Lightning."""
```
the sentence is incomplete

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -241,3 +245,10 @@ def _should_track_batch_outputs_for_epoch_end(self) -> bool:
         if self.trainer.testing:
             return is_overridden("test_epoch_end", model)
         return is_overridden("validation_epoch_end", model)
+
+    @property
+    def _has_completed_validation_batch(self) -> bool:
+        return self.batch_progress.current.ready == self.batch_progress.current.completed
+
+    def _num_val_batches_reached(self) -> bool:
+        return self._has_completed_validation_batch and self.batch_progress.is_last_batch
```
awaelchli, line 254: one is a property, the other is a method. plus, they are protected yet never used by the class here. this makes little sense. imo, either

1) make them both public properties
2) move them to the loop that actually uses this information, then they can stay protected there

my preference is 2)

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -48,10 +48,15 @@ def __init__(self, min_steps: int, max_steps: int):
 
         self.batch_loop: Optional[TrainingBatchLoop] = None
         self.val_loop: Optional["loops.EvaluationLoop"] = None
+        self.dataloader_iter: Optional[Iterator] = None
 
         self._results = ResultCollection(training=True)
         self._epoch_output: Optional[List[List[STEP_OUTPUT]]] = None
 
+    @property
+    def _has_completed_training_batch(self) -> bool:
```
awaelchli, line 61: same comment as above

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -41,11 +41,16 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
         self._loaded_checkpoint: Dict[str, Any] = {}
 
     @property
-    def hpc_resume_path(self) -> Optional[str]:
+    def auto_restart_path(self) -> Optional[str]:
         dir_path_hpc = str(self.trainer.weights_save_path)
+        if dir_path_hpc is None or not os.path.isdir(dir_path_hpc):
+            return
```
awaelchli, line 47:
```suggestion
            return None
```
in theory, mypy should complain here

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -41,11 +41,16 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
         self._loaded_checkpoint: Dict[str, Any] = {}
 
     @property
-    def hpc_resume_path(self) -> Optional[str]:
+    def auto_restart_path(self) -> Optional[str]:
```
awaelchli, line 44: don't we need to rename this in a different PR?

tests/loops/test_loops.py
```diff
@@ -756,7 +757,7 @@ def test_fit_loop_reset(tmpdir):
     ["train_datasets", "val_datasets"],
     [([RandomDataset], [RandomDataset]), ([RandomDataset], [RandomDataset, RandomDataset])],
 )
-@pytest.mark.parametrize("val_check_interval", [0.5, 1.0])
+@pytest.mark.parametrize("val_check_interval", [0.5])
```
awaelchli, line 760: is 1.0 problematic?

tests/utilities/test_auto_restart.py
```diff
@@ -1060,3 +1062,142 @@ def run(should_fail, resume):
...
+        self.failure_on_step = failure_on_step
+        self.failure_on_training = failure_on_training
+        self.on_last_batch = on_last_batch
+        self.seen_train_batches = []
+
+    def _signal(self, message):
+        if self.should_signal:
+            print(message)
+            os.kill(os.getpid(), signal.SIGUSR1)
+            # small optimization to skip as soon as async signal as being triggered.
```
awaelchli, line 1080:
```suggestion
            # small optimization to skip as soon as async signal has being triggered.
```

tests/utilities/test_auto_restart.py
```diff
@@ -1060,3 +1062,142 @@ def run(should_fail, resume):
...
+
+
+@pytest.mark.parametrize("on_last_batch", [False, True])
+@pytest.mark.parametrize("val_check_interval", [0.5, 1.0])
+@pytest.mark.parametrize("failure_on_training", [False, True])
+@pytest.mark.parametrize("failure_on_step", [False, True])
+@mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": "1"})
+@RunIf(min_torch="1.7.0", skip_windows=True)
+def test_auto_restart_under_signal(on_last_batch, val_check_interval, failure_on_training, failure_on_step, tmpdir):
+
```
awaelchli, line 1152: the docstring is not attached to the function! 
black malfunction?

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -201,6 +211,12 @@ def on_advance_end(self):
             # progress global step according to grads progress
             self.global_step += 1
 
+        # we check this as we should have enough time to finish ``on_run_end``.
+        # TODO @(tchaton) Verify this assumption is True.
```
awaelchli, line 215:
```suggestion
        # TODO: @tchaton verify this assumption is True.
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2004,6 +2004,10 @@ def _results(self) -> Optional[ResultCollection]:
         if active_loop is not None:
             return active_loop._results
 
+    def should_exit_gracefully(self, status: str) -> None:
```
awaelchli, line 2007: any alternative names considered here? just `exit_gracefully`? the public api on the trainer also needs a docstring.

tests/conftest.py
awaelchli: > @awaelchli What is the intended purpose of this test ?

You were asking about test_hpc_max_ckpt_version?
It does what it says in the docstring and it does not set any env variables. I would have guessed maybe you have a newer tensorboard version and something gets triggered there. 
Also, I don't see how this env variable can be set by XLA if the test never runs on TPU

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

pytorch_lightning/utilities/auto_restart.py
tchaton: Good catch !

pytorch_lightning/utilities/exceptions.py
tchaton: Done !

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: Done !

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: Reverted back.

tests/loops/test_loops.py
tchaton: Good catch ! No, 0.5 does. Forgot to revert it back.

tests/utilities/test_auto_restart.py
tchaton: Done !

tests/utilities/test_auto_restart.py
tchaton: Done !

pytorch_lightning/loops/epoch/training_epoch_loop.py
tchaton: Done !

pytorch_lightning/trainer/trainer.py
tchaton: Changed to `_exit_gracefully_on_signal`.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
tchaton: @carmocca Any thoughts there ? Should we move them to the TrainingEpochLoop ?

pytorch_lightning/utilities/auto_restart.py
tchaton: I added it back as it can hep with wrong usage from the FastForwardSampler. 

The test `test_fast_forward_on_sequential_sampler` is an example where the FastForwardSampler isn't fully consumed before calling iter again. I think it is safer to always reset entirely when calling iter

awaelchli: but can the code not be simplified by

```
try:
    return next(self.sampler_iter)
except StopIteration:
    raise StopIteration
finally:
    self._current_iteration = 0
    self._cached_state_dict = None
    self.restarting = False
```
finally is guaranteed to run, so no need to jump using has_raised

awaelchli: no, I'm wrong. this won't work. no simplification possible because its in the `next` method, not `iter` with yield

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -43,9 +43,14 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
...
     def hpc_resume_path(self) -> Optional[str]:
         dir_path_hpc = str(self.trainer.weights_save_path)
+        if dir_path_hpc is None or not os.path.isdir(dir_path_hpc):
+            return None
         max_version = self.max_ckpt_version_in_folder(dir_path_hpc, "hpc_ckpt_")
         if max_version is not None:
             return os.path.join(dir_path_hpc, f"hpc_ckpt_{max_version}.ckpt")
+        if ".pl_auto_save.ckpt" in os.listdir(dir_path_hpc):
+            auto_save_checkpoint = os.path.join(dir_path_hpc, ".pl_auto_save.ckpt")
+            return auto_save_checkpoint
```
ananthsub, line 53: this only takes effect for HPC resume path? why isn't it part of the general `resume_from_checkpoint` flow? 

isn't fault tolerance a superset of the HPC support currently in Lightning? Why doesn't this supersede that so we can deprecate the explicit hpc hooks?

pytorch_lightning/trainer/trainer.py
```diff
@@ -2004,6 +2004,10 @@ def _results(self) -> Optional[ResultCollection]:
         if active_loop is not None:
             return active_loop._results
 
+    def _exit_gracefully_on_signal(self, status: str) -> None:
+        if _fault_tolerant_training() and self._terminate_gracefully:
+            raise ExitGracefullyException(f"Exiting on {status}")
```
ananthsub, line 2009: is this expected to be thrown during any part of the trainer execution? 
what's required for the error handling here? https://github.com/PyTorchLightning/pytorch-lightning/blob/019e69ffd2d4e1f86f9002901cf217f85660b4b1/pytorch_lightning/trainer/trainer.py#L558-L567

ananthsub (commented):
@tchaton this design is currently missing from https://docs.google.com/document/d/1pvhyNg17pm7-ECMTpgpZiJf6sbypeuaB8QsSP79n6k0/edit#

which makes it hard to see how this feature fits into the overall fault tolerance flow

pytorch_lightning/utilities/exceptions.py
```diff
@@ -19,3 +19,8 @@ class MisconfigurationException(Exception):
 
 class DeadlockDetectedException(Exception):
     """Exception used when a deadlock has been detected and processes are being killed."""
+
+
+class ExitGracefullyException(Exception):
+    """Exception used when a ``signal.SIGUSR1` is sent to the process which triggers Lightning to create fault
+    tolerant checkpoint once reached the next reproducible part of the codebase and then exits."""
```
ananthsub, line 26: how is "next reproducible part" defined? this is very vague to me

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: This isn't part of the `resume_from_checkpoint`, as users shouldn't have to modify their Trainer for their training for fault tolerant to resume. This is meant to enable auto_restart on preemptive failure.

Yes, the depreciation and modification would be done in the following PR.

pytorch_lightning/trainer/trainer.py
tchaton: Nothing to be changed on the exception handling part. 
It won't be thrown anywhere, just in safe places where all components have been properly exectuted.
It would be raised only if `self._terminate_gracefully` has been set to True through a signal notification.

pytorch_lightning/utilities/exceptions.py
tchaton: `next reproducible part` isn't fully baked yet, but it means optimization happened (optimizer.step, schedulers), logging, etc...

Here is the current mechanism to exit from the test.

```
    if failure_on_step:
        if on_last_batch:
            if failure_on_training:
                # Breaking on first validation batch.
                # This is done to capture the random state of the validation dataloader.
                status = "EvaluationEpochLoop:advance"
            else:
                # when breaking on last batch of validation, we should exist on `run_end` val_check_interval == 1.0
                status = (
                    "TrainingEpochLoop:on_run_end" if val_check_interval == 1.0 else "TrainingEpochLoop:on_advance_end"
                )
        else:
            status = "TrainingEpochLoop:on_advance_end" if failure_on_training else "EvaluationEpochLoop:advance"
    else:
        if val_check_interval == 1.0:
            status = "TrainingEpochLoop:on_run_end"
        else:
            # `training_epoch_end` happens after `validation_epoch_end` since Lightning v1.4
            status = "TrainingEpochLoop:on_run_end" if failure_on_training else "TrainingEpochLoop:on_advance_end"
```

tchaton:
> @tchaton this design is currently missing from https://docs.google.com/document/d/1pvhyNg17pm7-ECMTpgpZiJf6sbypeuaB8QsSP79n6k0/edit#
> 
> which makes it hard to see how this feature fits into the overall fault tolerance flow

Hey @ananthsub,

I added the description within the document.

`next reproducible part` means the part of the codebase where we know it is the safest to exit. Example on batch_end, on_epoch_end. I shared the exit strategy in the document and the test is covering most of them.

Best,
T.C

pytorch_lightning/trainer/connectors/checkpoint_connector.py
tchaton: resume_from_checkpoint is really meant to reload from a specific checkpoint generated by the user (c.f ModelCheckpoint)

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
tchaton: Done !

SeanNaren (approved):
LGTM

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
awaelchli: Other suggestion: 
If the location here is always supposed to be the function name, then maybe inspecting the stack for the class and function name is more user friendly. Here is an example after some quick testing:

```python
import inspect


class Trainer:
    def exit_gracefully_on_signal(self):
        stack = inspect.stack()
        class_name = stack[1][0].f_locals["self"].__class__.__name__
        method_name = stack[1][0].f_code.co_name
        print(f"Exiting gracefully on {class_name}:{method_name}")
        # Exiting gracefully on Loop:on_run_end


class Loop:
    def on_run_end(self):
        a = Trainer()
        a.exit_gracefully_on_signal()


Loop().on_run_end()
```

Not having to pass in the name could be slightly more user friendly when considering loop customization. What do you think?

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
tchaton: Great idea ! Let me add this :)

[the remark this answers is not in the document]
carmocca: To get the method name, this should work too

```python
        caller = inspect.stack()[1]
        class_name = caller[0].f_locals["self"].__class__.__name__
        print(f"Exiting gracefully on {class_name}:{caller.function}")
```

tests/loops/test_loops.py
carmocca: The tests seem to pass for me when these changes are undone.

And if what you say is still a problem, the key can be dropped after the assertion

tests/profiler/test_profiler.py
```diff
@@ -78,7 +78,7 @@ def test_simple_profiler_iterable_durations(simple_profiler, action: str, expect
         pass
 
     # we exclude the last item in the recorded durations since that's when StopIteration is raised
-    np.testing.assert_allclose(simple_profiler.recorded_durations[action][:-1], expected, rtol=0.2)
+    np.testing.assert_allclose(simple_profiler.recorded_durations[action][:-1], expected, rtol=0.25)
```
carmocca, line 81: Please keep the PR focused. This is tracked in a separate issue

```suggestion
    np.testing.assert_allclose(simple_profiler.recorded_durations[action][:-1], expected, rtol=0.2)
```

https://github.com/PyTorchLightning/pytorch-lightning/issues/9462

tests/trainer/connectors/test_signal_connector.py
```diff
@@ -38,16 +40,18 @@ def handler(*_):
...
-                if terminate_gracefully or register_handler:
-                    os.kill(os.getpid(), signal.SIGUSR1)
-                    sleep(0.1)
-                return super().training_step(batch, batch_idx)
+    with mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": str(int(terminate_gracefully))}):
 
-        model = TestModel()
         trainer = Trainer(default_root_dir=tmpdir, max_epochs=1, limit_train_batches=2, limit_val_batches=0)
-        trainer.fit(model)
+        with suppress(ExitGracefullyException):
```
carmocca, line 55:
```suggestion
        with pytest.raises(ExitGracefullyException):
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -256,10 +256,21 @@ def _on_evaluation_epoch_end(self) -> None:
...
+            state = getattr(self._data_fetcher.dataloader_iter, "previous_state", {})
+        state_dict["dataloader_state_dict"] = asdict(state) if is_dataclass(state) else state
         return state_dict
 
     def on_load_checkpoint(self, state_dict: Dict) -> None:
         # cache the dataloader state dict until the dataloader objects are available
         self._dataloader_state_dict = state_dict.get("dataloader_state_dict", {})
+
+    def _has_completed_evaluation_batch(self) -> bool:
+        return self.epoch_loop.batch_progress.current.ready == self.epoch_loop.batch_progress.current.completed
```
carmocca, line 273: This method should be in the epoch loop instead

carmocca (changes requested):

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -256,10 +256,21 @@ def _on_evaluation_epoch_end(self) -> None:
...
         state_dict = super().on_save_checkpoint()
-        if self._data_fetcher is not None and self._data_fetcher.dataloader_iter is not None:
-            state_dict["dataloader_state_dict"] = asdict(self._data_fetcher.dataloader_iter.previous_state)
+        if self._data_fetcher is None or self._num_val_batches_reached():
+            return state_dict
+        if self._has_completed_evaluation_batch():
+            state = self._data_fetcher.dataloader_iter.state
+        else:
+            state = getattr(self._data_fetcher.dataloader_iter, "previous_state", {})
+        state_dict["dataloader_state_dict"] = asdict(state) if is_dataclass(state) else state
```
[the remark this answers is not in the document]
tchaton: There are dataclass only if they exists. Replace the default by None to make it more explicit.

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -201,6 +207,12 @@ def on_advance_end(self):
             # progress global step according to grads progress
             self.global_step += 1
 
+        # we check this as we should have enough time to finish ``on_run_end``.
+        # TODO: @tchaton verify this assumption is True.
```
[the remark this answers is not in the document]
tchaton: This can't be done right now, only later on when benchmarking the feature.

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -256,10 +256,21 @@ def _on_evaluation_epoch_end(self) -> None:
...
 
     def on_load_checkpoint(self, state_dict: Dict) -> None:
         # cache the dataloader state dict until the dataloader objects are available
         self._dataloader_state_dict = state_dict.get("dataloader_state_dict", {})
+
+    def _has_completed_evaluation_batch(self) -> bool:
+        return self.epoch_loop.batch_progress.current.ready == self.epoch_loop.batch_progress.current.completed
+
+    def _num_val_batches_reached(self) -> bool:
+        return self._has_completed_evaluation_batch and self.epoch_loop.batch_progress.is_last_batch
```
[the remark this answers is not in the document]
tchaton: Great catch !

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -256,10 +256,21 @@ def _on_evaluation_epoch_end(self) -> None:
 
     def on_save_checkpoint(self) -> Dict:
         state_dict = super().on_save_checkpoint()
-        if self._data_fetcher is not None and self._data_fetcher.dataloader_iter is not None:
-            state_dict["dataloader_state_dict"] = asdict(self._data_fetcher.dataloader_iter.previous_state)
+        if self._data_fetcher is None or self._num_val_batches_reached():
```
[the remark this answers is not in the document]
tchaton: No, I believe it is correct. We don't want to save the state if we completed the epoch.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -248,9 +247,16 @@ def teardown(self) -> None:
...
-        if self.trainer.train_dataloader is not None:
-            state_dict["dataloader_state_dict"] = self.trainer.train_dataloader.state_dict(has_completed=False)
+
+        epoch_loop = getattr(self, "epoch_loop", None)
+        if not epoch_loop:
+            return state_dict
+
+        has_completed = self._has_completed_training_batch()
+        if self.trainer.train_dataloader is not None and not (
+            epoch_loop.batch_progress.is_last_batch and has_completed
```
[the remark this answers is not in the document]
tchaton: No, we want to skip adding the dataloader_state_dict when the epoch is completed.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -248,9 +247,16 @@ def teardown(self) -> None:
 
     def on_save_checkpoint(self) -> Dict:
         state_dict = super().on_save_checkpoint()
-        # TODO: update has_completed to its proper value
-        if self.trainer.train_dataloader is not None:
-            state_dict["dataloader_state_dict"] = self.trainer.train_dataloader.state_dict(has_completed=False)
+
+        epoch_loop = getattr(self, "epoch_loop", None)
+        if not epoch_loop:
+            return state_dict
```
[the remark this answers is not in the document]
tchaton: Done !

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -43,9 +43,14 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
     @property
     def hpc_resume_path(self) -> Optional[str]:
         dir_path_hpc = str(self.trainer.weights_save_path)
+        if dir_path_hpc is None or not os.path.isdir(dir_path_hpc):
+            return None
```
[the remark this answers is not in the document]
tchaton: Done !

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -43,9 +43,14 @@ def __init__(self, trainer: "pl.Trainer", resume_from_checkpoint: Optional[_PATH
...
         max_version = self.max_ckpt_version_in_folder(dir_path_hpc, "hpc_ckpt_")
         if max_version is not None:
             return os.path.join(dir_path_hpc, f"hpc_ckpt_{max_version}.ckpt")
+        if ".pl_auto_save.ckpt" in os.listdir(dir_path_hpc):
+            auto_save_checkpoint = os.path.join(dir_path_hpc, ".pl_auto_save.ckpt")
```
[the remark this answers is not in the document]
tchaton: Done !

tests/trainer/connectors/test_signal_connector.py
tchaton: Done !

pytorch_lightning/loops/dataloader/evaluation_loop.py
tchaton: c.f conversation above with @awaelchli. It was originally in the epoch loop but requested to be moved up :)

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
carmocca: @awaelchli one problem of option 2 is that if the child changes (for example if the batch_progress properties are customized) then this code cannot be easily adapted as it is defined in the parent.

awaelchli: accessing a protected member of one loop from another offers the same problem. option 1 in my comment above is still  possible I think

tests/utilities/test_auto_restart.py
```diff
@@ -1060,3 +1062,140 @@ def run(should_fail, resume):
...
+    def _signal(self, message):
+        if self.should_signal:
+            os.kill(os.getpid(), signal.SIGUSR1)
+            # small optimization to skip as soon as async signal has being triggered.
+            t0 = time()
+            while not self.trainer._terminate_gracefully:
+                t1 = time()
+                if (t1 - t0) > 0.1:
+                    self.trainer._terminate_gracefully = True
```
carmocca, line 1084: why does this not use sleep?

tchaton: I am using a while loop to make sure I exit as fast as possible. This helped saved couple of seconds from the tests compared to a sleep.

carmocca: I had a look after not believing it and that happens because `self.trainer._terminate_gracefully` becomes `True` right after the signal is sent so the while loop never runs. This can be entirely removed

tests/utilities/test_auto_restart.py
```diff
@@ -1060,3 +1061,137 @@ def run(should_fail, resume):
...
+    model_restarted = _fit_model(tmpdir, False, val_check_interval, failure_on_step, failure_on_training, on_last_batch)
+
+    # check the batches
+    actual = torch.cat(model_signaled.seen_train_batches + model_restarted.seen_train_batches)
+    expected = torch.cat(model_total.seen_train_batches)
+    assert torch.equal(actual, expected)
+
+    # check the weights
+    # FIXME
+    # assert not torch.equal(model_signaled.layer.weight, model_total.layer.weight)
```
carmocca, line 1183: I added this assertion but it fails. We should look into it

tchaton: Yes, this is normal. But I still added a FIXME for on_last_batch.

tests/utilities/test_auto_restart.py
tchaton: This doesn't seem to happen for all the environment in the CI. If you remove it, it would fail.

carmocca (approved):

PR #9749, opened 2021-09-29
Add logging messages to notify when `FitLoop` stopping conditions are met

### LINKED ISSUE #9636: `Trainer.fit()` multiple times
###  Context

I noticed in the unit test case ``test_dataloaders_reset_and_attach`` in `test_dataloaders.py` that `trainer.fit()` was called twice with different train_dataloaders. ([code pointer](https://github.com/PyTorchLightning/pytorch-lightning/blob/5a846d48ceb5412636bafc3037d04bc2f1bcc0e6/tests/trainer/test_dataloaders.py#L1310-L1320))
```python
    model = BoringModel()
    trainer = Trainer(default_root_dir=tmpdir, fast_dev_run=1)

    # 1st fit
    trainer.fit(model, train_dataloaders=dataloader_0, val_dataloaders=dataloader_1)
    assert trainer.train_dataloader.loaders.dataset is dataloader_0.dataset
    assert trainer.val_dataloaders[0].dataset is dataloader_1.dataset
    # 2nd fit
    trainer.fit(model, train_dataloaders=dataloader_2, val_dataloaders=dataloader_3)
    assert trainer.train_dataloader.loaders.dataset is dataloader_2.dataset
    assert trainer.val_dataloaders[0].dataset is dataloader_3.dataset

```

The original test case succeed under the current implementation that train/val_dataloader will be reattached before fit_loop.run() is called ([code pointer](https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/trainer/trainer.py#L1129-L1135))
```python
# reload data when needed
model = self.lightning_module

self.reset_train_val_dataloaders(model)

self.fit_loop.trainer = self
self.fit_loop.run()
```

However, the second fit_loop could never properly run, because the first fit_loop could property stop when either max_epochs or max_steps are reached, and meanwhile fit_loop.done = True, which leads to fit_loop.skip = True ([code pointer](https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/loops/fit_loop.py#L152-L177)). Without initializing a new trainer, the second fit_loop run is just skipped ([code pointer](https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/loops/base.py#L103-L104)). 

### Discussion:
Do we allow user to start multiple trainer.fit() with train_dataloaders? I understand the needs to have trainer.validate()/test()/predict(), but I think the pattern that allowing trainer.fit() multiple times could be problem. One edge case I can think of to call trainer.fit() multiple times is that trainer.fit() is interrupted by early stopping condition and resumed fit again with different training data. At least, we need to document this or add warning so that users could be aware of fit_loop actually did not start.  

### Pitch
- Update this test case by removing the second fit call
- Document the change

### Environment

- PyTorch Lightning Version (e.g., 1.3.0):
- PyTorch Version (e.g., 1.8)
- Python version:
- OS (e.g., Linux):
- CUDA/cuDNN version:

### Additional context

cc @borda @rohitgr7 @carmocca @justusschock @ananthsub @ninginthecloud

jinyoung-lim:
## What does this PR do?

Fix for fit_loop silently stopping without a warning when trainer's early stopping condition is not met but max_epochs or max_steps are met.

This is to raise a caution to users who might be calling Trainer.fit() multiple times and earlier fitting loop meeting max_epochs or max_steps condition so stopping. Then the later fit would start and silently stop as max_epochs or max_steps are met.

Fixes #9636 

None

## Before submitting

## PR review

## Did you have fun?


jinyoung-lim:
Cannot add labels and such as I do not have permission (not a contributor yet).

carmocca (approved):

src/pytorch_lightning/loops/fit_loop.py
```diff
@@ -150,31 +150,40 @@ def _results(self) -> _ResultCollection:
...
 
-        should_stop = False
         if self.trainer.should_stop:
             # early stopping
             met_min_epochs = self.epoch_progress.current.processed >= self.min_epochs if self.min_epochs else True
             met_min_steps = self.epoch_loop.global_step >= self.min_steps if self.min_steps else True
             if met_min_epochs and met_min_steps:
-                should_stop = True
+                self.trainer.should_stop = True
+                rank_zero_debug(f"`Trainer.fit` stopped: `should_stop` was set.")
```
rohitgr7, line 178:
```suggestion
                rank_zero_debug(f"`Trainer.fit` stopped: `trainer.should_stop` was set.")
```

src/pytorch_lightning/loops/fit_loop.py
```diff
@@ -150,31 +150,40 @@ def _results(self) -> _ResultCollection:
...
-                    f" ({self.min_epochs}) or minimum steps ({self.min_steps}) has"
-                    " not been met. Training will continue..."
+                rank_zero_info(
+                    f"Trainer was signaled to stop but the required `min_epochs={self.min_epochs!r}` or"
+                    f" `min_steps={self.min_steps!r}` has not been met. Training will continue..."
                 )
-        self.trainer.should_stop = should_stop
-
-        return stop_steps or should_stop or stop_epochs or self.trainer.num_training_batches == 0
+        self.trainer.should_stop = False
```
rohitgr7, line 185: let's say `min_epochs=2` and `max_epochs=10`
user sets `should_stop=True` after epoch 1. Now training doesn't stop since `min_epochs` haven't met, but should it continue until 10 epochs, or should be stopped after the 2nd epoch once `min_epochs` have met?

asking because we are setting `should_stop=False` here i.e overriding the user's decision.

rohitgr7 (approved):

src/pytorch_lightning/loops/fit_loop.py
carmocca: If I'm not mistaken, the changes in this PR do not change the logic regarding that decision.

Now, how it works is that whoever requested to stop will need to re-request it in the next iteration/epoch/whatever. I remember looking into this a while back and there were tests that require this behaviour.

carmocca: The problem is that the inner `TrainingEpochLoop` also checks for `should_stop` in it's `done`. So to do that change, we would need to check for `min_*` in all inner loops that check the attribute

rohitgr7: so currently the behavior is?

if someone sets `should_stop` in the middle of some epoch, then it breaks that epoch loop and continues running the next epochs completely until `min_epochs` is met.
```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
>>>>>>>>>>>>>>> (should_stop=True, epoch breaks)
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> (min_epochs)
```

should we add these checks there too to ensure that epoch is completed at least if min_condition is not met?

carmocca: Personally, I would say no (keep as it is). But this is a design question so I don't see a right or wrong answer. Feel free to open an issue to round up the choices and their impact.

rohitgr7: done! #13789

tests/tests_pytorch/loops/test_training_loop.py
```diff
@@ -136,6 +140,49 @@ def validation_step(self, *args):
     assert model.validation_called_at == (0, 5)
 
 
+def test_fit_loop_done_log_messages(caplog):
```
awaelchli, line 143: Thank you for writing a unit test instead of training 6 models!

awaelchli (approved):

PR #9755, opened 2021-09-29
Raise an exception if using `amp_level` with native `amp_backend`

### LINKED ISSUE #9739: There should be a warning or an exception if using `amp_backend="native"` and setting an `amp_level` (it doesn't use it)
## 🚀 Feature
There should be a warning or an exception if using `Trainer(amp_backend="native")` and setting a value to the `amp_level` argument (of `Trainer`).

### Motivation
It's not obvious for someone who hasn't looked into the code or who knows a lot about Apex that using Pytorch's native amp backend that the amp_level argument will have no effect, giving a false understanding of what is actually happening. Knowing how much of the model was made fp16 is really important to understand what is happening in the training, to make training reproducible.

With `amp_type="apex"` (amp_level used):
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L586
With `amp_type="native"` (not used):
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L573

### Pitch
There should be an exception or a warning added, maybe around https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L573, when `Trainer(amp_level=...)` is set and `Trainer(amp_backend="native")`. Again, Knowing how much of the model was made fp16 is really important to understand what is happening in the training, to make training reproducible.

### Alternatives
Rename `amp_level` to `apex_level` to make it more obvious that it only works with the `apex` backend. Breaks backwards compatibility.

rohitgr7:
## What does this PR do?

Fixes #9739

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -552,11 +552,18 @@ def select_precision_plugin(self) -> PrecisionPlugin:
                 return TPUHalfPrecisionPlugin()
 
             if self.amp_type == AMPType.NATIVE:
+                if self.amp_level is not None:
+                    raise MisconfigurationException(
```
Borda, line 556: I would just make it a a warning

pytorch_lightning/trainer/trainer.py
```diff
@@ -172,7 +172,7 @@ def __init__(
         prepare_data_per_node: Optional[bool] = None,
         plugins: Optional[Union[PLUGIN_INPUT, List[PLUGIN_INPUT]]] = None,
         amp_backend: str = "native",
```
ananthsub, line 174: Should amp backend be optional too? 

It doesn't match the defaults with precision=32 on the trainer.

On the other hand, if this were optional, users would need to set this flag and precision=16 for mixed precision training

tchaton: I think it is fine this way. Let's confusing to the user who knows native is the default one.

tchaton (approved):
LGMT !

pytorch_lightning/trainer/connectors/accelerator_connector.py
rohitgr7: it's now None by default. if someone sets it explicitly, we should expect them to set this correctly as well otherwise an exception seems ok to me.. also it happens during init so one doesn't have to wait to realize the mistake later at some point in case they ignore the warning.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -563,11 +563,18 @@ def select_precision_plugin(self) -> PrecisionPlugin:
...
+                        f"You have asked for amp_level={self.amp_level} which is not supported "
+                        "with amp_backend='native'."
```
awaelchli, line 569:
```suggestion
                        f"You have asked for `amp_level={self.amp_level}` which is not supported"
                        " with `amp_backend='native'`."
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -581,6 +588,10 @@ def select_precision_plugin(self) -> PrecisionPlugin:
                         "Sharded Plugin is not supported with Apex AMP, please using native AMP for 16-bit precision."
                     )
                 log.info("Using APEX 16bit precision.")
+
+                if self.amp_level is None:
+                    self.amp_level = "O2"
```
awaelchli, line 593:
```suggestion
               self.amp_level = self.amp_level or "O2"
```

tests/plugins/test_amp_plugins.py
```diff
@@ -224,3 +224,9 @@ def test_cpu_amp_precision_16_throws_error(tmpdir):
             default_root_dir=tmpdir,
             precision=16,
         )
+
+
+@RunIf(min_gpus=1, amp_native=True)
+def test_amp_level_raises_error_with_native(tmpdir):
```
awaelchli, line 230: this test should go to `tests/accelerators/test_accelerator_connector`

awaelchli (approved):

tests/accelerators/test_accelerator_connector.py
```diff
@@ -636,3 +636,9 @@ def test_validate_precision_type(tmpdir, precision):
 
     with pytest.raises(MisconfigurationException, match=f"Precision {precision} is invalid"):
         Trainer(precision=precision)
+
+
+@RunIf(min_gpus=1, amp_native=True)
+def test_amp_level_raises_error_with_native(tmpdir):
+    with pytest.raises(MisconfigurationException, match="not supported with amp_backend='native'"):
```
rohitgr7, line 643:
```suggestion
    with pytest.raises(MisconfigurationException, match="not supported with `amp_backend='native'`"):
```

kaushikb11 (approved):

PR #9765, opened 2021-09-29
add unit tests for `pl.utilities.grads`

awaelchli:
## What does this PR do?

Adds tests and coverage for `pytorch_lightning.utilities.grads.grad_norm` function.
Found an edge case where L0 "norm" (p=0) does not yield the correct result for the total norm. This is because the implementation of the total norm there only works for proper norms (the L0 is not a real norm). 

Option A: special case the L0 norm computation
Option B: don't support `norm_type=0`

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/tests/utilities/test_grads.py
+++ b/tests/utilities/test_grads.py
@@ -59,6 +59,8 @@ from pytorch_lightning.utilities import grad_norm
     ],
 )
 def test_grad_norm(norm_type, expected):
+    """Test utility function for computing the p-norm of individual parameter groups and norm in total."""
+
     class Model(nn.Module):
         def __init__(self):
             super().__init__()
--- a/tests/utilities/test_grads.py
+++ b/tests/utilities/test_grads.py
@@ -23,10 +23,10 @@ from pytorch_lightning.utilities import grad_norm
 @pytest.mark.parametrize(
     "norm_type,expected",
     [
-        # bug in for L0 "norm"? total norm should be 5
+        # bug for L0 "norm"! total norm should be 5 here
         (
             0,
-            {"grad_0.0_norm_param0": 3, "grad_0.0_norm_param1": 2, "grad_0.0_norm_total": 2},
+            {"grad_0.0_norm_param0": 3, "grad_0.0_norm_param1": 2, "grad_0.0_norm_total": 5},
         ),
         (
             1,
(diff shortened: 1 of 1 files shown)


tests/utilities/test_grads.py
```diff
@@ -0,0 +1,77 @@
...
+import torch
+import torch.nn as nn
+
+from pytorch_lightning.utilities import grad_norm
+
+
+@pytest.mark.parametrize(
+    "norm_type,expected",
+    [
+        # bug for L0 "norm"! total norm should be 5 here
```
carmocca, line 26: Why do you say it is a bug?

```python
np.linalg.norm(np.array([3., 2.]), 0)
Out[31]: 2.0
```

which matches the definition of vector 0-norm: `sum(x != 0)`

awaelchli: yes, 100% a bug in our implementation, I am sure. Agree your example is fine and I'm not saying `torch.norm` computes an incorrect result. It's our implementation that takes a shortcut for efficiency to avoid recomputing values when computing the total  norm of all parameters combined. That shortcut holds up for p>0 because of properties of the Lp-norm. But since for p=0 it is not a true norm, that formula breaks down. Here is what we do:

Let v0 and v1 be two vectors. 

Individual norms:
```
n0 = norm(v0, p) 
n1 = norm(v1, p)
```

Total norm of all elements combined:
```
t = norm([v0, v1], p)
```

Instead of the above, our implementation does this using the already computed values n0 and n1:
```
t = norm([n0, n1], p)
```

One can show that
```
norm([v0, v1], p) = norm([n0, n1], p) for p > 0
```

But for p = 0 this does not hold true. 
Example:

```
v0 = [-1, 2]
v1 = [3, 4]

```
We will get `n0 = n1 = 2`
but
```
2 = norm([2, 2], 0) != norm([-1, 2, 3, 4], 0) = 4
```

carmocca: This hasn't been an issue because:

https://github.com/PyTorchLightning/pytorch-lightning/blob/44aed17affc396a82ea3bb93345c1892ebaee82b/pytorch_lightning/loops/optimization/optimizer_loop.py#L485

but if we don't want to support 0-norm, it should raise an exception instead of silently skipping it

PR #6231, opened 2021-02-26
feat(wandb): log models as artifacts

borisdayma:
## What does this PR do?

Fixes #4903

should be merged after #5931 

## Before submitting
- [X] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [X] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [X] Did you make sure to update the documentation with your changes? (if necessary)
- [X] Did you write any new necessary tests? (not for typos and docs)
- [X] Did you verify new and existing tests pass locally with your changes?
- [X] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [ ] Is this pull request ready for review? (if not, please submit in draft mode)
 - [ ] Check that all items from **Before submitting** are resolved
 - [ ] Make sure the title is self-explanatory and the description concisely explains the PR
 - [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #9783, opened 2021-10-01
[bugfix] Run `logger.after_save_checkpoint` in model checkpoint's `on_train_end` hook

dalessioluca:
## What does this PR do?

This PR makes sure that the logger.after_save_checkpoint is called after ModelCheckpoint creates a checkpoint "on_train_end".

Note: A bigger refactor has been discussed there: https://github.com/PyTorchLightning/pytorch-lightning/pull/6231#discussion_r592769726

Details:
```py
import torch
from pytorch_lightning.loggers import CSVLogger 
from pytorch_lightning.trainer import Trainer


class Logger(CSVLogger):
    call_counter = 0
    
    def after_save_checkpoint(self, checkpoint_callback: "ReferenceType[ModelCheckpoint]") -> None:
        print("called after new ckpt is saved")
        self.call_counter += 1


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        return {"loss": loss}

def run():

    # These two together make sure that I save a ckpt every XXX epochs AND at the end of training
    ckpt_train_end = ModelCheckpoint(
        filename="last",
        save_on_train_epoch_end=False,
        save_last=True,
    )

    ckpt_train_interval = ModelCheckpoint(
        filename="my_checkpoint-{epoch}",
        save_on_train_epoch_end=True,
        save_last=False,
        every_n_epochs=6,
    )

    model = BoringModel()

    trainer = Trainer(
        logger=Logger(save_dir='./logging_dir', flush_logs_every_n_steps=1),
        max_epochs=9,
        log_every_n_steps=1,
        weights_save_path="saved_ckpt",
        callbacks=[ckpt_train_interval, ckpt_train_end],
    )

    assert trainer.logger.call_counter == 2

if __name__ == "__main__":
    run()
```

This code saves 2 ckpt_files in the "saved_ckpt" folder but the `logger.after_save_checkpoint` is called only once. 

-->

Fixes #

None

## Before submitting

- [x ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?

## PR review

## Did you have fun?


pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -346,6 +346,10 @@ def on_train_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -
         self._save_last_checkpoint(trainer, monitor_candidates)
         trainer.fit_loop.global_step += 1
 
+        # notify loggers
+        if trainer.is_global_zero and trainer.logger:
+            trainer.logger.after_save_checkpoint(proxy(self))
+
```
ananthsub, line 352: could you add a test for this to confirm this bug doesn't happen again?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -430,11 +434,6 @@ def __validate_init_configuration(self) -> None:
                     f"ModelCheckpoint(save_top_k={self.save_top_k}, monitor=None) is not a valid"
                     " configuration. No quantity for top_k to track."
                 )
-            if self.save_last:
-                rank_zero_warn(
-                    "ModelCheckpoint(save_last=True, save_top_k=None, monitor=None) is a redundant configuration."
-                    " You can save the last checkpoint with ModelCheckpoint(save_top_k=None, monitor=None)."
-                )
```
ananthsub, line 437: why's this being removed?

ananthsub (commented):
thanks for identifying this! did you see any other points in the checkpoint callback where the logger was not triggered?

pytorch_lightning/callbacks/model_checkpoint.py
dalessioluca: I added the test.

pytorch_lightning/callbacks/model_checkpoint.py
dalessioluca: As it stand, this warning is incorrect.
It is only correct if we make additional assumptions, for example every_n_epochs != 0.

For example:
`ModelCheckpoint(save_last=True, every_n_epochs=0, save_top_k=None, monitor=None)` only saves a ckpt at the end of training while `ModelCheckpoint(every_n_epochs=0, save_top_k=None, monitor=None)` does not save any ckpt.

carmocca: In that case, why not check it in the `if` condition?

rohitgr7:
hey @dalessioluca , thank you for this PR and identifying issues here. looks like it does log the model here atleast for wandb: https://github.com/PyTorchLightning/pytorch-lightning/blob/45c45dc7b018f9a2db60f5df1a3f7dbbb45ccb36/pytorch_lightning/loggers/wandb.py#L461-L464

but the way these are logged and how it connects to the logger after checkpointing isn't super convenient and can be done in a more optimal way. We should call after_save_checkpoint inside trainer.save_checkpoint(). I'll do another PR after sometime which might solve more issues and this one too. In case you want to make changes to loggers and checkpoint then it would be great else you can close this one. Again thanks a lot for this.. was really helpful identifying more underlying issues.

tchaton:
Dear @dalessioluca,

Based on @rohitgr7 latest comment, I believe it is better to make sure `after_save_checkpoint ` and `save_checkpoint` are called together.

Would you be willing to keep going with this PR based on @rohitgr7 comments ?

Best,
T.C

dalessioluca:
It indeed seems more clean to call  after_save_checkpoint inside save_checkpoint. 
I will modify the PR accordingly.

rohitgr7:
also you might need to update some loggers as well. for starters, wandb won't be needing the finalize method if we call `after_save_checkpoint` right after `save_checkpoint`.

carmocca:
I believe this is not necessary anymore. We call `after_save_checkpoint` already whenever we save: https://github.com/Lightning-AI/lightning/blob/233b36b185ae71b480ddd5c87ff5d12cd70c88fe/src/pytorch_lightning/callbacks/model_checkpoint.py#L386-L394

Please, correct me if I'm wrong.

PR #9786, opened 2021-10-01
Log LR using LearningRateMonitor even when LR Scheduler is not defined. 

### LINKED ISSUE #7468: Extend the Learning Rate Monitor callback to log the initial lr for optimizers
## 🚀 Feature
Log the initial learning rate of each optimizer

### Motivation

Currently the LR monitor callback only takes effect if the user specifies LR schedulers for training. However, one could want to record what the initial learning rate used for their optimizer's parameter groups even without any LR schedulers. 

### Pitch

We could log this once on train start following the logic here: https://github.com/PyTorchLightning/pytorch-lightning/blob/f6fe715e73cc87bfc106329f9103564097515379/pytorch_lightning/callbacks/lr_monitor.py#L138-L156

cc @simran2905

VirajBagal:
## What does this PR do?
In this PR, the `LearningRateMonitor` callback is extended to log learning rates of each optimizer even when LR schedulers are not specified by the user in training. 

When LR schedulers are not specified, `trainer.optimizers` is used instead of `trainer.lr_schedulers` for getting names of the optimizers, and then we iterate over the optimizers to log their learning rates.  

Extra tests are written to make sure that the learning rates are logged with correct names even when schedulers are not defined. 

## Motivation
Currently the LR monitor callback only takes effect if the user specifies LR schedulers for training. However, one could want to record what the initial learning rate used for their optimizer's parameter groups even without any LR schedulers. This PR tries to solve this issue. 
Fixes #7468 

### Does your PR introduce any breaking changes? 
No

## Before submitting

## PR review

### DIFF
--- a/pytorch_lightning/callbacks/lr_monitor.py
+++ b/pytorch_lightning/callbacks/lr_monitor.py
@@ -106,18 +106,14 @@ class LearningRateMonitor(Callback):
                 "Cannot use `LearningRateMonitor` callback with `Trainer` that has no logger."
             )
 
-        if not trainer.lr_schedulers:
-            rank_zero_warn(
-                "You are using `LearningRateMonitor` callback with models that"
-                " have no learning rate schedulers. Please see documentation"
-                " for `configure_optimizers` method.",
-                RuntimeWarning,
-            )
-
         if self.log_momentum:
 
             def _check_no_key(key):
-                return any(key not in sch["scheduler"].optimizer.defaults for sch in trainer.lr_schedulers)
+                if trainer.lr_schedulers:
+                    return any(key not in sch["scheduler"].optimizer.defaults for sch in trainer.lr_schedulers)
+                else:
+                    return any(key not in optimizer.defaults for optimizer in trainer.optimizers)
+
 
             if _check_no_key("momentum") and _check_no_key("betas"):
                 rank_zero_warn(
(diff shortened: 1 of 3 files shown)


pytorch_lightning/callbacks/lr_monitor.py
```diff
@@ -155,12 +154,30 @@ def on_train_epoch_start(self, trainer, *args, **kwargs):
...
+                    use_betas = "betas" in opt.defaults
+
+                    for i, pg in enumerate(param_groups):
+                        name_and_suffix = self._add_suffix(name, param_groups, i)
+                        lr = self._extract_lr(pg, name_and_suffix)
+                        latest_stat.update(lr)
+                        momentum = self._extract_momentum(
+                            param_group=pg, name=name_and_suffix.replace(name, f"{name}-momentum"), use_betas=use_betas
+                        )
+                        latest_stat.update(momentum)
```
ananthsub, line 174: this code is duplicated in the else case below. could we abstract this to a common helper function to reduce the risk of divergence?

pytorch_lightning/callbacks/lr_monitor.py
```diff
@@ -259,6 +276,35 @@ def _find_names(self, lr_schedulers: List, add_lr_sch_names: bool = True) -> Lis
...
+                    f"`name` values. {name} has duplicated parameter group names {duplicates}"
+                )
+
+            name = self._add_prefix(name, optimizer_cls, seen_optimizer_types)
+
+            names.extend(self._add_suffix(name, param_groups, i) for i in range(len(param_groups)))
+
+            if add_lr_sch_names:
+                self.lr_sch_names.append(name)
+
```
ananthsub, line 305: same here. there's a lot of duplicated code above. let's try to abstract these into common helper functions

VirajBagal: I have added helper methods.

pytorch_lightning/callbacks/lr_monitor.py
VirajBagal: I have now added helper methods

PR #9788, opened 2021-10-01
Fix Deepspeed and lightning calling scheduler

### LINKED ISSUE #9771: DeepSpeed & Lightning are both calling the scheduler
## 🐛 Bug

A user in our slack channel noticed that the scheduler step is being called by DeepSpeed and Lightning. Lightning should not call this, and DeepSpeed should be in control.

### To Reproduce

This test fails!
```python
import os

import mock
import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

        optimizer = torch.optim.SGD(self.layer.parameters(), lr=0.1)
        scheduler = torch.optim.lr_scheduler.StepLR(optimizer, 1, gamma=0.1)
        return {
            "optimizer": optimizer,
            "lr_scheduler": {
                "scheduler": scheduler,
            },
        }


@mock.patch('torch.optim.lr_scheduler.StepLR.step', autospec=True)
def test_deepspeed(mock_step):

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=0,
        num_sanity_val_steps=0,
        max_epochs=1,
        fast_dev_run=True,
        gpus=1,
        weights_summary=None,
        plugins='deepspeed'
    )
    # once before training, and once during training.
    assert mock_step.call_count == 2
```

### Expected behaviour

Above should succeed. I also see there may be an issue with `test` also calling the step function internally in DeepSpeed, but thats for a different issue.

cc @tchaton 

### Environment

* CUDA:
        - GPU:
                - NVIDIA GeForce RTX 3090
        - available:         True
        - version:           11.1
* Packages:
        - numpy:             1.21.2
        - pyTorch_debug:     False
        - pyTorch_version:   1.9.1
        - pytorch-lightning: 1.5.0dev
        - tqdm:              4.49.0
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.8.8

SeanNaren:
## What does this PR do?

Fixes #9771 

After speaking to @tchaton we decided to remove the scheduler responsibility from DeepSpeed, fully relying on Lightning to control when to call the lr scheduler step function. This also allows users to use epoch vs step calls which is not possible in DeepSpeed.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/plugins/training_type/deepspeed.py
+++ b/pytorch_lightning/plugins/training_type/deepspeed.py
@@ -441,7 +441,11 @@ class DeepSpeedPlugin(DDPPlugin):
 
         # although we set these here, deepspeed manages the specific optimizer logic
         self.lightning_module.trainer.optimizers = [deepspeed_optimizer]
+
+        deepspeed_scheduler = model.lr_scheduler
         if deepspeed_scheduler is not None:
+            # disable deepspeed lr scheduling as lightning manages scheduling
+            model.lr_scheduler = None
             lr_scheduler["scheduler"] = deepspeed_scheduler
             self.lightning_module.trainer.lr_schedulers = [lr_scheduler]
         self.model = model
--- a/tests/plugins/test_deepspeed_plugin.py
+++ b/tests/plugins/test_deepspeed_plugin.py
@@ -265,8 +265,6 @@ def test_deepspeed_run_configure_optimizers(tmpdir):
             assert isinstance(trainer.lr_schedulers[0]["scheduler"], torch.optim.lr_scheduler.StepLR)
             # check that the lr_scheduler config was preserved
             assert trainer.lr_schedulers[0]["name"] == "Sean"
-            # Ensure DeepSpeed engine has initialized with our lr_scheduler
-            assert isinstance(trainer.model.lr_scheduler, torch.optim.lr_scheduler.StepLR)
 
     class TestModel(BoringModel):
         def configure_optimizers(self):
@@ -303,8 +301,6 @@ def test_deepspeed_config(tmpdir, deepspeed_zero_config):
             assert isinstance(trainer.optimizers[0], FP16_DeepSpeedZeroOptimizer)
             assert isinstance(trainer.optimizers[0].optimizer, torch.optim.SGD)
             assert isinstance(trainer.lr_schedulers[0]["scheduler"], WarmupLR)
-            # Ensure DeepSpeed engine has initialized with our lr_scheduler
-            assert isinstance(trainer.model.lr_scheduler, WarmupLR)
 
     model = BoringModel()
     trainer = Trainer(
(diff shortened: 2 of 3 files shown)


tchaton (approved):
LGMT !

awaelchli (approved):

tests/plugins/test_deepspeed_plugin.py
```diff
@@ -921,3 +921,47 @@ def test_dataloader(self):
...
+    set to step."""
+    _run_scheduler_test(mock_step, max_epoch=2, limit_train_batches=2, interval="step")
+
+
+@mock.patch("torch.optim.lr_scheduler.StepLR.step", autospec=True)
+@RunIf(min_gpus=1, deepspeed=True, special=True)
+def test_deepspeed_scheduler_step_count_epoch(mock_step):
+    """Test to ensure that the scheduler is called the correct amount of times during training when scheduler is
+    set to epoch."""
+    _run_scheduler_test(mock_step, max_epoch=2, limit_train_batches=2, interval="epoch")
```
rohitgr7, line 939: maybe parametrize the interval here and put the below function logic inside the test?? else looks good!!

rohitgr7 (approved):

tests/plugins/test_deepspeed_plugin.py
SeanNaren: Oh sorry I should've mentioned, I couldn't make this work (probably because its a special test?) I had to split it up like this unfortunately :(

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -441,7 +441,11 @@ def _initialize_deepspeed_train(self, model):
 
         # although we set these here, deepspeed manages the specific optimizer logic
         self.lightning_module.trainer.optimizers = [deepspeed_optimizer]
+
+        deepspeed_scheduler = model.lr_scheduler
         if deepspeed_scheduler is not None:
+            # disable deepspeed lr scheduling as lightning manages scheduling
+            model.lr_scheduler = None
```
carmocca, line 448: why this option instead of not passing it to `deepspeed.initialize` above?

SeanNaren: The main reason is if the user has used the config file to determine what the optimizer/scheduler should be (this is the case if you're using the standard deepspeed recipe config and want to replicate something) then we still need to use deepspeed to select the appropriate lr scheduler if any

PR #9837, opened 2021-10-06
disable logger in tuner if it's disabled by user explicitly

rohitgr7:
## What does this PR do?

small fix:
We should disable logger and not assign it to dummy logger if it's explicitly disabled by user inside Trainer. Dummy logger won't create any logs but we still have some other console logs that relies on `if self.logger and other_condition`, so if we assign it to DummyLogger `if self.logger` will be True which is a bit incorrect. Found this while debugging another issue.

Does it need a changelog or any test? since it's just a minor update.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/connectors/debugging_connector.py
+++ b/pytorch_lightning/trainer/connectors/debugging_connector.py
@@ -60,7 +60,7 @@ class DebuggingConnector:
             self.trainer.fit_loop.max_epochs = 1
             val_check_interval = 1.0
             self.trainer.check_val_every_n_epoch = 1
-            self.trainer.logger = DummyLogger()
+            self.trainer.logger = DummyLogger() if self.trainer.logger is not None else None
 
             rank_zero_info(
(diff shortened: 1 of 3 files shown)


rohitgr7:
> ```python
> def __bool__(self):
>     return False
> ```
> 
> for the DummyLogger? This should solve the issue and not become a breaking change. Turning the logger off temporarily can be a useful feature and should not require changing the user code where e.g. `self.logger.experiment.log()` is called.

@awaelchli, good suggestion but if we don't assign it during `.fit` why do we need to assign it to a `DummyLogger` during `.tune` if the user is explicitly setting it to `Trainer(logger=False (not default))`. Isn't the error expected if someone is using `self.logger.experiment.something`?

I don't think it will be a breaking change because if someone is setting `Trainer(logger=False)` then `self.logger.experiment.something` won't work for them on `trainer.fit`.

Also this won't work if someone is using `if self.logger is None` somewhere in their code.
```py
def __bool__(self):
    return False
```

Issue #6074, opened 2021-02-19
Support training resume and saving best model for SWA


b02202050:
## 🚀 Feature
To support stop-resume and saving the model based on the best validation performance when using Stochastic weight averaging.

### Motivation
If one does not reach the end of the training, the SWA model would not be saved to the checkpoint.
For example:
1. One may want to save the best validation performance model before overfitting.
2. The training becomes unstable to reach the final epoch.

I observe that the saved checkpoint is still the original model even if I use an SWA callback.

### Pitch
1. Use SWA model to run validation step instead of the original model.
    1. We may consider to **Load SWA weight before validation** and **Restore original model weight after validation**
1. Save both original model weight and SWA model weight into checkpoint for resume training.
    1. With the concern of say one want to load the model for testing, he/she might expect the model be the SWA weight. Therefore, we can save the SWA weight to the `checkpotint['state_dict']` and let the original model weight saved as a callback state.

### Additional context
@Borda @MilesCranmer, I do not have much experience on coding neither am I familiar with PyTorch Lightning. would you help? Thx a lot. :wink:

adamreeve:
I've been looking into how batch normalization currently works when using SWA to better understand how this should be handled during validation. The mean and variance of inputs to the BatchNorm layers depend on the weights of lower layers so the running estimates computed by the underlying model won't be accurate when the weights are replaced with average weights. Therefore the mean and variance are computed with a full pass over the training set at the end of training, after replacing the weights with the averaged weights. This is done by increasing the number of epochs by one but setting `_skip_backward` to True for the last epoch to skip the optimization step, and modifying the batch norm layers to reset their mean and variance tensors before this final epoch and setting momentum to None so that they compute a simple cumulative moving average.

The validation passes should probably also update the batch norm parameters in order to accurately represent the model performance. Using the same approach of adding extra non-optimization training epochs before validation seems like it could get quite awkward though. I'm wondering whether it would be simpler to do this without modifying the training loop parameters by just using the training data fetcher directly. I've updated my draft PR to use this approach but I'm not that familiar with the PyTorch code base so am not sure what I might be missing by using the data fetcher directly rather than letting the training loop handle this.

The `StochasticWeightAveraging.restore_average_parameters_from_checkpoint` method will also need to be updated to take a data module or training data loader so that it can recompute batch norm parameters.

carmocca:
Implemented in https://github.com/Lightning-AI/lightning/pull/9938

PR #8578, opened 2021-07-27
[WIP] Fix current_epoch value on training end

carmocca:
# Blocked by #8477 and enabling restoring the ckpt progress tracking state by default.

## What does this PR do?

Fixes #5007
Part of #7406 

### Does your PR introduce any breaking changes? If yes, please list them.

With this PR, a `Trainer(max_epochs=1, limit_train_batches=1)` which saves a checkpoint`on_train_epoch_end` will have the `current_epoch=0, global_step=1` values saved. This is because we consider `on_train_epoch_end` to be part of the epoch itself:

```python
for epoch in epochs:
  ...
  on_train_epoch_end()
  epoch_count++
on_train_end()
```

`current_epoch` is now increased by 1 when training is over.
This means that if a model is run for 3 epochs (0, 1, 2), `trainer.current_epoch` will return `3`.
This is consistent with the fact that a new trainer returns `trainer.current_epoch == 0`, as in, the 0-th (first) epoch needs to run.
This will be breaking for anybody accessing the `trainer.current_epoch` attribute after `fit` is finished

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [ ] Is this pull request ready for review? (if not, please submit in draft mode)
 - [ ] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

🤪

Issue #6019, opened 2021-02-16
Support Tensorboard logs dir structure costumizaton


edenlightning:
Feature request:
As a user, i would like to have control over checkpoint dir name and structure.

Motivation: 
From @janhenriklambrechts (janhenrik.lambrechts@gmail.com):

My logging version directory of a single pytorch-lighting project with tensorboard logs looks something like this:

```
├── checkpoints
│   ├── epoch=86-max_acc=0.7446.ckpt
│   └── last.ckpt
├── events.out.tfevents.1613393562.lthpc.29645.0
├── hparams.yaml
├── train_acc_classifier_1
│   └── events.out.tfevents.1613393662.lthpc.29645.8
├── train_acc_classifier_2
│   └── events.out.tfevents.1613393662.lthpc.29645.7
├── train_acc_classifier_3
│   └── events.out.tfevents.1613393662.lthpc.29645.6
├── train_acc_classifier_4
│   └── events.out.tfevents.1613393662.lthpc.29645.5
├── val_acc_classifier_1
│   └── events.out.tfevents.1613393563.lthpc.29645.4
├── val_acc_classifier_2
│   └── events.out.tfevents.1613393563.lthpc.29645.3
├── val_acc_classifier_3
│   └── events.out.tfevents.1613393563.lthpc.29645.2
└── val_acc_classifier_4
    └── events.out.tfevents.1613393563.lthpc.29645.1
```
my network has multiple exits and I thus want to log the val_acc (or train_acc) of all these exits in a single graph which is why I need to log these setups as different runs.
As this is very unorganised and I would want to be able to in one scp -r call get all my tensorboard runs to my laptop.

Therefore I was wondering if there is a way in PL to log not directly in my version_0  folder but in version_0/tb_logs
I would want to be able to tell PL to save my logs in a subfolder tensorboard_logs as such that I only have two folders checkpoints  and tensorboard_logs in this version_0  directory.
I would have assumed that the log_dir  parameter of the TensorboardLogger would be this functionality, however in this case the checkpoints and logs are still together but now in a different main folder.

cc @borda @tchaton @justusschock @awaelchli @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy

janhenriklambrechts:
I propose a ```sub_dir``` argument for ```Logger``` objects. 

The default behaviour for multiple experiments and loggers right now looks something like this:

```
logs
└── default
    └── version_0
[...]

PR #9938, opened 2021-10-15
Support checkpoint save and load with Stochastic Weight Averaging

adamreeve:
## What does this PR do?

Partly addresses #6074 by supporting saving and loading the `StochasticWeightAveraging` callback data in checkpoints. Support for using SWA during validation will be done as a follow up PR.

No

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you list all the breaking changes introduced by this pull request?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [x] Check that all items from Before submitting are resolved

## Did you have fun?

- [x] Make sure you had fun coding 🙃


adamreeve:
There hasn't been any discussion on this approach on #6074 yet but I think this is at a point where it's ready for review.

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -198,68 +221,97 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
...
-    def reset_momenta(self):
-        """Adapted from https://github.com/pytorch/pytorch/blob/v1.7.1/torch/optim/swa_utils.py#L164-L165."""
-        for bn_module in self.momenta:
-            bn_module.momentum = self.momenta[bn_module]
+        # Recompute mean and variance for all batch norm layers by doing a full pass over the training data
+        for batch in data_loader:
+            if isinstance(batch, (list, tuple)):
+                batch = batch[0]
+            batch = batch.to(pl_module.device)
+            pl_module(batch)
```
justusschock, line 304: This seems to be pretty expensive to do. Especially if you do it every time you validate.

tests/callbacks/test_stochastic_weight_avg.py
```diff
@@ -63,22 +80,41 @@ def configure_optimizers(self):
...
+    validation_calls: int = 0
     update_parameters_calls: int = 0
     transfer_weights_calls: int = 0
 
     def update_parameters(self, *args, **kwargs):
         self.update_parameters_calls += 1
         return StochasticWeightAveraging.update_parameters(*args, **kwargs)
 
+    def on_validation_start(self, *args, **kwargs):
+        self.validation_calls += 1
```
justusschock, line 109: This doesn't check that the averaged model was used in validation, right? Could you maybe add a test for this?

justusschock (commented):
Looks good to me in general. Some concerns about looping over all training samples again for each validation.

In general this PR is quite big. Does it make sense to split it up into 2 PRs: one enabling only the checkpointing and a follow up enabling to use the averaged model in validation?

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: Yeah this isn't ideal. It's only done during SWA epochs and when batch normalization is used, but I'd guess that's fairly common. It seems like the right thing to do from a correctness point of view, but possibly some users would be OK with keeping the batch normalization parameters from training as they shouldn't be massively different, so maybe this behaviour could be configurable?

adamreeve:
Thanks @justusschock, yes it probably does make sense to split this up. I will revert the validation related changes from here and just keep the checkpointing changes, then follow up with another PR.

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I've reverted the validation changes so this can be addressed in a follow up PR.

tests/callbacks/test_stochastic_weight_avg.py
adamreeve: Yes I'll add this in the follow up PR for validation now that this is reverted.

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -182,6 +202,9 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
                 anneal_strategy=self._annealing_strategy,
                 last_epoch=trainer.max_epochs if self._annealing_strategy == "cos" else -1,
             )
+            if self._scheduler_step_count is not None:
+                # Restore scheduler step count from checkpoint
+                self._swa_scheduler._step_count = self._scheduler_step_count
```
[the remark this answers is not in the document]
tchaton: Yes, I don't see any other option either.

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -162,7 +167,10 @@ def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
             trainer.fit_loop.max_epochs += 1
 
     def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
-        if trainer.current_epoch == self.swa_start:
+        resuming_after_start = (not self._initialized) and (self.swa_start < trainer.current_epoch <= self.swa_end)
+        if trainer.current_epoch == self.swa_start or resuming_after_start:
```
tchaton, line 171:
```suggestion
        if (not self._initialized) and (self.swa_start <= trainer.current_epoch <= self.swa_end):
```

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -115,6 +115,7 @@ def __init__(
         if device is not None and not isinstance(device, (torch.device, str)):
             raise MisconfigurationException(f"device is expected to be a torch.device or a str. Found {device}")
 
+        self.n_averaged = None
```
tchaton, line 118:
```suggestion
        self.n_averaged: Optional[torch.Tensor] = None
```

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +291,44 @@ def avg_fn(
...
+
+    def on_load_checkpoint(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", callback_state: Dict[str, Any]
+    ) -> None:
+        if callback_state:
+            self.n_averaged = callback_state["n_averaged"]
+            self._scheduler_step_count = callback_state["scheduler_step_count"]
+            self._load_average_model_parameters(callback_state["average_model_parameters"])
+        else:
+            rank_zero_warn(
```
tchaton, line 313: I believe this should raise a MisConfigurationError. Do you see a use-case where a user would drop the callback state with SWA ?

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +291,44 @@ def avg_fn(
...
+            )
+
+    def _get_average_model_parameters(self, trainer: "pl.Trainer") -> Any:
+        if self._average_model is None or not (self.swa_start <= trainer.current_epoch <= self.swa_end):
+            # If we're not within the SWA epochs then when loading checkpoint data we would want
+            # to use parameters from the underlying model rather than the SWA parameters.
+            return None
+        parameters = []
+        for p_swa in self._average_model.parameters():
+            parameters.append(p_swa.detach())
```
tchaton, line 324: Do we need the detach here ?

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +291,44 @@ def avg_fn(
...
+        if callback_state:
+            self.n_averaged = callback_state["n_averaged"]
+            self._scheduler_step_count = callback_state["scheduler_step_count"]
+            self._load_average_model_parameters(callback_state["average_model_parameters"])
+        else:
+            rank_zero_warn(
+                f"Checkpoint has no data for the {self.state_key} callback, not initializing the callback state."
+            )
+
+    def _get_average_model_parameters(self, trainer: "pl.Trainer") -> Any:
```
tchaton, line 317: Might be better to rely on the training_type_plugin to reload the weights. This won't work with sharded models.

tests/callbacks/test_stochastic_weight_avg.py
```diff
@@ -62,8 +69,23 @@ def configure_optimizers(self):
...
+        if not self.crash_after_epoch:
+            return
+        self._epoch_count += 1
+        if self._epoch_count >= self.crash_after_epoch:
+            raise RuntimeError("Crash test")
+
 
 class SwaTestCallback(StochasticWeightAveraging):
+    def __init__(self, *args, **kwargs):
+        if "resuming_from_epoch" in kwargs:
```
tchaton, line 82: Do we need `resuming_from_epoch `. Can't we use trainer.current_epoch ?

tchaton:
Hey @adamreeve,

Great progress :)

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I had imagined this might happen if upgrading the version of PyTorch lightning and then resuming from a checkpoint saved with a previous version. Although I'm not sure that's something that's supposed to be supported so I'd be happy to changes this to an exception.

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: Fixed, plus I added annotations to some of the other members below too

tests/callbacks/test_stochastic_weight_avg.py
adamreeve: Yes you're right, I've simplified this to use `trainer.current_epoch`

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: No we don't, I've simplified this and also removed an extra detach in `_load_average_model_parameters` that also wasn't needed.

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I had a go at setting up a unit test with a sharded model following https://pytorch-lightning.readthedocs.io/en/stable/advanced/advanced_gpu.html, setting `strategy='fsdp'` with 2 gpus and defining `configure_sharded_model` on `SwaTestModel`. But the test crashes in `StochasticWeightAveraging.on_fit_start` with ``MisconfigurationException: SWA currently works with 1 `optimizer`.`` When I check the number of optimizers at this point it's actually zero. Does this mean SWA already doesn't work with sharded models or am I doing something wrong?

I'd be happy to change to using the `training_type_plugin` here regardless, but I'm struggling to see how to use it to save and load the weights.

This did also prompt me to test with multiple processes and I found there was a bug in how I handled `n_averaged` as after resuming, each process was using the same instance of `n_averaged`, resulting in it being incremented too many times, so I've fixed that.

adamreeve: If I remove the check for 1 optimizer in `on_fit_start`, then by the time we hit `on_train_epoch_start` the number of optimizers is 1 and things work up until calling `update_parameters`, which fails with:
```
        for p_swa, p_model in zip(average_model.parameters(), model.parameters()):
            device = p_swa.device
            p_swa_ = p_swa.detach()
            p_model_ = p_model.detach().to(device)
            src = p_model_ if n_averaged == 0 else avg_fn(p_swa_, p_model_, n_averaged.to(device))
>           p_swa_.copy_(src)
E           RuntimeError: The size of tensor a (32) must match the size of tensor b (528) at non-singleton dimension 1
```
so it looks like there's already work to do elsewhere to make SWA work with sharded models.

adamreeve:
Thanks @tchaton, I've addressed a few of your comments but still have some questions on a couple where I've left a response. As far as I can tell, SWA doesn't currently work with sharded models, so I would think that was outside the scope of this PR to support.

pytorch_lightning/callbacks/stochastic_weight_avg.py
tchaton: I believe it would be safer to use a `MisConfigurationError` and re-investigate this later on.

pytorch_lightning/callbacks/stochastic_weight_avg.py
tchaton: Yes, it isn't supported and might be complex to do so.

I believe it would safer to raise a MisConfigurationExecption if this Callback is used with DeepSpeed, DDPSharded, or FSDP.

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I tried testing this but realised this method will never be called if the state isn't present in the checkpoint as the trainer already checks that the state is truthy before calling `on_load_checkpoint`: https://github.com/PyTorchLightning/pytorch-lightning/blob/8bf7f9cce744bbc05ebce03bfd0567e3bc414128/pytorch_lightning/trainer/callback_hook.py#L290-L292
So I've just removed the check here.

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: OK thanks, I've added a check for these now and raise a MisconfigurationException

adamreeve:
Thanks for the follow up @tchaton, I think the comments so far have all been addressed now. There's just one lot of tests that is consistently timing out and being cancelled (macOS-10.15, 3.7, latest, stable), but it looks like that's happening on other PRs too.

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -16,18 +16,25 @@
...
+from pytorch_lightning.plugins.training_type.deepspeed import DeepSpeedPlugin
+from pytorch_lightning.plugins.training_type.fully_sharded import DDPFullyShardedPlugin
 from pytorch_lightning.trainer.optimizers import _get_default_scheduler_config
 from pytorch_lightning.utilities import rank_zero_info, rank_zero_warn
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 
+try:
+    from pytorch_lightning.plugins.training_type.fully_sharded import FullyShardedDataParallel
+except ImportError:
+    FullyShardedDataParallel = None
```
awaelchli, line 36: You are importing the wrong thing. From the code below it seems you want to import the plugin. This can be done without checking for import error.

```suggestion
from pytorch_lightning.plugins.training_type import DDPFullyShardedPlugin
```

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -145,6 +158,12 @@ def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
         optimizers = trainer.optimizers
         lr_schedulers = trainer.lr_schedulers
 
+        sharded_plugins = [DDPFullyShardedPlugin, DeepSpeedPlugin]
+        if FullyShardedDataParallel:
+            sharded_plugins.append(FullyShardedDataParallel)
+        if isinstance(trainer.training_type_plugin, tuple(sharded_plugins)):
+            raise MisconfigurationException("SWA does not currently support sharded models.")
```
awaelchli, line 165: FullyShardedDataParallel is not a plugin :) 

```suggestion
        if isinstance(trainer.training_type_plugin, (DDPFullyShardedPlugin, DeepSpeedPlugin):
            raise MisconfigurationException("SWA does not currently support sharded models.")
```

You will also have to include these:
DDPShardedPlugin, DDPSpawnShardedPlugin

awaelchli (changes requested):

awaelchli:
> here's just one lot of tests that is consistently timing out and being cancelled (macOS-10.15, 3.7, latest, stable), but it looks like that's happening on other PRs too.

Yes, we are aware and working on it :)

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +304,35 @@ def avg_fn(
...
+
+    def on_save_checkpoint(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", checkpoint: Dict[str, Any]
+    ) -> dict:
+        checkpoint_data = {
+            "n_averaged": 0 if self.n_averaged is None else self.n_averaged.item(),
+            "scheduler_step_count": None if self._swa_scheduler is None else self._swa_scheduler._step_count,
+            "average_model_parameters": self._get_average_model_parameters(trainer),
+        }
+        return checkpoint_data
```
tchaton, line 316:
```suggestion
        return {
            "n_averaged": 0 if self.n_averaged is None else self.n_averaged.item(),
            "scheduler_step_count": None if self._swa_scheduler is None else self._swa_scheduler._step_count,
            "average_model_parameters": self._get_average_model_parameters(trainer),
        }
```

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: Whoops, thanks yes I'm already importing DDPFullyShardedPlugin above, not sure why I thought I needed this.

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I think DDPShardedPlugin and DDPSpawnShardedPlugin are actually OK. At least from my quick testing they seem to work with SWA. It's only when the model parameters are sharded that we get problems, but sharding the optimizer and gradients seems fine. DeepSpeed is a bit different as it looks like parameter sharding is optional so maybe it could be made to work fairly easily, but with the default deepspeed strategy the SWA callback raises an exception due to there not being 1 optimizer in on_fit_start.

adamreeve:
Hi @awaelchli, thanks for the review, I've addressed the issues you pointed out.

awaelchli:
@adamreeve The master branch of your fork is 417 commits behind. Could you update it, then merge master into this branch here? Our CI will not run on your branch otherwise.

CHANGELOG.md
```diff
@@ -239,6 +239,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed unnessesary `_move_optimizer_state` method overrides from `TPUSpawnPlugin` and `SingleTPUPlugin` ([#10849](https://github.com/PyTorchLightning/pytorch-lightning/pull/10849))
 
 
+- Fixed resuming from a checkpoint when using Stochastic Weight Averaging (SWA) ([#9938](https://github.com/PyTorchLightning/pytorch-lightning/pull/9938))
```
awaelchli, line 242: oops, probably a merge error :) 
this entry needs to move to the Fixed section

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +295,34 @@ def avg_fn(
...
+            "average_model_parameters": self._get_average_model_parameters(trainer),
+        }
+
+    def on_load_checkpoint(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", callback_state: Dict[str, Any]
+    ) -> None:
+        self._init_n_averaged = callback_state["n_averaged"]
+        self._scheduler_step_count = callback_state["scheduler_step_count"]
+        self._load_average_model_parameters(callback_state["average_model_parameters"])
+
```
awaelchli, line 314: since on_save and on_load_checkpoint were not implemented in older versions, it would be good to raise an error or provide a default behavior if the keys are no in the checkpoint (when loading a checkpoint that was saved with an older Lightning version). Right now, it would result in a `KeyError`. What are your thoughts here?

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +295,34 @@ def avg_fn(
...
+    def _get_average_model_parameters(self, trainer: "pl.Trainer") -> Optional[List[nn.Parameter]]:
+        if self._average_model is None or not (self.swa_start <= trainer.current_epoch <= self.swa_end):
+            # If we're not within the SWA epochs then when loading checkpoint data we would want
+            # to use parameters from the underlying model rather than the SWA parameters.
+            return None
```
awaelchli, line 319:
```suggestion
            return
```

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -280,3 +295,34 @@ def avg_fn(
...
+        return list(self._average_model.parameters())
+
+    def _load_average_model_parameters(self, parameter_state: Any):
```
awaelchli, line 322:
```suggestion
    def _load_average_model_parameters(self, parameter_state: Any) -> None:
```

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -145,6 +152,9 @@ def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
         optimizers = trainer.optimizers
         lr_schedulers = trainer.lr_schedulers
 
+        if isinstance(trainer.training_type_plugin, (DDPFullyShardedPlugin, DeepSpeedPlugin)):
+            raise MisconfigurationException("SWA does not currently support sharded models.")
```
awaelchli, line 156: @ananthsub @daniellepintz Here is an example where we would have implemented `on_init_end` to fail early instead of spawning processes and then error on fit start.

CHANGELOG.md
adamreeve: Whoops, yes this was originally under fixed, will move it back

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I believe this method will never be called if we load a checkpoint saved from an older version, as `TrainerCallbackHookMixin` checks that the callback data is truthy before calling this `on_load_checkpoint`:
https://github.com/PyTorchLightning/pytorch-lightning/blob/06c5903600b8ed288757a71a00d686120a5fd7a7/pytorch_lightning/trainer/callback_hook.py#L82-L86
(and it only saves something to the checkpoint if it gets a truthy value back from `on_save_checkpoint`).

So it doesn't seem like there's much that can be done from the callback's point of view to handle checkpoints from an older version without data, we'll just silently continue without realising we're restarting without any saved state. Maybe we could detect this based on `trainer.current_epoch` when starting a training epoch? As things are, if we restart after the SWA starting epoch without any checkpoint data, we'll now reinitialize SWA when we wouldn't have before, but `n_averaged` and the scheduler step count will start at zero.

adamreeve: @awaelchli, I've done some testing to verify this behaviour. If loading an old checkpoint without SWA data everything works fine if the checkpoint was from before the SWA start epoch. If loading after the start epoch then the behaviour will be a bit different than if checkpoint data was loaded but there isn't a crash. I think this is an improvement on what currently happens on master where SWA isn't reinitialized at all, but it might be nice to log a warning about this?

adamreeve: I've added that warning now. Happy to change the approach if you disagree though.

Looks like the release of PyTorch 1.10.1 has caused the build to start failing.

pytorch_lightning/callbacks/stochastic_weight_avg.py
daniellepintz: would using `setup` work?

tchaton (approved):
Great work ! LGTM !

adamreeve:
Hi @awaelchli, can you please take another look at this?

pytorch_lightning/callbacks/stochastic_weight_avg.py
awaelchli: thanks, looks good :)

awaelchli (approved):

awaelchli:
LGTM! First of all, thanks for the patience. We can resume here since our GPU CI is back in operation! 
One minor thing we may want to double check: 

The test `test_swa_resume_training_from_checkpoint_ddp` warns with the following message:

```
/Users/adrian/repositories/pytorch-lightning/pytorch_lightning/loops/fit_loop.py:197: UserWarning: You're resuming from a checkpoint that ended mid-epoch. Training will start from the beginning of the next epoch. This can cause unreliable results if further training is done, consider using an end of epoch checkpoint or use fault-tolerant training to restart as if training did not stop.
```

I'm not yet sure why this happens, the model checkpoint is configured to run on_validation_end in this test, and since we raise the error after that, we should be good imo. Will get back to you if I find out why

justusschock (approved):

Borda (approved):

tests/callbacks/test_stochastic_weight_avg.py
```diff
@@ -281,3 +307,89 @@ def on_train_epoch_start(self):
...
+        num_processes=num_processes,
+        strategy=strategy,
+    )
+
+    exception_type = Exception if ddp else RuntimeError
+    backward_patch = mock.patch.object(Strategy, "backward", wraps=trainer.strategy.backward)
+    with backward_patch, pytest.raises(exception_type):
+        trainer.fit(model)
+
+    checkpoint_dir = Path(tmpdir) / "lightning_logs" / "version_0" / "checkpoints"
```
rohitgr7, line 338:
```suggestion
    checkpoint_dir = trainer.checkpoint_callback.dirpath
```

rohitgr7 (changes requested):
sorry for blocking this. Just need to ensure if it fixes another similar issue: https://github.com/PyTorchLightning/pytorch-lightning/issues/6019

give me a day.

rohitgr7:
> LGTM! First of all, thanks for the patience. We can resume here since our GPU CI is back in operation! One minor thing we may want to double check:
> 
> The test `test_swa_resume_training_from_checkpoint_ddp` warns with the following message:
> 
> ```
> /Users/adrian/repositories/pytorch-lightning/pytorch_lightning/loops/fit_loop.py:197: UserWarning: You're resuming from a checkpoint that ended mid-epoch. Training will start from the beginning of the next epoch. This can cause unreliable results if further training is done, consider using an end of epoch checkpoint or use fault-tolerant training to restart as if training did not stop.
> ```
> 
> I'm not yet sure why this happens, the model checkpoint is configured to run on_validation_end in this test, and since we raise the error after that, we should be good imo. Will get back to you when I find out why

this was fixed here: https://github.com/PyTorchLightning/pytorch-lightning/pull/11556

tests/callbacks/test_stochastic_weight_avg.py
adamreeve: I tried making this change but this caused the `test_swa_resume_training_from_checkpoint_ddp` test to fail as for some reason `trainer.checkpoint_callback.dirpath` is None when using the `ddp_spawn_find_unused_parameters_false` strategy.

adamreeve: If I debug this, I can see `__resolve_ckpt_dir` being called on an instance of `ModelCheckpoint`, but this happens in a spawned process and the instance is different to the one referenced by `trainer.checkpoint_callback` in the test method.

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -159,7 +169,9 @@ def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
             trainer.fit_loop.max_epochs += 1
 
     def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
-        if trainer.current_epoch == self.swa_start:
+        if (not self._initialized) and (self.swa_start <= trainer.current_epoch <= self.swa_end):
```
rohitgr7, line 172: okay... as I anticipated the fix is not complete. When it resumes the training:
- the checkpoint contains the scheduler states as well.
- Now if in the previous run, SWALR gets configured, then the scheduler states, will be of SWALR.
- While resuming in such a condition, when we reload the scheduler states, SWALR is not configured by that point and SWALR scheduler states will be loaded inside another scheduler configured by the user explicitly which might reload the states sometimes (as it does in the test) but might not reload the state if some other incompatible scheduler is configured (as in https://github.com/PyTorchLightning/pytorch-lightning/issues/11665 issue)

need to think more on this.

tests/callbacks/test_stochastic_weight_avg.py
rohitgr7: hmm. weird

pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: I've pushed a commit with a unit test that reproduces this crash to a new branch: https://github.com/adamreeve/pytorch-lightning/commit/33f67eb7b918468e28b32805f0112e375d6f6dc5

Not sure yet how to handle this though. I've tried configuring the scheduler within `StochasticWeightAveraging.on_load_checkpoint` but that doesn't work as the optimizers haven't yet been configured at the point the callback state is restored, and that only happens right before restoring the scheduler state.

If we could get the correct SWALR scheduler configured before its checkpoint state is loaded that would probably mean I don't need my somewhat hacky saving and restoring of the scheduler's `_step_count`.

adamreeve: Okay I have a fix for this, although it's still on a separate branch for now: https://github.com/adamreeve/pytorch-lightning/commit/0107ff1205ed39509f0487ca5039899f64b36428

It feels a little bit flaky as it relies on the order in which checkpoint state is restored. The idea is to hide the scheduler config so that the CheckpointConnector doesn't try to restore state for them, then restore the configs with the scheduler state from the checkpoint when training restarts.

If this seems reasonable I can merge it into this PR.

adamreeve: I've merged those changes into this PR now

adamreeve:
Hi @carmocca, tests I've added in this PR have started failing since merging in your changes from #8578 and the new behaviour doesn't look correct to me, but I don't fully understand what's going on.

In the tests where I resume training after crashing, the number of total training batches processed in `on_train_end` is less than expected, as on the first epoch after resuming from a checkpoint, all training batches for the epoch are skipped.

This is because the `TrainingEpochLoop` returns True from `done()` immediately without any training batches actually running in the epoch. `_num_ready_batches_reached` is evaluating to True as `self.batch_progress.current.ready` is 5, which is equal to the number of training batches in these tests. But prior to merging your change, this was 0 after resuming from a checkpoint.

I think this is due to your change here, where you always save all progress loop state even if fault tolerant mode isn't enabled: https://github.com/PyTorchLightning/pytorch-lightning/pull/8578/files#diff-4076755145b42e55247c8c12a7a7961a91f4d28a3e962c80a98161163bfa3ec9L295

This new behaviour doesn't seem correct to me, as if we're restarting at the beginning of an epoch then shouldn't the current training epoch loop progress be reset to zero rather than restored to what it was at the end of the previous epoch when the checkpoint was saved?

carmocca:
@adamreeve I can see there's an issue but your hypothesis is not correct. I will look at what's the problem.

tests/callbacks/test_stochastic_weight_avg.py
```diff
@@ -61,10 +67,18 @@ def configure_optimizers(self):
             },
         }
 
+    def training_epoch_end(self, _):
+        if not self.crash_after_epoch:
+            return
+        if self.trainer.current_epoch + 1 >= self.crash_after_epoch:
+            raise RuntimeError("Crash test")
```
carmocca, line 74: We should use a dummy exception class just in case.

tests/callbacks/test_stochastic_weight_avg.py
```diff
@@ -61,10 +67,18 @@ def configure_optimizers(self):
             },
         }
 
+    def training_epoch_end(self, _):
+        if not self.crash_after_epoch:
+            return
+        if self.trainer.current_epoch + 1 >= self.crash_after_epoch:
```
carmocca, line 73: Note that raising an exception here is a bit silly because the epoch that raised it will be lost since checkpointing has not triggered yet.

The checkpoint loaded will be for the end of the previous epoch.

tests/callbacks/test_stochastic_weight_avg.py
```diff
@@ -38,17 +44,17 @@ def __init__(self, batchnorm: bool = True, interval: str = "epoch", iterable_dat
         self.layer = nn.Sequential(*layers)
         self.interval = interval
         self.iterable_dataset = iterable_dataset
+        self.crash_after_epoch = crash_after_epoch
+        self.save_hyperparameters()
```
carmocca, line 48: Is adding `save_hyperparameters` relevant?

tests/callbacks/test_stochastic_weight_avg.py
adamreeve: Yes probably makes sense to raise it on the epoch start rather than doing unnecessary work and raising it after the training epoch ends.

tests/callbacks/test_stochastic_weight_avg.py
adamreeve: No I don't remember why I put this in but it doesn't appear to be necessary now, I'll remove it.

pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -194,14 +225,14 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
             else:
                 trainer.lr_scheduler_configs.append(default_scheduler_cfg)
 
-            self.n_averaged = torch.tensor(0, dtype=torch.long, device=pl_module.device)
+            if self.n_averaged is None:
+                self.n_averaged = torch.tensor(self._init_n_averaged, dtype=torch.long, device=pl_module.device)
 
         if self.swa_start <= trainer.current_epoch <= self.swa_end:
             self.update_parameters(self._average_model, pl_module, self.n_averaged, self._avg_fn)
```
carmocca, line 232: this should check whether we already updated parameters for the current epoch as this hook can get called again for the same epoch when the trainer restarts.

adamreeve: How would that happen? Is that when restoring from a mid-epoch checkpoint?

carmocca: I believe doing that will fix `test_swa_resume_training_from_checkpoint[4]`

carmocca: That happens because when you save a checkpoint `on_train_epoch_end` (ModelCheckpoint default), the epoch has technically not finished. See description in https://github.com/PyTorchLightning/pytorch-lightning/pull/8578

So after it is reloaded, the FitLoop (loop that manages epochs) does one "dummy" iteration to get to the correct state. During that dummy iteration the `TrainingEpochLoop` is skipped as you observed above.

This has the side effect that epoch-level hooks get called again on restoration. So `update_parameters` would be getting called again for the previous epoch.

adamreeve: Ok thanks for the explanation, that makes sense. I'm not sure that matches up with what I was observing though. Eg. if I crash during epoch 3, then the checkpoint from epoch 2 is loaded and training restarts, but trainer.current_epoch for the first dummy iteration that doesn't do any training is 3, whereas if I'm understanding you correctly, it should be 2?

adamreeve: I've made the change you suggested here but as suspected, it doesn't actually fix the tests. When resuming from a checkpoint, the first "dummy" epoch number is the epoch after the checkpoint was saved, not the epoch from the checkpoint. So things look like this in the case where we crash on epoch number 3:
```
epoch 0: 5 batches
epoch 1: 5 batches
epoch 2: 5 batches
epoch 3: crash
epoch 3: 0 batches
epoch 4: 5 batches
epoch 5: 5 batches (extra epoch due to batchnorm)
```

So epoch 3 effectively never happens. This is because when loading the checkpoint data, `fit_loop.epoch_progress.current` in the checkpoint is `{'ready': 3, 'completed': 2, 'started': 3, 'processed': 3}`, so the setter for `FitLoop.restarting` overrides `completed` and sets it to 3. If I remove [that line](https://github.com/PyTorchLightning/pytorch-lightning/pull/8578/files#diff-a3f9154e7c08e7e6f2c703a4a483679193a2f4268cf1cf495d466a872e7cd5fdR145), then all my tests pass, but there are some other checkpoint related tests that start to fail (test_checkpoint_repeated_strategy_extended and test_trainer_model_hook_system_fit_no_val_and_resume).

The logic in that `restarting` setter seems to contradict the idea that there should be a dummy epoch with no batches though.

adamreeve:
Hi @rohitgr7 and @carmocca, I've pushed a change to the tests to work around the current behaviour when resuming from a checkpoint. If it's going to take a while to sort out the checkpoint resume behaviour it would be great to get this PR merged soon and I'm happy to help later if any further SWA changes are needed, but hopefully it will just be a matter of reverting the changes in that commit (d2ed468a3adea2e3e14e592fb604dae61ef35d0c).

I've also had to make a change related to #11576, as when resuming from a checkpoint an exception was being thrown due to the validation dataloader not being set. This only caused a problem when restarting, as `TrainingEpochLoop._is_validation_done` short-circuits on `return not self.restarting or self.val_loop.done`, and I guess no existing tests covered this scenario.

stackedsax:
@rohitgr7  and @carmocca any chance you could take a look at this?  It would be great to get this merged. Thanks!

pytorch_lightning/callbacks/stochastic_weight_avg.py
carmocca: Exploring the fix in https://github.com/PyTorchLightning/pytorch-lightning/pull/12429

adamreeve:
I've had to change how the LR scheduler configs are handled to work around the deprecation of `on_load_checkpoint`. The new approach is pretty hacky and it would be a lot simpler if only considering the case where `restore_checkpoint_after_setup` is `False`. Given `False` is the default and it looks like only deepspeed sets this to `True`, and SWA already doesn't work with deepspeed, it might make sense to drop support for handling the case when `restore_checkpoint_after_setup` is `True`?

adamreeve:
Hi @awaelchli and @rohitgr7, apologies for the delay but I've now made that change to remove the workaround for `restore_checkpoint_after_setup = True`.

adamreeve:
Hi @rohitgr7, can you please take another look at this? I think this is just blocked on your change request.

The current test failures look unrelated to my changes, I can see the same errors on other PRs.

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -273,3 +308,38 @@ def avg_fn(
...
+        # restored, and doesn't work if restore_checkpoint_after_setup is True, but at the time of
+        # writing that is only True for deepspeed which is already not supported by SWA.
+        # See https://github.com/PyTorchLightning/pytorch-lightning/issues/11665 for background.
+        if trainer.lr_scheduler_configs:
+            assert len(trainer.lr_scheduler_configs) == 1
+            self._scheduler_configs = list(trainer.strategy.lr_scheduler_configs)
+            trainer.lr_scheduler_configs.clear()
+
+    def _load_average_model_parameters(self, parameter_state: Any) -> None:
+        if self._average_model is None or parameter_state is None:
```
rohitgr7, line 340:
```suggestion
        if self._average_model is None:
```
just making some strong assertions to ensure bugs are caught if any in future.

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -273,3 +308,38 @@ def avg_fn(
     ) -> FloatTensor:
         """Adapted from https://github.com/pytorch/pytorch/blob/v1.7.1/torch/optim/swa_utils.py#L95-L97."""
         return averaged_model_parameter + (model_parameter - averaged_model_parameter) / (num_averaged + 1)
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {
+            "n_averaged": 0 if self.n_averaged is None else self.n_averaged.item(),
+            "latest_update_epoch": self._latest_update_epoch,
+            "scheduler_state": None if self._swa_scheduler is None else self._swa_scheduler.state_dict(),
+            "average_model_parameters": None if self._average_model is None else list(self._average_model.parameters()),
```
rohitgr7, line 317: what if we save the `self._average_model.state_dict` instead?
have you tried?

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -273,3 +308,38 @@ def avg_fn(
...
+        # load state into the wrong type of schedulers when restoring scheduler checkpoint state.
+        # We'll configure the scheduler and re-load its state in on_train_epoch_start.
+        # Note that this relies on the callback state being restored before the scheduler state is
+        # restored, and doesn't work if restore_checkpoint_after_setup is True, but at the time of
+        # writing that is only True for deepspeed which is already not supported by SWA.
+        # See https://github.com/PyTorchLightning/pytorch-lightning/issues/11665 for background.
+        if trainer.lr_scheduler_configs:
+            assert len(trainer.lr_scheduler_configs) == 1
+            self._scheduler_configs = list(trainer.strategy.lr_scheduler_configs)
+            trainer.lr_scheduler_configs.clear()
```
rohitgr7, line 337: Neat!

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
```diff
@@ -255,3 +281,103 @@ def on_train_epoch_start(self):
...
+        "strategy": "ddp_spawn_find_unused_parameters_false" if ddp else None,
+        "devices": 2 if ddp else 1,
+        "limit_train_batches": 5,
+        "limit_val_batches": 0,
+        "accumulate_grad_batches": 2,
+        "enable_progress_bar": False,
+    }
+    trainer = Trainer(callbacks=SwaTestCallback(swa_epoch_start=swa_start, swa_lrs=0.1), **trainer_kwargs)
+
+    with _backward_patch(trainer), pytest.raises(Exception if ddp else DummyError):
```
rohitgr7, line 301: you can raise Exception(msg) instead and match it with the error msg
```python
pytest.raises(..., match='Crash test')
```

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
```diff
@@ -255,3 +281,103 @@ def on_train_epoch_start(self):
...
+        trainer.fit(model)
+
+    checkpoint_dir = Path(tmpdir) / "lightning_logs" / "version_0" / "checkpoints"
```
rohitgr7, line 304: is this correct?
I remember it should be just `Path(tmpdir) / "checkpoints"`

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
```diff
@@ -255,3 +281,103 @@ def on_train_epoch_start(self):
...
+def test_misconfiguration_error_with_ddp_fully_sharded(tmpdir):
+    _test_misconfiguration_error_with_sharded_model(tmpdir, "fsdp")
+
+
+@RunIf(deepspeed=True, min_cuda_gpus=1)
+def test_misconfiguration_error_with_deep_speed(tmpdir):
+    _test_misconfiguration_error_with_sharded_model(tmpdir, "deepspeed")
+
+
+def _backward_patch(trainer: Trainer) -> ContextManager:
```
rohitgr7, line 382: why it needs to patch the backward method?

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
```diff
@@ -255,3 +281,103 @@ def on_train_epoch_start(self):
...
+
+
+@RunIf(fairscale_fully_sharded=True, min_cuda_gpus=1)
+def test_misconfiguration_error_with_ddp_fully_sharded(tmpdir):
+    _test_misconfiguration_error_with_sharded_model(tmpdir, "fsdp")
+
+
+@RunIf(deepspeed=True, min_cuda_gpus=1)
+def test_misconfiguration_error_with_deep_speed(tmpdir):
+    _test_misconfiguration_error_with_sharded_model(tmpdir, "deepspeed")
```
rohitgr7, line 379: can parametrize instead?

rohitgr7 (commented):
apologies for the delay here!
looks good. Just have a few more questions and need to review new tests again.

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
```diff
@@ -255,3 +281,103 @@ def on_train_epoch_start(self):
...
+    # Reproduces the bug reported in https://github.com/PyTorchLightning/pytorch-lightning/issues/11665
+    model = CustomSchedulerModel(crash_on_epoch=crash_on_epoch)
+    resume_model = CustomSchedulerModel()
+    _swa_resume_training_from_checkpoint(tmpdir, model, resume_model)
+
+
+@RunIf(skip_windows=True, min_torch="1.8")
+def test_swa_resume_training_from_checkpoint_ddp(tmpdir):
+    # Requires PyTorch >= 1.8 to include this segfault fix:
+    # https://github.com/pytorch/pytorch/pull/50998
```
carmocca, line 350: We don't test 1.8 anymore


```suggestion
@RunIf(skip_windows=True)
def test_swa_resume_training_from_checkpoint_ddp(tmpdir):
```

adamreeve:
Thanks for the updated feedback, I should have some time next week to follow up on this.

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
adamreeve: I think these weren't parameterized due to the different `RunIf` parameters (`deepspeed=True` vs `fairscale_fully_sharded=True`), so as far as I know they need to be separate tests

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
adamreeve: This was already patched previously, I've just pulled it out into a function for reuse. The patch is needed because we check the number of times backward is called in `SwaTestCallback.on_train_end`, to make sure it isn't called for the batchnorm update epoch.

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -174,10 +191,25 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
...
+                    "This may be caused by loading a checkpoint from an older version of PyTorch Lightning."
+                )
+
             # We assert that there is only one optimizer on fit start, so know opt_idx is always 0
             default_scheduler_cfg = LRSchedulerConfig(self._swa_scheduler, opt_idx=0)
             assert default_scheduler_cfg.interval == "epoch" and default_scheduler_cfg.frequency == 1
 
+            if self._scheduler_configs:
+                trainer.lr_scheduler_configs[:] = self._scheduler_configs
+                self._scheduler_configs = None
```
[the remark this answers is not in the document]
adamreeve: Good point, it looks like we only use the original schedulers to verify the interval and frequency, but if we're restoring from a snapshot then we can assume they were previously verified. I'll remove this.

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
rohitgr7: you can do it like this: https://github.com/Lightning-AI/lightning/blob/eb233ea12dacf010edda49470396e85707d8c00e/tests/tests_pytorch/trainer/test_trainer.py#L1073

adamreeve: Neat, thanks!

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
adamreeve: Good idea, no I hadn't thought of that. That's much cleaner and does seem to work fine.

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
adamreeve: Yeah that's probably better. I was previously raising a `RuntimeError` but Carlos asked to have this changed to a custom dummy exception class to make sure we can only catch the specific exception we're expecting. After adding tests with DDP I ran into the problem that the exceptions that are thrown in subprocesses get wrapped in a plain `Exception` object though, so the type is different. Just always raising an `Exception` with a message to match on seems reasonable to simplify this.

tests/tests_pytorch/callbacks/test_stochastic_weight_avg.py
adamreeve: Yes this does seem to be correct. This matches `trainer.checkpoint_callback.dirpath`, except that is None when using DDP so I can't use that property and have to specify the path here.

awaelchli: Yes that's correct, dirpath wouldn't be available, so this here looks fine to me.

FYI, `checkpoint_callback.best_model_path` would however be available, as we specifically restore it when returning to the main process. You can try it to save a few lines but I think the current explicit way is good for the test!

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -46,6 +46,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 -
 
 
+## [1.8.0] - YYYY-MM-DD
```
carmocca, line 49: This was already added in https://github.com/Lightning-AI/lightning/pull/13982

carmocca (approved):

rohitgr7 (approved):
Perfect!
Great work!

Apologies for the delayed review! but it was complex and you pulled it out smartly 😃 🚀

awaelchli:
@rohitgr7 Will this go into 1.7.x?

rohitgr7:
@awaelchli yes

kampelmuehler:
@adamreeve 
Am I correct that the "follow up PR" on using the `_average_model` for validation hasn't yet been done?

kampelmuehler:
@adamreeve 
I think that these features should also somehow be documented. Currently the docs include nothing on how to actually use the SWA model (or that indeed only in the end of training the weights are transferred to the original model). What do you think?

kampelmuehler:
And another also:
The checkpointing is not in any sense automated, or is it?
Only way I see thus far is to manually load the checkpoint, get the `state_dict` for swa and load it manually into the callback instance.

If I just declare the model, add an swa callback to the trainer and pass the checkpoint path to the trainer it won't load the swa callback states.

kampelmuehler:
cc @awaelchli

adamreeve:
Hi @kampelmuehler 

> Am I correct that the "follow up PR" on using the _average_model for validation hasn't yet been done?

Yes, that was originally part of this PR but the scope was reduced, so #6074 probably shouldn't have been closed. I had started working towards that in a separate branch (https://github.com/adamreeve/pytorch-lightning/commits/swa_validation) but that's now quite far behind master and doesn't include all the changes from this PR. I don't currently have any plans to continue with that work.

This PR didn't really add any new user-visible feature but just fixed checkpointing to work correctly with SWA so that training could be resumed. Eg. it fixed #11665. But I agree that the documentation could be improved to better explain how SWA works.

It's a while since I looked at this now so I'm not sure whether `trainer.predict()` not loading the SWA parameters is intentional or just a limitation of the current approach, but it sounds consistent with the behaviour of the averaged parameters not being transferred until training is completed.

kampelmuehler:
Hi @adamreeve - thanks for the quick response and all the insights!

zhong-yy:
Hi, is swa_validation removed in the latest version?

PR #9987, opened 2021-10-18
Lightning Lite Examples

awaelchli:
## What does this PR do?

This is the V1 for the new Lightning Lite package. It bundles all major changes together, but individual PRs will be done to get it merged (e.g., #10175, #10176)
Planned to be released as part of 1.5.

### Demo
Lightning Lite Demo

### TODOs

- [x] Precision support
    - [x] NativeMixedPrecision
    - [x] Delay support ApexMixedPrecision
    - [x] TPUBfloat (verify it works)
    - [x] Double 

- [x] Plugin support
    - [x] DeepSpeed
    - [x] DeepSpeed with multiple models (@tchaton, @awaelchli)
    - [x] TPUSpawn: DDPSpawn changes need to be carried down into this plugin. A minor refactor could be merged to PL master. (@kaushikb11)
    - [x] Sharded (optimizer sharding)
    - [x]  Delay support FullySharded (model sharding)
    
- [x] Move data to device automatically
- [x] Move model to device automatically
- [x] Allow only one model per setup() call
- [x] DataLoader setup: Currently, there is no distributed sampler. 
- [x] Resolve miscellaneous TODOs in the code base
- [x] Fix changes that broke Lightning tests
- [x] Make self.setup() take  model and optimizers positionally.
- [x] Unit testing, parity tests
- [x] Typing (mypy)

### Discussions

- [x] LightningLite constructor arguments: We are currently changing the Trainer constructor arguments to support a new pattern: `Trainer(accelerator="cpu/tpu/gpu", strategy="ddp/deepspeed/...", devices=X)` Should we start promoting this directly in the LightningLite API? 

- [x] Deepspeed API for backward: The user can't call `loss.backward()`, it needs to be called on the model. Which API do we want to offer: 
    ```py
    A) self.backward(loss, model) # model is optional for plugins other than deepspeed
    B) model.backward(loss)
    ```
    In both cases, user needs to change their code if they switch from one strategy to the next. 

- [x] Deepspeed API for optimization step: Plain Deepspeed requires a call to `model.step()` as opposed to the usual `optimizer.step()`. Since we wrap the optimizers of the user anyway, we could still offer `optimizer.step()` and redirect to `model.step()`. It would mean less code changes for the user switching between plugins, but for deepspeed users it might be confusing! 

Related work:

- We are currently also [redesigning the Precision/TrainingType plugin interactions](https://docs.google.com/document/d/1E5t8auWf5DrNHzutvMmrJC_KqBqVyZuYX0thra69Ad8/edit) in collaboration with FB.


aniketmaurya:
in the documentation, I see it is recommended to initialize the model inside the `run` method to avoid OOM.

Will it be possible to initialize the model in `__init__` method and not worry about OOM?

I want to do something like this - 

```python
class BaseModel(LightningLite):
    def __init__(self):
        self.model: nn.Module = Model()
        super().__init__()
    
    def run(self):
        "train model"
```


> We recommend instantiating the models within the run() method as large models would cause OOM Error otherwise.

awaelchli:
Hey @aniketmaurya 
Thanks for checking out the Lite docs and examples. 

To clarify, the OOM remark is about sharded models. If the model fits on the GPU anyway then there is no problem. 
Another point is, the user is not meant to define their Lite class as a model. Lite is closer to a Trainer than a model. If you do this: 

```py
class BaseModel(LightningLite):
    def __init__(self):
        self.model: nn.Module = Model()
        super().__init__()
    
    def run(self):
        "train model"
```

Then we are losing all features of Lite, because `BaseModel(accelerator="ddp", devices=2).run()` will not make sense anymore. What we intend users to do is this:

```py
class Lite(LightningLite):
    # not override init
    
    def run(self, model):
        "train model"
```
or

```py
class Lite(LightningLite):
    # not override init
    
    def run(self, *parameters):
        model = Model(*parameters)
        "train model"
```

aniketmaurya:
hi @awaelchli, thanks for the explanation. I have more doubt, is it possible to utilize the features of Lite with this syntax? Since I am passing all the configs required for `Lite` class then it should be able to instantiate the required components, right?

```python
class BaseModel(LightningLite):
    def __init__(self, lite_configs={"accelerator":"ddp"}):
        self.model: nn.Module = Model()
        super().__init__(**lite_configs)
    
    def run(self):
        "train model"
```

tchaton:
Hey @aniketmaurya,

Yes, you could do something like this.

The only side note is: If you train massive models, this would fail, better to instantiate the model under the run function.

```py
class BaseModel(LightningLite):
    def __init__(self, *args, model_config = None, **kwargs):
        super().__init__(*args, **kwargs)

        self.model = from_config(model_config)

    def run(self):
        "train model"
```

.gitignore
```diff
@@ -137,6 +137,11 @@ ENV/
 Datasets/
 mnist/
 legacy/checkpoints/
+*.gz
+*ubyte
+grid_generated*
+grid_ori*
```
kaushikb11, line 143: I don't think this is of use for the Users.

pl_examples/basic_examples/mnist_examples/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+# cpu
+python image_classifier_5_lightning_datamodule.py
+
+# gpus (any number)
+python image_classifier_5_lightning_datamodule.py --trainer.gpus 2
+
+# dataparallel
+python image_classifier_5_lightning_datamodule.py --trainer.gpus 2 --trainer.accelerator 'dp'
```
kaushikb11, line 66:
```suggestion
# distributed data parallel
python image_classifier_5_lightning_datamodule.py --trainer.gpus 2 --trainer.strategy 'ddp'
```

pl_examples/basic_examples/mnist_examples/image_classifier_2_lite.py
```diff
@@ -0,0 +1,157 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Here are 5 required steps to convert to LightningLite.
+
+1. Subclass LightningLite and override its run method.
```
kaushikb11, line 17:
```suggestion
1. Subclass ``LightningLite`` and override it's ``run`` method.
```

tests/special_tests.sh
```diff
@@ -88,8 +88,8 @@ fi
 
 # test that a user can manually launch individual processes
 args="--trainer.gpus 2 --trainer.accelerator ddp --trainer.max_epochs=1 --trainer.limit_train_batches=1 --trainer.limit_val_batches=1 --trainer.limit_test_batches=1"
```
kaushikb11, line 90:
```suggestion
args="--trainer.gpus 2 --trainer.strategy ddp --trainer.max_epochs=1 --trainer.limit_train_batches=1 --trainer.limit_val_batches=1 --trainer.limit_test_batches=1"
```

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
+python image_classifier_5_lightning_datamodule.py
 
 # gpus (any number)
-python autoencoder.py --trainer.gpus 2
+python image_classifier_5_lightning_datamodule.py --trainer.gpus 2
 
-# Distributed Data Parallel
-python autoencoder.py --trainer.gpus 2 --trainer.accelerator ddp
+# dataparallel
+python image_classifier_5_lightning_datamodule.py --trainer.gpus 2 --trainer.accelerator 'dp'
```
kaushikb11, line 70:
```suggestion
# Distributed Data parallel
python image_classifier_5_lightning_datamodule.py --trainer.gpus 2 --trainer.strategy 'ddp'
```

pl_examples/basic_examples/mnist_examples/image_classifier_2_lite.py
awaelchli: it's is a contraction for it is
so the grammar is not correct. 
Also, here LightningLite is considered a name, like PyTorch, so it is fine to not specially format it.

kaushikb11 (approved):
LGTM! 🥳

rohitgr7 (approved):

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
 
 Use these examples to test how Lightning works.
 
-#### MNIST
+## MNIST Examples
 
-Trains MNIST where the model is defined inside the `LightningModule`.
+Here are 5 MNIST examples showing you how to gradually convert from pure PyTorch to PyTorch Lightning.
+
+The transition through [LightningLite](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.rst) from pure PyTorch is optional but it might be helpful to learn about it.
```
Borda, line 9:
```suggestion
The transition through [LightningLite](https://pytorch-lightning.readthedocs.io/en/stable/starter/lightning_lite.rst) from pure PyTorch is optional but it might be helpful to learn about it.
```

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
+
+#### 1. Image Classifier with Vanilla PyTorch
+
+Trains a simple CNN over MNIST using vanilla PyTorch.
 
 ```bash
 # cpu
-python simple_image_classifier.py
+python image_classifier_1_pytorch.py
```
Borda, line 17: what is the number? could be called it case/chapter?
```suggestion
python image_classifier_case-1_pytorch.py
```

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
+#### 2. Image Classifier with LightningLite
 
-# Distributed Data Parallel
-python simple_image_classifier.py --trainer.gpus 2 --trainer.accelerator ddp
+This script shows you how to scale the previous script to enable GPU and multi-GPU training using [LightningLite](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html).
+
+```bash
+# CPU / multiple GPUs if available
+python image_classifier_2_lite.py
 ```
```
Borda, line 29: do we need this call while it is args free so just refer to the script?
also without cloning/forking this repo they won't have these files...

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
 ```bash
-python dali_image_classifier.py
+# CPU / multiple GPUs if available
+python image_classifier_3_lite_to_lightning.py
 ```
 
-#### Image classifier
+#### 4. Image Classifier with LightningModule
```
Borda, line 44: the above already say:

> LightningLite to `LightningModule`.

```suggestion
#### 4. Image Classifier with Lightning
```

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
+
+```bash
+# CPU / multiple GPUs if available
+python image_classifier_2_lite.py
 ```
 
-#### MNIST with DALI
+#### 3. Image Classifier - Conversion from Lite to Lightning
```
Borda, line 33:
```suggestion
#### 3. Image Classifier - Introducing LightningModule powered by Lite
```

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
+python image_classifier_3_lite_to_lightning.py
 ```
 
-#### Image classifier
+#### 4. Image Classifier with LightningModule
 
-Generic image classifier with an arbitrary backbone (ie: a simple system)
+This script shows you the result of the conversion to the `LightningModule` and finally all the benefits you get from Lightning.
```
Borda, line 46:
```suggestion
This script shows you the result of the conversion to the `LightningModule` and finally all the benefits you get from Lightning ecosystem.
```

pl_examples/basic_examples/README.md
```diff
@@ -2,62 +2,70 @@
...
+python image_classifier_5_lightning_datamodule.py
 
-# gpus (any number)
-python autoencoder.py --trainer.gpus 2
+# GPUs (any number)
+python image_classifier_5_lightning_datamodule.py --trainer.gpus 2
 
-# Distributed Data Parallel
-python autoencoder.py --trainer.gpus 2 --trainer.accelerator ddp
+# Distributed Data parallel
```
Borda, line 69:
```suggestion
# Distributed Data Parallel (DDP)
```

pl_examples/basic_examples/mnist_examples/README.md
```diff
@@ -0,0 +1,67 @@
...
+The transition through [LightningLite](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.rst) from pure PyTorch is optional but it might be helpful to learn about it.
+
+#### 1. Image Classifier with Vanilla PyTorch
+
+Trains a simple CNN over MNIST using vanilla PyTorch.
+
+```bash
+# cpu
+python image_classifier_1_pytorch.py
+```
```
Borda, line 14: can we rather include the code as @ethanwharris does in Flash?

pl_examples/basic_examples/mnist_examples/README.md
```diff
@@ -0,0 +1,67 @@
+## MNIST Examples
```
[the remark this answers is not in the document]
Borda: what is the difference to the other readme, seems to me as identical...

pl_examples/basic_examples/mnist_examples/image_classifier_1_pytorch.py
```diff
@@ -0,0 +1,154 @@
...
+
+class Net(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv1 = nn.Conv2d(1, 32, 3, 1)
+        self.conv2 = nn.Conv2d(32, 64, 3, 1)
+        self.dropout1 = nn.Dropout(0.25)
+        self.dropout2 = nn.Dropout(0.5)
+        self.fc1 = nn.Linear(9216, 128)
+        self.fc2 = nn.Linear(128, 10)
```
Borda, line 37: can we simplify this with a wrap to sequence, for the later froward is simpler...

pl_examples/basic_examples/mnist_examples/image_classifier_1_pytorch.py
```diff
@@ -0,0 +1,154 @@
...
+        for batch_idx, (data, target) in enumerate(train_loader):
+            data, target = data.to(device), target.to(device)
+            optimizer.zero_grad()
+            output = model(data)
+            loss = F.nll_loss(output, target)
+            loss.backward()
+            optimizer.step()
+            if (batch_idx == 0) or ((batch_idx + 1) % hparams.log_interval == 0):
+                print(
+                    "Train Epoch: {} [{}/{} ({:.0f}%)]\tLoss: {:.6f}".format(
```
Borda, line 90: rather f-string?

pl_examples/basic_examples/mnist_examples/image_classifier_1_pytorch.py
```diff
@@ -0,0 +1,154 @@
...
+
+        if hparams.dry_run:
+            break
+
+    if hparams.save_model:
+        torch.save(model.state_dict(), "mnist_cnn.pt")
+
+
+def main():
+    parser = argparse.ArgumentParser(description="PyTorch MNIST Example")
```
Borda, line 132: I may consider to user (only for examples) Fire or jsonparser as CLI

pl_examples/basic_examples/mnist_examples/image_classifier_2_lite.py
```diff
@@ -0,0 +1,157 @@
...
+from torchmetrics.classification import Accuracy
+
+from pl_examples.basic_examples.mnist_datamodule import MNIST
+from pl_examples.basic_examples.mnist_examples.image_classifier_1_pytorch import Net
+from pytorch_lightning import seed_everything
+from pytorch_lightning.lite import LightningLite  # import LightningLite
+
+
+class Lite(LightningLite):
+    def run(self, hparams):
```
Borda, line 47: add typing? and why not as
```suggestion
    def run(self, **hparams):
```

pl_examples/basic_examples/mnist_examples/image_classifier_3_lite_to_lightning.py
```diff
@@ -0,0 +1,168 @@
...
+    # Methods for the `LightningModule` conversion
+
+    def forward(self, x):
+        return self.model(x)
+
+    def training_step(self, batch, batch_idx):
+        """Here you compute and return the training loss and compute extra training metrics."""
+        x, y = batch
+        logits = self.forward(x)
+        loss = F.nll_loss(logits, y.long())
```
Borda, line 115: train does not have accuracy?

pl_examples/basic_examples/mnist_examples/image_classifier_3_lite_to_lightning.py
```diff
@@ -0,0 +1,168 @@
...
+        optimizer = optim.Adadelta(self.model.parameters(), lr=self.hparams.lr)
+        return [optimizer], [StepLR(optimizer, step_size=1, gamma=self.hparams.gamma)]
+
+    # Methods for the `LightningDataModule` conversion
+
+    @property
+    def transform(self):
+        return T.Compose([T.ToTensor(), T.Normalize((0.1307,), (0.3081,))])
+
+    def prepare_data(self) -> None:
```
Borda, line 136:
```suggestion
    @staticmethod
    def prepare_data() -> None:
```

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
```diff
@@ -0,0 +1,85 @@
+# Copyright The PyTorch Lightning team.
```
Borda, line 1: rename as `_4_lightning_module`

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
```diff
@@ -0,0 +1,85 @@
...
+        return loss
+
+    def configure_optimizers(self):
+        optimizer = torch.optim.Adadelta(self.model.parameters(), lr=self.hparams.lr)
+        return [optimizer], [torch.optim.lr_scheduler.StepLR(optimizer, step_size=1, gamma=self.hparams.gamma)]
+
+    # Methods for the `LightningDataModule` conversion
+
+    @property
+    def transform(self):
```
Borda, line 61: I would keep it in Lite as it won't stay in LM and will be moved again to DM

pl_examples/loop_examples/mnist_lite.py
```diff
@@ -0,0 +1,186 @@
...
+
+from pl_examples.basic_examples.mnist_datamodule import MNIST
+from pl_examples.basic_examples.mnist_examples.image_classifier_1_pytorch import Net
+from pytorch_lightning import seed_everything
+from pytorch_lightning.lite import LightningLite
+from pytorch_lightning.loops import Loop
+
+
+class TrainLoop(Loop):
+    def __init__(self, lite, args, model, optimizer, scheduler, dataloader):
```
Borda, line 32: typing?

pl_examples/loop_examples/mnist_lite.py
```diff
@@ -0,0 +1,186 @@
...
+        scheduler = StepLR(optimizer, step_size=1, gamma=hparams.gamma)
+
+        MainLoop(self, hparams, model, optimizer, scheduler, train_loader, test_loader).run()
+
+        if hparams.save_model:
+            self.save(model.state_dict(), "mnist_cnn.pt")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="LightningLite MNIST Example with Lightning Loops.")
```
Borda, line 162: rather `LightningCLI`?

pl_examples/run_examples.sh
```diff
@@ -11,6 +11,23 @@ args="
...
+
+args="
+  --trainer.max_epochs=1
+  --trainer.limit_train_batches=2
+  --trainer.limit_val_batches=2
+  --trainer.limit_test_batches=2
+  --trainer.limit_predict_batches=2
+"
+
+python "${dir_path}/basic_examples/mnist_examples/image_classifier_4_lightning.py" ${args} "$@"
```
Borda, line 26: why this strange order 4-5-1-2-3?

pl_examples/loop_examples/mnist_lite.py
awaelchli: LightningCLI is not compatible with Lite

pl_examples/loop_examples/mnist_lite.py
awaelchli: I'd rather not add typing to the examples, it can get quite distracting

pl_examples/basic_examples/README.md
awaelchli: > Introducing LightningModule powered by Lite

But that's not what it is. There is no LightningModule in that step

pl_examples/basic_examples/README.md
awaelchli: the number just tells the reader in which order they need to look at the examples. 
The files get sorted normally, so it should be pretty straightforward

pl_examples/run_examples.sh
tchaton: Because 4-5 is using PyTorch Lightning and therefore the arg above. The others won't

pl_examples/loop_examples/mnist_lite.py
tchaton: Yes, agree.

pl_examples/basic_examples/README.md
awaelchli: That's because it has the default arguments set.
But in Lightning/Lite, we can highlight how easy it is to set it without code changes.

pl_examples/basic_examples/mnist_examples/image_classifier_1_pytorch.py
awaelchli: It's taken directly from the pytorch repo. 
https://github.com/pytorch/examples/blob/master/mnist/main.py

It's a standard example, so I think we should keep it this way.

pl_examples/basic_examples/mnist_examples/image_classifier_1_pytorch.py
awaelchli: It's taken 1:1 from the pure pytorch example here: https://github.com/pytorch/examples/blob/master/mnist/main.py
They don't use f-string

pl_examples/basic_examples/mnist_examples/image_classifier_2_lite.py
awaelchli: Because hparams here is a argparse Namespace.
Later on in example 3-5, self.hparams will become the LightningModule.hparams.

pl_examples/basic_examples/mnist_examples/image_classifier_3_lite_to_lightning.py
awaelchli: we only care about the test accuracy.

pl_examples/basic_examples/mnist_examples/image_classifier_3_lite_to_lightning.py
awaelchli: prepare_data on the datamodule is also not a static method
it would be better to stay consistent, otherwise it is confusing to the user

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
awaelchli: you mean remove "image_classifier"? Why?
@tchaton

pl_examples/run_examples.sh
awaelchli: we can switch the order back

pl_examples/basic_examples/mnist_examples/image_classifier_2_lite.py
tchaton: Because it is similar to the action of save_hyperparameters

pl_examples/basic_examples/mnist_examples/image_classifier_1_pytorch.py
tchaton: This is taken from PyTorch examples. I would prefer to keep argparse as it is more common to users. And json-argparse won't provide more here.

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
awaelchli: @tchaton do you understand what @Borda means?

pl_examples/basic_examples/mnist_examples/README.md
awaelchli: Like what exactly? There is no readme file in flash: https://github.com/PyTorchLightning/lightning-flash/tree/master/flash_examples

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
tchaton: I would prefer not, this would look weird.

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
Borda: you move the `transform` from inside the Lite/run and then in the next step you move it to Data module so it is confusing why you need it LM while it is not the final destination...

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
Borda: `image_classifier_4_lightning_module` just have this with module as it is introducing module

pl_examples/basic_examples/README.md
Borda: yes I know what it is for, but it is a bit confusing without context... but it is minor comment

pl_examples/basic_examples/README.md
Borda: yes, then lets add this to the comment :]

pl_examples/run_examples.sh
Borda: ok, so just let just add comment for our future us :]

pl_examples/loop_examples/mnist_lite.py
Borda: well that is one side, but my side as a user is that I have no idea what the inputs really are, just guessing bu name

Borda (approved):

pl_examples/basic_examples/README.md
tchaton: I prefer the LightningModule for 4 :)

pl_examples/basic_examples/mnist_examples/image_classifier_4_lightning.py
tchaton: Done !

pl_examples/basic_examples/mnist_examples/README.md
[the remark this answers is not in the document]
tchaton: Yes, the same except they contain the mnist_examples/*

pytorch_lightning/lite/wrappers.py
```diff
@@ -164,6 +164,11 @@ def __init__(self, dataloader: Iterable, device: Optional[torch.device] = None)
         self._dataloader = dataloader
         self._device = device
 
+    def __len__(self) -> Union[int, float]:
+        if isinstance(self._dataloader, Sized):
+            return len(self._dataloader)
+        return float("inf")
```
awaelchli, line 170: This does not belong in this PR. Why did we add this? 
Needs to be addressed in #10297

PR #10034, opened 2021-10-20
1/n Simplify spawn plugins: Simplify handling of multiprocessing queue

awaelchli:
## What does this PR do?

Part of #10059, implements the first 3 steps which are highly coupled. Please read steps 1-3 on the issue to better follow the thought process behind this PR. Steps 1-3 are combined in this PR because they are tightly entangled. Steps 4 and 5 will be independent follow-ups.

Notes to reviewer:
- TPUSpawnPlugin subclasses from DDPSpawnPlugin and it is therefore not very feasible to split this PR into two pieces. Due to the inheritance, it makes the most sense to update both plugins in the same PR.
- Please have a look at #10865 if you find the time.

A very tiny edge case breaking change in two hooks, `LightningModule.add_to_queue` and `LightningModule.get_from_queue` which now accept a `_SimpleQueue` as input instead of a `torch.multiprocessing.SimpleQueue`. This is however not expected to make an impact directly, since only the type hint changes while the user interface remains exactly the same.

## Before submitting
   _No new tests are required. We have enough (quite expensive) ddp spawn tests that will pass and confirm that these refactors and redesigns retain the existing functionality and performance._

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @akihironitta @tchaton @kaushikb11


pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -208,28 +199,11 @@ def new_process(self, trainer: "pl.Trainer", mp_queue: SimpleQueue) -> None:
...
 
         # ensure that spawned processes go through teardown before joining
         trainer._call_teardown_hook()
-
-    def post_dispatch(self, trainer: "pl.Trainer"):
-        # restore main state with best weights
-        best_path = self.mp_queue.get()
-        last_path = self.mp_queue.get()
-        self._results = self.mp_queue.get()
-        # get the `callback_metrics` and set it to the trainer
```
awaelchli, line 223: Reviewer: these get calls (and corresponding put calls) are no longer needed. The core of this PR is to move the whole tuple of results all at once. This now happens centrally inside the `spawn()` method, aligning both Lite and Lightning.

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -208,28 +199,11 @@ def new_process(self, trainer: "pl.Trainer", mp_queue: SimpleQueue) -> None:
...
-        # only in case the user does not override it.
-        # TODO: Remove the if in v1.7
-        if is_overridden("get_from_queue", self.lightning_module):
-            self.lightning_module.get_from_queue(self.mp_queue)
-        else:
-            self.get_from_queue(trainer, self.mp_queue)
-
-        # recover the weights of the processes trained in the children
-        self.__recover_child_process_weights(best_path, last_path)
```
awaelchli, line 232: Reviewer: this has moved to `__recover_results_in_main_process`

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -413,3 +400,16 @@ def teardown(self) -> None:
...
+    """Simulates a :class:`torch.multiprocessing.queue.SimpleQueue` interface using the Python list."""
+
+    def get(self) -> Any:
+        return self.pop(0)
+
+    def put(self, item: Any) -> None:
+        self.append(item)
+
+    def empty(self) -> bool:
+        return len(self) == 0
```
awaelchli, line 415: Reviewer: This is needed to preserve backward compatibility with everyone overriding the `get_form_queue` and `add_to_queue` hooks expecting a queue object. See #10865 for how I propose to go forward.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -290,18 +289,6 @@ def lightning_module(self) -> Optional["pl.LightningModule"]:
         """Returns the pure LightningModule without potential wrappers."""
         return unwrap_lightning_module(self._model) if self._model is not None else None
 
-    @property
-    def results(self) -> Optional[Union[_EVALUATE_OUTPUT, _PREDICT_OUTPUT]]:
-        """Enables plugin-agnostic access to the result returned by the training/evaluation/prediction run.
```
awaelchli, line 295: Reviewer: I can introduce a deprecation if you request it.

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -285,18 +261,19 @@ def __transfer_distrib_spawn_state_on_fit_end(self, trainer: "pl.Trainer", resul
...
+        extra = _ExtraQueue()
         if is_overridden("add_to_queue", self.lightning_module):
-            self.lightning_module.add_to_queue(self.mp_queue)
+            # TODO: Remove the if in v1.7
+            self.lightning_module.add_to_queue(extra)
         else:
-            self.add_to_queue(trainer, self.mp_queue)
+            self.add_to_queue(trainer, extra)
+
+        return best_model_path, last_path, results, extra
```
carmocca, line 272: If this is not going to be removed soon in a follow-up, I'd suggest a NamedTuple here, it'll aid with the typing too

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -413,3 +400,16 @@ def teardown(self) -> None:
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+
+class _ExtraQueue(list):
```
carmocca, line 405:
```suggestion
class _MockQueue(list):
```

or

```suggestion
class _FakeQueue(list):
```

carmocca (approved):
Nice!

pytorch_lightning/core/lightning.py
```diff
@@ -1917,7 +1917,7 @@ def model_size(self) -> float:
             )
         return get_model_size_mb(self)
 
-    def add_to_queue(self, queue: torch.multiprocessing.SimpleQueue) -> None:
+    def add_to_queue(self, queue: pl.plugins.training_type.ddp_spawn._ExtraQueue) -> None:
```
four4fish, line 1920: Curious why do we have this in lightningModule? Can we call directly from DDPSpawnPlugin?

pytorch_lightning/plugins/training_type/ddp_spawn.py
ananthsub: Agreed, NamedTuple/TypedDict/Dataclass will make this more explicit

ananthsub (approved):

pytorch_lightning/core/lightning.py
four4fish: nvm!

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -145,16 +132,22 @@ def set_world_ranks(self, process_idx: int = 0) -> None:
     def get_mp_spawn_kwargs(self, trainer: Optional["pl.Trainer"] = None) -> Dict[str, Any]:
```
four4fish, line 132: Do we still need this function?

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -169,9 +164,7 @@ def init_dist_connection(self, global_rank: int, world_size: int) -> None:
     def set_world_ranks(self, process_idx: int = 0) -> None:
         pass
 
-    def new_process(self, trainer: "pl.Trainer", mp_queue: SimpleQueue) -> None:
-        self.mp_queue = mp_queue
-
+    def new_process(self, trainer: "pl.Trainer") -> Optional[Tuple[Optional[str], Optional[str], Any, "_ExtraQueue"]]:
```
four4fish, line 167: In ddp (non-spawning plugins), we only need pass trainer to strategy for setup() which calls init_optimizer. After we clean up TrainerOptimizersMixin, most plugin doesn't need trainer anymore. 
Do you think it's possible to spawn individual functions instead of entire trainer for spawning in the furture?

four4fish:
This is super cool!! love it!

pytorch_lightning/plugins/training_type/ddp_spawn.py
awaelchli: Ah true, I was thinking about this too. 
Any of the three options containers would be ok I guess, as long as it's pickleable. 
What's the preference?

pytorch_lightning/plugins/training_type/ddp_spawn.py
awaelchli: I agree with _FakeQueue

pytorch_lightning/plugins/training_type/ddp_spawn.py
awaelchli: The original motivation for it to be a standalone method is that users/authors of plugins can just override only this part if they need to. If we remove it, they would instead override the spawn() method directly, or perhaps we could expose the `start_method` as an attribute to the plugin. Let's discuss it in an issue.

pytorch_lightning/plugins/training_type/ddp_spawn.py
four4fish: +1 namedTuple

pytorch_lightning/plugins/training_type/tpu_spawn.py
awaelchli: In step 4 and 5 of  #10059 we will automatically get rid of some of these trainer references yes. There are some occurrences in the plugins where we need to check things like `trainer.stage == FITTING` but these trainer instances can also be accessed through model.trainer if people think this is better.

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -101,15 +100,6 @@ def num_nodes(self, num_nodes: int) -> None:
     def local_rank(self) -> int:
         return self._local_rank
 
-    def __getstate__(self):
```
tchaton, line 104: Neat !

tchaton (approved):
Looks great! Awesome work.

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -413,3 +398,23 @@ def teardown(self) -> None:
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+
+class _FakeQueue(list):
```
justusschock, line 403: afaik subclassing mutable builtins (especially dict and list) can sometimes lead to unexpected results. Maybe use [`UserList`](https://docs.python.org/3/library/collections.html#collections.UserList) as base class instead?

justusschock (approved):
LGTM. Just one minor comment :)

pytorch_lightning/plugins/training_type/ddp_spawn.py
awaelchli: going with namedTuple for now

pytorch_lightning/plugins/training_type/ddp_spawn.py
awaelchli: #10895

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -161,9 +156,7 @@ def init_dist_connection(self, global_rank: int, world_size: int) -> None:
     def set_world_ranks(self, process_idx: int = 0) -> None:
         pass
 
-    def new_process(self, trainer: "pl.Trainer", mp_queue: SimpleQueue) -> None:
-        self.mp_queue = mp_queue
-
+    def new_process(self, trainer: "pl.Trainer") -> Optional[Tuple[Optional[str], Optional[str], Any, "_FakeQueue"]]:
```
carmocca, line 159: Isn't this missing the SpawnOutput return?

pytorch_lightning/plugins/training_type/sharded_spawn.py
```diff
@@ -115,12 +114,12 @@ def pre_backward(self, closure_loss: torch.Tensor) -> None:
     def post_training_step(self):
         pass
 
-    def new_process(self, trainer: "pl.Trainer", mp_queue: SimpleQueue) -> None:
+    def new_process(self, trainer: "pl.Trainer") -> Optional[Tuple[Optional[str], Optional[str], Any, _FakeQueue]]:
```
carmocca, line 117: Same comment

pytorch_lightning/plugins/training_type/ddp_spawn.py
carmocca: @justusschock I don't think that's true nowadays: https://stackoverflow.com/questions/25464647/list-vs-userlist-and-dict-vs-userdict

`list` is the recommended approach for simple extension
and `collections.*` for complex extension

pytorch_lightning/plugins/training_type/sharded_spawn.py
awaelchli: true, I forgot to update the subclasses. Fortunately new_process will become obsolete in 2/n if everything goes according to plan.

pytorch_lightning/plugins/training_type/ddp_spawn.py
carmocca: Quoting Raymond Hettinger (Python core dev):

"FWIW, the UserDict class in Py3.x was relocated to the collections
module.  It isn't gone.  We made an effort to kill it but found that
there were compelling use cases that were not a cleanly solved by any
other approach."

https://bugs.python.org/issue5402

pytorch_lightning/plugins/training_type/sharded_spawn.py
awaelchli: resolved in #10934

PR #9920, opened 2021-10-13
Add support for init_meta_context, materialize_module

tchaton:
## What does this PR do?

Fixes #9375

This PR builds on top of https://github.com/pytorch/pytorch/pull/66317. A code section will be dropped once merged to PyTorch.

The goal is no code change for the end users.

```py
from pytorch_lightning.utilities.meta import init_meta_context, materialize_module

class MLP(nn.Module):
    def __init__(self, num_convs: int):
        super().__init__()
        self.lins = []
        for _ in range(num_convs):
            self.lins.append(nn.Linear(1, 1))
        self.layer = nn.Sequential(*self.lins)

with init_meta_context():
    m = nn.Linear(in_features=1, out_features=1)
    assert m.weight.device.type == "meta"
    mlp = MLP(4)
    assert mlp.layer[0].weight.device.type == "meta"

    materialize_module(mlp)
    assert mlp.layer[0].weight.device.type == "cpu"

m = nn.Linear(in_features=1, out_features=1)
assert m.weight.device.type == "cpu"

with init_meta_context():
    m = nn.Linear(in_features=1, out_features=1)
    assert m.weight.device.type == "meta"

m = nn.Linear(in_features=1, out_features=1)
assert m.weight.device.type == "cpu"
```

TODO:
- [x] Verify it works with DeepSpeed on BoringModel

```py
    with init_meta_context():
        model = BoringModel()
    assert model.layer.weight.device.type == "meta"
    trainer = Trainer(
        default_root_dir=tmpdir, plugins=[DeepSpeedPlugin(stage=3)], gpus=2, fast_dev_run=True, precision=16
    )
    trainer.fit(model)
    assert model.layer.weight.device.type == "cpu"
```

- [x] Verify it works with DeepSpeed on minGPT. Find the associated example here: https://github.com/SeanNaren/minGPT/blob/meta_device/train.py

Screenshot 2021-10-15 at 12 59 53

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
[...]

PR #8208, opened 2021-06-29
Fix double precision casting complex buffers

ethanwharris:
## What does this PR do?

Fixes #8178 

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [ ] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [ ] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #10079, opened 2021-10-22
Add context manager to properly convert the precision

carmocca:
## What does this PR do?

The fix in #8208 isn't working as expected. The check for `complex64` should be

`assert self.complex_buffer.dtype == torch.complex128`.

This is because (from https://pytorch.org/docs/stable/generated/torch.set_default_dtype.html):

`when the default floating point type is float64 the default complex type is complex128.`

However, I didn't find a way to fix it, because moving the model `model.double()` or `model.to(torch.double)` does not convert it to `complex128` after it has been created.

The only option I see is that the model instantiation is done after `torch.set_default_dtype(torch.double)` but that's outside of Lightning's scope.
To make this easier, this PR adds the following context manager:

```python
with autodtype(torch.double):
    model = Model()
# or
with trainer.precision_plugin.forward_context():
    model = Model()
```

This is similar to how #9920 works.

Some other changes included:
- Ensure the default dtype is restored in case of failure.
- Use `set_default_dtype` instead of `set_default_tensor_type`.
- Generalize the classes. This is not useful right now because `set_default_dtpye` only supports floating types currently.

The model is no longer moved. 

## Before submitting

## PR review

### DIFF
--- a/pytorch_lightning/plugins/precision/double.py
+++ b/pytorch_lightning/plugins/precision/double.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from contextlib import contextmanager
-from typing import Any, Generator, List, Tuple
+from typing import Any, Generator, List, Optional, Tuple
 
 import torch
 import torch.nn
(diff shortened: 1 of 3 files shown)


carmocca:
This is no longer necessary after the addition of `init_module` and `configure_model`

PR #10123, opened 2021-10-25
Changed the model size calculation using `ByteCounter`

### LINKED ISSUE #10074: `get_model_size_mb` (`LightningModule.model_size`) shouldn't create temporary files in the current directory
Currently, the [`get_model_size_mb` function](https://github.com/PyTorchLightning/pytorch-lightning/blob/c3614f1c07caf2c1519fcbe37140f2aeef7a8308/pytorch_lightning/utilities/memory.py#L165) is implemented like this:
```python
def get_model_size_mb(model: Module) -> float:
    """Calculates the size of a Module in megabytes by saving the model to a temporary file and reading its size.
    The computation includes everything in the :meth:`~torch.nn.Module.state_dict`,
    i.e., by default the parameteters and buffers.
    Returns:
        Number of megabytes in the parameters of the input module.
    """
    # TODO: Implement a method without needing to download the model
    tmp_name = f"{uuid.uuid4().hex}.pt"
    torch.save(model.state_dict(), tmp_name)
    size_mb = os.path.getsize(tmp_name) / 1e6
    os.remove(tmp_name)
    return size_mb
```

This writes the model to disk and also introduces a race condition where the temporary file may be left behind if the user "Ctrl+C"s the program while the `torch.save` and `os.path.getsize` calls are running. I originally found this bug by investigating, what was creating the mysterious `cb0e0c0d62ee4c20986c67462c1105d3.pt` files in my project directory.

You can achieve the same result without touching the disk by doing
```python
import io

target = io.BytesIO()
torch.save(model.state_dict(), target)
size_mb = target.getbuffer().nbytes / 1e6
```

rohitgr7:
## What does this PR do?

Fixes #10074
all credits to @RuRo :)
Tests are already there for `get_model_size_mb`

## Before submitting

## PR review

## Did you have fun?


rohitgr7:
@tchaton couldn't find a way to support this for Deepspeed/FSDP. Discussed this with @SeanNaren too and seems like its right not practically possible since weights are sharded and are not kept in a readable format. I'd suggest for now we can merge this for v1.5 and later we can explore more.

pytorch_lightning/utilities/memory.py
```diff
@@ -25,6 +24,17 @@
...
 
 
+class ByteCounter:
+    def __init__(self) -> None:
+        self.nbytes: int = 0
+
+    def write(self, data: bytes) -> None:
+        self.nbytes += len(data)
+
+    def flush(self) -> None:
```
kaushikb11, line 34:
```suggestion
    def write(self, data: bytes) -> None:
        self.nbytes += len(data)

    def flush(self) -> None:
```
Would like to some docstrings in these methods

kaushikb11 (approved):
LGTM! Could you add it to the class that it's not supported by DeepSpeed and Sharded plugins.

tchaton (approved):
LGTM !

SeanNaren (approved):
Looks good! can we add an issue around FSDP/DeepSpeed support for summarization of model size? As @tchaton mentioned it is possible, just will require some thought/implementation

CHANGELOG.md
```diff
@@ -119,6 +119,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     * Updated precision attributes in `DeepSpeedPlugin` ([#10164](https://github.com/PyTorchLightning/pytorch-lightning/pull/10164))
     * Added the ability to return a result from rank 0 in `DDPSpawnPlugin.spawn` ([#10162](https://github.com/PyTorchLightning/pytorch-lightning/pull/10162))
     * Added `pytorch_lightning.lite` package ([#10175](https://github.com/PyTorchLightning/pytorch-lightning/pull/10175))
+- Added `ByteCounter` to calculate model_size ([#10123](https://github.com/PyTorchLightning/pytorch-lightning/pull/10123))
```
awaelchli, line 122: I suggest to mention this in the changed section, that we are now using a byte counter instead of file io.

awaelchli (approved):

pytorch_lightning/utilities/memory.py
```diff
@@ -25,6 +24,20 @@
...
+
+    def __init__(self) -> None:
+        self.nbytes: int = 0
+
+    def write(self, data: bytes) -> None:
+        """Stores the total bytes of the data."""
+        self.nbytes += len(data)
+
+    def flush(self) -> None:
+        pass
```
ananthsub, line 38: why do we need a byte counter class? the write/flush APIs aren't used anywhere

pytorch_lightning/utilities/memory.py
```diff
@@ -25,6 +24,20 @@
 from pytorch_lightning.utilities.apply_func import apply_to_collection
 
 
+class ByteCounter:
```
carmocca, line 27: This class should be protected

PR #10192, opened 2021-10-27
Set devices to 1 when it's just `Trainer(accelerator='auto')`

kaushikb11:
## What does this PR do?

Set devices to 1 when it's just `Trainer(accelerator='auto')`

## Before submitting

## PR review

## Did you have fun?


kaushikb11:
I am putting this PR on hold. IMO, if a user just does Trainer(accelerator="auto"), we should crash it and ask them to pass devices=x or gpus if it had GPUs.(like we do right now)

Rather than assigning only one device for them.
For this particular reason, we support `devices="auto"`

@PyTorchLightning/core-contributors

carmocca:
It could also default to `devices="auto"` if only `accelerator="auto"` was passed.

SkafteNicki (approved):

Borda (approved):

ananthsub:
Is there any discussion of this feature in an issue or RFC? It is very hard to review this without the broader context of what's trying to be achieved here

akihironitta:
I'm not aware of the context outside this PR at all, but I don't think we should accept this change. If it were more common to use only one device out of multiple devices than using all of them, it'd be fine, but it's not common, right?

In my opinion, it should default to `devices="auto"` as carmocca suggests because I think that's what users (at least I) expect, or at least raise an exception asking users to specify `devices` as kaushikb11 suggests.

tchaton:
IMO, I believe accelerator="auto" should be the default and devices become a required argument.

kaushikb11:
`devices` is set to `auto` when it's not passed. Closing the PR.

PR #10247, opened 2021-10-29
Add warning if multiple batch_sizes are found from ambiguous batch

### LINKED ISSUE #10093: `utilities.data.extract_batch_size` can silently fail -- maybe it needs to be less automatic?
## Motivation

Hey! So from what I understand `extract_batch_size` is used by `ResultCollection` to reduce/re-weight various logging metrics such as validation loss, etc.

We had an issue where our batch collation function aggregated Tensors into a list instead of stacking them (needed because of size mismatches), and `extract_batch_size` would recursively crawl through the batch object until it found a Tensor and returned its first dimention size -- which is wrong (the list length is correct batch size). So we would silently get "random" batch sizes and the val_loss logging becomes incorrect and then the LR scheduling / early stopping incorrectly triggers etc.

### Pitch

1. Maybe there should be some warning / note in the documentation (didn't find any)
2. If the batch object isn't a tensor (like a dict/dataclass), maybe try to `getattr(batch, 'batch_size')` and only then produce a warning and try to infer it automatically.

rohitgr7:
## What does this PR do?

Fixes #10093

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/utilities/data.py
```diff
@@ -12,34 +12,57 @@
...
+    if isinstance(batch, (Iterable, dict)):
+        if isinstance(batch, dict):
+            batch = batch.values()
+
+        batch_sizes = list(flatten_list([_extract_batch_size(sample, recursive=True) for sample in batch]))
+        batch_size = batch_sizes[0]
+
+        if not all(bs == batch_size for bs in batch_sizes) and not recursive:
+            warning_cache.warn(
+                f"Lightning is trying to infer the `batch_size` from an ambiguous collection. The batch size we"
```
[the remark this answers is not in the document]
rohitgr7: haha... true!!

pytorch_lightning/utilities/data.py
```diff
@@ -12,34 +12,57 @@
...
+def _extract_batch_size(batch: BType, recursive=False) -> List[int]:
+    if isinstance(batch, torch.Tensor):
+        return [batch.size(0)]
+    if isinstance(batch, str):
+        return [len(batch)]
+    if isinstance(batch, (Iterable, dict)):
+        if isinstance(batch, dict):
+            batch = batch.values()
+
+        batch_sizes = list(flatten_list([_extract_batch_size(sample, recursive=True) for sample in batch]))
```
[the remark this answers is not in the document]
rohitgr7: hmm. yeah.. let me try some DSA :sweat_smile:

pytorch_lightning/utilities/data.py
```diff
@@ -18,28 +18,55 @@
...
+
+
+def _extract_batch_size(batch: BType, batch_sizes) -> bool:
+    current_batch_size = None
+
+    if isinstance(batch, torch.Tensor):
+        current_batch_size = batch.size(0)
+    elif isinstance(batch, str):
+        current_batch_size = len(batch)
+    elif not isinstance(batch, (Iterable, dict)):
```
eladsegal, line 35: @rohitgr7 
Instead of using `dict`, `Mapping` should be used.
Example for a case in which the batch will be `Mapping` but not `dict` is when using huggingface.
Currently, 9 is returned instead of 2 for this HF example code:
```
from transformers import AutoTokenizer
tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
pt_batch = tokenizer(
    ["We are very happy to show you the 🤗 Transformers library.", "We hope you don't hate it."],
    padding=True,
    truncation=True,
    max_length=512,
    return_tensors="pt"
)
print(f"Inferred batch size is {extract_batch_size(pt_batch)}")
```

rohitgr7: true.. it's a `UserDict`.. thanks you :)

CHANGELOG.md
```diff
@@ -231,6 +231,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `ckpt_path` argument for `trainer.fit()` ([#10061](https://github.com/PyTorchLightning/pytorch-lightning/pull/10061))
 
 
+- Added warning if multiple batch_sizes are found from ambiguous batch ([#10247](https://github.com/PyTorchLightning/pytorch-lightning/pull/10247))
```
awaelchli, line 234:
```suggestion
- Added a warning if multiple batch sizes are found from ambiguous batch ([#10247](https://github.com/PyTorchLightning/pytorch-lightning/pull/10247))
```

pytorch_lightning/utilities/data.py
```diff
@@ -18,28 +18,55 @@
...
 
 from pytorch_lightning.utilities import rank_zero_warn
+from pytorch_lightning.utilities.warnings import WarningCache
 
 BType = Union[torch.Tensor, str, Mapping[Any, "BType"], Iterable["BType"]]
 
+warning_cache = WarningCache()
+
+
+def _extract_batch_size(batch: BType, batch_sizes) -> bool:
```
awaelchli, line 28: isn't there a more elegant implementation using a Generator (yield)? 
Then there is no special return value needed, and the warning can be handled externally.

rohitgr7: I'll try

tchaton (approved):
LGTM ! We can provide a more elegant version of it in another PR too.

pytorch_lightning/utilities/data.py
```diff
@@ -64,9 +46,18 @@ def extract_batch_size(batch: BType) -> int:
...
         ``len(tensor)`` when found, or ``1`` when it hits an empty or non iterable.
     """
-    batch_sizes = []
-    _extract_batch_size(batch, batch_sizes)
-    return batch_sizes[0]
+    batch_size = None
+    for bs in _extract_batch_size(batch):
+        if batch_size is None:
+            batch_size = bs
+        elif batch_size != bs:
```
tchaton, line 53: Neat !

tchaton (commented):
Nice !

pytorch_lightning/utilities/data.py
```diff
@@ -18,28 +18,46 @@
...
-        return extract_batch_size(sample)
-
-    return 1
+    batch_size = None
+    for bs in _extract_batch_size(batch):
+        if batch_size is None:
+            batch_size = bs
+        elif batch_size != bs:
+            warning_cache.warn(
+                f"Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
```
kaushikb11, line 55:
```suggestion
                "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
```

pytorch_lightning/utilities/data.py
```diff
@@ -18,28 +18,46 @@
...
 
 from pytorch_lightning.utilities import rank_zero_warn
+from pytorch_lightning.utilities.warnings import WarningCache
 
 BType = Union[torch.Tensor, str, Mapping[Any, "BType"], Iterable["BType"]]
 
+warning_cache = WarningCache()
+
+
+def _extract_batch_size(batch: BType) -> int:
```
awaelchli, line 28:
```suggestion
def _extract_batch_size(batch: BType) -> Generator[int, None, None]:
```
did mypy not error?

kaushikb11 (approved):

pytorch_lightning/utilities/data.py
```diff
@@ -18,28 +18,46 @@
...
-        return len(batch)
-    if isinstance(batch, dict):
-        sample = next(iter(batch.values()), 1)
-        return extract_batch_size(sample)
-    if isinstance(batch, Iterable):
-        sample = next(iter(batch), 1)
-        return extract_batch_size(sample)
-
-    return 1
+    batch_size = None
```
awaelchli, line 49:
```suggestion
    batch_size = 1
```
or annotation as `Optional[int]`

awaelchli (approved):

tests/utilities/test_data.py
```diff
@@ -23,6 +23,22 @@ def test_extract_batch_size():
...
+    warning_cache.clear()
+
+    batch = {"test": [{"test": [torch.zeros(10, 10), torch.zeros(11, 10)]}]}
+    with pytest.warns(UserWarning, match="Trying to infer the `batch_size` .* we found is 10."):
+        extract_batch_size(batch)
+
+    warning_cache.clear()
+
+    batch = [{"test": torch.zeros(10, 10), "test_1": torch.zeros(11, 10)}]
+    with pytest.warns(UserWarning, match="Trying to infer the `batch_size` .* we found is 10."):
```
awaelchli, line 39: I think this test needs a check that no warning is raised if the collection has a uniform batch size

pytorch_lightning/utilities/data.py
rohitgr7: it can never be optional ? since above it will return 1 in the last else condition. The reason I initalized it with None to first batch_size since there is a possibility of getting 1 in the first call itself.

pytorch_lightning/utilities/data.py
rohitgr7: why None here?

pytorch_lightning/utilities/data.py
awaelchli: that's interesting, I didn't see that but that means mypy can infer that it's never None, i.e., that the for loop always runs?

rohitgr7: I think yes, it will.

awaelchli (approved):

pytorch_lightning/utilities/data.py
awaelchli: On my prs, it never wants to be smart...

rohitgr7: lucky day maybe 😆

PR #10249, opened 2021-10-29
Add `LightningModule.lr_scheduler_step`

### LINKED ISSUE #10190: Support timm schedulers and other custom schedulers
## 🚀 Feature

Previous discussions:
https://github.com/PyTorchLightning/pytorch-lightning/issues/9964
https://github.com/PyTorchLightning/pytorch-lightning/issues/5555

### Motivation

[timm](https://github.com/rwightman/pytorch-image-models) seems to be a very popular package (14k stars). Lightning currently raises an error when using it and doesn't support it out of the box.

### Pitch
Since it has all the basic interface that we need inside lightning, we can add a hook `LightningModule.scheduler_step` similar to `optimizer_step`.

This will be the default implementation
```py
class LightningModule(...):
    ...
    def scheduler_step(self, scheduler, opt_idx, step, monitor, ...):
       
        if monitor is None:
            scheduler.step()
        else:
            scheduler.step(monitor)
```

and for timm or some other schedulers one can configure and can do:
```py
class LitModel(pl.LightningModule):
    ...
    def scheduler_step(self, scheduler, opt_idx, step, monitor, ...):
        scheduler.step(epoch=step)
```

### Alternatives

Can be done using manual optimization if we remove some assertions within codebase, but doesn't seem to be a reliable solution.

### Additional context

As always open to discussion/better suggestions :)

<sub>

</sub>

cc @borda

rohitgr7:
## What does this PR do?

Fixes #10190

## Before submitting

## PR review

## Did you have fun?

cc @borda @carmocca @justusschock @awaelchli @ananthsub @ninginthecloud @jjenniferdai @rohitgr7


rohitgr7:
sure :)

docs/source/common/optimizers.rst
```diff
@@ -252,6 +252,31 @@ If you want to call schedulers that require a metric value after each epoch, con
...
+
+.. testcode:: python
+
+    def configure_optimizers(self):
+        optimizer = ...
+        scheduler = ...
+        return [optimizer], [scheduler]
+
+
+    def lr_scheduler_step(self, scheduler, step, optimizer_idx, monitor_val=None):
```
tchaton, line 275: Should we pass both the step and the epoch there ?

pytorch_lightning/core/lightning.py
```diff
@@ -1578,6 +1578,44 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
             optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
         )
 
+    def lr_scheduler_step(
+        self,
+        scheduler: Any,
+        step: int,
```
tchaton, line 1584: Personally, if I want to implement a custom scheduling mechanism I might want to have access to both the global_step and epoch.

docs/source/common/optimizers.rst
carmocca: I wouldn't considering the Trainer object is available

pytorch_lightning/core/lightning.py
carmocca: You have access through `self.trainer`.

pytorch_lightning/core/lightning.py
```diff
@@ -1262,7 +1262,7 @@ def configure_optimizers(self):
                 # `scheduler.step()`. 1 corresponds to updating the learning
```
carmocca, line 1262: It might be worth splitting this docstring now that this adds a hook for the scheduler logic specifically.

pytorch_lightning/core/lightning.py
rohitgr7: yes thought so.. but these are some of the arguments that actually takes part in scheduler step.
we have some others like:
```py
optimizer_step(epoch, ...) <- epoch can be accessed using `trainer.current_epoch`.
transfer_batch_to_device(..., device) <- device can be accessed using `self.device`.
```
since I personally know these are available, so not sure whether to remove them from user's point of view. but if we choose to remove them, we should update the above 2 hook as well IMO.

pytorch_lightning/core/lightning.py
rohitgr7: well, frequency is still a part of `configure_optimizers`. the hook is used to configure how to do the `scheduler.step`, similar to `optimzer_step`. when to update is still decided internally using the frequency specified by the user.

pytorch_lightning/core/lightning.py
carmocca: Passing arguments that aren't used explicitly opens the door to the question of, which should be passed and which should not.

I also don't like having it in `optimizer_step`. But in the case of `transfer_batch_to_device`, it's good to have because the default implementation uses it.

rohitgr7: btw `transfer_batch_to_device` doesn't need one either. it's a lightningmodule, hook, device is a part of module itself.
for now, I'll remove it from `scheduler_step`, we can think of others later.

ananthsub:
Where is the issue/RFC describing this feature and what it aims to solve? The PR has 0 summary for what this PR does. This makes it very difficult for anyone in the community to review. 

This is touching the core LightningModule! There needs to be much more information about this before changes are just merged

rohitgr7:
> Where is the issue/RFC describing this feature and what it aims to solve? The PR has 0 summary for what this PR does. This makes it very difficult for anyone in the community to review. 
> 
> This is touching the core LightningModule! There needs to be much more information about this before changes are just merged

It's still draft. Attached the issue link in the description.

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
             optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
         )
 
+    def lr_scheduler_step(
+        self,
+        scheduler: Any,
+        optimizer_idx: Optional[int] = None,
+        monitor_val: Optional[Union[float, torch.Tensor]] = None,
```
tchaton, line 1500: Let's call this metrics as it is the expected API from PyTorch.

tchaton: Should we provide the epoch and the current optimizer step count ?

pytorch_lightning/trainer/optimizers.py
```diff
@@ -18,13 +18,28 @@
...
+from typing_extensions import Protocol, runtime_checkable
 
 import pytorch_lightning as pl
 from pytorch_lightning.core.optimizer import LightningOptimizer
 from pytorch_lightning.utilities import rank_zero_warn
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 
 
+@runtime_checkable
+class _SupportedLRScheduler(Protocol):
```
tchaton, line 30: Neat !

docs/source/common/optimizers.rst
```diff
@@ -252,6 +252,31 @@ If you want to call schedulers that require a metric value after each epoch, con
...
+.. testcode:: python
+
+    def configure_optimizers(self):
+        optimizer = ...
+        scheduler = ...
+        return [optimizer], [scheduler]
+
+
+    def lr_scheduler_step(self, scheduler, optimizer_idx, monitor_val=None):
+        scheduler.step(epoch=self.current_epoch)  # timm's scheduler need the epoch value
```
tchaton, line 276: Let's add an example with timm directly .

```suggestion

   from timm.scheduler import  TanhLRScheduler

    def configure_optimizers(self):
        optimizer = ...
        scheduler = TanhLRScheduler(optimizer, ...)
        return [optimizer], [{"scheduler": scheduler, "interval": "step"}]


    def lr_scheduler_step(self, scheduler, optimizer_idx, monitor_val=None):
        scheduler.step(epoch=self.trainer.current_epoch)  # timm's scheduler need the epoch value
```

pytorch_lightning/core/lightning.py
rohitgr7: we can, but it's directly accessible using `self`.
also Carlos comment here: https://github.com/PyTorchLightning/pytorch-lightning/pull/10249#discussion_r741031873

pytorch_lightning/core/lightning.py
```diff
@@ -1177,7 +1177,7 @@ def configure_optimizers(self):
                 # `scheduler.step()`. 1 corresponds to updating the learning
                 # rate after every epoch/step.
                 "frequency": 1,
-                # Metric to to monitor for schedulers like `ReduceLROnPlateau`
+                # Metric to to monitor for schedulers like ``ReduceLROnPlateau``
```
carmocca, line 1180: This is a code-block so we don't need double backticks to render in monospace.

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
             optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
         )
 
+    def lr_scheduler_step(
+        self,
+        scheduler: Any,
+        optimizer_idx: Optional[int] = None,
+        monitor_val: Optional[Union[float, torch.Tensor]] = None,
+    ):
```
carmocca, line 1501:
```suggestion
    ) -> None:
```

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
             optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
         )
 
+    def lr_scheduler_step(
+        self,
+        scheduler: Any,
+        optimizer_idx: Optional[int] = None,
```
carmocca, line 1499: Should this be always int?

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -505,14 +505,15 @@ def _update_learning_rates(
...
+                self.trainer._call_lightning_module_hook(
+                    "lr_scheduler_step",
+                    lr_scheduler["scheduler"],
+                    optimizer_idx=lr_scheduler["opt_idx"],
+                    monitor_val=monitor_val,
+                )
                 self.scheduler_progress.increment_completed()
 
-    def _get_monitor_value(self, key: str) -> Any:
+    def _get_monitor_value(self, key: str) -> Optional[Union[float, torch.Tensor]]:
```
carmocca, line 516: Technically nothing forbids that the user adds custom metrics of any type to the dictionary. Is the strict typing necessary?

pytorch_lightning/trainer/optimizers.py
```diff
@@ -187,10 +202,13 @@ def _configure_schedulers(
                     lr_schedulers.append(
                         {**default_config, "scheduler": scheduler, "reduce_on_plateau": True, "monitor": monitor}
                     )
-                elif isinstance(scheduler, optim.lr_scheduler._LRScheduler):
-                    lr_schedulers.append({**default_config, "scheduler": scheduler})
                 else:
-                    raise ValueError(f'The provided lr scheduler "{scheduler}" is invalid')
+                    lr_schedulers.append({**default_config, "scheduler": scheduler})
+
+            current_scheduler = lr_schedulers[-1]["scheduler"]
```
carmocca, line 208: Why `[-1]` and not in the loop?

rohitgr7: schedulers config is being created differently under different conditions up there. So check whether it's a valid scheduler or not inside every condition might make it too verbose. here is will just check the recently added one in the list.

docs/source/common/optimizers.rst
```diff
@@ -252,6 +252,31 @@ If you want to call schedulers that require a metric value after each epoch, con
...
+----------------------------------------------
+Lightning allows custom learning rate schedulers which are not present in
+`PyTorch natively <https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate>`_.
+One good example is `Timm Schedulers <https://github.com/rwightman/pytorch-image-models/blob/master/timm/scheduler/scheduler.py>`_.
+You can configure how your learning rate will be updated based on your custom implementation
+and lightning will handle when they should be updated based on the scheduler config provided inside
+:meth:`~pytorch_lightning.core.lightning.LightningModule.configure_optimizers`. For you custom
+implementation you must override :meth:`~pytorch_lightning.core.lightning.LightningModule.lr_scheduler_step`
+if necessary. If you are using native PyTorch schedulers, there is no need to override this hook since
+Lightning will handle it optimally by default.
```
ananthsub, line 265:
```suggestion
and Lightning will handle when they should be updated based on the scheduler config provided inside
:meth:`~pytorch_lightning.core.lightning.LightningModule.configure_optimizers`. For your custom
implementation, you must override :meth:`~pytorch_lightning.core.lightning.LightningModule.lr_scheduler_step`. If you are using native PyTorch schedulers, there is no need to override this hook since
Lightning will handle it optimally by default.
```

pytorch_lightning/core/lightning.py
rohitgr7: ideally yes... there should be.. but doesn't seem to be the case on master even with multiple optimizers. There might be an underlying bug here. Will investigate.

pytorch_lightning/trainer/optimizers.py
```diff
@@ -18,13 +18,28 @@
...
+
+    def step(self, *args: Any, **kwargs: Any) -> None:
+        ...
+
+    def state_dict(self) -> Dict[str, Any]:
+        ...
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        ...
+
```
ananthsub, line 41: this should go in `core/optimizers.py` https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/core/optimizer.py

given ongoing efforts to remove the trainer optimizer mixin: https://github.com/PyTorchLightning/pytorch-lightning/issues/10681

rohitgr7: yes I am waiting for that one to merge to make further changes here.

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
...
+    ) -> None:
+        r"""
+        Override this method to adjust the default way the
+        :class:`~pytorch_lightning.trainer.trainer.Trainer` calls each scheduler.
+        By default, Lightning calls ``step()`` and as shown in the example
+        for each scheduler based on its ``interval``.
+
+        Args:
+            scheduler: Learning rate scheduler.
+            optimizer_idx: Index of the optimizer associated with scheduler.
```
awaelchli, line 1510:
```suggestion
            optimizer_idx: Index of the optimizer associated with this scheduler.
```

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
...
+
+            # Alternative way to update schedulers if it requires an epoch value
+            def lr_scheduler_step(self, scheduler, optimizer_idx, metrics):
+                scheduler.step(epoch=self.current_epoch)
+
+        """
+        if metrics is None:
+            scheduler.step()
+        else:
+            scheduler.step(metrics=metrics)
```
awaelchli, line 1530: This doesn't make sense considering that most schedulers won't take this argument. Previously, it was passed positionally, and wasn't plural

docs/source/common/optimizers.rst
```diff
@@ -252,6 +252,29 @@ If you want to call schedulers that require a metric value after each epoch, con
 
 -----
 
+Bring your own Custom Learning Rate Schedulers
```
awaelchli, line 255: This is ordered under "Manual Optimization", probably not intentional.

rohitgr7: the structure of this doc is still not appropriate. I am reviewing this one in another PR. Will be restructuring it there.

pytorch_lightning/core/lightning.py
rohitgr7: updaed!

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
...
+            monitor_val: Value of the metric used for schedulers like ``ReduceLROnPlateau``.
+
+        Examples::
+
+            # DEFAULT
+            def lr_scheduler_step(self, scheduler, optimizer_idx, metrics):
+                if monitor_val is None:
+                    scheduler.step()
+                else:
+                    scheduler.step(monitor_val)
```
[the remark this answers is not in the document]
rohitgr7: good catch! updated!

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -400,7 +399,7 @@ def _setup_model_and_optimizers(self, model: Module, optimizers: List[Optimizer]
         return self.model, [optimizer]
 
     def _setup_model_and_optimizer(
-        self, model: Module, optimizer: Optimizer, lr_scheduler: Optional[_LRScheduler] = None
+        self, model: Module, optimizer: Optimizer, lr_scheduler: Optional[LRSchedulerTypeUnion] = None
```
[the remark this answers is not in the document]
rohitgr7: updated!

pytorch_lightning/core/optimizer.py
```diff
@@ -394,3 +405,17 @@ def zero_grad(self, set_to_none: Optional[bool] = False) -> None:
 
     def __repr__(self) -> str:
         return "No Optimizer"
+
+
+@runtime_checkable
+class _SupportedLRScheduler(Protocol):
+    """This class is used to detect if an object is stateful using `isinstance(obj, _SupportedLRScheduler)`"""
```
awaelchli, line 412:
```suggestion
    """This class is used to detect if a learning rate scheduler is stateful using `isinstance(obj, _SupportedLRScheduler)`"""
```

pytorch_lightning/utilities/types.py
```diff
@@ -46,12 +46,11 @@
...
 
 
-# Copied from `torch.optim.lr_scheduler.pyi`
-# Missing attributes were added to improve typing
+# Inferred from `torch.optim.lr_scheduler.pyi`
 class _LRScheduler:
     optimizer: Optimizer
 
-    def __init__(self, optimizer: Optimizer, last_epoch: int = ...) -> None:
+    def __init__(self, optimizer: Optimizer, *args: Any, **kwargs: Any) -> None:
```
awaelchli, line 53: why is it necessary to deviate from the pytorch definitions?

tests/trainer/optimization/test_optimizers.py
```diff
@@ -661,3 +662,97 @@ def on_save_checkpoint(self, checkpoint):
...
+        limit_val_batches=0,
+    )
+    trainer.fit(model)
+
+    for param_group in trainer.optimizers[1].param_groups:
+        assert param_group["lr"] == lr / math.factorial(max_epochs)
+
+
+def test_invalid_lr_scheduler(tmpdir):
+    """Test that custom lr_schedulers works and `lr_scheduler_hook` is called at appropriate time."""
```
awaelchli, line 738: the docstring does not express what is being tested

tests/trainer/optimization/test_optimizers.py
```diff
@@ -661,3 +662,97 @@ def on_save_checkpoint(self, checkpoint):
     model.training_epoch_end = None
     trainer.fit(model)
     assert model.on_save_checkpoint_called
+
+
+def test_lr_scheduler_step_hook(tmpdir):
+    """Test that custom lr_schedulers works and `lr_scheduler_hook` is called at appropriate time."""
```
awaelchli, line 668:
```suggestion
    """Test that custom lr schedulers work and `lr_scheduler_step` is called at appropriate time."""
```

awaelchli (approved):

pytorch_lightning/utilities/types.py
rohitgr7: the main motivation is to support custom schedulers that follow certain protocol API, but not necessarily the arguments. One example: https://github.com/rwightman/pytorch-image-models/blob/master/timm/scheduler/scheduler.py

tests/trainer/optimization/test_optimizers.py
rohitgr7: my bad!

pytorch_lightning/core/optimizer.py
```diff
@@ -394,3 +405,17 @@ def zero_grad(self, set_to_none: Optional[bool] = False) -> None:
 
     def __repr__(self) -> str:
         return "No Optimizer"
+
+
+@runtime_checkable
+class _SupportedLRScheduler(Protocol):
```
carmocca, line 411:
```suggestion
class _SupportsLRScheduler(Protocol):
```

.gitignore
```diff
@@ -142,7 +142,7 @@ mnist/
 legacy/checkpoints/
 *.gz
 *ubyte
-
+MNIST/
```
carmocca, line 145: huge nit: put this beside the other `mnist/` ignore

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
             optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
         )
 
+    def lr_scheduler_step(
+        self,
+        scheduler: LRSchedulerTypeUnion,
+        optimizer_idx: Optional[int] = None,
+        metric: Optional[Union[float, torch.Tensor]] = None,
```
carmocca, line 1500: Technically this is `Any` as this comes from `trainer.callback_metrics`

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
...
+        r"""
+        Override this method to adjust the default way the
+        :class:`~pytorch_lightning.trainer.trainer.Trainer` calls each scheduler.
+        By default, Lightning calls ``step()`` and as shown in the example
+        for each scheduler based on its ``interval``.
+
+        Args:
+            scheduler: Learning rate scheduler.
+            optimizer_idx: Index of the optimizer associated with this scheduler.
+            metric: Value of the metric used for schedulers like ``ReduceLROnPlateau``.
```
carmocca, line 1511:
```suggestion
            metric: Value of the monitor used for schedulers like ``ReduceLROnPlateau``.
```

pytorch_lightning/core/optimizer.py
```diff
@@ -298,10 +301,13 @@ def _configure_schedulers_automatic_opt(schedulers: list, monitor: Optional[str]
...
             )
-        elif isinstance(scheduler, optim.lr_scheduler._LRScheduler):
-            lr_schedulers.append({**default_config, "scheduler": scheduler})
         else:
-            raise ValueError(f'The provided lr scheduler "{scheduler}" is invalid')
+            lr_schedulers.append({**default_config, "scheduler": scheduler})
+
+        current_scheduler = lr_schedulers[-1]["scheduler"]
+        if not isinstance(current_scheduler, _SupportedLRScheduler):
+            raise ValueError(f"The provided lr scheduler `{current_scheduler.__class__.__name__}` is invalid.")
```
carmocca, line 309: I would remove this personally.

We had a ValueError there before to fill in the `else` condition, but it's not necessary anymore. Also checking the last one is a bit arbitrary.

rohitgr7: the first condition `if isinstance(scheduler, dict):` doesn't check for the scheduler protocol API in its block and the else condition we had won't be triggered if the user puts in an invalid scheduler even though all the keys are present. This avoids duplicate assertions and checks the recently added scheduler or not at one place.

carmocca: Still, we don't *need* to check this here. This function does not call any of the methods defined by "_SupportedLRScheduler".

Additionally, we don't *need* the scheduler to define this interface as the user can override the actually calls and do `scheduler.foo()` instead of `scheduler.step()`

rohitgr7: that is something @tchaton and @ananthsub proposed to have certain boundaries, for now, to make sure that basic API should be followed while implementing the even a custom scheduler and at the lowest level it should have a step, state_dict and load_state_dict methods defined else such a scheduler won't work within lightning.

carmocca: If the matter is with state_dict, then we can use the existing `_SupportsStateDict` to check it. But it should be in the checkpoint connector IMO.

And as I mentioned in my previous comment, we don't need to enforce `step()` anymore since this PR allows customization.

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -399,7 +398,7 @@ def _setup_model_and_optimizers(self, model: Module, optimizers: List[Optimizer]
         return self.model, [optimizer]
 
     def _setup_model_and_optimizer(
-        self, model: Module, optimizer: Optimizer, lr_scheduler: Optional[_LRScheduler] = None
+        self, model: Module, optimizer: Optimizer, lr_scheduler: Optional[LRSchedulerTypeUnion] = None
```
akihironitta, line 401: Does this typing mean LR scheduler customisation is not supported with DeepSpeed strategy? If so, should we note in the docs that only native PyTorch LR schedulers are supported in DeepSpeed?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -468,3 +469,14 @@ def hpc_save_path(folderpath: _PATH) -> str:
         ckpt_number = (max_suffix if max_suffix is not None else 0) + 1
         filepath = os.path.join(folderpath, f"hpc_ckpt_{ckpt_number}.ckpt")
         return filepath
+
+
+@runtime_checkable
+class _SupportsStateDict(Protocol):
```
akihironitta, line 475: Not sure how types are organised in this repo, but this type should be defined in `pl.utilities.types`?

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -577,6 +576,8 @@ def _reload_dataloader_state_dict_manual(dataloader: DataLoader, state_dict: Dic
     # In manual mode, we don't wrap the user objects with `CaptureMapDataset` or `CaptureIterableDataset`
     # therefore, we need to reload the states manually.
 
+    from pytorch_lightning.trainer.connectors.checkpoint_connector import _SupportsStateDict
+
```
akihironitta, line 580: nit: Shouldn't this import statement be at the top?

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -656,6 +646,8 @@ def __accumulate_state(self, sampler_state: Dict[str, Any]) -> None:
 
     def _store_sampler_state(self) -> None:
         """This function is used to extract the sampler states if any."""
+        from pytorch_lightning.trainer.connectors.checkpoint_connector import _SupportsStateDict
+
```
akihironitta, line 650: nit: Shouldn't this import statement be at the top?

rohitgr7: was causing some issue while running tests.. not sure why.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
rohitgr7: this isn't actually a type, used to check `isinstance`. I think since it's somewhat related to checkpointing API, so added it here and @carmocca suggested it too.

pytorch_lightning/core/optimizer.py
```diff
@@ -298,15 +302,21 @@ def _configure_schedulers_automatic_opt(schedulers: list, monitor: Optional[str]
...
             )
-        elif isinstance(scheduler, optim.lr_scheduler._LRScheduler):
-            lr_schedulers.append({**default_config, "scheduler": scheduler})
         else:
-            raise ValueError(f'The provided lr scheduler "{scheduler}" is invalid')
+            lr_schedulers.append({**default_config, "scheduler": scheduler})
+
+        current_scheduler = lr_schedulers[-1]["scheduler"]
+        if not isinstance(current_scheduler, _SupportsStateDict):
+            raise ValueError(f"The provided lr scheduler `{current_scheduler.__class__.__name__}` is invalid.")
```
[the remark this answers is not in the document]
rohitgr7: how about `... is invalid. Should have 'state_dict' and 'load_state_dict' methods be defined`?

pytorch_lightning/strategies/deepspeed.py
rohitgr7: lightning handles the deepspeed scheduler: https://github.com/PyTorchLightning/pytorch-lightning/blob/98ea79b8b07fd30ed718d28052d1bd8a586d3796/pytorch_lightning/strategies/deepspeed.py#L481-L487

pytorch_lightning/core/optimizer.py
tchaton, line 310:
```suggestion
        if not isinstance(current_scheduler, _SupportsStateDict):
            raise ValueError(f"The provided lr scheduler `{current_scheduler.__class__.__name__}` is invalid.")
         if not isinstance(current_scheduler, _LR_Scheduler_Protocol) and is_overriden("lr_scheduler_step", self.lightning_module):
            raise ValueError(f"The provided lr scheduler doesn't follow the PyTorch LR Scheduler Protocol and you should override the `lr_scheduler_step` hook of the LightningModule with your own logic.")
```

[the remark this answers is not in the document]
akihironitta: Sounds good :]

pytorch_lightning/strategies/deepspeed.py
akihironitta: Thanks for the reference. As the customisation is supported with DeepSpeed, isn't the type `LRSchedulerTypeUnion` too tight? I might be missing something, but it should probably be something looser such as `_SupportsStateDict` instead of `LRSchedulerTypeUnion`, no? Same applies for the type of `LightningModule.lr_scheduler_step`'s `scheduler`.

https://github.com/PyTorchLightning/pytorch-lightning/blob/65bda5f4f1f0e898cb710ea762a7c89eb81dcb8f/pytorch_lightning/utilities/types.py#L96

rohitgr7: these types are used for LightningCLI to generate docs for `help`. If you notice, there are separate types defined for _LRScheduler and ReduceLROnPlateau but they are not used here for the same reason. I asked Carlos, he suggested leaving it for now and in case this comes up for a CLI request, we will update this later.

pytorch_lightning/core/optimizer.py
rohitgr7: there is no `_LR_Scheduler_Protocol`, using the existing `_SupportsStateDict` since with custom hook, we don't force users to have the `step` method now.

pytorch_lightning/strategies/deepspeed.py
akihironitta: Ah, that totally makes sense. Thanks for the clarification!

pytorch_lightning/core/lightning.py
```diff
@@ -1493,6 +1493,42 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
             optimizer, gradient_clip_val=gradient_clip_val, gradient_clip_algorithm=gradient_clip_algorithm
         )
 
+    def lr_scheduler_step(
+        self,
+        scheduler: LRSchedulerTypeUnion,
+        optimizer_idx: Optional[int] = None,
+        metric: Any = None,
```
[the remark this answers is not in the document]
rohitgr7: it can be anything as long as the user has control over it.
https://github.com/PyTorchLightning/pytorch-lightning/pull/10249#discussion_r772383389

pytorch_lightning/core/optimizer.py
tchaton: Hey @rohitgr7.

Yes, I meant we should implement a protocol for PyTorch LR Scheduler and if the user scheduler doesn't match and they haven't overridden the new hook, it should break.

rohitgr7: ah! good edge case 👍 thank you

pytorch_lightning/core/optimizer.py
```diff
@@ -427,3 +430,18 @@ def zero_grad(self, set_to_none: Optional[bool] = False) -> None:
...
     def __repr__(self) -> str:
         return "No Optimizer"
+
+
+@runtime_checkable
+class _SupportsLRScheduler(Protocol):
+    """This class is used to detect if a learning rate scheduler is supported for default configuration using
+    `isinstance(obj, _SupportsLRScheduler)`."""
+
+    def step(self, *args: Any, **kwargs: Any) -> None:
```
tchaton, line 440: Can we make this one more strict? It should raise an expectation with timm schedulers.

rohitgr7: runtime_checkable doesn't check for types, just check the presence of the methods.
ref: https://docs.python.org/3/library/typing.html#typing.runtime_checkable

I think we should rather check whether the scheduler is an instance of native PT schedulers or not. if it's not, then the user should always override the new hook.

pytorch_lightning/utilities/auto_restart.py
awaelchli: must be a circular import, looks ok to me. could add a comment

carmocca: Or move it to `types.py`. It would solve the circular dependency

pytorch_lightning/core/optimizer.py
```diff
@@ -325,9 +329,29 @@ def _configure_schedulers_manual_opt(schedulers: list, monitor: Optional[str]) -
...
+                f"The provided lr scheduler `{scheduler.__class__.__name__}` is invalid."
+                " It should have `state_dict` and `load_state_dict` methods defined."
+            )
+
+        if not isinstance(scheduler, LRSchedulerTypeTuple) and not is_overridden("lr_scheduler_step", model):
+            raise MisconfigurationException(
+                f"The provided lr scheduler `{scheduler.__class__.__name__}` doesn't follow the PyTorch LR Scheduler"
+                " Protocol. You should override the `LightningModule.lr_scheduler_step` hook with your own logic if"
+                " you are using a custom LR scheduler."
+            )
```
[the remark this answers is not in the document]
rohitgr7: couldn't find a way to check the argument types in case it has a step method. https://github.com/PyTorchLightning/pytorch-lightning/pull/10249#discussion_r778752315, so I thought maybe let users override this hook because we can't for sure say whether default implementation will work or not. is there any other way around?

pytorch_lightning/core/lightning.py
rohitgr7: yes: https://github.com/PyTorchLightning/pytorch-lightning/pull/11247

pytorch_lightning/core/optimizer.py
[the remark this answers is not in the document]
carmocca: The other option is doing

```python
try:
    lightning_module.lr_scheduler_step()
except TypeError as e:
    if "step()" in str(e):
            f"The provided lr scheduler `{scheduler.__class__.__name__}` doesn't follow the PyTorch LR Scheduler"
            " API. You should override the `LightningModule.lr_scheduler_step` hook with your own logic if"
            " you are using a custom LR scheduler."
        ) from e
    raise
```

Maybe also check `AttributeError: .* object has no attribute 'step'`

This would be similar to #10719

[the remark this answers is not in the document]
rohitgr7: yeah, thought of doing this.. but wasn't sure if it's clean...
why didn't timm just use PyTorch scheduler 😄 

will update.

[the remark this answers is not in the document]
rohitgr7: @carmocca after thinking a bit, I think keeping it as it is should be okay. The reason is if someone copies the exact implementation of let's stay StepLR without subclassing LR_Scheduler then it won't work anyway because it uses its superclass methods too to work. Raising an error during the actual step might be too late for cases where it happens at the epoch level. We can consider it like this: if someone's using native schedulers then Lightning will handle it by default else one has to override this hook. One case as you mentioned can be user copying the whole implementation including the superclass methods and not subclassing might be very very rare, and I don't think anyone might be doing that plus if someone chooses to do it anyways then we can't help since we don't do any such checks for Optimizer or any other PyTorch components in a similar case.

pytorch_lightning/utilities/auto_restart.py
rohitgr7: done!

pytorch_lightning/core/lightning.py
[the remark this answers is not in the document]
carmocca: Applied

pytorch_lightning/core/lightning.py
carmocca: Passing always int for now with `or 0`, we do the type narrowing in another PR

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -454,11 +454,12 @@ def _update_learning_rates(
...
                 # update LR
-                if lr_scheduler["reduce_on_plateau"]:
-                    lr_scheduler["scheduler"].step(monitor_val)
-                else:
-                    lr_scheduler["scheduler"].step()
-
+                self.trainer._call_lightning_module_hook(
+                    "lr_scheduler_step",
+                    lr_scheduler["scheduler"],
+                    lr_scheduler["opt_idx"] or 0,
```
rohitgr7, line 460: will it be `None` in any case? @carmocca 
https://github.com/PyTorchLightning/pytorch-lightning/blob/b56d8677ad0ff8513e566334f4a78a24b88480c3/pytorch_lightning/core/optimizer.py#L345-L358

pytorch_lightning/core/optimizer.py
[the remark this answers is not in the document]
carmocca: Consider this case, adapted from your tests

```python
def test_user_is_forced_to_override_unnecesarily():

    class CustomScheduler:
        def __init__(self, optimizer):
            self.optimizer = optimizer

        def step(self):
            ...

        def state_dict(self):
            ...

        def load_state_dict(self, state_dict):
            ...

    class CustomBoringModel(BoringModel):
            opt = torch.optim.SGD(self.parameters(), lr=1e-2)
            lr_scheduler = CustomScheduler(opt)
            return {"optimizer": opt, "lr_scheduler": lr_scheduler}

    model = CustomBoringModel()
    model.trainer = Trainer()
    _init_optimizers_and_lr_schedulers(model)
```

This fails unnecessarily. Forces you to override the method when the base method would work.

If you prefer not to do this for now, or leave this for a follow-up, I'm fine with that. Just wanted to point out that this would be a source of boilerplate for that small percentage of users.

carmocca (commented):
I've simplified two of the tests

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: Nope, I added it thinking mypy would complain

But it doesn't at the moment, so we can remove it.

```suggestion
                    lr_scheduler["opt_idx"],
```

It should since `lr_scheduler.opt_idx` is an `Optional[int]` because it doesn't know the `opt_idx` was set.

pytorch_lightning/core/optimizer.py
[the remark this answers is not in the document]
rohitgr7: I see it as a very rare case. If there's a way to test it during init then it would be okay.. but my only concern regarding adding the exception during the actual step might be too late.

tests/trainer/optimization/test_optimizers.py
```diff
@@ -679,3 +680,128 @@ def on_save_checkpoint(self, checkpoint):
     model.training_epoch_end = None
     trainer.fit(model)
     assert model.on_save_checkpoint_called
+
+
+def test_lr_scheduler_step_hook(tmpdir):
+    """Test that custom lr scheduler works and `lr_scheduler_step` is called at appropriate time."""
```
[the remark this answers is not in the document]
rohitgr7: updated!

carmocca (approved):
Epic!

pytorch_lightning/core/optimizer.py
```diff
@@ -325,9 +329,27 @@ def _configure_schedulers_manual_opt(schedulers: list, monitor: Optional[str]) -
...
             lr_schedulers.append({**default_config, "scheduler": scheduler})
+
     return lr_schedulers
 
 
+def _validate_scheduler_api(lr_schedulers: List[Dict[str, Any]], model: "pl.LightningModule") -> None:
+    for scheduler_config in lr_schedulers:
+        scheduler = scheduler_config["scheduler"]
+        if not isinstance(scheduler, _SupportsStateDict):
+            raise ValueError(
```
kaushikb11, line 340: Should this be `MisConfigurationError`?

kaushikb11 (approved):
Love it!

pytorch_lightning/core/optimizer.py
carmocca: No: #10918

But maybe it should be `TypeError`, not `ValueError`

rohitgr7: good catch! updated.

PR #10272, opened 2021-10-30
PoC: LightningSystem

tchaton:
## What does this PR do

This is a weekend PoC to open a discussion and get feedbacks. Just exploration for now 🤗

This PR introduces a LightningSystem class, subclassing the LightningModule. At the difference of the LightningModule, it is more restrictive to force right patterns for production.
This forces the Lightning Module to be a vessel for training and shouldn t be while serving the model.

```py
    class AbstractGAN(ABC):

        generator: nn.Module
        discriminator: nn.Module

    class Model(AbstractGAN, nn.Module):
        def __init__(self, dim: int):
            super().__init__()
            self.generator = nn.Linear(dim, 1)
            self.discriminator = nn.Linear(1, 1)

        def forward(self, x):
            return self.generator(x)

    class GANSystem(LightningSystem):
        def __init__(self, module_cls, *args, **kwargs):
            super().__init__()

            self.gan = module_cls(*args, **kwargs)
            self.automatic_optimization = False

            # Implement a GAN there.
            pass

            opt_g = torch.optim.Adam(self.gan.generator.parameters(), lr=0.01)
            opt_d = torch.optim.Adam(self.gan.discriminator.parameters())
            return [opt_g, opt_d], []

        def train_dataloader(self):
            return DataLoader(RandomDataset(2, 2))

    system = GANSystem(Model, 5)
    trainer = Trainer(fast_dev_run=False, max_epochs=1)
    trainer.fit(system)
    trainer.save_checkpoint(p)
    model = load_from_checkpoint(Model, p)
```

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/core/system.py
+++ b/pytorch_lightning/core/system.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """The LightningSystem - an LightningModule """
+import functools
 from collections import OrderedDict
-from typing import Optional, Union
+from typing import Callable, Dict, IO, Optional, Type, Union
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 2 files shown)


maximsch2:
I like it! We have a few patterns internally, both using LightningModule as all-in-one and a training wrapper. An older system uses things as all-in-one as it seemed pointless to introduce another layer of abstraction, but with the experience of using it, I think it's right for us to encourage people to do the split. There can be a bunch of things that are used during training (e.g. metrics most prominently) that are not useful for inference/deployment and keeping them on the LightningModule makes things more bloated when you script them if you try to deploy the whole module.

A few questions/potential annoying issues:
1. Where does the loss belong here now? I'm assuming we would encourage that to be not part of the underlying model and be a field in the `system`? 
2. What do we do with logit/probabilities split? For a simple classifier, during training we want it to return `logits` to be able to use a faster and more numerically stable loss functions that operate on logits directly. During deployment though, we need to attach an output gate on top of those logits (e.g. softmax/sigmoid). This is simple thing, but having to do it all the time, I think we should make it clear what's the canonical way to do it.

ananthsub:
I am strongly opposed to this.

- The motivation for this is not clear from the PR? Why do we want to introduce a new abstraction? 
- Extending the LightningModule and then breaking behaviors related to the LightningModule extending nn.Module violates base class/child class expectations and responsibilities. 
- It creates uncertainty as to whether the LightningSystem and LightningModule are both supported for training. Is LightningModule being phased out? If not, why not?

maximsch2:
@ananthsub , I think we can separate two things: 1) do we want to encourage people to split out the model that is being trained from the LightningModule that is doing training (IMO yes, but let's discuss) and then 2) how should we do it? It might be just reworking documentation/examples or introducing a layer of code-enforcement like this PR suggest. 

I don't have a strong opinion on the 2nd point and maybe you are right that we shouldn't introduce another layer here - instead of failing things when LightningSystem is used "incorrectly", we can move the same code to LightningModule and send out warnings that it's used "suboptimally"?

aazzolini:
Hmm not sure i understand this . Is this LightningModuleV2? Why are we sub-classing LightningModule?
If we want to make it more restrictive , then maybe this is really an opportunity to simplify the implementation as well.

aazzolini:
In general I would strongly prefer to simplify the existing LightningModule rather than introduce yet another concept. It will just create more mess and will be increasilngly harder for users to know what they should do when using Lightingn.

LightningModule
LightingLite
LightningSystem.

What am I supposed to do with all this?

aazzolini:
I believe that there's no magic to good production. Ands it's to make APIs more stable over time. Which is the opposite of adding new experimental APis every week :)

oliland:
Some context to this PR: I provided some feedback on Slack:

> I really like the idea of distinguishing between “Models” and “Systems”. It makes sense to me that a system is a composition of a model + an optimization strategy. But I feel as though subclassing pl.LightningModule is the wrong approach to this. In my work I take a compositional approach to further separate lightning from my model:

```
class System(pl.LightningModule):
    def __init__(self, model):
        self.model = model
    
    # configure optimizers etc.

# initialize my model
model = MyModel()

# wrap my model with a system to train it
system = System(model)

# Train the system
trainer.fit(system)
```
> The benefit of this composition is that I can deploy MyModel on an embedded device without the need for a lightning dependency. Lightning installs a ton of stuff that isn’t really appropriate for embedded systems (like TensorBoard), so I feel that decoupling the model from the system makes sense.

> This is kindof supported by lightning if you’re willing to do some plumbing, but I’d like to see better support for it - for example, having the option to separately save the system parameters from the model parameters.

Hopefully this explains some of the context as to why decoupling "models" and "systems" might make more sense for some people.

oliland:
As for my feedback on this PR:  The initial decision to subclass `nn.Module` is clever design because it ensures that the entire system state is trivially serializable. However, this has come at an additional cost, which is that the system state parameters and the model state parameters are now tangled together in a way that isn't explicit enough for PyTorch Lightning effectively disambiguate between system (optimizer) parameters and model parameters.

When I deploy a model, I really just want to deploy the model parameters, and not have to ship around whatever state my optimizer was in during training time.

In this PR, it appears that the proposed checkpoint now only saves the model parameters, which allows you to deserialize your model from the checkpoint:
```
trainer.save_checkpoint(p)
model = load_from_checkpoint(Model, p)
```
But doesn't this break a really important use case of checkpoints, which is that if your optimizers have some non-trivial state (such as a variable learning rate), that this state is no longer resumable?

Given that our model parameters and our system parameters are neatly separated in this abstraction, we could instead add a new method, `trainer.save_model_state_dict(filename)` which saves the model parameters (and nothing else) for deployment. This can then be loaded and deserialized by `Model` without having to install Lightning on an embedded system.

tchaton:
Hey @oliland 

> In this PR, it appears that the proposed checkpoint now only saves the model parameters, which allows you to deserialize your model from the checkpoint:

This is not accurate. The checkpoint is generated by the trainer and therefore contains all the information required if you want to use your system afterwards.

But more importantly, this enables you to not use Lightning on the embedded systems if you need to do so.

This enables you to have both LightningModule and not Lightning usable from the same checkpoint.

Best,
T.C

Issue #10270, opened 2021-10-30
[RFC] Improve Lightning for production


tchaton:
## 🚀 Feature

Lot of users are complaining for PyTorch Lightning to be un-usable in production due to the extra dependencies, code, etc...

Furthermore, Lightning makes a bad job at conveying the best practices behind production around the LightningModule usage.

The LightningModule should be seen as vessel for training, users call it a System.

Here are some proposals to improve best practices around Lightning production usage.

[] Start converting examples (docs, example) to the System format.
[] Dedicate page to best practices for PyTorch Lightning in production.
[] Improve checkpointing / reloading with the System design
[] Have a `pip install pytorch-lightning[prod]` which is the smallest possible (no extra libraries like TensorBoard, etc..)
[] Introduce a System design which is not a nn.Module and accept (need to be explored)

Any extra bullet points are welcome.

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

</sub>

cc @borda

lantiga:
Hey @tchaton, great discussion! My 2 cents here.

I think we need to have the concept of “production” defined precisely. Exporting a model to ONNX or TorchScript alone is unfortunately a far cry from deploying to production in many circumstances, so there’s typically a good chunk of (typically) Python code you need (or you need to replace) for full inference. However it may not even make sense to execute it on the same machine in some applications.

[...]

PR #10276, opened 2021-10-31
[WIP] Trim PyTorch Lightning dependencies

tchaton:
## What does this PR do?

> The lightweight PyTorch wrapper for high-performance AI research. Scale your models, not the boilerplate.

PyTorch Lightning claims to be a light wrapper over PyTorch, however it has been taking more and more dependencies making it not suitable for production. Users are starting to complain Lightning isn't right for the task.

Find the Reddit Thread about users complaning about this. https://www.reddit.com/r/MachineLearning/comments/qgqq07/d_is_pytorch_lightning_production_ready/

On Slack: https://pytorch-lightning.slack.com/archives/CU5PF2E2J/p1635596286018500

This PR makes only torch the unique dependency to PyTorch Lightning.

```py
pip install `pytorch-ligtning[base]` -> install only torch and should work with LightningModule + weight loading + forward
pip install pytorch-ligtning -> No change, current behaviour.

```

As a user, I want to be able to use PyTorch Lightning & PyTorch and this code to work.

```py
import os
import tempfile

import torch

    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

def run():
    with tempfile.TemporaryDirectory() as tmpdir:
        checkpoint_path = os.path.join(tmpdir, "model.pt")
        state_dict = {"state_dict": BoringModel().state_dict()}
        torch.save(state_dict, checkpoint_path)
        BoringModel.load_from_checkpoint(checkpoint_path)


if __name__ == "__main__":
    run()

```

Parts of #10270

TODO:
- [ ] Update CI to release the `base` version with only the torch dep
- [ ] Update CI to release the normal version with all default dependencies to prevent BC from pypi
- [ ] Add test with the prod version
- [ ] Make sure typing is properly applied
- [ ] Add tests to make sure prod is properly setup.

I believe we can make this change without introducing any breaking change either on pypi or local dev. I will update this when the PR is ready. 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/utilities/imports.py
+++ b/pytorch_lightning/utilities/imports.py
@@ -12,13 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """General utilities."""
+import functools
 import importlib
 import operator
 import os
 import platform
 import sys
 from importlib.util import find_spec
-from typing import Callable
+from typing import Callable, List, Union
 
 import pkg_resources
 import torch
@@ -82,6 +83,7 @@ _DEEPSPEED_AVAILABLE = _module_available("deepspeed")
 _FAIRSCALE_AVAILABLE = not _IS_WINDOWS and _module_available("fairscale.nn")
 _FAIRSCALE_OSS_FP16_BROADCAST_AVAILABLE = _FAIRSCALE_AVAILABLE and _compare_version("fairscale", operator.ge, "0.3.3")
 _FAIRSCALE_FULLY_SHARDED_AVAILABLE = _FAIRSCALE_AVAILABLE and _compare_version("fairscale", operator.ge, "0.3.4")
+_FSSPEC_AVAILABLE: bool = _module_available("fsspec")
 _GROUP_AVAILABLE = not _IS_WINDOWS and _module_available("torch.distributed.group")
 _HOROVOD_AVAILABLE = _module_available("horovod.torch")
 _HYDRA_AVAILABLE = _module_available("hydra")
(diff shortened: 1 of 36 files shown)


Borda:
if I see it correctly you just push dependencies from install to runtime?

pytorch_lightning/utilities/imports.py
```diff
@@ -110,6 +116,34 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
     _IPU_AVAILABLE = False
 
 
+def requires(module_paths: Union[str, List]):
```
Borda, line 119: is it the same we have in Flash?

requirements/default.txt
```diff
@@ -0,0 +1,11 @@
+# the default package dependencies
```
Borda, line 1: is default some keyword in extras?

Borda:
> This PR makes only torch the unique dependency to PyTorch Lightning.
> 
> ```python
> pip install `pytorch-ligtning[base]` -> install only torch and should work with LightningModule + weight loading + forward
> pip install pytorch-ligtning -> No change, current behaviour.
> ```

do you know how the extra works, is it a full alternative to install the general? or it is just an addition?
https://stackoverflow.com/questions/41268863/difference-between-extras-require-and-install-requires-in-setup-py

Borda:
just thinking what dependencies you found as problematic?
```
numpy>=1.17.2
torch>=1.6
future>=0.17.1  # required for builtins in setup.py
tqdm>=4.41.0
PyYAML>=5.1
fsspec[http]>=2021.05.0, !=2021.06.0
tensorboard>=2.2.0
torchmetrics>=0.4.1
pyDeprecate==0.3.1
packaging>=17.0
typing-extensions
```
I may say only `tensorboard` as it pulls TensorFlow with it... but others are basic and present in most env already...

tchaton:
Dear @Borda,

Aligning on PyTorch, I would keep the following:

```py
torch>=1.6
future>=0.17.1  # required for builtins in setup.py
packaging>=17.0
typing-extensions ? 
```

justusschock:
@tchaton i like the idea in general. Since during prediction we also write stuff to disk sometimes, does that mean we need fsspec as well?

I think we don't need to include typing-extensions since they are only necessary for old python versions. 
How do we deal with pydeprecate? That's just a very thin wrapper of @Borda . I think since we moved out metrics, we don't need it anymore but currently it is used in the codebase. 

Also we need to deactivate progressbar per default if tqdm is not present :)

Borda:
> I think since we moved out metrics, we don't need it anymore but currently it is used in the codebase.

yes, we can drop it if there is no other need for it... :]

tchaton:
I am closing this PR as it doesn't seem feasible with setup.py. The only option would be to release another package called `pytorch_lightning_trimmed`.
@Borda Should we consider this ?

PR #10279, opened 2021-10-31
Add custom dataloader support with Lite 

tchaton:
## What does this PR do?

This PR enables the following Custom DataLoader not to crash.

This would work

```py
class CustomDataLoader(DataLoader):
    def __init__(self, value: int, *args, **kwargs):
        super().__init__(range(value), *args, **kwargs)
```

```py
class LiteWithCustomDataLoader(LightningLite):
    def run(self):
        # This doesn't fail as the context manager would save all the arguments provided
        # to the dataloaders.
        dataloader = CustomDataLoader(2, batch_size=2)
        self.setup_dataloaders(dataloader)

LiteWithCustomDataLoader().run()
```

Would we failed as the value has been captured.

```py
dataloader = CustomDataLoader(2, batch_size=2)
lite_dataloader = lite.setup_dataloaders(dataloader)
``` 

This would enable to remove this ugliness from PyTorch Geometric: https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/loader/dataloader.py#L74 if they use LightningLite.

This would be added to Lightning in another PR.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +100,38 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
         return self._device
 
-    def __iter__(self) -> Union[Iterator[Any], Generator[Any, None, None]]:
-        iterator = super().__iter__()
-        if self._device is None:
-            return iterator
-
-        for item in iterator:
-            yield move_data_to_device(item, self._device)
+    def __iter__(self) -> "_LiteDataLoader":
```
awaelchli, line 123: I think we can keep the previous code and just iterate over `self._dataloader.` There should be no need to implement the next method. We can just remove the subclassing of DataLoader. The rest of the changes look good

tchaton: Done !

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
         self._device = device
+        self._dataloader_iter: Optional[Iterator] = None
 
     @property
     def device(self) -> Optional[torch.device]:
         return self._device
 
     def __iter__(self) -> Union[Iterator[Any], Generator[Any, None, None]]:
-        iterator = super().__iter__()
+        dataloader_iter = iter(self._dataloader)
```
awaelchli, line 172: the name here can be kept general, after all, the PR is aiming at supporting custom non-dataloader iterators.

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
             device: The device to which the data should be moved. By default the device is `None` and no data
                 transfers will be made (identical behavior as :class:`~torch.utils.data.DataLoader`).
-            **dl_kwargs: Accepts all arguments that the PyTorch :class:`~torch.utils.data.DataLoader` accepts.
         """
-        super().__init__(**dl_kwargs)
+        super().__init__()
+        self.__dict__.update(getattr(dataloader, "__dict__", {}))
+        self._dataloader = dataloader
         self._device = device
+        self._dataloader_iter: Optional[Iterator] = None
```
awaelchli, line 165:
```suggestion
```
unused variable

tests/lite/test_lite.py
```diff
@@ -164,6 +164,59 @@ def test_setup_dataloaders_return_type():
...
+        def __init__(self, value: int = 2, *args, **kwargs):
+            self.value = value
+            super().__init__(range(value), *args, **kwargs)
+
+    dataloader = CustomDataLoader(2, batch_size=2)
+
+    # single dataloader
+    lite_dataloader = lite.setup_dataloaders(dataloader)
+    assert lite_dataloader._dataloader
+    assert lite_dataloader._dataloader_iter is None
```
awaelchli, line 181:
```suggestion
```
it's always true

tests/lite/test_lite.py
```diff
@@ -164,6 +164,59 @@ def test_setup_dataloaders_return_type():
...
+        def __init__(self, range, *args, **kwargs):
+            self.range = range
+            super().__init__(range, *args, **kwargs)
+
+    dataloader = CustomDataLoader2(range(2), batch_size=2)
+
+    # single dataloader
+    lite_dataloader = lite.setup_dataloaders(dataloader)
+    assert lite_dataloader._dataloader
+    assert lite_dataloader._dataloader_iter is None
```
awaelchli, line 196:
```suggestion
```

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+    cls._old_init = cls.__init__
+    cls.__init__ = _wrap_init(cls.__init__)
+
+
+def _disable_class(cls: Type[Any]) -> None:
+    cls.__init__ = cls._old_init
+
+
+@contextmanager
+def _replace_dataloader_init_function() -> Generator:
```
awaelchli, line 142: init is a method, not a function

pytorch_lightning/lite/lite.py
```diff
@@ -233,17 +238,18 @@ def _setup_dataloader(
...
-        if isinstance(self._strategy, TPUSpawnPlugin):
-            dataloader = DataLoader(**kwargs)
-        else:
-            dataloader = _LiteDataLoader(device=device, **kwargs)
-
+        dataloader_kwargs = TrainerDataLoadingMixin._get_dataloader_init_kwargs(dataloader, sampler)
+        try:
+            dataloader = type(dataloader)(**dataloader_kwargs)
+        except TypeError:
+            dataloader_kwargs.pop("dataset")
```
awaelchli, line 245: why are we re-trying to populate the arguments without dataset?

awaelchli (approved):
I am not sure on this one, but I think we should not do it. This is diverging from how Lightning handles the dataloader preparation. Also, it only solves the problem partially. If the dataloader is created outside the run() method, there is nothing we can do (currently).

pytorch_lightning/lite/wrappers.py
tchaton: The PR doesn't support custom iterators. The PR adds support for subclassed dataloader.

pytorch_lightning/lite/wrappers.py
tchaton: Done !

tests/lite/test_lite.py
tchaton: Done !

tests/lite/test_lite.py
tchaton: Done !

pytorch_lightning/lite/wrappers.py
tchaton: Done !

pytorch_lightning/lite/lite.py
tchaton: Because a custom dataloader might not take a dataset. Actually it is pretty likely if you subclass it.

Example:

```py
class CustomDataLoader(DataLoader):

    def __init__(self, start, end, *args, **kwargs):
        super().__init__(range(start, end), *args, **kwargs)
```

When doing the re-instantiation, there would be 2 datasets being provided to the super method.

awaelchli: There are infinitely many ways one can construct subclasses and call super() with different arguments that don't match the init. So this is an extreme assumption

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+    return set(subclass_list)
+
+
+def _enable_class(cls: Type[Any]) -> None:
+    cls._old_init = cls.__init__
+    cls.__init__ = _wrap_init(cls.__init__)
+
+
+def _disable_class(cls: Type[Any]) -> None:
+    cls.__init__ = cls._old_init
```
ethanwharris, line 138: Maybe add `del(cls._old_init)` here to fully tear down

ethanwharris (approved):
LGTM, small comment

pytorch_lightning/lite/wrappers.py
tchaton: Good catch !

kaushikb11 (approved):
LGTM!

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+    for subclass in _get_all_subclasses(DataLoader):
+        _enable_class(subclass)
+    yield
+    for subclass in _get_all_subclasses(DataLoader):
+        _disable_class(subclass)
+
+
+class _LiteDataLoader:
+    def __init__(self, dataloader: Iterable, device: Optional[torch.device] = None) -> None:
+        """The LiteDataLoader is an extension of an Iterator. It would move move the data to the device
```
kaushikb11, line 154:
```suggestion
        """The LiteDataLoader is an extension of an Iterator. It would move the data to the device
```

pytorch_lightning/lite/lite.py
```diff
@@ -208,7 +213,7 @@ def setup_dataloaders(
 
     def _setup_dataloader(
         self, dataloader: DataLoader, replace_sampler: bool = True, move_to_device: bool = True
-    ) -> Union[Iterable, DataLoader]:
+    ) -> Iterable:
```
carmocca, line 216: `isinstance(DataLoader, Iterable)` returns False so I'm not sure these typing changes are correct.

pytorch_lightning/lite/lite.py
```diff
@@ -233,17 +238,18 @@ def _setup_dataloader(
...
-        device = self.device if move_to_device else None
-        if isinstance(self._strategy, TPUSpawnPlugin):
-            dataloader = DataLoader(**kwargs)
-        else:
-            dataloader = _LiteDataLoader(device=device, **kwargs)
-
+        dataloader_kwargs = TrainerDataLoadingMixin._get_dataloader_init_kwargs(dataloader, sampler)
+        try:
+            dataloader = type(dataloader)(**dataloader_kwargs)
+        except TypeError:
```
carmocca, line 244: Relying on an exception is very ugly. Whatever this intends to resolve, should have been implemented inside `TrainerDataLoadingMixin._get_dataloader_init_kwargs`

pytorch_lightning/lite/lite.py
```diff
@@ -233,17 +238,18 @@ def _setup_dataloader(
                 )
             sampler = self._get_distributed_sampler(dataloader, **self._strategy.distributed_sampler_kwargs)
 
-        kwargs = TrainerDataLoadingMixin._get_dataloader_init_kwargs(dataloader, sampler)
-        device = self.device if move_to_device else None
-        if isinstance(self._strategy, TPUSpawnPlugin):
```
carmocca, line 238: why is this check no longer required?

pytorch_lightning/lite/lite.py
```diff
@@ -183,7 +188,7 @@ def setup(
 
     def setup_dataloaders(
         self, *dataloaders: DataLoader, replace_sampler: bool = True, move_to_device: bool = True
-    ) -> Union[DataLoader, List[DataLoader], Iterable]:
+    ) -> Union[Iterable, List[Iterable]]:
```
carmocca, line 191: Same here

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+    for subclass in _get_all_subclasses(DataLoader):
+        _enable_class(subclass)
+    yield
+    for subclass in _get_all_subclasses(DataLoader):
+        _disable_class(subclass)
+
+
+class _LiteDataLoader:
+    def __init__(self, dataloader: Iterable, device: Optional[torch.device] = None) -> None:
+        """The LiteDataLoader is an extension of an Iterator. It would move the data to the device automatically if
```
carmocca, line 154: This is not correct as it does not extend the Iterator class

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+    """This context manager is used to support custom :class:`~torch.utils.data.DataLoader."""
+    for subclass in _get_all_subclasses(DataLoader):
+        _enable_class(subclass)
+    yield
+    for subclass in _get_all_subclasses(DataLoader):
+        _disable_class(subclass)
+
+
+class _LiteDataLoader:
+    def __init__(self, dataloader: Iterable, device: Optional[torch.device] = None) -> None:
```
carmocca, line 153: Missing Union here

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
 
         Args:
+            dataloader: The current dataloader to be used.
             device: The device to which the data should be moved. By default the device is `None` and no data
                 transfers will be made (identical behavior as :class:`~torch.utils.data.DataLoader`).
-            **dl_kwargs: Accepts all arguments that the PyTorch :class:`~torch.utils.data.DataLoader` accepts.
         """
-        super().__init__(**dl_kwargs)
+        super().__init__()
+        self.__dict__.update(getattr(dataloader, "__dict__", {}))
```
carmocca, line 163: this should be

```python
if isinstance(dataloader, DataLoader)
  self.__dict__.update(dataloader.__dict__)
```

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+        the device is specified.
 
         Args:
+            dataloader: The current dataloader to be used.
             device: The device to which the data should be moved. By default the device is `None` and no data
                 transfers will be made (identical behavior as :class:`~torch.utils.data.DataLoader`).
-            **dl_kwargs: Accepts all arguments that the PyTorch :class:`~torch.utils.data.DataLoader` accepts.
         """
-        super().__init__(**dl_kwargs)
+        super().__init__()
```
carmocca, line 162: Calling __init__ is not necessary here

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+        self.__dict__.update(getattr(dataloader, "__dict__", {}))
+        self._dataloader = dataloader
         self._device = device
 
     @property
     def device(self) -> Optional[torch.device]:
         return self._device
 
     def __iter__(self) -> Union[Iterator[Any], Generator[Any, None, None]]:
```
carmocca, line 171: Won't this get possibly overridden by the `__dict__.update` above?

carmocca (commented):
cc @awaelchli

pytorch_lightning/lite/lite.py
```diff
@@ -400,7 +406,7 @@ def _run_impl(self, run_method: Callable, *args: Any, **kwargs: Any) -> Any:
             return run_method(*args, **kwargs)
 
     def _run_with_sharded_context(self, run_method: Callable, *args: Any, **kwargs: Any) -> Any:
-        with self._strategy.model_sharded_context():
+        with self._strategy.model_sharded_context(), _replace_dataloader_init_method():
```
carmocca, line 409: I'd rather try and iterate on our existing dataloader argument logic than messing with the __init__

tests/lite/test_lite.py
```diff
@@ -164,6 +164,57 @@ def test_setup_dataloaders_return_type():
...
 
 
+def test_setup_custom_dataloaders():
+    """Test that the setup_dataloaders method returns the dataloaders wrapped as LiteDataLoader."""
+    lite = EmptyLite()
+
+    class CustomDataLoader(DataLoader):
+        def __init__(self, value: int = 2, *args, **kwargs):
+            self.value = value
+            super().__init__(range(value), *args, **kwargs)
```
carmocca, line 174: There's no reason for us to add support for this, as this is a bug in the DataLoader implementation.

Adding arbitrary `**kwargs` when opens the door to duplicate `dataset` arguments, but this should be fixed by the user, not by us.

The fix only works for the first argument and if the name is dataset, but this would still break for any of the other arguments

pytorch_lightning/lite/wrappers.py
```diff
@@ -100,27 +103,75 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
-        """The LiteDataLoader is an extension of the PyTorch :class:`~torch.utils.data.DataLoader` that adds
-        additional features such as moving the data to the device automatically.
+def _wrap_init(f: Callable) -> Callable:
+    @functools.wraps(f)
+    def wrapper(module: Any, *args: Any, **kwargs: Dict[str, Any]) -> None:
+        params = dict(inspect.signature(module._old_init).parameters)
+        params.pop("args")
+        params.pop("kwargs")
+        for init_name, init_arg in zip(params, args):
+            setattr(module, init_name, init_arg)
```
carmocca, line 113: This could overwrite the existing attribute, which could be a problem if they have different values

awaelchli:
@carmocca We can accumulate these smaller things like typing in #10297 and make separate PRs where needed as discussed in  #10329

pytorch_lightning/lite/lite.py
awaelchli: hopefully addressed in  #10625

pytorch_lightning/lite/lite.py
awaelchli: same here

pytorch_lightning/lite/wrappers.py
awaelchli: hopefully addressed in  #10625

pytorch_lightning/lite/wrappers.py
awaelchli: the `__*` methods don't get affected by direct update to `__dict__`

Issue #5345, opened 2021-01-04
Error with `load_from_checkpoint` 


bnaman50:
I am trying to fine-tune a language model and facing some issues with loading the model from the saved checkpoint. 

I have defined my own model which takes in `argparse.NameSpace` as the input as is suggested in the documentation [here](https://pytorch-lightning.readthedocs.io/en/latest/hyperparameters.html). 

```
class FineTuner(pl.LightningModule):
    def __init__(self, hparams):  # pass all the arguments
        super(FineTuner, self).__init__()
        self.hparams = hparams
```

After I am done with the fine-tuning, I try to load the model from saved checkpoint **in a separate file** as follows - 
```
from model_file import FineTuner
model = FineTuner.load_from_checkpoint(ckpt_path)
```

but it gives me the following error - 
`AttributeError: 'dict' object has no attribute 'tokenizer_name'`

I tried to look into the source code and found that the error is happening at this [line](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/core/saving.py#L192). 
I believe the error is occurring because the aforementioned line in `saving.py` file expects a `**dict` which is different from my model definition. 

**It would be great if you could solve this issue as I have seen many others are also having similar issues with model loading.** 

### For others who might still be looking for the answer, here is the workaround that I am using right now 
(but would still like to use the cleaner official version) - 
```
ckpt_path = abs_path('./path/to/ckpt/file')
checkpoint = torch.load(ckpt_path, map_location=lambda storage, loc: storage)
model = my_model(checkpoint['hyper_parameters']) #(better solution than explicitly providing all the hyperparameters)
model.load_state_dict(checkpoint['state_dict'])
```
If one were to look at the source code, you would notice that they are kind of doing the same steps but with some extra stuff ;) 
But hey, who does not like one-liners. 

I am using version version `1.0.4`.

tchaton:
Hey @bnaman50,

Thanks for reporting this behaviour.
For now, you could change your `FineTuner` class as follow:

```
class FineTuner(...):
    def __init__(self, *args, arg_1, ...,arg_n, kwarg_1=..., ..., kwarg_n=...,  **kwargs):
```

So extra parameters doesn't break the loading.

Best,
T.X

awaelchli:
[...]

Issue #4390, opened 2020-10-27
LightningModule.load_from_checkpoint with module_arguments is broken


lagph:
## 🐛 Bug
Saving a `LightningModule` whose constructor takes arguments and attempting to load using `load_from_checkpoint` errors with

```
TypeError: __init__() missing 1 required positional argument: 'some_param'
```

## Please reproduce using [the BoringModel and post here](https://colab.research.google.com/drive/1HvWVVTK8j2Nj52qU4Q4YCyzOm0_aLQF3?usp=sharing)

https://colab.research.google.com/drive/1fbzDt4mEjRQuy4L_DYkkZqBRhQCxZ6kP?usp=sharing

### Expected behavior
```python

class BoringModel(LightningModule):

    def __init__(self, some_param):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)
        self.some_param = some_param



model = BoringModel(some_param=10)

# Initialize a trainer
trainer = pl.Trainer(
    max_epochs=1, 
    progress_bar_refresh_rate=20
)

# Train the model ⚡
trainer.fit(model, train, val)
trainer.save_checkpoint("example.ckpt")

model = BoringModel.load_from_checkpoint("example.ckpt")
assert model.some_param == 10
```
### Environment

* CUDA:
	- GPU:
		- Tesla T4
	- available:         True
	- version:           10.1
* Packages:
	- numpy:             1.18.5
	- pyTorch_debug:     False
	- pyTorch_version:   1.6.0+cu101
	- pytorch-lightning: 1.0.0
	- tqdm:              4.41.1
* System:
	- OS:                Linux
	- architecture:
		- 64bit
	- processor:         x86_64
	- python:            3.6.9
	- version:           #1 SMP Thu Jul 23 08:00:38 PDT 2020

### Additional context

Part of the problem seems to stem from `checkpoint_connector.py`:
```python
        # add the module_arguments and state_dict from the model
        model = self.trainer.get_model()

        checkpoint["state_dict"] = model.state_dict()

        if model.hparams:
            if hasattr(model, "_hparams_name"):
                checkpoint[
                    LightningModule.CHECKPOINT_HYPER_PARAMS_NAME
                ] = model._hparams_name
            # add arguments to the checkpoint
            if OMEGACONF_AVAILABLE:
                checkpoint[LightningModule.CHECKPOINT_HYPER_PARAMS_KEY] = model.hparams
                if isinstance(model.hparams, Container):
[...]

Issue #4333, opened 2020-10-24
Saving / loading hparams type in checkpoint


chiragraman:
## 🐛 Bug

There seems to be an issue with saving or loading hyperparams type in checkpoints. Related to #924 (unrelated to #3998). 

## Please reproduce using [the BoringModel and post here](https://colab.research.google.com/drive/1HvWVVTK8j2Nj52qU4Q4YCyzOm0_aLQF3?usp=sharing)

Here is the BoringModel gist using the snippet from @awaelchli in #924 
 (running in colab has been unreliable, see #3998)
https://gist.github.com/chiragraman/c235f6f2a25b2432bde1a08ae6ed1b03

### Behavior

With my local master at 155f4e9a, I'm getting the following error: 

```
type of hparams <class 'pytorch_lightning.utilities.parsing.AttributeDict'>
class of hparams type AttributeDict
accessing hparams 1 no problem
GPU available: True, used: False
TPU available: False, using: 0 TPU cores
/home/chirag/miniconda3/envs/ml/lib/python3.8/site-packages/pytorch_lightning/utilities/distributed.py:45: UserWarning: GPU available but not used. Set the --gpus flag when calling the script.
  warnings.warn(*args, **kwargs)
/home/chirag/miniconda3/envs/ml/lib/python3.8/site-packages/pytorch_lightning/utilities/distributed.py:45: UserWarning: The dataloader, val dataloader 0, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 8 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  warnings.warn(*args, **kwargs)
/home/chirag/miniconda3/envs/ml/lib/python3.8/site-packages/pytorch_lightning/utilities/distributed.py:45: UserWarning: The dataloader, train dataloader, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 8 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  warnings.warn(*args, **kwargs)
Epoch 0: 100%|██████████████████| 2/2 [00:00<00:00, 427.58it/s, loss=2.280, v_num=11]
type of hparams <class 'pytorch_lightning.utilities.parsing.AttributeDict'>
class of hparams type AttributeDict
Traceback (most recent call last):
[...]

Issue #3998, opened 2020-10-08
Error in  `load_from_checkpoint` when LightningModule init contains 'hparams'


chiragraman:
#### What is your question?

Just pulled master today, and `load_from_checkpoint` no longer works. I am wondering if this is a backwards compatibility issue, or I need to do something differently to load now? I've provided the HEAD commits of my local repo I'm installing from in the environment section below.

#### Error in loading

Here's the error I'm getting now without changing any code:

```
File "/home/chirag/miniconda3/envs/ml/lib/python3.8/site-packages/pytorch_lightning/core/saving.py", line 153, in load_from_checkpoint
    model = cls._load_model_state(checkpoint, strict=strict, **kwargs)
  File "/home/chirag/miniconda3/envs/ml/lib/python3.8/site-packages/pytorch_lightning/core/saving.py", line 193, in _load_model_state
    model = cls(**_cls_kwargs)
TypeError: __init__() missing 1 required positional argument: 'hparams'
```
The lightning module has always had a member variable hparams, so everything should have been saved (and indeed worked before this pull) in accordance with the docs on stable: [https://pytorch-lightning.readthedocs.io/en/stable/hyperparameters.html#lightningmodule-hyperparameters](https://pytorch-lightning.readthedocs.io/en/stable/hyperparameters.html#lightningmodule-hyperparameters)

#### What's your environment?

```
❯ git rev-parse --short HEAD                                             
1d3c7dc8    # DOES NOT WORK

❯ git rev-parse --short HEAD@{1}                                          
90929fa4    # WORKS OKAY
```

rohitgr7:
can you put your `LightningModule` init code and `load_from_checkpoint` line here?

chiragraman:
Hi Rohit, 

Sure. Here's the init code:

```
class ProcessSystemBase(pl.LightningModule):

    def __init__(self, hparams: Namespace, builder: RecurrentBuilder) -> None:
        """ Initialize the Process module.

        Args:
            hparams             -- The hyperparameters for the experiment
            builder                -- The factory to build backbone components

        """
        super().__init__()

        ##----------------------------------------
        # Hack to combat lightning loading hparams as dicts
        # Refer issue #924 
        if isinstance(hparams, dict):
            hparams = Namespace(**hparams)
        ##----------------------------------------
        self.hparams = hparams
        self.components = builder.init_components(hparams)
```
[...]

PR #10280, opened 2021-10-31
Hyperparameters docs refresh

### LINKED ISSUE #9984: save_hyperparameters() saves all **kwargs arguments - which includes unpickable plugins for DDP
## 🐛 Bug

The call to save `self.save_hyperparameters()` (NOTE: no args provided) as described in the docs (see [here](https://pytorch-lightning.readthedocs.io/en/latest/common/hyperparameters.html#lightningmodule-hyperparameters
)) is by default saving the content of **kwargs to hparams. This is an issue because:

1. this is not matching the documentation (see the 'equivalent' call part)
2. one of the fields (I suspect introduced between 1.2.8 to 1.4.9?) is 'plugins' and there there is a DDP related object which cannot be pickled. This causes model checkpoints saved like this to be impossible to be loaded outside of a similar env (boh with PL loading functionality or the torch.load method).

### To Reproduce

Save any model according to the docs and try to load it in another script which runs afterwards (i.e. in a inference only script) not powered by PL

### Expected behavior

It should keep working as it did in version 1.2.8

### Environment

- PyTorch Lightning Version 1.4.9

awaelchli:
## What does this PR do?

A refresh of the docs around the `save_hyperparameters()` method. The current text is dated and does no longer accurately reflect the recommended, best possible ways to deal with hyperparameters. 
The `save_hyperparameters` method should not be seen as a magical feature. There is one way to use it and it must be cristal clear how to instantiate a model from a checkpoint.

1. Several examples had a "**kwargs" in their signature. While this can certainly be handled by our saving logic, it's not the first thing we want to recommend to users. It's a bad practice as unwanted objects could be accidentally getting pickled into the checkpoints, and make them potentially not able to load again.

2. It mentioned that one can save whole containers as `save_hyperparametes(conf)`. While this is possible, this should be avoided at all cost and is an artifact of a backward compatibility with the "old" way (pre 1.0) of saving hyperparameters. If one wants to save a  container of paramters (e.g. OmegaConf dict), then one should treat it like any other hyperaparameter and do `self.save_hyperparameters("conf")`. Otherwise, it causes too many problems, especially when the user does not know how to use it properly and later on finds that they can't reload their checkpoint with a simple `Model.load_from_checkpoint` as advertised in the docs. This has been a pain point for ages. Now, we will NO LONGER document this.

See all these issues from the past:
#3998, #4333, #4390, #5345

Fixes #9984
Should help with repeated questions asked on Slack

## Before submitting

## PR review

## Did you have fun?


docs/source/common/hyperparameters.rst
```diff
@@ -113,78 +113,75 @@ LightningModule hyperparameters
...
-                # save the config and any extra arguments
-                self.save_hyperparameters(conf)
-                self.save_hyperparameters(kwargs)
+load_from_checkpoint
+""""""""""""""""""""
 
-                self.layer_1 = nn.Linear(28 * 28, self.hparams.layer_1_dim)
-                self.layer_2 = nn.Linear(self.hparams.layer_1_dim, self.hparams.layer_2_dim)
-                self.layer_3 = nn.Linear(self.hparams.layer_2_dim, 10)
+LightningModule's that have hyperparameters automatically saved with :meth:`~pytorch_lightning.core.lightning.LightningModule.save_hyperparameters`
```
awaelchli, line 169:
```suggestion
LightningModules that have hyperparameters automatically saved with :meth:`~pytorch_lightning.core.lightning.LightningModule.save_hyperparameters`
```

docs/source/common/hyperparameters.rst
```diff
@@ -113,78 +113,75 @@ LightningModule hyperparameters
 Often times we train many versions of a model. You might share that model or come back to it a few months later
 at which point it is very useful to know how that model was trained (i.e.: what learning rate, neural network, etc...).
 
-Lightning has a few ways of saving that information for you in checkpoints and yaml files. The goal here is to
+Lightning has a standardized way saving that information for you in checkpoints and yaml files. The goal here is to
```
tchaton, line 116:
```suggestion
Lightning has a standardized way of saving that information for you in checkpoints and YAML files. The goal here is to
```

tchaton (approved):
LGTM !

docs/source/common/hyperparameters.rst
```diff
@@ -113,78 +113,75 @@ LightningModule hyperparameters
...
+            self.save_hyperparameters(exclude=["loss_fx", "generator_network"])
 
-    .. code-block:: python
 
-        class LitMNIST(LightningModule):
-            def __init__(self, conf: Optional[Union[Dict, Namespace, DictConfig]] = None, **kwargs):
-                super().__init__()
-                # save the config and any extra arguments
-                self.save_hyperparameters(conf)
-                self.save_hyperparameters(kwargs)
```
carmocca, line 169: I understand you don't want to document this as it can be a footgun for users.

But in that case, should also deprecate this behaviour? What's the point of having support but keep it undocumented?

awaelchli: It is still documented here in the method signature: https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bd4fbd96c1e73e251859d99b207008008de87d/pytorch_lightning/core/mixins/hparams_mixin.py#L33

I'm also ok going for a deprecation, but I should create an issue so we can discuss it there. Nonetheless, I think it's better to go with this documentation change regardless.

carmocca (approved):

docs/source/common/hyperparameters.rst
carmocca: Sure

awaelchli: Created #10375 for everyone to see :)

docs/source/common/hyperparameters.rst
```diff
@@ -113,78 +113,75 @@ LightningModule hyperparameters
...
-                super().__init__()
-                self.layer_1_dim = layer_1_dim
-                self.loss_fx = loss_fx
+Excluding hyperparameters
+"""""""""""""""""""""""""
 
-                # call this to save (layer_1_dim=128) to the checkpoint
-                self.save_hyperparameters("layer_1_dim")
+By default, every parameter of the ``__init__`` method will be considered a hyperparameter to the LightningModule.
+However, sometimes some parameters need to be excluded from saving, for example when they are not serializable.
```
tchaton, line 148:
```suggestion
However, sometimes some parameters need to be excluded from saving, for example when they are not serializable.
Those parameters should be provided back when reloading the LightningModule.
```

tchaton (approved):
LGTM! !

docs/source/common/hyperparameters.rst
```diff
@@ -113,78 +113,75 @@ LightningModule hyperparameters
 Often times we train many versions of a model. You might share that model or come back to it a few months later
 at which point it is very useful to know how that model was trained (i.e.: what learning rate, neural network, etc...).
 
-Lightning has a few ways of saving that information for you in checkpoints and yaml files. The goal here is to
+Lightning has a standardized way of saving that information for you in checkpoints and YAML files. The goal here is to
```
rohitgr7, line 116:
```suggestion
Lightning has a standardized way of saving that information for you in checkpoints and YAML files. The goal here is to
```

docs/source/common/hyperparameters.rst
```diff
@@ -113,78 +113,75 @@ LightningModule hyperparameters
...
+            self.layer_1_dim = layer_1_dim
+            self.loss_fx = loss_fx
 
+            # call this to save only (layer_1_dim=128) to the checkpoint
+            self.save_hyperparameters("layer_1_dim")
 
-3.  You can also convert full objects such as ``dict`` or ``Namespace`` to ``hparams`` so they get saved to the
-    checkpoint.
+            # equivalent
+            self.save_hyperparameters(exclude=["loss_fx", "generator_network"])
```
rohitgr7, line 163:
```suggestion
            self.save_hyperparameters(ignore=["loss_fx", "generator_network"])
```

rohitgr7 (approved):

PR #10301, opened 2021-11-02
Add `leave` argument to RichProgressBar

kaushikb11:
## What does this PR do?

Add `display_every_n_epochs` argument to `RichProgressBar`. Set to a non-negative integer to display progress bar every n epochs.

Screen Shot 2021-11-02 at 1 27 05 PM

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,7 +84,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     * Added Support for iterable datasets ([#9734](https://github.com/PyTorchLightning/pytorch-lightning/pull/9734))
     * Added `RichModelSummary` callback ([#9546](https://github.com/PyTorchLightning/pytorch-lightning/pull/9546))
     * Added `configure_columns` method to `RichProgressBar` ([#10288](https://github.com/PyTorchLightning/pytorch-lightning/pull/10288))
-    * Added `display_every_n_epochs` argument to `RichProgressBar` ([#10301](https://github.com/PyTorchLightning/pytorch-lightning/pull/10301))
+    * Added `leave` argument to `RichProgressBar` ([#10301](https://github.com/PyTorchLightning/pytorch-lightning/pull/10301))
 - Added input validation logic for precision ([#9080](https://github.com/PyTorchLightning/pytorch-lightning/pull/9080))
 - Added support for CPU AMP autocast ([#9084](https://github.com/PyTorchLightning/pytorch-lightning/pull/9084))
 - Added `on_exception` callback hook ([#9183](https://github.com/PyTorchLightning/pytorch-lightning/pull/9183))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     * Added Rich progress bar ([#8929](https://github.com/PyTorchLightning/pytorch-lightning/pull/8929), [#9559](https://github.com/PyTorchLightning/pytorch-lightning/pull/9559))
     * Added Support for iterable datasets ([#9734](https://github.com/PyTorchLightning/pytorch-lightning/pull/9734))
     * Added `RichModelSummary` callback ([#9546](https://github.com/PyTorchLightning/pytorch-lightning/pull/9546))
+    * Added `configure_columns` method to `RichProgressBar` ([#10288](https://github.com/PyTorchLightning/pytorch-lightning/pull/10288))
+    * Added `display_every_n_epochs` argument to `RichProgressBar` ([#10301](https://github.com/PyTorchLightning/pytorch-lightning/pull/10301))
 - Added input validation logic for precision ([#9080](https://github.com/PyTorchLightning/pytorch-lightning/pull/9080))
 - Added support for CPU AMP autocast ([#9084](https://github.com/PyTorchLightning/pytorch-lightning/pull/9084))
 - Added `on_exception` callback hook ([#9183](https://github.com/PyTorchLightning/pytorch-lightning/pull/9183))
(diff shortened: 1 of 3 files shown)


tchaton:
Hey @kaushikb11.

By the same occasion, do you think you could look into the fickering on the progress bar. The last line isn't aligned properly due the loss.

Best,
T.C

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -195,6 +196,7 @@ class RichProgressBar(ProgressBarBase):
 
     Args:
         refresh_rate_per_second: the number of updates per second. If refresh_rate is 0, progress bar is disabled.
+        display_every_n_epochs: Set to a non-negative integer to display progress bar every n epochs.
```
rohitgr7, line 199: should it be restricted to just -1?

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -323,9 +331,19 @@ def on_train_epoch_start(self, trainer, pl_module):
         total_batches = total_train_batches + total_val_batches
 
         train_description = self._get_train_description(trainer.current_epoch)
+        if (
+            self.main_progress_bar_id is not None
+            and self._display_every_n_epochs
+            and (not trainer.current_epoch % self._display_every_n_epochs)
```
rohitgr7, line 337: so it's like if I set it to `2` it won't show at the 3rd, 5th, 7th epoch?

rohitgr7, line 337: also I think current_epoch is zero-indexed, so should it be `(trainer.current_epoch + 1) % self._display_every_n_epochs` ?

tests/callbacks/test_rich_progress_bar.py
```diff
@@ -159,3 +160,29 @@ def configure_columns(self, trainer, pl_module):
...
+        progress_bar = RichProgressBar(display_every_n_epochs=2)
+        trainer = Trainer(
+            default_root_dir=tmpdir,
+            num_sanity_val_steps=0,
+            limit_train_batches=1,
+            max_epochs=6,
+            callbacks=progress_bar,
+        )
+        trainer.fit(model)
+    assert mock_progress_stop.call_count == 3
```
rohitgr7, line 182: should we instead check at what epoch the progress bar comes up?

pytorch_lightning/callbacks/progress/rich_progress.py
kaushikb11: What do you mean by `restricted to just -1`?

pytorch_lightning/callbacks/progress/rich_progress.py
kaushikb11: It will show, but it won't start as a new bar

kaushikb11: No, I don't believe so. It should go with every 2 epochs starting with zero index

pytorch_lightning/callbacks/progress/rich_progress.py
rohitgr7: I meant.. currently setting it to -2, -100 will show it on every epochs. so should we just make it -1? else raise a Misconfig.. the reason being, we have certain other flags across codebase that takes just `-1` as an indication for no restriction.

tests/callbacks/test_rich_progress_bar.py
kaushikb11: Updated the test! `reset` means continuing on the same bar.

pytorch_lightning/callbacks/progress/rich_progress.py
kaushikb11: No, it will currently [raise](https://github.com/PyTorchLightning/pytorch-lightning/blob/support%2Fdisplay_every_n_epoch/pytorch_lightning/callbacks/progress/rich_progress.py#L219) a Misconfiguration exception.

rohitgr7: okay.. sorry my bad.. it's non-negative.. 🤦‍♂️

rohitgr7 (approved):

kaushikb11:
> do you think you could look into the fickering on the progress bar.

@tchaton The flickering issue seems to exist for VS code terminal. It works as expected on ITerm2. I think we would need to raise an issue on Rich repo. cc @SeanNaren

tchaton (approved):
LGMT !

pytorch_lightning/callbacks/progress/rich_progress.py
awaelchli: "display" is misleading as a name imo. I wasn't sure what you mean by it, and I tested it out. It actually means you want to keep the last line of the progress bar in the stdout print at the end of an epoch, if I interpret correctly. 

Have we considered/discussed the name to be something else? Why do you think it is important to set the frequency to a different value than 0 or 1? 

`RichProgressBar(leave=True|False)`
`RichProgressBar(leave_every_epoch=True|False)`

CHANGELOG.md
```diff
@@ -128,6 +128,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for empty `gpus` list to run on CPU ([#10246](https://github.com/PyTorchLightning/pytorch-lightning/pull/10246))
 - Added `configure_columns` method to `RichProgressBar` ([#10288](https://github.com/PyTorchLightning/pytorch-lightning/pull/10288))
 - Added a warning if multiple batch sizes are found from ambiguous batch ([#10247](https://github.com/PyTorchLightning/pytorch-lightning/pull/10247))
+- Added `display_every_n_epochs` argument to `RichProgressBar` ([#10301](https://github.com/PyTorchLightning/pytorch-lightning/pull/10301))
```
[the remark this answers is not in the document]
kaushikb11: Updated!

pytorch_lightning/callbacks/progress/rich_progress.py
SeanNaren: how about `bar_every_epoch`? and set it to true false rather than providing an actual value?

awaelchli: But if I understand (hopefully), the question is not about whether the user wants to see a bar or not (it will be there the entire time). The question is whether once it is full, should it delete the full bar and reset completely or should it make a new line to start a new bar (and leave the old one behind).

kaushikb11: I am okay with boolean values.

`display_bar_every_epoch` or `bar_every_epoch`?

Tbh, `display_every_epoch` sounds good to me. Because you are technically displaying each epoch progress bar to user on the terminal.

rohitgr7: `progress_every_epoch`?

kaushikb11: Or should we go with `reset_every_epoch` with True as default.

carmocca: I don't have the full context. But one clarification:

The progress bar still appears and gets updated during the epoch and this feature is about whether it stays printed on the terminal afterward?

If that's correct, I'm in favor of naming it `leave` which matches exactly the `tqdm`
 API

Borda: so if I read it correctly, keeps showing last n bars from particular epochs, so `preserve_n_last_epochs_bars`?

justusschock: > I think we could be more intuitive about it and we don't need to match the tqdm api

I agree, but I think it is easier for people if we provide the same interface for the same functionality from different backends.

kaushikb11: @tchaton @SeanNaren I need two more votes, before going ahead with it. Let's make it quick!

awaelchli (approved):

PR #10330, opened 2021-11-03
Update bfloat16 docs

### LINKED ISSUE #10090: Improve/Update bfloat16 docs
## 📚 Documentation

We can improve the bfloat16 docs now that 1.10 is out, with also a link on how it works (how it chooses which operations to run in bf16 + the behaviour in autocast if we set to bf16).

This is motivated by https://github.com/PyTorchLightning/pytorch-lightning/pull/9049#issuecomment-949932101

SeanNaren:
## What does this PR do?

Closes #10090 

Tries to update the `bfloat16` docs a bit with information following the motivations from the related issue.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/docs/source/advanced/mixed_precision.rst
+++ b/docs/source/advanced/mixed_precision.rst
@@ -23,7 +23,7 @@ FP16 Mixed Precision
 
 In most cases, mixed precision uses FP16. Supported torch operations are automatically run in FP16, saving memory and improving throughput on GPU and TPU accelerators.
 
-Since computation happens in FP16, there is a chance of numerical instability. This is handled internally by a dynamic grad scaler which skips steps that are invalid, and adjusts the scaler to ensure subsequent steps fall within a finite range. For more information `see the autocast docs <https://pytorch.org/docs/stable/amp.html#gradient-scaling>`__.
+Since computation happens in FP16, there is a chance of numerical instability during training. This is handled internally by a dynamic grad scaler which skips steps that are invalid, and adjusts the scaler to ensure subsequent steps fall within a finite range. For more information `see the autocast docs <https://pytorch.org/docs/stable/amp.html#gradient-scaling>`__.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


docs/source/advanced/mixed_precision.rst
```diff
@@ -39,14 +39,16 @@ BFloat16 Mixed Precision
...
-    BFloat16 requires PyTorch 1.10 or later. Currently this requires installing `PyTorch Nightly <https://pytorch.org/get-started/locally/>`__.
+    BFloat16 requires PyTorch 1.10 or later.
 
     BFloat16 is also experimental and may not provide large speedups or memory improvements, but offer better numerical stability.
 
     Do note for GPUs, largest benefits require `Ampere <https://en.wikipedia.org/wiki/Ampere_(microarchitecture)>`__ based GPUs, such as A100s or 3090s.
 
 BFloat16 Mixed precision is similar to FP16 mixed precision, however we maintain more of the "dynamic range" that FP32 has to offer. This means we are able to improve numerical stability, compared to FP16 mixed precision. For more information see `this TPU performance blog post <https://cloud.google.com/blog/products/ai-machine-learning/bfloat16-the-secret-to-high-performance-on-cloud-tpus>`__.
 
+A technical note: unlike TPUs which accumulate in FP32, PyTorch does not accumulate into a master set of FP32 weights. Instead, PyTorch relies on `autocast's <https://pytorch.org/docs/stable/amp.html#autocasting>`__ list of supported bfloat16 ops to prevent instabilities and remain in bfloat16 where possible.
```
SeanNaren, line 50: cc @zplizzi is this correct? going to try ping a few people in the FB team to help confirm here :)

cc @blefaudeux @rohan-varma

blefaudeux: I'm not very familiar with how Pytorch handles bfloat16, but I can tell about float16 and AMP. My guess is that the two are not the same, float16 is handled through *mixed* precision (some ops are fp16 when low dynamic range is enough + general loss scaling, some ops are kep fp32, emphasis on the mixed because you will have a mix of both during the training).

bfloat16 does not have the same issue, [the dynamic range is the same as fp32](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) so you don't need any scaling and underflow/overflow catch (well, not more than fp32), it's just not as precise as fp32 (you get quantization noise basically). In that case I would expect the training to be monolithic, not mixed. 

I had a quick look at the Pytorch doc and bfloat16 does not seem mentioned other than in terms of an available type, so my suspicion is that it does not work with autocast, it's just that the whole model and data which are bfloat16 and you run with it.

zplizzi: So how @SeanNaren implemented bfloat16 in lightning is to use torch native AMP just like fp16, except using bfloat16 instead of fp16 as the fast type (and with grad scaling disabled). As such, the model is left in fp32.

@SeanNaren , I don't think your addition here is quite right given the above. The weights are still kept in fp32, so there's no "accumulation" necessary. I would probably say something like "bfloat16 support in lightning is implemented by using torch.cuda.amp with fp16 replaced by bfloat16, and without using gradient scaling". I would probably also clean up the text above your addition - or maybe just remove it. It's not clear to me whether bfloat16 is actually more or less numerically stable than fp16 when used this way (fp16 has more precision, and gradient scaling was used to compensate for the lack of dynamic range). And I think (?) that bfloat16 is *only* available on Ampere.

SeanNaren: Thanks guys! I've updated based on @zplizzi suggestions :)

ethanwharris (approved):
LGTM :smiley:

Borda (approved):

docs/source/advanced/mixed_precision.rst
```diff
@@ -39,15 +39,15 @@ BFloat16 Mixed Precision
...
+    BFloat16 requires PyTorch 1.10 or later.
 
     BFloat16 is also experimental and may not provide large speedups or memory improvements, but offer better numerical stability.
 
     Do note for GPUs, largest benefits require `Ampere <https://en.wikipedia.org/wiki/Ampere_(microarchitecture)>`__ based GPUs, such as A100s or 3090s.
 
 BFloat16 Mixed precision is similar to FP16 mixed precision, however we maintain more of the "dynamic range" that FP32 has to offer. This means we are able to improve numerical stability, compared to FP16 mixed precision. For more information see `this TPU performance blog post <https://cloud.google.com/blog/products/ai-machine-learning/bfloat16-the-secret-to-high-performance-on-cloud-tpus>`__.
 
-Since BFloat16 is more stable than FP16 during training, we do not need to worry about any gradient scaling or nan gradient values that comes with using FP16 mixed precision.
+Under the hood we use `torch.cuda.amp <https://pytorch.org/docs/stable/amp.html>`__ with the the dtype set to `bfloat16`, with no gradient scaling.
```
carmocca, line 50: We use `torch.autocast`

awaelchli (approved):

justusschock (approved):

PR #10075, opened 2021-10-22
Upgrade CI after the 1.10 release

carmocca:
## What does this PR do?

- Test torch 1.10
- Rename the jobs for consistency and clarity
- Update the REDAME table

### Does your PR introduce any breaking changes? If yes, please list them.

None

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

PR #10367, opened 2021-11-05
Drop torch 1.6 support

carmocca:
## What does this PR do?

We have a 4 PyTorch version policy. With the release of 1.10, that means we can drop 1.6

[1.6, 1.7, 1.8, 1.9, 1.10] -> [1.7, 1.8, 1.9, 1.10]

Part of #10075

None

## Before submitting

## PR review


environment.yml
```diff
@@ -29,7 +29,7 @@ dependencies:
     - python>=3.6
     - pip>20.1
     - numpy>=1.17.2
-    - pytorch>=1.6
+    - pytorch>=1.7.1
```
justusschock, line 32: I think that should still work since 1.7 and 1.7.1 should be identical from the API perspective

```suggestion
    - pytorch>=1.7
```

.github/workflows/events-nightly.yml
```diff
@@ -123,7 +125,7 @@ jobs:
       matrix:
         # the config used in '.github/workflows/ci_test-conda.yml'
         python_version: ["3.8"]
-        pytorch_version: ["1.6", "1.7", "1.8", "1.9", "1.10"]
+        pytorch_version: ["1.7.1", "1.8", "1.9", "1.10"]
```
justusschock, line 128:
```suggestion
        pytorch_version: ["1.7", "1.8", "1.9", "1.10"]
```

That should work since we test with 3.8 here, no?

.github/workflows/ci_test-conda.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.8"]  # previous to last Python version as that one is already used in test-full
-        pytorch-version: ["1.6", "1.7", "1.8", "1.9", "1.10"]
+        pytorch-version: ["1.7.1", "1.8", "1.9", "1.10"]
```
justusschock, line 18:
```suggestion
        pytorch-version: ["1.7", "1.8", "1.9", "1.10"]
```
That should work since we test with 3.8 here, no?

.github/workflows/ci_dockers.yml
```diff
@@ -93,12 +93,12 @@ jobs:
       matrix:
         # the config used in '.github/workflows/ci_test-conda.yml'
         python_version: ["3.8"]
-        pytorch_version: ["1.6", "1.7", "1.8", "1.9", "1.10"]
+        pytorch_version: ["1.7.1", "1.8", "1.9", "1.10"]
```
justusschock, line 96:
```suggestion
        pytorch_version: ["1.7", "1.8", "1.9", "1.10"]
```
That should work since we run 3.8 here, no?

justusschock (approved):
LGTM overall.

I think whereever possible (all runs/images with python 3.8) we should still fall back to 1.7 and only for the 3.9 versions lift it to 1.7.1

pytorch_lightning/overrides/torch_distributed.py
```diff
@@ -1,99 +0,0 @@
-import logging
```
awaelchli, line 1: shouldn't this have been done in another PR?

awaelchli (approved):

requirements/adjust_versions.py
```diff
@@ -14,7 +14,6 @@
     dict(torch="1.8.0", torchvision="0.9.0", torchtext="0.9.0"),
     dict(torch="1.7.1", torchvision="0.8.2", torchtext="0.8.1"),
     dict(torch="1.7.0", torchvision="0.8.1", torchtext="0.8.0"),
```
akihironitta, line 16: If we drop support for `torch<1.7.1`, should this line also be removed?
```suggestion
```

.github/workflows/events-nightly.yml
```diff
@@ -6,6 +6,8 @@ on:
   schedule:
     # At the end of every day
     - cron: "0 0 * * *"
+  # FIXME
+  push: {}
```
akihironitta, line 10: Should be removed before merge?
```suggestion
```

pytorch_lightning/overrides/torch_distributed.py
carmocca: I've done it all together in a single unit of work because the PR is not too long and it helps validate the CI runs properly.

But if you insist, I could split it

requirements/adjust_versions.py
carmocca: Chose to keep it for now in case we use 1.7.0 somewhere. Using 1.7.1 is a limitation due to Python 3.9 support, not in general

carmocca:
@justusschock we could do that but would complicate CI a bit since we would need logic to handle 1.7 and 1.7.1 separately, additionally, we would need to generate both docker images.

I think 1.7.1 is fine looking at the release notes: https://github.com/pytorch/pytorch/releases/tag/v1.7.1
and download stats per version: https://pepy.tech/project/torch?versions=1.7.0&versions=1.7.1&versions=1.10.0&versions=1.6.0

And technically this should have been done before the 1.5 release, because now that supports 1.6-1.10 (5 versions)

so we should actually drop 1.7 too before the 1.6 release

akihironitta (approved):
LGTM!

Does this mean we can also deprecate the following lines and remove it in pl v1.7, right?

https://github.com/PyTorchLightning/pytorch-lightning/blob/3a6acadd115e86f02d83a788f1978372ab6764f3/pytorch_lightning/overrides/torch_distributed.py#L13-L14

carmocca:
> Does this mean we can also deprecate the following lines and remove it in pl v1.7, right?

yes, in #10390

Borda:
@carmocca I would rather keep it on master as we are going to merge bug fixes which will be conservatively cherry picked to `release/1.5.x` where we still support PT 1.6

carmocca:
@Borda This PR should have landed with 1.5 already to comply with the 4 torch versions window.

Still, this can be merged to master and the milestone changed if you want to land a specific fix for 1.5.0 with 1.6

requirements/adjust_versions.py
```diff
@@ -33,28 +32,59 @@ def find_latest(ver: str) -> Dict[str, str]:
...
+        replace = f"{lib}=={version}" if version else ""
+        req = re.sub(rf"\b{lib}(?!\w).*", replace, req)
 
-    print(req)  # on purpose - to debug
-    with open(path_req, "w") as fp:
-        fp.write(req)
+    return req
+
+
+def test():
```
tchaton, line 53: Not sure if this test should be there.

tchaton (approved):
LGTM !

requirements/adjust_versions.py
carmocca: I couldn't think of any other way to test this, as this is outside of our CI.

It should be very quick to run

PR #10369, opened 2021-11-05
Fixed uploading best model checkpoint in NeptuneLogger

### LINKED ISSUE #10602: Neptune Logger assumes unix-style filepath for model checkpoint
The Neptune Logger can automatically upload model checkpoints if the callback is used. It generates an "expected_model_path" for that. This path has hard-coded a '/' at the end (see [here](https://github.com/PyTorchLightning/pytorch-lightning/blob/261ea90822e2bf1cfa5d56171ab1f95a81d5c571/pytorch_lightning/loggers/neptune.py#L547)), which can cause problems on Windows systems because the actual path has (or can have) a backslash at this position. A simple fix would be to use `os.path.sep` insteat of '/'.

cc @justusschock @awaelchli @akihironitta

Raalsky:
## What does this PR do?

Fixes #10602

Adds best model checkpoint uploading to NeptuneLogger.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loggers/neptune.py
```diff
@@ -523,6 +523,17 @@ def after_save_checkpoint(self, checkpoint_callback: "ReferenceType[ModelCheckpo
...
+        # log best model path
+        if checkpoint_callback.best_model_path:
+            self.experiment[
+                self._construct_path_with_prefix("model/best_model_path")
+            ] = checkpoint_callback.best_model_path
+
+            model_name = self._get_full_model_name(checkpoint_callback.best_model_path, checkpoint_callback)
+            file_names.add(model_name)
+            self.experiment[f"{checkpoints_namespace}/{model_name}"].upload(
+                checkpoint_callback.best_model_path)
```
justusschock, line 535: Can we enable this optionally with a flag? I think this flag should also default to `False` to replicate current behaviour.

Background: There are people having only a limited bandwith during training

Raalsky: Hey @justusschock,
We already have `log_model_checkpoints` which is set to `True` at default. In the case of the user calling ModelCheckpoint directly, it should work pretty fine with uploading model checkpoints (before this bugfix). We're trying to document in all known places that `log_model_checkpoints=True` is equal to uploading every checkpoint that was saved by a Trainer. In such a context, I think that we're filling the gap instead of breaking some expected behavior. So I don't think it's necessary to add any new flag :wink:

justusschock (approved):

awaelchli:
Seems our codeowners github trigger did not work. They were not requested for review.
https://github.com/PyTorchLightning/pytorch-lightning/blob/edbf27430d02b59da717a870d98154f3dbea8ff1/.github/CODEOWNERS#L29

Pinging them manually for now:
@shnela @HubertJaworski @pkasprzyk @pitercl @Raalsky @aniezurawski @kamil-kaczmarek
Could one of you have a quick look? Thx

kamil-kaczmarek:
@awaelchli 
approved.

pytorch_lightning/loggers/neptune.py
```diff
@@ -523,6 +523,16 @@ def after_save_checkpoint(self, checkpoint_callback: "ReferenceType[ModelCheckpo
             file_names.add(model_name)
             self.experiment[f"{checkpoints_namespace}/{model_name}"].upload(key)
 
+        # log best model path and checkpoint
+        if checkpoint_callback.best_model_path:
+            self.experiment[
+                self._construct_path_with_prefix("model/best_model_path")
+            ] = checkpoint_callback.best_model_path
```
rohitgr7, line 530: quick ques: why aren't we following the similar upload process as that of last model path?
for eg:
we are not doing this while uploading last model:
```py
self._construct_path_with_prefix("model/best_model_path")
```

pytorch_lightning/loggers/neptune.py
```diff
@@ -523,6 +523,16 @@ def after_save_checkpoint(self, checkpoint_callback: "ReferenceType[ModelCheckpo
             file_names.add(model_name)
             self.experiment[f"{checkpoints_namespace}/{model_name}"].upload(key)
 
+        # log best model path and checkpoint
+        if checkpoint_callback.best_model_path:
```
rohitgr7, line 527: not an expert with neptune, but just wondering since we already upload all the best k models, shouldn't the best one already there and somehow if possible we can instead upload the best model checkpoint path/name?

rohitgr7 (commented):
mind update the chlog since 1.5.1 has been released?

pytorch_lightning/loggers/neptune.py
Raalsky: Internally Neptune has an asynchronous worker to consume operations (file upload, value assign, etc.) Essentially only the last file will be uploaded to the same namespace (it's like overriding which file should be uploaded).

awaelchli (approved):

pytorch_lightning/loggers/neptune.py
awaelchli: @Raalsky could you have a look?

Raalsky: @kamil-kaczmarek Would you mind? :wink:

CHANGELOG.md
```diff
@@ -136,6 +136,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed an issue where the model wrapper in Lite converted non-floating point tensors to float ([#10429](https://github.com/PyTorchLightning/pytorch-lightning/pull/10429))
 - Fixed an issue with inferring the dataset type in fault-tolerant training ([#10432](https://github.com/PyTorchLightning/pytorch-lightning/pull/10432))
 - Fixed dataloader workers with `persistent_workers` being deleted on every iteration ([#10434](https://github.com/PyTorchLightning/pytorch-lightning/pull/10434))
+- Fixed uploading best model checkpoint in NeptuneLogger ([#10369](https://github.com/PyTorchLightning/pytorch-lightning/pull/10369))
```
Borda, line 139: pls move this to unreleased section

tests/loggers/test_neptune.py
```diff
@@ -276,14 +276,15 @@ def test_after_save_checkpoint(self, neptune):
...
             cb_mock = MagicMock(
-                dirpath="path/to/models",
-                last_model_path="path/to/models/last",
+                dirpath=models_root_dir,
+                last_model_path=os.path.join(models_root_dir, "last"),
                 best_k_models={
-                    "path/to/models/model1": None,
-                    "path/to/models/model2/with/slashes": None,
+                    f"{os.path.join(models_root_dir, 'model1')}": None,
+                    f"{os.path.join(models_root_dir, 'model2/with/slashes')}": None,
```
Borda, line 285: maybe still replace / by path join here

Raalsky: By `model2/with/slashes` we mean checkpoints produced via (we already saw a few examples of it):
```
ModelCheckpoint(dirpath="model/checkpoints/",
                               filename="{epoch:02d}-{val/loss:.2f}",
                               save_weights_only=True,
                               save_top_k=3,
                               save_last=True,
                               monitor="val/loss",
                               every_n_epochs=1)
```
which produces checkpoints like `epoch=00-val/loss=0.16.ckpt`. So it's with a "/" slash symbol in the filename and I don't know how exactly Windows handle it :wink: and that's the reason that it wasn't joined via `os.path.join`.

Check these as an reference: https://app.neptune.ai/o/common/org/pytorch-lightning-integration/e/PTL-1/all?path=training%2Fmodel%2Fcheckpoints%2Fepoch%3D00-val

CHANGELOG.md
awaelchli: What @Borda meant is to move it to the 1.6.0 section which is the unreleased next version.

Raalsky: :facepalm: I was thinking that 1.6.0 is some "future and breaking changes" and I was looking for something like "1.5.3" etc. Ok, should be fixed in a moment.

tchaton (approved):
LGTM !

pytorch_lightning/loggers/neptune.py
Raalsky: Sorry that it takes so long to answer :relieved:.

The main idea was to provide an exact local path of the corresponding model checkpoint files that were uploaded. At our design level, there was some lack of information about how we should fully handle every behavior that ModelCheckpoint could provide. After that, at the implementation, we missed and messed something that it's got only applied to the best model.

I think it should be translated to something like `local_path` instead of `best_model_path` and be prepared for every last/k-best/best (especially be exactly right to every `cpkt` file). We would like to do some additional discussions to take proper care of it and I think it's out-of-scope for this PR. It may introduce new functionality or change the behavior instead of only bug-fixing.

awaelchli:
The hanging tests were fixed on master. I tried to update the branch but don't have write access. Could you please do it so we can merge the fix?

Raalsky:
@awaelchli Is there anything should I take care of?

rohitgr7 (approved):

awaelchli:
Nope, thanks for the reminder. Will check back once tests have run

awaelchli:
Thanks @kamil-kaczmarek , this fix will be included next week in the bugfix release 1.5.5.

kamil-kaczmarek:
Great thanks, kudos to @Raalsky for leading this bug fix :tada:

PR #10374, opened 2021-11-05
Enable distributed training with CombinedDataLoader and max_size_cycle

### LINKED ISSUE #10373: CombinedLoader does not work in DDP when using `max_size_cyle` option
## 🐛 Bug

When using a `CombinedLoader` with the `max_size_cycle` option and DDP, all the GPUs get all validation data. 
This bug is related to https://github.com/PyTorchLightning/pytorch-lightning/issues/7013 - however, the fix in PR #7102 only affect the default `min_size` option of the `CombinedLoader`

@tchaton ?

### To Reproduce

[Repro](https://gist.github.com/ant0nsc/e4ce39c76c3176c4d1bea133c2875391)

### Expected behavior

For the above repro, the validation data has length 8. I would expect that each of the 2 GPUs only get 4 batches each, but in fact they get 8 batches.

### Environment

* CUDA:
        - GPU:
                - Tesla K80
                - Tesla K80
        - available:         True
        - version:           10.2
* Packages:
        - numpy:             1.21.2
        - pyTorch_debug:     False
        - pyTorch_version:   1.8.0
        - pytorch-lightning: 1.5.0
        - tqdm:              4.62.3
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - 
        - processor:         x86_64
        - python:            3.7.3
        - version:           18.04.1-Ubuntu SMP Wed Jul 28 23:14:18 UTC 2021

tchaton:
## What does this PR do?

This PR adds support for CombinedDataLoader with max_size_cycle and distributed training.
The current `prepare_dataloader` wasn't injected a DistributedSampler within the CycleIterator loader.

Fixes #10373

No.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/supporters.py
```diff
@@ -457,6 +464,19 @@ def _wrap_loaders_max_size_cycle(self) -> Any:
             )
             state.reset()
 
+    def _apply_cycle_iterator_length(self):
```
tchaton, line 467: Note for reviewing: I am not sure happy with this function, but unfortunately, the length computation requires to be computed only once all the dataloaders have been modified.
Alternatives: use `self.length / world_size` within the CycleIterator, but this could lead to incorrect results if the users had custom dataloaders, samplers.

tests/trainer/test_supporters.py
```diff
@@ -379,3 +380,34 @@ def _assert_dataset(loader):
...
+
+    dataloader = CombinedLoader(
+        {"a": DataLoader(RandomDataset(32, 8), batch_size=1), "b": DataLoader(RandomDataset(32, 8), batch_size=1)},
+    )
+
+    trainer = Trainer(strategy="ddp", gpus=2, replace_sampler_ddp=replace_sampler_ddp)
+    dataloader = trainer.prepare_dataloader(dataloader, shuffle=False)
+    assert len(dataloader) == 4 if replace_sampler_ddp else 8
+
+    for a_length in [6, 8, 10]:
```
awaelchli, line 392: can we add a comment for this section of the code? it seems it is independent of the blocks above. perhaps these sections should be their own tests?

pytorch_lightning/trainer/supporters.py
awaelchli: this info could be added as a comment or docstring to this method.

pytorch_lightning/trainer/supporters.py
```diff
@@ -457,6 +464,19 @@ def _wrap_loaders_max_size_cycle(self) -> Any:
...
+        if self.mode == "max_size_cycle":
```
awaelchli, line 468:
```suggestion
        if self.mode != "max_size_cycle":
            return
```
invert the if statement?

pytorch_lightning/trainer/supporters.py
awaelchli, line 467:
```suggestion
    def _apply_cycle_iterator_length(self) -> None:
```

pytorch_lightning/trainer/supporters.py
```diff
@@ -217,6 +217,13 @@ def __next__(self) -> Any:
     def __len__(self) -> Union[int, float]:
         return self.length
 
+    @staticmethod
+    def get_len(cycle_iterator: "CycleIterator") -> Union[int, float]:
+        if has_len(cycle_iterator.loader):
+            return len(cycle_iterator.loader)
+
+        return float("inf")
```
awaelchli, line 225:
```suggestion
        return len(cycle_iterator.loader
```
why not just this?

And it is probably problematic to have both `__len__` and the new `get_len` defined. Can we combine them?

pytorch_lightning/trainer/supporters.py
tchaton: Done !

tests/trainer/test_supporters.py
tchaton: Done !

pytorch_lightning/trainer/supporters.py
tchaton: Done !

pytorch_lightning/trainer/supporters.py
tchaton: Done !

pytorch_lightning/trainer/supporters.py
tchaton: Done !

pytorch_lightning/trainer/supporters.py
```diff
@@ -457,6 +457,26 @@ def _wrap_loaders_max_size_cycle(self) -> Any:
...
+        """When the model is `max_size_cycle`, compute the length across all ``CycleIterator`` and re-assign it to
+        all dataloaders."""
+        if self.mode != "max_size_cycle":
+            return
+
+        def get_cycle_iterator_len(cycle_iterator: CycleIterator) -> Union[float, int]:
+            return get_len(cycle_iterator.loader)
+
+        all_lengths = apply_to_collection(
+            self.loaders, CycleIterator, get_cycle_iterator_len, wrong_dtype=(Sequence, Mapping)
```
awaelchli, line 470:
```suggestion
            self.loaders, CycleIterator, function=(lambda c: get_len(c.loader)), wrong_dtype=(Sequence, Mapping)
```

pytorch_lightning/trainer/data_loading.py
```diff
@@ -136,14 +136,22 @@ def prepare_dataloader(self, dataloader: Any, shuffle: bool, mode: Optional[Runn
         if isinstance(dataloader, CombinedLoader):
             # apply `prepare_dataloader` on all the collection of loaders
             dataloader.loaders = apply_to_collection(
-                dataloader.loaders, DataLoader, self.prepare_dataloader, shuffle, mode=mode
+                dataloader.loaders, (DataLoader, CycleIterator), self.prepare_dataloader, shuffle, mode=mode
             )
+            # the length need to recomputed across all dataloaders in case of special behavior.
+            dataloader._apply_cycle_iterator_length()
```
awaelchli, line 142: it seems that we cannot do this length computation at the time of instantiation, of `CombinedLoader` (i.e. in `CombinedLoader._wrap_loaders_max_size_cycle`. 
The result is that the dataloader preparation code is interleaving sections corresponding to different (optional) features and this is probably a sign that we will revisit this code (and problems) the future 😅

pytorch_lightning/trainer/data_loading.py
```diff
@@ -153,6 +161,10 @@ def prepare_dataloader(self, dataloader: Any, shuffle: bool, mode: Optional[Runn
             sampler = self._resolve_sampler(dataloader, shuffle=shuffle, mode=mode)
             dataloader = self._update_dataloader(dataloader, sampler, mode=mode)
 
+        if cycle_iterator is not None:
```
awaelchli, line 164: shouldn't the dataloader variable be set to the cycle_iterator reference and this if-statement be an instance check instead?

awaelchli (approved):

pytorch_lightning/trainer/supporters.py
tchaton: Done !

pytorch_lightning/trainer/data_loading.py
tchaton: Yes, improved it slightly, but I agree, might be worth to clean this up.

pytorch_lightning/trainer/data_loading.py
tchaton: Hey @awaelchli, not sure to fully get your suggestion.

The cycle_iterator isn't defined in some case. How would you like to implement this ?

pytorch_lightning/trainer/supporters.py
```diff
@@ -473,11 +490,12 @@ def __getstate__patch__(*_):
         return iterator
 
     @staticmethod
-    def _calc_num_batches(loaders: Any) -> Union[int, float]:
+    def _calc_num_batches(loaders: Any, mode="min_size") -> Union[int, float]:
```
carmocca, line 493: Shouldn't this be "min_size_cycle"?

pytorch_lightning/trainer/supporters.py
```diff
@@ -457,6 +457,23 @@ def _wrap_loaders_max_size_cycle(self) -> Any:
...
             state.reset()
 
+    def _apply_cycle_iterator_length(self) -> None:
+        """When the model is `max_size_cycle`, compute the length across all ``CycleIterator`` and re-assign it to
+        all dataloaders."""
+        if self.mode != "max_size_cycle":
+            return
+
+        all_lengths = apply_to_collection(
+            self.loaders, CycleIterator, lambda c: get_len(c.loader), wrong_dtype=(Sequence, Mapping)
```
carmocca, line 467: Wouldn't this work too?

```suggestion
            self.loaders, DataLoader, get_len
```

Also, why do you set `wrong_dtype`? Sequence and Mapping do not subclass CycleIterator

pytorch_lightning/trainer/supporters.py
```diff
@@ -457,6 +457,23 @@ def _wrap_loaders_max_size_cycle(self) -> Any:
...
+        )
+        length = _nested_calc_num_data(all_lengths, max)
+
+        def _apply_fn(cycle_iterator: CycleIterator) -> None:
+            nonlocal length
```
carmocca, line 472: This could take the value to avoid the nonlocal

```suggestion
        def _apply_fn(cycle_iterator: CycleIterator, length: int) -> None:
```

carmocca (commented):
Definitely an esoteric part of the codebase

pytorch_lightning/trainer/supporters.py
tchaton: There isn't any cycle if min_size is used. I believe it is why it was chosen this way. It should be changed within another PR if we want to.

tests/trainer/test_supporters.py
```diff
@@ -379,3 +381,65 @@ def _assert_dataset(loader):
             assert isinstance(d, CustomDataset)
 
     apply_to_collection(dataloader.loaders, DataLoader, _assert_dataset)
+
+
+@pytest.mark.skipif(not _TORCH_GREATER_EQUAL_1_7, reason="Requires at least PyTorch 1.7")
```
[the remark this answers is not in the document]
tchaton: Removed !

pytorch_lightning/trainer/supporters.py
tchaton: Yes, removed !

tchaton: And it needs to apply on the CycleIterator which doesn't subclass DataLoader.

pytorch_lightning/trainer/supporters.py
tchaton: Done !

tests/trainer/test_supporters.py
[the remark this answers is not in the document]
carmocca: Looks like you didn't push the commit

justusschock (approved):

tests/trainer/test_supporters.py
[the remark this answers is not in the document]
tchaton: Done !

tests/trainer/test_supporters.py
```diff
@@ -379,3 +381,64 @@ def _assert_dataset(loader):
...
+        dataloader = trainer.prepare_dataloader(dataloader, shuffle=False)
+        assert len(dataloader) == length // 2 if replace_sampler_ddp else length
+        if replace_sampler_ddp:
+            batches = [batch for batch in dataloader]
+            if a_length == 6:
+                assert batches[-1] == {"a": torch.tensor([0]), "b": torch.tensor([6])}
+            elif a_length == 8:
+                assert batches[-1] == {"a": torch.tensor([6]), "b": torch.tensor([6])}
+            elif a_length == 10:
+                assert batches[-1] == {"a": torch.tensor([8]), "b": torch.tensor([0])}
```
carmocca, line 425:
```suggestion
            last_batch = list(dataloader)[-1]
            if a_length == 6:
                assert last_batch == {"a": torch.tensor([0]), "b": torch.tensor([6])}
            elif a_length == 8:
                assert last_batch == {"a": torch.tensor([6]), "b": torch.tensor([6])}
            elif a_length == 10:
                assert last_batch == {"a": torch.tensor([8]), "b": torch.tensor([0])}
```

carmocca (approved):

pytorch_lightning/trainer/supporters.py
```diff
@@ -457,6 +457,20 @@ def _wrap_loaders_max_size_cycle(self) -> Any:
...
+        if self.mode != "max_size_cycle":
+            return
+
+        all_lengths = apply_to_collection(self.loaders, CycleIterator, lambda c: get_len(c.loader))
+
+        def _apply_fn(cycle_iterator: CycleIterator, length) -> None:
+            cycle_iterator.length = length
+
+        length = _nested_calc_num_data(all_lengths, max)
+        apply_to_collection(self.loaders, CycleIterator, partial(_apply_fn, length=length))
```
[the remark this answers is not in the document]
justusschock: @carmocca but the `_apply_fn` is still needed, no? Just the partial isn't

Issue #10329, opened 2021-11-03
DataLoader wrapping, re-instantiation and patching 


awaelchli:
## 🐛 Bug

Recent changes to the Lite DataLoader in #10279 have not been properly reviewed and tested. Several clean ups need to be done and better tests need to be written. 

The following issue exists:

### To Reproduce

```python
def test_setup_custom_dataloaders_new():

    class DataLoaderSubclass1(DataLoader):
        def __init__(self, attribute1, *args, **kwargs):
            # UNCOMMENT THE LINE BELOW, NO DIFFERENCE (EXPECTED)
            # self.attribute1 = attribute1
            super().__init__(*args, **kwargs)

    class DataLoaderSubclass2(DataLoaderSubclass1):
        def __init__(self, attribute1, attribute2, *args, **kwargs):
            # UNCOMMENT THE LINE BELOW, TEST PASSES
            # self.attribute2 = attribute2
            super().__init__(attribute1, *args, **kwargs)

    class LiteWithCustomDataLoader(LightningLite):
        def run(self):
            dataloader = DataLoaderSubclass2("attribute1", "attribute2", dataset=range(4), batch_size=2)
            lite_dataloader = self.setup_dataloaders(dataloader)
            
            # THIS ASSERTION PASSES
            assert lite_dataloader._dataloader.attribute1 == "attribute1"
            
            # THIS ASSERTION FAILS
            assert lite_dataloader._dataloader.attribute2 == "attribute2"

    LiteWithCustomDataLoader().run()
```

### Expected behavior

According to the changes in the PR #10279, this is supposed to work and was the main motivation of it. But it does not because the tests were insufficient. This issue was found by me in #10297 while trying to clean up the code. 

### Environment

Lightning v1.5.0!

cc @tchaton @rohitgr7 @carmocca @justusschock @awaelchli

awaelchli:
The issue exists because this wrapper here

https://github.com/PyTorchLightning/pytorch-lightning/blob/93caa7cda9277f86f17b1eab0793019aee8d40c2/pytorch_lightning/lite/wrappers.py#L106-L116

pops args and kwargs. 
When we get to this line here

https://github.com/PyTorchLightning/pytorch-lightning/blob/93caa7cda9277f86f17b1eab0793019aee8d40c2/pytorch_lightning/lite/lite.py#L242-L244

We are re-instantiating (with the new init) but passing arguments by kwargs, strictly! This means the for loop here runs empty!

[...]

PR #10334, opened 2021-11-03
Fix DataLoader inspection and re-instantiation in Lite

awaelchli:
## What does this PR do?

Part of #10329

A detailed explanation of why the bug occurs can be found in the linked issue.
A newly added test demonstrates the failure on master. 

This fix should be reviewed carefully, at least by @carmocca and @tchaton .

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

Part of #1 (it's a lie, this is just here to avoid noisy GitHub bot)

PR #10376, opened 2021-11-05
Revert part of #10279

carmocca:
## What does this PR do?

We shouldn't support this, as this is a bug in the DataLoader implementation.

Adding arbitrary `**kwargs` to a custom DataLoader allows for duplicate arguments.
This is something that should be checked by the user.

The fix that this PR removes only works for the first argument and if the name is "dataset", but this would still break for any of the other duplications.

Instead, the author of the DataLoader should write something like:

```python
def __init__(self, value: int = 2, *args, **kwargs):
    if "dataset" in kwargs:
        dataset = kwargs["dataset"]
    else:
        dataset = range(value)
    super().__init__(dataset, *args, **kwargs)
```

Partly reverts #10279
Part of https://github.com/PyTorchLightning/pytorch-lightning/issues/10329

A follow-up PR will add a `MisconfigurationException` to help the user in these cases.

This will no longer work for anybody already relying on this. Lite is experimental

## Before submitting

## PR review

### DIFF
--- a/tests/lite/test_lite.py
+++ b/tests/lite/test_lite.py
@@ -460,7 +460,9 @@ def test_replace_dataloader_init_method():
 
     with _replace_dataloader_init_method():
         dataloader = CustomDataLoader(extra_argument=1, dataset=range(1))
+        assert dataloader.extra_argument == 1
         dataloader = lite.setup_dataloaders(dataloader)
 
         dataloader = CustomDataLoader(1, range(1))
+        assert dataloader.extra_argument == 1
         dataloader = lite.setup_dataloaders(dataloader)
--- a/tests/lite/test_lite.py
+++ b/tests/lite/test_lite.py
@@ -24,7 +24,12 @@ from torch import nn
 from torch.utils.data import DataLoader, DistributedSampler, Sampler
 
 from pytorch_lightning.lite import LightningLite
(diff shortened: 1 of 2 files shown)


tests/lite/test_lite.py
```diff
@@ -164,57 +164,6 @@ def test_setup_dataloaders_return_type():
     assert lite_dataloader1.dataset is dataset1
 
 
-def test_setup_custom_dataloaders():
```
tchaton, line 167: We could have a custom dataloader using only a dataset as a test.

tchaton (approved):
LGTM !

tests/lite/test_lite.py
carmocca: We already test that in `tests/trainer/test_data_loading.py`

awaelchli (approved):

tests/lite/test_lite.py
tchaton: I meant, we should unittest the context manager to make sure the params are

tchaton (approved):
LGTM ! @carmocca I added a test to ensure the new context manager is being tested.

rohitgr7 (approved):

tests/lite/test_lite.py
carmocca: @tchaton isn't that covered by the test that @awaelchli added in https://github.com/PyTorchLightning/pytorch-lightning/pull/10334?

PR #10406, opened 2021-11-08
make `get_nvidia_gpu_stats` public

rohitgr7:
## What does this PR do?

This can be a public method so that it can be used when training with Lite.
Fixes: #10414

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @justusschock @awaelchli

### DIFF
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -73,7 +73,7 @@ class GPUAccelerator(Accelerator):
         """
         if _TORCH_GREATER_EQUAL_1_8:
             return torch.cuda.memory_stats(device)
-        return _get_nvidia_gpu_stats(device)
+        return get_nvidia_gpu_stats(device)
 
     def teardown(self) -> None:
         super().teardown()
@@ -85,7 +85,7 @@ class GPUAccelerator(Accelerator):
         return torch.cuda.device_count()
 
 
-def _get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
+def get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
     """Get GPU stats including memory, fan speed, and temperature from nvidia-smi.
 
     Args:
@@ -98,6 +98,10 @@ def _get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
         FileNotFoundError:
             If nvidia-smi installation not found
     """
+    nvidia_smi_path = shutil.which("nvidia-smi")
+    if nvidia_smi_path is None:
+        raise FileNotFoundError("nvidia-smi: command not found")
+
     gpu_stat_metrics = [
         ("utilization.gpu", "%"),
         ("memory.used", "MB"),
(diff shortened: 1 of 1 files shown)


kaushikb11 (approved):
LGTM!

ethanwharris (commented):
Looks good :smiley: Could you also update the forwarding references in https://github.com/PyTorchLightning/pytorch-lightning/blob/e61a98a3d0cbe28f1df41841b4766166dd9ea531/pytorch_lightning/utilities/memory.py#L111

rohitgr7:
> Looks good 😃 Could you also update the forwarding references in
> 
> https://github.com/PyTorchLightning/pytorch-lightning/blob/e61a98a3d0cbe28f1df41841b4766166dd9ea531/pytorch_lightning/utilities/memory.py#L111

good catch. thanks 😃

ethanwharris (approved):
LGTM :smiley:

rohitgr7:
> @rohitgr7 I'm not sure about this. The accelerator api for accessing device stats is already public.
> 
> > This can be a public method so that it can be used when training with Lite.
> 
> Is there an issue connected to this? This to me does not seem sufficient for a motivation. In theory, everything in Lightning can become a utility. But should it?

I was just trying out Lite over the weekend. Comparing lite/lightning. So was trying to check the GPU usage on kaggle. But couldn't find a way, so thought this might be a good utility. There is no issue related to it.

pytorch_lightning/accelerators/gpu.py
```diff
@@ -83,63 +82,3 @@ def teardown(self) -> None:
     def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return torch.cuda.device_count()
-
-
-def _get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
```
carmocca, line 88: I'm not against making it public, but I would keep it in this file. the utilities file keeps growing and growing

and `from pytorch_lightning.accelerators.gpu import get_nvidia_gpu_stats` seems more natural to me.

awaelchli:
@rohitgr7 Such a problem like you describe would have been perfect for a GitHub issue to discuss. For example, if this was one, then I would have asked what the reasons are for a user to access

- the utility in a device specific way, as opposed to a device agnostic way
- why rely on a particular call to nvidia-smi when one could simply call built-in functions https://pytorch.org/docs/stable/cuda.html#memory-management

rohitgr7:
@awaelchli opened an issue here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10414

pytorch_lightning/accelerators/gpu.py
tchaton: Same here !

pytorch_lightning/utilities/gpu.py
```diff
@@ -0,0 +1,80 @@
...
+        ("temperature.gpu", "°C"),
+        ("temperature.memory", "°C"),
+    ]
+    gpu_stat_keys = [k for k, _ in gpu_stat_metrics]
+    gpu_query = ",".join(gpu_stat_keys)
+
+    gpu_id = _get_gpu_id(device.index)
+    nvidia_smi_path = shutil.which("nvidia-smi")
+    if nvidia_smi_path is None:
+        raise FileNotFoundError("nvidia-smi: command not found")
```
ananthsub, line 51: this should be at the start of the function to fail faster

pytorch_lightning/accelerators/gpu.py
rohitgr7: fine with me. I just want it to be public

rohitgr7: updated!

pytorch_lightning/utilities/gpu.py
rohitgr7: done!

carmocca (approved):

Issue #10349, opened 2021-11-04
Trying to infer the `batch_size` from an ambiguous collection


BttMA:
## 🐛 Bug

Yesterday, I was working with my code and it is was perfect, running with no error. But toady, I keep getting this UserWarning saying that the model trying to infer the `batch_size` from an ambiguous collection. I am not sure where it come from. I did not do any changes on my code. The error keeps getting different batch size with the sale warning : 

```
/usr/local/lib/python3.7/dist-packages/pytorch_lightning/utilities/data.py:57: UserWarning: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 55. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.
  "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
/usr/local/lib/python3.7/dist-packages/pytorch_lightning/utilities/data.py:57: UserWarning: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 1. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.
  "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
/usr/local/lib/python3.7/dist-packages/pytorch_lightning/utilities/data.py:57: UserWarning: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 28. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.
  "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
/usr/local/lib/python3.7/dist-packages/pytorch_lightning/utilities/data.py:57: UserWarning: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 51. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.
  "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
```

### Expected behavior

Running epochs smoothly.

### Environment

- PyTorch Lightning Version: 1.5
- PyTorch Version: 1.9.0
- Python version: 3.7.12
- OS (e.g., Linux): Linux (Colab)
- CUDA/cuDNN version: 11.2
- GPU models and configuration:
- How you installed PyTorch (`conda`, `pip`, source): pip
- If compiling from source, the output of `torch.__config__.show()`:
PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
[...]

PR #9791, opened 2021-10-01
Fix `self.log(on_epoch=True, reduce_fx=sum)` on_batch_start

carmocca:
## What does this PR do?

Minimal repro:

<details>

```python
import os

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer

from pytorch_lightning.callbacks import Callback


class ACallback(Callback):

    def on_train_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
        print(trainer.callback_metrics)

    def on_validation_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
        print(trainer.sanity_checking, trainer.callback_metrics)

    def on_train_batch_start(self, trainer, pl_module, batch, batch_idx, dataloader_idx):
        pl_module.log("on_train_batch_start", 1.0, reduce_fx="sum")

    def on_train_batch_end(self, trainer, pl_module, outputs, batch, batch_idx, dataloader_idx):
        pl_module.log("on_train_batch_end", 1.0, reduce_fx="sum")

    def on_validation_batch_start(self, trainer, pl_module, batch, batch_idx, dataloader_idx):
        print("on_validation_batch_start batch", batch_idx)
        #print("BEFORE", trainer._results.get("on_validation_batch_start.on_validation_batch_start"))
        pl_module.log("on_validation_batch_start", 1.0, reduce_fx="mean")
        #print("AFTER", trainer._results.get("on_validation_batch_start.on_validation_batch_start"))

    def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx, dataloader_idx):
        print("on_validation_batch_end batch", batch_idx)
        #print("BEFORE", trainer._results.get("on_validation_batch_start.on_validation_batch_start"))
        pl_module.log("on_validation_batch_end", 1.0, reduce_fx="sum")
        #print("AFTER", trainer._results.get("on_validation_batch_start.on_validation_batch_start"))


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        loss = self(batch).sum()
[...]

PR #10408, opened 2021-11-08
Fix an issue that caused Lightning to extract the batch size even though it was set by the user in `LightningModule.log`

rohitgr7:
## What does this PR do?

Fixes: https://github.com/PyTorchLightning/pytorch-lightning/issues/10349#issuecomment-962490098
Fixes: #10576

Seems like after we added the warning to check for batch_size for an ambiguous batch, a few users realized that the batch_size used internally isn't correct. We always make a call to extract batch size even when it's set by the user explicitly inside `self.log(batch_size=...)` which IMO is not required because we have a recursive call to extract batch size which causes unnecessary extraction for it even when it's not required plus it will raise a warning if someone specifies it and we still look for it. So now the call will be made inside `.log` only when batch_size is not specified.

In short: this fixes an issue that caused Lightning to extract the batch size even though it was set by the user in `LightningModule.log`.

So an overview of the solution is:
track the batch and use it to extract batch_size inside `self.log` call when it's not specified explicitly and metric is being logged with `on_epoch=True` and will be averaged at the end.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -194,8 +192,6 @@ def advance(self, *args: Any, **kwargs: Any) -> None:
             with self.trainer.profiler.profile("run_training_batch"):
                 batch_output = self.batch_loop.run(batch, batch_idx)
 
-        self.trainer._results.batch_size = batch_size
```
rohitgr7, line 197: reviewer: this is not required since we don't keep track of it now. if it's needed, it will be extracted using the `current_batch` attached to `ResultCollection`

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -253,28 +252,20 @@ def _log_gpus_metrics(self) -> None:
...
     def on_epoch_start(self) -> None:
         self._epoch_end_reached = False
 
-    def on_batch_start(self, batch_idx: int, batch: Any) -> int:
+    def on_batch_start(self, batch_idx: int, batch: Any) -> None:
         self._batch_idx = batch_idx
         self._epoch_end_reached = False
-        return self.on_new_batch(batch)
+        assert self.trainer._results is not None
+        self.trainer._results.current_batch = batch
```
rohitgr7, line 262: reviewer: we keep track of current batch instead of batch size since we are extracting it inside `self.log` call.

pytorch_lightning/core/lightning.py
```diff
@@ -358,122 +359,25 @@ def log(
             rank_zero_only: Whether the value will be logged only on rank 0. This will prevent synchronization which
                 would produce a deadlock as not all processes would perform this log call.
         """
-        if tbptt_reduce_fx is not None:
```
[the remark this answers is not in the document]
carmocca: Please, do not do this change here, keep the PR focused to the fix. This change should be done separately if so.

CHANGELOG.md
```diff
@@ -116,16 +116,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
 
-- Fixed an issue where the model wrapper in Lite converted non-floating point tensors to float ([#10429](https://github.com/PyTorchLightning/pytorch-lightning/pull/10429))
+- Fixed dataloader workers with `persistent_workers` being deleted on every iteration ([#10434](https://github.com/PyTorchLightning/pytorch-lightning/pull/10434))
 
 
--
-
+- Fixed an issue where the model wrapper in Lite converted non-floating point tensors to float ([#10429](https://github.com/PyTorchLightning/pytorch-lightning/pull/10429))
 
--
```
carmocca, line 128: Keep these, they are here to try and avoid conflicts after the CHANGELOG is empty

carmocca (changes requested):

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
carmocca: Why not put that inside `on_new_batch`?

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -356,7 +355,7 @@ class ResultCollection(dict):
     def __init__(self, training: bool, device: Optional[Union[str, torch.device]] = None) -> None:
         super().__init__()
         self.training = training
-        self._batch_size = torch.tensor(1, device=device)
+        self._current_batch = None
```
carmocca, line 358: I'm not a fan of this decision. Saving only the batch size value was done for memory and performance savings.

If each result object saves the current batch, then it's on more reference that we need to clean-up

Why can't the loop do it? I don't want the loop batch logic to leak into the ResultCollection

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: if we do that and someone specifies `self.log(batch_size=...)` and uses an ambiguous batch, then it will raise a warning suggesting user to use `self.log(batch_size=...)` which seems odd. so we can't really extract the batch_size outside the `LightningModule.log` because we don't know until then whether the user specified the batch_size explicitly or not.

pytorch_lightning/trainer/connectors/logger_connector/result.py
rohitgr7: this is cleaned inside `on_batch_end`. we somehow need to access the current batch inside `LightningModule.log`, can't find an alternative way for it. any suggestions?

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: also in the current implementation if someone is not even logging, they will get this warning because we try to extract the batch size regardless of the fact whether its required or not

pytorch_lightning/core/lightning.py
```diff
@@ -428,6 +429,29 @@ def log(
                 "With `def training_step(self, dataloader_iter)`, `self.log(..., batch_size=...)` should be provided."
             )
 
+        reduce_fx = reduce_fx.lower() if isinstance(reduce_fx, str) else reduce_fx
+
+        # check if we have extracted the batch size already
```
tchaton, line 434: Can we move this logic within the ResultCollection instead?

tchaton: I think the ResultCollection should keep track of the batch and recompute the batch_size only if the batch changed and log is been performed.

rohitgr7: done! result collection keeps track of the batch, and update & stores the batch size whenever it's required.

pytorch_lightning/loops/epoch/training_epoch_loop.py
awaelchli: It was added by @carmocca as a fix in #9791 (checked git blame) so I am rather confused by your comment 😅

CHANGELOG.md
```diff
@@ -121,6 +121,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `CombinedLoader` and `max_size_cycle` didn't receive a `DistributedSampler` ([#10374](https://github.com/PyTorchLightning/pytorch-lightning/issues/10374))
 
 
+- Fixed the call to `extract_batch_size` only when its required ([#10408](https://github.com/PyTorchLightning/pytorch-lightning/pull/10408))
```
awaelchli, line 124: The changelog is not clear what the issue was that got fixed. IMO it should always mention the problem. Describing the solution is optional.

pytorch_lightning/loops/epoch/training_epoch_loop.py
rohitgr7: it's changed a bit now.. we do keep track of it inside resultCollection. For tbptt I checked and found we don't split it across the batch_size dimension, rather we do it across time dimension

CHANGELOG.md
rohitgr7:
```suggestion
- Fixed an issue to not extract batch_size when batch_size inside `LightningModule.log` is set by the user ([#10408](https://github.com/PyTorchLightning/pytorch-lightning/pull/10408))
```

rohitgr7: better?

awaelchli: Sorry, I think it is worse 😅 

I think it is either

> Fixed an issue that prevented Lightning to extract the batch size even though it was set by the user in LightningModule.log


or

> Fixed an issue that caused Lightning to extract the batch size even though it was set by the user in LightningModule.log


Totally not sure what you are trying to fix. Given the description of the PR, I probably think it is the latter.

rohitgr7, line 124:
```suggestion
- Fixed an issue that caused Lightning to extract the batch size even though it was set by the user in `LightningModule.log` ([#10408](https://github.com/PyTorchLightning/pytorch-lightning/pull/10408))
```

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
tchaton: I believe this line should clean the previous batch_size.

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -290,6 +282,9 @@ def on_batch_end(self) -> None:
         self._progress_bar_metrics.update(metrics["pbar"])
         self._callback_metrics.update(metrics["callback"])
         self._logged_metrics.update(metrics["log"])
+        assert self.trainer._results is not None
```
tchaton, line 285:
```suggestion
        # drop reference to current batch and batch_size
        assert self.trainer._results is not None
```

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -253,28 +252,21 @@ def _log_gpus_metrics(self) -> None:
...
         self._epoch_end_reached = False
 
-    def on_batch_start(self, batch_idx: int, batch: Any) -> int:
+    def on_batch_start(self, batch_idx: int, batch: Any) -> None:
         self._batch_idx = batch_idx
         self._epoch_end_reached = False
-        return self.on_new_batch(batch)
+        assert self.trainer._results is not None
+        self.trainer._results.current_batch = batch
+        self.trainer._results.current_batch_size = None
```
tchaton, line 263: Add a comment why this line needs to be there.

pytorch_lightning/trainer/connectors/logger_connector/result.py
tchaton: @carmocca We moved the logic within the ResultCollection as the other alternative was to keep this logic within the LightningModule.log method. This would make Trainer / LightningModule separation more complex.

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -458,10 +467,27 @@ def log(
                 f"You called `self.log({name}, ...)` twice in `{fx}` with different arguments. This is not allowed"
             )
 
-        if batch_size is not None:
-            self.batch_size = batch_size
+        # check if we have extracted the batch size already
+        if batch_size is None:
```
tchaton, line 471: Let's move this logic within its own private method

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -575,19 +601,10 @@ def fn(item: ResultMetric) -> None:
...
-        except RecursionError:
-            batch_size = 1
-        self.batch_size = batch_size  # the setter converts it to `Tensor`
-        return batch_size
-
     def to(self, *args: Any, **kwargs: Any) -> "ResultCollection":
         """Move all data to the given device."""
         self.update(apply_to_collection(dict(self), (torch.Tensor, Metric), move_data_to_device, *args, **kwargs))
 
-        self._batch_size = self._batch_size.to(*args, **kwargs)
```
tchaton, line 590: This would still be required with current_batch_size if we were to move the ResultCollection on CPU mid batch iteration.

tests/trainer/logging_/test_train_loop_logging.py
```diff
@@ -715,19 +715,15 @@ def on_validation_epoch_end(self):
             assert all(v == 3 for v in self.trainer.callback_metrics.values())
 
         def on_train_batch_start(self, batch, batch_idx):
-            assert self.trainer._results.batch_size == 2
```
tchaton, line 718: Why did you remove this ?

pytorch_lightning/trainer/connectors/logger_connector/result.py
rohitgr7: I wonder do we even need to move the batch_size to the device at all. It's a scalar and wherever it's used, it's used along with other tensors which are already on their respective devices.

```bash
>>> a = torch.tensor(10.).cuda()
>>> a + 7
torch(17., device='cuda:0')
>>> a * 7
torch(70., device='cuda:0')
```

tests/trainer/logging_/test_train_loop_logging.py
rohitgr7: because reduce_fx = 'sum', so we don't track the batch_size in such a case.

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: done

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: done

pytorch_lightning/trainer/connectors/logger_connector/result.py
rohitgr7: done

pytorch_lightning/trainer/connectors/logger_connector/result.py
tchaton: This is done to ensure there is no memory left on the GPU. We have tests for this and they would fail.

awaelchli (approved):

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -390,14 +392,22 @@ def append_fn(v: ResultMetric) -> None:
...
+    def _extract_batch_size(self, batch_size: Optional[int], meta: _Metadata) -> int:
+        # check if we have extracted the batch size already
+        batch_size = batch_size or self.batch_size
+        if batch_size is not None:
+            return batch_size
+        batch_size = 1
+
+        # extract it
+        fx_validate = _FxValidator.functions.get(meta.fx.split(".")[0])
+        if meta.on_epoch and fx_validate is not None and (True in fx_validate["on_step"]) and meta.is_mean_reduction:
```
carmocca, line 404: why exactly do we need the `True in fx_validate["on_step"]` check?

carmocca (commented):
LGTM! I took the liberty of pushing several commits since it was easier than posting comments. Please take a look at the changes.

Two comments and you'll have my approval ✅

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -390,14 +392,22 @@ def append_fn(v: ResultMetric) -> None:
...
+            try:
+                batch_size = extract_batch_size(self.batch)
+                self.batch_size = batch_size
+            except RecursionError:
```
ninginthecloud, line 408: @rohitgr7  There could be other exceptions coming from `extract_batch_size()`. For example, if batch is ```
batch={'a':torch.tensor([1, 2, 3, 4, 5]), 'b': [torch.tensor(1), torch.tensor(2)]}
```
it raises index error when batch.size(0) is called for torch.tensor(1). 
I understand the good intention to automatically retrieve batch_size from batch, but we can't control how user set up their batch input. It's better to involve  user to provide batch_size.

File: tests/trainer/logging_/test_train_loop_logging.py [hunk 15]
```diff
@@ -749,3 +746,36 @@ def validation_epoch_end(self, *_) -> None:
...
+def test_no_batch_size_extraction_with_specifying_explictly(tmpdir):
+    batch_size = BoringModel().train_dataloader().batch_size + 1
+    fast_dev_run = 2
+    log_val = 7
+
+    class CustomBoringModel(BoringModel):
+        def on_before_batch_transfer(self, batch, *args, **kwargs):
+            # This is an ambiguous batch which have multiple potential batch sizes
+            if self.trainer.training:
+                batch = {"batch1": torch.randn(batch_size, 10), "batch2": batch}
```
ninginthecloud, line 760: Hi, @rohitgr7  could we add more test cases to reflect users have mentioned in issue #10576 ?  This test case only captures a very ideal situation.

tests/trainer/logging_/test_train_loop_logging.py
```diff
@@ -749,3 +746,36 @@ def validation_epoch_end(self, *_) -> None:
     train_data = DataLoader(RandomDataset(32, 64), batch_size=2)
     val_data = DataLoader(RandomDataset(32, 64), batch_size=2)
     trainer.fit(model, train_dataloaders=train_data, val_dataloaders=val_data)
+
+
+def test_no_batch_size_extraction_with_specifying_explictly(tmpdir):
```
ninginthecloud, line 751: Maybe I have missed here. do we have test case when user set batch_size in `LightningModule.log` and our code did not spam warning for ambiguous batch?

ninginthecloud (commented):
Hi, @rohitgr7  Thanks a lot for tackling this issue! I left a few comments in code, feel free to let me know if you have any questions. 
A generic idea from my point of view, I think `extract_data_batch()` is this kind function we could convert as hook and delegate to user to implement, because they know their batch, it will be easily for them to define the batch_size or how to get batch_size. While here we had a lot assumptions about what the batch is going to look like, but we still can only extract valid batch_size under very limited situations.  
Anyway, I don't want to interrupt this PR discussion thread, I will initialize an issue to start the discussion. 
cc: @tchaton @ananthsub

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -390,14 +392,22 @@ def append_fn(v: ResultMetric) -> None:
...
-    def batch_size(self) -> torch.Tensor:
-        # performance: cache the `batch_size` tensor instead of re-creating it
-        return self._batch_size
-
-    @batch_size.setter
-    def batch_size(self, value: int) -> None:
-        self._batch_size = torch.tensor(value, device=self.device)
+    def _extract_batch_size(self, batch_size: Optional[int], meta: _Metadata) -> int:
+        # check if we have extracted the batch size already
+        batch_size = batch_size or self.batch_size
```
rohitgr7, line 397: @carmocca thanks for updates.. looks good.
this will fail with `self.log(..., batch_size=0)`. if the user is making a mistake, we shouldn't touch it.

pytorch_lightning/trainer/connectors/logger_connector/result.py
rohitgr7: well, we need to ensure that batch_size should only be extracted for the cases where step-level logging is allowed. Also in other cases `results.batch` will be None too. We can avoid this check by allowing the call for `extract_batch_size(None)` and returning 1 from there, but it didn't seem ideal to me.

pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -390,14 +392,25 @@ def append_fn(v: ResultMetric) -> None:
...
+        if batch_size is not None:
+            return batch_size
 
-    @batch_size.setter
-    def batch_size(self, value: int) -> None:
-        self._batch_size = torch.tensor(value, device=self.device)
+        # extract it
+        fx_validate = _FxValidator.functions.get(meta.fx.split(".")[0])
```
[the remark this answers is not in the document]
rohitgr7: we append the dataloader_idx to meta.fx if we have multiple dataloaders

tests/trainer/logging_/test_train_loop_logging.py
rohitgr7: yes. 2nd log inside `training_step`
```
 self.log("epoch_log_val", log_val, batch_size=batch_size, ...)
```

pytorch_lightning/trainer/connectors/logger_connector/result.py
rohitgr7: good catch! I guess this is the only edge case we had with tensors. Added a check for it.

tchaton:
> Hi, @rohitgr7 Thanks a lot for tackling this issue! I left a few comments in code, feel free to let me know if you have any questions. A generic idea from my point of view, I think `extract_data_batch()` is this kind function we could convert as hook and delegate to user to implement, because they know their batch, it will be easily for them to define the batch_size or how to get batch_size. While here we had a lot assumptions about what the batch is going to look like, but we still can only extract valid batch_size under very limited situations. Anyway, I don't want to interrupt this PR discussion thread, I will initialize an issue to start the discussion. cc: @tchaton @ananthsub

Hey @ninginthecloud, I believe this is a great idea and this would make the logic more efficient. Personally, I am in favor of this.

@awaelchli @carmocca

rohitgr7:
> Hi, @rohitgr7 Thanks a lot for tackling this issue! I left a few comments in code, feel free to let me know if you have any questions. A generic idea from my point of view, I think `extract_data_batch()` is this kind function we could convert as hook and delegate to user to implement, because they know their batch, it will be easily for them to define the batch_size or how to get batch_size. While here we had a lot assumptions about what the batch is going to look like, but we still can only extract valid batch_size under very limited situations. Anyway, I don't want to interrupt this PR discussion thread, I will initialize an issue to start the discussion. cc: @tchaton @ananthsub

if we do this, then we will have two different ways by which user can define their batch size. One will be this hook and another one will be `self.log(..., batch_size=)`. They still can define methods within their lightning module and can call it to retrieve the batch_size and use it within `self.log`. Plus we already have too many hooks and this seems more like a utility method to me. Just a personal opinion :)

tchaton:
To add to @rohitgr7 point,

> self.log(..., batch_size=) was more a rescue solution and I personally don't find it elegant to provide the batch size, especially if you are logging outside of the *_step methods of the LightningModule where the current batch is unavailable. But this might affect a minority of users.

> Plus we already have too many hooks
This is entirely true and we want to enforce only 1 way to do anything within Lightning. I believe we would have to weigh the hook vs the self.log(..., batch_size=) in terms of usability/simplicity. 

@carmocca @awaelchli @SeanNaren  Any thoughts ?

rohitgr7:
@tchaton let's discuss more here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10635

tests/trainer/logging_/test_train_loop_logging.py
rohitgr7: well this test is just to check in what cases we should make a call to `extract_batch_size`. for the warning/error check we have tests here: https://github.com/PyTorchLightning/pytorch-lightning/blob/29797720cdbd74f5440c8e7186c82fe39b2002aa/tests/utilities/test_data.py#L18

Since I am not super familiar with torch-geometric, feel free to add some more here :)

carmocca (approved):

tests/trainer/logging_/test_train_loop_logging.py
carmocca: Resolving this, we can add more when we find new examples that do not work

PR #10415, opened 2021-11-08
Remove DDP teardown move model to CPU

### LINKED ISSUE #10294: teardown function casts model to cpu after testing
https://github.com/PyTorchLightning/pytorch-lightning/blob/6e124e7207f6459cb43f540cfb5a1c6cc9b00f7a/pytorch_lightning/plugins/training_type/single_device.py#L83-L88

Issue: This part of code forces model to be on CPU after the testing is finished

expected behaviour: the teardown logic after testing should not change the model device

reproducible code snippet:

```
    # model.device --> 'cuda'
    test_loader = data_module.test_dataloader()
    tester = pl.Trainer(precision=precision, gpus=gpus, logger=False)
    test_results = tester.test(model, test_loader)
    # model.device --> 'cpu'
```

cc @borda @akihironitta

four4fish:
## What does this PR do?

I didn't see logic depends on ttp teardown to move model to cpu after fit or test

Other than the error mentioned in the issue, we saw internal failures for FSDP. Model is too big doesn't fit into cpu which caused OOM. 

Fixes #10294 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/plugins/training_type/ddp.py
+++ b/pytorch_lightning/plugins/training_type/ddp.py
@@ -525,7 +525,5 @@ class DDPPlugin(ParallelPlugin):
             self.model = self.lightning_module
 
         if self.on_gpu:
-            # GPU teardown
-            self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
--- a/pytorch_lightning/plugins/training_type/ddp_spawn.py
+++ b/pytorch_lightning/plugins/training_type/ddp_spawn.py
@@ -419,7 +419,5 @@ class DDPSpawnPlugin(ParallelPlugin):
             self.model = self.lightning_module
 
         if self.on_gpu:
-            # GPU teardown
-            self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
--- a/pytorch_lightning/plugins/training_type/ddp_spawn.py
+++ b/pytorch_lightning/plugins/training_type/ddp_spawn.py
@@ -419,5 +419,7 @@ class DDPSpawnPlugin(ParallelPlugin):
             self.model = self.lightning_module
 
         if self.on_gpu:
+            # GPU teardown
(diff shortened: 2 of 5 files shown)


justusschock:
@four4fish The idea behind this is that the model is usually on CPU before fit and thus after the entrypoint we want to reset everything back to it's original state (except the weights of course ^^) to have a minimal impact.

awaelchli:
To address the linked issue, wouldn't we need to move the model back to the device it was originally on?

four4fish:
> To address the linked issue, wouldn't we need to move the model back to the device it was originally on?

Hi @awaelchli, I thought user is `Issue: This part of code forces model to be on CPU after the testing is finished; expected behaviour: the teardown logic after testing should not change the model device` which means we don't want to move the model to cpu?

Also we saw failures in internal FB for FSDP. Model is too big doesn't fit into cpu which caused OOM.
For DDP, all rank trying to move model back to cpu which cause OOM too (I need go verify this more)

awaelchli:
Given the code snippet in the linked issue and the comment from @ananthsub https://github.com/PyTorchLightning/pytorch-lightning/issues/10294#issuecomment-956455783 I interpreted that the goal was to leave the model on the device it was originally on (before fit). This is strictly speaking not the same as not moving it back to CPU.

**Just one important remark everyone should be aware of:** 
The code in the teardown method moving the model to CPU is there for a reason, I am 99% sure. The problem is that it is not documented, so we don't remember why. There is a chance that if we make the change then we also break what this was fixing originally. There is also a chance that it has become redundant. 
In any case, it's a breaking change: The user needs to add a line `model.cpu()` to preserve the behavior if they need.

> Also we saw failures in internal FB for FSDP. Model is too big doesn't fit into cpu which caused OOM.
> For DDP, all rank trying to move model back to cpu which cause OOM too (I need go verify this more

Have we considered changing `teardown()` for FSDP?

PR #10483, opened 2021-11-11
Fix SignalConnector._has_already_handler check for callable type

mauvilsa:
## What does this PR do?

Fixes an issue where a user's registered SIGTERM signal handler is overwritten by `SignalConnector.register_signal_handlers`. One case in which this happens is when a method of a class instance is registered. That is:

```python
import signal

class Callback:
    def __init__(self):
        signal.signal(signal.SIGTERM, self.sigterm_handler)
    def sigterm_handler(self):
        pass

callback = Callback()
```

The current check `isinstance(signal.getsignal(signal.SIGTERM), FunctionType)` evaluates to false, thus overwriting it.

No related issue. This was discussed over slack.

## Before submitting

## PR review

## Did you have fun?


carmocca (commented):
LGTM. Can you add a test to avoid regressions? Can be a simple unit-test. `_has_already_handler` can be made static for it.

mauvilsa:
> Can you add a test to avoid regressions? Can be a simple unit-test.

Done.

tests/trainer/connectors/test_signal_connector.py
```diff
@@ -57,3 +58,27 @@ def training_step(self, batch, batch_idx):
         else:
             trainer.fit(model)
         assert trainer._terminate_gracefully == (False if register_handler else terminate_gracefully)
+
+
+def signal_handler():
+    pass
+
+
+class C:
```
awaelchli, line 67:
```suggestion
class SignalHandlers:
```

tests/trainer/connectors/test_signal_connector.py
```diff
@@ -57,3 +58,27 @@ def training_step(self, batch, batch_idx):
...
+@pytest.mark.parametrize(
+    ["handler", "expected_return"],
+    [
+        (signal.Handlers.SIG_DFL, False),
+        (signal_handler, True),
+        (C().signal_handler, True),
+    ],
+)
+def test_has_already_handler(handler, expected_return):
+    trainer = Trainer()
```
awaelchli, line 81:
```suggestion
def test_has_already_handler(handler, expected_return):
    """Test that the SignalConnector detects whether a signal handler is already attached."""
    trainer = Trainer()
```

awaelchli (approved):

carmocca (approved):

tchaton (approved):
LGTM !

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -94,7 +94,7 @@ def _is_on_windows(self) -> bool:
 
     def _has_already_handler(self, signum: Signals) -> bool:
         try:
-            return isinstance(signal.getsignal(signum), FunctionType)
+            return callable(signal.getsignal(signum))
```
carmocca, line 97: Seeing test failures here

```python
_________________ test_fault_tolerant_sig_handler[True-False] __________________

register_handler = False, terminate_gracefully = True
tmpdir = local('/tmp/pytest-of-root/pytest-0/test_fault_tolerant_sig_handle0')

    @pytest.mark.parametrize("register_handler", [False, True])
    @pytest.mark.parametrize("terminate_gracefully", [False, True])
    @RunIf(skip_windows=True)
    def test_fault_tolerant_sig_handler(register_handler, terminate_gracefully, tmpdir):
    
        if register_handler:
    
            def handler(*_):
                pass
    
            signal.signal(signal.SIGUSR1, handler)
    
        class TestModel(BoringModel):
                if terminate_gracefully or register_handler:
                    os.kill(os.getpid(), signal.SIGUSR1)
                    sleep(0.1)
    
        model = TestModel()
    
        with mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": str(int(terminate_gracefully))}):
    
            if terminate_gracefully and not register_handler:
>                   trainer.fit(model)
E                   Failed: DID NOT RAISE <class 'pytorch_lightning.utilities.exceptions.ExitGracefullyException'>

/__w/pytorch-lightning/pytorch-lightning/tests/trainer/connectors/test_signal_connector.py:56: Failed
```

awaelchli:
Interesting, wasn't aware of this PR, but I have also one here #10611 where I changed the `has_already_sampler` to return

`return signal.getsignal(signum) is not signal.SIG_DFL`

Is the check for callable necessary? Which way is the correct way to check this, anyone knows? :)

pytorch_lightning/trainer/connectors/signal_connector.py
mauvilsa: I don't know why this is failing. Locally all tests in `test_signal_connector.py` are run and are successful. And the initial runs for this pull request were successful, see e.g. https://github.com/PyTorchLightning/pytorch-lightning/actions/runs/1451925254. So it must be related to some other pull request that was merged recently.

mauvilsa:
> Is the check for callable necessary? Which way is the correct way to check this, anyone knows? :)

I guess the callable is not necessary. What should be the behavior if the value is `signal.SIG_IGN` which is part of the same enum that defines `signal.SIG_DFL`.

awaelchli:
Good point, I think we should respect it and not add a handler in that case too?

Borda (approved):

pytorch_lightning/trainer/connectors/signal_connector.py
awaelchli: Hey, I added a few commits here but couldn't solve the issue. But I believe it could be related to the missing teardown I have seen over here #10611. In fact, if I include your change in signal connector, the tests pass there. So I suggest we merge my teardown fix first, then attempt to run tests again here.

Issue #10362, opened 2021-11-04
`RichProgressBar` deadlocks distributed training


bryant1410:
## 🐛 Bug

I have run into many distributed training now that have gone completely frozen after enabling the `RichProgressBar` callback.

It gets stuck in between epochs after it finished one. Sometimes it's after the first one, sometimes after the second one, sometimes after even 9 epochs.

The weird thing is that if I do `Ctrl+C`, the program doesn't interrupt. I have to kill it with SIGKILL (`kill -9`) because SIGTERM doesn't do it. So I don't know the stack trace. I'm also inside a Docker container, so I can't do `strace` (I don't have access to the host computer). Any help here to check the stack trace is welcome.

### To Reproduce

I'm sorry but I can't share my code. All I can say is that I use a machine with 8 GPUs and the `ddp_find_unused_parameters_false` strategy, and that the problem only appears with `RichProgressBar` and that otherwise it doesn't.

I'd be happy to provide more details or try stuff out. Ask me for it! But I can't share my code, I'm sorry.

### Expected behavior

The training to continue successfully.

### Environment

- PyTorch Lightning Version (e.g., 1.3.0): 1.5.0
- PyTorch Version (e.g., 1.8): 1.10.0
- Python version: 3.8.12
- OS (e.g., Linux): Linux
- CUDA/cuDNN version: 11.3
- GPU models and configuration: 8x A100
- How you installed PyTorch (`conda`, `pip`, source): conda
- If compiling from source, the output of `torch.__config__.show()`:
- Any other relevant information:

bryant1410:
I'm using rich v10.12.0 from Conda.

tchaton:
Dear @bryant1410,

Are you using ddp_spawn ? We noticed similar issue within our own CI and didn't manage to resolve it yet.

Best,
T.C

SeanNaren:
Hey @bryant1410 could you try this branch?

https://github.com/PyTorchLightning/pytorch-lightning/tree/feat/enable_rich_default
should be able to do something like `pip install https://github.com/PyTorchLightning/pytorch-lightning/archive/refs/heads/feat/enable_rich_default.zip`

kaushikb11:
@bryant1410 It's okay if you can't share your code, but could you try running and reproducing it in a non Conda environment? Thank you!

bryant1410:
> Are you using ddp_spawn ? We noticed similar issue within our own CI and didn't manage to resolve it yet.

[...]

PR #9647, opened 2021-09-22
Enable RichProgressBar by default if Rich is available

SeanNaren:
## What does this PR do?

Closes #9580. By default if Rich (10.0.2 or later) is available, we use the Rich Progress Bar! 

Related to #9616, we throw a warning if the user sets `progress_bar_refresh_rate` from the Trainer and default to TQDM (the `ProgressBar`). This is because the meaning of refresh rate is different for Rich/TQDM, and are not related. Since in #9616 the argument is being deprecated, I felt it made sense to ask the users to pass the `RichProgressBar` to callbacks like `callbacks=RichProgressBar(refresh_rate_per_second=20)`.

cc @daniellepintz 

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #10497, opened 2021-11-12
Add `refresh_rate` to RichProgressBar

### LINKED ISSUE #10607: pdb - based debugging with RichProgressBar doesn't show input
## 🐛 Bug

RichProgressBar swallows the input for pdb or pdbr which makes debugging with those tools virtually impossible.
This is a known issue in rich which does not yet have a solution: [See here](https://github.com/willmcgugan/rich/issues/1465)
With RichProgressBar on the way of replacing tqdm as default it would be good to either document a workaround, or to reconsider that decision. 
 
### To Reproduce
Use RichProgressBar in the Trainer callback and put
```python
import pdb; pdb.set_trace()
```
or
```python
breakpoint()
```
somewhere inside the `training_step(...)` or `validation_step(...)` of your module.

If the execution stops at that breakpoint you will still get the (Pdb) repl, which also works. However the input will not be shown which makes you blind when typing pdb commands.

For example with the RichProgressBar:
```python
...
(Pdb) 
Validation Sanity Check ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0/2 0:00:08 • -:--:-- 0.00it/s
'Hello World!' 
(Pdb)
```
Note that the input `p "Hello World!" is not shown. It should appear in the repl behind (Pdb)
The same Example with ProgressBar
```python
...
(Pdb) p "Hello World!"
'Hello World!'
(Pdb)
``` 
### Expected behavior
The input to pdb is shown when typed.

### Environment

- PyTorch Lightning Version (e.g., 1.3.0): 1.5.1
- PyTorch Version (e.g., 1.8): 1.10.0+cu113
- Python version: 2.8.10
- OS (e.g., Linux): Linux
- CUDA/cuDNN version: 11.3 / 8.2
- GPU models and configuration: Not relevant

### Additional context

This is not a bug that can/should be fixed in pytorch-lighting. However now that the progress-bar is in PytorchLightning we need this to be pointed out in the documentation.

kaushikb11:
## What does this PR do?

closes #10607 

Changed `refresh_rate_per_second` param to `referesh_rate` for `RichProgressBar` signature to match `TQDMProgressBar`'s interface

## Before submitting

## PR review

## Did you have fun?


CHANGELOG.md
```diff
@@ -121,7 +121,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `CombinedLoader` and `max_size_cycle` didn't receive a `DistributedSampler` ([#10374](https://github.com/PyTorchLightning/pytorch-lightning/issues/10374))
 
 
--
+- Changed `refresh_rate_per_second` param to `referesh_rate` for `RichProgressBar` signature ([#10497](https://github.com/PyTorchLightning/pytorch-lightning/pull/10497))
```
awaelchli, line 124: If this gets included as a fix I think we should mention the issue. For example:

```suggestion
- Fixed misleading argument name `refresh_rate_per_second` in `RichProgressBar` signature by renaming it to `referesh_rate` ([#10497](https://github.com/PyTorchLightning/pytorch-lightning/pull/10497))
```
or move it to the changed section is also an option

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -289,14 +285,18 @@ def _init_progress(self, trainer):
             self.progress = CustomProgress(
                 *self.configure_columns(trainer),
                 self._metric_component,
-                refresh_per_second=self.refresh_rate_per_second,
+                auto_refresh=False,
```
awaelchli, line 288: I think this is the fix I proposed in #9647 to @SeanNaren to prevent threading issues in the render function.
Did you check that this might enable #9647 to pass the CI now?

kaushikb11: Talked to Sean regarding it, it didn't work!

CHANGELOG.md
kaushikb11: Better option! Updated

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -222,7 +223,7 @@ class RichProgressBar(ProgressBarBase):
 
     def __init__(
         self,
-        refresh_rate_per_second: int = 10,
+        refresh_rate: int = 1,
```
ananthsub, line 226: This is a breaking change. Removing `refresh_rate_per_second` should follow the deprecation process before its removal

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -289,14 +285,18 @@ def _init_progress(self, trainer):
...
-                refresh_per_second=self.refresh_rate_per_second,
+                auto_refresh=False,
                 disable=self.is_disabled,
                 console=self._console,
             )
             self.progress.start()
             # progress has started
             self._progress_stopped = False
 
+    def refresh(self):
```
ananthsub, line 296:
```suggestion
    def refresh(self) -> None:
```

could you add mypy/typehint coverage for this class in a separate PR?

kaushikb11: Yup!

pytorch_lightning/callbacks/progress/rich_progress.py
kaushikb11: I and Sean discussed about it, it would be really messy and unintuitive to support both the parameters for RichProgressBar. Also, as we do manual refresh now in this PR

SeanNaren:
As discussed offline, I'm not sure I agree with this change unless it allows us to move towards making this progress bar default.

My argument was that this is closer to Rich's capabilities, and allows the progress bar to be updated outside of the main thread (meaning time elapses without progress being made). I also wonder how it looks when rendered in a notebook without per second refresh.

I understand the reasoning that its nice to be close to the TQDMProgressBar, but since now both are imported as separate callbacks + the argument isn't exposed anymore in the trainer, maybe the functionalities can be different. Thoughts?

pytorch_lightning/callbacks/progress/rich_progress.py
tchaton: Also, 

I would recommend the following for the default value:

```py
import os
from typing import Optional

class TQDMProgressBar(ProgressBarBase):

    def __init__(self, refresh_rate: Optional[int] = None, process_position: int = 0):
        super().__init__()

        self._refresh_rate = self._resolve_refresh_rate(refresh_rate)

    # moved in the `ProgressBarBase`
    @staticmethod
    def _resolve_refresh_rate(refresh_rate: Optional[int]) -> int:
        if refresh_rate is None:
            refresh_rate = 20 if os.getenv("COLAB_GPU", "1") == "1" else 1
        return refresh_rate
```

IMO, the progress bar should be responsible to set up properly for colab.

SeanNaren (approved):
Looks good! Thanks for this @kaushikb11.

Regarding the breaking change, we could make a deprecation process for it but I'm indifferent here.

tchaton (approved):
LGTM!

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -364,52 +367,63 @@ def on_validation_epoch_start(self, trainer, pl_module):
...
 
     def on_validation_epoch_end(self, trainer, pl_module):
         super().on_validation_epoch_end(trainer, pl_module)
         if self.val_progress_bar_id is not None:
-            self._update(self.val_progress_bar_id, visible=False)
+            self._update(self.val_progress_bar_id, self.val_batch_idx, self.total_val_batches, visible=False)
 
     def on_test_epoch_start(self, trainer, pl_module):
-        super().on_train_epoch_start(trainer, pl_module)
+        super().on_test_epoch_start(trainer, pl_module)
```
ananthsub, line 392: is this a bugfix?

ananthsub (approved):

pytorch_lightning/callbacks/progress/rich_progress.py
kaushikb11: Nope, it's not even required actually. Updated!

CHANGELOG.md
```diff
@@ -37,7 +37,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Raise `MisconfigurationException` when `enable_progress_bar=False` and a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
 
 
--
+- Changed `refresh_rate_per_second` param to `referesh_rate` for `RichProgressBar` signature ([#10497](https://github.com/PyTorchLightning/pytorch-lightning/pull/10497))
```
awaelchli, line 40:
```suggestion
- Renamed `refresh_rate_per_second` parameter to `referesh_rate` for `RichProgressBar` signature ([#10497](https://github.com/PyTorchLightning/pytorch-lightning/pull/10497))
```

awaelchli (approved):

pytorch_lightning/callbacks/progress/rich_progress.py
akihironitta: Hi @kaushikb11, I'm working on #13937 and debugging it has led me to this PR. Do you remember if there is any particular reason why we changed `auto_refresh` from `True` to `False`? Was it for #10362?

Issue #9932, opened 2021-10-14
TODOs for achieving stability for Trainer Strategy API


kaushikb11:
### Motivation

We have added the `strategy` argument for Trainer in [this](https://github.com/PyTorchLightning/pytorch-lightning/pull/8597) PR.

### The next immediate refactors required are:
- [x] Tests #9931 
- [x] Code/warnings updates #9937 
- [x] Docs: general usage #10000 
- [x] Docs: Trainer flag documentation (accelerator, strategy) #10042 
- [x] tutorials

### Post the above refactors, we will be going through the below internal updates:

- [x]  Remove DistributedType.DDP_CPU. This should not be a valid `strategy` value.
- [x]  Remove DistributedType.TPU_SPAWN
- [x]  Rename DistributedType to StrategyType #10505
- [x]  Rename DeviceType to AcceleratorType #10503 
- [x]  Remove user-facing distributed_backend references in code and error messages. #10017
- [x]  Rename the inernal `self.distributed_backend` references.
- [x]  Rename AcceleratorConnector._device_type to _accelerator_type
- [x]  Rename AcceleratorConnector._distrib_type to _strategy_type
- [ ]  Define all non-supported accelerator-strategy combinations. Is this already done?
- [x]  Rename TrainingTypePlugin and its references to StrategyPlugin #10548 
- [ ] Remove "test_accelerator_x" tests in `tests/accelerators/test_accelerator_connector.py` that test for the deprecated usage of `accelerator="strategy_name"`

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

</sub>

daniellepintz:
Do we want to also rename DeviceType here to AcceleratorType?
[...]

Issue #9956, opened 2021-10-15
Is `precision="mixed"` redundant?


carmocca:
## Proposed refactoring or deprecation

Does `precision="mixed"` act differently to `precision=16` in any way?

I understand that "mixed" is more correct as 16-bit precision can still run some computations in 32-bit.

### Motivation

In https://github.com/PyTorchLightning/pytorch-lightning/pull/9763 I noticed we did not even have a `PrecisionType` for `"mixed"`.

There's a single test in the codebase passing the "mixed" value:

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/plugins/test_deepspeed_plugin.py#L153

And no mention at all of this value in the docs.

### Pitch

Have one value to set this, whether it is `16` or `"mixed"`. Most likely `16` since its the one widely used.

Otherwise, add tests for passing `"mixed"`

### Additional context

```python
$ grep -iIrn '"mixed"' pytorch_lightning
pytorch_lightning/plugins/training_type/sharded.py:62:                    is_fp16 = precision in ("mixed", 16)
pytorch_lightning/plugins/training_type/fully_sharded.py:135:            mixed_precision=precision == "mixed",
pytorch_lightning/plugins/training_type/ipu.py:42:        if self.precision in ("mixed", 16):
pytorch_lightning/plugins/training_type/deepspeed.py:405:            dtype = torch.float16 if self.precision in (16, "mixed") else torch.float32
pytorch_lightning/plugins/training_type/deepspeed.py:473:            dtype = torch.float16 if self.precision in (16, "mixed") else torch.float32
pytorch_lightning/plugins/training_type/deepspeed.py:602:        if precision in (16, "mixed"):
pytorch_lightning/plugins/precision/mixed.py:26:    precision: Union[str, int] = "mixed"
pytorch_lightning/plugins/precision/fully_sharded_native_amp.py:26:    precision = "mixed"
```

```python
$ grep -iIrn '"mixed"' tests
tests/plugins/test_deepspeed_plugin.py:153:@pytest.mark.parametrize("precision", [16, "mixed"])
```

```python
$ grep -Irn 'mixed' docs | grep 'precision='
# no mention in the docs!
```
______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

[...]

PR #10503, opened 2021-11-12
Deprecate `DeviceType` in favor of `_AcceleratorType`

kaushikb11:
## What does this PR do?

Part of #9932 

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @akihironitta


pytorch_lightning/utilities/enums.py
```diff
@@ -111,16 +111,16 @@ def is_interactive_compatible(self) -> bool:
     DDP_FULLY_SHARDED = "ddp_fully_sharded"
 
 
-class DeviceType(LightningEnum):
```
ethanwharris, line 114: Can we deprecate this rather than delete? See example of Enum which prints a deprecation message from Flash: https://github.com/PyTorchLightning/lightning-flash/blob/79b943f858417717b926dcf4bd4d3e2be0788a1f/flash/core/data/data_source.py#L21

kaushikb11: This is cool! https://github.com/PyTorchLightning/lightning-flash/blob/79b943f858417717b926dcf4bd4d3e2be0788a1f/flash/core/utilities/on_access_enum_meta.py#L4

kaushikb11: Added in #10505

pytorch_lightning/lite/lite.py
```diff
@@ -469,11 +469,11 @@ def _check_strategy_support(self, strategy: Optional[Union[str, TrainingTypePlug
             )
 
     @staticmethod
-    def _supported_device_types() -> Sequence[DeviceType]:
+    def _supported_device_types() -> Sequence[AcceleratorType]:
         return (
-            DeviceType.CPU,
-            DeviceType.GPU,
-            DeviceType.TPU,
+            AcceleratorType.CPU,
```
tchaton, line 474: Should this be private similar to _StrategyType ?

pytorch_lightning/utilities/enums.py
tchaton: Dope !

pytorch_lightning/lite/lite.py
```diff
@@ -470,11 +470,11 @@ def _check_strategy_support(self, strategy: Optional[Union[str, TrainingTypePlug
             )
 
     @staticmethod
-    def _supported_device_types() -> Sequence[DeviceType]:
+    def _supported_device_types() -> Sequence[_AcceleratorType]:
```
kaushikb11, line 473: Could change the method name as well

pytorch_lightning/utilities/enums.py
```diff
@@ -256,3 +257,22 @@ def interactive_compatible_types() -> List["_StrategyType"]:
     def is_interactive_compatible(self) -> bool:
         """Returns whether self is interactive compatible."""
         return self in _StrategyType.interactive_compatible_types()
+
+
+class _AcceleratorType(LightningEnum):
```
justusschock, line 262: Why is this protected now?

pytorch_lightning/utilities/enums.py
```diff
@@ -256,3 +257,22 @@ def interactive_compatible_types() -> List["_StrategyType"]:
...
+    >>> # you can match the type with string
+    >>> _AcceleratorType.GPU == 'GPU'
+    True
+    >>> # which is case invariant
+    >>> _AcceleratorType.TPU in ('tpu', 'CPU')
+    True
+    """
+
+    CPU = "CPU"
+    GPU = "GPU"
```
justusschock, line 276: Just a question: Should we also change `GPU` to `CUDA` to also allow for ROCm devices in the future?

justusschock (approved):

pytorch_lightning/utilities/enums.py
kaushikb11: Then user would have to pass `accelerator="cuda"`, are we okay with that change?

pytorch_lightning/utilities/enums.py
kaushikb11: Related discussion: https://github.com/PyTorchLightning/pytorch-lightning/pull/10505#discussion_r748313996

pytorch_lightning/utilities/enums.py
tchaton: This would be an additional breaking change, but definitely cleaner and more aligned with PyTorch API. @ananthsub.

I believe it might be worth waiting for adding support `ROCm` as the GPUAccelerator might be renamed CUDAAccelerator too.

justusschock: We could also add it now and deprecate the gpu thing in the meantime. Fine with both ways :)

rohitgr7 (approved):

pytorch_lightning/utilities/enums.py
```diff
@@ -141,24 +141,25 @@ def deprecate(self) -> None:
...
-    True
-    >>> # you can match the type with string
-    >>> DeviceType.GPU == 'GPU'
-    True
-    >>> # which is case invariant
-    >>> DeviceType.TPU in ('tpu', 'CPU')
-    True
+    Deprecated since v1.6.0 and will be removed in v1.8.0.
+
+    Use `__AcceleratorType` instead.
```
carmocca, line 149:
```suggestion
    Use `_AcceleratorType` instead.
```

pytorch_lightning/utilities/enums.py
carmocca: I would like to avoid adding things prematurely which are not used yet. Avoids issues like #9956

carmocca (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -1598,7 +1604,7 @@ def _distrib_type(self) -> _StrategyType:
         return self._accelerator_connector._distrib_type
 
     @property
-    def _device_type(self) -> DeviceType:
+    def _device_type(self) -> _AcceleratorType:
```
awaelchli, line 1607: will you also rename this?
Since this property is protected and directly tied to this enum, you could rename it here too.

pytorch_lightning/utilities/enums.py
```diff
@@ -143,24 +143,25 @@ def deprecate(self) -> None:
...
 
     CPU = "CPU"
     GPU = "GPU"
     IPU = "IPU"
     TPU = "TPU"
 
+    def deprecate(self) -> None:
+        rank_zero_deprecation(
+            "`DeviceType` Enum has been deprecated in v1.6 and will be removed in v1.8."
+            " Use the string value `{self.value!r}` instead."
```
awaelchli, line 162: I don't understand this. Shouldn't it be "use _AcceleratorType instead"? Or rather delete this, because there is no real alternative on the public api side.

pytorch_lightning/utilities/model_summary.py
```diff
@@ -261,7 +261,11 @@ def _forward_example_input(self) -> None:
         input_ = model.example_input_array
         input_ = model._apply_batch_transfer_handler(input_)
 
-        if trainer is not None and trainer.amp_backend == AMPType.NATIVE and trainer._device_type != DeviceType.TPU:
+        if (
+            trainer is not None
+            and trainer.amp_backend == AMPType.NATIVE
+            and trainer._device_type != _AcceleratorType.TPU
```
awaelchli, line 267:
```suggestion
            and not isinstance(trainer.accelerator, TPUAccelerator)
```
Outside of the trainer/accelerator connector, this is how we want to check for the accelerator type in the future. Just a reminder, let's update this in follow up PRs

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
kaushikb11: Let me do a follow up for this with multiple other changes

pytorch_lightning/utilities/model_summary.py
kaushikb11: Yup!

pytorch_lightning/utilities/enums.py
kaushikb11: We tell them to use the value instead directly. Tbh, I think it makes sense

Issue #9939, opened 2021-10-15
Fail the test when a `DeprecationWarning` is raised


carmocca:
## Proposed refactoring or deprecation

Set our test suite to fail when a `DeprecationWarning` is raised.

### Motivation

Since we don't track warnings raised by our tests, it's too easy to introduce a deprecation path but forget to update all deprecated usages inside our codebase and tests.

This is very bad in terms of UX because the users will see this warning appear but won't be able to do anything to fix it.

This has been exacerbated lately with the number of deprecations happening.

For example consider this simple training script:

<details>

```python
import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer, LightningDataModule


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def validation_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("valid_loss", loss)

    def test_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("test_loss", loss)

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)


class BoringDataModule(LightningDataModule):
    def train_dataloader(self):
        return DataLoader(RandomDataset(32, 64), batch_size=2)

    def val_dataloader(self):
        return DataLoader(RandomDataset(32, 64), batch_size=2)

    def test_dataloader(self):
        return DataLoader(RandomDataset(32, 64), batch_size=2)

    def predict_dataloader(self):
        return DataLoader(RandomDataset(32, 64), batch_size=2)


def run():
    model = BoringModel()
    datamodule = BoringDataModule()
    trainer = Trainer(
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        limit_predict_batches=1,
        max_epochs=1,
    )
    trainer.fit(model, datamodule)
    trainer.test(verbose=False)
    trainer.predict()


if __name__ == "__main__":
    run()
````

</details>

[...]

PR #10520, opened 2021-11-13
Avoid deprecated `progress_bar_refresh_rate` usage

carmocca:
## What does this PR do?

Part of #9939

None

## Before submitting

## PR review

cc @awaelchli

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,7 +134,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Squeeze the early stopping monitor to remove empty tensor dimensions ([#10461](https://github.com/PyTorchLightning/pytorch-lightning/issues/10461))
 
 
--
+- Fixed issue where `trainer.progress_bar_callback` was not getting set with `enable_progress_bar=False` if a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
 
 
 ## [1.5.1] - 2021-11-09
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Squeeze the early stopping monitor to remove empty tensor dimensions ([#10461](https://github.com/PyTorchLightning/pytorch-lightning/issues/10461))
 
 
+-
+
 ## [1.5.1] - 2021-11-09
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `monitor` argument in the `EarlyStopping` callback is no longer optional ([#10328](https://github.com/PyTorchLightning/pytorch-lightning/pull/10328))
 
 
--
+- Raise `MisconfigurationException` with `enable_progress_bar=False` if a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
 
 
 -
@@ -132,9 +132,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
 - Squeeze the early stopping monitor to remove empty tensor dimensions ([#10461](https://github.com/PyTorchLightning/pytorch-lightning/issues/10461))
-
-
-- Fixed issue where `trainer.progress_bar_callback` was not getting set with `enable_progress_bar=False` if a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
 
 
 ## [1.5.1] - 2021-11-09
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `monitor` argument in the `EarlyStopping` callback is no longer optional ([#10328](https://github.com/PyTorchLightning/pytorch-lightning/pull/10328))
 
 
-- Raise `MisconfigurationException` with `enable_progress_bar=False` if a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
+- Raise `MisconfigurationException` when `enable_progress_bar=False` and a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
 
 
 -
--- a/pytorch_lightning/trainer/connectors/callback_connector.py
+++ b/pytorch_lightning/trainer/connectors/callback_connector.py
@@ -94,12 +94,9 @@ class CallbackConnector:
                 " bar pass `enable_progress_bar = False` to the Trainer."
             )
 
-        if enable_progress_bar:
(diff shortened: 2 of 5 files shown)


tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -32,54 +32,40 @@
...
-        ([], None),
-        ([], 1),
-        ([], 2),
-        ([TQDMProgressBar(refresh_rate=1)], 0),
-        ([TQDMProgressBar(refresh_rate=2)], 0),
-        ([TQDMProgressBar(refresh_rate=2)], 1),
+        # won't print but is still set
+        {"callbacks": TQDMProgressBar(refresh_rate=0)},
+        {"callbacks": TQDMProgressBar()},
+        # callbacks takes precedence
```
carmocca, line 40: I'm assuming that passing the instance as a callback should take precedence over `enable_progress_bar`

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -276,20 +263,21 @@ def test_tqdm_progress_bar_default_value(tmpdir):
     trainer = Trainer(default_root_dir=tmpdir)
     assert trainer.progress_bar_callback.refresh_rate == 1
 
-    trainer = Trainer(default_root_dir=tmpdir, progress_bar_refresh_rate=None)
```
carmocca, line 279: This is exactly the same as in the assertion above

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -276,20 +263,21 @@ def test_tqdm_progress_bar_default_value(tmpdir):
...
 @mock.patch.dict(os.environ, {"COLAB_GPU": "1"})
 def test_tqdm_progress_bar_value_on_colab(tmpdir):
     """Test that Trainer will override the default in Google COLAB."""
     trainer = Trainer(default_root_dir=tmpdir)
     assert trainer.progress_bar_callback.refresh_rate == 20
 
-    trainer = Trainer(default_root_dir=tmpdir, progress_bar_refresh_rate=None)
-    assert trainer.progress_bar_callback.refresh_rate == 20
+    trainer = Trainer(default_root_dir=tmpdir, callbacks=TQDMProgressBar())
+    assert trainer.progress_bar_callback.refresh_rate == 1  # FIXME: should be 20
```
carmocca, line 274: Is this a bug? Can be done in a separate PR.

tchaton: Yes and no :) As you provide manual the progress bar, the default value is set to 1. However, it would be better to set it to 20 in the COLAB env.

I believe a way to resolve this would be for this logic to live within the Callback directly as follows:

```py
import os
from typing import Optional

class TQDMProgressBar(ProgressBarBase):

    def __init__(self, refresh_rate: Optional[int] = None, process_position: int = 0):
        super().__init__()

        if refresh_rate is None:
            refresh_rate = 20 if os.getenv("COLAB_GPU", "1") == "1" else 1

        self._refresh_rate = refresh_rate
```

carmocca: Yes, exactly.

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -94,12 +94,9 @@ def on_trainer_init(
                 " bar pass `enable_progress_bar = False` to the Trainer."
             )
 
-        if enable_progress_bar:
-            self.trainer._progress_bar_callback = self.configure_progress_bar(
-                progress_bar_refresh_rate, process_position
-            )
-        else:
-            self.trainer._progress_bar_callback = None
+        self.trainer._progress_bar_callback = self.configure_progress_bar(
```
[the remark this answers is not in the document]
carmocca: This is a fix so i'd rather not change the API here.

tests/callbacks/test_tqdm_progress_bar.py
kaushikb11: Added it here #10542. And tested that `RichProgressBar` wouldn't require this setup

kaushikb11 (approved):

tests/callbacks/test_tqdm_progress_bar.py
carmocca: Thanks @kaushikb11, let's get this merged so you can fix it on top

tchaton (approved):
LGTM !

daniellepintz:
Hmm, are we sure this is the desired behavior here? If a user explicitly overrides the default and sets `enable_progress_bar=False` it is weird that we still enable a progress bar.

pytorch_lightning/trainer/connectors/callback_connector.py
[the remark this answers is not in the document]
ananthsub: 1. The model checkpoint callback raises a misconfiguration error if the flags are inconsistent (`enable_checkpointing=False` yet `callbacks` contains at least one `ModelCheckpoint` callback instance). IMO we should have the same behavior for progress bars for consistency.
2. (for a separate PR) I'm not sure why `progress_bar_callback` has a separate property on the Trainer. I'd prefer defining a property like this which is derived from the `callbacks` property: https://github.com/PyTorchLightning/pytorch-lightning/blob/40945a51765e72400dd3be8f1a84a52a69effa73/pytorch_lightning/trainer/trainer.py#L1841-L1846

carmocca:
> Hmm, are we sure this is the desired behavior here? If a user explicitly overrides the default and sets enable_progress_bar=False it is weird that we still enable a progress bar.

The user sends a mixed signal, both passing it as a callback and disabling it, we can go for either of them.

daniellepintz:
I agree with @ananthsub, I think we should just raise a MisConfigException

pytorch_lightning/trainer/connectors/callback_connector.py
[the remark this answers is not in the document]
carmocca: I'll update the behaviour on conflict.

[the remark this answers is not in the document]
carmocca: @ananthsub feel free to create an issue for (2)

CHANGELOG.md
```diff
@@ -134,7 +134,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Squeeze the early stopping monitor to remove empty tensor dimensions ([#10461](https://github.com/PyTorchLightning/pytorch-lightning/issues/10461))
 
 
--
+- Fixed issue where `trainer.progress_bar_callback` was not getting set with `enable_progress_bar=False` if a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
```
daniellepintz, line 137: update this to added misconfig check when `enable_progress_bar=False` and pbar instance has been passed?

daniellepintz (approved):
Looks good!

CHANGELOG.md
```diff
@@ -31,7 +31,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `monitor` argument in the `EarlyStopping` callback is no longer optional ([#10328](https://github.com/PyTorchLightning/pytorch-lightning/pull/10328))
 
 
--
+- Raise `MisconfigurationException` with `enable_progress_bar=False` if a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
```
daniellepintz, line 34:
```suggestion
- Raise `MisconfigurationException` when `enable_progress_bar=False` and a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
```

PR #10541, opened 2021-11-15
Raise an error if batch_size cannot be inferred from current batch

### LINKED ISSUE #10458: Fix `extract_batch_size` for string values and custom batches

String values and custom batches are not handled correctly while extracting batch sizes.

### Motivation and Pitch

**For string values:**
Currently when we try to infer the batch_size to be used for accumulating the metrics, for string values we do this:
https://github.com/PyTorchLightning/pytorch-lightning/blob/aad86423f731d1cbc175291825da161c71123006/pytorch_lightning/utilities/data.py#L33-L34
when torch dataset returns a string, the dataloader just convert it to a list of strings and since we iterate over the list and once reach the point where we have a string we just return its len, which is not correct. More discussions here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10349

Solution 1 (I prefer):
return None and raise the warning if multiple batch sizes are found.

Solution 2:
check all the values of the iterable and see if all are strings, if they are return the len of the list. This one might increase the time complexity a little bit.

---

**For custom batches:**
we return `1`
https://github.com/PyTorchLightning/pytorch-lightning/blob/aad86423f731d1cbc175291825da161c71123006/pytorch_lightning/utilities/data.py#L41-L42
which is super silent and highly inaccurate IMO. For this one too we should return None and raise the appropriate warning.

### Additional context

As always open to suggestion/discussion/thoughts and approvals :)

rohitgr7:
## What does this PR do?

Fixes #10458

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Raise exception in `init_dist_connection()` when torch distibuted is not available ([#10418](https://github.com/PyTorchLightning/pytorch-lightning/issues/10418))
+- Raised exception in `init_dist_connection()` when torch distibuted is not available ([#10418](https://github.com/PyTorchLightning/pytorch-lightning/issues/10418))
 
 
 - The `monitor` argument in the `EarlyStopping` callback is no longer optional ([#10328](https://github.com/PyTorchLightning/pytorch-lightning/pull/10328))
@@ -34,7 +34,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Do not fail if batch size could not be inferred for logging when using DeepSpeed ([#10438](https://github.com/PyTorchLightning/pytorch-lightning/issues/10438))
 
 
-- Raise `MisconfigurationException` when `enable_progress_bar=False` and a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
+- Raised `MisconfigurationException` when `enable_progress_bar=False` and a progress bar instance has been passed in the callback list ([#10520](https://github.com/PyTorchLightning/pytorch-lightning/issues/10520))
 
 
 - Changes in `LightningCLI` required for the new major release of jsonargparse v4.0.0 ([#10426](https://github.com/PyTorchLightning/pytorch-lightning/pull/10426))
@@ -44,6 +44,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
 - Moved ownership of the `PrecisionPlugin` into `TrainingTypePlugin` and updated all references ([#10570](https://github.com/PyTorchLightning/pytorch-lightning/pull/10570))
+
+
+- Raised an error if the `batch_size` cannot be inferred from current batch incase if its a string or some custom batch ([#10541](https://github.com/PyTorchLightning/pytorch-lightning/pull/10541))
+
+
+-
 
 
 -
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Moved ownership of the `PrecisionPlugin` into `TrainingTypePlugin` and updated all references ([#10570](https://github.com/PyTorchLightning/pytorch-lightning/pull/10570))
 
 
-- Raised an error if the `batch_size` cannot be inferred from current batch incase if its a string or some custom batch ([#10541](https://github.com/PyTorchLightning/pytorch-lightning/pull/10541))
+- Raised an error if the `batch_size` cannot be inferred from the current batch if it contained a string or was a custom batch object ([#10541](https://github.com/PyTorchLightning/pytorch-lightning/pull/10541))
 
 
 -
--- a/pytorch_lightning/utilities/data.py
+++ b/pytorch_lightning/utilities/data.py
@@ -60,8 +60,8 @@ def extract_batch_size(batch: BType) -> int:
         "We could not infer the batch_size from the batch. Either simplify its structure"
         " or provide the batch_size as `self.log(..., batch_size=batch_size)`."
     )
+    batch_size = None
     try:
-        batch_size = None
         for bs in _extract_batch_size(batch):
             if batch_size is None:
                 batch_size = bs
--- a/pytorch_lightning/utilities/data.py
+++ b/pytorch_lightning/utilities/data.py
@@ -58,9 +58,6 @@ def extract_batch_size(batch: BType) -> int:
     """
     batch_size = None
     for bs in _extract_batch_size(batch):
-        if bs is None:
-            continue
-
         if batch_size is None:
             batch_size = bs
         elif batch_size != bs:
--- a/pytorch_lightning/utilities/data.py
+++ b/pytorch_lightning/utilities/data.py
@@ -57,9 +57,8 @@ def extract_batch_size(batch: BType) -> int:
         ``len(tensor)`` when found, or ``1`` when it hits an empty or non iterable.
     """
     error_msg = (
-        "Trying to infer `batch_size` from an ambiguous collection but couldn't find one. Either"
-        " set `self.log(..., batch_size=batch_size)` or simplify the batch collection passed"
-        " inside `*_step` functions."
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -46,6 +46,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Moved ownership of the `PrecisionPlugin` into `TrainingTypePlugin` and updated all references ([#10570](https://github.com/PyTorchLightning/pytorch-lightning/pull/10570))
 
 
+- Raised an error if the `batch_size` cannot be inferred from current batch incase if its a string or some custom batch ([#10541](https://github.com/PyTorchLightning/pytorch-lightning/pull/10541))
```
tchaton, line 49:
```suggestion
- Raised an error if the `batch_size` cannot be inferred from the current batch if it contained a string or was a custom batch object ([#10541](https://github.com/PyTorchLightning/pytorch-lightning/pull/10541))
```

pytorch_lightning/utilities/data.py
```diff
@@ -69,6 +67,13 @@ def extract_batch_size(batch: BType) -> int:
             )
             break
 
+    if batch_size is None:
+        raise MisconfigurationException(
+            "Trying to infer `batch_size` from an ambiguous collection but couldn't find one. Either"
+            " set `self.log(..., batch_size=batch_size)` or simplify the batch collection passed"
+            " inside `*_step` functions."
+        )
```
tchaton, line 75:
```suggestion
        raise MisconfigurationException(
            "Trying to infer `batch_size` from an ambiguous collection but couldn't find one. Either"
            " set `self.log(..., batch_size=batch_size)`.
        )
```

tchaton: Let's not introduce a new API here. It should be an int.

tests/utilities/test_data.py
```diff
@@ -55,6 +58,20 @@ def _check_warning_raised(data, expected):
...
     _check_warning_raised(batch, 10)
 
+    # Error raised
+    batch = "test string"
+    _check_error_raised(batch)
+
+    data = {"test": ["some text"] * 7}
+    _check_error_raised(data)
+
+    class CustomBatchCollector:
```
tchaton, line 68: What is this CustomBatchCollector has a len ?

pytorch_lightning/utilities/data.py
justusschock: We could use `-1` for that if you want to stick with ints? Although I think `None` to be more intuitive

justusschock (approved):

tests/utilities/test_data.py
```diff
@@ -55,6 +58,20 @@ def _check_warning_raised(data, expected):
...
+        def __init__(self):
+            self.x = torch.randn(7, 2)
+
+    data = CustomBatchCollector()
```
rohitgr7, line 72:
```suggestion
    class CustomBatch:
        def __init__(self):
            self.x = torch.randn(7, 2)

    data = CustomBatch()
```

tests/utilities/test_data.py
rohitgr7: It's just a custom batch class.I'll update the name

tchaton: I meant, a `CustomBatch` object could have a `__len__` defined as the `batch_size`. Should we consider this too ?

pytorch_lightning/utilities/data.py
tchaton: Hey @rohitgr7,

As you said, this would raise an Exception. I just wanted to share that I don't understand what you meant to say here

>  Or simplify the batch collection passed " inside `*_step` functions.

To me, if the batch_size can't be inferred properly, then the user should pass the `batch_size`, not the batch.

tchaton (approved):
LGTM ! Waiting for open questions.

pytorch_lightning/utilities/data.py
rohitgr7: there are two options. Either they can pass the batch_size to self.log(), or they can simplify the batch being passed to step functions. I have added both suggestions in the error msg.

tests/utilities/test_data.py
rohitgr7: not sure if we should do this. I haven't seen any case where user have a custom batch structure and have defined `__len__`. Plus we have `__len__` inside PT dataset so it might confuse users more.

pytorch_lightning/utilities/data.py
```diff
@@ -58,18 +56,29 @@ def extract_batch_size(batch: BType) -> int:
     Returns:
         ``len(tensor)`` when found, or ``1`` when it hits an empty or non iterable.
     """
-    batch_size = None
-    for bs in _extract_batch_size(batch):
+    error_msg = (
+        "Trying to infer `batch_size` from an ambiguous collection but couldn't find one. Either"
+        " set `self.log(..., batch_size=batch_size)` or simplify the batch collection passed"
+        " inside `*_step` functions."
```
carmocca, line 62: When would the user "simplify the batch collection passed..."?

If the user has that collection structure, it's probably for a reason

pytorch_lightning/utilities/data.py
```diff
@@ -58,18 +56,29 @@ def extract_batch_size(batch: BType) -> int:
...
+    )
+    try:
```
carmocca, line 64: Minimize code under the try/except

```python
batch_size = None
try:
    for bs in _extract_batch_size(batch):
        if batch_size is None:
            batch_size = bs
        elif batch_size != bs:
            warning_cache.warn(
                "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
                f" found is {batch_size}. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`."
            )
            break
except RecursionError:
    raise RecursionError(error_msg)

if batch_size is None:
    raise MisconfigurationException(error_msg)

return batch_size
```

pytorch_lightning/utilities/data.py
tchaton: Oh I understand what you meant now.

> `simplify the batch collection passed` is quite confusing.
Maybe something like. Lightning couldn't infer the `batch_size` from your batch. Either simplify its structure or provide the batch_size as self.log(..., batch_size=batch_size) `

tests/utilities/test_data.py
tchaton: I have seen several actually.

pytorch_lightning/utilities/data.py
rohitgr7: better! updated

pytorch_lightning/utilities/data.py
rohitgr7: done!

pytorch_lightning/utilities/data.py
rohitgr7: sometimes they are not aware of how the collate_fn works so I guess it's good to have.

pytorch_lightning/utilities/data.py
```diff
@@ -58,16 +56,26 @@ def extract_batch_size(batch: BType) -> int:
...
-                "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"
-                f" found is {batch_size}. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`."
-            )
-            break
+    error_msg = (
+        "We could not infer the batch_size from the batch. Either simplify its structure"
+        " or provide the batch_size as `self.log(..., batch_size=batch_size)`."
+    )
+    try:
+        batch_size = None
```
carmocca, line 64:
```suggestion
    batch_size = None
    try:
```

carmocca (approved):

ninginthecloud:
Hi, @rohitgr7 this change breaks user's normal training flow. As auto inferencing of batch_size is very fragile, this misconfigurationexception will break down the whole training. Why not just keeping the default value?

rohitgr7:
@ninginthecloud We have a strict rule too to only auto inference the batch_size when it's required for logging. Keeping the default value of 1 might not end up generating correct aggregated metrics.

ninginthecloud:
> @ninginthecloud We have a strict rule too to only auto inference the batch_size when it's required for logging. Keeping the default value of 1 might not end up generating correct aggregated metrics.

Thanks for quick reply~ Your statement emphasized the importance of correctly inferencing batch_size. if inferenced batch_size is wrong, it also ends up generating wrong aggregated metrics. Therefore, we need gracefully address the limitation of auto-inferencing for batch_size. If `extract_batch_size` can't extract valid or correct value, let user know the limitation instead of throwing out exception.  I will forward the discussion to the open issue.

PR #10548, opened 2021-11-15
Introduce `Strategy` in favour of `TrainingTypePlugin`

kaushikb11:
## What does this PR do?

Part of #9932

As this PR is already huge enough, breaking the task into these follow up PRs:

* Deprecate TrainingTypePlugin (as it has a lot of references, seperate PR)
* Deprecate `_training_type_plugin` in favour of `_strategy_plugin`
* Add `strategy_plugin` argument to `Accelerator`
* Update Training Type Registry

## Before submitting

## PR review

## Did you have fun?


ananthsub:
I'd drop `Plugin` from the name of the class:
1. Matches conventions with the other "pluggable" arguments Lightning offers: none of Accelerator, Callback, Profiler, Logger have the plugin suffix.
2. So people know they should specify this via the `strategy` argument on the Trainer and not the `plugins` argument.

kaushikb11:
> So people know they should specify this via the strategy argument on the Trainer and not the plugins argument.

Do you think it should be part of `plugins` directory anymore?

four4fish:
> I'd drop `Plugin` from the name of the class:
> 
> 1. Matches conventions with the other "pluggable" arguments Lightning offers: none of Accelerator, Callback, Profiler, Logger have the plugin suffix.
> 2. So people know they should specify this via the `strategy` argument on the Trainer and not the `plugins` argument.
> 3. to match convention again, the base class + implementations should be under `pytorch_lightning/strategy`
> 
> and over time, these classes could be renamed like `<blank>Strategy` the same way we have `SimpleProfiler` and `TensorboardLogger`
> 
> @four4fish what do you think?

Yeah, agree, I think we should drop plugin this word completely. Also remove the `plugins` flag from trainer

pytorch_lightning/plugins/training_type/ddp2.py
```diff
@@ -11,65 +11,15 @@
...
+
+
+class DDP2Plugin(DDP2Plugin):
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        rank_zero_deprecation(
+            "`pytorch_lightning.plugins.training_type.ddp2.DDP2Plugin` has been deprecated in v1.6"
+            " and will be removed in v1.8. Use `pytorch_lightning.plugins.strategy.ddp2.DDP2Plugin`"
+            " instead."
+        )
```
carmocca, line 25: Since you are making these changes to each file, git does not properly move the file and interprets it as a deletion + addition.

I strongly suggest we do this in two PRs:

1. Move the code.
2. Add the backwards compatible paths.

This is to preserve the git history.

In between, the backwards compatibility would be broken but should be okay as there will not be a release in-between.

Since the Meta team relies on master, we just need them to skip the commits between step (1) until step (2) is merged. cc @ananthsub

four4fish: And move files lost the historical tracks, it's hard to tracking back changes and old issues related to it. Is this nessesary?

carmocca: `git mv` supports moving the location and changing the filename.

The problem is then adding changes to the file.

four4fish: > `git mv` supports moving the location and changing the filename.
> 
> The problem is then adding changes to the file.

Got you! that's why you suggested separate the steps, make sense! thanks~~

Issue #10107, opened 2021-10-24
RFC: Validate Trainer settings against cluster environment settings


awaelchli:
## 🚀 Feature

Introduce a validation check that verifies the Trainer settings like gpus, and number of nodes are set compatible with the cluster environment currently used. 

### Motivation

There have been several issues where users did not set the number of devices or the number of nodes correctly on clusters (#10098,  #8993, #4612). The result: Processes hang for unknown reason. The user is confused, does not know what to do. It's also hard to debug for us because often the user just reports "DDP hangs, help!", and there could be a million other reasons why ddp could stall. 

### Pitch

Introduce a `ClusterEnvironment.validate_distributed_settings` method (name up for negotiation):

```python
class ClusterEnvironment:

    def validate_distributed_settings(num_devices, num_nodes, ...):
        # implementation may vary depending on environment
        if self.world_size() != num_devices * num_nodes:
            # raise warning, e.g., user should check SLURM settings for num nodes, num tasks per node, etc.

```

This would be an optional method on the ClusterEnvironment. The argument list could include also other arguments.

In the accelerator connector, when setting up the environments, we would call this method. Roughly here:
https://github.com/PyTorchLightning/pytorch-lightning/blob/c9bc10ce8473a2249ffa4e00972c0c3c1d2641c4/pytorch_lightning/trainer/connectors/accelerator_connector.py#L794-L807
### Alternatives

Alternative suggestions welcome.

### Additional context

This would resolve #10098,  #8993, #4612, and perhaps others.

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

</sub>

Issue #6303, opened 2021-03-02
ClusterEnvironment interface and improvements


awaelchli:
## 🚀 Feature

In #5915 we started discussing what a good interface for the cluster environment could be and where it is headed in the future.
We should iterate on these ideas.

Some ideas / issues / refactors to discuss:

- [x] there is a small amount of code duplication.
- [x] Review #5102 and determine if interface needs changes
- [x] Add docs for all environments
- [x] Teardown logic: cluster environments that set environment variables (globally) should also be responsible for cleaning them up (#6942)
- [ ] Convert methods to proper setters/getters (#10103)
- [x] Consider renaming "creates_children". It is not clear if it means the environment within Lightning or the actual cluster. cc @borda @justusschock @awaelchli @akihironitta @rohitgr7 @tchaton @ananthsub #10339 
- [x] Move left over slurm logic in slurm connector to the slurm environment (#10622 #10615)

- [x] A `detect` method:

```python
# returns true if we can detect the cluster environment.
def detect():
    if ...
        return True
```

leezu:
Trainer has a SLURM specific property. Should it be deprecated as part of the ClusterEnvironment interface improvements?

https://github.com/PyTorchLightning/pytorch-lightning/blob/8c0ea92af237542f5b36ae684543f871da829379/pytorch_lightning/trainer/properties.py#L176-L189

awaelchli:
Hi, good observation. I would say the logic inside of it can and should definitely move into the cluster plugin. I would still keep the property as an easy way to access the cluster environment IF that is something users actually access (I have no idea).

ananthsub:
> Consider renaming "creates_children". It is not clear if it means the environment within Lightning or the actual cluster. cc @ananthsub

Since it came up in #9389 - we need a consistent value here during the course of the job. relying on the mutable os.environ for instance can lead to surprising side effects. a naming nit would be to to rename this as `created_children` to be ultra clear that this either happened or didn't, as opposed to a continuous/ongoing behavior.

carmocca:
@awaelchli is this done? can we close?

awaelchli:
@carmocca there is one item open

> Convert methods to proper setters/getters

[...]

PR #10564, opened 2021-11-16
Introduce `ClusterEnvironment.detect()`

awaelchli:
## What does this PR do?

Part of #6303 
Also addresses https://github.com/PyTorchLightning/pytorch-lightning/issues/10107#issuecomment-950848971

Introduces an interface for detecting the cluster environment. Previously, each enironment had a custom method:

```py
TorchElasticEnvironment.is_using_torch_elastic
LSFEnvironment.is_using_lsf
KubeflowEnvironment.is_using_kubeflow
```

Now these implementations will be under the static method `detect()` inherited from the base environment.

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @akihironitta @tchaton @kaushikb11


pytorch_lightning/plugins/environments/cluster_environment.py
```diff
@@ -40,6 +40,12 @@ def main_address(self) -> str:
     def main_port(self) -> int:
         """An open and configured port in the main node through which all processes communicate."""
 
+    @staticmethod
+    @abstractmethod
+    def detect() -> bool:
+        """Detects the environment settings corresponding to this cluster and returns ``True`` if they match."""
+        pass
```
carmocca, line 47: nit: not necessary with the docstring

```suggestion
```

pytorch_lightning/plugins/environments/kubeflow_environment.py
```diff
@@ -47,6 +50,14 @@ def main_address(self) -> str:
     def main_port(self) -> int:
         return int(os.environ["MASTER_PORT"])
 
+    @staticmethod
+    def detect() -> bool:
+        """Returns ``True`` if the current process was launched using Kubeflow PyTorchJob."""
+        required_env_vars = ("KUBERNETES_PORT", "MASTER_ADDR", "MASTER_PORT", "WORLD_SIZE", "RANK")
+        # torchelastic sets these. Make sure we're not in torchelastic
+        excluded_env_vars = ("GROUP_RANK", "LOCAL_RANK", "LOCAL_WORLD_SIZE")
+        return all(v in os.environ for v in required_env_vars) and not any(v in os.environ for v in excluded_env_vars)
```
carmocca, line 59: we can be cool here with sets

```suggestion
        required_env_vars = {"KUBERNETES_PORT", "MASTER_ADDR", "MASTER_PORT", "WORLD_SIZE", "RANK"}
        # torchelastic sets these. Make sure we're not in torchelastic
        excluded_env_vars = {"GROUP_RANK", "LOCAL_RANK", "LOCAL_WORLD_SIZE"}
        env_vars = os.environ.keys()
        return env_vars == required_env_vars and env_vars.isdisjoint(excluded_env_vars)
```

pytorch_lightning/plugins/environments/lsf_environment.py
```diff
@@ -66,6 +68,12 @@ def main_port(self) -> int:
         """The main port gets calculated from the LSF job ID."""
         return self._main_port
 
+    @staticmethod
+    def detect() -> bool:
+        """Returns ``True`` if the current process was launched using the jsrun command."""
+        required_env_vars = ("LSB_JOBID", "LSB_HOSTS", "JSM_NAMESPACE_LOCAL_RANK", "JSM_NAMESPACE_SIZE")
```
carmocca, line 74: Same set proposal

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -806,17 +806,14 @@ def select_cluster_environment(self) -> ClusterEnvironment:
...
-        elif LSFEnvironment.is_using_lsf():
-            env = LSFEnvironment()
-        else:
-            env = LightningEnvironment()
-        return env
+            return SLURMEnvironment()
+
+        for env_type in (TorchElasticEnvironment, KubeflowEnvironment, LSFEnvironment, LightningEnvironment):
+            if env_type.detect():
+                return env_type()
```
carmocca, line 814: cool!

carmocca (approved):

pytorch_lightning/plugins/environments/kubeflow_environment.py
awaelchli: that seems all good, but should we really do it now? I meant to only move things around here and not change at the same time.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -806,17 +806,14 @@ def select_cluster_environment(self) -> ClusterEnvironment:
...
-        elif KubeflowEnvironment.is_using_kubeflow():
-            env = KubeflowEnvironment()
-        elif LSFEnvironment.is_using_lsf():
-            env = LSFEnvironment()
-        else:
-            env = LightningEnvironment()
-        return env
+            return SLURMEnvironment()
+
+        for env_type in (TorchElasticEnvironment, KubeflowEnvironment, LSFEnvironment, LightningEnvironment):
```
ananthsub, line 812: 1. I don't think we need `LightningEnvironment` in this list. its inclusion will mean we never execute L816 because we will always return True because of `LightningEnvironment.detect()`'s implementation
2. Should we keep this list of non-LightningEnvironmen cluster environments as a constant somewhere? then if support for other clusters are added, we can update the registry once for wherever it's used
3. does the ordering of the list matter?

pytorch_lightning/plugins/environments/kubeflow_environment.py
carmocca: why not? tests should catch any issues with the change (assuming we have tests for this)

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: 1) agree, this can be done :) 
2) that has crossed my mind too. SLURM will be an outlier though. It's detection is slightly different and guarded through a method _is_slurm_managing_tasks in the accelerator connector.
3) Yes, TorchElasticEnv has priority over the Lightning because they share some env variable names. But we can take it out and make sure it comes last anyway, so the rest should not matter in ordering.

pytorch_lightning/plugins/environments/cluster_environment.py
```diff
@@ -40,6 +40,11 @@ def main_address(self) -> str:
     def main_port(self) -> int:
         """An open and configured port in the main node through which all processes communicate."""
 
+    @staticmethod
+    @abstractmethod
+    def detect() -> bool:
+        """Detects the environment settings corresponding to this cluster and returns ``True`` if they match."""
```
ananthsub, line 46: total n00b question: any idea how to make adding a new abstractmethod not a breaking change? i imagine this will repeatedly come up

ananthsub (approved):

pytorch_lightning/plugins/environments/cluster_environment.py
awaelchli: one could avoid making it abstract now and remember to do it once the deprecation ends. not sure if there is a better way

kaushikb11:
Let's introduce a generic `environment` label instead

pytorch_lightning/plugins/environments/kubeflow_environment.py
awaelchli: it's not about correctness of the implementation.
I don't feel comfortable moving code and changing it at the same time because of the review process.

pytorch_lightning/plugins/environments/cluster_environment.py
tchaton: We could have a soft detect for custom cluster env. Return an Optional[bool] and have the default implementation passing. We could detect it hasn't been implemented and add a warning this function would be moved to abstract later on.

pytorch_lightning/plugins/environments/kubeflow_environment.py
tchaton: Let's make this change in another PR then.

tchaton (approved):
LGTM !

Issue #10416, opened 2021-11-08
[Main Issue] Accelerator and Plugin refactor


four4fish:
### Motivation
Accelerator is not stable API yet, we can improve the Accelerator related logic and move towards stable Accelerator version for 1.6

### Pitch
**Steps**
1. Collective refactor #7534
   - [x] deprecate accelerator collective directly call from TTP #9677
   - [ ] collective refactor #9414
2. Move Precision Plugin into TTP #7324 
3. Move Accelerator into Strategy #10648 
4. Simplify the Spawning logic #10059 
5. [RFC] Simplifying the Accelerator Connector logic and flags (can be done in parallel with aboves) #11449  11449
6. [RFC] Revisit the inheritance of TTP #11863
 
 
 

**More details in**: [Accelerator Refactor Proposal](https://docs.google.com/document/d/1E5t8auWf5DrNHzutvMmrJC_KqBqVyZuYX0thra69Ad8/edit#heading=h.nehmb825f40s)
[updating] 

### FAQ
1. Will this be a lot of breaking changes?
Not much user facing API changes from 1,2,3,4.(Unless we found out other existing bugs during refactor) The only breaking change will be for custom plugins
5 and 6 is still RFC stage, may have breaking changes which impact user facing APIs

2. How does this impact lightningLite?
Should be helpful for lightningLite too, there maybe function refactor/simplification could happen for lightningLite. (@awaelchli  any suggestion about this part?)

### Follow up TODOs:
- [ ] check if trainer.FITTING before setting up optimizers https://github.com/PyTorchLightning/pytorch-lightning/pull/10596#discussion_r753587321

- [ ] Send a PR to  https://github.com/ray-project/ray_lightning  to update their plugins

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

[...]

PR #9940, opened 2021-10-15
Fail the test when a `DeprecationWarning` is raised

carmocca:
## What does this PR do?

Fixes #9939

### Does your PR introduce any breaking changes? If yes, please list them.

None

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [n/a] Did you make sure to **update the documentation** with your changes? (if necessary)
- [n/a] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [n/a] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

cc @borda @tchaton @carmocca @akihironitta

PR #10570, opened 2021-11-16
1/n Move precision plugin into strategy - update reference

### LINKED ISSUE #7324: Precision Plugins should be part of Training Type Plugins

four4fish:
## What does this PR do?

**1/n Move Precision Plugin into Strategy and update references**
2/n Move optimizer, lr to Strategy, remove accelerator routing functions 
3/n Simplify and reduce hooks and redundancies

Fixes #7324 which is part of #10416

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -45,9 +45,7 @@ class Accelerator:
     One to handle differences from the training routine and one to handle different precisions.
     """
 
-    def __init__(
-        self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None
-    ) -> None:
+    def __init__(self, precision_plugin: Optional[PrecisionPlugin], training_type_plugin: TrainingTypePlugin) -> None:
         """
         Args:
             training_type_plugin: the plugin to handle different training routines
@@ -60,12 +58,9 @@ class Accelerator:
 
         self.training_type_plugin = training_type_plugin
 
-        if precision_plugin:
-            """
-            .. deprecated
-                precision_plugin parameter is deprecated will be removed soon.
-                Use :`training_type_plugin(precision_plugin) instead.
-            """
+        if precision_plugin is not None:
+            """precision_plugin is deprecated and will be removed soon User
+            `training_type_plugin(precision_plugin)` instead."""
             self.training_type_plugin._precision_plugin = precision_plugin
 
         self.optimizers: List = []
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -276,11 +276,10 @@ class Accelerator:
     def precision(self) -> Union[str, int]:
         """The type of precision being used with this accelerator.
 
-        Use `training_type_plugin.precision_plugin.precision` instead.
-
         .. deprecated::
-            The ``precision_plugin`` parameter has been deprecated and will be removed soon.
-            Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
+            This property been deprecated and will be removed soon.
+            Use ``training_type_plugin.precision_plugin.precision`` instead.
+
         """
         rank_zero_deprecation(
             f"`{self.__class__.__name__}.precision` has been deprecated and will be removed soon"
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -48,18 +48,17 @@ class Accelerator:
     def __init__(self, precision_plugin: Optional[PrecisionPlugin], training_type_plugin: TrainingTypePlugin) -> None:
         """
         Args:
-            training_type_plugin: the plugin to handle different training routines
             precision_plugin: the plugin to handle precision-specific parts
-
-        Notes:
-            precision_plugin is deprecated and will be removed soon
-            User `training_type_plugin(precision_plugin)` instead
+                .. deprecated::
+                    The ``precision_plugin`` parameter has been deprecated and will be removed soon.
+                    Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
+            training_type_plugin: the plugin to handle different training routines
         """
 
         self.training_type_plugin = training_type_plugin
 
         if precision_plugin is not None:
-            """precision_plugin is deprecated and will be removed soon User
+            """precision_plugin is deprecated and will be removed soon, use
             `training_type_plugin(precision_plugin)` instead."""
             self.training_type_plugin._precision_plugin = precision_plugin
 
@@ -275,12 +274,13 @@ class Accelerator:
 
     @property
     def precision(self) -> Union[str, int]:
-        """This method is deprecated and will be removed soon.
+        """The type of precision being used with this accelerator.
 
-        Use `training_type_plugin.precision_plugin.precision` instead.
+        .. deprecated::     The ``precision_plugin`` parameter has been deprecated and will be removed soon.     Pass
+        the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
         """
         rank_zero_deprecation(
-            f"`{self.__class__.__name__}.precision` was and will be removed soon"
+            f"`{self.__class__.__name__}.precision` has been deprecated and will be removed soon"
             f" Use `training_type_plugin.precision_plugin.precision` instead."
         )
         return self.training_type_plugin.precision_plugin.precision
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -44,14 +44,19 @@ class Accelerator:
     One to handle differences from the training routine and one to handle different precisions.
     """
 
-    def __init__(self, precision_plugin: PrecisionPlugin, training_type_plugin: TrainingTypePlugin) -> None:
+    def __init__(
+        self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None
+    ) -> None:
         """
         Args:
             precision_plugin: the plugin to handle precision-specific parts
             training_type_plugin: the plugin to handle different training routines
         """
-        self.precision_plugin = precision_plugin
+
         self.training_type_plugin = training_type_plugin
+
+        if precision_plugin:
+            self.training_type_plugin._precision_plugin = precision_plugin
 
         self.optimizers: List = []
         self.lr_schedulers: List = []
@@ -84,7 +89,7 @@ class Accelerator:
         if self.training_type_plugin.setup_optimizers_in_pre_dispatch:
             self.setup_optimizers(trainer)
 
-        self.precision_plugin.pre_dispatch()
+        self.training_type_plugin.precision_plugin.pre_dispatch()
 
     def _move_optimizer_state(self, device: Optional[torch.device] = None) -> None:
         """Moves the state of the optimizers to the GPU if needed."""
@@ -96,12 +101,12 @@ class Accelerator:
     def dispatch(self, trainer: "pl.Trainer") -> None:
         """Hook to do something before the training/evaluation/prediction starts."""
         self.training_type_plugin.dispatch(trainer)
-        self.precision_plugin.dispatch(trainer)
+        self.training_type_plugin.precision_plugin.dispatch(trainer)
 
     def post_dispatch(self, trainer: "pl.Trainer") -> None:
         """Hook to do something after the training/evaluation/prediction starts."""
         self.training_type_plugin.post_dispatch(trainer)
-        self.precision_plugin.post_dispatch()
+        self.training_type_plugin.precision_plugin.post_dispatch()
 
     @property
     def model(self) -> Module:
@@ -159,7 +164,7 @@ class Accelerator:
 
         See :meth:`~pytorch_lightning.core.lightning.LightningModule.training_step` for more details
         """
-        with self.precision_plugin.train_step_context():
+        with self.training_type_plugin.precision_plugin.train_step_context():
             return self.training_type_plugin.training_step(*step_kwargs.values())
 
     def validation_step(self, step_kwargs: Dict[str, Union[Any, int]]) -> Optional[STEP_OUTPUT]:
@@ -167,7 +172,7 @@ class Accelerator:
 
         See :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_step` for more details
         """
-        with self.precision_plugin.val_step_context():
+        with self.training_type_plugin.precision_plugin.val_step_context():
             return self.training_type_plugin.validation_step(*step_kwargs.values())
 
     def test_step(self, step_kwargs: Dict[str, Union[Any, int]]) -> Optional[STEP_OUTPUT]:
@@ -175,7 +180,7 @@ class Accelerator:
 
         See :meth:`~pytorch_lightning.core.lightning.LightningModule.test_step` for more details
         """
-        with self.precision_plugin.test_step_context():
+        with self.training_type_plugin.precision_plugin.test_step_context():
             return self.training_type_plugin.test_step(*step_kwargs.values())
 
     def predict_step(self, step_kwargs: Dict[str, Union[Any, int]]) -> STEP_OUTPUT:
@@ -183,7 +188,7 @@ class Accelerator:
 
         See :meth:`~pytorch_lightning.core.lightning.LightningModule.predict_step` for more details
         """
-        with self.precision_plugin.predict_step_context():
+        with self.training_type_plugin.precision_plugin.predict_step_context():
             return self.training_type_plugin.predict_step(*step_kwargs.values())
 
     def backward(self, closure_loss: Tensor, *args: Any, **kwargs: Any) -> Tensor:
@@ -193,11 +198,11 @@ class Accelerator:
             closure_loss: a tensor holding the loss value to backpropagate
         """
         self.training_type_plugin.pre_backward(closure_loss)
-        closure_loss = self.precision_plugin.pre_backward(self.lightning_module, closure_loss)
+        closure_loss = self.training_type_plugin.precision_plugin.pre_backward(self.lightning_module, closure_loss)
 
-        self.precision_plugin.backward(self.lightning_module, closure_loss, *args, **kwargs)
+        self.training_type_plugin.precision_plugin.backward(self.lightning_module, closure_loss, *args, **kwargs)
 
-        closure_loss = self.precision_plugin.post_backward(self.lightning_module, closure_loss)
+        closure_loss = self.training_type_plugin.precision_plugin.post_backward(self.lightning_module, closure_loss)
         self.training_type_plugin.post_backward(closure_loss)
 
         return closure_loss
@@ -220,7 +225,7 @@ class Accelerator:
             **kwargs: Any extra arguments to ``optimizer.step``
         """
         model = model or self.lightning_module
-        self.precision_plugin.optimizer_step(model, optimizer, opt_idx, closure, **kwargs)
+        self.training_type_plugin.precision_plugin.optimizer_step(model, optimizer, opt_idx, closure, **kwargs)
 
     def optimizer_zero_grad(self, current_epoch: int, batch_idx: int, optimizer: Optimizer, opt_idx: int) -> None:
         """Zeros all model parameter's gradients."""
@@ -248,26 +253,29 @@ class Accelerator:
 
     def setup_precision_plugin(self) -> None:
         """Attaches the precision plugin to the accelerator."""
-        model, optimizers, schedulers = self.precision_plugin.connect(self.model, self.optimizers, self.lr_schedulers)
+        model, optimizers, schedulers = self.training_type_plugin.precision_plugin.connect(
+            self.model, self.optimizers, self.lr_schedulers
+        )
         self.model = model
         self.optimizers = optimizers
         self.lr_schedulers = schedulers
 
     @property
     def amp_backend(self) -> Optional[LightningEnum]:
-        if isinstance(self.precision_plugin, ApexMixedPrecisionPlugin):
+        if isinstance(self.training_type_plugin.precision_plugin, ApexMixedPrecisionPlugin):
             return AMPType.APEX
-        if isinstance(self.precision_plugin, NativeMixedPrecisionPlugin):
+        if isinstance(self.training_type_plugin.precision_plugin, NativeMixedPrecisionPlugin):
             return AMPType.NATIVE
         return None
 
     @property
     def precision(self) -> Union[str, int]:
-        return self.precision_plugin.precision
+        """deprecated."""
+        return self.training_type_plugin.precision
 
     @property
     def scaler(self) -> Optional["GradScaler"]:
-        return getattr(self.precision_plugin, "scaler", None)
+        return getattr(self.training_type_plugin.precision_plugin, "scaler", None)
 
     def optimizer_state(self, optimizer: Optimizer) -> Dict[str, Tensor]:
         """Returns state of an optimizer.
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -49,17 +49,21 @@ class Accelerator:
         """
         Args:
             precision_plugin: the plugin to handle precision-specific parts
+
                 .. deprecated::
                     The ``precision_plugin`` parameter has been deprecated and will be removed soon.
                     Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
+
             training_type_plugin: the plugin to handle different training routines
         """
 
         self.training_type_plugin = training_type_plugin
 
         if precision_plugin is not None:
-            """precision_plugin is deprecated and will be removed soon, use
-            `training_type_plugin(precision_plugin)` instead."""
+            """precision_plugin is deprecated and will be removed soon.
+
+            Use `training_type_plugin(precision_plugin)` instead.
+            """
             self.training_type_plugin._precision_plugin = precision_plugin
 
         self.optimizers: List = []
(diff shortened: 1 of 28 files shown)


pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +44,18 @@ class Accelerator:
     One to handle differences from the training routine and one to handle different precisions.
     """
 
-    def __init__(self, precision_plugin: PrecisionPlugin, training_type_plugin: TrainingTypePlugin) -> None:
+    def __init__(self, training_type_plugin: TrainingTypePlugin, precision_plugin: PrecisionPlugin = None) -> None:
```
kaushikb11, line 47:
```suggestion
    def __init__(self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None) -> None:
```

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +44,20 @@ class Accelerator:
...
         Args:
             precision_plugin: the plugin to handle precision-specific parts
             training_type_plugin: the plugin to handle different training routines
         """
-        self.precision_plugin = precision_plugin
+
         self.training_type_plugin = training_type_plugin
 
+        if precision_plugin:
+            self.training_type_plugin._precision_plugin = precision_plugin
```
ananthsub, line 59: should this include a deprecation message for passing precision plugin this way?

ananthsub (commented):
It'd be very good to list what APIs this lets us simplify or remove from the Accelerator/strategy/precision plugin. Those could be listed as immediate followups to this PR

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -33,16 +34,27 @@ class TrainingTypePlugin(ABC):
...
     @property
     def checkpoint_io(self) -> CheckpointIO:
         return self._checkpoint_io
 
+    @property
+    def precision(self) -> Union[str, int]:
+        return self._precision_plugin.precision
+
+    @property
+    def precision_plugin(self) -> Optional[PrecisionPlugin]:
```
awaelchli:
```suggestion
    def precision_plugin(self) -> PrecisionPlugin:
```
I believe this should not be optional given the instantiation above.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -407,6 +406,7 @@ def training_type_plugin(self) -> TrainingTypePlugin:
         # attach checkpoint plugin to the training type plugin
         if self._checkpoint_io is not None:
             self._training_type_plugin.checkpoint_io = self._checkpoint_io
+        self._training_type_plugin._precision_plugin = self.precision_plugin
```
awaelchli, line 410: Perhaps we need to None check here too?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1669,7 +1669,7 @@ def amp_backend(self) -> Optional[str]:
 
     @property
     def precision(self) -> Union[str, int]:
-        return self.accelerator.precision
+        return self.training_type_plugin.precision
```
awaelchli:
```suggestion
        return self.training_type_plugin.precision_plugin.precision
```
in case we want to skip the property definition as shared in my comment above.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +44,20 @@ class Accelerator:
     One to handle differences from the training routine and one to handle different precisions.
     """
 
-    def __init__(self, precision_plugin: PrecisionPlugin, training_type_plugin: TrainingTypePlugin) -> None:
+    def __init__(
+        self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None
+    ) -> None:
         """
         Args:
             precision_plugin: the plugin to handle precision-specific parts
```
awaelchli, line 52: note to self: update

pytorch_lightning/plugins/training_type/sharded_spawn.py
```diff
@@ -118,7 +118,7 @@ def post_training_step(self):
     def new_process(self, trainer: "pl.Trainer", mp_queue: SimpleQueue) -> None:
         # Ensure that the scaler points to the correct process group
         # which is re-initialized in a new process
-        precision_plugin = trainer.accelerator.precision_plugin
+        precision_plugin = trainer.precision_plugin
         if isinstance(precision_plugin, ShardedNativeMixedPrecisionPlugin):
             precision_plugin.scaler = ShardedGradScaler()
```
awaelchli, line 123:
```suggestion
        if isinstance(self.precision_plugin, ShardedNativeMixedPrecisionPlugin):
            self.precision_plugin.scaler = ShardedGradScaler()
```

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -56,11 +57,14 @@ def __init__(
         self,
         parallel_devices: Optional[List[int]] = None,
         checkpoint_io: Optional[CheckpointIO] = None,
+        precision_plugin: Optional[PrecisionPlugin] = None,
```
awaelchli, line 60: note to self: update trainer.precision_plugin references here

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: fixes mypy

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -248,26 +253,29 @@ def setup_training_type_plugin(self) -> None:
...
             return AMPType.APEX
-        if isinstance(self.precision_plugin, NativeMixedPrecisionPlugin):
+        if isinstance(self.training_type_plugin.precision_plugin, NativeMixedPrecisionPlugin):
             return AMPType.NATIVE
         return None
 
     @property
     def precision(self) -> Union[str, int]:
-        return self.precision_plugin.precision
+        """deprecated."""
```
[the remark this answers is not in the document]
four4fish: Sorry for the poor wording, will update

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -33,16 +34,27 @@ class TrainingTypePlugin(ABC):
...
         self._checkpoint_io = checkpoint_io
+        self._precision_plugin = precision_plugin if precision_plugin is not None else PrecisionPlugin()
 
     @property
     def checkpoint_io(self) -> CheckpointIO:
         return self._checkpoint_io
 
+    @property
+    def precision(self) -> Union[str, int]:
+        return self._precision_plugin.precision
```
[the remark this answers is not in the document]
four4fish: I don't see too much value. trainer already have it as well, I think we can remove this?

pytorch_lightning/trainer/connectors/accelerator_connector.py
four4fish: If it's none, in training_type_plugin init() it will be set to Precision_plugin(), should we raise misconfig() or runtimeError if precision_plugin is None though? I feel just default Precision_plugin() could cause silent failures?

pytorch_lightning/plugins/training_type/training_type_plugin.py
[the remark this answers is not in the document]
awaelchli: I think so too :)

awaelchli (approved):

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: Actually it can never be None. In the AcceleratorConnector.select_precision_plugin we always return a plugin and have a misconfiguration when no precision is detected. So we should be fine here.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +45,29 @@ class Accelerator:
...
+        self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None
+    ) -> None:
         """
         Args:
-            precision_plugin: the plugin to handle precision-specific parts
             training_type_plugin: the plugin to handle different training routines
+            precision_plugin: the plugin to handle precision-specific parts
+
+                .. deprecated:: v1.6
+                    The ``precision_plugin`` parameter has been deprecated in v1.6 and will be removed in 1.7.
```
kaushikb11, line 57:
```suggestion
                    The ``precision_plugin`` parameter has been deprecated in v1.6 and will be removed in 1.8.
```
Should we remove it for v1.8 release instead?

justusschock (approved):

pytorch_lightning/accelerators/accelerator.py
awaelchli: I'm not sure. A few thoughts:

- I'm not even sure we can uphold this deprecation/backward compatibility if we go all the way with https://docs.google.com/document/d/1E5t8auWf5DrNHzutvMmrJC_KqBqVyZuYX0thra69Ad8/edit#heading=h.nehmb825f40s
- We want accelerator/plugins to be stable at 1.6 - so we might as well introduce larger breaking changes now 
- So far we have always chosen a shorter deprecation of only one release for all plugin/accelerator related items.

RFC @ananthsub @four4fish

four4fish: Yeah I agree, in step 3 move accelerator into TTP, there is no way we can keep backwards compatibility. Precision plugin has to be removed then. 

> We want accelerator/plugins to be stable at 1.6 - so we might as well introduce larger breaking changes now

Agree! how about "this is deprecated and will be removed soon"?

ananthsub: The precision plugin would be removed from here once we invert the Strategy/Accelerator hierarchy. @four4fish I don't think the deprecation warning would last through that refactor.

If that's the case, then we don't need the warning here. We can add the basic support in the interim for people on master, but the Accelerator shouldn't accept a precision plugin when 1.6 is released, since the accelerator won't have a reference to the strategy/training type plugin by then.

four4fish: Agree, will update.
I think as soon as the step 3: moving accelerator start, accelerator won't accept precision plugin anymore.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -33,16 +34,23 @@ class TrainingTypePlugin(ABC):
...
         checkpoint_io = checkpoint_io if checkpoint_io is not None else TorchCheckpointIO()
         self._checkpoint_io = checkpoint_io
+        self._precision_plugin = precision_plugin if precision_plugin is not None else PrecisionPlugin()
 
     @property
     def checkpoint_io(self) -> CheckpointIO:
         return self._checkpoint_io
 
+    @property
+    def precision_plugin(self) -> PrecisionPlugin:
```
kaushikb11, line 51: Should we have a setter method for it as well?

four4fish: `training_type_plugin._precision_plugin=`
There are overrides for precision_plugin(self) logic in deepspeed.

kaushikb11:
Also, we would be breaking it by changing the signature of the `Accelerator` from

```python
def __init__(self, precision_plugin: PrecisionPlugin, training_type_plugin: TrainingTypePlugin)

# to

def __init__(self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None)
```

four4fish:
> Also, we would be breaking it by changing the signature of the `Accelerator` from
> 
> ```python
> def __init__(self, precision_plugin: PrecisionPlugin, training_type_plugin: TrainingTypePlugin)
> 
> # to
> 
> def __init__(self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None)
> ```

Yes it will for positional parameter passing in. But precision_plugin is optional option now, has to switch the order.  Soon we will do step 3 which is move Accelerator into strategy, we have to remove precision_plugin entirely to avoid circle import. So it will be breaking change

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -407,7 +407,9 @@ def training_type_plugin(self) -> TrainingTypePlugin:
         # attach checkpoint plugin to the training type plugin
         if self._checkpoint_io is not None:
             self._training_type_plugin.checkpoint_io = self._checkpoint_io
-        self._training_type_plugin._precision_plugin = self.precision_plugin
+        precision_plugin = self.precision_plugin
+        if precision_plugin is not None:
```
kaushikb11, line 411: But it could never be None?

four4fish: Yeah, as we discussed above. But consider all the refactor happening now, feel safer have a none check? 
Training_type_plugin(precision_plugin=None) will have default value, but training_type_plugin.precision_plugin will have not default value.

ananthsub (approved):

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +45,29 @@ class Accelerator:
...
-        self.precision_plugin = precision_plugin
+
         self.training_type_plugin = training_type_plugin
 
+        if precision_plugin:
+            """
+            .. deprecated
+                precision_plugin parameter is deprecated will be removed soon.
+                Use :`training_type_plugin(precision_plugin) instead.
+            """
```
carmocca, line 68: nit: Triple quotes should be for docstrings, which this is not. Can you use comments instead?

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +45,29 @@ class Accelerator:
     One to handle differences from the training routine and one to handle different precisions.
     """
 
-    def __init__(self, precision_plugin: PrecisionPlugin, training_type_plugin: TrainingTypePlugin) -> None:
+    def __init__(
+        self, training_type_plugin: TrainingTypePlugin, precision_plugin: Optional[PrecisionPlugin] = None
```
carmocca, line 49: This change is not backward compatible, as the order is reversed now compared to before.

The only solution would be to set a token default value for training type plugin.

On a tangential note, I don't get the point for deprecating just the constructor, either the API is stable or it isn't, doing it for the constructor seems arbitrary.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -248,26 +263,36 @@ def setup_training_type_plugin(self) -> None:
...
     def precision(self) -> Union[str, int]:
-        return self.precision_plugin.precision
+        """This method is deprecated and will be removed soon.
+
+        Use `training_type_plugin.precision_plugin.precision` instead.
+        """
+        rank_zero_deprecation(
+            f"`{self.__class__.__name__}.precision` was and will be removed soon"
+            f" Use `training_type_plugin.precision_plugin.precision` instead."
+        )
```
carmocca, line 290: Can you merge/rebase master into this PR? To avoid any potential failures after #9940

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -456,8 +459,7 @@ def init_deepspeed(self):
                 "DeepSpeed currently does not support different `accumulate_grad_batches` at different epochs."
             )
 
-        precision = self.lightning_module.trainer.accelerator.precision
```
carmocca, line 459: awesome!

pytorch_lightning/trainer/trainer.py
```diff
@@ -1568,7 +1568,7 @@ def training_type_plugin(self) -> TrainingTypePlugin:
 
     @property
     def precision_plugin(self) -> PrecisionPlugin:
-        return self.accelerator.precision_plugin
+        return self.training_type_plugin.precision_plugin
```
carmocca, line 1571: Related to my comment about it being Optional, if that's the case, then this should check `None` and the typing would need to be fixed. Same for `precision` below

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         parallel_devices: Optional[List[torch.device]] = None,
         cluster_environment: Optional[ClusterEnvironment] = None,
         checkpoint_io: Optional[CheckpointIO] = None,
+        precision_plugin: Optional[PrecisionPlugin] = None,
```
carmocca, line 90: Why is it `Optional` here whereas it was required for the `Accelerator`?

This comment also applies to all other additions to each TTP.

pytorch_lightning/trainer/trainer.py
four4fish: It won't be None as 1. in accelerator we make sure precision plugin is not None (there was a resolved conversation about about this) 2. in ttp initialization, if precision_plugin is none, it will auto assign precision_plugin()

pytorch_lightning/accelerators/accelerator.py
kaushikb11: Comments for deprecation notes?

awaelchli: This must have been a merge resolution issue because I had already fixed this, including the formatting @four4fish.

pytorch_lightning/accelerators/accelerator.py
awaelchli: Reviewers need to agree here or we can't move forward. 

Either we add deprecation and backward compatibility for a few commits for all the 0.0001% users who actually install master and use these APIs, OR we remove the reference here, breaking change, and move on.

> either the API is stable or it isn't

The API is not stable and we know that from the docs. We prioritize making it stable for 1.6.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -44,15 +45,29 @@ class Accelerator:
...
+        Notes:
+            precision_plugin is deprecated and will be removed soon
+            User `training_type_plugin(precision_plugin)` instead
         """
-        self.precision_plugin = precision_plugin
+
         self.training_type_plugin = training_type_plugin
 
+        if precision_plugin:
+            """
```
tchaton, line 64: Let's move the depreciation within the docstring of the init function.

pytorch_lightning/accelerators/accelerator.py
tchaton: Hey @awaelchli. Adding to @carmocca  point, let's add the deprecation warning and not change the init API. Better to do it at once when we actually break the API.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -159,31 +174,31 @@ def training_step(self, step_kwargs: Dict[str, Union[Any, int]]) -> STEP_OUTPUT:
 
         See :meth:`~pytorch_lightning.core.lightning.LightningModule.training_step` for more details
         """
-        with self.precision_plugin.train_step_context():
+        with self.training_type_plugin.precision_plugin.train_step_context():
```
tchaton, line 177: Should the precision plugin own the `step_context` instead? I don't have a strong opinion there.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -84,7 +99,7 @@ def pre_dispatch(self, trainer: "pl.Trainer") -> None:
         if self.training_type_plugin.setup_optimizers_in_pre_dispatch:
             self.setup_optimizers(trainer)
 
-        self.precision_plugin.pre_dispatch()
+        self.training_type_plugin.precision_plugin.pre_dispatch()
```
tchaton, line 102: Should we use strategy already instead ?

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -407,6 +407,9 @@ def training_type_plugin(self) -> TrainingTypePlugin:
         # attach checkpoint plugin to the training type plugin
         if self._checkpoint_io is not None:
             self._training_type_plugin.checkpoint_io = self._checkpoint_io
+        precision_plugin = self.precision_plugin
+        if precision_plugin is not None:
+            self._training_type_plugin._precision_plugin = self.precision_plugin
```
tchaton, line 412:
```suggestion
        if self.precision_plugin is not None:
            self._training_type_plugin._precision_plugin = self.precision_plugin
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -795,12 +798,10 @@ def select_accelerator(self) -> Accelerator:
...
         else:
             acc_cls = CPUAccelerator
-        # as precision_plugin is dependent on training_type_plugin, make sure
-        # that we first select training_type_plugin, then precision_plugin
-        accelerator = acc_cls(training_type_plugin=self.training_type_plugin, precision_plugin=self.precision_plugin)
+
+        accelerator = acc_cls(training_type_plugin=self.training_type_plugin)
         # transfer ownership of the plugins to the accelerator
         self._training_type_plugin = proxy(self.training_type_plugin)
-        self._precision_plugin = proxy(self.precision_plugin)
```
tchaton, line 803: Neat !

pytorch_lightning/accelerators/accelerator.py
carmocca: Sorry if this wasn't clear, I'm fine with both deprecating OR not deprecating, I was just raising a question. I don't want to block the PR over that.

but IF the deprecation is chosen, then my point about setting a token default value stands.

pytorch_lightning/accelerators/accelerator.py
four4fish: Yep, all steps will be either move to training_type_plugin or precision in 2/n or 3/n

pytorch_lightning/accelerators/accelerator.py
four4fish: Agree we should unify the term. we haven't make that change yet? How about address in a different PR?

pytorch_lightning/trainer/connectors/accelerator_connector.py
four4fish: There is selection logic build in property precision plugin(): It's better call it once instead multiple times

https://github.com/PyTorchLightning/pytorch-lightning/blob/c5cabf836e0729bd48c16f538d48bb72651a73b8/pytorch_lightning/trainer/connectors/accelerator_connector.py#L393-L397

The self.select_precision_plugin() is quite heavy logic

pytorch_lightning/plugins/training_type/ddp.py
four4fish: Optional for user who construct Strategy(), they can not pass percision_plugin which default is None. Then we set to default for them. If it's TPU or IPU we know which precision_plugin to set when it's None, if not we use precision_plugin().

If the precision plugin choiced by accelerator_connector, it guaranteed won't be None

pytorch_lightning/accelerators/accelerator.py
awaelchli: Yes, the renaming has to happen in a separate PR and it will touch many files. This can definitely not be done here.

four4fish:
@tchaton @carmocca Addressed the comments, could you guys take a look?
1. Now this 1/n is backward compatible. 
2. Updated the deprecation message and format. Add to @awaelchli mentioned before: there will breaking change in refactor step3 "move accelerator into strategy". Which means we will have to remove precision_plugin param from Accelerator BEFORE we have a stable version to release. So the deprecation message is "will be deprecated soon" instead of "deprecated in 1.6 remove in 1.7"

pytorch_lightning/accelerators/accelerator.py
awaelchli: Yep, that's mostly it, but you can keep the text ".. deprecated soon" instead of v1.7 as the target :)

four4fish: is 
```
 .. deprecated:: soon
                    The ``precision_plugin`` parameter has been deprecated and will be removed in soon
                    Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
```
A valid format? Or do you think use comments better?

awaelchli: Nope :) 

This is our standard format: 

```
.. deprecated:: v1.6
                    The ``precision_plugin`` parameter has been deprecated in v1.6 and will be removed soon
                    Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
```

four4fish: How about we just use comments then? we gonna remove before 1.6, don't want to misleading user

awaelchli: Comments don't show in the documentation. We dont expect them to read the source files, right?
We could just drop the version number

.. deprecated::
    The ``precision_plugin`` parameter has been deprecated and will be removed soon
    Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.

pytorch_lightning/accelerators/accelerator.py
four4fish: Checked

pytorch_lightning/accelerators/accelerator.py
awaelchli: @carmocca Is the current way sufficient?

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: Sounds good @four4fish , just a note, it is currently also used multiple times, so you could make it also 

```suggestion
        precision_plugin = self.precision_plugin
        if precision_plugin is not None:
            self._training_type_plugin._precision_plugin = precision_plugin
```

four4fish: Yeah! typo!! thank you for catching it :)

carmocca (approved):

PR #10575, opened 2021-11-16
Re-design `call_hook` interface

### LINKED ISSUE #8506: [RFC] Re-design `call_hook` interface
## 🚀 Feature

The current `Trainer.call_hook` implementation has several drawbacks:

1. The call order cannot be modified. Forces that `Callback` hooks are called before `LightningModule` hooks of the same name, however, this is not always true. An example is `on_load_checkpoint`, where the order is the opposite.
2. Any trainer method (which maybe is not part of the `Callback` hook API) with the same name as the hook to call will be silently called. This is potentially dangerous if somebody names a trainer method with the same name as a hook.
3. Any accelerator method with the same names as the hook to call will be silently called. This is unwanted for some hook names as `setup` and `teardown`. Also dangerous.
4. It doesn't allow the `Callback` hook to return something

https://github.com/PyTorchLightning/pytorch-lightning/blob/e1442d247e0e4967dd2772bdcf5166226c974f89/pytorch_lightning/trainer/trainer.py#L1282-L1317

### Motivation

The previously mentioned reasons mean that the current implementation is error-prone. Additionally, when we support loop customization, we will want users to be able to call these hooks themselves.

### Pitch

Provide a lambda call_hook function that just handles any necessary state maintenance before the hook call.

Option (a):

```python
def call_hook(self, fn: Callable, hook_name: Optional[str], *args: Any, **kwargs: Any) -> Any:
    _set_current_fx_name()

    if hook_name is None:
        output = fn()
    else:
        with self.profiler.profile(hook_name):
            output = fn(*args, **kwargs)

    _unset_current_fx_name()

    return output
```

The state maintenance is currently just the `_current_fx_name` required for `self.log()`

This option means that the code will be more verbose but the calls will be made explicit and the user will have more responsibility.
This extra responsibility means that the user might break things if the hooks are not called in the proper order - whatever the proper order is for the hook in particular

### Alternatives

Option (b):

```python
def call_hook(self, hook_name: str, *args: Any, **kwargs: Any) -> Any:
    _set_current_fx_name()

    with self.profiler.profile(hook_name):
        _call_callback_hook(hook_name, *args, **kwargs)
        output = _call_model_hook(hook_name, *args, **kwargs)

    _unset_current_fx_name()

    return output
```

Does not resolve (1) and (4)

Option (c):

```python
def call_hook(
    self,
    hook_name: str,
    should_call_accelerator: bool = False,
    should_call_trainer: bool = False,
    *args: Any,
    **kwargs: Any
) -> Any:
    _set_current_fx_name()

    with self.profiler.profile(hook_name):
        _call_callback_hook(hook_name, *args, **kwargs)
        if should_call_trainer:
            # note: in the master impl this is silently done with the call above
            _call_trainer_hook(hook_name, **args, **kwargs)
        output = _call_model_hook(hook_name, *args, **kwargs)
        if should_call_accelerator:
            # note: in the master impl this can also return something
            _call_accelerator_hook(hook_name, **args, **kwargs)

    _unset_current_fx_name()

    return output
```

Does not resolve (1) and (4)

### Questions

Do the profilers merge profile timings with the same name?

#### If you enjoy PL, check out our other projects:

- [Metrics](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.
- [Flash](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning
- [Bolts](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch
- [Lightning Transformers](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc: @tchaton @awaelchli @ananthsub  

cc @borda @justusschock @awaelchli @akihironitta @tchaton

daniellepintz:
## What does this PR do?
Fixes #8506

Replaces `call_hook` with 4 methods: `call_callback_hooks`, `call_lightning_module_hook`, `call_accelerator_hook` and `call_ttp_hook`. The last two can hopefully be removed at some point in the future.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -196,22 +198,26 @@ def _on_evaluation_model_train(self) -> None:
...
+        self.trainer._call_hook(self.trainer, "on_epoch_start", *args, **kwargs)
+        self.trainer._call_hook(self.trainer.lightning_module, "on_epoch_start", *args, **kwargs)
 
         if self.trainer.testing:
-            self.trainer.call_hook("on_test_epoch_start", *args, **kwargs)
+            self.trainer._call_hook(self.trainer, "on_test_epoch_start", *args, **kwargs)
+            self.trainer._call_hook(self.trainer.lightning_module, "on_test_epoch_start", *args, **kwargs)
         else:
-            self.trainer.call_hook("on_validation_epoch_start", *args, **kwargs)
+            self.trainer._call_hook(self.trainer, "on_validation_epoch_start", *args, **kwargs)
```
daniellepintz, line 219: what should I do in the case of hooks like `on_validation_epoch_start` where the hook in callbacks takes trainer + LM, but the hook on LM takes no input args?

carmocca: That's managed already by the `TrainerCallbackHookMixin`

four4fish:
[RFC] @daniellepintz @awaelchli what's our plan for TraininerOptimizerMixin? TraininerOptimizerMixin is basically unrelated to trainer, the setup optimizer and lr logic can stay in Strategy. Strategy is the only place using setup optimizer and lr logic and it owns optimizers and lr. 
But TraininerOptimizerMixin is calling trainer.call_hook(). Could we separate call_hook from trainer, then we can remove reference to Trainer from strategy and remove TraininerOptimizerMixin.

ananthsub:
@four4fish that is a separate from reworking `call_hook`, so let's write up a separate issue/doc to discuss it

daniellepintz:
@four4fish the plan for TrainerOptimizerMixin is to move all of its logic to Strategy, and remove it. My understanding is we don't need to separate `call_hook` from the trainer in order to remove TrainerOptimizerMixin. re. the trainer reference on Strategy, is `call_hook` the only reason it is needed?

four4fish:
@daniellepintz after the refactor (include simplify spawning logic ), yeah i think so. Now we need pass trainer to setup, setup_optizer because call_hook in TrainerOptimizerMixin. Otherwise we only need trainer.fn.state.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1494,6 +1500,42 @@ def call_hook(
 
         return output
 
+    def _profile_and_call(self, obj: object, hook_name: str, *args: Any, **kwargs: Any) -> Optional[Any]:
+        fn = getattr(obj, hook_name)
+        with self.profiler.profile(f"{obj.__class__.__name__}.{hook_name}"):
+            return fn(*args, **kwargs)
+
+    def _call_hook(
```
ananthsub, line 1508: why add a new method instead of migrating `call_hook` in place?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1494,6 +1500,42 @@ def call_hook(
 
         return output
 
+    def _profile_and_call(self, obj: object, hook_name: str, *args: Any, **kwargs: Any) -> Optional[Any]:
+        fn = getattr(obj, hook_name)
+        with self.profiler.profile(f"{obj.__class__.__name__}.{hook_name}"):
+            return fn(*args, **kwargs)
+
```
ananthsub, line 1507: i'd rather this be directly in the implementation of `call_hook` to avoid making `_profile_and_call` broadly used elsewhere in the trainer codebase. what do you think?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1494,6 +1500,42 @@ def call_hook(
...
+            pl_module._current_fx_name = hook_name
+
+        output = None
+        if isinstance(obj, Trainer):
+            for obj in self.callbacks:
+                if hook_name in ("on_init_start", "on_init_end"):
+                    # these `Callback` hooks are the only ones that do not take a lightning module
+                    output = self._profile_and_call(obj, hook_name, self, *args, **kwargs)
+                else:
+                    output = self._profile_and_call(obj, hook_name, self, pl_module, *args, **kwargs)
```
ananthsub, line 1529: callbacks have a `state_key` which we should use to distinguish entries in the profiler output: https://github.com/PyTorchLightning/pytorch-lightning/blob/af0bb96f0ff645102680a7adc99dc131cfeb9c0b/pytorch_lightning/callbacks/base.py#L36-L44

pytorch_lightning/trainer/trainer.py
```diff
@@ -1494,6 +1500,42 @@ def call_hook(
...
+        with self.profiler.profile(f"{obj.__class__.__name__}.{hook_name}"):
+            return fn(*args, **kwargs)
+
+    def _call_hook(
+        self,
+        obj: object,
+        hook_name: str,
+        *args: Any,
+        pl_module: Optional["pl.LightningModule"] = None,
+        **kwargs: Any,
```
ananthsub, line 1514: i'm confused why `obj` and `pl_module` are passed in here. we should always be able to derive the pl_module from the trainer.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1494,6 +1500,42 @@ def call_hook(
...
+        pl_module = pl_module or self.lightning_module
+
+        if pl_module is not None:
+            prev_fx_name = pl_module._current_fx_name
+            pl_module._current_fx_name = hook_name
+
+        output = None
+        if isinstance(obj, Trainer):
+            for obj in self.callbacks:
+                if hook_name in ("on_init_start", "on_init_end"):
```
ananthsub, line 1525: these hooks never made sense to me. they leak the trainer constructor implementation detail. this is worth a separate discussion

pytorch_lightning/trainer/trainer.py
```diff
@@ -1494,6 +1500,42 @@ def call_hook(
...
+        hook_name: str,
+        *args: Any,
+        pl_module: Optional["pl.LightningModule"] = None,
+        **kwargs: Any,
+    ) -> Optional[Any]:
+        pl_module = pl_module or self.lightning_module
+
+        if pl_module is not None:
+            prev_fx_name = pl_module._current_fx_name
+            pl_module._current_fx_name = hook_name
```
ananthsub, line 1520: this means we're setting the hook name even for callback hooks, not lightning module hooks. is that correct?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1427,7 +1432,8 @@ def _call_teardown_hook(self) -> None:
         if self.datamodule is not None:
             self.datamodule.teardown(stage=fn)
 
-        self.call_hook("teardown", stage=fn)
+        self._call_hook(self, "teardown", stage=fn)
```
ananthsub, line 1435: passing `self` like this is very roundabout. it's already an instance method. `self` is accessible without needing to pass this arg.

pytorch_lightning/trainer/trainer.py
ananthsub: do you see a major inconvenience if we break this up like:
`_call_lightning_module_hook`
`_call_callback_hook`
`_call_strategy_hook` (most dubious if we need this)

each of which is profiled accordingly. and then `call_hook` can call these in order

daniellepintz: > why add a new method instead of migrating call_hook in place?

this is just temporary, the plan is to delete `call_hook`

> and then call_hook can call these in order

I thought one of the main points of this refactor was so that we have the flexibility to call the hooks in any order? i.e. now we do callback hooks first and then LM hooks, and then strategy hooks

> (most dubious if we need this)

I agree, can we simplify things by only having hooks for LM and callbacks?

pytorch_lightning/trainer/trainer.py
daniellepintz: Yes, makes sense

pytorch_lightning/trainer/trainer.py
daniellepintz: It could still make sense to break it up into:
```
_call_lightning_module_hook
_call_callback_hook
_call_strategy_hook
```

but rather than call them sequentially check which obj is being passed in `_call_hook` and then call the correct function?

pytorch_lightning/trainer/trainer.py
daniellepintz: for pl_module see @carmocca's answer here: https://github.com/PyTorchLightning/pytorch-lightning/issues/8506#issuecomment-969343535

I agree it would be much cleaner if we could not pass `pl_module`.

for obj, we could avoid passing it by creating the three methods you mentioned above, and each method is called directly, i.e. we wouldn't need `_call_hook`:
`_call_lightning_module_hook`
`_call_callback_hook`
`_call_strategy_hook`

The pros to this approach are the code is easier to read and understand, and less args passed, with a con being some code duplication. I think the three methods approach makes sense

pytorch_lightning/trainer/trainer.py
daniellepintz: https://github.com/PyTorchLightning/pytorch-lightning/discussions/10677

pytorch_lightning/trainer/trainer.py
daniellepintz: yeah I agree. this can be remedied by creating three diff methods as discussed above

pytorch_lightning/trainer/trainer.py
daniellepintz: I don't think so. it seems to me that `_current_fx_name` is just for the LM hooks so if we go with the three method approach, we can just include this logic directly in `_call_lightning_module_hook`?  

I am also a bit confused because we sometimes just call the LM hook directly, instead of calling `call_hook`:
https://github.com/PyTorchLightning/pytorch-lightning/blob/823bfa6f8a16a8ff2f93c8d5893549424e5b471b/pytorch_lightning/loops/dataloader/evaluation_loop.py#L235

Is this intentional or is this a mistake?

pytorch_lightning/trainer/trainer.py
daniellepintz: Good call, added

pytorch_lightning/trainer/trainer.py
daniellepintz: I checked and there are only two places where we actually pass in `pl_module` to call_hook:
https://github.com/PyTorchLightning/pytorch-lightning/blob/dcafc95f2b0fd3f176d425139ca99676ce943a12/pytorch_lightning/trainer/optimizers.py#L35
https://github.com/PyTorchLightning/pytorch-lightning/blob/dcafc95f2b0fd3f176d425139ca99676ce943a12/pytorch_lightning/trainer/data_loading.py#L577

Will create an issue for both

daniellepintz:
Marking this PR as ready for review but there are still tests failing with the same error:
```
def validation_epoch_end(self, outputs) -> None:
>       torch.stack([x["x"] for x in outputs]).mean()
E       RuntimeError: stack expects a non-empty TensorList
```
because for some reason `outputs` is empty. I have been unable to figure out the source of the issue, and would welcome any help.

stack trace: https://gist.github.com/daniellepintz/db722539210af762ae94ac7c59488c84

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -174,16 +173,18 @@ def _on_evaluation_start(self, *args: Any, **kwargs: Any) -> None:
         self._results.to(device=self.trainer.lightning_module.device)
 
         if self.trainer.testing:
-            self.trainer.call_hook("on_test_start", *args, **kwargs)
+            self.trainer._call_callback_hooks("on_test_start", *args, **kwargs)
+            self.trainer._call_lightning_module_hook("on_test_start", *args, **kwargs)
```
carmocca, line 177: TTP call missing

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -174,16 +173,18 @@ def _on_evaluation_start(self, *args: Any, **kwargs: Any) -> None:
...
         else:
-            self.trainer.call_hook("on_validation_start", *args, **kwargs)
+            self.trainer._call_callback_hooks("on_validation_start", *args, **kwargs)
+            self.trainer._call_lightning_module_hook("on_validation_start", *args, **kwargs)
```
carmocca, line 180: TTP call missing

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -196,22 +197,27 @@ def _on_evaluation_model_train(self) -> None:
     def _on_evaluation_end(self, *args: Any, **kwargs: Any) -> None:
         """Runs ``on_{validation/test}_end`` hook."""
         if self.trainer.testing:
-            self.trainer.call_hook("on_test_end", *args, **kwargs)
+            self.trainer._call_callback_hooks("on_test_end", *args, **kwargs)
+            self.trainer._call_lightning_module_hook("on_test_end", *args, **kwargs)
```
carmocca, line 201: TTP call missing

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -196,22 +197,27 @@ def _on_evaluation_model_train(self) -> None:
...
         else:
-            self.trainer.call_hook("on_validation_end", *args, **kwargs)
+            self.trainer._call_callback_hooks("on_validation_end", *args, **kwargs)
+            self.trainer._call_lightning_module_hook("on_validation_end", *args, **kwargs)
```
carmocca, line 204: TTP call missing

pytorch_lightning/loops/dataloader/prediction_loop.py
```diff
@@ -110,8 +110,10 @@ def _on_predict_start(self) -> None:
         self.trainer.lightning_module.zero_grad()
 
         # hook
-        self.trainer.call_hook("on_predict_start")
-        self.trainer.call_hook("on_predict_epoch_start")
+        self.trainer._call_callback_hooks("on_predict_start")
+        self.trainer._call_lightning_module_hook("on_predict_start")
```
carmocca, line 114: TTP call missing

pytorch_lightning/loops/dataloader/prediction_loop.py
```diff
@@ -110,8 +110,10 @@ def _on_predict_start(self) -> None:
...
+        self.trainer._call_callback_hooks("on_predict_epoch_start")
```
carmocca, line 115: nit: a newline would be nice here

```suggestion

        self.trainer._call_callback_hooks("on_predict_epoch_start")
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -225,18 +231,15 @@ def _evaluation_epoch_end(self, outputs: EPOCH_OUTPUT) -> None:
...
+            self.trainer._call_lightning_module_hook("validation_epoch_end", outputs)
 
     def _on_evaluation_epoch_end(self) -> None:
         """Runs ``on_{validation/test}_epoch_end`` hook."""
         hook_name = "on_test_epoch_end" if self.trainer.testing else "on_validation_epoch_end"
-        self.trainer.call_hook(hook_name)
-        self.trainer.call_hook("on_epoch_end")
+        self.trainer._call_callback_hooks(hook_name)
+        self.trainer._call_lightning_module_hook(hook_name)
+        self.trainer._call_callback_hooks("on_epoch_end")
```
carmocca, line 243: nit: a newline would be nice here
```suggestion

        self.trainer._call_callback_hooks("on_epoch_end")
```

pytorch_lightning/loops/dataloader/prediction_loop.py
```diff
@@ -133,7 +136,8 @@ def _on_predict_end(self) -> None:
         self.epoch_batch_indices = []
 
         # hook
-        self.trainer.call_hook("on_predict_end")
+        self.trainer._call_callback_hooks("on_predict_end")
+        self.trainer._call_lightning_module_hook("on_predict_end")
```
carmocca, line 140: TTP call missing

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -221,7 +221,9 @@ def _evaluation_step(self, batch: Any, batch_idx: int, dataloader_idx: int) -> O
     def _evaluation_step_end(self, *args: Any, **kwargs: Any) -> Optional[STEP_OUTPUT]:
```
carmocca, line 221: The step calls in the method above this one should also be converted

pytorch_lightning/loops/epoch/prediction_epoch_loop.py
```diff
@@ -125,7 +125,8 @@ def _predict_step(self, batch: Any, batch_idx: int, dataloader_idx: int) -> None
 
         model_ref = self.trainer.lightning_module
 
-        self.trainer.call_hook("on_predict_batch_start", batch, batch_idx, dataloader_idx)
+        self.trainer._call_callback_hooks("on_predict_batch_start", batch, batch_idx, dataloader_idx)
+        self.trainer._call_lightning_module_hook("on_predict_batch_start", batch, batch_idx, dataloader_idx)
 
         self.batch_progress.increment_started()
```
carmocca, line 131: There is a call right below this which also needs to be updated to use the TTP

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,84 @@ def _call_teardown_hook(self) -> None:
...
+        **kwargs: Any,
+    ) -> Optional[Any]:
+        pl_module = self.lightning_module
+        output = None
+
+        if pl_module:
+            prev_fx_name = pl_module._current_fx_name
+            pl_module._current_fx_name = hook_name
+
+        # TODO: remove if in v1.7
```
carmocca, line 1478:
```suggestion
        # TODO: remove it in v1.7
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,84 @@ def _call_teardown_hook(self) -> None:
...
+        if hook_name in ("on_train_batch_start", "on_train_batch_end"):
+            fn = getattr(self, hook_name)
+            if callable(fn):
+                with self.profiler.profile(hook_name):
+                    output = fn(*args, **kwargs)
+        else:
+            for callback in self.callbacks:
+                if hook_name in ("on_init_start", "on_init_end"):
+                    # these `Callback` hooks are the only ones that do not take a lightning module.
+                    # we also don't profile bc profiler hasn't been set yet
```
carmocca, line 1488: is this a minor optimization? the `PassThroughProfiler` should be set

carmocca (changes requested):
Several occurrences were missed. I suggest you look at the PR I had which was mostly complete and compare the two!

https://github.com/PyTorchLightning/pytorch-lightning/pull/9029/files

(I didn't review everything yet)

pytorch_lightning/trainer/trainer.py
daniellepintz: you are right for `on_init_end` the profiler has been set, but not for `on_init_start` since that happens here:
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d6262b7a91215e72019d720e742e6261e1636dc/pytorch_lightning/trainer/trainer.py#L502

and we configure profiler here:
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d6262b7a91215e72019d720e742e6261e1636dc/pytorch_lightning/trainer/trainer.py#L560

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,100 @@ def _call_teardown_hook(self) -> None:
...
+        pl_module = self.lightning_module
+        output = None
+
+        if pl_module:
+            prev_fx_name = pl_module._current_fx_name
+            pl_module._current_fx_name = hook_name
+
+        # TODO: remove code in if statement in v1.7
+        if hook_name in ("on_train_batch_start", "on_train_batch_end"):
+            fn = getattr(self, hook_name)
```
tchaton, line 1480: Quick optimization question. Should we explore the space of available overridden hooks on callbacks and run only them?

pytorch_lightning/trainer/callback_hook.py
```diff
@@ -32,136 +31,6 @@ class TrainerCallbackHookMixin(ABC):
...
-        """Called when either of train/val/test epoch ends."""
-        for callback in self.callbacks:
-            callback.on_epoch_end(self, self.lightning_module)
-
-    def on_train_start(self):
-        """Called when the train begins."""
-        for callback in self.callbacks:
-            callback.on_train_start(self, self.lightning_module)
-
-    def on_train_end(self):
```
tchaton, line 140: Neat !

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,100 @@ def _call_teardown_hook(self) -> None:
...
+        output = None
+        fn = getattr(self.training_type_plugin, hook_name)
+        if callable(fn):
+            with self.profiler.profile(hook_name):
+                output = fn(*args, **kwargs)
+
+        return output
+
+    # TODO: eventually no longer need this
+    def _call_accelerator_hook(
```
tchaton, line 1520: Do you see a need for call connectors hook ?

pytorch_lightning/trainer/callback_hook.py
```diff
@@ -32,136 +31,6 @@ class TrainerCallbackHookMixin(ABC):
...
-        """Called when the trainer initialization begins, model has not yet been set."""
-        for callback in self.callbacks:
-            callback.on_fit_start(self, self.lightning_module)
-
-    def on_fit_end(self):
-        """Called when the trainer initialization begins, model has not yet been set."""
-        for callback in self.callbacks:
-            callback.on_fit_end(self, self.lightning_module)
-
-    def on_sanity_check_start(self):
```
tchaton, line 75: Neat !

pytorch_lightning/trainer/trainer.py
daniellepintz: hmm, how do you envision changing the code? isn't that kind of what we are doing with `getattr` and then `if callable(fn)`?

pytorch_lightning/trainer/trainer.py
daniellepintz: Hmm I think it's better to leave the "call hook" functions to user defined hooks? The "hooks" in the connectors are not user defined right?

daniellepintz:
@carmocca I fixed the missed occurrences and the PR should be ready for another review

pytorch_lightning/trainer/trainer.py
```diff
@@ -1251,9 +1253,11 @@ def _pre_training_routine(self):
         # --------------------------
         # Pre-train
         # --------------------------
-        self.call_hook("on_pretrain_routine_start")
+        self._call_callback_hooks("on_pretrain_routine_start")
+        self._call_lightning_module_hook("on_pretrain_routine_start")
 
-        self.call_hook("on_pretrain_routine_end")
+        self._call_callback_hooks("on_pretrain_routine_end")
+        self._call_lightning_module_hook("on_pretrain_routine_end")
```
rohitgr7, line 1260: quick ques: why `_call_callback_hooks` with an `s`?

daniellepintz: I see `callback hook` as a compound noun composed of two nouns, similar to `rain boot` or `car seat`. To make a compound noun plural you only pluralize the second noun, i.e. `rain boots`, `car seats`, `callback hooks`

pytorch_lightning/trainer/trainer.py
tchaton: I think we could possibly pre-compute them before. Not sure if it would be worth it.

justusschock (approved):

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -218,19 +218,23 @@ def start_predicting(self, trainer: "pl.Trainer") -> None:
         self._results = trainer.run_stage()
 
     def training_step(self, *args, **kwargs):
-        return self.model.training_step(*args, **kwargs)
+        trainer = self.lightning_module.trainer
+        return trainer._call_lightning_module_hook("training_step", *args, **kwargs)
```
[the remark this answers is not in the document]
daniellepintz: Good point. This was done to address situations like this one:
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d6262b7a91215e72019d720e742e6261e1636dc/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py#L220-L222

My understanding was in general for accelerator and ttp hooks we don't want to set `pl_module._current_fx_name` so I didn't include that in the implementations of `_call_ttp_hook` and `_call_accelerator_hook`. However in this case we do bc the accel/ttp hooks are just calling the LM hooks directly. I can fix this by setting `pl_module._current_fx_name` in `_call_ttp_hook` and `_call_accelerator_hook`. wdyt? 
cc @carmocca

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,100 @@ def _call_teardown_hook(self) -> None:
...
-    ) -> Any:
-        pl_module = self.lightning_module or pl_module
+    def _call_lightning_module_hook(
+        self,
+        hook_name: str,
+        *args: Any,
+        pl_module: Optional["pl.LightningModule"] = None,
+        **kwargs: Any,
+    ):
+        pl_module = pl_module or self.lightning_module
```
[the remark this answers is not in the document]
daniellepintz: Good call. Added!

pl_examples/loop_examples/yielding_training_step.py
```diff
@@ -88,8 +88,9 @@ def _training_step(self, generator):
         training_step_output = next(generator)
         self.trainer.training_type_plugin.post_training_step()
 
-        training_step_output = self.trainer.call_hook("training_step_end", training_step_output)
-
+        model_output = self.trainer._call_lightning_module_hook("training_step_end", training_step_output)
+        ttp_output = self.trainer._call_ttp_hook("training_step_end", training_step_output)
+        training_step_output = ttp_output if model_output is None else model_output
```
carmocca, line 93: nit: keep the whitespace

```suggestion
        training_step_output = ttp_output if model_output is None else model_output

```

pytorch_lightning/loops/fit_loop.py
```diff
@@ -250,7 +252,8 @@ def on_run_end(self) -> None:
         self.current_epoch = max(self.current_epoch - 1, 0)
 
         # hook
-        self.trainer.call_hook("on_train_end")
+        self.trainer._call_callback_hooks("on_train_end")
+        self.trainer._call_lightning_module_hook("on_train_end")
```
carmocca, line 256: Missing TTP call

pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -426,14 +427,12 @@ def _training_step(self, split_batch: Any, batch_idx: int, opt_idx: int) -> Clos
...
-            with self.trainer.profiler.profile("training_step"):
-                training_step_output = self.trainer.accelerator.training_step(step_kwargs)
-                self.trainer.training_type_plugin.post_training_step()
+            training_step_output = self.trainer._call_accelerator_hook("training_step", step_kwargs)
+            self.trainer.training_type_plugin.post_training_step()
 
             del step_kwargs
 
-            training_step_output = self.trainer.call_hook("training_step_end", training_step_output)
+            training_step_output = self.trainer._call_lightning_module_hook("training_step_end", training_step_output)
```
carmocca, line 435: This is missing the TTP call and output selection. Should match what you have already in manual_loop

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -288,7 +288,7 @@ def dataloader(self) -> Union[TRAIN_DATALOADERS, EVAL_DATALOADERS]:
             return self.instance
 
         if isinstance(self.instance, LightningModule):
-            return self.instance.trainer.call_hook(self.name, pl_module=self.instance)
```
carmocca, line 291: why did this remove?

`pl_module=self.instance`

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,105 @@ def _call_teardown_hook(self) -> None:
...
+
+        assert pl_module is not None, "No lightning module is available to call hooks on"
```
carmocca, line 1449: Use an actual exception, asserts should be just for tests and typing enforcement

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,105 @@ def _call_teardown_hook(self) -> None:
...
+
+        fn = getattr(pl_module, hook_name)
+        assert callable(fn), "Hook is not callable"
```
carmocca, line 1452: Same here

pytorch_lightning/trainer/trainer.py
carmocca: I wouldn't explore this in this PR.

And there's probably more valuable optimizations to explore.

pytorch_lightning/trainer/trainer.py
carmocca: Yes

pytorch_lightning/plugins/training_type/training_type_plugin.py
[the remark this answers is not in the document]
carmocca: Yes. The problem here is that an object that contains called hooks (the TTP) needs to call a different object (the LM) that contains called hooks.

Since the interface is in the Trainer, we are forced to get it from the LM.

It's something we should avoid, but I would leave it as it is for this PR. It's already large enough.

carmocca (commented):
This is very close to the finish line. I'll review again and approve when CI is green.

pytorch_lightning/trainer/connectors/data_connector.py
daniellepintz: not sure, I put it back. I wish we didn't need this tho :/

pytorch_lightning/trainer/trainer.py
carmocca: I generally suggest `Misconfiguration` if it's actionable by the end-user (meaning, the user can do something about it and it's communicated in the message)

Otherwise one of https://docs.python.org/3/library/exceptions.html#concrete-exceptions

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,109 @@ def _call_teardown_hook(self) -> None:
...
+    ) -> Optional[Any]:
+        output = None
+        fn = getattr(self.accelerator, hook_name)
+        if not callable(fn):
+            raise MisconfigurationException("Hook is not callable")
+
+        with self.profiler.profile(hook_name):
+            output = fn(*args, **kwargs)
+
+        return output
```
ananthsub, line 1541: @four4fish @awaelchli  fyi since you are actively refactoring the accelerator/training type plugin structure. this should go away after that refactor

four4fish: Will be call_strategy_hook right?

ananthsub: yes, we can rename the `_call_ttp_hook` to `_call_strategy_hook` once the rename is formalized, and remove this `_call_accelerator_hook` entirely

four4fish: Issue created https://github.com/PyTorchLightning/pytorch-lightning/issues/10905

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -1431,58 +1437,109 @@ def _call_teardown_hook(self) -> None:
         # summarize profile results
         self.profiler.describe()
 
-    def call_hook(
```
awaelchli, line 1434: will you deprecate this one?

pytorch_lightning/trainer/trainer.py
daniellepintz: yup, I am planning on doing this as a follow up to this PR

pytorch_lightning/trainer/trainer.py
daniellepintz: because the new `_call_hook` functions are replacing it completely. There are no more calls of `call_hook` left in the codebase

daniellepintz: I am just removing it, not deprecating, since is it meant to be used only in our internal codebase

daniellepintz:
@ananthsub @carmocca I'm not sure we should be raising an exception if the hook is not callable. Current behavior is we just silently pass, returning output=None, and changing this behavior is causing a lot of tests to fail

awaelchli:
Probably many tests are making use of this current behavior by setting things like
`model.training_epoch_end = None` (none is not callable) to deactivate that hook on an existing model (say BoringModel or EvalModelTemplate)

pytorch_lightning/trainer/trainer.py
ananthsub: @daniellepintz to be safe, I think we should go with the deprecation out of precaution. This deprecation could be handled in a separate PR to keep this one focused on introducing the granular hook funtions & improved profiling

pytorch_lightning/trainer/trainer.py
```diff
@@ -1441,58 +1447,105 @@ def _call_teardown_hook(self) -> None:
...
+            if callable(fn):
+                with self.profiler.profile(hook_name):
+                    output = fn(*args, **kwargs)
+        elif hook_name in ("on_init_start", "on_init_end"):
+            # these `Callback` hooks are the only ones that do not take a lightning module.
+            # we also don't profile bc profiler hasn't been set yet
+            for callback in self.callbacks:
+                fn = getattr(callback, hook_name)
+                if callable(fn):
+                    output = fn(self, *args, **kwargs)
```
ananthsub, line 1503: thanks @daniellepintz for starting the discussion around this. It looks like there's appetite to deprecate these hooks  at which point this can be removed in v1.8: https://github.com/PyTorchLightning/pytorch-lightning/issues/10894

daniellepintz:
There are 12 failing tests remaining in the CI, and they are all passing locally for me so I am quite confused:
```
FAILED tests/accelerators/test_accelerator_connector.py::test_plugin_accelerator_choice[ddp_spawn-ddp_sharded]
FAILED tests/accelerators/test_accelerator_connector.py::test_plugin_accelerator_choice[None-ddp_sharded]
FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[1-ddp_sharded-DDPShardedPlugin]
FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[1-ddp_sharded_spawn-DDPSpawnShardedPlugin]
FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[2-ddp_sharded-DDPShardedPlugin]
FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[2-ddp_sharded_spawn-DDPSpawnShardedPlugin]
FAILED tests/plugins/test_cluster_integration.py::test_ranks_available_manual_plugin_selection[DDPShardedPlugin]
FAILED tests/plugins/test_cluster_integration.py::test_ranks_available_automatic_plugin_selection[trainer_kwargs1]
FAILED tests/plugins/test_plugins_registry.py::test_ddp_find_unused_parameters_training_type_plugins_registry[ddp_sharded_spawn_find_unused_parameters_false-DDPSpawnShardedPlugin]
FAILED tests/plugins/test_plugins_registry.py::test_ddp_find_unused_parameters_training_type_plugins_registry[ddp_sharded_find_unused_parameters_false-DDPShardedPlugin]
FAILED tests/trainer/test_trainer.py::test_trainer_config_strategy[trainer_kwargs27-expected27]
FAILED tests/trainer/test_trainer.py::test_trainer_config_strategy[trainer_kwargs28-expected28]
```

ananthsub:
> There are 12 failing tests remaining in the CI, and they are all passing locally for me so I am quite confused:
> 
> ```
> FAILED tests/accelerators/test_accelerator_connector.py::test_plugin_accelerator_choice[ddp_spawn-ddp_sharded]
> FAILED tests/accelerators/test_accelerator_connector.py::test_plugin_accelerator_choice[None-ddp_sharded]
> FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[1-ddp_sharded-DDPShardedPlugin]
> FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[1-ddp_sharded_spawn-DDPSpawnShardedPlugin]
> FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[2-ddp_sharded-DDPShardedPlugin]
> FAILED tests/accelerators/test_accelerator_connector.py::test_accelerator_choice_multi_node_gpu[2-ddp_sharded_spawn-DDPSpawnShardedPlugin]
> FAILED tests/plugins/test_cluster_integration.py::test_ranks_available_manual_plugin_selection[DDPShardedPlugin]
> FAILED tests/plugins/test_cluster_integration.py::test_ranks_available_automatic_plugin_selection[trainer_kwargs1]
> FAILED tests/plugins/test_plugins_registry.py::test_ddp_find_unused_parameters_training_type_plugins_registry[ddp_sharded_spawn_find_unused_parameters_false-DDPSpawnShardedPlugin]
> FAILED tests/plugins/test_plugins_registry.py::test_ddp_find_unused_parameters_training_type_plugins_registry[ddp_sharded_find_unused_parameters_false-DDPShardedPlugin]
> FAILED tests/trainer/test_trainer.py::test_trainer_config_strategy[trainer_kwargs27-expected27]
> FAILED tests/trainer/test_trainer.py::test_trainer_config_strategy[trainer_kwargs28-expected28]
> ```

The CI is failing with this
```
pytorch_lightning.utilities.exceptions.MisconfigurationException: `DDPShardedPlugin` requires `fairscale` to be installed. Install it by running `pip install fairscale`.
```

daniellepintz:
no, I didn't change anything with fairscale. maybe @awaelchli knows something since he has been working with DDP recently?

awaelchli:
Happy to help. 

Take the example of this test that fails: `tests/trainer/test_trainer.py::test_trainer_config_strategy`
You see the error in CI and not locally because:
- fairscale not installed in CPU CI, locally you have it installed
- the test initializes a Trainer with sharded plugin
- during Trainer init, there is a call to `call_callback_hooks("on_init_start")`
- This guy accesses training_type_plugin.lightning_module
- the sharded plugin cannot work if fairscale is not installed, hence the error
- even if sharded plugin behaved differently, a lightningmodule reference is not available in Trainer init anyhow. 

This all occurs because this PR replaced
`self.on_init_start()`
with
`self._call_callback_hooks("on_init_start")`
And that is no longer equivalent to what it was before. The same problem would occur if we replaced `self.on_init_start()` with `self.call_hook("on_init_start")` on master. `on_init_start` only exists for callbacks, not for the LightningModule. 

My suggestion for resolution:
- Special case the cases for on_init_start/end not to access `self.lightning_module` (this is partially already done inside `_call_callback_hooks`
- Eventually remove on_init_start/end

daniellepintz:
@awaelchli thank you SO much for your help! That makes total sense. Fixed!!

pytorch_lightning/trainer/trainer.py
```diff
@@ -1457,7 +1457,7 @@ def _call_lightning_module_hook(
         pl_module = pl_module or self.lightning_module
 
         if pl_module is None:
-            raise MisconfigurationException("No lightning module is available to call hooks on")
+            raise Exception("No Lightning Module is available to call hooks on")
```
carmocca, line 1460: TypeError would fit better here? cc @justusschock 

#10918

pytorch_lightning/trainer/trainer.py
```diff
@@ -1441,58 +1447,106 @@ def _call_teardown_hook(self) -> None:
...
+    def _call_accelerator_hook(
+        self,
+        hook_name: str,
+        *args: Any,
+        **kwargs: Any,
+    ) -> Optional[Any]:
+        fn = getattr(self.accelerator, hook_name)
+        if not callable(fn):
+            return None
+
```
carmocca, line 1544: We discussed not setting the `current_fx_name` for accelerator but since we do it manually in all places that call this method, might as well put it inside.

carmocca (approved):
Epic!

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -176,7 +179,12 @@ def advance(self, data_fetcher: AbstractDataFetcher) -> None:  # type: ignore[ov
             )
 
             # hook
-            response = self.trainer.call_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)
+            self.trainer._call_callback_hooks("on_train_batch_start", batch, batch_idx, **extra_kwargs)
+            model_response = self.trainer._call_lightning_module_hook(
+                "on_train_batch_start", batch, batch_idx, **extra_kwargs
+            )
+            ttp_response = self.trainer._call_ttp_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)
+            response = ttp_response if model_response is None else model_response
```
carmocca, line 187: The TTP is annotated as returning `None`

```suggestion
            response = self.trainer._call_lightning_module_hook(
                "on_train_batch_start", batch, batch_idx, **extra_kwargs
            )
            self.trainer._call_ttp_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)            
```

pytorch_lightning/trainer/trainer.py
daniellepintz: Yeah, I ended up having to change it because some tests were failing. Also at least this way we don't use the LM's trainer reference in TTP.py

pytorch_lightning/loops/epoch/training_epoch_loop.py
daniellepintz: it's not directly annotated as returning None, it is just missing the return type annotation. The reason it is missing the return type annotation is because it previously had `-> Optional[Any]` as return type but this caused a mypy error:
`pytorch_lightning/loops/optimization/optimizer_loop.py:439: error: Argument 1 to "_extract_hiddens" has incompatible type "Optional[Any]"; expected "Union[Tensor, Dict[str, Any]]"  [arg-type]`

so I removed it. But in order to maintain the current behavior of call_hook, ttp hooks _should_ be allowed to return something.

pytorch_lightning/trainer/trainer.py
daniellepintz: I don't feel like TypeError exactly fits but I guess it is the best option

daniellepintz:
Merged, but @carmocca if you have any follow ups I can add them in subsequent PRs

pytorch_lightning/trainer/trainer.py
awaelchli: I have some mind boggling problems in #10890. One question, why do these new call_xyz methods not reset the current_fx_name to its previous value? Was this forgotten or a mistake?

daniellepintz: After discussion with @carmocca I learned that only LM hooks and callback hooks need to set current_fx_name, so I only included that in the `_call_LM_hook` and `_call_callback_hooks`, and not for `_call_ttp_hook` and `_call_accelerator_hook`. However, before this PR, all the places where we called the accelerator hook had a line like this:
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d6262b7a91215e72019d720e742e6261e1636dc/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py#L220

so I included `self.lightning_module._current_fx_name = hook_name` in `_call_accelerator_hook` and got rid of the individual lines like the above

carmocca:
Just opened https://github.com/PyTorchLightning/pytorch-lightning/pull/10957 with what I think was missing,

@awaelchli, for your PR, you'll have to set and unset in `_call_ttp_hook` as you've moved the calls from the accelerator to the ttp.

PR #10578, opened 2021-11-17
Use `update_wrapper` in test_hooks.py

carmocca:
## What does this PR do?

Fixes a bug in the hook tests. Found with #9029

None

## Before submitting

## PR review

cc @borda @carmocca @awaelchli

### DIFF
--- a/tests/models/test_hooks.py
+++ b/tests/models/test_hooks.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from functools import partial
+from functools import partial, update_wrapper
 from inspect import getmembers, isfunction
 from unittest import mock
 from unittest.mock import ANY, PropertyMock
@@ -223,7 +223,9 @@ class HookedCallback(Callback):
 
         for h in get_members(Callback):
             attr = getattr(self, h)
-            setattr(self, h, partial(call, h, attr))
+            partial_h = partial(call, h, attr)
+            update_wrapper(partial_h, attr)
+            setattr(self, h, partial_h)
 
     def on_save_checkpoint(*args, **kwargs):
         return {"foo": True}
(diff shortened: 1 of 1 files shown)


awaelchli:
@carmocca can you explain what the bug was? It would help the reviewer

tests/models/test_hooks.py
```diff
@@ -872,20 +878,12 @@ def call(hook, fn, *args, **kwargs):
     called = []
     dm = HookedDataModule(called)
     trainer.fit(model, datamodule=dm)
-    batch_transfer = [
-        dict(name="on_before_batch_transfer", args=(ANY, 0)),
-        dict(name="transfer_batch_to_device", args=(ANY, torch.device("cpu"), 0)),
-        dict(name="on_after_batch_transfer", args=(ANY, 0)),
-    ]
```
carmocca, line 879: This test only aims to test the order for the `LightningDataModule` hooks, however, the transfer of batches is part of the LightningModule. Why were these here then?

The reason is that `LightningDataModule` subclasses `DataHooks`.
https://github.com/PyTorchLightning/pytorch-lightning/blob/ba036fdeeade11aeb43f3f35ff1b924b3780adc1/pytorch_lightning/core/datamodule.py#L26
And `DataHooks` implements the transfer hooks.
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/core/hooks.py#L709
So for this test both the `HookedDataModule` and `BoringModel` share the same implementation (none of them override it).

This means that when we were wrapping it with partial above, we were doing it for this shared implementation and thus the hook was getting included in `called` for the DataModule when in reality it was for the `LightningModule`.

`update_wrapper` fixes that.

cc @awaelchli

awaelchli: Thanks, that's pretty nasty.
Do you think a testing approach with unittest.Mock would have helped to prevent this here? (aware we are currently limited by some internals preventing to use Mock)

awaelchli (approved):

tests/models/test_hooks.py
carmocca: I wouldn't know without trying

tchaton (approved):
LGTM !

Borda (approved):

PR #10603, opened 2021-11-18
Avoid patching common `DataHooks` to the `LightningModule`

### LINKED ISSUE #10498: Avoid patching `DataHooks`

We need to avoid patching some of these datahooks:
https://github.com/PyTorchLightning/pytorch-lightning/blob/09cf167237e867f1ec67a5db87e5a02c2cea4b69/pytorch_lightning/trainer/connectors/data_connector.py#L238-L242

We have already removed the patching for dataloader related hooks here: https://github.com/PyTorchLightning/pytorch-lightning/pull/9764

### Motivation

It can fail in some cases. One example:

<p>

```py
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len

class BoringData(LightningDataModule):
    def __init__(self):
        super().__init__()
        self.batch_size = 16
        self.total_len = 1000
        
    def forward(self, x):
        return self.layer(x)
        
    def train_dataloader(self):
         return DataLoader(RandomDataset(32, self.total_len), batch_size=self.batch_size)
    
    def val_dataloader(self):
        return DataLoader(RandomDataset(32, self.total_len), batch_size=self.batch_size*2)
    
    def transfer_batch_to_device(self, *args, **kwargs):
        print(f'{self.trainer.state.stage}: DataModule.transfer_batch_to_device')
        return super().transfer_batch_to_device(*args, **kwargs)

    def __init__(self):
        super().__init__()
        self.lr = 1e-2
        
    def transfer_batch_to_device(self, *args, **kwargs):
        print(f'{self.trainer.state.stage}: LightningModule.transfer_batch_to_device')
        return super().transfer_batch_to_device(*args, **kwargs)
        
    def forward(self, x):
        return self.layer(x)

        logits = self(batch)
        loss = torch.nn.functional.mse_loss(logits, torch.randn(*logits.shape))
        return loss
    
        logits = self(batch)
        loss = torch.nn.functional.mse_loss(logits, torch.randn(*logits.shape))
    
        logits = self(batch)
        loss = torch.nn.functional.mse_loss(logits, torch.randn(*logits.shape))

        opt = torch.optim.SGD(self.layer.parameters(), lr=1e-2)
        return opt

test_dl = DataLoader(RandomDataset(32, 1000), batch_size=16)
model = BoringModel()
dm = BoringData()

trainer = Trainer(fast_dev_run=True)
trainer.fit(model, datamodule=dm)
trainer.test(model, dataloaders=test_dl)
```
</p>
</details>

Prints:
```
train: DataModule.transfer_batch_to_device
validate: DataModule.transfer_batch_to_device
test: DataModule.transfer_batch_to_device
```

Expected:
```
train: DataModule.transfer_batch_to_device
validate: DataModule.transfer_batch_to_device
test: LightningModule.transfer_batch_to_device
```

### Additional context

cc @justusschock @awaelchli @akihironitta @rohitgr7 @ninginthecloud

rohitgr7:
## What does this PR do?

Fixes #10498

If any of these hooks are overridden in both model and datamodule, then we raise a warning and prioritize datamodule.

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @rohitgr7 @carmocca @ananthsub @ninginthecloud @jjenniferdai @akihironitta


pytorch_lightning/core/lightning.py
```diff
@@ -263,9 +263,13 @@ def _apply_batch_transfer_handler(
         self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0
     ) -> Any:
         device = device or self.device
-        batch = self.on_before_batch_transfer(batch, dataloader_idx)
-        batch = self.transfer_batch_to_device(batch, device, dataloader_idx)
-        batch = self.on_after_batch_transfer(batch, dataloader_idx)
+        batch = self.trainer._data_connector._datahook_source.get_hook("on_before_batch_transfer")(
```
tchaton, line 266: Does it mean only the objects providing the data loaders can provide the implementation for those hooks?
I am fine with this change, but this would be behavior BC.

rohitgr7: I would say yes.. but we can make adjustments by checking where these hooks are overridden and using that one. In case the same hook is overridden in both Module and DataModule, then we can either raise an error or prioritize datamodule with a warning. wdyt?

tchaton: I am fine with this change, just wanted to be sure.

tchaton (approved):
LGTM !

pytorch_lightning/core/lightning.py
```diff
@@ -263,9 +264,13 @@ def _apply_batch_transfer_handler(
         self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0
     ) -> Any:
         device = device or self.device
-        batch = self.on_before_batch_transfer(batch, dataloader_idx)
-        batch = self.transfer_batch_to_device(batch, device, dataloader_idx)
-        batch = self.on_after_batch_transfer(batch, dataloader_idx)
+        datahook_source = (
+            _DataHookSource(self) if self.trainer is None else self.trainer._data_connector._datahook_source
```
ananthsub, line 268: can we do this in a way without adding more dependencies to referring to `self.trainer` inside of the LightningModule? 

Why is `_apply_batch_transfer_handler` defined in the LightningModule at all at this point? Should this logic be in the Trainer, so it can select which datasource to call these hooks on?

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -308,3 +306,22 @@ def is_module(self) -> bool:
...
+    2. a method from the :class:`~pytorch_lightning.core.datamodule.LightningDataModule`,
+
+    Arguments:
+        instance: A LightningModule or a LightningDataModule
+    """
+
+    instance: Optional[Union["pl.LightningModule", "pl.LightningDataModule"]]
+
+    def get_hook(self, hook_name):
+        return getattr(self.instance, hook_name)
```
ananthsub, line 327: setup and teardown are also part of DataHooks: https://github.com/PyTorchLightning/pytorch-lightning/blob/2c7c4aab8087d4c1c99c57c7acc66ef9a8e815d4/pytorch_lightning/core/hooks.py#L373-L399

This could easily be misused to call `setup` or `teardown` only once between DataModule and LightningModule instead of on both

pytorch_lightning/core/lightning.py
```diff
@@ -263,9 +264,13 @@ def _apply_batch_transfer_handler(
...
+        )
+
+        batch = datahook_source.get_hook("on_before_batch_transfer")(batch, dataloader_idx)
+        batch = datahook_source.get_hook("transfer_batch_to_device")(batch, device, dataloader_idx)
+        batch = datahook_source.get_hook("on_after_batch_transfer")(batch, dataloader_idx)
```
ananthsub, line 273: while the current patch is bad, this doesn't address what we want the fundamental hook order to be here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10498#issuecomment-973228003

pytorch_lightning/trainer/connectors/data_connector.py
rohitgr7: but these are not specific to just data. For eg one can initialize a module component inside `LightningModule.setup` and some data-specific component inside `Datamodule.setup`, so we need to call both in such a case. A similar case can happen for teardown.

pytorch_lightning/core/lightning.py
rohitgr7: left a comment here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10498#issuecomment-973852552
I'd say we can prioritize datamodule with a warning.

pytorch_lightning/core/lightning.py
rohitgr7: we use `_apply_batch_transfer_handler` at other places like model_summary, to_torchscript, to_onnx. so here I added this check to make sure it doesn't rely on the trainer completely. If a trainer is not attached, it will use its own logic to move the data around.

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -308,3 +308,35 @@ def is_module(self) -> bool:
         from pytorch_lightning import LightningDataModule, LightningModule  # prevent cyclic import
 
         return isinstance(self.instance, (LightningModule, LightningDataModule))
+
+
+@dataclass
+class _DataHookSource:
```
awaelchli, line 314: I disagree with this. The name and usage is inspired by the class I wrote above, but this not how it is meant to work. 
A XSource is meant to declare one source and one way to retrieve an instance from that source. 
As you can see in the design above, the source only contains one reference and what that reference is, is determined before instantiating the source.

It is unclear to me why we need two classes instead of just one.

awaelchli: Given the name _DataHookSource, perhaps your intention was to scrap the _DataLoaderSource above? After all, the methods x_dataloader() are hooks too.

rohitgr7: yes but those hooks are more like data providers and the data source can also be explicit too (`.fit(train_dataloader=...`), but these hooks are sort of like data-handlers. For BC and the possibility of overriding the same method in both module and datamodule, I thought of creating an additional class that can work with any hooks that are common in both. For eg, if users provide datamodule but override one of these hooks within LightningModule, then we can't rely on just the datamodule and ignore the implementation within LightningModule.

One possible improvement I can think of is we can start raising a warning there if the data provider is different from the component where the hook is overridden and later merge both of them. Also possibly create a temp datamodule for explicit dataloaders as suggested here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10498#issuecomment-973812617

CHANGELOG.md
```diff
@@ -168,6 +168,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed `DeepSpeedPlugin.{precision,amp_type,amp_level}` properties ([#10657](https://github.com/PyTorchLightning/pytorch-lightning/pull/10657))
 
 
+- Removed patching of common `DataHooks` to the `LightningModule` ([#10603](https://github.com/PyTorchLightning/pytorch-lightning/pull/10603))
```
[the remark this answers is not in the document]
ninginthecloud: agree. There are 3 hooks  ("on_before_batch_transfer", "transfer_batch_to_device", "on_after_batch_transfer") being updated in this PR, let's be specific and list them.

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -615,3 +615,35 @@ def is_module(self) -> bool:
...
+
+    Arguments:
+        model: A LightningModule
+        datamodule: A LightningDataModule
+    """
+
+    model: Optional["pl.LightningModule"]
+    datamodule: Optional["pl.LightningDataModule"]
+
+    def get_hook(self, hook_name):
```
ninginthecloud, line 637: This class currently seems only to be used for  3 hook methods, then let's check if the hook_name is valid.

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -615,3 +615,35 @@ def is_module(self) -> bool:
...
+
+        if is_overridden(hook_name, self.datamodule):
+            if is_overridden(hook_name, self.model):
+                warning_cache.warn(
+                    f"You have overridden `{hook_name}` in both LightningModule and LightningDataModule."
+                    " We will use the implementation from LightningDataModule instance."
+                )
+            return getattr(self.datamodule, hook_name)
+
+        return getattr(self.model, hook_name)
```
ninginthecloud, line 655: For datamodule is not None, but the specific hook function has not been overridden, current logic is to use hooks from model. Is this a valid behavior? How about throwing a warning that requested hooks come from model instead of datamodule? Because *_dataloader hook source may come from datamodule, but the transfer_* hook may come from lightningmodule.

rohitgr7: I kept that for BC since there's no warning in such a case as of now. Do you think we should change this behavior and mention it in the warning that if datamodule is passed then it will strictly use datamodule hooks only in v1.7?

or just a warning is fine?

ninginthecloud: I'm fine with just warning. User can specify specific data source in `.fit()` , but it did not specify where the transfer hook source come from, so it will be better to communicate and clarifying the underlying logics.

ninginthecloud (commented):
Hi, @rohitgr7 I'm ok with this update, thanks a lot for all the change! Would you add the warnings into the codebase as you proposed in the comment? Additionally, we currently have both `_DataHookSource` and `_DataLoaderSource`, could we simplify as just one? Afterall, `*_dataloader`'s are also data hooks, the purpose of source is providing a source of truth handling different overridden. Let me know what you think.

rohitgr7:
> Hi, @rohitgr7 I'm ok with this update, thanks a lot for all the change! Would you add the warnings into the codebase as you proposed in the comment? Additionally, we currently have both `_DataHookSource` and `_DataLoaderSource`, could we simplify as just one? Afterall, `*_dataloader`'s are also data hooks, the purpose of source is providing a source of truth handling different overridden. Let me know what you think.

yes @ninginthecloud, I was going to propose this in a follow-up issue. Also, we need to ensure to have strict rules there:
1. strictly use LDM for dataloader hooks just like we do now. No warnings in such a case.
2. Strictly follow LDM, if passed, for both common hooks and dataloader hooks. For other hooks, IMO we could start raising the warning for the current behavior i.e if LDM is passed, use hooks for LM if there are not overridden in LDM but overridden in LightningModule. Making LDM as a source of truth always.

ninginthecloud (approved):
Thanks for the update!

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -628,3 +628,48 @@ def is_module(self) -> bool:
...
+            return getattr(self.datamodule, hook_name)
+
+        warning_cache.warn(
+            f"You have overridden `{hook_name}` in `LightningModule` but have passed in a"
+            " `LightningDataModule`. It will use the implementation from `LightningModule` instance."
+        )
+        return getattr(self.model, hook_name)
+
+    def __post_init__(self):
+        self._valid_hooks = ("on_before_batch_transfer", "transfer_batch_to_device", "on_after_batch_transfer")
```
carmocca, line 675: Why do this in the post init instead of making it a keyword field?

pytorch_lightning/trainer/connectors/data_connector.py
carmocca: Perhaps `DataHookSelector` would be a better name?

pytorch_lightning/core/lightning.py
carmocca: Resolving as I think you've implemented this already

pytorch_lightning/core/lightning.py
```diff
@@ -255,9 +256,13 @@ def _apply_batch_transfer_handler(
...
     ) -> Any:
         device = device or self.device
-        batch = self.on_before_batch_transfer(batch, dataloader_idx)
-        batch = self.transfer_batch_to_device(batch, device, dataloader_idx)
-        batch = self.on_after_batch_transfer(batch, dataloader_idx)
+        datahook_source = (
+            _DataHookSource(self, None) if self.trainer is None else self.trainer._data_connector._datahook_source
+        )
+
+        batch = datahook_source.get_hook("on_before_batch_transfer")(batch, dataloader_idx)
```
carmocca, line 263: Can you separate the two calls?

```suggestion
        hook = datahook_source.get_hook("on_before_batch_transfer")
        batch = hook(batch, dataloader_idx)
```

pytorch_lightning/core/lightning.py
```diff
@@ -255,9 +256,13 @@ def _apply_batch_transfer_handler(
         self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0
     ) -> Any:
         device = device or self.device
-        batch = self.on_before_batch_transfer(batch, dataloader_idx)
-        batch = self.transfer_batch_to_device(batch, device, dataloader_idx)
-        batch = self.on_after_batch_transfer(batch, dataloader_idx)
+        datahook_source = (
+            _DataHookSource(self, None) if self.trainer is None else self.trainer._data_connector._datahook_source
```
carmocca, line 260: Is it expected that `self.trainer` would ever be `None` here?

tests/trainer/connectors/test_data_connector.py
```diff
@@ -71,6 +74,115 @@ def test_dataloader_source_request_from_module():
     module.foo.assert_called_once()
 
 
+@pytest.mark.parametrize(
+    "hook_name", ("on_before_batch_transfer", "transfer_batch_to_device", "on_after_batch_transfer")
+)
+@mock.patch("pytorch_lightning.strategies.strategy.Strategy.lightning_module", new_callable=PropertyMock)
+def test_shared_datahook_call(module_mock, hook_name):
```
carmocca, line 81: I think it would be simpler and cleaner to unittest the dataclass itself.

There's many other tests that already check that the hooks get called.

pytorch_lightning/core/lightning.py
rohitgr7: if its None, then it would select `_DataHookSoure(self, None)`.

tests/trainer/connectors/test_data_connector.py
rohitgr7: done!

pytorch_lightning/core/lightning.py
carmocca: I meant if it can happen that this method (`_apply_batch_transfer_handler`) can be called and the trainer reference not be set. Meaning, should it be

```python
assert self.trainer is not None
datahook_source = self.trainer._data_connector._datahook_source
```

(or a ValueError instead of an assertion if it's not just a mypy issue)

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -555,3 +555,48 @@ def is_module(self) -> bool:
...
+
+    Arguments:
+        model: A LightningModule
+        datamodule: A LightningDataModule
+    """
+
+    model: "pl.LightningModule"
+    datamodule: Optional["pl.LightningDataModule"]
+    _valid_hooks: List[str] = field(
+        default=("on_before_batch_transfer", "transfer_batch_to_device", "on_after_batch_transfer")
```
carmocca, line 577: nit: the annotation says list but this is a tuple

tests/trainer/connectors/test_data_connector.py
```diff
@@ -71,6 +72,84 @@ def test_dataloader_source_request_from_module():
...
+
+        warning_cache.clear()
+        assert hook == getattr(model, hook_name)
+
+    _no_datamodule_no_overridden()
+    _with_datamodule_no_overridden()
+    _override_model_hook()
+    _override_datamodule_hook()
+    _override_both_model_and_datamodule()
+    _with_datamodule_override_model()
```
carmocca, line 144: Why not write 6 separate tests instead of 1 that calls 6 local functions?

Or a testing class: https://docs.pytest.org/en/6.2.x/getting-started.html#group-multiple-tests-in-a-class

pytorch_lightning/core/lightning.py
rohitgr7: some tests were failing since this is being used within ModelSummary, `to_onnx`, `to_torchscript` where we don't need a trainer every time.

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -555,3 +555,48 @@ def is_module(self) -> bool:
...
+                f"`{hook_name}` is not a shared hook within `LightningModule` and `LightningDataModule`."
+                f" Valid hooks are {self._valid_hooks}."
+            )
+
+        if self.datamodule is None:
+            return getattr(self.model, hook_name)
+
+        if is_overridden(hook_name, self.datamodule):
+            if is_overridden(hook_name, self.model):
+                warning_cache.warn(
```
awaelchli, line 592: I don't like this. I may not be the author of these implementations. I'm just reusing existing code from others and I can't really influence this. A warning sends me the signal that I could be doing something wrong.

awaelchli: Not sure, I could be contradicting a previous reviewer here.

rohitgr7: but isn't that a good thing? If I am importing someone else's LM where all the training/eval logic is defined and not injecting my own LDM after overriding these hooks again, it should warn users that the old training logic rely on some additional logic as well defined under these hooks, now datamodule has overridden it so the old one will no longer be used at all.

Also, this is done for BC but now with a warning.
https://github.com/PyTorchLightning/pytorch-lightning/blob/1fa0639bcaf4a5ab058178b284c52ce5c13c1ca3/pytorch_lightning/trainer/connectors/data_connector.py#L204-L207

do you think we should raise an error now?

awaelchli (approved):

Issue #10431, opened 2021-11-09
[RFC] Fault Tolerant Improvement


tchaton:
## 🚀 Feature

PyTorch Lightning 1.5 introduces Fault Tolerant Training and some users are already reporting some issues with it.
This was expected as the supported use case are still limited and the users are trying to feature hoping it would work for their own use case.

To improve the experience, here are 3 proposed steps:
- [] Implement a Fault-Tolerant Supported use case validator which could raise an expectation if it is not supported with the dataloader of the user

- [] Implement a manual mode where users would be responsible to implement their own logic. Lightning would be responsible to collect the state and restore them.

- [] Create advanced lectures on Fault-Tolerant Training demonstrating how to implement fault-tolerant for real-world use and how to test it out. Lightning could provide a utility to ensure fault-tolerant training is properly implemented.
   
### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

PR #10605, opened 2021-11-18
Fault Tolerant: move signal to SIGTERM

tchaton:
## What does this PR do?

This PR move the signal triggered from SIGUSR1 to SIGTERM and adds a test example for the community to validate fault tolerant training is properly working for them.

This is useful for any scheduler builder which wants to create e2e tests with Lightning.

Parts of #10431

Fault Tolerant relies on `signal.SIGTERM` to gracefully exit instead of `signal.SIGUSR1`

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @akihironitta @carmocca @ninginthecloud


pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -102,3 +102,12 @@ def _has_already_handler(self, signum: Signals) -> bool:
...
             signal.signal(signum, handlers)
+
+    @staticmethod
+    def _reset_signal(signum: Signals) -> None:
+        if isinstance(signal.getsignal(signum), HandlersCompose):
+            signal.signal(signum, signal.SIG_DFL)
+
+    def teardown(self) -> None:
+        self._reset_signal(signal.SIGTERM)
+        self._reset_signal(signal.SIGUSR1)
```
awaelchli, line 113: this part is conflicting with  #10611

awaelchli: is this needed for the goal of this PR ("move signal to SIGTERM")?

tests/utilities/fault_tolerant.py
```diff
@@ -0,0 +1,101 @@
...
+if not auto_restart_checkpoint_path_exists:
+    checkpoint_path = os.path.join(tmpdir, ".pl_auto_save.ckpt")
+    assert os.path.exists(checkpoint_path)
+    print(".pl_auto_save.ckpt exists.")
+
+if auto_restart_checkpoint_path_exists:
+    print([w for w in weights])
+
+os.environ.clear()
+os.environ.update(env_backup)
```
awaelchli, line 101: this code all runs on import. this is probably not intentional

awaelchli (changes requested):

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: No, let's use yours instead.

tchaton: I will remove this code.

tests/utilities/fault_tolerant.py
tchaton: This is intended. This code is used to validate Fault Tolerant Training works and I was thinking it could be exposed and used by organisation to validate their auto restart is working as expected.

awaelchli: But the docs you added say it is meant to be run as a script.

tests/utilities/fault_tolerant.py
```diff
@@ -0,0 +1,143 @@
...
+        super().__init__()
+        self.layer = torch.nn.Linear(1, 2)
+        self.seen_batches = []
+        self.fail_on_step = fail_on_step
+
+    def training_step(self, batch, batch_idx):
+        if self.global_step == self.fail_on_step:
+            log.info("READY TO BE KILLED WITH SIGTERM SIGNAL.")
+            while not self.trainer._terminate_gracefully:
+                sleep(0.00001)
```
nmiculinic, line 55:
```suggestion
                sleep(0.1)
```

This seems like it'd burn many CPUs cycles for no good reason

tests/utilities/fault_tolerant.py
tchaton: I have updated the docs.

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Done !

tests/utilities/fault_tolerant.py
awaelchli: I insist. I see no other way. As you can see, the test suite never starts, because it hangs at collecting the tests. You have this pice of code:
```py
 log.info("READY TO BE KILLED WITH SIGTERM SIGNAL.")
            while not self.trainer._terminate_gracefully:
                sleep(0.1)
```
which stalls execution indefinitely.
This is because the code runs at import!! I cannot see how this is intentional. Please consider adding if `if __name__ == "__main__"` check or let me know why this is not possible.

tchaton: Done !

pytorch_lightning/trainer/trainer.py
```diff
@@ -685,6 +685,9 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
             self.on_exception(exception)
             # shutdown workers
             self._data_connector.teardown()
+            if isinstance(exception, ExitGracefullyException):
+                # this is required to ensure schedulers can properly restart the training.
+                return
```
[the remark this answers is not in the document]
tchaton: Good question. I am still a bit ensure of this right now but here is my current reasoning.

Schedulers will be able to reschedule the training (TorchElastic, Slurm, etc...) only it the process finishes with exit 0.

Therefore, I see 2 options right there.

* 1: return and let the script continue. The user using Fault Tolerant Training is responsible to ensure his script would exit with 0, but might want to perform some additional operation outside the trainer method execution 
* 2: call sys.exit(0) to exit without failure as it was artificially created and everything was successfully completed.

tests/utilities/fault_tolerant.py
```diff
@@ -0,0 +1,142 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Here are the steps to use this file to ensure Lightning is properly setup for Fault Tolerant Training.
+
+1. Launch this script with `python tests/utilities/fault_tolerant.py`
+2. Once ``READY TO BE KILLED WITH SIGTERM SIGNAL`` is detected within
+   the logs, send a SIGTERM to this process. The process is waiting for it.
+3. You should detect `.pl_auto_save.ckpt exists` within the logs.
```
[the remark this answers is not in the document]
tchaton: Done !

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
tchaton: Hey @awaelchli,

After reflection, I have applied the second option.

tests/trainer/connectors/test_signal_connector.py
```diff
@@ -27,7 +27,7 @@
 from tests.helpers.runif import RunIf
 
 
-@pytest.mark.parametrize("register_handler", [False, True])
+@pytest.mark.parametrize("register_handler", [True])
```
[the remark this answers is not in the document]
tchaton: Reverted back.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2099,6 +2101,12 @@ def _exit_gracefully_on_signal(self) -> None:
             class_name = caller[0].f_locals["self"].__class__.__name__
             raise ExitGracefullyException(f"Exiting gracefully on {class_name}:{caller.function}")
 
+    @staticmethod
+    def _on_exit_gracefully_exception(exception: Exception) -> None:
+        if isinstance(exception, ExitGracefullyException):
+            # this is required to ensure schedulers can properly restart the training.
+            os._exit(0)
```
awaelchli, line 2108: minor suggestion:
- move this code into `_on_exception` method above
- in the test, mock.patch("pytorch_lightning.trainer.trainer.os._exit")  (it will be a no-op)
- optionally assert that `os._exit` was called

if it does not work, ignore my comment

pytorch_lightning/trainer/trainer.py
```diff
@@ -1544,13 +1544,15 @@ def _log_device_info(self) -> None:
                 " `Trainer(ipus=8)` or script `--ipus=8`."
             )
 
-    def _on_exception(self):
+    def _on_exception(self, exception):
```
awaelchli, line 1547: would be great to add the typing on that occasion :)

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -36,7 +36,7 @@ def register_signal_handlers(self) -> None:
         sigterm_handlers: List[Callable] = []
 
         if _fault_tolerant_training():
-            sigusr1_handlers.append(self.fault_tolerant_sigusr1_handler_fn)
+            sigterm_handlers.append(self.fault_tolerant_sigterm_handler_fn)
```
awaelchli, line 39: you missed updating the actually SIGUSR1, which I believe was the indention of this PR.
remember to also update all references to "SIGUSR1" in this file, and there are few others in other files in Lightning.

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
awaelchli: 2) seems reasonable, however I believe you accidentally reverted some of the changes to SIGUSR1 (see my comment below)

pytorch_lightning/trainer/trainer.py
tchaton: Donee !

pytorch_lightning/trainer/trainer.py
tchaton: Done !

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Hey @awaelchli,

I am not sure to fully understand your comment. `sigusr1` is still use for Slurm and I believe all reference left are SLURM related.

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
awaelchli: nvm SIGUSR1 is still used for SLURM, i missed that

awaelchli (approved):

tests/trainer/connectors/test_signal_connector.py
```diff
@@ -37,29 +38,32 @@ def test_fault_tolerant_sig_handler(register_handler, terminate_gracefully, tmpd
...
-    with mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": str(int(terminate_gracefully))}):
+    with mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": str(int(terminate_gracefully))}), mock.patch(
+        "os._exit"
+    ):
 
         trainer = Trainer(default_root_dir=tmpdir, max_epochs=1, limit_train_batches=2, limit_val_batches=0)
         if terminate_gracefully and not register_handler:
             with pytest.raises(ExitGracefullyException):
                 trainer.fit(model)
+            assert os._exit._mock_call_args == call(0)
```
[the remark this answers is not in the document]
tchaton: Done !

tests/utilities/test_auto_restart.py
```diff
@@ -879,13 +879,13 @@ def configure_optimizers(self):
         return torch.optim.SGD(self.layer.parameters(), lr=0.1)
 
 
-def _run_training(trainer_kwargs, dataset_classes, fail_on_step: int = -1, ckpt_path=None):
+def _run_training(trainer_kwargs, dataset_classes, fail_on_step: int = -1, ckpt_path=None, model_cls=TestModel):
```
carmocca, line 882: If unused, let's not add the argument

pytorch_lightning/trainer/trainer.py
```diff
@@ -1550,13 +1550,17 @@ def _log_device_info(self) -> None:
...
+    def _on_exception(self, exception: Exception) -> None:
         if not _fault_tolerant_training():
             return
         # save a checkpoint for fault tolerant training. we don't use `log_dir` to minimize the chances of failure.
         file_path = os.path.join(self.default_root_dir, ".pl_auto_save.ckpt")
         self.save_checkpoint(file_path)
 
+        # this is required to ensure schedulers can properly restart the run.
+        if isinstance(exception, ExitGracefullyException):
+            os._exit(0)
```
carmocca, line 1562: why not `raise SystemExit(0)` or `exit(0)`?

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -83,7 +83,8 @@ def slurm_sigusr1_handler_fn(self, signum: Signals, frame: FrameType) -> None:
             if self.trainer.logger:
                 self.trainer.logger.finalize("finished")
 
-    def fault_tolerant_sigusr1_handler_fn(self, signum: Signals, frame: FrameType) -> None:
+    def fault_tolerant_sigterm_handler_fn(self, signum: Signals, frame: FrameType) -> None:
+        log.info(f"Received signal {signum} saving fault tolerant checkpoint and terminating.")
```
carmocca, line 87:
```suggestion
        log.info(f"Received signal {signum}. Saving a fault-tolerant checkpoint and terminating.")
```

tests/utilities/fault_tolerant.py
```diff
@@ -0,0 +1,142 @@
+# Copyright The PyTorch Lightning team.
```
carmocca, line 1: Our CI will not test this, right? Why should we have it here then?

tests/utilities/fault_tolerant.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if not auto_restart_checkpoint_path_exists:
+        checkpoint_path = os.path.join(tmpdir, ".pl_auto_save.ckpt")
+        assert os.path.exists(checkpoint_path)
+        log.info(".pl_auto_save.ckpt exists.")
+
+    if auto_restart_checkpoint_path_exists:
+        params = [w for w in weights]
+        log.info(params)
+        assert "[-1.0939, -0.4306]" in str(params[-1])
```
carmocca, line 135: This could change with different torch versions.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2102,13 +2106,13 @@ def _results(self) -> Optional[ResultCollection]:
     def _exit_gracefully_on_signal(self) -> None:
         if not _fault_tolerant_training():
             return
-        if not self._should_terminated_gracefully():
+        if not self._should_terminate_gracefully():
```
carmocca, line 2109:
```suggestion
        if not _fault_tolerant_training() or not self._should_terminate_gracefully():
```

carmocca: > No, you don't want to force a reduce when fault tolerant is not active. This would be costly for the users who don't want to opt-in.

If fault tolerant training is not active, `not _fault_tolerant_training() == True` and the right hand side is not evaluated

tests/utilities/test_auto_restart.py
tchaton: Done !

pytorch_lightning/trainer/connectors/signal_connector.py
tchaton: Done !

pytorch_lightning/trainer/trainer.py
tchaton: I shared it before, we don't want to merge this. This would impact performance for user which don't opt-in on fault tolerant training.

pytorch_lightning/trainer/trainer.py
carmocca: Or even better, we can have `ExitGracefullyException` subclass `SystemExit`

tests/utilities/fault_tolerant.py
tchaton: This is used by Grid to validate its fault tolerant CI works as expected with Lightning.
And this script should be used by any other platform to provide a seamless experience with Fault Tolerant Training and Lightning.

pytorch_lightning/trainer/trainer.py
carmocca: See my last comment

```python
def _fault_tolerant_training():
    return False


def _should_terminate_gracefully():
    assert False


def foo():
    if not _fault_tolerant_training() or not _should_terminate_gracefully():
        print("Just return")
        return
    print("Doing stuff")

foo()
```

tests/utilities/fault_tolerant.py
carmocca: But how does Grid run it? Does it clone the repository just to access this file?

Why can't it be a `gist` then? It would be more convenient to update then.

tchaton: Grid clone Pytorch Lightning. I am still debuting it right now.

carmocca: If our CI does not run it, I would not add it here.

Normally, it would be in the Grid codebase, but if you expect other platforms to use it, then I'd publish it as a public gist.

This has the advantage of not having to get reviews + 3 approvals in Lightning for any change. If Grid CI breaks because this fails and the fix is in this script in particular, you want to be able to move fast.

pytorch_lightning/trainer/trainer.py
tchaton: Done !

pytorch_lightning/trainer/trainer.py
tchaton: Done !

tests/utilities/fault_tolerant.py
tchaton: Removed !

tests/utilities/test_auto_restart.py
```diff
@@ -1093,7 +1094,18 @@ def _fit_model(
         num_sanity_val_steps=0,
     )
 
-    trainer = Trainer(**trainer_kwargs)
+    class ExitGracefullyException(Exception):
+        pass
+
+    class TestTrainer(Trainer):
```
carmocca, line 1100: Awesome

tests/trainer/connectors/test_signal_connector.py
```diff
@@ -37,18 +37,20 @@ def test_fault_tolerant_sig_handler(register_handler, terminate_gracefully, tmpd
...
+                os.kill(os.getpid(), signal.SIGTERM)
                 sleep(0.1)
             return super().training_step(batch, batch_idx)
 
     model = TestModel()
 
-    with mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": str(int(terminate_gracefully))}):
+    with mock.patch.dict(os.environ, {"PL_FAULT_TOLERANT_TRAINING": str(int(terminate_gracefully))}), mock.patch(
+        "os._exit"
+    ):
```
carmocca, line 53: Do we need the exit mock still?

carmocca (approved):

tests/trainer/connectors/test_signal_connector.py
awaelchli: yes, it is supposed to prevent an actual call to the real exit, though theoretically, even though this works, one should set the target of the patch where os._exit is used (the signal connector)

carmocca: Tested locally and seems to be passing without it. It doesn't look to be necessary after the lastest commits

awaelchli: Then something is not working as intended. Based on the explanation of @tchaton, when exiting gracefully we trigger os._exit. During running the pytest, this should actually exit the test run but this is not what we want. Therefore the mock.

tchaton: I removed them !

tests/utilities/fault_tolerant.py
tchaton: Improve the code and moved it as an example.

tchaton: Done !

pl_examples/fault_tolerant/automatic.py
```diff
@@ -0,0 +1,128 @@
...
+    trainer = Trainer(default_root_dir=default_root_dir, max_epochs=max_epochs)
+    trainer.fit(model, train_dataloaders=train_dataloader, ckpt_path=ckpt_path)
+    return model.seen_batches, model.parameters()
+
+
+def main(args):
+    seed_everything(42)
+    os.environ["PL_FAULT_TOLERANT_TRAINING"] = "1"  # active fault tolerant automatic
+
+    ckpt_path = os.path.join(".", ".pl_auto_save.ckpt")
```
carmocca, line 106: Isnt this the same?

```suggestion
    ckpt_path = ".pl_auto_save.ckpt"
```

pl_examples/fault_tolerant/automatic.py
```diff
@@ -0,0 +1,128 @@
...
+    if not auto_restart_ckpt_path_exists and args.should_fail:
+        assert os.path.exists(ckpt_path)
+
+    if auto_restart_ckpt_path_exists or not args.should_fail:
+        log.info([w for w in weights])
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser(description="Fault Tolerant Example")
+    parser.add_argument("--should_fail", type=int, default=1, help="Whether the training should fail.")
```
carmocca, line 127: I think a `--fail` boolean would look nicer

```suggestion
    parser.add_argument("--fail", action="store_true", help="Whether the training should fail.")
```

pl_examples/fault_tolerant/automatic.py
```diff
@@ -0,0 +1,128 @@
...
+        p = torch.tensor([python_random.random() for _ in range(self.size)])
+        sample = (index + (t + n + p) / 10).float()
+        return sample
+
+    def __len__(self):
+        return self.len
+
+
+def _run_training(default_root_dir=".", max_epochs=3, fail_on_step: int = -1, ckpt_path=None):
+    seed_everything(1)
```
carmocca, line 94: `main` seeds already

pl_examples/fault_tolerant/automatic.py
```diff
@@ -0,0 +1,128 @@
...
+    train_dataloader = DataLoader(RandomGetItemDataset(3, 1))
```
carmocca, line 95: nit: put this on the lightningmodule?

pl_examples/fault_tolerant/automatic.py
tchaton: Done !

pl_examples/fault_tolerant/automatic.py
tchaton: Done !

pl_examples/fault_tolerant/automatic.py
tchaton: Done !

pl_examples/fault_tolerant/automatic.py
tchaton: Done !

pl_examples/fault_tolerant/automatic.py
```diff
@@ -0,0 +1,129 @@
...
+    2. Run this command within another terminal.
+        - You should see `Received signal 15. Saving a fault-tolerant checkpoint and terminating.` in the logs.
+    3. Launch `python pl_examples/fault_tolerant/automatic.py --emulate_failure` again.
+        - You should see `Restored all states from the checkpoint file at ./.pl_auto_save.ckpt`
+        - And you should see `[-1.0939, -0.4306]` in the logs.
+
+    To restart the process, just run `rm .pl_auto_save.ckpt` to delete the auto restart checkpoint.
+
+This means the weights with the failure matches the weight without and
+the training has been properly resumed and is fully reproduced.
```
SeanNaren, line 35:
```suggestion
This example shows that the weights trained with failures matches the weight trained without failures,
 thus the training has been properly resumed whilst being fully reproducible.
```

SeanNaren (approved):

pl_examples/fault_tolerant/automatic.py
tchaton: Done !

PR #10297, opened 2021-11-01
LiteDataLoader wrapper improvements

awaelchli:
## What does this PR do?

Follow up to #10279
Related: #10329

Adds small improvements around the init wrapper in Lite.
- Minimal docs
- Clearer variable names
- Separation of tests

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

Part of #1 (it's a lie, this is just here to avoid noisy GitHub bot)

cc @borda @justusschock @awaelchli @akihironitta @carmocca

PR #10625, opened 2021-11-19
LiteDataLoader code improvements and docs

awaelchli:
## What does this PR do?

Part of #10297 (follow-up to #10279)
Related: #10329

Updates docs, fixes typing (a DataLoader is not an iterator). 

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @akihironitta @ninginthecloud @carmocca


pytorch_lightning/lite/wrappers.py
```diff
@@ -157,29 +157,27 @@ def _replace_dataloader_init_method() -> Generator:
...
         super().__init__()
-        self.__dict__.update(getattr(dataloader, "__dict__", {}))
+        self.__dict__.update(dataloader.__dict__)
         self._dataloader = dataloader
         self._device = device
 
-    def __len__(self) -> Union[int, float]:
-        if isinstance(self._dataloader, Sized):
-            return len(self._dataloader)
-        return float("inf")
```
awaelchli, line 177: returning inf here is a strong assumption. LiteDataLoader should wrap the underlying dataloader and there is currently no reason why we would change the behavior of ``len``.

pytorch_lightning/lite/wrappers.py
```diff
@@ -157,29 +157,27 @@ def _replace_dataloader_init_method() -> Generator:
...
 
         Args:
-            dataloader: The current dataloader to be used.
+            dataloader: The dataloader to wrap
             device: The device to which the data should be moved. By default the device is `None` and no data
                 transfers will be made (identical behavior as :class:`~torch.utils.data.DataLoader`).
         """
         super().__init__()
-        self.__dict__.update(getattr(dataloader, "__dict__", {}))
+        self.__dict__.update(dataloader.__dict__)
```
awaelchli, line 170: AFAIK, every python object has a dict (unless it is slotted)

tchaton (approved):
LGTM ! This code was added for people not using a DataLoader

pytorch_lightning/lite/wrappers.py
```diff
@@ -157,29 +157,27 @@ def _replace_dataloader_init_method() -> Generator:
...
-        if isinstance(self._dataloader, Sized):
-            return len(self._dataloader)
-        return float("inf")
-
     @property
     def device(self) -> Optional[torch.device]:
         return self._device
 
+    def __len__(self) -> int:
+        return len(self._dataloader)
```
kaushikb11, line 179: What if it doesn't have the `len` attribute? Shouldn't we handle it?

awaelchli: If the thing we wrap doesn't have a len defined, how would we be able to handle it? 

If the user called len(dataloader) in their PyTorch code and it didn't have the attribute (AttributeError) they should get the same error in Lite. 
If it doesn't work in pure PyTorch, it won't work in Lite!

pytorch_lightning/lite/wrappers.py
justusschock: But what's the issue with doing it the "old" way?

pytorch_lightning/lite/wrappers.py
```diff
@@ -157,29 +157,27 @@ def _replace_dataloader_init_method() -> Generator:
 
 
 class _LiteDataLoader:
-    def __init__(self, dataloader: Union[Iterable, DataLoader], device: Optional[torch.device] = None) -> None:
-        """The LiteDataLoader is an extension of an Iterator. It would move the data to the device automatically if
-        the device is specified.
+    def __init__(self, dataloader: DataLoader, device: Optional[torch.device] = None) -> None:
```
justusschock, line 160: That means raw iterables are not supported anymore?

pytorch_lightning/lite/wrappers.py
justusschock: The most important thing here is that unlike in the trainer, we don't access that attribute and therefore don't need to handle it.

awaelchli:
> LGTM ! This code was added for people not using a DataLoader

In order to provide a seamless experience with all accelerators and strategies, we currently assume in both Lightning and Lite only DataLoaders. 
If there is a desire to support non-DataLoader iterables, we will have to rethink a few components to make this work.

pytorch_lightning/lite/wrappers.py
```diff
@@ -157,29 +157,27 @@ def _replace_dataloader_init_method() -> Generator:
...
+        """The LiteDataLoader is a wrapper for the :class:`~torch.utils.data.DataLoader`. It moves the data to the
+        device automatically if the device is specified.
 
         Args:
-            dataloader: The current dataloader to be used.
+            dataloader: The dataloader to wrap
             device: The device to which the data should be moved. By default the device is `None` and no data
                 transfers will be made (identical behavior as :class:`~torch.utils.data.DataLoader`).
         """
         super().__init__()
```
justusschock, line 169: Should we maybe use `typing.cast(type(dataloader), self)` here so that user type checks still work? This could be done for models as well...

justusschock (approved):

justusschock:
@awaelchli 
> In order to provide a seamless experience with all accelerators and strategies, we currently assume in both Lightning and Lite only DataLoaders.

AFAIK this is not true for Lightning, where we explicitly have type checks against the data loader when replacing samplers etc. to make sure it also works with raw iterables (just not all features then)

pytorch_lightning/lite/wrappers.py
kaushikb11: You are right!

pytorch_lightning/lite/wrappers.py
awaelchli: The code is misleading, indicating that there is a possibility to hit the default value of {}. But it can't, right? 
So it's dead code essentially.

pytorch_lightning/lite/wrappers.py
awaelchli: I'm a typing noob so I can't judge if that's important or not. Will it have any implications on our typing?

awaelchli:
> AFAIK this is not true for Lightning, where we explicitly have type checks against the data loader when replacing samplers etc. to make sure it also works with raw iterables (just not all features then)

This claim has been made in the past but where do we see that in our tests, docs and examples? I am failing to create an example based on our boring model, as it will just throw an error.

pytorch_lightning/lite/wrappers.py
carmocca, line 169:
```suggestion
```

There's no subclass

carmocca (approved):
Thanks Adrian!

Would you mind also addressing the open reviews in #10279 ?

pytorch_lightning/lite/wrappers.py
justusschock: It basically sets `LiteDataLoader=Dataloader` from a typing perspective

awaelchli: I couldn't figure out how this is supposed to work. For mypy to recognize the cast, I need to assign it back to the object, but here 

`self = cast(...)` won't work as we can't change the self reference.

awaelchli: done in setup_dataloader directly now

Issue #10595, opened 2021-11-17
[RFC] Depreceate the move_metrics_to_cpu Trainer argument.


tchaton:
## 📚 Documentation

After investigating this community issue https://github.com/PyTorchLightning/pytorch-lightning/issues/10379, it seems the parameter `move_metrics_to_cpu` isn't working as expected with `ddp`.

Bug reason: The ResultCollection object storing the metrics is moved on the CPU [here](https://github.com/PyTorchLightning/pytorch-lightning/search?q=_results.cpu%28%29&type=) and on epoch end when performing compute, the tensors are being reduced and this would raise `RuntimeError: Tensors must be CUDA and dense`.
Note, this should also fail with `sync_dist` argument within the self.log method.

Background: 

Before: Logging prior to the ResultCollection object, tensor metrics used to have an o(num_batches_in_epoch) memory footprint as there were stored within a list and reduced on epoch ended. When the epoch had a very large number of batches, this would raise an OOM.

Now: The ResultCollection has a memory space of o(1) for tensor metrics therefore, the `move_metrics_to_cpu` argument isn't as impactful as before.

As believe there is 2 option forward:

Option 1 🚀: 

Make the ResultCollection of `move_metrics_to_cpu`. The ResultCollection would be responsible to move back and forth the ResultMetric on device before distributed collection and on CPU right after. 

Pros:
* This might be impactful for large Metric from TM but a user could alternatively do this manually within its LightningModule

Cons:
* Engineering heavy
* Perform drop.

Here is the pseudo-code for such solution. This would be spread across several code parts.

```py
class ResultCollection

    def __init__(self, ..., move_metrics_to_cpu):
        self.move_metrics_to_cpu = move_metrics_to_cpu

    def metrics(self, on_step = False):
        if on_step
            for result_metric in self.result_metrics:
                # move the metric back the device
                if self.move_metrics_to_cpu:
                    result_metric.to(self.device)
                ... = result_metric.compute() # perform distributed reduction
                # move the metric back to cpu
                if self.move_metrics_to_cpu:
                    result_metric.to("cpu")
```

Option 2 😋:

Depreciate and remove the section of code moving the ResultCollection to CPU. 

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

[...]

PR #10631, opened 2021-11-19
Fix `move_metrics_to_cpu` with evaluation

carmocca:
## What does this PR do?

Fixes the bug uncovered by the analysis here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10595#issuecomment-972490354

- The outputs are no longer moved to CPU with `move_metrics_to_cpu=True`. This is consistent with how the training loops work.
- The recursive `detach()` call has been removed, it's not necessary as gradients arent computed during evaluation.
- `move_metrics_to_cpu=True` now actually moves the metrics to CPU. This is consistent with how the training loops work.
 
- If the user relies on the outputs being on CPU in `{validation,test}_epoch_end`.
- If the user memory blows up due to this change. Keep in mind that the user is always able to do `return your_tensor_on_gpu.cpu()`.

## Before submitting

## PR review

cc @borda @carmocca @edward-io @ananthsub @awaelchli

### DIFF
--- a/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
@@ -24,7 +24,6 @@ from pytorch_lightning.loops.utilities import _update_dataloader_iter
 from pytorch_lightning.trainer.progress import BatchProgress
 from pytorch_lightning.utilities.auto_restart import MergedIteratorState, reload_dataloader_state_dict
 from pytorch_lightning.utilities.fetching import AbstractDataFetcher, DataFetcher
-from pytorch_lightning.utilities.memory import recursive_detach
 from pytorch_lightning.utilities.model_helpers import is_overridden
 from pytorch_lightning.utilities.types import EPOCH_OUTPUT, STEP_OUTPUT
 
@@ -134,10 +133,13 @@ class EvaluationEpochLoop(Loop):
         self.trainer.logger_connector.update_eval_step_metrics()
 
         # track epoch level outputs
-        if self._should_track_batch_outputs_for_epoch_end():
-            output = recursive_detach(output, to_cpu=self.trainer.move_metrics_to_cpu)
-            if output is not None:
-                self.outputs.append(output)
+        if self._should_track_batch_outputs_for_epoch_end() and output is not None:
+            self.outputs.append(output)
+
+        if self.trainer.move_metrics_to_cpu:
+            # the evaluation step output is not moved as they are not considered "metrics"
+            assert self.trainer._results is not None
+            self.trainer._results.cpu()
 
         if not self.batch_progress.is_last_batch:
             # if fault tolerant is enabled and process has been notified, exit.
--- a/tests/trainer/logging_/test_eval_loop_logging.py
+++ b/tests/trainer/logging_/test_eval_loop_logging.py
@@ -26,6 +26,7 @@ from pytorch_lightning.loggers import TensorBoardLogger
 from pytorch_lightning.trainer.connectors.logger_connector import LoggerConnector
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from tests.helpers import BoringModel, RandomDataset
+from tests.helpers.runif import RunIf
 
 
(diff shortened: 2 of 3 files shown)


tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -699,3 +699,27 @@ def test_filter_metrics_for_dataloader(kwargs, expected):
...
+
+        def validation_epoch_end(self, outputs):
+            # the step outputs were not moved
+            assert all(o.device == self.device for o in outputs), outputs
+            # but the logging results were
+            assert self.trainer.callback_metrics["foo"].device.type == "cpu"
+
+    model = TestModel()
+    trainer = Trainer(
+        default_root_dir=tmpdir, limit_val_batches=2, move_metrics_to_cpu=True, accelerator="auto", devices=1
```
rohitgr7, line 723: should we run this on just GPU? since it will pass on CPU anyway.

carmocca: Done

rohitgr7 (approved):

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -134,10 +133,13 @@ def advance(
...
-        if self._should_track_batch_outputs_for_epoch_end():
-            output = recursive_detach(output, to_cpu=self.trainer.move_metrics_to_cpu)
-            if output is not None:
-                self.outputs.append(output)
+        if self._should_track_batch_outputs_for_epoch_end() and output is not None:
+            self.outputs.append(output)
+
+        if self.trainer.move_metrics_to_cpu:
+            # the evaluation step output is not moved as they are not considered "metrics"
+            assert self.trainer._results is not None
```
justusschock, line 141: why do we assert this? If somebody doesn't log metrics during validation but during training only and still uses the flag, it would error out here. Why would this be expected?

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -134,10 +133,13 @@ def advance(
...
+            self.trainer._results.cpu()
```
tchaton, line 142: This would break if there is reduction performed on epoch end.

tchaton (approved):
LGTM !

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
rohitgr7: I think it's due to mypy.

carmocca: Exactly. `Trainer._results` is `Optional` because the predict loop does not define it, but all other loops do

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
carmocca: You mean as described in https://github.com/PyTorchLightning/pytorch-lightning/issues/10595 right?

It's something to address separately, this PR just makes it works exactly as for training

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
justusschock: wouldn't

```
if self.trainer._results is not None:
    self.trainer._results.cpu()
```

work?
Seems slightly cleaner than asserting, but as long as we are sure this is never triggered, I'm also fine with the current way

justusschock (approved):

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
carmocca: I'd prefer that if that was something that could happen, but it's not.

I use assertions because they are short, the other option would be.

```python
if self.trainer.move_metrics_to_cpu:
    if self.trainer._results is None:
        raise InternalError("This shouldn't have happened")
    self.trainer._results.cpu()
```

PR #10568, opened 2021-11-16
Add support for manual fault tolerant

tchaton:
## What does this PR do?

This PR adds support for manual fault tolerant training.

There is currently 5 different modes defined as follows:

```py
class FaultTolerantTrainingModes(LightningEnum):

    DISABLED = "0"
    AUTOMATIC = "1"
    MANUAL = "2"
    AUTOMATIC_ELASTIC = "3"
    MANUAL_ELASTIC = "4"
```

In order to prevent too many patching, this PR introduces a `_SingleProcessDataLoaderIterStateful` and  `_MultiProcessingDataLoaderIterStateful`.

In further PR, the code between automatic and manual would be further merged.

Parts of #10431

TODO:
- [x] Validate it works with 1 worker
- [x] Validate it works with 2 workers

To be merged after
* #10674

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @carmocca @justusschock @awaelchli

PR #10638, opened 2021-11-19
Fault Tolerant Manual: Add support for DDP

tchaton:
## What does this PR do?

This PR adds support for Fault Tolerant Manual with DDP.

Parts of #10431

To be merged after
* #10568

## Before submitting

## PR review

## Did you have fun?

cc @borda @carmocca @justusschock @awaelchli @ninginthecloud


rohitgr7 (approved):

justusschock (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2094,7 +2094,11 @@ def _results(self) -> Optional[ResultCollection]:
             return active_loop._results
 
     def _exit_gracefully_on_signal(self) -> None:
-        if _fault_tolerant_training() and self._terminate_gracefully:
+        if _fault_tolerant_training():
+            # the signal should be sent to rank 0
```
awaelchli, line 2098: are you sure? if the signal does not get sent to all processes, only some of them will enter this broadcast call and the rest will hang (they will never run the _exit_gracefully_on_signal method).

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -737,3 +739,15 @@ def _teardown_dataloader_get_iterators() -> None:
     if get_iterator:
         DataLoader._get_iterator = get_iterator
         del DataLoader._ori_get_iterator
+
+
+def _collect_states_on_rank_zero_over_collection(state_dict: Any, device: torch.device) -> Any:
+    """This utility collects the state across processes for a collection of state."""
+
+    def fn(state: Dict):
+        nonlocal device
```
awaelchli, line 748: this should not be necessary. the closure should bind the local variable

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -737,3 +739,15 @@ def _teardown_dataloader_get_iterators() -> None:
...
+        if state.get("state", None) is not None:
```
awaelchli, line 749:
```suggestion
        if "state" in state:
```
 or not?

awaelchli: I believe the function should document that it will only collect dictionaries with a key "state" in it.

pytorch_lightning/trainer/trainer.py
tchaton: This code works only if the signal is sent to global rank 0. I didn't manage to make it work otherwise. This would need more investigation.

pytorch_lightning/utilities/auto_restart.py
tchaton: removed !

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -174,11 +181,20 @@ def on_save_checkpoint(self) -> Dict:
...
     def on_load_checkpoint(self, state_dict: Dict) -> None:
         # cache the dataloader state dict until the dataloader objects are available
-        self._dataloader_state_dict = state_dict.get("dataloader_state_dict")
+        # dataset states are collected across all ranks
+        if _fault_tolerant_training():
+            dataloader_state_dict = state_dict.get("dataloader_state_dict", None)
+            if not dataloader_state_dict:
+                return
+            rank = torch.distributed.get_rank() if distributed_available() else 0
+            self._dataloader_state_dict = dataloader_state_dict[rank]
```
awaelchli, line 197: isn't it possible to do this before we call the `on_load_checkpoint` hook? I believe this shouldn't be only loop specific

tchaton: Yes, I was thinking about this too. And I believed the cleanest was there. 

If you do it within the base loop, the user won't be its full state_dict and this logic is hidden from the user.

awaelchli: I wouldn't do it in the base loop, but in the checkpoint connector on the full state dict, before we send it to the objects to restore. This line here

`self._dataloader_state_dict = dataloader_state_dict[rank]
`
tells me we can filter the dict already earlier before giving it to the objects. Isn't this better?

tchaton: I am afraid this would break the paradigm where state_dict should not being altered through connecting components such as the CheckpointConnector. I am personally not in favour of this and prefer to have the logic visible the reader.

This might also result in unwanted behaviour for users subclassing the loops and overriding those functions. How do you detect and automate the extraction of rank state dict in this case ?

@carmocca @ananthsub @justusschock Any thoughts ?

awaelchli: I think it's fine to have the loop call the collect function, but the rest should be made strategy agnostic as much as possible. For example, instead of `torch.distributed.get_rank()` use call Lightning to determine the global_rank.

pytorch_lightning/trainer/supporters.py
```diff
@@ -403,6 +404,10 @@ def create_loader_iters(dataloader: DataLoader, state_dict: Dict) -> Iterator:
             if isinstance(dataloader, CycleIterator):
                 dataloader = dataloader_to_iter_on.loader
 
+            # dataset states are collected across all ranks
+            rank = torch.distributed.get_rank() if distributed_available() else 0
```
awaelchli, line 408: alternatively, we could access Lightnings global rank variable, i.e., `_get_rank()` to make it more strategy agnostic, although granted it is not much cleaner.

pytorch_lightning/utilities/auto_restart.py
awaelchli: I believed you missed my comment here.
You also changed `_collect_states_on_rank_zero` so this code might be outdated.

pytorch_lightning/utilities/distributed.py
```diff
@@ -378,7 +378,8 @@ def init_dist_connection(
         )
 
 
-def _collect_states_on_rank_zero(state: Dict[str, Any], device: torch.device) -> Optional[Dict[int, Any]]:
+# TODO: Refactor with the Strategy Collectives once finalized.
+def _collect_states_on_rank_zero(state: Dict[str, Any]) -> Optional[Dict[int, Any]]:
```
awaelchli, line 382: I believe this cannot return None anymore?

awaelchli (approved):

pytorch_lightning/utilities/distributed.py
tchaton: Done !

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

pytorch_lightning/trainer/supporters.py
tchaton: Yes, I believe we should do this when we have a cleaner API.

Right now, I still want the code to be unittested on this own and we haven't tested fault tolerant on TPU, IPU, horovod yet, so I believe there might be more caviats.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -174,11 +179,18 @@ def on_save_checkpoint(self) -> Dict:
         state: Optional[MergedIteratorState] = getattr(self._data_fetcher.dataloader_iter, state_to_save, None)
         if state:
             state_dict["dataloader_state_dict"] = asdict(state)
+        state_dict["dataloader_state_dict"] = _collect_states_on_rank_zero_over_collection(
+            state_dict["dataloader_state_dict"]
+        )
```
carmocca, line 184: nit: access the dictionary just once
```suggestion
        if state:
            state = asdict(state)
        state = _collect_states_on_rank_zero_over_collection(state)
        state_dict["dataloader_state_dict"] = state
```

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -320,9 +321,14 @@ def on_save_checkpoint(self) -> Dict:
...
         ):
             return state_dict
+
         state_dict["dataloader_state_dict"] = self.trainer.train_dataloader.state_dict(
             has_completed=self._has_completed()
         )
+
+        state_dict["dataloader_state_dict"] = _collect_states_on_rank_zero_over_collection(
+            state_dict["dataloader_state_dict"]
+        )
```
carmocca, line 331: Same comment about avoiding 3 dictionary accesses

pytorch_lightning/trainer/trainer.py
```diff
@@ -2094,7 +2094,11 @@ def _results(self) -> Optional[ResultCollection]:
...
+            should_terminate_gracefully = self.training_type_plugin.broadcast(self._terminate_gracefully)
```
carmocca, line 2099: shouldn't this use `reduce_boolean_decision`?

tests/utilities/test_distributed.py
```diff
@@ -64,15 +64,14 @@ def foo():
 def _test_collect_states(rank, world_size):
     os.environ["MASTER_ADDR"] = "localhost"
 
+    torch.cuda.set_device(f"cuda:{rank}")
+
```
carmocca, line 68: Why is this required now?

carmocca (approved):

tests/utilities/test_distributed.py
tchaton: I believe there is a bug with `torch.distributed.broadcast_object_list` or it might be a design choice.

But [here](https://github.com/pytorch/pytorch/blob/master/torch/distributed/distributed_c10d.py#L1855) they rely on torch.cuda.current_device() instead of torch.distributed.get_rank() meaning both tensors are still on `cuda:0` and the broadcast would fail.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
tchaton: Done !

pytorch_lightning/loops/epoch/training_epoch_loop.py
tchaton: Done !

pytorch_lightning/trainer/trainer.py
tchaton: Actually, no because the signal can be set to 1, many or all processes. Used reduce instead.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2094,14 +2094,17 @@ def _results(self) -> Optional[ResultCollection]:
...
-            raise ExitGracefullyException(f"Exiting gracefully on {class_name}:{caller.function}")
+        if not _fault_tolerant_training():
+            return
+        if not self._should_terminated_gracefully():
+            return
+        caller = inspect.stack()[1]
+        class_name = caller[0].f_locals["self"].__class__.__name__
+        raise ExitGracefullyException(f"Exiting gracefully on {class_name}:{caller.function}")
+
+    def _should_terminated_gracefully(self) -> bool:
```
carmocca, line 2105:
```suggestion
    def _should_terminate_gracefully(self) -> bool:
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2094,14 +2094,17 @@ def _results(self) -> Optional[ResultCollection]:
...
-            caller = inspect.stack()[1]
-            class_name = caller[0].f_locals["self"].__class__.__name__
-            raise ExitGracefullyException(f"Exiting gracefully on {class_name}:{caller.function}")
+        if not _fault_tolerant_training():
+            return
+        if not self._should_terminated_gracefully():
+            return
+        caller = inspect.stack()[1]
+        class_name = caller[0].f_locals["self"].__class__.__name__
+        raise ExitGracefullyException(f"Exiting gracefully on {class_name}:{caller.function}")
```
carmocca, line 2103: I strongly suggest not doing this, this is what stacktraces are for.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2094,10 +2094,17 @@ def _results(self) -> Optional[ResultCollection]:
...
 
     def _exit_gracefully_on_signal(self) -> None:
-        if _fault_tolerant_training() and self._terminate_gracefully:
-            caller = inspect.stack()[1]
-            class_name = caller[0].f_locals["self"].__class__.__name__
-            raise ExitGracefullyException(f"Exiting gracefully on {class_name}:{caller.function}")
+        if not _fault_tolerant_training():
+            return
+        if not self._should_terminated_gracefully():
+            return
```
carmocca, line 2100:
```suggestion
        if not _fault_tolerant_training() or not self._should_terminated_gracefully():
            return
```

pytorch_lightning/trainer/trainer.py
tchaton: Done !

pytorch_lightning/trainer/trainer.py
tchaton: No, you don't want to force a reduce when fault tolerant is not active. This would be costly for the users who don't want to opt-in.

PR #10668, opened 2021-11-22
docs: remove import of datasets separately since unused

shabie:
It is a simple PR that removes an unnecessary import of `datasets` from `torchvision` which is not used anywhere in the guide.


tchaton (approved):
LGTM !

carmocca (approved):

awaelchli (approved):

PR #10674, opened 2021-11-22
Fault Tolerant Manual: Add stateful dataloader iter

tchaton:
## What does this PR do?

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/utilities/auto_restart.py
```diff
@@ -441,7 +449,9 @@ def _find_current_worker(iterator: Iterator) -> Dict[str, Optional[int]]:
     return {"num_workers": num_workers, "previous_worker": previous_worker}
 
 
-def _capture_metadata_collate(samples: List, dataset: Dataset, default_collate: Callable) -> Dict:
+def _capture_metadata_collate(
+    samples: List, dataset: Dataset, default_collate: Callable, fault_tolerant_mode: _FaultTolerantMode
+) -> Dict:
```
justusschock, line 454: the return type is not accurate due to the early return path

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -441,7 +449,9 @@ def _find_current_worker(iterator: Iterator) -> Dict[str, Optional[int]]:
     return {"num_workers": num_workers, "previous_worker": previous_worker}
 
 
-def _capture_metadata_collate(samples: List, dataset: Dataset, default_collate: Callable) -> Dict:
+def _capture_metadata_collate(
+    samples: List, dataset: Dataset, default_collate: Callable, fault_tolerant_mode: _FaultTolerantMode
```
justusschock, line 453:
```suggestion
    samples: List, dataset: Dataset, collate: Callable, fault_tolerant_mode: _FaultTolerantMode
```

this is not necessarily the default collate

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -570,3 +602,126 @@ def reload_dataloader_state_dict(dataloader: DataLoader, state_dict: Dict[str, A
 
     else:
         raise MisconfigurationException("This shouldn't happen. Please, open an issue on PyTorch Lightning Github.")
+
+
+def _is_obj_stateful(obj: Any) -> bool:
+    """In order to be stateful, an object should implement a ``state_dict`` and ``load_state_dict`` method."""
```
justusschock, line 608: If you say it should implement both, we also should check for both methods

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -454,10 +461,26 @@ def _capture_metadata_collate(samples: List, dataset: Dataset, default_collate:
             "__pl_restart_meta": {"sampler_name0": state_dict0, "sampler_name1": state_dict1},
         }
     """
-    data = default_collate(samples)
-    if not isinstance(dataset, (CaptureIterableDataset, CaptureMapDataset)):
+    data = collate(samples)
+    fault_tolerant_mode
```
awaelchli, line 465: @tchaton accidentally placed?

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -580,3 +609,116 @@ def state_dict(self) -> Dict[str, Any]:
...
+                num_batches_fetched=self.num_batches_fetched,
+            )
+        ]
+        # ensures there is an alignement between the sampler state and currently fetched batch
+        assert sampler_state_idx == self.num_batches_fetched
+        self._data_fetcher._store_dataloader_iter_state(self, state)
+        return batch
+
+
+class _SingleProcessDataLoaderIterStateful(_StatefulMixin, _SingleProcessDataLoaderIter):
```
awaelchli, line 686: "StatefulMixin" is not a mixin if the implementation of this class relies on it. Currently, what I recognize here is multiple inheritance for the purpose of code reuse. So let's not call it a mixin?

tests/utilities/test_auto_restart.py
```diff
@@ -1237,3 +1241,109 @@ def test_fault_tolerant_mode_enum():
...
+
+    def state_dict(self):
+        self.counter += 1
+        return {"counter": self.counter}
+
+    def load_state_dict(self, state_dict):
+        self.counter = state_dict["counter"]
+
+
+class FailingStatefulRandomDataset(RandomDataset):
```
awaelchli, line 1258: what makes this failing?

tests/utilities/test_auto_restart.py
```diff
@@ -1237,3 +1241,109 @@ def test_fault_tolerant_mode_enum():
...
+
+
+class StatefulRandomDataset(FailingStatefulRandomDataset):
+    def state_dict(self):
+        info = get_worker_info()
+        worker_id = info.id if info else 0
+        return {worker_id: {"counter": self.counter}}
+
+
+@pytest.mark.parametrize("num_workers", [0, 2])
```
awaelchli, line 1281: will you exclude num_workers=2 before merging?

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

tests/utilities/test_auto_restart.py
tchaton: Cleaned up !

tests/utilities/test_auto_restart.py
tchaton: It takes 3 sec in total. I would prefer to include it as it covers the `_MultiProcessingDataLoaderIterStateful` functionalities.

awaelchli: I know, but unfortunately 3 secs are an eternity considering we have thousands of tests. Recently we have hit our CI timeout of 40 minutes! cc @carmocca

tchaton: Fine for now, but let's re-explore the others auto restart tests. Some are taking 20 secs.

tchaton: Done !

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -447,10 +455,25 @@ def _capture_metadata_collate(samples: List, dataset: Dataset, default_collate:
...
-    metadata = dataset.state_dict()
+    data = collate_fn(samples)
+    metadata = None
+    if fault_tolerant_mode.is_automatic:
+        metadata = dataset.state_dict()
+    else:
+        state_dict_fn = getattr(dataset, "state_dict", None)
+        info = get_worker_info()
+        worker_id = info.id if info else 0
+        if state_dict_fn:
```
ethanwharris, line 466: Better explicit here:
```suggestion
        if state_dict_fn is not None:
```

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -589,3 +621,116 @@ def state_dict(self) -> Dict[str, Any]:
...
+    def _reset(self, loader: DataLoader, first_iter: bool = False):
+        super()._reset(loader, first_iter=first_iter)
+        self._loader = loader
+        self.num_batches_fetched = 0
+
+    def __accumulate_state(self, sampler_state: Dict[str, Any]) -> None:
+        # initialize the queue if it doesn't exist.
+        if not hasattr(self, "_sampler_state"):
+            self._sampler_state = []
+            self._sampler_state_idx = 0
```
ethanwharris, line 638: Can be removed as this will be handled by the next line:
```suggestion
```

ethanwharris (approved):

pytorch_lightning/trainer/data_loading.py
```diff
@@ -441,9 +440,7 @@ def request_dataloader(
     def _add_sampler_metadata_collate(dataloader: DataLoader) -> None:
         """Wrap default collate function to retrive ``FastForwardSampler`` state dict when fault tolerant is
         enabled."""
-        dataloader.collate_fn = partial(
-            _capture_metadata_collate, dataset=dataloader.dataset, default_collate=dataloader.collate_fn
-        )
+        _add_capture_metadata_collate(dataloader)
```
carmocca, line 443: If `_add_sampler_metadata_collate` only calls `_add_capture_metadata_collate`, shouldn't all occurrences of the former be replaced with the latter?

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -480,6 +503,9 @@ def patch_dataloader_iterator(
     will extract the current iteration as part of the metadata returned by a custom batch.
     """
 
+    if not _FaultTolerantMode.detect_current_mode().is_automatic:
+        return
```
carmocca, line 507: Shouldn't this be checked wherever this gets called?

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

pytorch_lightning/utilities/auto_restart.py
tchaton: Done !

pytorch_lightning/trainer/data_loading.py
tchaton: Yes, entirely right !

pytorch_lightning/utilities/auto_restart.py
tchaton: Not sure to fully understand your question. Yes, it needs to be checked every time it gets called. But this code will be refactored soon.

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -447,10 +455,25 @@ def _capture_metadata_collate(samples: List, dataset: Dataset, default_collate:
...
+            if worker_id not in metadata:
+                if info and info.num_workers > 1:
+                    raise MisconfigurationException(
+                        f"The state_dict returned by {dataset} needs to be indexed by `worker_id` integer keys."
+                    )
+                metadata = {0: metadata}
+        if metadata is None:
+            metadata = {worker_id: {}}
+
     return {"data": data, AutoRestartBatchKeys.PL_RESTART_META: metadata}
```
justusschock, line 477: I know this doesn't come from this PR, but still related: for every batch we also get the corresponding state, right? Have you seen how large they are? Asking because they will also be transferred through shared memory. Does it slow down the process a lot? Probably not if you have some io-heavy stuff in your dataset, but what if returning preloaded data?

pytorch_lightning/utilities/auto_restart.py
```diff
@@ -589,3 +621,106 @@ def state_dict(self) -> Dict[str, Any]:
...
+def _get_iterator(self) -> "_BaseDataLoaderIter":
+    if not hasattr(self, "_lightning_fetcher"):
+        raise MisconfigurationException(
+            "A stateful iterator should be used only when a DataFetcher has been attached to the DataLoader."
+        )
+    if self.num_workers == 0:
+        return _SingleProcessDataLoaderIterStateful(self)
+    else:
+        if hasattr(self, "check_worker_number_rationality"):
+            self.check_worker_number_rationality()
```
justusschock, line 709: What does this do and where does it come from? Is this the warning we print if people don't use the correct number of workers? If yes, why is it here? Don't we already do that somewhere else?

justusschock (approved):
LGTM!

One general and one minor question.

pytorch_lightning/utilities/auto_restart.py
tchaton: Yes, with the BoringModel, we get a 30 % slowdown. But with Resnet50 on Cifar10, it is close to 0. @awaelchli performed the benchmark.

pytorch_lightning/utilities/auto_restart.py
tchaton: This is coming from the DataLoader `get_iterators`. In 1.10, `check_worker_number_rationality` was added, but doesn't exist on all versions. As I am patching their function, I had to still support this functionality.

SeanNaren (approved):
Looks good. imo we should offer a mixin class to add `state_dict` an `load_state_dict` for samplers/datasets, primarily out of good practice. But this can come at a later date :)

PR #10680, opened 2021-11-22
Support re-instantiation for custom DataLoader in Lightning

awaelchli:
## What does this PR do?

Part of #10329

Adds support for re-instantiating custom DataLoaders in Lightning, a feature that was recently added to Lite first in  #10279. 

- The utilities in Lite were moved to general data utilities
- The place in Lightning where we call `*_dataloader` on the LightningModule or DataModule gets wrapped with a context manager.

Limitations:
The patching will only work if the DataLoader gets instantiated inside the `*_dataloader` methods.
The custom DataLoader needs to be compatible with the patching, i.e., it needs to allow Lightning to write the arguments to it's dict.

## Before submitting

## PR review

## Did you have fun?


carmocca:
> TODO: add test

I suggest we move the tests we have in Lite to Lightning

tests/trainer/test_data_loading.py
```diff
@@ -74,15 +71,12 @@ def on_test_start(self) -> None:
         def create_dataset(self):
             dataset = IndexedRandomDataset(32, 64)
             if self._mode == 1:
-                # this case will raise an error
-                return FailureCustomDataLoader(32, dataset)
```
awaelchli, line 78: Reviewer: I removed/renamed the original mode 1 here because what was expected to fail here is now supported.

pytorch_lightning/lite/wrappers.py
```diff
@@ -110,49 +106,6 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
         return output
 
 
-def _wrap_init(init: Callable) -> Callable:
```
awaelchli, line 113: Reviewer: These three functions moved to utilities/data.py and are now shared between Lightning and Lite

tests/utilities/test_data.py
```diff
@@ -112,3 +113,29 @@ def test_has_len_all_rank():
         assert not has_len_all_ranks(DataLoader(RandomDataset(0, 0)), trainer.training_type_plugin, model)
 
     assert has_len_all_ranks(DataLoader(RandomDataset(1, 1)), trainer.training_type_plugin, model)
+
+
+def test_replace_dataloader_init_method():
```
carmocca, line 118: Why not remove the test of the same name in `tests/lite/test_lite.py`?

This test here is a better unit-test for the functionality.

Additionally, no tests were added to Lite for functionality already existing in Lightning (accelerator connector parsing for example).

Tests should try to be orthogonal, and these 2 are not.

carmocca (approved):

pytorch_lightning/utilities/data.py
```diff
@@ -305,3 +308,46 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+
+    recurse(cls)
+    return subclasses
+
+
+@contextmanager
+def _replace_dataloader_init_method() -> Generator[None, None, None]:
+    """This context manager is used to add support for re-instantiation of custom (subclasses) of
+    :class:`~torch.utils.data.DataLoader`. It patches the ``__init__`` method."""
+    for subclass in _get_all_subclasses(DataLoader):
```
carmocca, line 347: we should call `_get_all_subclasses` once

pytorch_lightning/trainer/data_loading.py
```diff
@@ -430,7 +431,8 @@ def request_dataloader(
 
         hook = f"{stage.dataloader_prefix}_dataloader"
         self.call_hook("on_" + hook, pl_module=model)
-        dataloader = source.dataloader()
+        with _replace_dataloader_init_method():
```
tchaton, line 434: Add a comment ?

tchaton (approved):
LGTM !

tests/utilities/test_data.py
awaelchli: I simplified the test, but the test has to stay I'm pretty sure.

pytorch_lightning/trainer/data_loading.py
awaelchli: done :)

justusschock (approved):

rusty1s:
@awaelchli This is really cool. This comes in handy with our custom data loaders in PyG. Thank you!

One issue I ran into is the use of "nested inheritance" of data loaders:
   ```
   torch.utils.data.DataLoader
          |
   MyBaseDataLoader
          |
   MyFinalLoader
   ```
This seems to do some weird things in the [`_replace_dataloader_init_method` context manager](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/utilities/data.py#L354-L365) since we overwrite the `__init__` function of both data loaders. I can try to set up a reproducible example later.

carmocca:
Hi @rusty1s! Glad that you find it useful, we actually worked on this with PyG as a motivation.

Happy to take a look once you define what those "weird things" are. You can find existing tests here if you are interested in poking at it yourself:

https://github.com/PyTorchLightning/pytorch-lightning/blob/1acb0dcbf4720b49b3de557d75f57a06c7276083/tests/utilities/test_data.py#L153-L176

Also, it would be better if you can open a new issue once you have a repro. Thank you!

PR #10682, opened 2021-11-22
Configure mypy to install dependencies in CI and update pyproject.toml

daniellepintz:
## What does this PR do?

* Update `code-checks.yml` to install all dependencies - fixes #9751
* Fix some of the 36 type errors revealed
* Update `pyproject.toml` for the files that still have type errors

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #10743, opened 2021-11-24
Improve typing for Lite

awaelchli:
## What does this PR do?

Follow up to #10682
#10682 will remove Lite from mypy checked typing, but we need to keep it there. 

Part of #7037 

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/lite/wrappers.py
```diff
@@ -131,6 +131,7 @@ def __iter__(self) -> Union[Iterator[Any], Generator[Any, None, None]]:
         iterator = iter(self._dataloader)
         if self._device is None:
             yield from iterator
+            return None
```
carmocca, line 134: was having no return a bug? or not because the iterator would've been exhausted in the loop below?

carmocca (approved):

pytorch_lightning/lite/wrappers.py
awaelchli: Not a bug: 
The iterator is exhausted but mypy does not know that. It pretends that self.device can be None after the if statement.

carmocca:
```suggestion
            return
```

None is not required here

awaelchli: this is mypy business, always have to be explicit

carmocca: That's not true. I ran it locally and the everything is green.

Mypy only requires it when there's another return

See how only `fn2` has an error here: https://mypy-play.net/?mypy=latest&python=3.10&gist=6bd880ade68f34c5ff86f47662022302

kaushikb11 (approved):

rohitgr7 (approved):

tchaton (approved):
LGTM !

justusschock (approved):

pyproject.toml
```diff
@@ -63,8 +63,6 @@ module = [
     "pytorch_lightning.core.mixins.hparams_mixin",
```
daniellepintz, line 63: you can delete the comment in line 39 "# Changes mypy default to ignore all errors" as it is no longer relevant

Issue #10702, opened 2021-11-23
ModuleNotFoundError: No module named 'rich.progress_bar'


Vozf:
## 🐛 Bug

Updated pl from 1.4.9 to 1.5.2 and got this
```
Traceback (most recent call last):
File "train.py", line 9, in <module>
from pytorch_lightning import Trainer, seed_everything
File ".venv/lib/python3.8/site-packages/pytorch_lightning/__init__.py", line 20, in <module>
from pytorch_lightning.callbacks import Callback  # noqa: E402
File ".venv/lib/python3.8/site-packages/pytorch_lightning/callbacks/__init__.py", line 25, in <module>
from pytorch_lightning.callbacks.progress import ProgressBar, ProgressBarBase, RichProgressBar, TQDMProgressBar
File ".venv/lib/python3.8/site-packages/pytorch_lightning/callbacks/progress/__init__.py", line 23, in <module>
from pytorch_lightning.callbacks.progress.rich_progress import RichProgressBar  # noqa: F401
File ".venv/lib/python3.8/site-packages/pytorch_lightning/callbacks/progress/rich_progress.py", line 26, in <module>
from rich.progress_bar import ProgressBar
ModuleNotFoundError: No module named 'rich.progress_bar'
```
P.s. import rich works, but rich.progress_bar - no

### Expected behavior

You can also fill out the list below manually.
-->

kaushikb11:
@Vozf What version of rich are you using?

Vozf:
I'm not using it at all
It seems like it was installed as pl subdependency 
8.0.0

kaushikb11:
Could you use the latest version of rich and test it out?

Vozf:
Works fine after update to 10.14, but still It should be handled in setup.py and not by manual user installation

Runkli:
Would this be resolved just by adding 
`rich>=10.14` to the requirements.txt file? If yes, I can quickly add it

Vozf:
I think so, but I can't be sure. I can test it on a branch or fork

awaelchli:
Let's do that @Runkli

kaushikb11:
@Runkli We have `rich>=10.2.2` in requirements/extras.txt.

We also check `_RICH_AVAILABLE = _module_available("rich") and _compare_version("rich", operator.ge, "10.2.2")` for the User before instantiating `RichProgressBar`

Vozf:
[...]

PR #10761, opened 2021-11-25
rich moved from extra to base requirements

Runkli:
## What does this PR do?

Related to issue https://github.com/PyTorchLightning/pytorch-lightning/issues/10702

This change moves rich from being an extra requirement to being a base requirement so it's installed upon setup. rich.progress_bar is used when training with PL, so it should be indicated in the requirements file.


tchaton:
Hey @Runkli,

I am not fully sure we want rich to become a default without fully supporting it across all our tests.

PR #10793, opened 2021-11-28
Integrate Intel Neural Compressor tool to quantize fp32 model

PenghuiCheng:
## What does this PR do?

This PR will integrate the Intel Neural Compressor tool to quantize an fp32 model. The example is in pl_examples/quantization. The usage is in README in this folder.

Below is The Resnet50 test result on AWS  c6i.16xlarge instance [Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz]
BS=1 Cores/instance=4 Instance=8

|	| fp32	| int8 |
|-----|----------|------------|
| accuracy	| 75.932	| 76.13 |
| throughput	| 184.98	| 542.60 |


awaelchli:
@PenghuiCheng Thanks for the contribution. Could you create a feature request issue too, so discussion can take place there first? Thanks a lot.

pl_examples/quantization/README.md
```diff
@@ -0,0 +1,86 @@
+Step-by-Step
+============
+
+This document describes the step-by-step instructions for reproducing PyTorch ResNet50/ResNet18/ResNet101 tuning results with Intel® Neural Compressor.
```
tchaton, line 4: Link to the library

pl_examples/quantization/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+# Saving and loading model:
+
+* Saving model:
+  After tuning with Neural Compressor, we can get neural_compressor.model:
+
+```
+quantizer = pl.IncQuantization('config/quantization.yaml', model, calib_dataloader=model.val_dataloader(),
+                                   metric_name="val_acc1")
```
tchaton, line 70:
```suggestion
quantizer = pl.IncQuantization(
       'config/quantization.yaml',
       model,
       calib_dataloader=model.val_dataloader(),
       metric_name="val_acc1"
)
```

pl_examples/quantization/README.md
```diff
@@ -0,0 +1,86 @@
...
+quantizer.fit_dynamic()
+quantizer.save("path/to/checkpoint")
+```
+
+
+* loading model:
```
tchaton, line 76:
```suggestion
* Create and quantize the model:
```

pl_examples/quantization/config/quantization.yaml
```diff
@@ -0,0 +1,18 @@
+version: 1.0
+
+model:                                               # mandatory.
+  name: imagenet
+  framework: pytorch_fx                                 # mandatory.
+
+device: cpu
+
+quantization:                                        # optional. tuning constraints on model-wise for advance user to reduce tuning space.
+  approach: post_training_static_quant
```
tchaton, line 10: More details on the configuration would be appreciated.

pl_examples/quantization/imagenet.py
```diff
@@ -0,0 +1,374 @@
...
+import torchvision.datasets as datasets
+import torchvision.models as models
+import torchvision.transforms as transforms
+
+import pytorch_lightning as pl
+from pl_examples import cli_lightning_logo
+from pytorch_lightning.core import LightningModule
+
+
+class ImageNetLightningModel(LightningModule):
```
tchaton, line 50: There is already an example with imagenet under `domain_templates`. Let's prevent code duplication.

pl_examples/quantization/imagenet.py
```diff
@@ -0,0 +1,374 @@
...
+        q_model = quantizer.fit_static()
+    elif quantizer.approach == pl.QuantizationMode.QAT.value:
+        q_model = quantizer.fit_aware_training()
+    else:
+        raise ValueError(f"Unknown quantization approach.")
+
+    quantizer.save(args.checkpoint)
+
+
+def validate(val_loader, model, args):
```
tchaton, line 267: This counter Lightning philosophy. You should use Trainer.validate to gather the metric and let's use TorchMetric.

pl_examples/quantization/imagenet.py
```diff
@@ -0,0 +1,374 @@
...
+
+    model = ImageNetLightningModel(**vars(args))
+    if args.evaluate:
+        if args.precision_mode == 'int8':
+            model = pl.compressor.intel.neural_compressor.load(model, args.checkpoint)
+        dataloader = model.val_dataloader()
+        validate(dataloader, model, args)
+        return
+
+    quantizer = pl.IncQuantization('config/quantization.yaml', model, calib_dataloader=model.val_dataloader(),
```
tchaton, line 246: This should be a callback. Find the [SparseML Integration from DeepSparse](https://github.com/PyTorchLightning/lightning-bolts/blob/master/pl_bolts/callbacks/sparseml.py)

pl_examples/quantization/imagenet.py
```diff
@@ -0,0 +1,374 @@
...
+        self.sum += val * n
+        self.count += n
+        self.avg = self.sum / self.count
+
+    def __str__(self):
+        fmtstr = '{name} {val' + self.fmt + '} ({avg' + self.fmt + '})'
+        return fmtstr.format(**self.__dict__)
+
+
+class ProgressMeter(object):
```
tchaton, line 337: Let's use exciting callbacks or create new ones.

pl_examples/quantization/imagenet.py
```diff
@@ -0,0 +1,374 @@
...
+        print('\t'.join(entries))
+
+    def _get_batch_fmtstr(self, num_batches):
+        num_digits = len(str(num_batches // 1))
+        fmt = '{:' + str(num_digits) + 'd}'
+        return '[' + fmt + '/' + fmt.format(num_batches) + ']'
+
+
+def run_cli():
+    parent_parser = ArgumentParser(add_help=False)
```
tchaton, line 355: Let's use the LightningCLI

pl_examples/quantization/requirements.txt
```diff
@@ -0,0 +1,4 @@
+neural-compressor
+--find-links https://download.pytorch.org/whl/torch_stable.html
+torch>=1.9.0+cpu
```
tchaton, line 3: Does this run only a CPU ?

pl_examples/quantization/imagenet.py
tchaton: Once the callback is done, let's have it along `QuantizationAwareTraining` in `pytorch_lightning/callbacks/quantization.py`

pyproject.toml
```diff
@@ -75,5 +75,7 @@ module = [
     "pytorch_lightning.utilities.parsing",
     "pytorch_lightning.utilities.seed",
     "pytorch_lightning.utilities.xla_device",
+    "pytorch_lightning.compressor.intel.neural_compressor.quantization",
+    "pytorch_lightning.compressor.intel.neural_compressor.utils",
```
tchaton, line 79: Let's move this under `pytorch_lightning/utilities/quantization.py` as a single file.

pytorch_lightning/__init__.py
```diff
@@ -16,9 +16,10 @@
 from pytorch_lightning.callbacks import Callback  # noqa: E402
 from pytorch_lightning.core import LightningDataModule, LightningModule  # noqa: E402
 from pytorch_lightning.trainer import Trainer  # noqa: E402
+from pytorch_lightning.compressor.intel.neural_compressor import IncQuantization, QuantizationMode  # noqa: E402
 from pytorch_lightning.utilities.seed import seed_everything  # noqa: E402
 
-__all__ = ["Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
+__all__ = ["Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything", "IncQuantization", "QuantizationMode"]
```
tchaton, line 22:
```suggestion
__all__ = ["Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
```

tchaton: Only core Lightning objects should be exposed.

pytorch_lightning/compressor/intel/neural_compressor/quantization.py
```diff
@@ -0,0 +1,119 @@
...
+            calib_dataloader (:obj:`DataLoader`, `optional`):
+                DataLoader for calibration.
+            metric_name(:obj:`str`):
+                metric name to symbolic trace to get FX GraphModule
+            trainer_args:
+                arguments for PyTorch Lightning trainer
+        Returns:
+            quantizer: IncQuantization object.
+        """
+        from neural_compressor.conf.config import Quantization_Conf
```
tchaton, line 38: This should fail with a good error message if neural_compressor isn't installed.

pytorch_lightning/compressor/intel/neural_compressor/quantization.py
```diff
@@ -0,0 +1,119 @@
...
+import pytorch_lightning as pl
+
+
+class QuantizationMode(Enum):
+    PTQ_DYNAMIC = "post_training_dynamic_quant"
+    PTQ_STATIC = "post_training_static_quant"
+    QAT = "quant_aware_training"
+
+
+class IncQuantization:
```
tchaton, line 14: This should be refactored into a callback. We won't merge objects wrapping a Trainer. The Trainer should be the entry point for anything else.

tchaton (changes requested):

pl_examples/README.md
```diff
@@ -51,3 +51,6 @@ Contains implementations leveraging [loop customization](https://pytorch-lightni
 
 - [K-fold Cross Validation Loop](./loop_examples/kfold.py): Implemenation of cross validation in a loop and special datamodule.
 - [Yield Loop](./loop_examples/yielding_training_step.py): Enables yielding from the training_step like in a Python generator. Useful for automatic optimization with multiple optimizers.
+## quantization
```
vkuzo, line 54: Can we call this something like "Intel quantization" instead of "quantization" everywhere, to reduce confusion between this workflow and PyTorch's native quantization workflow?

pl_examples/quantization/README.md
```diff
@@ -0,0 +1,86 @@
+Step-by-Step
```
vkuzo, line 1: can we make sure the folder is called "intel_quantization" and something similar, not "quantization"?

vkuzo (changes requested):

pl_examples/quantization/README.md
tchaton: I would prefer not, this would be generally confusing to the user as we currently have 3 ways to perform quantization and this folder shouldn't contain them 3 ideally.

vkuzo: I'd expect the word "quantization" in PyTorch first party repos to refer to PyTorch's native quantization solution.  Since this PR is proposing integrating a 3p tool with similar functionality, it would be good to be clear in the naming.

hshen14: Thanks for your comments @tchaton @vkuzo. WIP to enable INC (for Intel Neural Compressor) Callback to support various quantization approaches e.g., post-training static quantization, post-training dynamic quantization, quantization-aware training, etc. Regarding Intel quantization w/ INC vs. PyTorch native quantization, Intel quantization follows the native quantization API/flow to allow the users to migrate from PyTorch native to Intel quantization easily and provides additional value to maximize the productivity powered by automatic tuning.

pl_examples/intel_compressor/README.md
```diff
@@ -0,0 +1,82 @@
+# Step-by-Step
```
tchaton, line 1: Let's call it `pl_examples/quantization/intell_neural_compressor/...`

pl_examples/intel_compressor/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### 1. ResNet50
+
+```shell
+python imagenet.py --arch resnet50 --pretrained --data-path /path/to/imagenet --batch-size 300 --default_root_dir ./checkpoint
+```
+
+> **Note**
+>
+> batch size should be the munber which is sampling size divisible by. sampling size is a setting in config/quantization.yaml
```
tchaton, line 32:
```suggestion
> batch size should be the number which is sampling size divisible by. sampling size is a setting in config/quantization.yaml
```

pl_examples/intel_compressor/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### 1. Installation
+
+```shell
+pip install -r requirements.txt
+```
+
+### 2. Prepare Dataset
+
+Download [ImageNet](http://www.image-net.org/) Raw image to dir: /path/to/imagenet.  The dir include below folder:
```
tchaton, line 15:
```suggestion
Download [ImageNet](http://www.image-net.org/) raw images to the following directory: /path/to/imagenet.  The directory should include a train and val folders:
```

pl_examples/intel_compressor/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+# Saving and loading model:
+
+- Saving model:
+  After tuning with Neural Compressor, we can get neural_compressor.model:
+
+```
+quantizer = pl.IncQuantization(
+        'config/quantization.yaml',
+        metric_name="val_acc1",
```
tchaton, line 66:
```suggestion
        monitor="val_acc1",
```

We use monitor in our components, let's keep it consistent.

pl_examples/intel_compressor/README.md
```diff
@@ -0,0 +1,82 @@
...
+        save_path=args.default_root_dir)
```
tchaton, line 67:
```suggestion
        dirpath=args.default_root_dir)
```

Same as above.

pl_examples/intel_compressor/README.md
```diff
@@ -0,0 +1,82 @@
+# Step-by-Step
+
+This document describes the step-by-step instructions for reproducing PyTorch ResNet50/ResNet18/ResNet101 tuning results with [Intel® Neural Compressor](https://github.com/intel/neural-compressor).
+
+# Prerequisite
+
+### 1. Installation
+
+```shell
+pip install -r requirements.txt
```
tchaton, line 10: I would add and make intel neural compressor installation explicit here.

pl_examples/intel_compressor/config/quantization.yaml
```diff
@@ -0,0 +1,20 @@
+version: 1.0
+
+model:
+  name: imagenet
+  framework: pytorch_fx                              # supported values are pytorch, pytorch_fx, pytorch_ipex...and so on.
+
+device: cpu
+
+quantization:
```
tchaton, line 9: In the readme, more notes on each quantization method would be appreciated.

pl_examples/intel_compressor/config/quantization.yaml
```diff
@@ -0,0 +1,20 @@
...
+  approach: post_training_static_quant               # supported values are post_training_static_quant, post_training_dynamic_quant, quant_aware_training.
+  calibration:
+    sampling_size: 300
+
+tuning:
+  accuracy_criterion:
```
tchaton, line 15: This is quite confusing. My metric might not be accuracy and yet it is used as the decisive criterion to stop tuning. Might be worth to reconsider a better terminoligy.

pl_examples/README.md
```diff
@@ -51,3 +51,7 @@ Contains implementations leveraging [loop customization](https://pytorch-lightni
 
 - [K-fold Cross Validation Loop](./loop_examples/kfold.py): Implemenation of cross validation in a loop and special datamodule.
 - [Yield Loop](./loop_examples/yielding_training_step.py): Enables yielding from the training_step like in a Python generator. Useful for automatic optimization with multiple optimizers.
+
+## Intel Compressor
```
tchaton, line 55:
```suggestion
## Quantization
```

pl_examples/README.md
```diff
@@ -51,3 +51,7 @@ Contains implementations leveraging [loop customization](https://pytorch-lightni
...
+
+This folder contains quantization Image Classifier example with Intel Neural Compressor tool
```
tchaton, line 57:
```suggestion
- [Intel Neural Compressor] This folder contains quantization Image Classifier example with Intel Neural Compressor tool.
```

Adds links to documentation.

pl_examples/intel_compressor/imagenet.py
```diff
@@ -0,0 +1,88 @@
...
+    parent_parser = pl.Trainer.add_argparse_args(parent_parser)
+    parent_parser.add_argument("--data-path", metavar="DIR", type=str, help="path to dataset")
+    parent_parser.add_argument("--seed", type=int, default=42, help="seed for initializing training.")
+    parent_parser.add_argument(
+        "-e", "--evaluate", dest="evaluate", action="store_true", help="evaluate model on validation set"
+    )
+    parser = ImageNetLightningModel.add_model_specific_args(parent_parser)
+    parser.set_defaults(profiler="simple", deterministic=True, max_epochs=90)
+    args = parser.parse_args()
+    main(args)
```
tchaton, line 83: Let's use the LightningCLI. Sync with @carmocca @mauvilsa if you are experiencing issues.

pytorch_lightning/callbacks/__init__.py
```diff
@@ -48,6 +48,7 @@
     "ProgressBar",
     "ProgressBarBase",
     "QuantizationAwareTraining",
+    "INCQuantization",
```
tchaton, line 51: Let's keep this list alphabetically ordered.

pytorch_lightning/core/hooks.py
```diff
@@ -304,6 +304,12 @@ def configure_sharded_model(self) -> None:
         implementation of this hook is idempotent.
         """
 
+    def on_compress_model(self) -> None:
```
tchaton, line 307: Let's rely on the existing hooks. Take an example from https://github.com/tchaton/SparseMLCallback.

It seems the `on_compress_model` could be replaced with `on_fit_end`.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1075,6 +1077,91 @@ def tune(
 
         return result
 
+    def compressor(
```
tchaton, line 1080: Let's remove the compressor method, we don't plan to add any new methods to the trainer.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+        self.save_path = save_path
+
+    def on_compress_model(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        """Called when the compressor loop begins."""
+        try:
+            from neural_compressor.experimental import common, Quantization
+        except ImportError as e:
+            raise RuntimeError("Please install neural-compressor with: pip install neural-compressor") from e
+
+        def eval_func(model):
```
tchaton, line 375: Let's make them static private methods on the `INCQuantization` and not local ones.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+            return out[0][self.metric_name]
+
+        def train_func(model):
+            pl_module_tmp = copy.deepcopy(pl_module)
+            pl_module_tmp.model = model
+            trainer.fit(pl_module_tmp)
+            trainer.state.fn = TrainerFn.COMPRESSING
+            return pl_module_tmp.model
+
+        assert hasattr(
```
tchaton, line 389: I would make this more flexible. Either provide the module name from the init arguments or try to infer this automatically.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+            pl_module, "model"
+        ), "The LightningModule should have the backbone model which is a native PyTorch model!"
+        quantizer = Quantization(self.config)
+        quantizer.model = common.Model(pl_module.model)
+
+        if self.approach == QuantizationMode.PTQ_STATIC.value:
+            quantizer.calib_dataloader = pl_module.train_dataloader()
```
tchaton, line 396: This doesn't seem very pythonic. Any better API to setup the quantizer ?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+            trainer.state.fn = TrainerFn.COMPRESSING
+            return pl_module_tmp.model
+
+        assert hasattr(
+            pl_module, "model"
+        ), "The LightningModule should have the backbone model which is a native PyTorch model!"
+        quantizer = Quantization(self.config)
+        quantizer.model = common.Model(pl_module.model)
+
+        if self.approach == QuantizationMode.PTQ_STATIC.value:
```
tchaton, line 395: Does your team plan to add more methods there ?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+
+    def on_compress_model(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        """Called when the compressor loop begins."""
+        try:
+            from neural_compressor.experimental import common, Quantization
+        except ImportError as e:
+            raise RuntimeError("Please install neural-compressor with: pip install neural-compressor") from e
+
+        def eval_func(model):
+            pl_module_tmp = copy.deepcopy(pl_module)
```
tchaton, line 376: This is usually a bad idea to deepcopy the model as it might trigger an OOM. Let's remove it.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+            pl_module_tmp.model = model
+            out = trainer.validate(pl_module_tmp)
```
tchaton, line 378: This won't play nice when called within fit.

tchaton: @carmocca @awaelchli Any ideas ?

pytorch_lightning/callbacks/quantization.py
hshen14: No, these are all three quantization approaches. We plan to add pruning approaches in the near future, but those will be put in pruning.py probably.

pl_examples/intel_compressor/README.md
PenghuiCheng: OK， will change it

pl_examples/intel_compressor/imagenet.py
PenghuiCheng: Hi,@carmocca, Could you tell me where is the LightningCLI example? 
I just refer pl_examples/domain_templates/imagenet.py here.

pytorch_lightning/core/hooks.py
PenghuiCheng: @tchaton , thanks for your comments. 
We use on_compress_model hook for two reasons:
1. On_fit_end is for train loop, but static and dynamic quantization didn't need to train, only need to evaluate. 
2. for quantization aware training, we need to tune the quantized model in INC tool, INC will generate quantization configure by specified strategy, and train the model with the lightning trainer, then INC will check the evaluation metric with the lightning trainer, if the metric matches the target, then stop turning，otherwise，will tune again. The whole process is completed in INC tool, so I think the fit processing does not match the requirement.

pytorch_lightning/trainer/trainer.py
PenghuiCheng: @tchaton ， the reasons is the same as above.
In INC tool, the tuning process used trainer.fit, trainer.validate and so on. if we use trainer.fit to quantize the model with INC, there is conflict when we use the early stop callback hook.

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: I consider that if INC tune failure(such as can't find a quantized model match the target), then it will destroy the original model.

pytorch_lightning/callbacks/quantization.py
tchaton: Won't it be better to have a single callback to perform both quantization and pruning? This would simplify Lightning users onboarding.

pytorch_lightning/core/hooks.py
tchaton: If I understand properly your requirements, you should be able to provide those functionalities by relying on more hooks such as hooks `on_train_end`, `on_validation_end`, and `on_test_end` which could share the code to perform the quantization.

By utilizing the `on_epoch_end` hook should be able to check the validation metrics and decide to stop tuning.

Am I missing something? 

Furthermore, if you believe a hook should be added, please open an issue with a clear explanation on why this is required.

pytorch_lightning/trainer/trainer.py
tchaton: We don't plan to add a new method to the Trainer. Therefore, we would either have to use more hooks within the Callbacks or add new hooks.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
     def on_predict_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         if "predict" in self._observer_disabled_stages:
             self._restore_last_observer_enabled()
+
+
+class INCQuantization(Callback):
```
tchaton, line 331: Let's add some tests for this component.

pytorch_lightning/callbacks/quantization.py
tchaton: In the extreme case, we might want to save a temporary checkpoint there. With large model, this would likely raise an OOM.

pl_examples/intel_compressor/imagenet.py
mauvilsa: @PenghuiCheng probably you should first have a quick look at the documentation https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_cli.html. Regarding examples, there are multiple which use LightningCLI, see the search in the repo https://github.com/PyTorchLightning/pytorch-lightning/search?l=Python&q=LightningCLI.

hshen14:
Hi @tchaton @vkuzo, to better address your review comments, please let me give a short summary of what is Intel Neural Compressor (INC) and how INC fits PyTorchLightning.

INC is an open-source python library providing various compression techniques (e.g., quantization, pruning, etc.) across mainstream deep learning frameworks to accelerate inference performance. This PR is for quantization so the context is around this. Typically, there are three quantization approaches (following the definition by PyTorch): 1) post-training static quantization, 2) post-training dynamic quantization, and 3) quantization-aware training. The first two approaches are training-free, while the last one requires few training. Moreover, INC adds the unique feature for post-training quantization with machine learning-driven automatic tuning (https://github.com/intel/neural-compressor#workflow) to make the model quantization faster.

With the above notes, I'd like to address some major comments first:
1. Callback. We read the code and considered your comments carefully, however, we think it's worthwhile to introduce a new API like "compress" or "train_free" (same level as "train" or "tune") to show up the easy quantization using post-training approach, which is complementary as quantization-aware training using "train".
2. Deepcopy. INC tuning process relies on the original FP32 model so we expect this PR would also provide this capability for effective tuning. So deepcopy is needed, unless we may want to sacrifice the accuracy (or other objectives/metrics) w/o tuning.

One more question: we were aware of quantization-aware training is already in PyTorchLightning using the first version of APIs provided by PyTorch. I am wondering whether you are aware of the new set of APIs based on FX which is becoming the official APIs (for quantization and others) in PyTorch, and whether you have the plan to migrate to the new APIs. An alternative solution might be using INC to unify the user-facing APIs for quantization in the three approaches aforementioned above. 

Please let us know your comments/suggestions. Thanks.

vkuzo:
> One more question: we were aware of quantization-aware training is already in PyTorchLightning using the first version of APIs provided by PyTorch. I am wondering whether you are aware of the new set of APIs based on FX which is becoming the official APIs (for quantization and others) in PyTorch, and whether you have the plan to migrate to the new APIs. An alternative solution might be using INC to unify the user-facing APIs for quantization in the three approaches aforementioned above.

hi @hshen14 , wondering if you could clarify what would be helpful to know about PyTorch quantization to help this PR?  I work on PyTorch quantization, so happy to share what I can (not sure yet what would be helpful).

tchaton:
Dear @hshen14,

Thanks for providing a better overview of the current method and a cleared definition around quantization.

Concerning your first point, PyTorch Lightning is extremely strict on new addition such as a new hook on the Callback and even more onto the Trainer. I would recommend you to open an RFC to open such discussion with the clear motivation for such addition, but IMO, this addition is quite opinionated and would be confronting our core values.

Furthermore, I believe the INCQuantization Callback and existing hook should be enough to fully support the functionality provided by the INC Quantizer.

I believe the core of the problem is coming from the current approach this PR is taking. By assigning the current `train_func` as the calibration method, this would trigger a fit within a fit for QAT.

I believe the right approach would be to fit the PyTorch Callback API by mapping your [internal hooks](https://github.com/intel/neural-compressor/blob/master/neural_compressor/experimental/component.py#L42) and data loaders transformation onto ours. Right now, it seems to me the current PR is forcing the Quantizer API (which is very similar to a callback) onto a Lightning Callback. 

Concerning your second, you are right. I am just raising it might raise OOM and some users might open issues once it happens. If you believe it is fine for now, I am good too.

Best,
T.C

hshen14:
Thanks for your additional comments/suggestions. I understood the design philosophy that you described and appreciated your thoughts on how to better integrate INC into PyTorchLightning. Just want to raise some points to help us get more aligned in this discussion:

* Post-training quantization & Tuning. As mentioned earlier, INC supports automatic tuning, which means it may require several iterations to generate a quantized model that meets the accuracy & performance requirement, where each iteration includes calibration, quantization, and evaluation. Back to your suggestion on "INCQuantization Callback and existing hook", it's true that we can continue following this way, however, my concern is the tuning (multiple-times) capability will be degenerate to one-time quantization. The new PR is trying to address this concern, so want to get your insights on whether one-time post-training quantization is okay for the users.

* Hook mapping. As you pointed out, INC defines several hooks that can be mapped to PyTorchLightning hooks probably. Please note that INC hooks are only for quantization-aware training (not for post-training quantization as described above). I believe it would be relatively easy to integrate quantization-aware training with well-mapped hooks than post-training quantization, but this could be addressed in the future PR.

Thanks, - Haihao

> Dear @hshen14,
> 
> Thanks for providing a better overview of the current method and a cleared definition around quantization.
> 
> Concerning your first point, PyTorch Lightning is extremely strict on new addition such as a new hook on the Callback and even more onto the Trainer. I would recommend you to open an RFC to open such discussion with the clear motivation for such addition, but IMO, this addition is quite opinionated and would be confronting our core values.
> 
> Furthermore, I believe the INCQuantization Callback and existing hook should be enough to fully support the functionality provided by the INC Quantizer.
> 
> I believe the core of the problem is coming from the current approach this PR is taking. By assigning the current `train_func` as the calibration method, this would force a fit within a fit and it is why you introduced a new trainer method.
> 
> However, I believe the right approach would be to rely on the Trainer fit, validate, test methods and the PyTorch Callback API by mapping your [internal hooks](https://github.com/intel/neural-compressor/blob/master/neural_compressor/experimental/component.py#L42) and data loaders transformation onto ours. Right now, it seems to me the current PR is forcing the Quantizer API (which is very similar to a callback with some Dataloaders manipulation) onto a single Lightning Callback hook. Instead, your hooks should map on yours and I believe it would work just fine.
> 
> Concerning your second, you are right. I am just raising it might raise OOM and some users might open issues once it happens. If you believe it is fine for now, I am good too.
> 
> Best, T.C

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: I think not all the models can be save to .pth file for pytorch FX graph.

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: Now it is in compress processing.

PenghuiCheng: What are your suggestion?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+        if self.approach == QuantizationMode.QAT.value:
+            quantizer.q_func = train_func
+        else:
+            trainer.post_quantizing = True
+
+        quantizer.eval_func = eval_func
+        model = quantizer()
+        pl_module.model = model.model
+        if self.save_path is not None:
+            model.save(self.save_path)
```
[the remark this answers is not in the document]
PenghuiCheng: This model is INC model and it will save quantized configure and weights. I think Trainer save_checkpoint can handle this case.

PenghuiCheng:
> 



> Thanks for your additional comments/suggestions. I understood the design philosophy that you described and appreciated your thoughts on how to better integrate INC into PyTorchLightning. Just want to raise some points to help us get more aligned in this discussion:
> 
> * Post-training quantization & Tuning. As mentioned earlier, INC supports automatic tuning, which means it may require several iterations to generate a quantized model that meets the accuracy & performance requirement, where each iteration includes calibration, quantization, and evaluation. Back to your suggestion on "INCQuantization Callback and existing hook", it's true that we can continue following this way, however, my concern is the tuning (multiple-times) capability will be degenerate to one-time quantization. The new PR is trying to address this concern, so want to get your insights on whether one-time post-training quantization is okay for the users.
> * Hook mapping. As you pointed out, INC defines several hooks that can be mapped to PyTorchLightning hooks probably. Please note that INC hooks are only for quantization-aware training (and other training-involved compressions, but not suitable for post-training quantization). I believe it would be relatively easy to integrate quantization-aware training (with the well-mapped hooks) than post-training quantization, but this could be addressed in future PR.
> 
> Thanks, - Haihao
> 
> > Dear @hshen14,
> > Thanks for providing a better overview of the current method and a cleared definition around quantization.
> > Concerning your first point, PyTorch Lightning is extremely strict on new addition such as a new hook on the Callback and even more onto the Trainer. I would recommend you to open an RFC to open such discussion with the clear motivation for such addition, but IMO, this addition is quite opinionated and would be confronting our core values.
> > Furthermore, I believe the INCQuantization Callback and existing hook should be enough to fully support the functionality provided by the INC Quantizer.
> > I believe the core of the problem is coming from the current approach this PR is taking. By assigning the current `train_func` as the calibration method, this would force a fit within a fit and it is why you introduced a new trainer method.
> > However, I believe the right approach would be to rely on the Trainer fit, validate, test methods and the PyTorch Callback API by mapping your [internal hooks](https://github.com/intel/neural-compressor/blob/master/neural_compressor/experimental/component.py#L42) and data loaders transformation onto ours. Right now, it seems to me the current PR is forcing the Quantizer API (which is very similar to a callback with some Dataloaders manipulation) onto a single Lightning Callback hook. Instead, your hooks should map on yours and I believe it would work just fine.
> > Concerning your second, you are right. I am just raising it might raise OOM and some users might open issues once it happens. If you believe it is fine for now, I am good too.
> > Best, T.C

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: The eval_func method is used by INC, so it can't be static private methods here.

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: Hi, @tchaton , Do you mean that we should add a new python file like INC_compress.py, and it will cover quantization and pruning with INC tool?

pl_examples/quantization/README.md
PenghuiCheng: now the folder is called "quantization/intel_neural_compressor"

pl_examples/intel_compressor/README.md
PenghuiCheng: changed

pl_examples/intel_compressor/README.md
PenghuiCheng: changed

pl_examples/intel_compressor/README.md
PenghuiCheng: changed

pl_examples/intel_compressor/README.md
PenghuiCheng: changed

pl_examples/intel_compressor/README.md
PenghuiCheng: Changed

pl_examples/intel_compressor/config/quantization.yaml
PenghuiCheng: done

pl_examples/README.md
PenghuiCheng: changed

pl_examples/intel_compressor/imagenet.py
PenghuiCheng: Done

pytorch_lightning/callbacks/__init__.py
PenghuiCheng: Done

pytorch_lightning/core/hooks.py
PenghuiCheng: Now we only integrate PTQ, so only use on_fit_start hook.

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: done

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: I removed deepcopy now.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+        assert hasattr(
+            pl_module, "model"
+        ), "The LightningModule should have the backbone model which is a native PyTorch model!"
+        quantizer = Quantization(self.config)
+        quantizer.model = common.Model(pl_module.model)
+
+        if self.approach == QuantizationMode.PTQ_STATIC.value:
+            quantizer.calib_dataloader = pl_module.train_dataloader()
+
+        if self.approach == QuantizationMode.QAT.value:
```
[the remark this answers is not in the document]
PenghuiCheng: done

pytorch_lightning/callbacks/quantization.py
```diff
@@ -318,3 +326,82 @@ def on_predict_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule
...
+            quantizer.q_func = train_func
+        else:
```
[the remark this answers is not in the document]
PenghuiCheng: done

hshen14:
> > One more question: we were aware of quantization-aware training is already in PyTorchLightning using the first version of APIs provided by PyTorch. I am wondering whether you are aware of the new set of APIs based on FX which is becoming the official APIs (for quantization and others) in PyTorch, and whether you have the plan to migrate to the new APIs. An alternative solution might be using INC to unify the user-facing APIs for quantization in the three approaches aforementioned above.
> 
> hi @hshen14 , wondering if you could clarify what would be helpful to know about PyTorch quantization to help this PR? I work on PyTorch quantization, so happy to share what I can (not sure yet what would be helpful).

Hi @vkuzo , thanks for asking. This PR provides ease of use quantization experience to generate a quantized model meeting accuracy faster on PyTorch powered by automatic tuning feature in Intel Neural Compressor. So getting familiar with PyTorch quantization will definitely help this PR move forward. As mentioned early, Intel Neural Compressor is built on top of the framework and leverages the framework quantization flow and APIs (PyTorch FX).  As you may know, PyTorch supports three quantization use cases (post-training static quantization, post-training dynamic quantization, and quantization-aware training) that are well supported by Intel Neural Compressor. Please let me know if I can provide more info. Thanks.

vkuzo:
> One more question: we were aware of quantization-aware training is already in PyTorchLightning using the first version of APIs provided by PyTorch. I am wondering whether you are aware of the new set of APIs based on FX which is becoming the official APIs (for quantization and others) in PyTorch, and whether you have the plan to migrate to the new APIs. An alternative solution might be using INC to unify the user-facing APIs for quantization in the three approaches aforementioned above.

I see, thank you, I think understand your question now.  Looks like here (https://github.com/PyTorchLightning/pytorch-lightning/blob/7a9a08c5d3ca4699fb439f691c40e1320b37507a/pytorch_lightning/callbacks/quantization.py), Lightning has an integration with Eager mode quantization, and you were asking if there are plans to also add an FX graph mode quantization integration?  Please correct me if I didn't get the question correctly.

If that is the question, here are my thoughts:
1. Integrating FX graph mode quantization into PyTorch Lightning makes sense strategically.
2. The PyTorch quantization team does not have any immediate plans to do so ourselves in the near future.  We would be happy to work with someone in the community to help make this happen, if someone is interested in taking this on.

Hope this helps!

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: done

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: A custom fit loop may help, but it could be a kind of duplication of the tuning capability already provided by Intel Neural Compressor (INC). As alternative, we suggest to keep the current integration flow which treats eval_func as the callback function for INC and reuses the existing tuning capability.

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: Done

pytorch_lightning/callbacks/quantization.py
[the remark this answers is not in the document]
PenghuiCheng: I am afraid that Trainer save_checkpoint is not sufficient, since a complete quantized model for deployment needs both quantized model/weight and quantization config file. That’s why we introduce the save method for now, but we are open to use any official save API in the future and commit for the change, as long as it can support well for INT8 model deployment.

pl_examples/quantization/imagenet.py
PenghuiCheng: dnoe

pl_examples/quantization/imagenet.py
PenghuiCheng: removed

pl_examples/quantization/imagenet.py
PenghuiCheng: callback is done

pl_examples/quantization/imagenet.py
PenghuiCheng: removed

pl_examples/quantization/imagenet.py
PenghuiCheng: done

pl_examples/quantization/requirements.txt
PenghuiCheng: yes, For now, INC focus on CPU and will support intel GPU in the future.

pyproject.toml
PenghuiCheng: Done

pytorch_lightning/__init__.py
PenghuiCheng: removed

pytorch_lightning/compressor/intel/neural_compressor/quantization.py
PenghuiCheng: done

pytorch_lightning/compressor/intel/neural_compressor/quantization.py
PenghuiCheng: done

pl_examples/README.md
PenghuiCheng: Done

pl_examples/intel_compressor/config/quantization.yaml
PenghuiCheng: @tchaton , How about "metric_criterion" or "metric"?

pl_examples/README.md
PenghuiCheng: Done

pytorch_lightning/trainer/trainer.py
PenghuiCheng: removed

pytorch_lightning/callbacks/quantization.py
PenghuiCheng: I changed quantizer.model = common.Model(pl_module.model) to quantizer.model = getattr(pl_module.model)

pl_examples/quantization/intel_neural_compressor/README.md
```diff
@@ -0,0 +1,118 @@
...
+### Saving and loading model:
+
+- Saving quantization configure and weights:
+  After tuning with Neural Compressor, we can get neural_compressor.model:
+
+```
+cli = MyLightningCLI(ImageNetLightningModel,
+                     seed_everything_default=42,
+                     save_config_overwrite=True, run=False)
+callback = pl.callbacks.INCQuantization(
```
tchaton, line 62: This can be added in the trainer_defaults.

pl_examples/quantization/intel_neural_compressor/imagenet.py
```diff
@@ -0,0 +1,65 @@
...
+            parser.add_argument(
+                "-e", "--evaluate", dest="evaluate", action="store_true", help="evaluate model on validation set"
+            )
+
+    cli = MyLightningCLI(ImageNetLightningModel, seed_everything_default=42, save_config_overwrite=True, run=False)
+    if cli.config["evaluate"]:
+        from neural_compressor.utils.pytorch import load
+
+        print("Load quantized configure from ", cli.trainer.default_root_dir)
+        cli.model.model = load(cli.trainer.default_root_dir, cli.model.model)
```
tchaton, line 49: This seems to be a hack.

tchaton:
> * Post-training quantization & Tuning. As mentioned earlier, INC supports automatic tuning, which means it may require several iterations to generate a quantized model that meets the accuracy & performance requirement, where each iteration includes calibration, quantization, and evaluation. 

Concerning QAT, I believe we should have a separation of concerns. The Callback should be running 1 iteration of the quantization procedure, and the Loops should be running the iteration to refine the performance requirement. IMO, this should come with a custom `fit_loop`. Have a look at the [KFold Custom Loop](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pl_examples/loop_examples/kfold.py#L278) 

Back to your suggestion on "INCQuantization Callback and existing hook", it's true that we can continue following this way, however, my concern is the tuning (multiple-times) capability will be degenerate to one-time quantization. The new PR is trying to address this concern, so want to get your insights on whether one-time post-training quantization is okay for the users.

Yes, I believe post-training quantization is entirely fine for the users as applied at the end of the training.

pytorch_lightning/callbacks/intel_compressor.py
```diff
@@ -0,0 +1,116 @@
...
+        self.monitor = monitor
+        assert (
+            module_name_to_quant is not None
+        ), "Please set module_name_to_quant with module name which you want to quantize"
+        self.module_name_to_quant = module_name_to_quant
+        self.datamodule = datamodule
+        self.dirpath = dirpath
+
+    def on_fit_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        """Called when the compressor loop begins."""
```
tchaton, line 80:
```suggestion
```

pytorch_lightning/callbacks/intel_compressor.py
```diff
@@ -0,0 +1,116 @@
...
+        try:
```
tchaton, line 81: Use our imports mechanism to check this. This is cleaner.

pytorch_lightning/callbacks/intel_compressor.py
```diff
@@ -0,0 +1,116 @@
...
+            quantizer.calib_dataloader = (
+                pl_module.train_dataloader() if self.datamodule is None else self.datamodule.train_dataloader()
+            )
+        elif self.approach != QuantizationMode.PTQ_DYNAMIC.value:
+            raise MisconfigurationException("Unsupport quantization approach:" + self.approach)
+
+        quantizer.eval_func = eval_func
+        model = quantizer()
+        setattr(pl_module, self.module_name_to_quant, model.model)
+        if self.dirpath is not None:
```
tchaton, line 109: This won't work with all strategy (e.g DeepSpeed). Rely on the trainer.save_checkpoint

pytorch_lightning/callbacks/intel_compressor.py
```diff
@@ -0,0 +1,116 @@
...
+            model.save(self.dirpath)
+
+        trainer.limit_train_batches = 0
```
tchaton, line 112: This looks like a hack.

pytorch_lightning/callbacks/intel_compressor.py
```diff
@@ -0,0 +1,116 @@
...
+            setattr(pl_module, self.module_name_to_quant, model)
+            out = trainer.validate(pl_module, self.datamodule.val_dataloader() if self.datamodule is not None else None)
+            return out[0][self.monitor]
+
+        trainer.quantizing = True
+
+        if self.dirpath is None:
+            self.dirpath = trainer.default_root_dir
+
+        quantizer = Quantization(self.config)
```
tchaton, line 96: As shared above, I believe the Callback should be subclassing `Quantization` or even its base primitive, but not be instantiated internally. This is a hack and won't work well with other callbacks.

pl_examples/quantization/intel_neural_compressor/imagenet.py
PenghuiCheng: You are right, will remove deepcopy in INC load function next version.

pytorch_lightning/callbacks/intel_compressor.py
kaushikb11: Agreed. @PenghuiCheng

kaushikb11: I will pushing some changes here.

pytorch_lightning/callbacks/intel_compressor.py
PenghuiCheng: done

pytorch_lightning/callbacks/intel_compressor.py
PenghuiCheng: done

pl_examples/quantization/intel_neural_compressor/README.md
PenghuiCheng: done

pytorch_lightning/callbacks/intel_compressor.py
```diff
@@ -0,0 +1,115 @@
...
+from pytorch_lightning.trainer.states import TrainerFn, TrainerStatus
+from pytorch_lightning.utilities import _NEURAL_COMPRESSOR_AVAILABLE
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+
+if _NEURAL_COMPRESSOR_AVAILABLE:
+    from neural_compressor.conf.config import Quantization_Conf
+    from neural_compressor.experimental import Quantization
+
+
+class QuantizationMode(Enum):
```
kaushikb11, line 33: @PenghuiCheng Why did you remove quantization aware training from here? Doesn't neural compressor have accuracy-driven tuning enhancement for it?

I have pinged you on Lightning Slack. I would like us to collaborate on this and land the PR by thinking of multiple approaches.

PenghuiCheng: @kaushikb11 , I'm glad to receive your suggestion, and let's collaborate on this PR. I removed the QAT approach because it seems INC processing is not matched PyTorch lightning's. So I removed QAT temporarily. And at the same time, we still continue to find ways to integrate methods for QAT. Your suggestions are very welcome. 
 BTW, How to use Lightning Slack? Could you give me a link?

pl_examples/quantization/intel_neural_compressor/imagenet.py
PenghuiCheng: @kaushikb11，we meet a problem that PyTorch quantization function for FX mode is not an in-place function([prepare_fx](https://github.com/pytorch/pytorch/blob/b4a75af75810237d58b53d0bb8e515d84929a897/torch/ao/quantization/quantize_fx.py#L337),...), so we can't modify the original model(cli.model.model) directly in load function. What is your suggestion? thanks very much!

pytorch_lightning/callbacks/intel_compressor.py
kaushikb11: @PenghuiCheng Hey, use this [link](https://join.slack.com/t/pytorch-lightning/shared_invite/zt-pw5v393p-qRaDgEk24~EjiZNBpSQFgQ) to join Lightning Slack. You could ping me at Kaushik B.

Borda:
@PenghuiCheng @kaushikb11, how is it going here? :otter:

carmocca:
Hi! Carlos from the core team here.

This is a very interesting integration, but we are not looking to merge it into PyTorch Lightning Core. Since it's implemented as a Callback, it should be fine to import it from a 3rd party library. This way you maintain control over it.

If you still want to take this direction, we are open to adding a reference inside our docs.

Closing otherwise. Thank you for your time.

hshen14:
Hi @carmocca , thanks for reaching. It would be super interesting to add a reference inside the doc. We will follow up.

PR #10802, opened 2021-11-29
Fix typing in `pl.callbacks.lr_monitor`

awaelchli:
## What does this PR do?

Part of #7037 

## Before submitting

## PR review

## Did you have fun?

cc @borda


pytorch_lightning/utilities/types.py
```diff
@@ -43,7 +56,30 @@
     Dict[str, Sequence[DataLoader]],
 ]
 EVAL_DATALOADERS = Union[DataLoader, Sequence[DataLoader]]
-# todo: improve LRSchedulerType naming/typing
+
+
+@runtime_checkable
+class _LRScheduler(Protocol):
+    optimizer: Optimizer
+
```
awaelchli, line 52: @carmocca 
I couldn't figure out what's the best way to tell mypy that LRScheduler has an attribute "optimizer". Is this the right way?

carmocca: Mmm I don't know for sure.

AFAIK the runtime checkable is for `isinstance` checks
And I don't know if it works with class attributes like that.

Ideally they would define it here:https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.pyi#L4-L10
Or you would extend that

awaelchli: updated as discussed

awaelchli: unfortunately, the changes I make here are messing with LightningCLI, not sure yet why

pytorch_lightning/utilities/types.py
```diff
@@ -43,7 +44,75 @@
...
 # todo: improve LRSchedulerType naming/typing
-LRSchedulerTypeTuple = (_LRScheduler, ReduceLROnPlateau)
-LRSchedulerTypeUnion = Union[_LRScheduler, ReduceLROnPlateau]
-LRSchedulerType = Union[Type[_LRScheduler], Type[ReduceLROnPlateau]]
+LRSchedulerTypeTuple = (torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau)
+LRSchedulerTypeUnion = Union[torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau]
+LRSchedulerType = Union[Type[torch.optim.lr_scheduler._LRScheduler], Type[torch.optim.lr_scheduler.ReduceLROnPlateau]]
+
+
+class LRSchedulerConfig(TypedDict):
```
carmocca, line 110: Since we have this now, we can remove `_get_default_scheduler_config`

pytorch_lightning/utilities/types.py
carmocca: Pushed a commit that fixes the CLI, but breaks mypy, we might need separate variables.

Feel free to revert

pytorch_lightning/utilities/types.py
```diff
@@ -43,7 +44,75 @@
     Dict[str, Sequence[DataLoader]],
 ]
 EVAL_DATALOADERS = Union[DataLoader, Sequence[DataLoader]]
+
+
+# Copied from `torch.optim.lr_scheduler.pyi`
+# Missing attributes were added to improve typing
+class _LRScheduler:
+    optimizer: Optimizer
```
tchaton, line 52: Should we open an issue on PyTorch too ?

pytorch_lightning/utilities/types.py
tchaton: This would be a BC for Flash.

pytorch_lightning/utilities/types.py
```diff
@@ -43,7 +44,75 @@
...
+    def get_lr(self) -> float:
+        ...
+
+    def step(self, epoch: Optional[int] = ...) -> None:
+        ...
+
+
+# Copied from `torch.optim.lr_scheduler.pyi`
+# Missing attributes were added to improve typing
+class ReduceLROnPlateau:
```
tchaton, line 75: Is it used ?

tchaton (approved):
LGTM !  Ideally, this would be done in PyTorch.

pytorch_lightning/utilities/types.py
awaelchli: @carmocca could you elaborate how it is now possible to remove it?

pytorch_lightning/utilities/types.py
awaelchli: yes, below in the LRSchedulerConfig. It is required because in the lr_monitor we access `scheduler_config["scheduler"].optimizer` and mypy needs to know about the optimizer attribute

rohitgr7 (approved):

pytorch_lightning/utilities/types.py
carmocca: We cannot set defaults for the `TypedDict`.

Ideally, this would be a dataclass, which then is used in our code. then we don't need the function to create the dict

awaelchli: Right now the code expects a dict everywhere. I think it would be best to leave this for a separate PR in which this can be updated across the code base if needed and provide a way to make it work with flash

Borda (approved):

pytorch_lightning/utilities/types.py
awaelchli: will look into it #11211

pytorch_lightning/utilities/types.py
```diff
@@ -43,7 +44,75 @@
...
+        ...
+
+
 # todo: improve LRSchedulerType naming/typing
-LRSchedulerTypeTuple = (_LRScheduler, ReduceLROnPlateau)
-LRSchedulerTypeUnion = Union[_LRScheduler, ReduceLROnPlateau]
-LRSchedulerType = Union[Type[_LRScheduler], Type[ReduceLROnPlateau]]
+LRSchedulerTypeTuple = (torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau)
+LRSchedulerTypeUnion = Union[torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau]
+LRSchedulerType = Union[Type[torch.optim.lr_scheduler._LRScheduler], Type[torch.optim.lr_scheduler.ReduceLROnPlateau]]
```
rohitgr7, line 107: @awaelchli quick ques: is there any reason why you didn't use the _LRScheduler and ReduceLROnPlateau types you created here?

awaelchli: something about CLI, I don't know. @carmocca knows

carmocca: Because these are used in the CLI and it needs the actual classes. This is because it uses types to extract the help messages and do input validation

PR #10803, opened 2021-11-29
Fix typing in `pl.trainer.config_validator`

awaelchli:
## What does this PR do?

Fixes

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/configuration_validator.py
+++ b/pytorch_lightning/trainer/configuration_validator.py
@@ -32,11 +32,11 @@ def verify_loop_configurations(trainer: "pl.Trainer", model: "pl.LightningModule
         __verify_train_val_loop_configuration(trainer, model)
         __verify_manual_optimization_support(trainer, model)
         __check_training_step_requires_dataloader_iter(model)
-    elif trainer.state.fn == TrainerFn.VALIDATING:
+    elif trainer.state.fn is TrainerFn.VALIDATING:
         __verify_eval_loop_configuration(trainer, model, "val")
-    elif trainer.state.fn == TrainerFn.TESTING:
+    elif trainer.state.fn is TrainerFn.TESTING:
         __verify_eval_loop_configuration(trainer, model, "test")
-    elif trainer.state.fn == TrainerFn.PREDICTING:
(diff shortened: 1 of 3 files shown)


pytorch_lightning/trainer/configuration_validator.py
```diff
@@ -32,11 +32,11 @@ def verify_loop_configurations(trainer: "pl.Trainer", model: "pl.LightningModule
         __verify_train_val_loop_configuration(trainer, model)
         __verify_manual_optimization_support(trainer, model)
         __check_training_step_requires_dataloader_iter(model)
-    elif trainer.state.fn == TrainerFn.VALIDATING:
+    elif trainer.state.fn is TrainerFn.VALIDATING:
```
carmocca, line 35: Even though Python recommends `is` to check Enums, since we use our own weird enums which act as strings, `==` is preferred over `is` because strings shouldn't be compared with the latter.

To fix the `mypy` error, we should instead

```
if trainer.state.fn is None:
    raise ValueError("Unexpected, should be set")
```

awaelchli: I checked already, it works fine. You can create a TrainerFn from a string and then compare it with "is". 

```py
from pytorch_lightning.trainer.states import TrainerFn
state = TrainerFn.FITTING
state is TrainerFn.FITTING
Out[4]: True
state = TrainerFn("fit")
state is TrainerFn.FITTING
Out[8]: True
state is TrainerFn.VALIDATING
Out[9]: False
```

carmocca: That assumes that `trainer.state.fn` will be a `TrainerFn` enum instance, but since we don't want the users to deal with the Enums, it is possible for an user to do:

```py
trainer.state.fn = "fit"
```

and then this would happen

```python
"fit" is TrainerFn.FITTING 
<string>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
Out[6]: False
```

tchaton (approved):
LGTM !

carmocca (approved):

tchaton (approved):
LGTM !

justusschock (approved):

PR #10805, opened 2021-11-29
Remove setting None to TensorBoardLogger warning

kaushikb11:
## What does this PR do?

Was added in #10777

The warning won't be necessary for majority of users.

And this warning pops up every time a user trains with `ddp_spawn`,` ddp_spawn_sharded` and `tpu_spawn` as TensorBoard logger is the default.

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda

PR #10838, opened 2021-11-30
Skip hanging spawn tests

carmocca:
## What does this PR do?

Since we haven't found a solution for all loggers, we've decided to skip these tests under the circumstances where they hang. This was attempted in #10826 

That is `torch<1.9` and `Python>=3.8`.

We don't expect this to be a problem for users as they won't be spawning as many processes together as we do when we run `pytest`.

Closes #10805

None

## Before submitting

## PR review

## Did you have fun?

Nope

cc @borda

### DIFF
--- a/tests/utilities/test_all_gather_grad.py
+++ b/tests/utilities/test_all_gather_grad.py
@@ -54,7 +54,7 @@ def _test_all_gather_ddp(rank, world_size):
     assert torch.allclose(grad2, tensor2.grad)
 
 
-@RunIf(skip_windows=True, skip_49370=True)
+@RunIf(skip_windows=True, skip_49370=True, skip_hanging_spawn=True)
 def test_all_gather_ddp_spawn():
     world_size = 3
     torch.multiprocessing.spawn(_test_all_gather_ddp, args=(world_size,), nprocs=world_size)
--- a/pytorch_lightning/plugins/training_type/ddp_spawn.py
+++ b/pytorch_lightning/plugins/training_type/ddp_spawn.py
@@ -25,7 +25,6 @@ from torch.nn import Module
 from torch.nn.parallel.distributed import DistributedDataParallel
 
 import pytorch_lightning as pl
-from pytorch_lightning.loggers import LoggerCollection, TensorBoardLogger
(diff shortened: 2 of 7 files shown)


awaelchli (approved):

justusschock (approved):

tests/utilities/test_all_gather_grad.py
```diff
@@ -54,7 +54,7 @@ def _test_all_gather_ddp(rank, world_size):
     assert torch.allclose(grad2, tensor2.grad)
 
 
-@RunIf(skip_windows=True, skip_49370=True)
+@RunIf(skip_windows=True, skip_49370=True, skip_hanging_spawn=True)
```
rohitgr7, line 57: why not add a min PyTorch and python requirements in the test with a comment instead of a separate flag?

Borda (approved):

awaelchli:
@carmocca should we add the pytest-timeout also to the regular CPU tests?

tests/utilities/test_all_gather_grad.py
carmocca: I feel like this is cleaner. Makes it easy to find the tests later

There's nothing guaranteeing that all the comments will be equal in the future if more tests are skipped. It also runs the risk of somebody forgetting one of the 2.

And we can add and remove code from tests no problem, so BC is not a concern.

carmocca:
> should we add the pytest-timeout also to the regular CPU tests?

We totally can, but we haven't seen any hangs there at the moment. The current failures we have are caused by the MacOs tests actually hitting the global timeout because they are slow.

awaelchli:
Ok, I commented this only because I saw them timeout and feared that skipping the tests had global impact on others.

carmocca:
The "Slow CI" addition might alleviate that since the worker tests are particularly slow on MacOS.

tests/utilities/test_all_gather_grad.py
rohitgr7: if we skip more tests with this flag, then we also cannot guarantee that they will fall under the same version constraints.

but these are tests, not for users.. so its fine I guess.

carmocca:
Finally 😓

PR #10839, opened 2021-11-30
Raise exception if rich is less than 10.2.2

kaushikb11:
## What does this PR do?

Raise an exception if `rich` version is less than `10.2.2`, or else it's very hard for the user to know the error.
Discussed in #10702 

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -227,10 +228,15 @@ def __init__(
         leave: bool = False,
         theme: RichProgressBarTheme = RichProgressBarTheme(),
     ) -> None:
-        if not _RICH_AVAILABLE:
+        if not _module_available("rich"):
```
carmocca, line 231: This could still use `_RICH_AVAILABLE` as the message already suggests using `-U`

See how it's done for the CLI:

https://github.com/PyTorchLightning/pytorch-lightning/blob/5d9df39b1c35629c25ed3faba094a03b07dec15c/pytorch_lightning/utilities/imports.py#L87

https://github.com/PyTorchLightning/pytorch-lightning/blob/5d9df39b1c35629c25ed3faba094a03b07dec15c/pytorch_lightning/utilities/cli.py#L107-L111


There's no need for 2 separate checks in this case

kaushikb11: Yup, we had it initially for Rich as well. But imo it is better error handling for the user stating what's missing.

carmocca: You can still state it in that case, by keeping only this if but the message in the bottom error.

carmocca: if rich is not installed, it also misses the requirement of `rich >= 10.2.2`

tchaton (approved):
LGTM !

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -227,10 +228,15 @@ def __init__(
...
             raise ModuleNotFoundError(
                 "`RichProgressBar` requires `rich` to be installed. Install it by running `pip install -U rich`."
             )
+        if not _RICH_GREATER_EQUAL_10_2_2:
+            raise MisconfigurationException(
+                "`RichProgressBar` requires `rich` >= 10.2.2. Install it by running `pip install -U rich`."
+            )
```
awaelchli, line 238: It looks to me that this message can directly replace the above message and be used for both the initial installation notice as well as for letting the user know they need to upgrade.

awaelchli (approved):

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -227,10 +228,15 @@ def __init__(
         leave: bool = False,
         theme: RichProgressBarTheme = RichProgressBarTheme(),
     ) -> None:
-        if not _RICH_AVAILABLE:
+        if not _module_available("rich"):
             raise ModuleNotFoundError(
                 "`RichProgressBar` requires `rich` to be installed. Install it by running `pip install -U rich`."
             )
+        if not _RICH_GREATER_EQUAL_10_2_2:
```
Borda, line 235: I would say that this shall be in requirements, not in code...

SeanNaren:
hey @kaushikb11 thanks for bringing this up :)

This is actually a pain I have had in the past with Flash, since a lot of the dependencies are quite touchy (specific versions and whatnot) it's hard to know the exact cause of the unavailability. I suggest following @awaelchli approach of combining the two together and mentioning that we require 10.2.2 and above :)

SeanNaren (approved):

pytorch_lightning/callbacks/progress/rich_progress.py
carmocca: Yes, this is what I meant with my comment above

kaushikb11: Updated!

Issue #10648, opened 2021-11-20
[Accelerator refactor] Move Accelerator into Strategy


four4fish:
## Proposed refactor

Part 3 of Accelerator and Plugin refactor #10416 

### Motivation

Moving towards stable version

After step 2 Move Precision Plugin into TTP Precision Plugins should be part of Training Type Plugins #7324
Accelerator is not the routing layer for strategy and precision anymore, optimizer related logic, steps, hooks all moved in to strategy. 

Accelerator only have device information - strategy should own accelerator. Reduce the code complexity and improve code maintainability  

### Pitch
move accelerator into ttp/strategy as device_plugin (similar to checkpoint_io) and updating logic in accelerator-connector, training, loops accordingly

[RFC] Should we have a new name for Accelerator?

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @justusschock @awaelchli @akihironitta @kaushikb11 @ananthsub

ananthsub:
> move accelerator into ttp/strategy as device_plugin (similar to checkpoint_io)

I would prefer keeping the Accelerator name. Other than that this looks good to me.

awaelchli:
Definitely 100% the Accelerator name should stay, and also its responsibilities. It should remain the main component to access hardware. For example, all the `*_to_device` methods and similar currently on the strategy should eventually call in the Accelerator to perform the action IMO. Does that make sense?

ananthsub:
[...]

PR #10890, opened 2021-12-02
4/n Move Accelerator into strategy - remove X_step() from accelerator

four4fish:
## What does this PR do?
Remove training_step() from Accelerator, and call strategy.training_step directly

In the following graph: Each color is one logic. 

[](https://postimg.cc/H8VVLFMQ)

[RFC] Should we have training_step() in Parallel Plugin, then DDPSpawning, DP and DDP doesn't need to duplicate the logic. Instead we need training_step in Horovod. 
cc: @awaelchli @ananthsub @carmocca @justusschock 

Removed TPUSpawning.training_step() as the logic is same as the super class DDPSpawn

Part of #10648

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -397,8 +397,9 @@ def reduce(self, tensor, group: Optional[Any] = None, reduce_op: Union[ReduceOp,
             tensor = sync_ddp_if_available(tensor, group, reduce_op=reduce_op)
         return tensor
 
-    def training_step(self, *args, **kwargs) -> Optional[Any]:
-        return self.model(*args, **kwargs)
+    def training_step(self, step_kwargs: Dict[str, Union[Any, int]]) -> STEP_OUTPUT:
+        with self.precision_plugin.train_step_context():
+            return self.model(*step_kwargs.values())
```
justusschock, line 402: Depending on the order they were assigned to the dict, this may not be the same. Also in general, there may be keyword only arguments

pytorch_lightning/plugins/training_type/ddp.py
```diff
@@ -397,8 +397,9 @@ def reduce(self, tensor, group: Optional[Any] = None, reduce_op: Union[ReduceOp,
             tensor = sync_ddp_if_available(tensor, group, reduce_op=reduce_op)
         return tensor
 
-    def training_step(self, *args, **kwargs) -> Optional[Any]:
-        return self.model(*args, **kwargs)
+    def training_step(self, step_kwargs: Dict[str, Union[Any, int]]) -> STEP_OUTPUT:
```
justusschock, line 400: Can we stick to `*args` and `**kwargs` please? We can rename them to `step_args` and `step_kwargs`, but I think it's important that we have positional and keyword arguments, as this is the only way to possibly deal with [PEP 570](https://www.python.org/dev/peps/pep-0570/) and [PEP 3102](https://www.python.org/dev/peps/pep-3102/)

justusschock (changes requested):

awaelchli:
> [RFC] Should we have training_step() in Parallel Plugin, then DDPSpawning, DP and DDP doesn't need to duplicate the logic. Instead we need training_step in Horovod.

I would keep the overrides for all plugins that wrap the model with a class like DistributedDataParallel for example. This way, the definition of how the model gets wrapped lies in the plugin together with how the training_step gets called through that wrapper.

pytorch_lightning/plugins/training_type/ddp.py
ananthsub: @justusschock @awaelchli do you recall why the accelerator had a different signature/convention compared to the training type plugin?

four4fish:
> > [RFC] Should we have training_step() in Parallel Plugin, then DDPSpawning, DP and DDP doesn't need to duplicate the logic. Instead we need training_step in Horovod.
> 
> I would keep the overrides for all plugins that wrap the model with a class like DistributedDataParallel for example. This way, the definition of how the model gets wrapped lies in the plugin together with how the training_step gets called through that wrapper.

I agree, even have duplicate logic is better than fail silently or inherent silencely. In the last step of the refactor, we can regrouping logics and flatten the inheritance

pytorch_lightning/plugins/training_type/ddp.py
four4fish: @justusschock Agree that *args, **kwargs is better, and it's align with what torch model training step will take and it's more flexible. I'm copying Accelerator's signature in this PR because all call to training_type_plugin.training_step was called through accelerator.training_step() before. And Accelerator's signature is 
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/accelerators/accelerator.py#L124-L130
If we use `*args, **kwargs` now will loose the requirement for the typing in a way, which I was concerned that could cause unexpected failures.

justusschock: @four4fish I get it. I think we still should move on with args and kwargs :)

@ananthsub yes I think that should be part of the loops. So that plugins are unrelated to that.

four4fish: @justusschock sounds good, will update and mark as breaking change

four4fish: Issue created, will address the issue before update this PR
https://github.com/PyTorchLightning/pytorch-lightning/issues/10907

ananthsub (approved):

pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -427,7 +427,7 @@ def _training_step(self, split_batch: Any, batch_idx: int, opt_idx: int) -> Clos
             )
 
             # manually capture logged metrics
-            training_step_output = self.trainer._call_accelerator_hook("training_step", *step_kwargs.values())
```
awaelchli, line 430: After the changes in this PR, there is now only one instance of `_call_accelerator_hook` left, one that calls the `on_train_end` hook on the accelerator.
cc @daniellepintz @carmocca

awaelchli (approved):
nice!

pytorch_lightning/loops/optimization/optimizer_loop.py
daniellepintz: do you mean `on_train_start`?
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d59a2faff72aa34254b156607458f2b9acf8523/pytorch_lightning/loops/fit_loop.py#L198

awaelchli: Yes

daniellepintz: Cool. It seems like the only reason we still need that is for the `torch.cuda.empty_cache()` line here:
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d59a2faff72aa34254b156607458f2b9acf8523/pytorch_lightning/accelerators/gpu.py#L50-L53

Could we put that line into `setup` instead?

carmocca (approved):

justusschock (approved):

pytorch_lightning/loops/optimization/optimizer_loop.py
awaelchli: I think we could do that yes.

pytorch_lightning/loops/optimization/manual_loop.py
```diff
@@ -102,7 +102,7 @@ def advance(self, batch: Any, batch_idx: int) -> None:  # type: ignore[override]
             )
 
             # manually capture logged metrics
-            training_step_output = self.trainer._call_accelerator_hook("training_step", *step_kwargs.values())
+            training_step_output = self.trainer._call_ttp_hook("training_step", *step_kwargs.values())
```
tchaton, line 105: Should we use an intermediate `_training_step` which contains:

```py
def _training_step(...):
      with self.precision_plugin.train_step_context():
               self.training_step(...)
```

I am not a huge fan of duplicating the logic within all *_step methods.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1521,7 +1521,8 @@ def _call_ttp_hook(
         hook_name: str,
         *args: Any,
         **kwargs: Any,
-    ):
+    ) -> Optional[Any]:
```
daniellepintz, line 1524: unfortunately this causes a mypy error, which is why I had to remove it in my PR

pytorch_lightning/trainer/trainer.py
```diff
@@ -1538,7 +1538,7 @@ def _call_accelerator_hook(
         **kwargs: Any,
     ) -> Optional[Any]:
         self.lightning_module._current_fx_name = hook_name
-        fn = getattr(self.accelerator, hook_name)
+        fn = getattr(self.training_type_plugin, hook_name)
```
daniellepintz, line 1541: why make this change? why not just call `_call_ttp_hook` instead of `_call_accelerator_hook`?

tchaton: Hey @awaelchli,

A small idea for the `test_fx_validator_integration` failure. The refactor didn't properly reset the hook name on the LightningModule as there are hooks that are nested such as the `training_step` with manual optimization. I think @carmocca has a draft PR to restore this behavior. Not sure if it could be the reason.

carmocca: This is fixed by #10575, but I don't have the time to finish it soon. If somebody wants to resolve the failing profiler tests, feel free to push to the PR.

pytorch_lightning/loops/optimization/manual_loop.py
four4fish: We have a issue for simplify precision_plugins's X_steps #10891
But here, different plugins has different behaviors of call model(arg,kwarg) or model.train/test/_step(), any idea how can we simplify strategies's x_steps?

pytorch_lightning/trainer/trainer.py
four4fish: @carmocca  you mean #10957?

Issue #10688, opened 2021-11-23
Exception: The wandb backend process has shutdown


morestart:
## 🐛 Bug

Exception: The wandb backend process has shutdown

full error info:
```
Traceback (most recent call last):
  File "/home/cat/PycharmProjects/torch-ocr/tools/train/det_train/train.py", line 59, in <module>
    trainer.fit(model, data)
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 735, in fit
    self._call_and_handle_interrupt(
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 682, in _call_and_handle_interrupt
    return trainer_fn(*args, **kwargs)
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 770, in _fit_impl
    self._run(model, ckpt_path=ckpt_path)
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1193, in _run
    self._dispatch()
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1272, in _dispatch
    self.training_type_plugin.start_training(self)
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/plugins/training_type/training_type_plugin.py", line 202, in start_training
    self._results = trainer.run_stage()
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1282, in run_stage
    return self._run_train()
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/trainer/trainer.py", line 1312, in _run_train
    self.fit_loop.run()
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/loops/base.py", line 145, in run
    self.advance(*args, **kwargs)
  File "/home/cat/miniconda3/envs/torch-ocr/lib/python3.8/site-packages/pytorch_lightning/loops/fit_loop.py", line 234, in advance
[...]

Issue #9876, opened 2021-10-09
WandBlogger doesn't log when working with TPU cores


rusty-electron:
## 🐛 Bug

I am trying to run a resnet200d model for document classification on Kaggle TPUs using python-lightning. I have set up WandB logger and connected it to my wandb account. The training starts well but at 50 steps the logger complains that It can't log as the process was called from a different pid. See error image below:

![wandb_issue](https://user-images.githubusercontent.com/22986666/136651234-fbcb7de9-feaa-4218-a036-1aa1a51f7c4a.png)

I have tried to use options like `sync_dist=True` or `rank_zero_only=True` (in the self.log calls) but it still doesn't work.

### To Reproduce

You can try running the mentioned kaggle notebook:
https://www.kaggle.com/rustyelectron/documentclassification-pytorch-tpu-wandb/

Wait till about 50 steps to get the error.

### Expected behavior

The wandblogger should work normally and log data.

### Environment

* CUDA:
	- GPU:
	- available:         False
	- version:           None
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     True
	- pyTorch_version:   1.7.0a0+7e71a98
	- pytorch-lightning: 1.2.10
	- tqdm:              4.62.1
* System:
	- OS:                Linux
	- architecture:
		- 64bit
	- processor:         x86_64
	- python:            3.7.10

### Additional context

None

cc @kaushikb11 @awaelchli @morganmcg1 @AyushExel @borisdayma @scottire

Programmer-RD-AI:
hi,

Can you check these issues?

https://forums.pytorchlightning.ai/t/tpu-failures-in-colab/310
https://github.com/wandb/client/issues/1994
https://github.com/PyTorchLightning/pytorch-lightning/issues/1062
https://discuss.huggingface.co/t/logging-experiment-tracking-with-w-b/498
https://wandb.ai/site/articles/pytorch-lightning-with-weights-biases
https://wandb.ai/cayush/pytorchlightning/reports/Use-Pytorch-Lightning-with-Weights-Biases--Vmlldzo2NjQ1Mw

Regards,

rusty-electron:
@Programmer-RD-AI sorry, 🤦 its public now. Ok, I will take a look at the links you provided.

Programmer-RD-AI:
It ok check the links and see if they solve the issue

Regards

borisdayma:
This should now be fixed.
[...]

PR #10896, opened 2021-12-02
2/n Simplify spawn plugins: Spawn immediately

### LINKED ISSUE #10059: Simplify multiprocessing logic in DDPSpawn plugins

When working on #9987 and the corresponding refactors around the spawn plugins, I realized that the logic around the multiprocessing queue and how results are handled and returned to the trainer is quite outdated and overly complicated. This logic has outlived many changes, but we never saw a way to make it simpler. With this issue I propose several steps towards a clean and easy to follow, easy to debug code path for 

1) spawning processes
2) moving results from the subprocess back to the main process
3) returning the result back to the trainer

### Motivation

There are several components in the DDPSpawn plugin around spawning processes and handling of results that are obscure and not well documented. 
On top of that, result handling bleeds into the TrainingType base class

https://github.com/PyTorchLightning/pytorch-lightning/blob/aa1540410ff55854e050ff117c2d66f22741d182/pytorch_lightning/plugins/training_type/training_type_plugin.py#L38

and also into the trainer:

https://github.com/PyTorchLightning/pytorch-lightning/blob/aa1540410ff55854e050ff117c2d66f22741d182/pytorch_lightning/trainer/trainer.py#L1123-L1125

**This is quite confusing to anyone not familiar with the peculiarities of ddp spawn. But it does not have to be that way. The situation can be drastically improved!**

#### Step 1
Remove the `self.mp_queue` attribute from the plugin. It is not required and can be locally created and used within the recently introduced `DDPSpawnPlugin.spawn` method #10018

#### Step 2
Instead of adding items like last_path, best_path, or results to the queue one by one, add all data at once as one result tuple to the queue. 

This logic

https://github.com/PyTorchLightning/pytorch-lightning/blob/aa1540410ff55854e050ff117c2d66f22741d182/pytorch_lightning/plugins/training_type/ddp_spawn.py#L224-L238

becomes 

```py

# inside the spawned processes
def new_process(self, ...):
    # ...
    return best_path, last_path, results

# in spawn wrapper:
def _wrapped_function(...):
    result = function(*args, **kwargs)
    if self.is_global_zero:
        queue.put(move_data_to_device(result, "cpu"))  # the tuple of data

# in main process:
def spawn(self, ...):
    queue = SimpleQueue()
    mp.spawn(self._wrapped_function, args=(function, args, kwargs, mp_queue), nprocs=self.num_processes)
    return mp_queue.get()  # the tuple of data
```

This allows us to standardize and limit the queue to a single `put()` and correspondingly a single `get`. This is less error prone and easier to understand for everyone working with custom plugins. The only really complicated part where a learning curve is steep for the reader is this code snippet above. Everything else will extremely simplify.

#### Step 3
With 1) and 2) in place, we can directly return the results from the spawned function instead of caching it in the attribute `self._results`. 

#### Step 4
Finally, we can get rid of dispatch and post dispatch 
   
https://github.com/PyTorchLightning/pytorch-lightning/blob/aa1540410ff55854e050ff117c2d66f22741d182/pytorch_lightning/trainer/trainer.py#L1102-L1107

and combine it into a single `plugin.run` call or alike:

```python
self.training_type_plugin.run(self.train)  # or whatever other trainer method
```
This then cleanly generalizes across all plugins. The confusing concept of dispatch and post dispatch is gone.

#### Step 5

Proposed by @ananthsub, the next step would be to directly spawn processes with the Trainer.fit() call. This how we do it in Lite as well: 

https://github.com/PyTorchLightning/pytorch-lightning/blob/412d507a73c79f5e4f7ef14289cefe2eb2af94a7/pytorch_lightning/lite/lite.py#L387-L396

The benefits of this last step are ultimately (https://github.com/PyTorchLightning/pytorch-lightning/issues/10059#issuecomment-954444367):

- consistent execution flow between spawn and non-spawn versions since _run is shared across both
   - No longer will it be required to exclude some hooks or change hook order based on the plugin spawn type. Example: setup() hook
   - No longer regressions with Loggers being accidentally created on the main process for spawn plugins. With step 5, we will no longer need to artificially delay any logger.experiment calls.
- we can move up initialization of distributed init for spawn plugins because processes are spawned earlier. this means no subtle difference in implementations of collectives, which @four4fish found out the tedious & difficult way with Deprecate LightningDistributed and keep logic in ddp/ddpSpawn directly #9691

### Nexts steps
A quick and dirty draft is available here in form of a PR (excluding some steps): #10034

### Additional context

The `add_to_queue` and `get_from_queue` methods were recently introduced, initially on the LightningModule and now they are in a deprecation phase. We would need to incorporate them into this design as well. Suggestions welcome.

<sub>

</sub>

awaelchli:
## What does this PR do?

Fixes #10059 (step 4 & 5)

Moves the spawning of process right at the beginning of `Trianer.fit/validate/test/predict`.
- Greatly improve readability due to simplified code path
- Aligns the code path of DDPSpawn with other plugins like DDP
- Hooks like setup() and teardown() now consistently run inside the worker processes and process group initialized
- Significantly decreases the chance that the `root_device` gets accessed in a hook that runs outside the spawned processes in TPU spawn. This today would cause the process to exit with a cryptic message. 

The core change of is that the spawning happens a level higher up in the trainer. Here is a simplified illustration using functions for before vs. after:

Before:

```py
def _run(model, ...):
    setup()
    other()
    # <-- processes spawn here
    mp.spawn(train_fn, args=model)
    # <-- processes join here

    teardown()
```

After:

```py
def _run(model):
    setup()
    other()
    train_fn(model)
    teardown()

# <-- processes spawn here
mp.spawn(_run, args=model)
# <-- processes join here
```

Possible issues that this PR might fix or help fix (need to check):
- #10688 
-  #9876

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @kaushikb11 @awaelchli @ninginthecloud


pytorch_lightning/trainer/trainer.py
```diff
@@ -676,7 +676,12 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
             **kwargs: keyword arguments to be passed to `trainer_fn`
         """
         try:
-            return trainer_fn(*args, **kwargs)
+            if isinstance(self.training_type_plugin, DDPSpawnPlugin):
+                spawn_output: _SpawnOutput = self.training_type_plugin.spawn(trainer_fn, *args, **kwargs)
+                self.training_type_plugin._recover_results_in_main_process(spawn_output, self)
```
ananthsub, line 681: I really like the simplification this offers. I am wary of tying this to DDP Spawn however. This forces everyone to subclass DDPSpawnPlugin and override different parts.

What do you think about offering an AbstractSpawn interace which requires people to implement `spawn` and `recover_results_in_main_process` and which defines the `spawn_output` struct?

Then DDP Spawn could both inherit from TrainingTypePlugin and implement `AbstractSpawn`

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -402,7 +402,6 @@ def test_model_checkpoint_no_extraneous_invocations(tmpdir):
         max_epochs=num_epochs,
     )
     trainer.fit(model)
-    assert trainer.state.finished, f"Training failed with {trainer.state}"
```
awaelchli, line 405: Here is one blocker I am facing, in short: 

```python
trainer = Trainer(strategy="ddp_spawn", devices=2) 
trainer.fit(model)
assert trainer.state == TrainerState.finished
```

The main process spawns immediately 2 processes with the fit call (as proposed by this new design). These workers will update the state in their own instance of the trainer but once processes terminate/join, these changes will not be reflected in the main process, which has never seen any updates (as expected). 


A solution to this would be to put the trainer.state variable onto the queue and restore it in the main process. I am wondering if this is worth it. RFC from reviewers.

pytorch_lightning/trainer/trainer.py
awaelchli: @ananthsub I was actually thinking in a similar direction, I imagined it with composition rather than inheritance. My idea was to have an (abstract) Executor interface, of which the TTP owns an instance like the checkpoint IO plugin. 
I'm definitely open to discuss such directions in more detail

justusschock: @awaelchli You mean something like an internal launcher? That either calls the script or uses spawn? I think we should do that to make it way more modular.

awaelchli: Yes I was thinking something like that.

tests/checkpointing/test_model_checkpoint.py
justusschock: Why don't we spawn N-1 processes and use the main proc as well?

Other then that we could probably send it through the queue as that should be quite cheap to do.

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -377,13 +350,3 @@ def checkpoint_io(self) -> CheckpointIO:
     @checkpoint_io.setter
     def checkpoint_io(self, plugin: CheckpointIO) -> None:
         raise MisconfigurationException("TPU Spawn Plugin currently does not support custom checkpoint plugins.")
-
-    @staticmethod
-    def _clean_logger(trainer: "pl.Trainer") -> None:
-        loggers = trainer.logger._logger_iterable if isinstance(trainer.logger, LoggerCollection) else [trainer.logger]
```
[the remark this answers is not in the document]
justusschock: Is it still necessary? It could be if the logger is still pickled right?

pytorch_lightning/trainer/trainer.py
justusschock: Will you open an issue for tracking this? :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -1113,25 +1124,19 @@ def _run(
              Lightning internal flow looks like this:
         {Trainer.fit} or {Trainer.test} or {Trainer.predict}  ||
                                 |                             ||
-                        create accelerator                    ||
-                                |                             ||
-                         {self._dispatch}                     ||
+                        setup accelerator                     ||
+                           and strategy                       ||
```
justusschock, line 1128: should we mention that this may include spawning processes?

tests/trainer/test_trainer.py
```diff
@@ -62,6 +62,10 @@
 if _OMEGACONF_AVAILABLE:
     from omegaconf import OmegaConf
 
+ProcessRaisedException = Exception
+if _TORCH_GREATER_EQUAL_1_8:
+    from torch.multiprocessing import ProcessRaisedException
```
justusschock, line 67:
```suggestion
if _TORCH_GREATER_EQUAL_1_8:
    from torch.multiprocessing import ProcessRaisedException
else:
    ProcessRaisedException = Exception
```

just a nit, but doesn't make sense to have a value and immediately overwrite it :)

justusschock (commented):
Will the processes live for the entire lifetime of the trainer or just for the duration of fit? 

I.e. when I first call fit and then test, will it use the same processes or will new processes be spawned for testing?

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -118,10 +118,23 @@ def connect(self, model: "pl.LightningModule") -> None:
         return super().connect(model)
 
     def pre_dispatch(self, trainer: "pl.Trainer") -> None:
-        super().pre_dispatch(trainer)
+        self._move_optimizer_state()
         if self.debug:
             os.environ["PT_XLA_DEBUG"] = str(1)
 
+        if self.tpu_global_core_rank != 0 and trainer.progress_bar_callback is not None:
+            trainer.progress_bar_callback.disable()
```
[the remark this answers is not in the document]
kaushikb11: Yup @awaelchli It could be removed.

awaelchli:
@justusschock The latter. This remains the same as before. The only difference is that the spawning happens a level higher up in the trainer. I updated the PR description with a visualization.

justusschock (approved):

pytorch_lightning/plugins/training_type/sharded_spawn.py
```diff
@@ -114,12 +114,12 @@ def pre_backward(self, closure_loss: torch.Tensor) -> None:
     def post_training_step(self):
         pass
 
-    def new_process(self, trainer: "pl.Trainer") -> Optional["_SpawnOutput"]:
+    def pre_dispatch(self, trainer: "pl.Trainer") -> None:
```
ananthsub, line 117: Is this override needed now? If not this could be cleanup for a separate PR

pytorch_lightning/trainer/trainer.py
```diff
@@ -1158,10 +1163,8 @@ def _run(
 
         self.checkpoint_connector.resume_end()
 
-        # dispatch `start_training` or `start_evaluating` or `start_predicting`
-        results = self._dispatch()
-
-        self._post_dispatch()
+        results = self.run_stage()
```
ananthsub, line 1166: nit: run_stage could be marked private as it's not used outside the class. For another PR 😀

pytorch_lightning/plugins/training_type/tpu_spawn.py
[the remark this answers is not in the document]
ananthsub: super nit for a separate issue/PR: are external callers expected to use tpu_global_core_rank or tpu_local_core_rank? why not use self.global_rank ?

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -263,6 +243,10 @@ def get_mp_spawn_kwargs(self, trainer: Optional["pl.Trainer"] = None) -> Dict[st
         }
 
     def spawn(self, function: Callable, *args: Any, **kwargs: Any) -> Optional[Union[Any, "_SpawnOutput"]]:
+        # TODO: this todo is unclear, does it still apply?
+        # todo: precision pluging is call in accelerator setup and should be moved
+        if "XLA_USE_BF16" in os.environ:
+            del os.environ["XLA_USE_BF16"]
```
ananthsub, line 249: n00b question: why is this key removed? 

before this was removed only before training, but not evaluation/prediction. was that a bug in the previous code that this is fixing?

pytorch_lightning/trainer/trainer.py
```diff
@@ -721,6 +727,7 @@ def fit(
 
             datamodule: An instance of :class:`~pytorch_lightning.core.datamodule.LightningDataModule`.
         """
+        self.training_type_plugin.model = model
```
ananthsub, line 730: n00b question: why model vs lightning_module?

tests/trainer/test_trainer.py
```diff
@@ -1411,7 +1415,7 @@ def predict(
         callbacks=[cb, cb_1] if use_callbacks else [],
     )
     if strategy == "ddp_spawn":
-        with pytest.raises(MisconfigurationException):
+        with pytest.raises(ProcessRaisedException, match="`return_predictions` should be set to `False`"):
```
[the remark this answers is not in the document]
ananthsub: this should be mentioned in the changelog in case people are catching specific exceptions from lightning

pytorch_lightning/trainer/trainer.py
tchaton: Yes, this is quite exciting. This would enable us to possibly merge ddp, ddp_spawn into ddp.py and provide a popen or spawn executor.

pytorch_lightning/trainer/trainer.py
tchaton: I believe this was to differentiate with `self.lightning_module`, but it might be confusing to the readers.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1223,25 +1226,18 @@ def _log_hyperparams(self) -> None:
...
-    def _dispatch(self) -> Any:
-        if self.evaluating:
-            return self.training_type_plugin.start_evaluating(self)
-        elif self.predicting:
-            return self.training_type_plugin.start_predicting(self)
-        else:
-            return self.training_type_plugin.start_training(self)
-
     def run_stage(self):
+        self.training_type_plugin.barrier("run-stage")
```
tchaton, line 1240: IMO, I would move this logic within the run function or its own method. More calls might need to be moved in dispatch.

pytorch_lightning/plugins/training_type/ddp_spawn.py
```diff
@@ -132,23 +132,6 @@ def set_world_ranks(self, process_idx: int = 0) -> None:
     def get_mp_spawn_kwargs(self, trainer: Optional["pl.Trainer"] = None) -> Dict[str, Any]:
         return {"nprocs": self.num_processes}
 
-    def start_training(self, trainer: "pl.Trainer") -> Any:
```
tchaton, line 135: This is aw !

pytorch_lightning/trainer/trainer.py
tchaton: Ideally, we would have `self.training_type_plugin.run(trainer_fn, *args, **kwargs)` and this would execute the associated trainer method with the spawn / none executor.

pytorch_lightning/trainer/trainer.py
tchaton: Yes, I would prefer so.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1171,15 +1174,15 @@ def _run(
...
-        # been included at the end of `new_process` functions)
-        if not isinstance(self.training_type_plugin, DDPSpawnPlugin):
-            self._call_teardown_hook()
+        self._call_teardown_hook()
 
         if self.state.status != TrainerStatus.INTERRUPTED:
             self.state.status = TrainerStatus.FINISHED
         self.state.stage = None
 
+        if isinstance(self.training_type_plugin, DDPSpawnPlugin):
```
tchaton, line 1183: Any chance we can prevent strategy leakage here ?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1223,25 +1226,18 @@ def _log_hyperparams(self) -> None:
...
-            return self.training_type_plugin.start_evaluating(self)
-        elif self.predicting:
-            return self.training_type_plugin.start_predicting(self)
-        else:
-            return self.training_type_plugin.start_training(self)
-
     def run_stage(self):
+        self.training_type_plugin.barrier("run-stage")
         self.training_type_plugin.dispatch(self)
         self.__setup_profiler()
```
tchaton, line 1242: Should this be moved to its own method ?

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -154,37 +166,6 @@ def init_dist_connection(self, global_rank: int, world_size: int) -> None:
...
-
-        results = trainer.run_stage()
-
-        outputs = self._collect_rank_zero_results(trainer, results)
-
-        # https://github.com/pytorch/xla/issues/1801#issuecomment-602799542
-        self.barrier("end-process")
-
-        # https://github.com/pytorch/xla/issues/2190#issuecomment-641665358
-        if self.local_rank == 0:
```
tchaton, line 181: This might be required for TPU. @kaushikb11

pytorch_lightning/plugins/training_type/tpu_spawn.py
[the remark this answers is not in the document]
ananthsub: progress bar disabling can be further refactored. posted this issue to discuss further: https://github.com/PyTorchLightning/pytorch-lightning/issues/10982

pytorch_lightning/trainer/trainer.py
```diff
@@ -1171,15 +1174,15 @@ def _run(
             self._call_callback_hooks("on_fit_end")
             self._call_lightning_module_hook("on_fit_end")
 
-        # teardown if necessary (similar calls for spawn plugins are excluded as they have
-        # been included at the end of `new_process` functions)
-        if not isinstance(self.training_type_plugin, DDPSpawnPlugin):
-            self._call_teardown_hook()
```
four4fish, line 1177: Where was teardown hook get called for spawning before?

pytorch_lightning/plugins/training_type/tpu_spawn.py
awaelchli: You are right, and we have this exact mechanism already in the `_wrapper_function()` at the very end. We added this when we added TPU support for Lite. So all good here, rank 0 will exit last, as required :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -1171,15 +1174,15 @@ def _run(
...
+        self._call_teardown_hook()
 
         if self.state.status != TrainerStatus.INTERRUPTED:
             self.state.status = TrainerStatus.FINISHED
         self.state.stage = None
 
+        if isinstance(self.training_type_plugin, DDPSpawnPlugin):
+            results = self.training_type_plugin._collect_rank_zero_results(self, results)
```
four4fish, line 1184: The _collect_rank_zero_results called after teardown hook, will there be issue if processes got killed in teardown hook?

pytorch_lightning/trainer/trainer.py
awaelchli: Awesome, we will continue discussion and drafting over here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10985

pytorch_lightning/plugins/training_type/tpu_spawn.py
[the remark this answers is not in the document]
awaelchli: Let's investigate this. https://github.com/PyTorchLightning/pytorch-lightning/issues/10986

pytorch_lightning/trainer/trainer.py
four4fish: +1 can we deprecate run_stage()?

four4fish: created a issue to discuss further. https://github.com/PyTorchLightning/pytorch-lightning/issues/10987

tests/trainer/test_trainer.py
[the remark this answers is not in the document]
awaelchli: good point. added this

pytorch_lightning/trainer/trainer.py
awaelchli: directly in the DDPSpawnPlugin, here: https://github.com/PyTorchLightning/pytorch-lightning/blob/aeb0b5595fd73d086f4ae0f99d3f1f112f6a4c29/pytorch_lightning/plugins/training_type/ddp_spawn.py#L205

But now, this is no longer split.

four4fish: Oh I see, it teared down in the same function. Should it be after _collect_rank_zero_results then? same as comments below

pytorch_lightning/trainer/trainer.py
awaelchli: _collect_rank_zero_results needs to be at the very and, because it needs to include the latest state of trainer.state. Whether teardown runs before or after this does not matter at the moment (process killing does not happen inside teardown). We may want to collect ideas how to polish this in #10985

pytorch_lightning/trainer/trainer.py
awaelchli: Not at this moment, but some ideas here #10985  might help to do that in the future.

pytorch_lightning/trainer/trainer.py
four4fish: But if the processes already teardown, will we lost state_dicts?
The old new_process logic seems collected the output first? I'm probably miss something here
https://github.com/PyTorchLightning/pytorch-lightning/blob/aeb0b5595fd73d086f4ae0f99d3f1f112f6a4c29/pytorch_lightning/plugins/training_type/ddp_spawn.py#L202-L205

pytorch_lightning/plugins/training_type/sharded_spawn.py
awaelchli: I'm not sure yet, I need to investigate in #10988, because the ShardedScaler gets set at instantiation of the precision plugin which originally happens in the main process on Trainer init.

pytorch_lightning/trainer/trainer.py
awaelchli: Good that you mention this. I looked at usages and found HorovodPlugin is the only one left that calls `trainer.run_stage()`. I completely missed this until now. Hashing it out in #10989 as a prerequisite for the PR here.

pytorch_lightning/plugins/training_type/tpu_spawn.py
awaelchli: True, something is odd here. Actually, I don't think the variable should be removed at all. It is being set initially in the precision plugin connect() method, so removing it here right after makes no sense. I will address this in #10990.

pytorch_lightning/trainer/trainer.py
awaelchli: I think not worth it, because dispatch() will disappear

pytorch_lightning/trainer/trainer.py
awaelchli: The teardown() call does not kill the process. Teardown() is just calling accelerator.teardown() etc. and teardown() hooks on LM etc. runs there. 

Here is what happens inside _run():

```py
def _run(...):
     # each worker process runs this function
     # <--- worker process starts here

     setup()
     ...
     run_stage()
     ...
     teardown() 

     _collect_rank_zero_results()

     # end of the function, all worker processes are done and will shut down
     return
```

pytorch_lightning/plugins/training_type/tpu_spawn.py
kaushikb11: Sounds good to me! We had this discussion during the (long & sweet) debugging for Lite haha.

Thanks for the clean up! It reflects the same expected behavior.

pytorch_lightning/plugins/training_type/tpu_spawn.py
awaelchli: Ok, as discussed in #10990, I will carry this bug over in this PR and fix it by teardown() in the linked PR. Resolving

pytorch_lightning/trainer/trainer.py
awaelchli: Thanks @four4fish, will follow up on it in  #11000

pytorch_lightning/trainer/trainer.py
awaelchli: `TTP.lightning_module` doesn't have a setter, `TTP.model` has a setter. 
Initially, `TTP.lightning_module == TTP.model`, and later during setup TTP.model will hold the wrapped model.
This line that I added here (and in the other Trainer methods) is needed for the case when:

```py
trainer.fit(model)
trainer.test()  # no model provided, need to reuse self.lightning_module

```
For spawn plugins, the model reference would be lost if we didn't save it in the main process.

pytorch_lightning/trainer/trainer.py
awaelchli: updated with your suggestions

pytorch_lightning/trainer/trainer.py
awaelchli: I prefer not to change the order in the Trainer._run() for existing behavior. 
It's not a big problem that teardown() is not the very very last hook to run. If we want to change the order, I believe we  should discuss the impact and do it in another PR.

awaelchli:
I have addressed all major questions and reviews. Thanks for the input and positive response so far!

ananthsub (approved):
excellent! super exciting to see this simplification

four4fish:
Looks great to me! I don't have permission to accept, but super happy to see this 💯  A big milestone for Accelerator stable version!!

kaushikb11 (approved):
Woohoo!

tchaton (approved):
Awesome work !

pytorch_lightning/plugins/training_type/tpu_spawn.py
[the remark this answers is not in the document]
tchaton: It might be, but only if the users doesn't something special

Issue #10904, opened 2021-12-02
Logging does not work on evaluation epoch end with multiple dataloaders


carmocca:
## 🐛 Bug

When multiple dataloaders are used during evaluation (validation or test), they are iterated like this:

```python

DataLoader 0: 0-1-2-3-4-5           
DataLoader 1:            0-1-2-3-4-5
                       (A)        (B)
```

Where the first iterator is completely exhausted before starting the next one. Each has 5 batches.

One might ask, what's considered the epoch end? (A) and (B) or just (B)?
Lightning runs `{validation,test}_epoch_end` ONLY on (B), I'm not sure why is that exactly but one can obtain the behavior of (A) and (B) by simply having separate `trainer.test(..)` calls.

The return of `trainer.test()` is a list of results (dictionaries) per dataloader. In the case of 1 DataLoader:

```python
class Model():
    def test_step(self, batch, batch_idx, dataloader_idx=None):
        self.log(f'foo', 123., add_dataloader_idx=True)

    def test_epoch_end(self, outputs) -> None:
        self.log("foobar", 123.)

x = trainer.test(model, datamodule=dm)
print(x)  # [{'foo': 123.0, 'foobar': 123.0}]
```

Notice how the DataLoader results include both items.

However, if we use 2 DataLoaders, the value of `x` is:

`[{'foo/dataloader_idx_0': 123.0}, {'foo/dataloader_idx_1': 123.0}]`

where neither of them include `foobar`.

This is correct in the sense that `foobar` was logged in `epoch_end` and it's not part of either of the dataloader states. One could also argue it's part of both and both should have it.

`[{'foo/dataloader_idx_0': 123.0, 'foobar': 123.0}, {'foo/dataloader_idx_1': 123.0, 'foobar': 123.0}]`

### To Reproduce

<details>

```python
import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer, LightningDataModule


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringData(LightningDataModule):
    def __init__(self):
        super().__init__()

    def test_dataloader(self):
        return [DataLoader(RandomDataset(32, 100), batch_size=16), DataLoader(RandomDataset(32, 100), batch_size=16)]
[...]

PR #10923, opened 2021-12-03
Sort out the dataloader idx logic for evaluation

carmocca:
## What does this PR do?

- Save the current dataloader idx in the `ResultCollection` instead of the `LightningModule`
- Avoid leaking the dataloader idx logic from the EvaluationLoop (where it should be contained) to the `EvaluationEpochLoop`

Follow-up to #10810

This will enable an optimization where we don't compute the results multiple times when multiple dataloaders are used during evaluation.

### Does your PR introduce any breaking changes? If yes, please list them.

None

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [n/a] Did you make sure to **update the documentation** with your changes? (if necessary)
- [n/a] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [n/a] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

cc @borda @justusschock @awaelchli @akihironitta @carmocca @edward-io @ananthsub @ninginthecloud

PR #10810, opened 2021-11-29
Fix filteration logic for eval results

rohitgr7:
## What does this PR do?

Fixes #10540

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @carmocca @edward-io @ananthsub

PR #10928, opened 2021-12-03
Track the evaluation loop outputs in the loop

carmocca:
## What does this PR do?

Follow-up to #10810 and #10923

None

## Before submitting

## PR review

### DIFF
--- a/pytorch_lightning/loops/dataloader/evaluation_loop.py
+++ b/pytorch_lightning/loops/dataloader/evaluation_loop.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 from typing import Any, List, Sequence, Union
 
+import torch
 from deprecate.utils import void
 from torch.utils.data.dataloader import DataLoader
 
 from pytorch_lightning.loops.dataloader import DataLoaderLoop
 from pytorch_lightning.loops.epoch import EvaluationEpochLoop
 from pytorch_lightning.trainer.connectors.logger_connector.result import _OUT_DICT, ResultCollection
+from pytorch_lightning.trainer.states import RunningStage, TrainerFn
 from pytorch_lightning.utilities.model_helpers import is_overridden
 from pytorch_lightning.utilities.types import EPOCH_OUTPUT
 
@@ -32,6 +34,7 @@ class EvaluationLoop(DataLoaderLoop):
 
         self._results = ResultCollection(training=False)
         self._outputs: List[EPOCH_OUTPUT] = []
+        self._logged_outputs: List[_OUT_DICT] = []
         self._max_batches: List[int] = []
         self._has_run: bool = False
 
@@ -76,6 +79,7 @@ class EvaluationLoop(DataLoaderLoop):
         self._max_batches = self._get_max_batches()
         # bookkeeping
         self._outputs = []
+        self._logged_outputs = []
 
         if isinstance(self._max_batches, int):
             self._max_batches = [self._max_batches] * len(self.dataloaders)
@@ -118,18 +122,32 @@ class EvaluationLoop(DataLoaderLoop):
             # indicate the loop has run
             self._has_run = True
 
+    def on_advance_end(self) -> None:
+        # inform logger the epoch loop has finished
+        self.trainer.logger_connector.epoch_end_reached()
+
+        self._logged_outputs.append(self.trainer.logger_connector.update_eval_epoch_metrics())
+
+        super().on_advance_end()
+
     def on_run_end(self) -> List[_OUT_DICT]:
         """Runs the ``_on_evaluation_epoch_end`` hook."""
         outputs, self._outputs = self._outputs, []  # free memory
-
-        # lightning module method
+        # hook
         self._evaluation_epoch_end(outputs)
 
         # hook
         self._on_evaluation_epoch_end()
 
-        # log epoch metrics
-        eval_loop_results = self.trainer.logger_connector.update_eval_epoch_metrics()
+        logged_outputs, self._logged_outputs = self._logged_outputs, []  # free memory
+        # include any logged outputs on epoch_end
+        if self.num_dataloaders < 2:  # TODO: remove this check
+            epoch_end_logged_outputs = self.trainer.logger_connector.update_eval_epoch_metrics()
+            for dl_outputs in logged_outputs:
+                dl_outputs.update(epoch_end_logged_outputs)
+
+        # log metrics
+        self.trainer.logger_connector.log_eval_end_metrics()
 
         # hook
         self._on_evaluation_end()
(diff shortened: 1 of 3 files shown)


pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -32,6 +34,7 @@ def __init__(self) -> None:
 
         self._results = ResultCollection(training=False)
         self._outputs: List[EPOCH_OUTPUT] = []
+        self._logged_outputs: List[_OUT_DICT] = []
```
carmocca, line 37: This substitutes the `eval_loop_results` that were in the `LoggerConnector`

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -118,26 +122,49 @@ def advance(self, *args: Any, **kwargs: Any) -> None:
             # indicate the loop has run
             self._has_run = True
 
+    def on_advance_end(self) -> None:
+        # inform logger the epoch loop has finished
+        self.trainer.logger_connector.epoch_end_reached()
+
+        self._logged_outputs.append(self.trainer.logger_connector.update_eval_epoch_metrics())
```
carmocca, line 129: Now, these are computed when each dataloader ends, instead of when all dataloaders end

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -118,26 +122,49 @@ def advance(self, *args: Any, **kwargs: Any) -> None:
...
         self._evaluation_epoch_end(outputs)
 
         # hook
         self._on_evaluation_epoch_end()
 
-        # log epoch metrics
-        eval_loop_results = self.trainer.logger_connector.update_eval_epoch_metrics()
+        logged_outputs, self._logged_outputs = self._logged_outputs, []  # free memory
+        # include any logged outputs on epoch_end
+        if self.num_dataloaders < 2:  # TODO: remove this check
```
carmocca, line 144: Removing this `if` should fix #10904. Left for a separate PR

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -237,3 +261,18 @@ def _on_evaluation_epoch_end(self) -> None:
         self.trainer.call_hook(hook_name)
         self.trainer.call_hook("on_epoch_end")
         self.trainer.logger_connector.on_epoch_end()
+
+    def _print_results(self, results: List[_OUT_DICT], stage: RunningStage) -> None:
+        # TODO: this could be updated to look nicer
```
carmocca, line 266: This was inside `LoggerConnector` before. I'd say it fits much better here, as it's a responsibility of the loop.

I've also added this TODO because we could make it look nicer, or use `rich` if possible.

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -150,58 +148,19 @@ def update_eval_step_metrics(self) -> None:
...
-
-        on_step = not self._epoch_end_reached
-        num_dataloaders = self.trainer._evaluation_loop.num_dataloaders
-        has_been_initialized = len(self.eval_loop_results) == num_dataloaders
-        assert self.trainer._evaluation_loop._results is not None
-        for dl_idx in range(num_dataloaders):
-            metrics = self.trainer._evaluation_loop._results.metrics(
-                on_step, dataloader_idx=dl_idx if num_dataloaders > 1 else None
-            )
-            callback_metrics = metrics["callback"]
```
carmocca, line 165: This no longer needs to recompute the metrics per dataloader even though we had already computed them in `update_eval_epoch_metrics`

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -118,26 +122,49 @@ def advance(self, *args: Any, **kwargs: Any) -> None:
...
         # enable train mode again
         self._on_evaluation_model_train()
 
-        return eval_loop_results
+        if (
+            self.trainer.state.fn not in (TrainerFn.FITTING, TrainerFn.TUNING)
+            and not self.trainer.sanity_checking
+            and self.trainer.is_global_zero
+            # TODO: this should be defined in this loop, not the Trainer
+            and self.trainer.verbose_evaluate
```
carmocca, line 163: Now that this is here, we can move it from the Trainer to this loop

carmocca: Waiting in https://github.com/PyTorchLightning/pytorch-lightning/pull/10931

pytorch_lightning/loops/dataloader/evaluation_loop.py
rohitgr7: quick ques: where are taking care of the condition that this method will only compute metrics for the current dataloader_idx?

carmocca: This will call `ResultCollection.metrics(False)` which in turn calls `ResultCollection.valid_items()` where we filter with the `ResultCollection.dataloader_idx` value

The LoggerConnector makes sure the `ResultCollection.dataloader_idx` value is the correct one

rohitgr7 (approved):
this is great!

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: > why is there so much space between lines? minimal distance rule for covid does not apply for code.

It's the general style used int he rest of this file. I also prefer tighter code. Feel free to push whitespace changes to your liking!

awaelchli (approved):

carmocca:
@tchaton this needs your review :)

pytorch_lightning/loops/dataloader/evaluation_loop.py
tchaton: Also, I think this might create some metric problems when doing cross-dataloaders aggregation as the first compute would be done on the first datalaoder and _compute would be stored.

Here is the failing test:

```py
def test_multiple_dataloaders_logging(tmpdir):

    class TestModel(BoringModel):

        def validation_step(self, batch, batch_idx, dataloader_idx):
            self.log("value_1", dataloader_idx, add_dataloader_idx=False)
            self.log("value_2", dataloader_idx, add_dataloader_idx=True)

        def val_dataloader(self):
            return [self.train_dataloader(), self.train_dataloader()]

    model = TestModel()
    model.validation_epoch_end = None
    trainer = Trainer(default_root_dir=tmpdir)
    results = trainer.validate(model)
    assert results == [{'value_2/dataloader_idx_0': 0.0, 'value_1': 0.5}, {'value_2/dataloader_idx_1': 1.0, 'value_1': 0.5}]
```

tchaton (changes requested):
Hey @carmocca, some investigation required around `add_dataloader_idx=False`. I added a test.

pytorch_lightning/loops/dataloader/evaluation_loop.py
tchaton: Here is a PR to discuss the right behaviour: https://github.com/PyTorchLightning/pytorch-lightning/pull/11125/files

tchaton: This would need better investigation, but it seems this is confusing. I will open an issue and unblock this PR.

tchaton (approved):

PR #10940, opened 2021-12-04
Deprecate callback hooks `on_init_start` and `on_init_end`

### LINKED ISSUE #10894: [RFC] Deprecate callback hooks `on_init_start` and `on_init_end`
### Discussed in https://github.com/PyTorchLightning/pytorch-lightning/discussions/10677

Part of #7740

<div type='discussions-op-text'>

<sup>Originally posted by **daniellepintz** November 22, 2021</sup>
These hooks are called when trainer initialization begins and ends, before the model has been set, essentially allowing the user to modify the Trainer constructor.  Should we be giving the user this much control over Trainer constructor? Are there scenarios where this is needed? Or can we deprecate these hooks?

https://github.com/PyTorchLightning/pytorch-lightning/blob/338f3cf63686935355c749920b2f298f3d18a26f/pytorch_lightning/trainer/callback_hook.py#L55-L63

@carmocca @tchaton @awaelchli do you know how these hooks are used? Have you seen any examples of these being used by the community? These hooks go way, way back, but I can't think of when they'd be needed given the user handles the Trainer initialization. It's also unclear when `on_init_start` actually happens: does that mean callbacks should be the first thing initialized? 

We also have correctness issues for these hooks when callbacks are passed directly to the Trainer vs configured from the LightningModule: https://github.com/PyTorchLightning/pytorch-lightning/blob/f407a00cec59886cbaf8816630f6760e34926bd5/pytorch_lightning/core/lightning.py#L1140-L1142

Therefore, as a user, it seems more concise to write this:
```python
trainer = Trainer(...)
run_all_my_fancy_logic_now(trainer)

# use the trainer here
```

vs this:

```python

class MyCallback(Callback):
    def on_init_start(self, trainer) -> None:
        ...
    
    def on_init_end(self, trainer) -> None:
        ...
 
cb = MyCallback()
trainer = Trainer(callbacks=[cb])
```

cc @awaelchli @ananthsub @tchaton @carmocca @borda @ninginthecloud </div>

daniellepintz:
## What does this PR do?

Fixes #10894

## Before submitting

## PR review

## Did you have fun?


tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -81,9 +81,6 @@ def test_tqdm_progress_bar_totals(tmpdir):
 
     trainer = Trainer(default_root_dir=tmpdir, max_epochs=1)
     bar = trainer.progress_bar_callback
-    assert float("inf") == bar.total_train_batches
-    assert 0 == bar.total_val_batches
-    assert 0 == bar.total_test_batches
```
daniellepintz, line 86: Unfortunately we lose this functionality since these props can only be accessed once the `setup` hook has been called. Is this ok?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1479,7 +1479,13 @@ def _call_callback_hooks(
         hook_name: str,
         *args: Any,
         **kwargs: Any,
+<<<<<<< HEAD
```
ananthsub, line 1482: Bad merge

ananthsub (approved):

pytorch_lightning/callbacks/early_stopping.py
```diff
@@ -125,7 +125,7 @@ def __init__(
     def state_key(self) -> str:
         return self._generate_state_key(monitor=self.monitor, mode=self.mode)
 
-    def on_init_end(self, trainer: "pl.Trainer") -> None:
+    def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
```
rohitgr7, line 128: quick ques:
there is one more hook called after on_init_end and before setup, `on_before_accelerator_backend_setup`. Any reason why not use that one?

daniellepintz: no particular reason. I thought about it but thought `setup` is more generic and should work as well so went with setup. I can change it if you think it makes more sense

awaelchli: Probably because setup() is more natural and the instructions inside this hook does not rely on whether accelerator is setup or not.

pytorch_lightning/trainer/configuration_validator.py
```diff
@@ -290,3 +292,14 @@ def _check_dl_idx_in_on_train_batch_hooks(trainer: "pl.Trainer", model: "pl.Ligh
...
                     " The `dataloader_idx` argument will be removed in v1.7."
                 )
+
+
+def _check_on_init_start_end(trainer: "pl.Trainer") -> None:
+    """Checks if on_init_start/end are overridden and sends a deprecation warning."""
+    for callback in trainer.callbacks:
+        if is_overridden(method_name="on_init_start", instance=callback):
+            rank_zero_deprecation(
+                "The `on_init_start` callback hook was deprecated in v1.6 and will be removed in v1.8."
```
awaelchli, line 302: Should we mention to use setup() instead?

tests/deprecated_api/test_remove_1-8.py
```diff
@@ -41,3 +43,32 @@ def test_v1_8_0_deprecated_torchtext_batch():
...
+        def on_init_end(self, trainer):
+            print("Trainer is init now")
+
+    model = BoringModel()
+
+    trainer = Trainer(
+        callbacks=[TestCallback()],
+        max_epochs=1,
+        limit_val_batches=0.1,
+        limit_train_batches=0.2,
```
awaelchli, line 62:
```suggestion
        fast_dev_run=True,
```

awaelchli (approved):

pytorch_lightning/callbacks/early_stopping.py
rohitgr7: either is fine. was just looking for a reason if there were any :)

pytorch_lightning/trainer/configuration_validator.py
rohitgr7: IMO we shouldn't because that's up to the user to look for another hook since the `Callback.setup` is called after the environment is already being setup up and we can't say for sure what all crazy stuff user might be doing inside `on_init` functions which could fail if they go for `setup`.

rohitgr7 (approved):

pytorch_lightning/trainer/configuration_validator.py
daniellepintz: Sure, added

daniellepintz: oh sorry just saw your comment Rohit. I actually agree. Will put it back to how it was before

tchaton (approved):
LGTM !

PR #10957, opened 2021-12-06
Follow-up changes to #10575

carmocca:
## What does this PR do?

Follow-up to #10575

None

## Before submitting

## PR review

cc @carmocca @awaelchli @borda @ninginthecloud

### DIFF
--- a/pytorch_lightning/loops/epoch/training_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/training_epoch_loop.py
@@ -180,11 +180,10 @@ class TrainingEpochLoop(loops.Loop[_OUTPUTS_TYPE]):
 
             # hook
             self.trainer._call_callback_hooks("on_train_batch_start", batch, batch_idx, **extra_kwargs)
-            model_response = self.trainer._call_lightning_module_hook(
+            response = self.trainer._call_lightning_module_hook(
                 "on_train_batch_start", batch, batch_idx, **extra_kwargs
             )
-            ttp_response = self.trainer._call_ttp_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)
-            response = ttp_response if model_response is None else model_response
+            self.trainer._call_ttp_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)
             if response == -1:
                 self.batch_progress.increment_processed()
                 raise StopIteration
--- a/pytorch_lightning/loops/optimization/optimizer_loop.py
+++ b/pytorch_lightning/loops/optimization/optimizer_loop.py
@@ -320,7 +320,7 @@ class OptimizerLoop(Loop[_OUTPUTS_TYPE]):
             return None
 
         def backward_fn(loss: Tensor) -> None:
-            self.trainer.training_type_plugin.backward(loss, optimizer, opt_idx)
+            self.trainer._call_ttp_hook("backward", loss, optimizer, opt_idx)
 
             # check if model weights are nan
             if self.trainer._terminate_on_nan:
@@ -403,7 +403,7 @@ class OptimizerLoop(Loop[_OUTPUTS_TYPE]):
             optimizer: the current optimizer
             opt_idx: the index of the current optimizer
         """
-        self.trainer.training_type_plugin.optimizer_zero_grad(self.trainer.current_epoch, batch_idx, optimizer, opt_idx)
+        self.trainer._call_ttp_hook("optimizer_zero_grad", self.trainer.current_epoch, batch_idx, optimizer, opt_idx)
         self.optim_progress.optimizer.zero_grad.increment_completed()
 
     def _training_step(self, split_batch: Any, batch_idx: int, opt_idx: int) -> ClosureResult:
--- a/pytorch_lightning/loops/optimization/optimizer_loop.py
+++ b/pytorch_lightning/loops/optimization/optimizer_loop.py
@@ -144,12 +144,10 @@ class Closure(AbstractClosure[ClosureResult]):
                 )
 
             if self._zero_grad_fn is not None:
-                with self._profiler.profile("zero_grad"):
-                    self._zero_grad_fn()
+                self._zero_grad_fn()
 
             if self._backward_fn is not None and step_output.closure_loss is not None:
-                with self._profiler.profile("backward"):
-                    self._backward_fn(step_output.closure_loss)
+                self._backward_fn(step_output.closure_loss)
 
         return step_output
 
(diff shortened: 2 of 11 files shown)


pytorch_lightning/trainer/trainer.py
```diff
@@ -1537,14 +1537,20 @@ def _call_accelerator_hook(
         *args: Any,
         **kwargs: Any,
     ) -> Optional[Any]:
-        self.lightning_module._current_fx_name = hook_name
+        pl_module = self.lightning_module
```
daniellepintz, line 1540: why do we have to do all this, since previously we just had one line? 
https://github.com/PyTorchLightning/pytorch-lightning/blob/3d6262b7a91215e72019d720e742e6261e1636dc/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py#L220

https://github.com/PyTorchLightning/pytorch-lightning/blob/3d6262b7a91215e72019d720e742e6261e1636dc/pytorch_lightning/loops/optimization/manual_loop.py#L105

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -180,11 +180,10 @@ def advance(self, data_fetcher: AbstractDataFetcher) -> None:  # type: ignore[ov
 
             # hook
             self.trainer._call_callback_hooks("on_train_batch_start", batch, batch_idx, **extra_kwargs)
-            model_response = self.trainer._call_lightning_module_hook(
+            response = self.trainer._call_lightning_module_hook(
                 "on_train_batch_start", batch, batch_idx, **extra_kwargs
             )
-            ttp_response = self.trainer._call_ttp_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)
-            response = ttp_response if model_response is None else model_response
+            self.trainer._call_ttp_hook("on_train_batch_start", batch, batch_idx, **extra_kwargs)
```
daniellepintz, line 186: `_call_ttp_hook` is not directly annotated as returning None, nor is it supposed to - it is just missing the return type annotation. The reason it is missing the return type annotation is because it previously had `-> Optional[Any]` as return type but this caused a mypy error:
`pytorch_lightning/loops/optimization/optimizer_loop.py:439: error: Argument 1 to "_extract_hiddens" has incompatible type "Optional[Any]"; expected "Union[Tensor, Dict[str, Any]]"  [arg-type]`

so I removed it. But in order to maintain the current behavior of `call_hook`, ttp hooks _should_ be allowed to return something. see previous behavior:
https://github.com/PyTorchLightning/pytorch-lightning/blob/a28b4cd0c0bba30c21cae571e650877f66cf5588/pytorch_lightning/trainer/trainer.py#L1482-L1488

daniellepintz: Also from the original motivation in https://github.com/PyTorchLightning/pytorch-lightning/issues/8506, don't we want to have the flexibility to return something from ttp hooks?

carmocca: `_call_ttp_hook` can return something at the moment. But the concrete `on_train_batch_start` hook doesn't.

```python
$ grep -iIrn "on_train_batch_start" pytorch_lightning/plugins/training_type/
pytorch_lightning/plugins/training_type//ipu.py:325:    def on_train_batch_start(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None:
pytorch_lightning/plugins/training_type//training_type_plugin.py:456:    def on_train_batch_start(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None:
```

pytorch_lightning/trainer/trainer.py
carmocca: Previously was incomplete, we should always set and unset

pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -144,12 +144,10 @@ def closure(self, *args: Any, **kwargs: Any) -> ClosureResult:
                 )
 
             if self._zero_grad_fn is not None:
-                with self._profiler.profile("zero_grad"):
```
carmocca, line 147: If we get rid of `profile("training_step_and_backward")` then we could remove the profiler reference from this class

awaelchli (approved):

ananthsub (approved):

pytorch_lightning/loops/optimization/optimizer_loop.py
tchaton: I think we could remove it. PyTorch Profiler would track those operations already at the PyTorch level.

pytorch_lightning/trainer/trainer.py
tchaton: This is required for some nested hooks such as training_step with manual optimization.

tchaton (approved):
LGTM !

justusschock (approved):

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -164,8 +164,7 @@ def optimizer_step(
 
     def optimizer_zero_grad(self, current_epoch: int, batch_idx: int, optimizer: Optimizer, opt_idx: int) -> None:
         """Zeros all model parameter's gradients."""
-        model_ref = self.lightning_module
-        model_ref.optimizer_zero_grad(current_epoch, batch_idx, optimizer, opt_idx)
+        self.lightning_module.optimizer_zero_grad(current_epoch, batch_idx, optimizer, opt_idx)
```
daniellepintz, line 167: why not `_call_LM_hook`?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1479,16 +1479,15 @@ def _call_callback_hooks(
         hook_name: str,
         *args: Any,
         **kwargs: Any,
-    ) -> Optional[Any]:
-        output = None
+    ) -> None:
```
daniellepintz, line 1482: why are we returning None for callback hooks? I thought part of the motivation for https://github.com/PyTorchLightning/pytorch-lightning/issues/8506 was to allow returning something for callback hooks? i.e. #4 in the description

pytorch_lightning/plugins/training_type/training_type_plugin.py
carmocca: Because it's called already with `_call_ttp_hook`.

https://github.com/PyTorchLightning/pytorch-lightning/blob/eec2bae6c984895d1d4df4e75cf628c58128928a/pytorch_lightning/loops/optimization/optimizer_loop.py#L403

No TTP subclasses this hook, so technically the optimizer_loop could call `_call_LM_hook` directly and this could be removed

pytorch_lightning/trainer/trainer.py
carmocca: 4) was described as a limitation, but it's not something we want to support with the refactor. However, the refactor allows for adding it easily when we want to.

PR #10991, opened 2021-12-08
Fix support for logging within callbacks returned from `LightningModule`

rohitgr7:
## What does this PR do?

The logging support for callbacks is added before the callbacks from LightningModule are configured. The PR fixes that. 
~Also added a small support of returning a single callback from `LightningModule.configure_callbacks` rather than a list always.~

## Before submitting

## PR review

## Did you have fun?

Part of #1

cc @borda @awaelchli @ananthsub @daniellepintz @carmocca @edward-io

### DIFF
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -1091,13 +1091,13 @@ class Trainer(
         if hasattr(model, "hparams"):
             parsing.clean_namespace(model.hparams)
 
-        self._callback_connector._attach_model_callbacks(model)
-        self._callback_connector.attach_model_logging_functions(model)
-
-        verify_loop_configurations(self, model)
-
         # attach model to the training type plugin
         self.training_type_plugin.connect(model)
+
+        self._callback_connector._attach_model_callbacks()
+        self._callback_connector.attach_model_logging_functions()
+
+        verify_loop_configurations(self, model)
 
         # hook
         self._data_connector.prepare_data()
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -1092,10 +1092,9 @@ class Trainer(
             parsing.clean_namespace(model.hparams)
 
         self._callback_connector._attach_model_callbacks(model)
-        verify_loop_configurations(self, model)
-
-        # attach model log function to callback
         self._callback_connector.attach_model_logging_functions(model)
+
+        verify_loop_configurations(self, model)
 
         # attach model to the training type plugin
         self.training_type_plugin.connect(model)
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -1091,6 +1091,7 @@ class Trainer(
         if hasattr(model, "hparams"):
             parsing.clean_namespace(model.hparams)
 
+        self._callback_connector._attach_model_callbacks(model)
         verify_loop_configurations(self, model)
 
         # attach model log function to callback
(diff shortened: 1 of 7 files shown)


tests/trainer/connectors/test_callback_connector.py
```diff
@@ -211,6 +211,22 @@ def test_attach_model_callbacks_override_info(caplog):
...
+
+def test_attach_model_callbacks_support_logging(tmpdir):
+    class TempCallback(Callback):
+        def on_train_epoch_start(self, *args, **kwargs):
+            self.log("on_train_epoch_end", 7.0)
+
+    class CustomBoringModel(BoringModel):
+        def configure_callbacks(self):
+            cb = TempCallback()
+            return cb
```
tchaton, line 227:
```suggestion
            return TempCallback()
```

tchaton (approved):
Great catch !

awaelchli:
> Also added a small support of returning a single callback from LightningModule.configure_callbacks rather than a list always.

This would have to be a separate PR. We want to include only the fix in the bugfix release.

tests/trainer/connectors/test_callback_connector.py
```diff
@@ -211,6 +211,21 @@ def test_attach_model_callbacks_override_info(caplog):
     trainer.model = model
     cb_connector = CallbackConnector(trainer)
     with caplog.at_level(logging.INFO):
-        cb_connector._attach_model_callbacks()
+        cb_connector._attach_model_callbacks(model)
 
     assert "existing callbacks passed to Trainer: EarlyStopping, LearningRateMonitor" in caplog.text
+
+
+def test_attach_model_callbacks_support_logging(tmpdir):
```
awaelchli, line 219: I believe the test_hooks.py is sufficient and no extra test needs to be added.

rohitgr7: yeah but no tests were failing, right?

rohitgr7: oh ! okay. got it.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1091,6 +1091,7 @@ def _run(
         if hasattr(model, "hparams"):
             parsing.clean_namespace(model.hparams)
 
+        self._callback_connector._attach_model_callbacks(model)
```
carmocca, line 1094: why before ""verify_loop_configurations" and not under `attach_model_logging_functions`? That way you don't need to pass the pl_module reference as we want to remove the argument

cc @daniellepintz

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
rohitgr7: under `verify_loop_configurations` we have certain checks for callback hooks. So here now we make sure we check them for callbacks returned from `LightningModule` as well.

carmocca (approved):

pytorch_lightning/trainer/trainer.py
carmocca: Pushed a change reordering so the argument change is not necessary.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1091,17 +1091,16 @@ def _run(
...
-        # attach model log function to callback
-        self._callback_connector.attach_model_logging_functions(model)
-
         # attach model to the training type plugin
         self.training_type_plugin.connect(model)
 
+        self._callback_connector._attach_model_callbacks()
+        self._callback_connector._attach_model_logging_functions()
+
+        verify_loop_configurations(self, model)
```
rohitgr7, line 1100: thanks, @carmocca,
I guess with the changes you made we can remove the model from `verify_loop_configurations` as well?

Issue #5090, opened 2020-12-11
ModelCheckpoint fails at garbage collecting checkpoint passed to Trainer.resume_from_checkpoint


ORippler:
## 🐛 Bug

When passing a checkpoint to `Trainer` via `resume_from_checkpoint`, it is not tracked/garbage collected by `ModelCheckpoint` class. Instead, a new checkpoint is instantiated and gargabe collected/updated as usual.

## Please reproduce using [the BoringModel and post here](https://colab.research.google.com/drive/1HvWVVTK8j2Nj52qU4Q4YCyzOm0_aLQF3?usp=sharing)

https://colab.research.google.com/drive/1QJrLngpOZg1MOgAtZH5kRo_s6u-Hjh0n?usp=sharing

### Expected behavior

Checkpoint passed to Trainer.resume_from_checkpoint is garbage collected.
If this is not desired behavior, I think a sentence or 2 should be added to the docs on the intended behavior.

### Environment
* CUDA:
	- GPU:
		- Tesla T4
	- available:         True
	- version:           10.1
* Packages:
	- numpy:             1.18.5
	- pyTorch_debug:     True
	- pyTorch_version:   1.7.0+cu101
	- pytorch-lightning: 1.1.0
	- tqdm:              4.41.1
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.6.9
	- version:           #1 SMP Thu Jul 23 08:00:38 PDT 2020

### Additional context

If this is due to epoch mismatching and not a design choice, #5007 #4655 #2401 could be possibly related.

haideraltahan:
Having the same issue. I believe this problem could be related to #5091
@ORippler is there a temporary solution?

ORippler:
I don't know how this is related to your problem, if your model is always initialized from scratch/fails resuming from a given checkpoint.

Resuming works properly for me, however the Checkpoint passed to `Trainer` via `resume_from_checkpoint` is no longer removed during the garbage collection (refer also the Colab reproduction linked above).

My temporary solutions to this is to perform the GC manually (i.e. removing the checkpoint passed to `resume_from_training`). I present a hotfix to #5091 inside the issue

Issue #4335, opened 2020-10-24
Clarify the model checkpoint arguments


ananthsub:
## 🐛 Proposals

This is not so much a bug report as an RFC to clarify the `ModelCheckpoint` callback arguments:
- `save_last`: to me, this means that whenever we save a checkpoint, we save a checkpoint with filename `"last.ckpt"`. This provides a pre-determined checkpoint name, which is very helpful for resuming from failures. Importantly, it should not determine *when* checkpoints are saved. Currently it's easy to confuse this parameter to mean "save the checkpoint after the last epoch," which I think should be split out as a separate argument. This distinction would also clarify the typing and validation: there's no need for it to be an `Optional[bool]`: either we save a checkpoint as `"last.ckpt"` or not. So it could be a regular `bool`. 

- There's an inefficiency right now where we dump the checkpoint twice if `save_last=True`. For techniques like ZeRO and sharded optimizer states, each checkpoint save triggers communications across all ranks. Instead, we should dump the checkpoint dict once, and then save to different file paths accordingly (cc @SeanNaren)

- `save_top_k`: since `monitor` is `None` by default, this should force `save_top_k` to be -1. The counterargument is that this can cause storage concerns. But I think this is easily correctable on the user-side: configure `save_top_k` + `monitor` 

- `period`: we should rename this as `every_n_epochs`. this opens up extensions for checkpointing after `every_n_steps` during training and checkpointing after a specified time interval. With those extensions in mind, `period` is ambiguous.

cc @awaelchli

awaelchli:
save_last: I agree, I believe this was the original intention when this feature was added. Definitely useful to have.
period: Regarding naming, we have a PR here: #3807 which needs to be finished, but implements exactly this :)

carmocca:
- `save_last`:  What about having a `symlink_to_last: bool = False` argument so there is always a symbolic link to the last saved epoch? It'd also avoid the inefficiency you talk about in your second point.
- `period`: I think we should keep the name `period` and have a mechanism to either track epochs or steps.

ananthsub:
> save_last: What about having a symlink_to_last: bool = False argument so there is always a symbolic link to the last saved epoch? It'd also avoid the inefficiency you talk about in your second point.

I didn't see symlink supported from the fsspec API, so this could be challenging without that. but yeah that's the general idea: maybe we call it `save_as_last` ? `save_most_recent` ? open to name ideas here haha 

[...]

PR #10995, opened 2021-12-08
Add required states for resumed ModelCheckpoint GC

### LINKED ISSUE #4911: ModelCheckpoint Callback save and restore extension
## 🚀 Feature

callbacks.ModelCheckpoint implements on_save_checkpoint and on_load_checkpoint to save and load it's state to/from a checkpoint. Save/Load all internal members to recover complete state from checkpoint. 

### Motivation

I use a ModelCheckpoint callback to save my Top-K (3) models during training. Sometimes I resume training at a certain point and want to train further epochs. But the loaded Callback has an empty best_k_models list, so this list is build up again. Older models are not tracked and I get three new models in same folder, and I do not know, which are the Top-3 ones. 
Therefore a full state of ModelCheckpoint would resolve this. 

### Pitch

Extend the ModelCheckpoint methods to export and load all internal states. I'm not sure if there is missing something, but here is a try:

    def on_save_checkpoint(self, trainer, pl_module) -> Dict[str, Any]:
        return {
            "best_model_score": self.best_model_score,
            "best_model_path": self.best_model_path,
            "best_k_models" : self.best_k_models,
        }

    def on_load_checkpoint(self, checkpointed_state: Dict[str, Any]):
        self.best_model_score = checkpointed_state["best_model_score"]
        self.best_model_path = checkpointed_state["best_model_path"]
        self.best_k_models = checkpointed_state["best_k_models"]

### Alternatives

Subclass the ModelCheckpoint to implement this, but I think this would complicate things.  

### Additional context

-/-

ORippler:
## What does this PR do?

Fixes #4911
Related: #5090

Currently, when resuming training the internal states required for continued ModelCheckpointing are neither saved nor restored. This leads to the fact that `k` new checkpoints are always generated due to this [check](https://github.com/PyTorchLightning/pytorch-lightning/blob/99adc45af136089abae8a1f6dc5a09e041d68605/pytorch_lightning/callbacks/model_checkpoint.py#L478-L480). These new checkpoints are properly gced/compared to against, but the old ones are not.

Note that this PR does not handle overrides of `monitor`, `dirpath` or `mode`, as also referred to in #4911

It might be that resuming training fails now if it did not fail before, if the paths were changed in the mean time (refer also #4911). I did not check/test for this, but confirmed that resumption of GC now works properly

## Before submitting

## PR review

## Did you have fun?

cc @carmocca @awaelchli @ninginthecloud @jjenniferdai


justusschock (commented):
@ORippler thanks for this fix. To avoid regression again, we need a test for this.

Do you think the following does reflect this issue sufficiently (If so, feel free to take it and commit it directly to your branch):

```python
def test_model_checkpoint_attributes(tmpdir):
    seed_everything()
    model = LogInTwoMethods()

    epochs = 2
    checkpoint_callback = ModelCheckpoint(monitor=None, dirpath=tmpdir, save_top_k=-1, save_last=True)
    trainer = Trainer(
        default_root_dir=tmpdir,
        callbacks=[checkpoint_callback],
        limit_train_batches=10,
        limit_val_batches=10,
        max_epochs=epochs,
        logger=False,
    )

    trainer.fit(model)

    checkpoint = torch.load(os.path.join(tmpdir, 'last.ckpt'))['callbacks'][checkpoint_callback.state_key]
    for k in ("best_models, kth_best_model_path", "kth_value", "last_model_path"):
        assert checkpoint[k] == getattr(checkpoint_callback, k)
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -342,13 +342,21 @@ def on_save_checkpoint(
...
 
     def on_load_checkpoint(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", callback_state: Dict[str, Any]
     ) -> None:
         self.best_model_score = callback_state["best_model_score"]
         self.best_model_path = callback_state["best_model_path"]
+        self.best_k_models = callback_state["best_k_models"]
+        self.kth_best_model_path = callback_state["kth_best_model_path"]
+        self.kth_value = callback_state["kth_value"]
+        self.last_model_path = callback_state["last_model_path"]
```
justusschock, line 359: This is required for backwards compatibility with legacy checkpoints who don't have this.

```suggestion
        self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
        self.kth_best_model_path = callback_state.get("kth_best_model_path", self.kth_best_model_path)
        self.kth_value = callback_state.get("kth_value", self.kth_value)
        self.last_model_path = callback_state.get("last_model_path", self.last_model_path)
```

ORippler: Thanks !

ORippler:
> @ORippler thanks for this fix. To avoid regression again, we need a test for this.
> 
> Do you think the following does reflect this issue sufficiently (If so, feel free to take it and commit it directly to your branch):
> 
> ```python
> def test_model_checkpoint_attributes(tmpdir):
>     seed_everything()
>     model = LogInTwoMethods()
> 
>     epochs = 2
>     checkpoint_callback = ModelCheckpoint(monitor=None, dirpath=tmpdir, save_top_k=-1, save_last=True)
>     trainer = Trainer(
>         default_root_dir=tmpdir,
>         callbacks=[checkpoint_callback],
>         limit_train_batches=10,
>         limit_val_batches=10,
>         max_epochs=epochs,
>         logger=False,
>     )
> 
>     trainer.fit(model)
> 
>     checkpoint = torch.load(os.path.join(tmpdir, 'last.ckpt'))['callbacks'][checkpoint_callback.state_key]
>     for k in ("best_models, kth_best_model_path", "kth_value", "last_model_path"):
>         assert checkpoint[k] == getattr(checkpoint_callback, k)
> ```

How would this integrate with the different functionality tests for `ModelCheckpoint` the overall testing framework ? For example, I see many tests checking whether attributes are written to the ckpt properly, but not whether they are loaded.

For example [here](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/checkpointing/test_model_checkpoint.py#L178), we check for `current_state` being written to disk/to the ckpt, but the current `on_load_checkpoint` never sets this attribute again.

Is this not something we want to test also? Am a bit confused here.

justusschock:
@ORippler that's true. I suppose for end-to-end testing you would have to extend that test by resuming it at a freshly created trainer and examining the properties there.

@carmocca do we want to test for different parametrizations of the callback here?

ORippler:
> @ORippler that's true. I suppose for end-to-end testing you would have to extend that test by resuming it at a freshly created trainer and examining the properties there.
> 
> @carmocca do we want to test for different parametrizations of the callback here?

I added your test and expanded it to check whether a freshly instantiated `ModelCheckpoint` also loads the properties.

This is still not a fully functional test though.

Off-Note: Do we have a test that compares for equivalence of results generated by one continuous training run and an interrupted one that is resumed by passing the checkpoint to `trainer.fit` ? This would be nice to have imo

justusschock:
@ORippler we have https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/trainer/test_trainer.py#L399 which doesn't check the results. However, this is on purpose as the checkpoint does not include any random data and thus continuing from the checkpoint doesn't have to yield the **exact** same results until now (this is currently in development).

cc @tchaton to add a similar test once fault tolerance is ready

tchaton:
Hey @ORippler,

Yes, we have multiple tests checking the weights are the same before and after for Fault Tolerance. Here they are: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/utilities/test_auto_restart.py

justusschock (approved):

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1202,3 +1202,32 @@ def test_check_val_every_n_epochs_top_k_integration(tmpdir):
     )
     trainer.fit(model)
     assert set(os.listdir(tmpdir)) == {"epoch=1.ckpt", "epoch=3.ckpt"}
+
+
+def test_model_checkpoint_attributes(tmpdir):
```
rohitgr7, line 1207: I think it's better to update this existing test with all the arguments that are being restored.
https://github.com/PyTorchLightning/pytorch-lightning/blob/5576fbc5f9a7d0bc71ad26b8b54775110e675808/tests/models/test_restore.py#L269-L272

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -342,13 +342,21 @@ def on_save_checkpoint(
...
+            "kth_value": self.kth_value,
+            "last_model_path": self.last_model_path,
         }
 
     def on_load_checkpoint(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", callback_state: Dict[str, Any]
     ) -> None:
         self.best_model_score = callback_state["best_model_score"]
         self.best_model_path = callback_state["best_model_path"]
+        self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
```
rohitgr7, line 356: why `.get`?
If those are saved. They should be available?

justusschock: Yes, but if you don't use get, this won't work with legacy checkpoints :)

rohitgr7: good point 👍 
do we have an interval or some sort of minimal version wrt to the current pl version until which we support legacy checkpoints?

justusschock: We test for all versions since 1.0.0

tests/checkpointing/test_model_checkpoint.py
ORippler: Imo we could differentiate between a:
* Unit tests that checks whether states are written to disk properly (i.e. testing `ModelCheckpoint.on_save_checkpoint` and `ModelCheckpoint.on_load_checkpoint`).
* The existing functional test that checks whether model states are restored properly when providing `ckpt` to `trainer.fit`.

Regarding the attbributes that are restored: Why don't we restore `current_score`, `monitor` and `dirpath` attributes that are written to `ckpt`. Is there an explicit reason for this ?

rohitgr7: `current_score` isn't required to be reloaded because it will eventually change right after the next call, and for `dirpath`, even while resuming it can be different and there won't be any problem. For `monitor`, I guess it can be changed too since we don't force users to use the old one but then best_model_path/score would be irrelevant.

Regarding unit testing, I am just concerned about repetitive tests. We test that while resuming `on_save/load_checkpoint` is called, elsewhere, all we need to test what's being saved and reloaded which I guess the existing test covers up.

awaelchli: I definitely prefer unit testing here. There are already plenty of tests that ensure that "checkpointing" works. 
Here, we would only want to instantiate the ModelCheckpoint and call the hooks to test their function (no trainer.fit, no loading checkpoint).

rohitgr7: I'd been fine with adding unit tests here only if there wasn't any existing test here that checks what we need to but adding another similar test makes it hard later to track which tests to update in case some new stuff gets added just like we missed here and later found out there's already a similar existing test.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -342,13 +342,21 @@ def on_save_checkpoint(
...
+            "best_k_models": self.best_k_models,
+            "kth_best_model_path": self.kth_best_model_path,
+            "kth_value": self.kth_value,
+            "last_model_path": self.last_model_path,
         }
 
     def on_load_checkpoint(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", callback_state: Dict[str, Any]
     ) -> None:
         self.best_model_score = callback_state["best_model_score"]
```
carmocca, line 354: Should this check that the same monitor is being used and only load in that case?

justusschock: @carmocca I think yes, but I also think that this might go beyond this PR and should be done in a separate PR. This PR is only a bugfix release and IMO your approach changes the behavior and would thereby go to feature releases.

carmocca: Oh right, I forgot about this magic :P

tests/checkpointing/test_model_checkpoint.py
ORippler: Hey, I just fixed the broken funciontality test `test_callbacks_state_fit_ckpt_path` (`ModelCheckpoint` actually never saved/triggered, since trainer was fitted with `max_steps = 1`  and `ModelCheckpoint` configured to trigger after every epoch).

However, tests fail now. I inspected states manually and `before` states are updated properly now, whereas `after` states remain empty/uninitialized (before my fix, both `before` and `after` states were same as upon their initialization). Any ideas why `before` isn't restored properly for the test ? This might hint at another bug somewhere in the code.

I would afterwards extend the test with the additional keys and then refactor the unit-test to do pure testing of `ModelCheckpoint.on_save_checkpoint` and `ModelCheckpoint.on_load_checkpoint`

justusschock: Hey @ORippler ,

You are correct and the test did not do what it was supposed to do any more. Updating `max_steps` to `max_epochs` seems correct to me (maybe we could additionally limit the number of batches here to 1 or 2 to make it faster?)

After adjusting the test it failed, because `save_on_train_epoch_end` was not specified and thereby internally set to True. However this caused the restore to fail since the state_keys did not match any longer. I fixed it in this test, but this is likely an issue beyond this test whenever `save_on_train_epoch_end` is set to `None` since we restore before this would be evaluated internally.
cc @awaelchli

awaelchli: Thanks for investigating guys and sorry, this was caused by a change in master recently (no release affected afaik). Let's take care of this in https://github.com/PyTorchLightning/pytorch-lightning/pull/11108

awaelchli: Resolved, fixed in master and updated this branch. Thanks

carmocca: This conversation got a bit sidetracked, but I agree with @rohitgr7 in that the added test overlaps a lot with the modified test

awaelchli: So, as the remainder of the discussion here, are we going with a unit test to replace this current test here?
Pseudo code:

```py
cb = ModelCheckpoint(...)
x = {}
cb.on_save_checkpoint(x)

restored_cb = ModelCheckpoint(...)
restored_cb.on_load_checkpoint(x)

for k in ("best_k_models", "kth_best_model_path", "kth_value", "last_model_path"):
        assert cb[k] == getattr(restored_cb, k)
```

No expensive training or saving and loading to disk

awaelchli: Thoughs?

rohitgr7: yes, better remove the old test since it's just checking for ModelCheckpoint states and nothing else.

tchaton: Yes, looks better to me.

ORippler: I replaced the test with a unit-test. Only question I stil have is why not all states returned by `on_save_checkpoint` are restored during `on_load_checkpoint` ([specifcially `current_score`, `dirpath` and `mode`](https://github.com/ORippler/pytorch-lightning/blob/fix/4911/pytorch_lightning/callbacks/model_checkpoint.py#L345-L346)). What I get is that `mode` is used for the state_key generation and therefore equivalence is already checked for this, but what about `current_score' and `dirpath` ?

carmocca (approved):

tests/checkpointing/test_model_checkpoint.py
awaelchli: There may be some confusion here. Don't get mislead by state_key, it does not matter.

The current_score does not get restored because you didn't implement that yet. You save it but you don't load it, so you need to fix it.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -344,13 +344,21 @@ def on_save_checkpoint(
             "best_model_path": self.best_model_path,
             "current_score": self.current_score,
```
awaelchli, line 345: current_score does not get loaded? I wonder if there was a reason for that

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1202,3 +1202,37 @@ def test_check_val_every_n_epochs_top_k_integration(tmpdir):
     )
     trainer.fit(model)
     assert set(os.listdir(tmpdir)) == {"epoch=1.ckpt", "epoch=3.ckpt"}
+
+
+def test_model_checkpoint_loadsave_ckpt(tmpdir):
```
awaelchli, line 1207: shouldn't the test first save then load? 
Also, it would be best if you re-instantiate the modelcheckpoint callback for each call.

pytorch_lightning/callbacks/model_checkpoint.py
rohitgr7: current_score changes at every iteration and is not used for anything else. Does it need a reload?

awaelchli: Yeah makes sense, I guess it just gets saved to the checkpoint so that it can be inspected if needed.  👍

tests/checkpointing/test_model_checkpoint.py
ORippler: > shouldn't the test first save then load? Also, it would be best if we re-instantiate the modelcheckpoint callback for each call.

Since these are pure unit tests the sequence strictly doesn't really matter that much (but I guess makes them easier to follow).

I applied your suggestions and clarified/documented that we currently don't restore all states that are written to disk.

tests/checkpointing/test_model_checkpoint.py
ORippler: @awaelchli Can you give an update whether `current_score` should be restored ? Currently, I leave it as is and just comment in the test that it's not restored. Might also do a different PR for this, since it's not directly related to #4911

awaelchli: As pointed out by @rohitgr7 https://github.com/PyTorchLightning/pytorch-lightning/pull/10995/files#r771482266 it is not required to restore the score.

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1202,3 +1202,37 @@ def test_check_val_every_n_epochs_top_k_integration(tmpdir):
...
+        setattr(cb_write, key, val)
+    written_ckpt = cb_write.on_save_checkpoint("", "", "")
+    for state in ckpt:
+        assert ckpt[state] == written_ckpt[state]
+
+    # test on_load_checkpoint
+    # Note: "current_score", "dirpath" and "monitor" are currently not restored by on_load_checkpoint.
+    # We therefore set "dirpath" and "monitor" to something different than for ckpt/cb_write so we can assert them.
+    # "current_score" is left as initialized, i.e. None, and can therefore also be asserted
+    cb_restore = ModelCheckpoint(dirpath=tmpdir + "restore", monitor=None, save_top_k=-1, save_last=True)
```
awaelchli, line 1232: don't understand this part, the dirppath and monitor don't have anything to do with the score :)

awaelchli (approved):
LGTM

tests/checkpointing/test_model_checkpoint.py
ORippler: True, `dirpath` and `monitor` are unrelated to `current_score`.

What I try to explain here is why we need to change `dirpath` and `monitor` for `cb_restore` compared to `cb_write` in order to check that they have not been changed by `.on_load_checkpoint`, and why we can leave `current_score` as is (it's set to `None` during `__init__` and can thus be already used to check that it's not been modified).

justusschock: It just states that this is intended behavior. I think this comment is fine :)

awaelchli: It makes sense, I would have written the test in a way that loads completely new values that we can assert regardless of the init values but your way is not wrong :)

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -344,13 +344,21 @@ def on_save_checkpoint(
...
             "dirpath": self.dirpath,
+            "best_k_models": self.best_k_models,
+            "kth_best_model_path": self.kth_best_model_path,
+            "kth_value": self.kth_value,
+            "last_model_path": self.last_model_path,
```
carmocca, line 350: I just noticed an issue with doing this.

Since we save each "ModelCheckpoint" mode sequentally, these attributes will not be correct depending on the order if more than 1 mode triggers a save for the same global step:

https://github.com/PyTorchLightning/pytorch-lightning/blob/fe940e195dceb18eb9f3bd512cea56ae3405d464/pytorch_lightning/callbacks/model_checkpoint.py#L366-L373

Currently, a "top-k" checkpoint will not include the `last_model_path` path even if it's saved right after for this global step.

I'm not sure what would be the best solution here. I think we should start recommending multiple `ModelCheckpoint` instances as a best practice because these interactions between flags can be unintuitive.

cc @awaelchli @ananthsub @jjenniferdai  
Related to #4335 and https://github.com/PyTorchLightning/pytorch-lightning/pull/11805#issuecomment-1067972348

Issue #10821, opened 2021-11-29
Generalize internal checks for precision plugin type, training type, accelerator type 


awaelchli:
## Proposed refactor

Internally, our checks against the type of Accelerator, Precision type, strategy is not robust towards custom instances passed in by the user. 

### Motivation

Internally, some operations in the optimization, logging, etc. need a different code path depending on 1) Accelerator type (cpu, gpu) or 2) Precision type (apex, native) or 3) strategy type (ddp, ddp-spawn, ...). Currently we have this pattern: 

```py
if trainer._device_type == DeviceType.CPU:
    # do something only for cpu


if trainer._amp_backend == AMPType.Apex:
    # do something differently for apex
```

### Pitch

Change these to

```py
if isinstance(trainer.accelerator, CPUAccelerator):
    # do something only for cpu


if isinstance(trainer.precision_plugin, ApexPrecisionPlugin):
    # do something differently for apex
```

This has the benefits:
- User passes in custom plugins (subclasses of our plugins)
- Encapsulation: Protected members _device_type, _strategy_type, won't be abusively accessed publicly anymore. They remain an implementation detail of AcceleratorConnector
- Minimally simplifies AcceleratorConnector logic

### Additional context

Discusson started in https://github.com/PyTorchLightning/pytorch-lightning/pull/10596

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

[...]

Issue #10410, opened 2021-11-08
[RFC] Future of `gpus/ipus/tpu_cores` with respect to `devices`


SeanNaren:
## Proposed refactoring or deprecation

Currently we have two methods to specifying devices. Let's take GPUs for example:

1. The standard case that we've all grown used to and are mostly aware of.
```python
trainer = Trainer(gpus=2)
```

2. Introduced in 1.5, tries to make the number of devices agnostic. This means if you specify `accelerator='tpu'` we automatically know to use 2 TPU cores.
```python
trainer = Trainer(devices=2, accelerator='gpu')
```

Recently, it has come up in https://github.com/PyTorchLightning/pytorch-lightning/pull/10404#discussion_r744562512 that we may want to deprecate and prevent further device specific names from appearing in the Trainer (such as `hpus`).

Related conversation https://github.com/PyTorchLightning/pytorch-lightning/issues/9053#issuecomment-904239610

I see two options:

🚀 We keep both device specific arguments (`gpus` `tpu_cores` `ipus` for the Trainer) and `devices`
👀 We drop `gpus` `tpu_cores` `ipus` in the future and fully rely on `devices`. (*Potentially this would likely be done in Lightning 2.0, instead of after 2 minor releases*)

cc @kaushikb11 @justusschock @ananthsub @awaelchli

ananthsub:
IMO we should follow the contributing guidelines: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md#main-core-value-one-less-thing-to-remember

Having multiple options in the public API to do the same thing is really confusing.

four4fish:
+1, totally agree

Current device related flags are confusing. Multiple flags partially overlap and interfere each other. When multiple flags passed in, we define prioritize and ignore some of the flags. 

For example: 
`gpu=2, device=3` `device` will be ignored.
`gpu=2, cpu=2, accelerator=cpu` what will happen? I think cpu with num_process=2?

I prefer option 2, drop `gpus`, `tpu_cores`,  `ipus` in the future and fully rely on `devices`
And  can we have `devices`  be `int`, not set =`auto`?
With this option `accelerator` flag for device_type, `devices` (probably rename to devices_num?) for device_number. It's also scalable for new device types like hpus

williamFalcon:
I think going from this:
```
Trainer(gpus=2)
```

to

```
[...]

PR #11001, opened 2021-12-09
1/n Generalize internal checks for Accelerator in Trainer - remove trainer._device_type

four4fish:
## What does this PR do?
#11002 could be follow ups for this

Part of #10821

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -27,7 +27,7 @@ import torch
 from torch.optim import Optimizer
 
 import pytorch_lightning as pl
-from pytorch_lightning.accelerators import Accelerator, IPUAccelerator
+from pytorch_lightning.accelerators import Accelerator, GPUAccelerator, IPUAccelerator, TPUAccelerator
 from pytorch_lightning.callbacks import Callback, EarlyStopping, ModelCheckpoint, ProgressBarBase
 from pytorch_lightning.callbacks.prediction_writer import BasePredictionWriter
 from pytorch_lightning.core.datamodule import LightningDataModule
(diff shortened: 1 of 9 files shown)


pytorch_lightning/trainer/trainer.py
```diff
@@ -1650,33 +1650,31 @@ def __setup_profiler(self) -> None:
...
                 "TPU available but not used. Set the `tpu_cores` flag in your trainer"
                 " `Trainer(tpu_cores=8)` or script `--tpu_cores=8`."
             )
 
-        if (
-            _IPU_AVAILABLE
-            and self._device_type != _AcceleratorType.IPU
-            and not isinstance(self.accelerator, IPUAccelerator)
-        ):
+        if _IPU_AVAILABLE and not isinstance(self.accelerator, IPUAccelerator):
```
ananthsub, line 1677: i find this logging pretty verbose. For instance, I don't want to  see `"TPU available: False, using: 0 TPU cores"` on every run. This is only useful if there are TPUs available and I'm not using them. I assume that if there are not TPUs available but I have specified them, then an exception would be thrown before this is called.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1650,33 +1650,31 @@ def __setup_profiler(self) -> None:
...
-            self.tpu_cores if self.tpu_cores is not None and self._device_type == _AcceleratorType.TPU else 0
+            self.tpu_cores if self.tpu_cores is not None and isinstance(self.accelerator, TPUAccelerator) else 0
         )
         rank_zero_info(f"TPU available: {_TPU_AVAILABLE}, using: {num_tpu_cores} TPU cores")
 
         num_ipus = self.ipus if self.ipus is not None else 0
         rank_zero_info(f"IPU available: {_IPU_AVAILABLE}, using: {num_ipus} IPUs")
 
-        if torch.cuda.is_available() and self._device_type != _AcceleratorType.GPU:
+        if torch.cuda.is_available() and isinstance(self.accelerator, GPUAccelerator):
```
ananthsub, line 1665: this is incorrect. the prior code was checking if the GPU accelerator was *not* being used.
But I don't think we need to check the accelerator instance at all. Why can't we check the `gpus` flag directly?

Same as above, I don't get why we have to check TPU Accelerator instead of `tpu_cores`

pytorch_lightning/trainer/trainer.py
kaushikb11: Fair point. What about when users pass `(accelerator="auto", devices="x")`?

justusschock: can we simply lower the logging level to maybe debug here? I also find it to verbose most of the times.

@kaushikb11 same goes here. Usually people know on what kind of machine they submitted their job.

pytorch_lightning/trainer/trainer.py
four4fish: With ppl reach agreement on gpu/cpus vs devices in issue #10410 Future of gpus/ipus/tpu_cores with respect to devices, I think check accelerator and device number will be better than tpu/gpus.

pytorch_lightning/trainer/trainer.py
four4fish: Totally agree it's too verbose, and why do we have _log_device_info() with all the warnings in trainer? Should we move _log_device_info to the end of the accelerator_connector init, or having the warning relocate to select functions in accelerator connector?

pytorch_lightning/trainer/trainer.py
tchaton, line 1665:
```suggestion
        if torch.cuda.is_available() and not isinstance(self.accelerator, GPUAccelerator):
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -1650,33 +1650,31 @@ def __setup_profiler(self) -> None:
...
             rank_zero_warn(
                 "GPU available but not used. Set the gpus flag in your trainer `Trainer(gpus=1)` or script `--gpus=1`.",
                 category=PossibleUserWarning,
             )
 
-        if _TPU_AVAILABLE and self._device_type != _AcceleratorType.TPU:
+        if _TPU_AVAILABLE and isinstance(self.accelerator, TPUAccelerator):
```
tchaton, line 1671:
```suggestion
        if _TPU_AVAILABLE and not isinstance(self.accelerator, TPUAccelerator):
```

pytorch_lightning/trainer/trainer.py
four4fish: created a issue #11014 to discuss and will address it in a different PR

Issue #11002, opened 2021-12-09
Remove trainer._device_type in favor of check Accelerator class


four4fish:
## Proposed refactor

Follow up for #11001: Generalize internal checks for precision plugin type, training type, accelerator type

### Motivation

Code simplification 

### Pitch

After #11001, _device_type is not needed anymore

```
    @property
    def _device_type(self) -> _AcceleratorType:
        return self._accelerator_connector.
```

Instead in tests and where we need check device type, use
```
isinstance(trainer.accelerator, XAccelerator)
```

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @justusschock @awaelchli @akihironitta

awaelchli:
Hey @four4fish 
Thanks for looking at this.
This is not something we need to deprecate because it's already declared as internal, not exposed to user. 
It can simply be removed together with the last part of #11001

four4fish:
Solved by https://github.com/PyTorchLightning/pytorch-lightning/pull/11992

Issue #10422, opened 2021-11-08
[RFC] Simplifying the Accelerator Connector logic and flags 


four4fish:
## Proposed refactoring or deprecation

No.5 of #10416  Accelerator and Plugin refactor
and part of #10417 Core Trainer Connectors
Related to #10410 Future of gpus/ipus/tpu_cores with respect to devices

### Motivation

Current flags and accelerator logic is confusing. Multiple Accelerator flags partially overlap and interfere each other. 

There are 30 MisconfigurationException in accelerator connector, half of them are caused by duplicated flags interfering with each other.

Multiple flag with same meaning doesn't add much value, but cause confusing and make accelerator_connector logic unnecessarily complicated.

**For example:**
1. The `devices` flags mentioned in #10410 Future of gpus/ipus/tpu_cores with respect to devices
`gpu=2, device=3`, the device will be ignored.
or

https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bd4fbd96c1e73e251859d99b207008008de87d/pytorch_lightning/trainer/connectors/accelerator_connector.py#L867-L869

2. Accelerator could have multiple meanings. It could be `device` which is a string, or `Accelerator()` which wraps the  precision and ttp.

https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bd4fbd96c1e73e251859d99b207008008de87d/pytorch_lightning/trainer/connectors/accelerator_connector.py#L784-L791

3.  Plugins and strategy are duplicated. If user specific both will be misconfig. Also we have to keep logic to handle both `strategy flag` and `plugin flag`. (There is `distributed_backend` too and it's deprecated)
    
    https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bd4fbd96c1e73e251859d99b207008008de87d/pytorch_lightning/trainer/connectors/accelerator_connector.py#L317-L322
    
    
    
Also, with increasing use cases of custom Plugins, it's critical to have more scalable solutions. For example, the current enum for distributed is not scalable for customized distributed
https://github.com/PyTorchLightning/pytorch-lightning/blob/db4e7700047519ff6e6365517d7e592c8ef023cb/pytorch_lightning/utilities/enums.py

### Pitch

**Every Flag should have One and Only One meaning, NO overlap between flags. Reduce user's Misconfig possibility** 
Deprecate `num_processes`, `tpu-cores`, `ipus`, `gpus`, `plugins` flag
Keep options:
```
[...]

PR #11022, opened 2021-12-10
3/n Move accelerator into Strategy

### LINKED ISSUE #10648: [Accelerator refactor] Move Accelerator into Strategy

four4fish:
## What does this PR do?
Last step of Move accelerator into Strategy

Fixes #10648 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/docs/source/extensions/accelerators.rst
+++ b/docs/source/extensions/accelerators.rst
@@ -25,10 +25,7 @@ One to handle differences from the training routine and one to handle different 
     from pytorch_lightning.accelerators import GPUAccelerator
     from pytorch_lightning.plugins import NativeMixedPrecisionPlugin, DDPPlugin
 
-    accelerator = GPUAccelerator(
-        precision_plugin=NativeMixedPrecisionPlugin(precision=16, device="cuda"),
-        training_type_plugin=DDPPlugin(),
-    )
+    accelerator = GPUAccelerator()
     trainer = Trainer(accelerator=accelerator)
 
 
--- a/docs/source/extensions/accelerators.rst
+++ b/docs/source/extensions/accelerators.rst
@@ -26,8 +26,8 @@ One to handle differences from the training routine and one to handle different 
     from pytorch_lightning.plugins import NativeMixedPrecisionPlugin, DDPPlugin
 
     accelerator = GPUAccelerator()
-    precision_plugin=NativeMixedPrecisionPlugin(precision=16, device="cuda")
-    training_type_plugin=DDPPlugin(accelerator=accelerator, precision_plugin=precision_plugin)
+    precision_plugin = NativeMixedPrecisionPlugin(precision=16, device="cuda")
+    training_type_plugin = DDPPlugin(accelerator=accelerator, precision_plugin=precision_plugin)
     trainer = Trainer(strategy=training_type_plugin)
 
 
--- a/docs/source/extensions/accelerators.rst
+++ b/docs/source/extensions/accelerators.rst
@@ -26,7 +26,9 @@ One to handle differences from the training routine and one to handle different 
     from pytorch_lightning.plugins import NativeMixedPrecisionPlugin, DDPPlugin
 
     accelerator = GPUAccelerator()
-    trainer = Trainer(accelerator=accelerator)
+    precision_plugin=NativeMixedPrecisionPlugin(precision=16, device="cuda")
+    training_type_plugin=DDPPlugin(accelerator=accelerator, precision_plugin=precision_plugin)
+    trainer = Trainer(strategy=training_type_plugin)
 
 
 We expose Accelerators and Plugins mainly for expert users who want to extend Lightning to work with new
--- a/docs/source/extensions/plugins.rst
+++ b/docs/source/extensions/plugins.rst
@@ -80,10 +80,7 @@ can then be passed into the Trainer directly or via a (custom) accelerator:
     trainer = Trainer(strategy=CustomDDPPlugin(), plugins=[CustomPrecisionPlugin()])
 
     # fully custom accelerator and plugins
-    accelerator = MyAccelerator(
-        precision_plugin=CustomPrecisionPlugin(),
-        training_type_plugin=CustomDDPPlugin(),
-    )
+    accelerator = MyAccelerator()
     trainer = Trainer(accelerator=accelerator)
 
 
--- a/docs/source/extensions/plugins.rst
+++ b/docs/source/extensions/plugins.rst
@@ -81,7 +81,9 @@ can then be passed into the Trainer directly or via a (custom) accelerator:
 
     # fully custom accelerator and plugins
     accelerator = MyAccelerator()
-    trainer = Trainer(accelerator=accelerator)
+    precision_plugin = MyPrecisionPlugin()
+    training_type_plugin = CustomDDPPlugin(accelerator=accelerator, precision_plugin=precision_plugin)
+    trainer = Trainer(strategy=training_type_plugin)
 
 
 The full list of built-in plugins is listed below.
--- a/pl_examples/bug_report/bug_report_model.py
+++ b/pl_examples/bug_report/bug_report_model.py
@@ -57,9 +57,6 @@ def run():
         num_sanity_val_steps=0,
         max_epochs=1,
         enable_model_summary=False,
-        accelerator="cpu",
-        strategy="ddp",
-        devices=2,
     )
     trainer.fit(model, train_dataloaders=train_data, val_dataloaders=val_data)
     trainer.test(model, dataloaders=test_data)
--- a/pl_examples/bug_report/bug_report_model.py
+++ b/pl_examples/bug_report/bug_report_model.py
@@ -57,6 +57,9 @@ def run():
         num_sanity_val_steps=0,
         max_epochs=1,
         enable_model_summary=False,
+        accelerator="cpu",
+        strategy="ddp",
+        devices=2,
     )
     trainer.fit(model, train_dataloaders=train_data, val_dataloaders=val_data)
     trainer.test(model, dataloaders=test_data)
--- a/pytorch_lightning/__init__.py
+++ b/pytorch_lightning/__init__.py
@@ -13,7 +13,7 @@ if not _root_logger.hasHandlers():
     _logger.addHandler(logging.StreamHandler())
     _logger.propagate = False
 
-from pytorch_lightning.accelerators import Accelerator
+from pytorch_lightning.accelerators import Accelerator  # noqa: E402
 from pytorch_lightning.callbacks import Callback  # noqa: E402
 from pytorch_lightning.core import LightningDataModule, LightningModule  # noqa: E402
 from pytorch_lightning.trainer import Trainer  # noqa: E402
--- a/pytorch_lightning/__init__.py
+++ b/pytorch_lightning/__init__.py
@@ -13,13 +13,12 @@ if not _root_logger.hasHandlers():
     _logger.addHandler(logging.StreamHandler())
     _logger.propagate = False
 
-from pytorch_lightning.accelerators import Accelerator  # noqa: E402
 from pytorch_lightning.callbacks import Callback  # noqa: E402
 from pytorch_lightning.core import LightningDataModule, LightningModule  # noqa: E402
 from pytorch_lightning.trainer import Trainer  # noqa: E402
 from pytorch_lightning.utilities.seed import seed_everything  # noqa: E402
 
-__all__ = ["Accelerator", "Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
+__all__ = ["Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
 
 # for compatibility with namespace packages
 __import__("pkg_resources").declare_namespace(__name__)
--- a/pytorch_lightning/__init__.py
+++ b/pytorch_lightning/__init__.py
@@ -13,12 +13,13 @@ if not _root_logger.hasHandlers():
     _logger.addHandler(logging.StreamHandler())
     _logger.propagate = False
 
+from pytorch_lightning.accelerators import Accelerator
 from pytorch_lightning.callbacks import Callback  # noqa: E402
 from pytorch_lightning.core import LightningDataModule, LightningModule  # noqa: E402
 from pytorch_lightning.trainer import Trainer  # noqa: E402
 from pytorch_lightning.utilities.seed import seed_everything  # noqa: E402
 
-__all__ = ["Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
+__all__ = ["Accelerator", "Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
 
 # for compatibility with namespace packages
 __import__("pkg_resources").declare_namespace(__name__)
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -30,7 +30,7 @@ class Accelerator:
     - IPU
     """
 
-    def setup_environment(self) -> None:
+    def setup_environment(self, root_device: torch.device) -> None:
         """Setup any processes or distributed connections.
 
         This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -31,27 +31,14 @@ class Accelerator:
     - GPU
     - TPU
     - IPU
-
-    Each Accelerator gets two plugins upon initialization:
-    One to handle differences from the training routine and one to handle different precisions.
     """
 
-    def __init__(self, precision_plugin: Optional[PrecisionPlugin], training_type_plugin: TrainingTypePlugin) -> None:
-        """
-        Args:
-            precision_plugin: the plugin to handle precision-specific parts
+    def __init__(self) -> None:
+        super().__init__()
 
-                .. deprecated::
-                    The ``precision_plugin`` parameter has been deprecated and will be removed soon.
-                    Pass the precision plugin as a parameter to the ``TrainingTypePlugin`` instead.
-
-            training_type_plugin: the plugin to handle different training routines
-        """
-
-        self.training_type_plugin = training_type_plugin
-
-        if precision_plugin is not None:
-            self.training_type_plugin._precision_plugin = precision_plugin
+    @property
+    def root_device(self):
+        return None
 
     def setup_environment(self) -> None:
         """Setup any processes or distributed connections.
@@ -59,7 +46,6 @@ class Accelerator:
         This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
         environment before setup is complete.
         """
-        self.training_type_plugin.setup_environment()
 
     def setup(self, trainer: "pl.Trainer") -> None:
         """Setup plugins for the trainer fit and creates optimizers.
@@ -67,40 +53,6 @@ class Accelerator:
         Args:
             trainer: the trainer instance
         """
-        self.training_type_plugin.setup(trainer)
-
-    @property
-    def model(self) -> Module:
-        """Returns the model.
-
-        This can also be a wrapped LightningModule. For retrieving the pure LightningModule use
-        :attr:`Accelerator.lightning_module`
-        """
-        return self.training_type_plugin.model
-
-    @model.setter
-    def model(self, new_model: Module) -> None:
-        self.training_type_plugin.model = new_model
-
-    @property
-    def lightning_module(self) -> "pl.LightningModule":
-        """Returns the pure LightningModule.
-
-        To get the potentially wrapped model use :attr:`Accelerator.model`
-        """
-        return self.training_type_plugin.lightning_module
-
-    @property
-    def root_device(self) -> torch.device:
-        """Returns the root device."""
-        return self.training_type_plugin.root_device
-
-    def teardown(self) -> None:
-        """This method is called to teardown the training process.
-
-        It is the right place to release memory and free other resources.
-        """
-        self.training_type_plugin.teardown()
 
     def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
         """Gets stats for a given device.
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -12,14 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from abc import abstractmethod
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Union
 
 import torch
-from torch.nn import Module
 
 import pytorch_lightning as pl
-from pytorch_lightning.plugins.precision import PrecisionPlugin
-from pytorch_lightning.plugins.training_type import TrainingTypePlugin
 
 
 class Accelerator:
@@ -32,13 +29,6 @@ class Accelerator:
     - TPU
     - IPU
     """
-
-    def __init__(self) -> None:
-        super().__init__()
-
-    @property
-    def root_device(self):
-        return None
 
     def setup_environment(self) -> None:
         """Setup any processes or distributed connections.
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -38,9 +38,7 @@ class GPUAccelerator(Accelerator):
         """
         super().setup_environment()
         if "cuda" not in str(self.root_device):
-            raise MisconfigurationException(
-                f"Device should be GPU, got {self.root_device} instead"
-            )
+            raise MisconfigurationException(f"Device should be GPU, got {self.root_device} instead")
         torch.cuda.set_device(self.root_device)
 
     def setup(self, trainer: "pl.Trainer") -> None:
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -21,7 +21,8 @@ import torch
 
 import pytorch_lightning as pl
 from pytorch_lightning.accelerators.accelerator import Accelerator
-from pytorch_lightning.utilities.exceptions import MisconfigurationException
+
+# from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.imports import _TORCH_GREATER_EQUAL_1_8
 
 _log = logging.getLogger(__name__)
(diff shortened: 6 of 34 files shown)


pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -405,11 +407,30 @@ def training_type_plugin(self) -> TrainingTypePlugin:
...
-
+        if (
+            # handle custom strategy with custom accelerator
+            (
+                isinstance(self.strategy, TrainingTypePlugin)
+                and (self.strategy.accelerator is None or not isinstance(self.strategy.accelerator, Accelerator))
+            )
+            or not isinstance(self.strategy, TrainingTypePlugin)
+        ):
+            self._training_type_plugin.accelerator = self.select_accelerator()
```
four4fish, line 433: @awaelchli this is really ugly, but this is the minimal impact way I can think about to handle custom plugins. How about we refactor/rewrite [Accelerator_connector](https://github.com/PyTorchLightning/pytorch-lightning/issues/10422) after this and keep logic as same as possible compare to before for now?

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -405,11 +407,17 @@ def training_type_plugin(self) -> TrainingTypePlugin:
...
+        ):
+            precision_plugin = self.precision_plugin
+            if precision_plugin is not None:
+                self._training_type_plugin._precision_plugin = precision_plugin
         self._training_type_plugin_resolved = True
-
+        if (hasattr(self.strategy, "accelerator") and self.strategy.accelerator is None) or not hasattr(
+            self.strategy, "accelerator"
+        ):
+            self._training_type_plugin.accelerator = self.select_accelerator()
```
four4fish, line 420: @awaelchli this is really ugly, but this is the minimal impact way I can think about to handle custom plugins. How about we refactor/rewrite Accelerator_connector after this and keep logic as same as possible compare to before for now?

tests/accelerators/test_accelerator_connector.py
```diff
@@ -1029,10 +1034,13 @@ def test_unsupported_tpu_choice(monkeypatch):
     with pytest.raises(MisconfigurationException, match=r"accelerator='tpu', precision=64\)` is not implemented"):
         Trainer(accelerator="tpu", precision=64)
 
-    with pytest.warns(UserWarning, match=r"accelerator='tpu', precision=16\)` but native AMP is not supported"):
-        Trainer(accelerator="tpu", precision=16)
-    with pytest.warns(UserWarning, match=r"accelerator='tpu', precision=16\)` but apex AMP is not supported"):
-        Trainer(accelerator="tpu", precision=16, amp_backend="apex")
+    with pytest.raises(ValueError, match="TPUAccelerator` can only be used with a `SingleTPUPlugin`"):
```
four4fish, line 1037: reviewers: the TPU misconfig check has been moved from TPUAccelerator setup to accelerator_connector, so it will get checked when trainer initialize

pytorch_lightning/accelerators/tpu.py
```diff
@@ -29,25 +25,6 @@
...
-                f"The `TPUAccelerator` can only be used with a `TPUPrecisionPlugin`,"
-                f" found: {self.training_type_plugin.precision_plugin}."
-            )
-        if not isinstance(self.training_type_plugin, (SingleTPUPlugin, TPUSpawnPlugin)):
-            raise ValueError(
-                "The `TPUAccelerator` can only be used with a `SingleTPUPlugin` or `TPUSpawnPlugin,"
-                f" found {self.training_type_plugin}."
-            )
-        return super().setup(trainer)
-
```
four4fish, line 50: reviewers: this has been moved to the end of the accelerator_connector, will be called at the end of the accelerator_connector init

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -394,22 +395,23 @@ def precision_plugin(self) -> PrecisionPlugin:
...
+        if (
+            isinstance(self.strategy, TrainingTypePlugin) and self.strategy._precision_plugin is None
+        ) or not isinstance(self.strategy, TrainingTypePlugin):
+            precision_plugin = self.precision_plugin
+            if precision_plugin is not None:
+                self._training_type_plugin.precision_plugin = precision_plugin
+        if (isinstance(self.strategy, TrainingTypePlugin) and self.strategy.accelerator is None) or not isinstance(
+            self.strategy, TrainingTypePlugin
+        ):
+            self._training_type_plugin.accelerator = self.select_accelerator()
```
four4fish, line 414: @awaelchli this is really ugly, but this is the minimal impact way I can think about to handle custom plugins. How about we refactor/rewrite Accelerator_connector after this and keep logic as same as possible compare to before for now?

pytorch_lightning/accelerators/tpu.py
ananthsub: why not move the gpu/cpu acelerator checks to the accelerator connector as well?

ananthsub (commented):
nit: updating references for `trainer.accelerator.training_type_plugin` to `trainer.training_type_plugin` ((e.g. in files test_deepspeed_plugin and test_ddp_plugin_wih_comm_hook) could be a separate PR to make this one smaller and easier to review

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -12,14 +12,11 @@
...
 
 import torch
-from torch.nn import Module
 
 import pytorch_lightning as pl
-from pytorch_lightning.plugins.precision import PrecisionPlugin
-from pytorch_lightning.plugins.training_type import TrainingTypePlugin
 
 
 class Accelerator:
```
ananthsub, line 25: I understand this PR is to facilitate the refactor. What should the ideal Accelerator API look like? It looks like the accelerator is now not much more than the torch.device being used, vendor specific setup, and a way to get device stats (which is not a bad thing)

pl_examples/bug_report/bug_report_model.py
```diff
@@ -57,6 +57,9 @@ def run():
         num_sanity_val_steps=0,
         max_epochs=1,
         enable_model_summary=False,
+        accelerator="cpu",
+        strategy="ddp",
+        devices=2,
```
carmocca, line 62: This shouldn't be changed

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -797,9 +799,9 @@ def select_accelerator(self) -> Accelerator:
         else:
             acc_cls = CPUAccelerator
 
-        accelerator = acc_cls(precision_plugin=None, training_type_plugin=self.training_type_plugin)
+        accelerator = acc_cls()
         # transfer ownership of the plugins to the accelerator
-        self._training_type_plugin = proxy(self.training_type_plugin)
+        # self._training_type_plugin = proxy(self.training_type_plugin)
```
carmocca, line 804: Remove this?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1671,11 +1672,11 @@ def _on_exception(self) -> None:
 
     @property
     def accelerator(self) -> Accelerator:
-        return self._accelerator_connector.accelerator
+        return self.training_type_plugin.accelerator
 
     @property
     def training_type_plugin(self) -> TrainingTypePlugin:
-        return self.accelerator.training_type_plugin
+        return self._accelerator_connector.training_type_plugin
```
carmocca, line 1679:
```suggestion
        return self.training_type_plugin
```

pytorch_lightning/accelerators/accelerator.py
four4fish: @ananthsub @justusschock, I agree that current Accelerator is more like a Device class and it's a good thing to have :) And more device related check/logic could be moved into this Accelerator subclasses after this 3/n. Especially as @justusschock mentioned logic like has_tpu(). Or mis configs/error specific to one Accelerator like cpu `num_processes` 
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L262-L264




>Make this the only source of truth for the used device type (e.g either replace the enum, or make it a property of this class) as I find it kind of confusing, that we check different things here and they could potentially diverge (although they shouldn't :)

Totally agree, we are planning on deprecating the trainer._device_type from and unify the check for device type to check accelerator class [Remove trainer._device_type in favor of check Accelerator class](https://github.com/PyTorchLightning/pytorch-lightning/issues/11002) and [Generalize internal checks for precision plugin type, training type, accelerator type](https://github.com/PyTorchLightning/pytorch-lightning/issues/10821) 
+1 make device_type a property of Accelerator class and deprecate Accelerator enum. (I think there will be circle import issue to solve before we can deprecate accelerator_enum)

pytorch_lightning/accelerators/tpu.py
four4fish: It's a bit different. This TPU check is checking what strategies and precision plugins to use when device is TPU, which should happens before in plugin selection phase like accelerator_connector.

For device specific checks like what's in cpu/gpu now, I think it will make sense to slim the accelerator_connector and moving to accelerator subclass.

pytorch_lightning/accelerators/gpu.py
```diff
@@ -30,22 +30,19 @@
...
-            raise MisconfigurationException(
-                f"Device should be GPU, got {self.training_type_plugin.root_device} instead"
-            )
-        torch.cuda.set_device(self.training_type_plugin.root_device)
+        if "cuda" not in str(root_device):
+            raise MisconfigurationException(f"Device should be GPU, got {self.root_device} instead")
+        torch.cuda.set_device(root_device)
 
     def setup(self, trainer: "pl.Trainer") -> None:
+        # TODO refactor input from trainer to local_rank @four4fish
```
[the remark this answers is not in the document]
four4fish: I don't think so. root_device not necessarily equal to root_device. Some strategies has root_device to be parallel_device[local_rank] like ddp, but some plugins has different implementation like dp and ddp2 (root_device always be device[0]), and single device doesn't have local rank property.

pytorch_lightning/trainer/trainer.py
```diff
@@ -451,6 +451,7 @@ def __init__(
             amp_level,
             plugins,
         )
+        self._accelerator_connector.training_type_plugin
```
awaelchli, line 454: note to self: need to find a way to avoid this. at the very least, this should be done within the connector

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -1016,3 +1014,18 @@ def _is_slurm_managing_tasks(self) -> bool:
         total_requested_devices = (self.num_gpus or self.num_processes) * self.num_nodes
         num_slurm_tasks = int(os.environ["SLURM_NTASKS"], 0)
         return num_slurm_tasks == total_requested_devices
+
+    def _check_tpu_mis_config(self) -> None:
+        # TODO moved from TPUAccelerator when refactor accelerator. Revisit when refactor
+        # accelerator_connector @four4fish
```
awaelchli, line 1025: @four4fish I would avoid such TODOs (from experience with other refactors) and rather note them to the related issue.

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: Thank you. Yes whatever makes the tests pass. At this point I have no clue anymore.

tests/accelerators/test_cpu.py
```diff
@@ -16,9 +16,10 @@
 
 def test_restore_checkpoint_after_pre_dispatch_default():
     """Assert default for restore_checkpoint_after_pre_dispatch is False."""
-    plugin = SingleDevicePlugin(torch.device("cpu"))
-    accelerator = CPUAccelerator(training_type_plugin=plugin, precision_plugin=PrecisionPlugin())
-    assert not accelerator.training_type_plugin.restore_checkpoint_after_pre_dispatch
+    plugin = SingleDevicePlugin(
+        accelerator=CPUAccelerator(), device=torch.device("cpu"), precision_plugin=PrecisionPlugin()
+    )
+    assert not plugin.restore_checkpoint_after_pre_dispatch
```
carmocca, line 22: Dupe

```suggestion
```

tests/accelerators/test_accelerator_connector.py
```diff
@@ -408,12 +415,11 @@ class Accel(Accelerator):
     class Prec(PrecisionPlugin):
         pass
 
-    class TrainTypePlugin(SingleDevicePlugin):
```
carmocca, line 411: Why this change?

tests/accelerators/test_accelerator_connector.py
```diff
@@ -397,7 +397,14 @@ def creates_processes_externally(self) -> bool:
...
 @mock.patch.dict(
     os.environ,
-    {"SLURM_NTASKS": "2", "SLURM_JOB_NAME": "SOME_NAME", "SLURM_NODEID": "0", "LOCAL_RANK": "0", "SLURM_LOCALID": "0"},
+    {
+        "SLURM_NTASKS": "2",
+        "SLURM_JOB_NAME": "SOME_NAME",
+        "SLURM_NODEID": "0",
+        "LOCAL_RANK": "0",
+        "SLURM_PROCID": "0",
+        "SLURM_LOCALID": "0",
```
carmocca, line 406: Removing this comma will undo this unrelated change

```suggestion
        "SLURM_LOCALID": "0"
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -1016,3 +1014,23 @@ def _is_slurm_managing_tasks(self) -> bool:
...
+        if isinstance(self.accelerator, TPUAccelerator):
+            if not isinstance(self.training_type_plugin.precision_plugin, TPUPrecisionPlugin):
+                raise ValueError(
+                    f"The `TPUAccelerator` can only be used with a `TPUPrecisionPlugin`,"
+                    f" found: {self.training_type_plugin.precision_plugin}."
+                )
+            if not isinstance(self.training_type_plugin, (SingleTPUPlugin, TPUSpawnPlugin)):
+                raise ValueError(
+                    "The `TPUAccelerator` can only be used with a `SingleTPUPlugin` or `TPUSpawnPlugin,"
```
carmocca, line 1034:
```suggestion
                    "The `TPUAccelerator` can only be used with a `SingleTPUPlugin` or `TPUSpawnPlugin`,"
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -997,14 +996,13 @@ def _set_distrib_type_if_training_type_plugin_passed(self):
...
+        conditions match:
 
-            - A SLURM cluster is detected
-            - A distributed plugin is being used
-            - The process is not launching in interactive mode
-            - The number of tasks in SLURM matches the requested number of devices and nodes in the Trainer
+        - A SLURM cluster is detected
+        - A distributed plugin is being used
+        - The process is not launching in interactive mode
+        - The number of tasks in SLURM matches the requested number of devices and nodes in the Trainer
```
carmocca, line 1005: Keep the space before the bullet points?

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -57,13 +61,25 @@ def __init__(
...
         return self._checkpoint_io
 
     @property
     def precision_plugin(self) -> PrecisionPlugin:
-        return self._precision_plugin
+        return self._precision_plugin if self._precision_plugin is not None else PrecisionPlugin()
+
+    @precision_plugin.setter
+    def precision_plugin(self, precision_plugin: Optional[PrecisionPlugin]) -> None:
+        self._precision_plugin = precision_plugin
```
carmocca, line 82: Adding this in #11071

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: this was modified by the weirdo pre-commit bot

docs/source/extensions/plugins.rst
```diff
@@ -80,10 +80,7 @@ can then be passed into the Trainer directly or via a (custom) accelerator:
     trainer = Trainer(strategy=CustomDDPPlugin(), plugins=[CustomPrecisionPlugin()])
 
     # fully custom accelerator and plugins
```
ananthsub, line 82: Custom plugins are dropped from the example?

pytorch_lightning/__init__.py
```diff
@@ -13,12 +13,13 @@
...
     _logger.propagate = False
 
+from pytorch_lightning.accelerators import Accelerator  # noqa: E402
 from pytorch_lightning.callbacks import Callback  # noqa: E402
 from pytorch_lightning.core import LightningDataModule, LightningModule  # noqa: E402
 from pytorch_lightning.trainer import Trainer  # noqa: E402
 from pytorch_lightning.utilities.seed import seed_everything  # noqa: E402
 
-__all__ = ["Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
+__all__ = ["Accelerator", "Trainer", "LightningDataModule", "LightningModule", "Callback", "seed_everything"]
```
ananthsub, line 22: Why this change?

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -31,76 +28,21 @@ class Accelerator:
...
-        To get the potentially wrapped model use :attr:`Accelerator.model`
-        """
-        return self.training_type_plugin.lightning_module
-
-    @property
-    def root_device(self) -> torch.device:
-        """Returns the root device."""
-        return self.training_type_plugin.root_device
-
-    def teardown(self) -> None:
```
ananthsub, line 98: Since the interface supports a setup, we should have a teardown

pytorch_lightning/__init__.py
awaelchli: to annotate the types in the plugins with "pl.Accelerator". Otherwise there appear to be circular imports unfortunately. Haven't found a different solution yet.

justusschock: so annotating with `"pl.accelerators.accelerator.Accelerator"` doesn't work? This might be a bit longer, but since we don't import it directly, that should be fine I think

pytorch_lightning/trainer/connectors/accelerator_connector.py
carmocca: The weirdo https://github.com/myint/docformatter project. pre-commit just obeys what we tell it to do.

carmocca: Note that the bot did this in reaction to a change by a human. Probably moving "Returns ``True..." to the first paragraph was the cause.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1124,7 +1125,7 @@ def _run(
         {Trainer.fit} or {Trainer.test} or {Trainer.predict}  ||
                                 |                             ||
                          spawn processes                      ||
-               {self.accelerator.setup_environment}           ||
+               {self.training_type_plugin.setup_environment}           ||
```
carmocca, line 1128: This needs adjustment

pytorch_lightning/accelerators/accelerator.py
carmocca: May I suggest we only add hooks when they have actual implementations? Adding just because is how we've ended up with that many deprecated hooks.

pytorch_lightning/__init__.py
carmocca: `__all__` is just to indicate what's imported with `from pytorch_lightning import *`

It shouldn't relate to circular imports at all

pytorch_lightning/accelerators/accelerator.py
awaelchli: After this refactor, we will have the opportunity to split out device specific teardown that currently lives in the ddp/spawn plugin teardown and move it to the accelerator teardown. So either way, it will be sorted out :)

awaelchli: https://github.com/PyTorchLightning/pytorch-lightning/blob/ffb1a754af5301608c71955c12468216e920af51/pytorch_lightning/plugins/training_type/ddp.py#L504-L508

pytorch_lightning/plugins/training_type/training_type_plugin.py
four4fish: Thank you @carmocca, will rebase if 11071 merge first. The accelerator_connector logic change in this PR need precision setters for now

pytorch_lightning/trainer/connectors/accelerator_connector.py
four4fish: Yeah make sense, will remove

pytorch_lightning/trainer/trainer.py
four4fish: Yeah agree, at least we should wet it in accelerator_connector

pytorch_lightning/__init__.py
awaelchli: okay whatever works

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -42,12 +42,16 @@ class TrainingTypePlugin(ABC):
     loop."""
 
     def __init__(
-        self, checkpoint_io: Optional[CheckpointIO] = None, precision_plugin: Optional[PrecisionPlugin] = None
+        self,
+        accelerator: Optional["pl.accelerators.accelerator.Accelerators.accelerator.Accelerator"] = None,
```
carmocca, line 46: I legitimally laughed at this 😂😂😂

Did find+replace trick you?

awaelchli: how is this possible ...

awaelchli:
```suggestion
        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
```

docs/source/extensions/accelerators.rst
```diff
@@ -25,11 +25,10 @@ One to handle differences from the training routine and one to handle different
     from pytorch_lightning.accelerators import GPUAccelerator
     from pytorch_lightning.plugins import NativeMixedPrecisionPlugin, DDPPlugin
 
-    accelerator = GPUAccelerator(
-        precision_plugin=NativeMixedPrecisionPlugin(precision=16, device="cuda"),
-        training_type_plugin=DDPPlugin(),
-    )
-    trainer = Trainer(accelerator=accelerator)
+    accelerator = GPUAccelerator()
+    precision_plugin = NativeMixedPrecisionPlugin(precision=16, device="cuda")
```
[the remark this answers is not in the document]
four4fish: Yeah make sense. Will create a issue to make the device param optional in native_mixed precision plugins and clean up here

docs/source/extensions/plugins.rst
four4fish: Will the example above sufficient?

tests/accelerators/test_accelerator_connector.py
four4fish: Was just for my own testing purpose. Good spotting :) will change it back

docs/source/extensions/plugins.rst
four4fish: added

four4fish:
Reviewers: most issue has been addressed, thank you for feedbacks and discussions!
Follow up for this PR:
1. Teardown hook, issue filed : https://github.com/PyTorchLightning/pytorch-lightning/issues/11091
2. Simplify NativeMixedPrecisionPlugin init param and remove device as required param. It's probably less trivial then it looks as it will require accelerator_connector changes.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -425,6 +443,7 @@ def teardown(self) -> None:
 
         It is the right place to release memory and free other resources.
         """
+        self._move_optimizer_state(torch.device("cpu"))
```
awaelchli, line 446: Note, if I read correctly this was previously only done for the GPU accelerator. Should be fine, just pointing it out in case.

pytorch_lightning/trainer/trainer.py
awaelchli: actually, since it is no longer a property I think this line is now redundant. It does not do anything and let's try to remove it.

awaelchli (approved):
Very nice!

pytorch_lightning/trainer/trainer.py
four4fish: Oh! good catch, it was added for debugging. Now accelerator_connector will initialized the ttp

pytorch_lightning/trainer/trainer.py
```diff
@@ -451,7 +451,7 @@ def __init__(
             amp_level,
             plugins,
         )
-        self.logger_connector = LoggerConnector(self, log_gpu_memory)
+        self.logger_connector: LoggerConnector = LoggerConnector(self, log_gpu_memory)
```
carmocca, line 454: nit: not required

```suggestion
        self.logger_connector = LoggerConnector(self, log_gpu_memory)
```

carmocca (approved):

pytorch_lightning/trainer/trainer.py
four4fish: @awaelchli, did you add this to fix mypy?

pytorch_lightning/plugins/training_type/training_type_plugin.py
four4fish: I don't remember you or me did this. But probably we should keep this in GPUAccelerator teardown. How about we address it together in https://github.com/PyTorchLightning/pytorch-lightning/issues/11091?

four4fish: I don't remember you or me did this. But probably we should keep this in GPUAccelerator teardown. How about we address it together in #11091?

pytorch_lightning/trainer/trainer.py
awaelchli: unfortunately, typing breaks if this is removed.

ananthsub (approved):

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: should be good either way. for tpu it's just redundant because the teardown runs in the spawned processes which, when they end, get completely destroyed so nothing is left in memory. 
But in the future, we may decide to bring the optimizer states back to the main process through the queue and restore them like we  do for the model weights. Anyway, this is open for discussion

pytorch_lightning/trainer/trainer.py
carmocca: I just pulled the PR, removed the annotation, and ran mypy (version 0.910). Everything is green

```suggestion
        self.logger_connector = LoggerConnector(self, log_gpu_memory)
```

four4fish: Awesome! I saw you have a mypy PR out, should I rebase on top of that after it merges and remove this?

carmocca: You can remove this already. This PR just needs to rebase on top of #11095

PR #11031, opened 2021-12-10
Simplify several profile calls

daniellepintz:
## What does this PR do?
Remove several calls to `profiler.profile` in favor doing the profiling inside `trainer._call_ttp_hook`

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/core/optimizer.py
```diff
@@ -149,6 +149,7 @@ def closure_dis():
                 with opt_dis.toggle_model(sync_grad=accumulated_grad_batches):
                     opt_dis.step(closure=closure_dis)
         """
+        # is separating optimizer_step_without_closure and optimizer_step_with_closure still needed?
```
daniellepintz, line 152: question to reviewers

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -107,8 +107,7 @@ def advance(  # type: ignore[override]
             raise StopIteration
 
         if not data_fetcher.store_on_device:
-            with self.trainer.profiler.profile("evaluation_batch_to_device"):
-                batch = self.trainer.training_type_plugin.batch_to_device(batch, dataloader_idx=(dataloader_idx or 0))
```
ananthsub, line 111: this is very interesting. now everything will show up as `batch_to_device`. this is positive for consistency with users since there's no "evaluation_batch_to_device" hook users implement. on the other hand, what happens if users implement these hooks differently based on the running stage the trainer is in?

we don't profile the `setup` or `teardown` hooks any differently, so i wonder if `batch_to_device` has just been an anomaly this whole time

four4fish: @ananthsub how can they implement these hooks differently based on the running stage the trainer? ttk passed in when create trainer which will be shared across stages right?

ananthsub: @four4fish the training type plugin calls the lightning module hook. the lightning module hook can be overridden by users like this:

```
def transfer_batch_to_device(self, batch, device, dataloader_idx):
    if self.trainer.training:
        .....
   elif self.trainer.testing:
        ....
```

pytorch_lightning/core/optimizer.py
carmocca: This is only relevant for manual optimization. It can be helpful information if the manual closure is particularly expensive.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
carmocca: In addition to checking the running stage, the performance could be different if each stage's data is very different. But I don't think it warrants a specific profile name

carmocca (approved):

ananthsub (approved):

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
awaelchli: this can (should) be mentioned in the changelog (and the ones below too)

awaelchli (approved):

pytorch_lightning/core/optimizer.py
awaelchli: remember to remove the line before resolving

PR #11033, opened 2021-12-10
WIP: Add calls to quant/dequant submodules for scripted QAT-trained modules

t-vi:
This does a bit of JIT graph surgery to insert the calls to quant/dequant as neeeded.
To keep backward compat, we leave things alone if the input is already quantized.

Fixes: #7552

### DIFF
--- a/pytorch_lightning/callbacks/quantization.py
+++ b/pytorch_lightning/callbacks/quantization.py
@@ -18,6 +18,7 @@ Quantization
 """
 import copy
 import functools
+import inspect
 from typing import Any, Callable, Dict, Optional, Sequence, Union
 
 import torch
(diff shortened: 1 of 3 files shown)


t-vi:
So a problem I'm still debugging is that something is cached somewhere and PyTorch will "remember" if we first script the model with no-input-compat and then script with input-compat... This is made hard to debug because with the `__torch_function__` mechanism, there 9999 exceptions being thrown and caught for everything involving scriptmodules, so I'll need to use replaying. :shrug:

t-vi:
So some more debugging, we end up with the same function and that will have optimized_graphs already, and so it doesn't look at the improved graph. I'll see what we can do to not have the same function...

pytorch_lightning/callbacks/quantization.py
```diff
@@ -81,6 +82,90 @@ def wrapper(data) -> Any:
     return wrapper
 
 
+def _is_quantize_forward_context_wrapped(model):
```
tchaton, line 85: Mind adding typing and docstring ?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -81,6 +82,90 @@ def wrapper(data) -> Any:
...
+    while True:
+        if first_line <= fw_method.__code__.co_firstlineno <= last_line and fw_method.__code__.co_filename == fn:
+            return True
+        fw_method = getattr(fw_method, "__wrapped__", None)
+        if fw_method is None:
+            return False
+
+
+def _fixup_quantization_torchscript(script_module):
+    # We do a bit of surgery here. We have quant and dequant submodules, but they are not
```
tchaton, line 101: Let's convert this note into docstring.

pytorch_lightning/callbacks/quantization.py
```diff
@@ -81,6 +82,90 @@ def wrapper(data) -> Any:
...
+    fw_method = model.forward
+    while True:
+        if first_line <= fw_method.__code__.co_firstlineno <= last_line and fw_method.__code__.co_filename == fn:
+            return True
+        fw_method = getattr(fw_method, "__wrapped__", None)
+        if fw_method is None:
+            return False
+
+
+def _fixup_quantization_torchscript(script_module):
```
tchaton, line 100: noob question, but do you believe this utility could have its place in PyTorch ?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -81,6 +82,90 @@ def wrapper(data) -> Any:
...
+    # So we add these calls.
+    # But for backward compat, we only do so if the input is not quantized yet.
+    # note that this changes the script_module in-place.
+
+    # We get the quant and dequant modules. In the JIT, scripted modules (and function) are their own types.
+    type_quant = list(script_module.quant.graph.inputs())[0].type()
+    type_dequant = list(script_module.dequant.graph.inputs())[0].type()
+
+    fw_graph = script_module.forward.graph  # .copy()
+    input_to_quantize = list(fw_graph.inputs())[1]
```
tchaton, line 112: noob question. Why take [1]? What happens if there are multiple inputs?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -81,6 +82,90 @@ def wrapper(data) -> Any:
...
+    is_not_quant_node.output().setType(torch.jit.annotations.BoolType.get())
+    is_not_quant_node.insertAfter(is_quant_node)
+
+    # if the input was not quantized, we quantize it. So we add an If block here.
+    if_node = fw_graph.create("prim::If", [is_not_quant_node.output()])
+    if_node.insertAfter(is_not_quant_node)
+    bl_if = if_node.addBlock()
+
+    # get the quant module
+    mod_node = fw_graph.create("prim::GetAttr", list(fw_graph.inputs())[:1])
```
tchaton, line 128: What is the difference between `list(fw_graph.inputs())[:1]` and `input_to_quantize` ?

pytorch_lightning/callbacks/quantization.py
```diff
@@ -81,6 +82,90 @@ def wrapper(data) -> Any:
...
+    call_node = fw_graph.create("prim::CallMethod", [mod_node.output(), input_to_quantize])
+    call_node.s_("name", "forward")
+    call_node.insertAfter(mod_node)
+    bl_if.registerOutput(call_node.output())
+    # in the else block (input was quantized), we do nothing
+    bl_else = if_node.addBlock()
+    bl_else.registerOutput(input_to_quantize)
+
+    # finally, we replace the use of the input with the quantized-if-needed output of the if_node.
+    input_to_quantize.replaceAllUsesAfterNodeWith(if_node, if_node.output())
```
tchaton, line 142: that's really cool.

tchaton (approved):

t-vi:
Thank you @tchaton for the comments. This is currently waiting for a way to clear PyTorch's cache. I've been digging quite a bit but I have yet to find a solution.

tchaton:
Hey @t-vi,

No problem. Even if I find this PR really insight full and quite magical/hacky, I believe it would be better suited within PyTorch.
Keep us updated on this PR status and PyTorch side progress.

Best,
T.C

PR #11069, opened 2021-12-15
Update the TQDM progress bar `on_train_epoch_end`

### LINKED ISSUE #10653: Progressbar showed different value vs tensorboard and myself output
## 🐛 Bug

code:
image

result:
image

We can see that the value on the progress bar is different from the output that I print. And the value on tensorboard also is the same as my output, so I guess there is something happening in the progress bar.

I only use one GPU.

### Environment

- PyTorch Lightning Version: 1.4.6
- PyTorch Version (e.g., 1.8): 1.8.1

cc @tchaton @rohitgr7 @akihironitta @carmocca @edward-io @ananthsub @kamil-kaczmarek @Raalsky @Blaizzy @SeanNaren @kaushikb11

carmocca:
## What does this PR do?

Fixes #10653

None

## Before submitting

## PR review

cc @borda @awaelchli

### DIFF
--- a/pytorch_lightning/callbacks/progress/tqdm_progress.py
+++ b/pytorch_lightning/callbacks/progress/tqdm_progress.py
@@ -250,7 +250,8 @@ class TQDMProgressBar(ProgressBarBase):
         if trainer.sanity_checking:
             reset(self.val_progress_bar, total=sum(trainer.num_sanity_val_batches), current=self.val_batch_idx)
         else:
-            self._update_bar(self.main_progress_bar)  # fill up remaining
+            if trainer.state.fn == pl.trainer.states.TrainerFn.FITTING:
+                self._update_bar(self.main_progress_bar)  # fill up remaining
             self.val_progress_bar = self.init_validation_tqdm()
             reset(self.val_progress_bar, total=self.total_val_batches, current=self.val_batch_idx)
 
@@ -258,14 +259,15 @@ class TQDMProgressBar(ProgressBarBase):
         super().on_validation_batch_end(trainer, pl_module, outputs, batch, batch_idx, dataloader_idx)
         if self._should_update(self.val_batch_idx):
             self._update_bar(self.val_progress_bar)
-            self._update_bar(self.main_progress_bar)
+            if trainer.state.fn == pl.trainer.states.TrainerFn.FITTING:
+                self._update_bar(self.main_progress_bar)
 
     def on_validation_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         if self.is_enabled:
             self._update_bar(self.val_progress_bar)
 
     def on_validation_end(self, trainer, pl_module):
-        if self.main_progress_bar is not None:
+        if self.main_progress_bar is not None and trainer.state.fn == pl.trainer.states.TrainerFn.FITTING:
             self.main_progress_bar.set_postfix(self.get_metrics(trainer, pl_module))
         self.val_progress_bar.close()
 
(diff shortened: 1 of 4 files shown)


pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -310,8 +319,8 @@ def print(
             s = sep.join(map(str, args))
             active_progress_bar.write(s, end=end, file=file, nolock=nolock)
 
-    def _should_update(self, current, total) -> bool:
-        return self.is_enabled and (current % self.refresh_rate == 0 or current == total)
```
awaelchli, line 314: I'm not yet fully understanding this fix. How are you handling the case where we need to update the bar at the end, when total is reached? the refresh rate may not align with the last batch.

awaelchli: Is that why you implement `on_test_epoch_end` now, to cover that case? In that case I agree.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -207,12 +208,10 @@ def init_test_tqdm(self) -> Tqdm:
         return bar
 
     def on_sanity_check_start(self, trainer, pl_module):
-        super().on_sanity_check_start(trainer, pl_module)
```
justusschock, line 210: why is it necessary to remove the `super` calls?

justusschock (commented):
Same as @awaelchli I'm struggling to understand why this fixes it. Could you maybe add a comment on that?

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -595,3 +596,64 @@ def test_tqdm_progress_bar_main_bar_resume():
...
+
+    model = MyModel()
+    pbar = MockedProgressBar()
+    trainer = Trainer(
+        default_root_dir=tmpdir,
+        limit_train_batches=2,
+        limit_val_batches=2,
+        limit_test_batches=2,
+        max_epochs=2,
+        enable_model_summary=False,
```
awaelchli, line 636: could also disable checkpointing

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: The parent doesn't override these. It's not necessary, I can put them back if you want.

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: So before, each `on_*_batch_end` checked whether it was the last batch and updated in that case (by passing `total` to `_should_update`)

Now, we do it in `on_*_epoch_end` always so this is not necessary

tests/callbacks/test_tqdm_progress_bar.py
carmocca:
```suggestion
        enable_model_summary=False,
        enable_checkpointing=False,
```

awaelchli:
@carmocca Perhaps the PR should leave the linked issue open until we have the same fix for rich?

pytorch_lightning/callbacks/progress/tqdm_progress.py
awaelchli: I see, I have tests for that so this should be fine!

awaelchli (approved):

carmocca:
> @carmocca Perhaps the PR should leave the linked issue open until we have the same fix for rich?

Sure, I don't how to unlink it, but we can re-open once this is merged

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: Yes, the number of calls are the same, but the logger connector state won't be the correct one in the last `on_train_batch_end` vs `on_train_epoch_end`

kaushikb11 (approved):

justusschock (approved):

PR #11068, opened 2021-12-15
Deprecate `Trainer.should_rank_save_checkpoint` property

jjenniferdai:
## What does this PR do?

see context: https://github.com/PyTorchLightning/pytorch-lightning/pull/11035#discussion_r767079698
original PR: #9433
original issue: #9074

Go through deprecation for public property

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #7684, opened 2021-05-24
Add `should_rank_save_checkpoint` property to Training Plugins

kaushikb11:
## What does this PR do?

Fixes #7678 
Related to [link
](https://github.com/PyTorchLightning/pytorch-lightning/pull/7296#discussion_r633377680)
## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [ ] Did you make sure to update the documentation with your changes? (if necessary)
- [ ] Did you write any new necessary tests? (not for typos and docs)
- [ ] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #7296, opened 2021-04-30
Support TPU Pod Training (n/n)

kaushikb11:
## What does this PR do?

Fixes #3058 

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [ ] Did you make sure to update the documentation with your changes? (if necessary)
- [ ] Did you write any new necessary tests? (not for typos and docs)
- [ ] Did you verify new and existing tests pass locally with your changes?
- [ ] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #11070, opened 2021-12-15
Remove `should_rank_save_checkpoint` property from TTP

carmocca:
## What does this PR do?

Finishes #9074

TTP API (experimental): removes the property

## Before submitting

## PR review

cc @borda @kaushikb11 @justusschock @awaelchli @akihironitta @ananthsub


pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -312,7 +312,16 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
...
-        return self.checkpoint_io.save_checkpoint(checkpoint, filepath)
+        self.checkpoint_io.save_checkpoint(checkpoint, filepath)
+
+    def remove_checkpoint(self, filepath: _PATH) -> None:
+        """Remove checkpoint filepath from the filesystem.
+
+        Args:
+            filepath: Path to checkpoint
+        """
+        if self.local_rank == 0:
```
carmocca, line 323: This change is equivalent to what's in master. However, I don't understand why we don't check local rank for `save_checkpoint` but we do it for remove.

pytorch_lightning/plugins/training_type/single_tpu.py
```diff
@@ -82,7 +82,7 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
             checkpoint: dict containing model and trainer state
             filepath: write-target file's path
         """
-        return self.checkpoint_io.save_checkpoint(checkpoint, filepath)
+        self.checkpoint_io.save_checkpoint(checkpoint, filepath)
```
carmocca, line 85: Can't we remove this override? Isn't global rank always 0 for single tpu?

pytorch_lightning/plugins/training_type/tpu_spawn.py
awaelchli: because xm.save needs to be called on all ranks. It internally checks if the local rank is 0. This is just how the library works. If we don't call xm.save on all ranks, it will hang. 

https://pytorch.org/xla/release/1.5/_modules/torch_xla/core/xla_model.html

pytorch_lightning/plugins/training_type/single_tpu.py
awaelchli: it was overridden because of my answer below.

awaelchli: I think you can drop the override

awaelchli (approved):

pytorch_lightning/plugins/training_type/tpu_spawn.py
carmocca: Okay, I'll add a comment

pytorch_lightning/plugins/training_type/single_tpu.py
carmocca: Done

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -312,7 +312,17 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
             checkpoint: dict containing model and trainer state
             filepath: write-target file's path
         """
-        return self.checkpoint_io.save_checkpoint(checkpoint, filepath)
```
four4fish, line 315: save_checkpoint() is same as training_type_plugin, should we remove it?

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -312,7 +312,17 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
...
+        # xm.save needs to be called on all ranks. It internally checks if the local rank is 0
+        self.checkpoint_io.save_checkpoint(checkpoint, filepath)
+
+    def remove_checkpoint(self, filepath: _PATH) -> None:
```
four4fish, line 318: RFC: should we have checkpoint_io for tpu and deepspeed instead of override functions in strategies directly?

pytorch_lightning/plugins/training_type/tpu_spawn.py
jjenniferdai: rank check is different (default global_zero v.s. all ranks here)

jjenniferdai:
@carmocca @ananthsub see https://github.com/PyTorchLightning/pytorch-lightning/pull/11068#discussion_r769903035 re: Trainer property removal v.s. deprecation?

pytorch_lightning/plugins/training_type/tpu_spawn.py
four4fish: @jjenniferdai the logic difference is in XLACheckpointIO.save_checkpoint() right? the strategy overrides here is same as training_type_plugin so it's not needed?

jjenniferdai: yep IO differences in `checkpointIO` but there's also different rank logic controlled by TTPs hence the TTP override:

training_type_plugin: `if global_zero: checkpointIO.save()`
here: `checkpointIO.save()` (all ranks)

pytorch_lightning/plugins/training_type/tpu_spawn.py
jjenniferdai: TPU: override is just rank logic which should be controlled by strategies right

deepspeed: my thoughts are it would be hard/inadvisable to fit inside `IOPlugin`, since it looks like it involves things like `broadcast` call, checking trainer state, which we ideally shouldn't do in `IOPlugin`. I think we'd need to pass `deepspeed_engine` too
https://github.com/PyTorchLightning/pytorch-lightning/blob/3c4d06bd42ff3282a017cd9fa879bbd68e8568df/pytorch_lightning/plugins/training_type/deepspeed.py#L755-L768

carmocca:
@jjenniferdai We don't need to keep BC for the property in the TTP. Since the property in the `Trainer` is read-only, we can copy its current implementation into it.

Whichever PR lands later should be updated.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1702,9 +1703,10 @@ def world_size(self) -> int:
     @property
     def should_rank_save_checkpoint(self) -> bool:
         rank_zero_deprecation(
-            "`Trainer.should_rank_save_checkpoint` is deprecated in v1.6 and will be removed in 1.8.", stacklevel=5
+            "`Trainer.should_rank_save_checkpoint` is deprecated in v1.6 and will be removed in v1.8.", stacklevel=5
         )
-        return self.training_type_plugin.should_rank_save_checkpoint
+        ttp = self.training_type_plugin
+        return isinstance(ttp, TPUSpawnPlugin) and ttp.local_rank == 0 or ttp.is_global_zero
```
jjenniferdai, line 1709: re: custom plugins like @awaelchli 's comment [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/11068#pullrequestreview-833560843):
could check for `hasattr(should_rank_save_checkpoint)` but could be misleading when TTP's `should_rank_save_checkpoint` is removed and doesn't actually control anything but the trainer property is not yet removed

carmocca: I'd say this is okay since it is read-only. IMO the deprecation here is more about not explicitly breaking the property access.

I also wouldn't worry about custom plugins since we are absolutely not recommending them for now. And when we do (1.6+), users won't implement `should_rank_save_checkpoint` as this PR will have removed it.

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -331,10 +341,6 @@ def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_gra
     def teardown(self) -> None:
         os.environ.pop("PT_XLA_DEBUG", None)
 
-    @property
-    def should_rank_save_checkpoint(self) -> bool:
-        return self.local_rank == 0
```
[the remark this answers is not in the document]
carmocca: I think this was answered here: https://github.com/PyTorchLightning/pytorch-lightning/pull/7296#discussion_r631789547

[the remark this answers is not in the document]
awaelchli: Yes exactly right. We wouldn't have this property in the first place if it wasn't for this difference. #7684

pytorch_lightning/trainer/trainer.py
awaelchli: I still think if we are not going all the way, we should remove the Trainer property without deprecation, or raise a custom error message. After all, it was just meant to be a short hand for `trainer.training_type_plugin.xyz`

carmocca: I'm totally fine either way. I proposed marking the property protected from the start
https://github.com/PyTorchLightning/pytorch-lightning/pull/7684#discussion_r639336673

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -312,7 +312,17 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
...
+        # `xla_model.save` needs to be called on all ranks. It internally checks if the local rank is 0
+        self.checkpoint_io.save_checkpoint(checkpoint, filepath)
+
+    def remove_checkpoint(self, filepath: _PATH) -> None:
+        """Remove checkpoint filepath from the filesystem.
+
+        Args:
+            filepath: Path to checkpoint
+        """
+        if self.local_rank == 0:
```
tchaton, line 324: This could be an issue in the future if the nodes filesystem becomes one a single volume.

Ideally, I think we should detect if the filesystem is shared. The condition would become (if self.global_rank if shared_filesystem else self.local_rank)

pytorch_lightning/plugins/training_type/tpu_spawn.py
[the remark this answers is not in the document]
tchaton: Instead we could have:

```py
save_rank = self.is_global_zero if self.checkpoint_io.shared_filesystem else self.local_rank
if save_rank:
    ...
```

pytorch_lightning/plugins/training_type/tpu_spawn.py
carmocca: I believe you have the same question as this comment: https://github.com/PyTorchLightning/pytorch-lightning/pull/11070/files#r770116930

tchaton (approved):
LGTM !

pytorch_lightning/plugins/training_type/tpu_spawn.py
tchaton: Yes, mind-opening an issue associated with this?

pytorch_lightning/trainer/trainer.py
carmocca: Keeping the deprecation since nobody seems to have a strong opinion and the work has been done already

Borda (approved):

PR #8743, opened 2021-08-05
Introduce CheckpointIO Plugin

SeanNaren:
## What does this PR do?

Closes #8118.

This is a V0 where we've abstracted the checkpointing logic into a class that eventually can be plugged in via the trainer.

Few outstanding issues that I'll tag throughout the PR!

The goal is to be able to do this:

```python
trainer = Trainer(DDPPlugin(checkpoint_plugin=MyCustomCheckpointIO()))
```

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #11071, opened 2021-12-15
Fix setter usage for checkpoint io and precision in TTP

### LINKED ISSUE #11097: Lazy initialize properties in Strategies
## Proposed refactor
Strategy has four properties in init : Accelerator, Precision_plugin, Checkpoint_io and Cluster_enviroment.

User could pass in a strategy class or a strategy str into trainer.
- If user pass a string, accelerator_connector will initialize the strategy with the properties selection. 
- If user pass a strategy class, the logics get a bit mess. We should track what user has passed in V.S strategy default value V.S. Trainer.accelerator_connector 

This issue propose lazy initialization for strategy properties. @awaelchli and @ananthsub also bought this up before

### Motivation
For correctness, maintenance and enable future simplifications 

### Pitch

Current Training_type_plugin __init__ will set default value to precision_plugin and checkpoint_io:
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/plugins/training_type/training_type_plugin.py#L40-L58
```python
    def __init__(
        self, checkpoint_io: Optional[CheckpointIO] = None, precision_plugin: Optional[PrecisionPlugin] = None
    ) -> None:
        self._model: Optional[Module] = None
        self._precision_plugin = precision_plugin if precision_plugin is not None else PrecisionPlugin()

```

Proposal

```python
def __init__(
        self, checkpoint_io: Optional[CheckpointIO] = None, precision_plugin: Optional[PrecisionPlugin] = None
    ) -> None:
        self._model: Optional[Module] = None

@property
def checkpoint_io(self):
       return self._checkpoint_io if self._checkpoint_io is not None else TorchCheckpointIO()

@setter
def checkpoint_io(checkpoint_io):

@property
def precision_plugin(self):
      return self._precision_plugin if self._precision_plugin is not None else PrecisionPlugin()

@setter
def precision_plugin(precision_plugin):
```

In accelerator_connector.py, we know:
`training_type_plugin._precision_plugin` before we call setter is what user passed in, not the default value.

### Additional context

carmocca:
## What does this PR do?

Fixes inconsistencies with the `precision_plugin` and `checkpoint_io` properties in the `TrainingTypePlugin`.

Fixes #11097 

None

## Before submitting

## PR review

cc @justusschock @awaelchli @akihironitta @borda @jjenniferdai @kaushikb11 @ananthsub


pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -57,17 +57,21 @@ def __init__(
...
-    def checkpoint_io(self) -> CheckpointIO:
-        return self._checkpoint_io
-
     @property
     def precision_plugin(self) -> PrecisionPlugin:
         return self._precision_plugin
 
+    @precision_plugin.setter
+    def precision_plugin(self, plugin: PrecisionPlugin) -> None:
+        self._precision_plugin = plugin
```
carmocca, line 66: This has been added for consistency. Ideally neither would have a setter but the `AcceleratorConnector` requires it.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -46,8 +46,8 @@ def __init__(
     ) -> None:
         self._model: Optional[Module] = None
         checkpoint_io = checkpoint_io if checkpoint_io is not None else TorchCheckpointIO()
-        self._checkpoint_io = checkpoint_io
-        self._precision_plugin = precision_plugin if precision_plugin is not None else PrecisionPlugin()
```
carmocca, line 50: The setters weren't getting used.

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -871,14 +870,6 @@ def register_plugins(cls, plugin_registry: Dict) -> None:
             offload_optimizer_device="nvme",
         )
 
-    @property
-    def checkpoint_io(self) -> CheckpointIO:
-        return self._checkpoint_io
-
-    @checkpoint_io.setter
-    def checkpoint_io(self, plugin: CheckpointIO) -> None:
-        raise MisconfigurationException("DeepSpeed currently does not support custom checkpoint plugins.")
```
carmocca, line 880: #8743 first implemented a `DeepSpeedCheckpointPlugin` which was removed before merge, so this is not necessary

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: And how are they getting used now? There can't be a public attribute and a setter with the same name.

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: How is this not creating an error? Something is not right here

pytorch_lightning/plugins/training_type/deepspeed.py
rohitgr7: if someones does this:
```py
plugin = DeepSpeedPlugin()
plugin.checkpoint_io = CustomCheckpointIO()
```
this should still raise this exception?

pytorch_lightning/plugins/training_type/tpu_spawn.py
```diff
@@ -339,10 +338,8 @@ def should_rank_save_checkpoint(self) -> bool:
...
-        return self._checkpoint_io
-
-    @checkpoint_io.setter
-    def checkpoint_io(self, plugin: CheckpointIO) -> None:
-        raise MisconfigurationException("TPU Spawn Plugin currently does not support custom checkpoint plugins.")
+    @DDPSpawnPlugin.checkpoint_io.setter
+    def checkpoint_io(self, io: XLACheckpointIO) -> None:
+        if not isinstance(io, XLACheckpointIO):
+            raise MisconfigurationException(f"{self.__class__.__name__}.checkpoint_io` must be a `XLACheckpointIO`.")
+        self._checkpoint_io = io
```
rohitgr7, line 345: quick ques: are we forcing users to always subclass `XLACheckpointIO` and not `CheckpointIO` even if their implementation is correct?

pytorch_lightning/plugins/training_type/training_type_plugin.py
carmocca: This will use the setter to set `self._precision_plugin`.

If we don't do it like this, any custom logic in the setter won't run

pytorch_lightning/plugins/training_type/training_type_plugin.py
carmocca: Why would it?

pytorch_lightning/plugins/training_type/tpu_spawn.py
carmocca: It's un-pythonic to have these `isinstance` checks. I prefer to remove them, however, in this PR I'm trying to keep the same functionality but with better consistency.

Before, the setter raised unconditionally, I think what it meant is to enforce a `XLACheckpointIO`

pytorch_lightning/plugins/training_type/deepspeed.py
carmocca: Why? This class does not have any custom checkpointing logic

carmocca: Mmmm I'm conflicted here.

This plugin doesn't use a checkpoint_io at all. Which might suggest it's good to raise

But the base implementation will set (through the setter) a default `checkpoint_io` which means the exception would always be raised.

I think it's more natural not to override the setter and leave it be. The problem is that this TTP implementation violates the base interface as the base requires the `checkpoint_io` but this subclass does not.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -46,8 +46,8 @@ def __init__(
...
+        self.checkpoint_io = checkpoint_io
+        self.precision_plugin = precision_plugin if precision_plugin is not None else PrecisionPlugin()
```
four4fish, line 50: We need this to be self._precision_plugin instead of direct set to self.precision_plugin
[In 3/n move accelerator into strategy](https://github.com/PyTorchLightning/pytorch-lightning/pull/11022) The accelerator_connector depends on distinguishing is user set up a non_None precision plugin or we set default to decide whether call _select_precision_plugin(), so self._precision_plugin is what user passed in, and self.precision_plugin is default value

four4fish:
```suggestion
        self._checkpoint_io = checkpoint_io
        self._precision_plugin = precision_plugin
```
```suggestion
        self.checkpoint_io = checkpoint_io
        self.precision_plugin = precision_plugin
```

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -57,17 +57,21 @@ def __init__(
                 f" Move your implementation to `{self.__class__.__name__}.teardown()` instead."
             )
 
-    @property
-    def checkpoint_io(self) -> CheckpointIO:
-        return self._checkpoint_io
-
     @property
     def precision_plugin(self) -> PrecisionPlugin:
         return self._precision_plugin
```
four4fish, line 62:
```suggestion
    def precision_plugin(self) -> PrecisionPlugin:
        return self._precision_plugin if self._precision_plugin is not None else PrecisionPlugin()
```

pytorch_lightning/plugins/training_type/training_type_plugin.py
four4fish: Add a issue to discuss future 
https://github.com/PyTorchLightning/pytorch-lightning/issues/11097

carmocca: Do the current changes fix #11097?

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -49,9 +49,8 @@ def __init__(
     ) -> None:
         self._accelerator = accelerator
         self._model: Optional[Module] = None
-        checkpoint_io = checkpoint_io if checkpoint_io is not None else TorchCheckpointIO()
-        self._checkpoint_io = checkpoint_io
-        self._precision_plugin = precision_plugin
+        self.checkpoint_io = checkpoint_io
+        self.precision_plugin = precision_plugin
```
four4fish, line 53:
```suggestion
        self._checkpoint_io = checkpoint_io
        self._precision_plugin = precision_plugin
```

carmocca: Can you explain why? Don't we want the subclasses that override the `setter` with validation (e.g. tpu) to run the validation on init?

See the test I just added in my last commit. It would fail with your proposed change

four4fish: Yeah sure. 
To distinguish user passed Trainer(strategy = CustomStrategy()) Versus. Trainer(stategy = CustomStrategy(precision_plugin = PrecisionPlugin())) in accelerator_connector
If we call strategy.precision_plugin, it will execute the @property which will assign a default value. Then we don't know did user pass in None or a precision_plugin. The current accelerator logic is depends on strategy._precision_plugin to decide whether user passed None and we should choose precision plugin for them. 
This is what I'm trying to say in #11097, we want to distinguish what user passed in vs default value

Could you run tests/accelerator/test_accelerator_test.py? I think with this change, the custom plugin test will fail

carmocca: CI looks green

pytorch_lightning/plugins/training_type/tpu_spawn.py
rohitgr7: yeah, I'd suggest by default it should use XLACheckpointIO but should allow users to be able to set a custom one too. can be updated in a separate PR.

pytorch_lightning/plugins/training_type/deepspeed.py
rohitgr7: if we remove this and someone calls plugin.checkpoint_io, it will return `TorchCheckpointIO` I think.
To avoid the conflict with the base one you can check
```py
def checkpoint_io(self, plugin: CheckpointIO)
    if plugin is not None:
        raise MisconfigurationException(...)
```
since it's optional anyways in the base class.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -71,7 +70,11 @@ def accelerator(self, accelerator: "pl.accelerators.accelerator.Accelerator") ->
 
     @property
     def checkpoint_io(self) -> CheckpointIO:
-        return self._checkpoint_io
+        return self._checkpoint_io if self._checkpoint_io is not None else TorchCheckpointIO()
```
rohitgr7, line 73: quick ques: why return `TorchCheckpointIO` here if there is no checkpoint_io?

pytorch_lightning/plugins/training_type/training_type_plugin.py
four4fish: I see why it worked. in accelerator_connector.py line 406
```
if (
            isinstance(self.strategy, TrainingTypePlugin) and self.strategy._precision_plugin is None
        ) or not isinstance(self.strategy, TrainingTypePlugin):
      call choose precision_plugin logic
```
Because training_type_plugin.precision_plugin haven't been called yet, the training_type_plugin._precision_plugin does not exist, so it guaranteed to be None, so it worked.

But this is VERY fragile, when the training_type_plugin.precision_plugin get called, both `training_type_plugin._precision_plugin` and `training_type_plugin.precision_plugin` will be the default value instead of what user set. If this happens before final_training_type_plugin(), even user themselves called it before pass into trainer, the precision plugin selection logic won't trigger and we will have silent bugs.  

As the test passes for now, feel free to land. We can discuss the longterm robust solution when we solve #11097

pytorch_lightning/plugins/training_type/training_type_plugin.py
carmocca: I'm following the same pattern used below for `precision_plugin`

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -873,14 +872,6 @@ def register_plugins(cls, plugin_registry: Dict) -> None:
             offload_optimizer_device="nvme",
         )
 
-    @property
-    def checkpoint_io(self) -> CheckpointIO:
```
carmocca, line 877: Just noticed that DeepSpeed overrides `save_checkpoint` and `load_checkpoint`, but not `remove_checkpoint`. Is that correct? Will the removal work with `TorchCheckpointIO`?

pytorch_lightning/plugins/training_type/training_type_plugin.py
carmocca: I fully agree. This is not a long-term solution. The accelerator connector needs a revamp

pytorch_lightning/plugins/training_type/deepspeed.py
carmocca: Done, although I still think we shouln't have these checks. They limit the flexibiltiy.

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -49,9 +49,8 @@ def __init__(
     ) -> None:
         self._accelerator = accelerator
```
four4fish, line 50: Why accelerator here is different?

carmocca: There wasn't a property for it here when I opened the PR. Let's fix it :)

```suggestion
        self.accelerator = accelerator
```

four4fish (approved):

pytorch_lightning/plugins/training_type/deepspeed.py
awaelchli: Yes, I think so because it has `recursive=True`

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -47,11 +47,10 @@ def __init__(
         checkpoint_io: Optional[CheckpointIO] = None,
         precision_plugin: Optional[PrecisionPlugin] = None,
     ) -> None:
-        self._accelerator = accelerator
+        self.accelerator = accelerator
```
awaelchli, line 50: why are these renamed? You can't have a property and attribute with the same name. What is the actual fix?

four4fish: I was confused about this PR and thought it changes logic too, but seems not. Only logic change is for checkpoint_io.
Before accelerator and precision_plugin in __init__ set self._accelerator/_precision_plugin directly, with changes in this PR it calling setter functions (`self.accelerator` is actually set `self._accelerator` in the accelerator setter method).

pytorch_lightning/plugins/training_type/training_type_plugin.py
four4fish: cc @awaelchli for this discussion, which is related to why ._accelerator become .accelerator

pytorch_lightning/plugins/training_type/training_type_plugin.py
carmocca: Exactly, also see the changed test

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -873,13 +873,10 @@ def register_plugins(cls, plugin_registry: Dict) -> None:
...
-        return self._checkpoint_io
-
-    @checkpoint_io.setter
-    def checkpoint_io(self, plugin: CheckpointIO) -> None:
-        raise MisconfigurationException("DeepSpeed currently does not support custom checkpoint plugins.")
+    @DDPPlugin.checkpoint_io.setter
+    def checkpoint_io(self, plugin: Optional[CheckpointIO]) -> None:
+        if plugin is not None:
+            raise MisconfigurationException("DeepSpeed does not support custom checkpoint plugins.")
```
rohitgr7, line 879: nit: just for consistency with your other changes.
```suggestion
    def checkpoint_io(self, io: Optional[CheckpointIO]) -> None:
        if io is not None:
            raise MisconfigurationException("DeepSpeed does not support custom `CheckpointIO` plugin.")
```

rohitgr7 (approved):

pytorch_lightning/plugins/training_type/deepspeed.py
tchaton: Might be better to delete the directory directly. This might be cleaner in `TorchCheckpointIO`.

```py
        fs = get_filesystem(path)
        if fs.exists(path):
            if fs.isdir(path):
                fs.rmdir(path)
            else:
                fs.rm(path, recursive=True)
            log.debug(f"Removed checkpoint: {path}")
```

tchaton (approved):
LGTM when all comments are addressed

pytorch_lightning/plugins/training_type/deepspeed.py
carmocca: Rmdir is only for empty directories.

awaelchli (approved):

Issue #11012, opened 2021-12-09
[RFC] Tuner Revamp


rohitgr7:
### Issues
1. The Tuner has been causing a lot of issues in the past and we plan to refactor it. The primary reason comes from the trainer state [snapshotting and restoration](https://github.com/PyTorchLightning/pytorch-lightning/blob/3fcfd0214cc95fa7de49f6404421b5c6cd113c57/pytorch_lightning/tuner/batch_size_scaling.py#L102-L140).
2. Auto batch size scaling doesn't work with validate/test/predict. Users might want to identify an optimal batch_size for inference to better utilize their available compute resources.
3. [LR Finder suggestion](https://github.com/PyTorchLightning/pytorch-lightning/blob/3fcfd0214cc95fa7de49f6404421b5c6cd113c57/pytorch_lightning/tuner/lr_finder.py#L164-L182) is not optimal. Sometimes it suggests bad LR as per its algorithm, and sometimes [it doesn't suggest anything at all](https://github.com/PyTorchLightning/pytorch-lightning/blob/3fcfd0214cc95fa7de49f6404421b5c6cd113c57/pytorch_lightning/tuner/lr_finder.py#L180-L182).
4. Doesn't work with [flash finetuning](https://lightning-flash.readthedocs.io/en/latest/general/finetuning.html#finetuning-in-flash). For eg. let's say the user might want to compute new LR or new batch_size after certain epochs of pre-training, then it's not easily configurable within a single call. One can achieve it with multiple calls but since we support strategies within Flash, this might be worth adding.

ps: please add more issues up here if you have any regarding the tuner.

### Possible solutions
- We can subclass Trainer for tuner and create independent states so that we don't do any sort of snapshotting and restoration with trainer states and it will stay independent.
```py
class Tuner(Trainer):
     # create independent states
     # create custom loops

trainer.tuner(auto_scale_batch_size=..., auto_lr_find=...).fit()
trainer.tuner(auto_scale_batch_size=...).predict()
```
well, this solution could possibly solve `1` & `2` but possibly can't be configured to solve `4`.

- Another solution proposed by @Borda is to make them as callbacks, so that they can be easily configured by users independently and can help resolve `4`. But this solution might not resolve `1` & `2`.

- Another solution @Borda and @SkafteNicki suggested, for now, is to move lr_finder to bolts and experiment there and improve scale_batch_size within lightning. But possibly it can't guarantee to solve `4`.

### Additional context

[...]

Issue #10918, opened 2021-12-03
[RFC] Revisit Usage of MisconfigurationException


justusschock:
## Proposed refactor
Replace Misconfiguration Exception with more diverse error types.

### Motivation

Currently we are using the `MisconfigurationException` for everything even if they are not directly related to the user's configuration. This means that the type of exception has no additional information about the kind of error. Python has some builtin exception types for that like `RuntimeError`, `ValueError`, `TypeError` etc. To gain more insights from the actual exception type, I propose to not use MisconfigurationException any longer and replace it with the other error types.

### Pitch
Have Lightning raising appropriate error types instead of just `MisconfigurationException` everywhere.

### Additional context

If we want to have some Lightning-Unique Exceptions we could still do something like 
```python
class LightningValueError(ValueError):
    pass

```
for the necessary error types and then raise these. This has the advantage that people could still have custom error handling also depending on types and statements like `except ValueError` would still catch those but they could also decide to only catch Lightning Errors.
______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

PS: Not sure under which part of the code to put it, since it probably touches all of them :D

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7 @ananthsub @tchaton @carmocca 

EDIT: current usage of the `MisconfigurationException`: https://grep.app/search?q=MisconfigurationException&filter[repo][0]=PyTorchLightning/pytorch-lightning

carmocca:
Fully agree!

[...]

PR #11089, opened 2021-12-15
Add BatchSizeFinder callback

rohitgr7:
## What does this PR do?

Part of #11012 

TODO:
- Add docs in followups
- Address added TODOs here in the followup

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub @daniellepintz @rohitgr7 @akihironitta


pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,319 @@
...
+            )
+
+        # Arguments we adjust during the batch size finder, save for restoring
+        self._dump_params(trainer)
+
+        # Set to values that are required by the algorithm
+        self._reset_params(trainer)
+
+        # Save initial model, that is loaded after batch size is found
+        save_path = os.path.join(trainer.default_root_dir, f"scale_batch_size_temp_model_{uuid.uuid4()}.ckpt")
```
tchaton, line 93: Might be better to make it hidden with `.scale_batch_size_temp_model_ `

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,319 @@
...
+        elif trainer.state.fn == TrainerFn.PREDICTING:
+            trainer.reset_predict_dataloader(pl_module)
+
+    def _dump_params(self, trainer):
+        self._dumped_params = {
+            "current_epoch": trainer.current_epoch,
+            "global_step": trainer.global_step,
+            "max_steps": trainer.max_steps,
+            "logger": trainer.logger,
+            "callbacks": trainer.callbacks,
```
tchaton, line 223: Should we remove deep copy the callbacks or even replace them with an empty list during tuning?

tchaton: Same with logger ?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,319 @@
...
+            trainer.fit_loop.max_steps = self.steps_per_trial
+        elif trainer.state.fn == TrainerFn.VALIDATING:
+            trainer.limit_val_batches = self.steps_per_trial
+        elif trainer.state.fn == TrainerFn.TESTING:
+            trainer.limit_test_batches = self.steps_per_trial
+        elif trainer.state.fn == TrainerFn.PREDICTING:
+            trainer.limit_predict_batches = self.steps_per_trial
+
+    def _restore_params(self, trainer):
+        trainer.fit_loop.current_epoch = self._dumped_params["current_epoch"]
```
tchaton, line 244: I am not sure if this would clean properly the progress objects.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,319 @@
...
+        elif trainer.state.fn == TrainerFn.VALIDATING:
+            trainer.reset_val_dataloader(pl_module)
+        elif trainer.state.fn == TrainerFn.TESTING:
+            trainer.reset_test_dataloader(pl_module)
+        elif trainer.state.fn == TrainerFn.PREDICTING:
+            trainer.reset_predict_dataloader(pl_module)
+
+    def _dump_params(self, trainer):
+        self._dumped_params = {
+            "current_epoch": trainer.current_epoch,
```
tchaton, line 219: Might be better to deecopy the loops

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: then the user would need to re-initialize the Trainer again?

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: I think saving and restoring loop state dict should work fine. wdyt?

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: using loop state dict now. I guess that should work??

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,476 @@
...
+            trainer.limit_test_batches = self._dumped_params["limit_test_batches"]
+        elif trainer.state.fn == TrainerFn.PREDICTING:
+            loop = trainer.predict_loop
+            trainer.limit_predict_batches = self._dumped_params["limit_predict_batches"]
+
+        loop.load_state_dict(deepcopy(self._dumped_params["loop_state_dict"]), force_load_progress=True)
+        if "loop_verbose" in self._dumped_params:
+            loop.verbose = self._dumped_params["loop_verbose"]
+
+    def pre_early_exit(self, trainer):
```
tchaton, line 335: Let's make those methods private.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,476 @@
...
+            trainer.num_test_batches = self._dumped_params["num_test_batches"]
+            loop = trainer.test_loop
+        if trainer.state.fn == TrainerFn.PREDICTING:
+            trainer.num_predict_batches = self._dumped_params["num_predict_batches"]
+            loop = trainer.predict_loop
+
+        loop.load_state_dict(self._dumped_params["loop_state_dict"], force_load_progress=True)
+        trainer.callbacks = [cb for cb in trainer.callbacks if not isinstance(cb, BatchSizeFinder)]
+
+    def on_fit_start(self, trainer, pl_module):
```
tchaton, line 374: Let's merge the duplicated code there.

And would on_fit_start + on_validation_start trigger 2 tuning ? It is an expected behaviour ?

pytorch_lightning/loops/base.py
```diff
@@ -275,10 +276,9 @@ def state_dict(self, destination: Optional[Dict] = None, prefix: str = "") -> Di
         destination[prefix + "state_dict"] = self.on_save_checkpoint()
 
         # do not get the mode from `self.trainer` because it might not have been attached yet
-        ft_enabled = _FaultTolerantMode.detect_current_mode().is_enabled
         for k, v in self.__dict__.items():
             key = prefix + k
-            if ft_enabled and isinstance(v, BaseProgress):
+            if isinstance(v, BaseProgress):
```
tchaton, line 281: Why did you modify this?

pytorch_lightning/loops/base.py
```diff
@@ -295,21 +295,30 @@ def load_state_dict(
...
 
-    def _load_from_state_dict(self, state_dict: Dict, prefix: str, metrics: Optional[Dict[str, Metric]] = None) -> None:
         for k, v in self.__dict__.items():
             key = prefix + k
             if key not in state_dict:
                 # no state for this object, maybe we are loading an old checkpoint
                 continue
 
-            if isinstance(v, BaseProgress):
+            if load_progress and isinstance(v, BaseProgress):
```
tchaton, line 321: Same here?

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -304,4 +305,7 @@ def _reorder_callbacks(callbacks: List[Callback]) -> List[Callback]:
         """
         checkpoints = [c for c in callbacks if isinstance(c, ModelCheckpoint)]
         not_checkpoints = [c for c in callbacks if not isinstance(c, ModelCheckpoint)]
-        return not_checkpoints + checkpoints
+        callbacks = not_checkpoints + checkpoints
+        batch_size_finder_callback = [c for c in callbacks if isinstance(c, BatchSizeFinder)]
+        other_callbacks = [c for c in callbacks if not isinstance(c, BatchSizeFinder)]
```
tchaton, line 310: oher callback is ambiguous, let's name it tuner callback.

pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -40,6 +40,8 @@ def scale_batch_size(
     batch_arg_name: str = "batch_size",
 ) -> Optional[int]:
     """See :meth:`~pytorch_lightning.tuner.tuning.Tuner.scale_batch_size`"""
+    raise MisconfigurationException("this is gone")
```
tchaton, line 43: ?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,476 @@
...
+    def on_predict_start(self, trainer, pl_module):
+        self.scale_batch_size(trainer, pl_module)
+
+        if self.early_exit:
+            self.pre_early_exit(trainer)
+        else:
+            trainer.callbacks = [cb for cb in trainer.callbacks if not isinstance(cb, BatchSizeFinder)]
+
+    def on_fit_end(self, trainer, pl_module):
+        if self.early_exit:
```
tchaton, line 410:
```suggestion
        if not self.early_exit:
```

tchaton: Mind providing more information about the pre and post early exit. What are you trying to achieve?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,476 @@
...
+            trainer.num_training_batches = 0
+        elif trainer.state.fn == TrainerFn.VALIDATING:
+            self._dumped_params["num_val_batches"] = trainer.num_val_batches
+            trainer.num_val_batches = [0]
+        elif trainer.state.fn == TrainerFn.TESTING:
+            self._dumped_params["num_test_batches"] = trainer.num_test_batches
+            trainer.num_test_batches = [0]
+        elif trainer.state.fn == TrainerFn.PREDICTING:
+            self._dumped_params["num_predict_batches"] = trainer.num_predict_batches
+            trainer.num_predict_batches = [0]
```
tchaton, line 351: I believe you should raise a custom expectation and catch it within the Trainer catching error mechanism.

pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: just a temporary thing.. will be removing this whole code later.

tchaton:
After chatting with @rohitgr7,

Here is the API design specification for the `BatchSizeFinder`.

# With the tune method.
User story: As a user, I want to fit the best batch size for each stage and get the associated maximal batch size.

```py
trainer = Trainer()

trainer.tune(auto_batch_size=True, method="fit", ...) # Compute the train_batch_size and val_batch_size
   # In the LightningModule, DataModule
    self.batch_size = train_batch_size # BC -> depreceated in favor of `{stage}_batch_size`.
    self.train_batch_size = train_max_batch_size
    self.val_batch_size = val_max_batch_size

trainer.tune(auto_batch_size=True, method="validate", ...) -> val dataloader batch_size
    # In the LightningModule, DataModule
    self.val_batch_size = max_batch_size

trainer.tune(auto_batch_size=True, method="test", ...) -> test dataloader batch_size
    # In the LightningModule, DataModule
    self.test_batch_size = max_batch_size

trainer.tune(auto_batch_size=True, method="predict", ...) -> predict dataloader batch_size
    self.predict_batch_size = max_batch_size
```

# With the Trainer entry point method.
User story: As a user, I want to fit the best batch size and run the trainer entry point into a single call.

```py
trainer = Trainer(callbacks=BatchSizeFinder())
trainer.fit(...) -> `on_fit_start` perform tune
trainer.validate(...) -> `on_validation_start` perform tune
trainer.test(...) -> `on_test_start` perform tune
trainer.predict(...) -> `on_predict_start` perform tune
```

Open questions: 

1. Some use case requires dynamic maximum batch size such as fine-tuning where the memory footprint of the model increases during trainer. Should we enable a scheduling strategy or re-compute the `max_batch_size` if the params group of the optimizer changed?

2. ?

pytorch_lightning/loops/base.py
rohitgr7: progress state is saved only when fault_tolerant is enabled, but now with the tuner, we need to reset the base progress as well after tuning if it's not early exiting to make sure while continuing with the `.fit/validate/..` call it should be reset completely.

pytorch_lightning/loops/base.py
rohitgr7: this was done to make sure the progress state is loaded if required else during state_dict it used to save only when fault_tolerant is True, but now it always saves it and while loading back, we have an argument(default=False) that force reloads the base progress state_dict if it's set to True else only reloads it if fault-tolerant is enabled.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: on_validation_start triggers tuning only when it's a `.validate` call.

rohitgr7:
update on PR:

all that's been discussed [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/11089#issuecomment-1005548992) is now added.

Only the following proposal is missing and I need more time to think about how to go about this. Maybe in a sep PR to avoid a lot of changes in a single one.
```
trainer.tune(auto_batch_size="fit | True (BC)", ...) # Compute the train_batch_size and val_batch_size
   # In the LightningModule, DataModule
    self.batch_size = train_batch_size # BC -> depreceated in favor of `{stage}_batch_size`.
    self.train_batch_size = train_max_batch_size
    self.val_batch_size = val_max_batch_size
```

Also the API is:
```py
trainer.tune(method="fit|validate|test|predict") (default fit)
```
instead of 
```py
trainer.tune(auto_batch_size="fit|validate|test|predict")
```
because, there is another option to run batch_size_finder using `trainer.tuner.scale_batch_size()` and `auto_batch_size` argument won't be a good one here. Also `auto_batch_size` is a Trainer argument already.

ananthsub (commented):
IMO a callback is not the right fit here. this will override the behavior of fit/validate/test if used this way. why doesn't this use the loops API instead?

rohitgr7:
> IMO a callback is not the right fit here. this will override the behavior of fit/validate/test if used this way. why doesn't this use the loops API instead?

I tried writing pseudocode to cover up all the possible use-cases we have and convert it into a component that users can easily integrate and plugin, but loops couldn't fit all possible cases. For eg, we have [flash finetuning](https://lightning-flash.readthedocs.io/en/latest/general/finetuning.html#finetuning-in-flash) and let's say the user might want to compute new LR or new batch_size after certain epochs of pre-training, then it's not easily configurable within a single call. One can achieve it with multiple calls but since we support strategies within Flash. With callbacks this is now pretty simple since we make sure after each tuning call, the respective loop state_dict is restored completely, and we still get an updated hparam using the tuner. Plus it now runs for val/test/predict.

If you have any ideas to configure it using loops, would love to discuss them.

tests/tuner/test_scale_batch_size.py
```diff
@@ -72,53 +79,45 @@ def test_scale_batch_size_method_with_model_or_datamodule(tmpdir, model_bs, dm_b
...
         "logger",
-        "max_steps",
+        "callbacks",
         "global_step",
+        "limit_val_batches",
+        "max_steps",
+        "limit_val_batches",
+        "limit_test_batches",
+        "limit_predict_batches",
     ]
```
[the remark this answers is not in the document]
rohitgr7: yeah, I had some doubts about it while implementing this. Looks like there are already some issues regarding this. Let's fix them all in a follow-up and update this test once we have standardized implementation as a callback?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,425 @@
...
+
+        self._dumped_params["loop_state_dict"] = deepcopy(loop.state_dict(force_save_progress=True))
+        if hasattr(loop, "verbose"):
+            self._dumped_params["loop_verbose"] = loop.verbose
+
+    def _reset_params(self, trainer: "pl.Trainer") -> None:
+        from pytorch_lightning.loggers.base import DummyLogger
+        from pytorch_lightning.trainer.states import TrainerFn
+
+        trainer.logger = DummyLogger() if trainer.logger is not None else None
```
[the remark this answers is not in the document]
rohitgr7: if someone calls self.trainer.logger.do_something, we don't want to raise an error, else it will require code change.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,425 @@
...
+            rank_zero_warn(
+                f"Field `model.{self.batch_arg_name}` and `model.hparams.{self.batch_arg_name}` are mutually exclusive!"
+                f" `model.{self.batch_arg_name}` will be used as the initial batch size for scaling."
+                " If this is not the intended behavior, please remove either one."
+            )
+
+        if not trainer._data_connector._train_dataloader_source.is_module():
+            raise MisconfigurationException(
+                "The batch scaling feature cannot be used with dataloaders passed directly to `.fit()`."
+                " Please disable the feature or incorporate the dataloader into the model."
```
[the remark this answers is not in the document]
rohitgr7: do we know the issue for the failing test? I'll add a TODO for now.

pytorch_lightning/loops/base.py
```diff
@@ -335,4 +339,5 @@ def _load_from_state_dict(self, state_dict: Dict, prefix: str, metrics: Optional
                     v.reset(metrics=False)
 
         self.on_load_checkpoint(state_dict[prefix + "state_dict"])
-        self.restarting = True
+        if _fault_tolerant_training():
+            self.restarting = True
```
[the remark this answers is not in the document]
rohitgr7: since we are using loop.load_state_dict it's restarting the loop by marking this flag as True. Updating this... IMO load_state_dict shouldn't decide the flag here. It should be set explicitly during the restart.

pytorch_lightning/callbacks/batch_size_finder.py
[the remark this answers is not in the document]
akihironitta: Not sure really since the PR doesn't point to any issue related to the test failure... Created an issue for this: #11604

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+        self.init_val = init_val
+        self.max_trials = max_trials
+        self.batch_arg_name = batch_arg_name
+        self.optimal_batch_size = init_val
+
+        self._early_exit = False
+
+        from pytorch_lightning.loggers.base import LightningLoggerBase
+
+        class _BatchSizeFinderDumpedParams(TypedDict):
```
tchaton, line 93: Mind taking this out of the init method and use a dataclass. "LightningLoggerBase" can be imported under if TYPE_CHECKING:

pytorch_lightning/loops/base.py
```diff
@@ -314,7 +317,6 @@ def load_state_dict(
         for k, v in self.__dict__.items():
             if isinstance(v, Loop):
                 v.load_state_dict(state_dict.copy(), prefix + k + ".")
-        self.restarting = True
```
tchaton, line 317: Why it is being removed? Instead, I would manually reset this value on the loops from the Tuner side. This would be cleaner IMO. @carmocca

pytorch_lightning/callbacks/batch_size_finder.py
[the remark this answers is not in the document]
tchaton: Mind adding a comment ?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1037,9 +1054,11 @@ def tune(
         model: "pl.LightningModule",
         train_dataloaders: Optional[Union[TRAIN_DATALOADERS, LightningDataModule]] = None,
         val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+        dataloaders: Optional[EVAL_DATALOADERS] = None,
         datamodule: Optional[LightningDataModule] = None,
         scale_batch_size_kwargs: Optional[Dict[str, Any]] = None,
         lr_find_kwargs: Optional[Dict[str, Any]] = None,
+        method="fit",
```
tchaton, line 1061:
```suggestion
        method: str = "fit",
```

pytorch_lightning/tuner/tuning.py
```diff
@@ -112,22 +149,33 @@ def scale_batch_size(
...
+        _check_tuner_configuration(self.trainer, train_dataloaders, val_dataloaders, dataloaders, method)
+
+        batch_size_finder = BatchSizeFinder(
+            mode=mode,
+            steps_per_trial=steps_per_trial,
+            init_val=init_val,
+            max_trials=max_trials,
+            batch_arg_name=batch_arg_name,
         )
+        batch_size_finder._early_exit = True
```
tchaton, line 164: Add a comment why it is being set from outside.

pytorch_lightning/tuner/tuning.py
```diff
@@ -45,14 +52,37 @@ def _tune(
         if self.trainer.auto_scale_batch_size:
             if isinstance(self.trainer.auto_scale_batch_size, str):
                 scale_batch_size_kwargs.setdefault("mode", self.trainer.auto_scale_batch_size)
-            result["scale_batch_size"] = scale_batch_size(self.trainer, model, **scale_batch_size_kwargs)
```
tchaton, line 48: Worth to share the value would be overridden. IMO, the provided one should override the trainer one.

pytorch_lightning/tuner/tuning.py
```diff
@@ -112,22 +149,33 @@ def scale_batch_size(
...
+
+        batch_size_finder = BatchSizeFinder(
+            mode=mode,
+            steps_per_trial=steps_per_trial,
+            init_val=init_val,
+            max_trials=max_trials,
+            batch_arg_name=batch_arg_name,
         )
+        batch_size_finder._early_exit = True
+        self.trainer.callbacks = [batch_size_finder] + self.trainer.callbacks
```
tchaton, line 165: Verify the `BatchSizeFinder` doesn't exist already within the callbacks.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+                f"Field `model.{self.batch_arg_name}` and `model.hparams.{self.batch_arg_name}` are mutually exclusive!"
+                f" `model.{self.batch_arg_name}` will be used as the initial batch size for scaling."
+                " If this is not the intended behavior, please remove either one."
+            )
+
+        # TODO: check if this can be enabled (#4040)
+        if not trainer._data_connector._train_dataloader_source.is_module():
+            raise MisconfigurationException(
+                "The batch scaling feature cannot be used with dataloaders passed directly to `.fit()`."
+                " Please disable the feature or incorporate the dataloader into the model."
```
tchaton, line 128:
```suggestion
                " Please disable the feature or incorporate the dataloader into the lightning_module or a datamodule."
```

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+            new_size = self._run_power_scaling(trainer, pl_module, new_size)
+        elif self.mode == "binsearch":
+            new_size = self._run_binary_scaling(trainer, pl_module, new_size)
+
+        garbage_collection_cuda()
+
+        if trainer.is_global_zero:
+            trainer.checkpoint_connector.restore(save_path)
+            fs = get_filesystem(save_path)
+            if fs.exists(save_path):
```
tchaton, line 156: Is this required as you have created the file before ?

tchaton: Should this be within the teardown in case the tuner fails in the process and the checkpoint requires to be cleaned out ?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+        # global step and current epoch are incremented before saved in checkpoint
+        trainer.fit_loop.global_step -= 1
+        trainer.fit_loop.current_epoch -= 1
+
+        self._restore_params(trainer)
+
+        if trainer.progress_bar_callback:
+            trainer.progress_bar_callback.enable()
+
+        print(f"New batch size: {new_size}")
```
tchaton, line 168: Let's use the logger. Would this print on all ranks? Might be worth adding the global rank too. How would it work if the batch_size isn't the same on all ranks? Do we have a check for this?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+        self._reset_params(trainer)
+
+        # Save initial model, that is loaded after batch size is found
+        save_path = os.path.join(trainer.default_root_dir, f".scale_batch_size_temp_model_{uuid.uuid4()}.ckpt")
+        trainer.save_checkpoint(save_path)
+
+        if trainer.progress_bar_callback:
+            trainer.progress_bar_callback.disable()
+
+        new_size, _ = self._adjust_batch_size(trainer, value=self.init_val)
```
tchaton, line 144:
```suggestion
        optimal_batch_size, _ = self._adjust_batch_size(trainer, value=self.init_val)
```

For consistency.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+        if desc:
+            rank_zero_info(f"Batch size {batch_size} {desc}, trying batch size {new_size}")
+
+        # TODO improve this for eval CombinedLoader and multi dataloaders
+        if trainer.state.fn == TrainerFn.FITTING:
+            if not self._is_valid_batch_size(new_size, trainer.train_dataloader, trainer):
+                new_size = min(new_size, len(trainer.train_dataloader.dataset))
+        if trainer.state.fn == TrainerFn.VALIDATING:
+            if not self._is_valid_batch_size(new_size, trainer.val_dataloaders[0], trainer):
+                new_size = min(new_size, len(trainer.val_dataloaders[0].dataset))
```
tchaton, line 406: Should we break right now if multiple validation dataloaders are provided ?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+        if trainer.state.fn == TrainerFn.TESTING:
+            if not self._is_valid_batch_size(new_size, trainer.test_dataloaders[0], trainer):
+                new_size = min(new_size, len(trainer.test_dataloaders[0].dataset))
+        if trainer.state.fn == TrainerFn.PREDICTING:
+            if not self._is_valid_batch_size(new_size, trainer.predict_dataloaders[0], trainer):
+                new_size = min(new_size, len(trainer.predict_dataloaders[0].dataset))
+
+        changed = new_size != batch_size
+        lightning_setattr(model, self.batch_arg_name, new_size)
```
tchaton, line 415: Let's add `"{stage}_batch_size"`, so they don't get overridden if the user was to reload the dataloaders later on.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+        return new_size, changed
+
+    @staticmethod
+    def _is_valid_batch_size(batch_size: int, dataloader: DataLoader, trainer: "pl.Trainer") -> bool:
+        from pytorch_lightning.utilities.data import has_len_all_ranks
+
+        module = trainer.lightning_module or trainer.datamodule
+        return not has_len_all_ranks(dataloader, trainer.strategy, module) or batch_size <= len(dataloader)
```
tchaton, line 423: @rohitgr7 This shouldn't be enough as has_len_all_ranks validates the additional length isn't bigger than the dataset. We also need to check that each batch_size found for each rank is the same. Otherwise, the DistributedSampler on each rank won't compute the same chunks and data would be duplicated.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,423 @@
...
+    def _run_power_scaling(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", new_size: int) -> int:
+        """Batch scaling mode where the size is doubled at each iteration until an OOM error is encountered."""
+        for _ in range(self.max_trials):
+            garbage_collection_cuda()
+
+            try:
+                self._try_loop_run(trainer)
+                new_size, changed = self._adjust_batch_size(trainer, factor=2.0, desc="succeeded")
+
+                if changed:
```
tchaton, line 183: changed and new_size should be reduced between all ranks. This can be a source of process divergence here.

pytorch_lightning/loops/base.py
rohitgr7: yeah, I'll be updating this code.

tchaton:
Hey @rohitgr7. Any updates on this one ?

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: if tuner fails in the process, then we handle that case within callback, so this will run eventually.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: I need to work and check if this works with DDP or not. Right now it doesn't as per the docs.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: well `new_size` might not be the `optimal_batch_size`. More runs can happen after this one.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: will address this when DDP support will be tested/added. Adding a Misconfiguration now that this will not work with DDP.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: if the user reloads the dataloader, then this provided argument must be updated rather than adding a new one, so that the new optimal batch_size will be used to initialize the dataloader on the next run.

pytorch_lightning/tuner/tuning.py
rohitgr7: already there :)

pytorch_lightning/tuner/tuning.py
rohitgr7: sorry, didn't get you here.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: will address this with DDP+batch size finder PR.

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -304,4 +305,7 @@ def _reorder_callbacks(callbacks: List[Callback]) -> List[Callback]:
         """
```
SkafteNicki, line 305: the docstring of the method does not match anymore

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -304,4 +305,7 @@ def _reorder_callbacks(callbacks: List[Callback]) -> List[Callback]:
...
         checkpoints = [c for c in callbacks if isinstance(c, ModelCheckpoint)]
         not_checkpoints = [c for c in callbacks if not isinstance(c, ModelCheckpoint)]
-        return not_checkpoints + checkpoints
+        callbacks = not_checkpoints + checkpoints
```
SkafteNicki, line 308: nit: there seems to be some repeated pattern here that could be refactored into its own function

pytorch_lightning/utilities/exceptions.py
```diff
@@ -27,3 +27,7 @@ class ExitGracefullyException(SystemExit):
     This signals Lightning to try to create a fault-tolerance checkpoint once the current batch or epoch is reached
     (assuming it can be done under 30 sec). After the checkpoint is saved, Lightning will exit.
     """
+
+
+class _TunerExitException(Exception):
```
SkafteNicki, line 32: why is this private when all other exceptions are public?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        mode = mode.lower()
+        if mode not in self.SUPPORTED_MODES:
+            raise MisconfigurationException(f"`mode` should be one of {self.SUPPORTED_MODES}")
+
+        self.mode = mode
+        self.steps_per_trial = steps_per_trial
+        self.init_val = init_val
+        self.max_trials = max_trials
+        self.batch_arg_name = batch_arg_name
```
SkafteNicki, line 101: nit: preferably we should also do some input validation of these

pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -39,6 +39,8 @@ def scale_batch_size(
...
```
kaushikb11, line 42: Reminder to remove this.

pytorch_lightning/tuner/tuning.py
```diff
@@ -47,14 +54,37 @@ def _tune(
...
+            # TODO: Remove this once LRFinder is converted to a Callback
+            # if a datamodule comes in as the second arg, then fix it for the user
+            if isinstance(train_dataloaders, LightningDataModule):
+                datamodule = train_dataloaders
+                train_dataloaders = None
+
+            # If you supply a datamodule you can't supply train_dataloader or val_dataloaders
+            if (train_dataloaders is not None or val_dataloaders is not None) and datamodule is not None:
+                raise MisconfigurationException(
+                    "You cannot pass `train_dataloader` or `val_dataloaders` to `trainer.tune(datamodule=...)`"
```
kaushikb11, line 77: The exception message seems incorrect.

tests/tuner/test_scale_batch_size.py
```diff
@@ -295,3 +294,142 @@ def test_dataloader_reset_with_scale_batch_size(tmpdir, scale_method):
...
+    fn(model)
+    after_batch_size = model.batch_size
+    loop = getattr(trainer, f"{trainer_fn}_loop")
+
+    if trainer_fn == "fit":
+        expected_steps = trainer.train_dataloader.loaders.dataset.len // after_batch_size
+        assert trainer.global_step == expected_steps * max_epochs
+        assert trainer.current_epoch == max_epochs
+        assert loop.epoch_loop.batch_progress.total.completed == expected_steps * max_epochs
+    else:
```
kaushikb11, line 347: nit: use elif

tests/tuner/test_scale_batch_size.py
```diff
@@ -295,3 +294,142 @@ def test_dataloader_reset_with_scale_batch_size(tmpdir, scale_method):
...
+    model = BoringModel()
+    dl = model.val_dataloader()
+
+    with pytest.raises(
+        MisconfigurationException, match="please consider setting `train_dataloaders` and `val_dataloaders` instead"
+    ):
+        trainer.tune(model, dataloaders=dl, method="fit")
+
+
+def test_batch_size_finder_with_distributed_strategies(tmpdir):
```
kaushikb11, line 410:
```suggestion
def test_batch_size_finder_with_distributed_strategies():
```

pytorch_lightning/tuner/tuning.py
```diff
@@ -197,3 +246,34 @@ def lr_find(
...
+    val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+    dataloaders: Optional[EVAL_DATALOADERS] = None,
+    method: str = "fit",
+):
+    supported_methods = ("fit", "validate", "test", "predict")
+    if method not in supported_methods:
+        raise MisconfigurationException(f"method {method!r} is invalid. Should be one of {supported_methods}.")
+
+    if method == "fit":
+        if dataloaders is not None:
```
kaushikb11, line 263: Could be a single condiiton statement

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -304,4 +305,7 @@ def _reorder_callbacks(callbacks: List[Callback]) -> List[Callback]:
...
+        tuner_callbacks = [c for c in callbacks if isinstance(c, BatchSizeFinder)]
+        non_tuner_callbacks = [c for c in callbacks if not isinstance(c, BatchSizeFinder)]
+        return tuner_callbacks + non_tuner_callbacks
```
kaushikb11, line 311: I think the iteration happening here could be optimized

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,425 @@
...
+                It is expected that the user has provided a model or datamodule that has a hyperparameter
+                with that name. We will look for this attribute name in the following places
+
+                - ``model``
+                - ``model.hparams``
+                - ``trainer.datamodule`` (the datamodule passed to the tune method)
+        """
+        mode = mode.lower()
+        if mode not in self.SUPPORTED_MODES:
+            raise MisconfigurationException(f"`mode` should be one of {self.SUPPORTED_MODES}")
```
kaushikb11, line 95:
```suggestion
            raise MisconfigurationException(f"`mode` should be either of {self.SUPPORTED_MODES}")
```

pytorch_lightning/callbacks/batch_size_finder.py
kaushikb11: Name it `new_batch_size` instead.

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,425 @@
...
+            trainer.progress_bar_callback.disable()
+
+        new_size, _ = self._adjust_batch_size(trainer, value=self.init_val)
+
+        if self.mode == "power":
+            new_size = self._run_power_scaling(trainer, pl_module, new_size)
+        elif self.mode == "binsearch":
+            new_size = self._run_binary_scaling(trainer, pl_module, new_size)
+
+        garbage_collection_cuda()
```
kaushikb11, line 172: Run it only for `GPUAccelerator`?

pytorch_lightning/callbacks/batch_size_finder.py
kaushikb11: Curious. What issues are limiting us to add DDP support?

pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,425 @@
...
+            trainer.limit_val_batches = self._dumped_params["limit_val_batches"]
+        elif trainer.state.fn == "validate":
+            loop = trainer.validate_loop
+            trainer.limit_val_batches = self._dumped_params["limit_eval_batches"]
+        elif trainer.state.fn == "test":
+            loop = trainer.test_loop
+            trainer.limit_test_batches = self._dumped_params["limit_eval_batches"]
+        elif trainer.state.fn == "predict":
+            loop = trainer.predict_loop
+            trainer.limit_predict_batches = self._dumped_params["limit_eval_batches"]
```
kaushikb11, line 353: This could be three lines with the help of f-strings and getattr

pytorch_lightning/utilities/exceptions.py
rohitgr7: this is not something that is user-facing. Also trying to avoid it being documented inside API references.

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: added a TODO for now.. will add them in a follow-up since we need to add tests for them too and if I add it here, the PR will grow more and will be hard for reviewers to review it properly.

pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: ah! must have come back while rebasing. removed!!

tests/tuner/test_scale_batch_size.py
rohitgr7: why? we have some common assertions for other states here.

pytorch_lightning/tuner/tuning.py
rohitgr7: then, if method is "fit" and dataloaders is None, it will fall under the else condition and will fail for the correct case I think

pytorch_lightning/callbacks/batch_size_finder.py
rohitgr7: done! is there any way to clear cache for TPU/IPUs? maybe we cab define a single utility for all/

pytorch_lightning/trainer/connectors/callback_connector.py
rohitgr7: good catch! updated!

pytorch_lightning/trainer/trainer.py
```diff
@@ -715,6 +719,18 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
...
+        except _TunerExitException as exception:
+            self.state.status = TrainerStatus.FINISHED
+            if distributed_available() and self.world_size > 1:
+                # try syncing remaing processes, kill otherwise
+                self.strategy.reconciliate_processes(traceback.format_exc())
+            self._on_exception()
+            # reset bookkeeping
+            self.state.stage = None
+            self._call_callback_hooks("on_exception", exception)
+            self._teardown()
```
rohitgr7, line 732:
```suggestion
            self._call_callback_hooks("on_exception", exception)
            self._teardown()
            self.state.stage = None
```

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -217,35 +234,78 @@ def _adjust_batch_size(
...
+
     module = trainer.lightning_module or trainer.datamodule
     return not has_len_all_ranks(dataloader, trainer.strategy, module) or batch_size <= len(dataloader)
+
+
+def _collect_garbage(trainer: "pl.Trainer") -> None:
+    from pytorch_lightning.accelerators.gpu import GPUAccelerator
+
+    if isinstance(trainer.accelerator, GPUAccelerator):
+        garbage_collection_cuda()
```
awaelchli, line 290: All this function does is basically call `torch.cuda.empty_cache()`, and it is no-op on non-cuda devices. So we don't need to have this check for GPUAccelerator

tests/tests_pytorch/tuner/test_scale_batch_size.py
```diff
@@ -281,9 +280,9 @@ def __init__(self):
         auto_scale_batch_size="ThisModeDoesNotExist",
     )
 
-    with pytest.raises(ValueError, match="could either be `power` or `binsearch`"):
+    with pytest.raises(MisconfigurationException, match="should be either of"):
         trainer.tune(model)
-    with pytest.raises(ValueError, match="could either be `power` or `binsearch`"):
+    with pytest.raises(MisconfigurationException, match="should be either of"):
```
awaelchli, line 285: ValueError was appropriate. We should keep it. 
See also #10918

tests/tests_pytorch/tuner/test_scale_batch_size.py
```diff
@@ -235,7 +234,7 @@ def test_error_on_dataloader_passed_to_fit(tmpdir):
 
     with pytest.raises(
         MisconfigurationException,
-        match="The batch scaling feature cannot be used with dataloaders passed directly",
+        match="Batch size finder cannot be used with dataloaders passed directly",
```
awaelchli, line 237: Why drop the article 🤣 The poor articles. They are important!

awaelchli (approved):
High-level review! Great work

src/pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,146 @@
...
+                with that name. We will look for this attribute name in the following places
+
+                - ``model``
+                - ``model.hparams``
+                - ``trainer.datamodule`` (the datamodule passed to the tune method)
+        """
+        # TODO: Add input validation.
+        mode = mode.lower()
+        if mode not in self.SUPPORTED_MODES:
+            raise MisconfigurationException(f"`mode` should be either of {self.SUPPORTED_MODES}")
```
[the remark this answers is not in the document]
rohitgr7: I prefer a list so that the possible values and easily be distinguished in a container in the error message.

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1017,9 +1032,11 @@ def tune(
         model: "pl.LightningModule",
         train_dataloaders: Optional[Union[TRAIN_DATALOADERS, LightningDataModule]] = None,
         val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+        dataloaders: Optional[EVAL_DATALOADERS] = None,
```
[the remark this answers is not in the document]
rohitgr7: if method is fit (default) and dataloaders is not None: Exception

src/pytorch_lightning/loggers/base.py
```diff
@@ -17,7 +17,7 @@
 import numpy as np
 
 import pytorch_lightning.loggers.logger as logger
-from pytorch_lightning.utilities.warnings import rank_zero_deprecation
+from pytorch_lightning.utilities.rank_zero import rank_zero_deprecation
```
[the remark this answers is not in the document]
rohitgr7: some issue was coming up. I don't remember but it's fine since this is the updated module.

tests/tests_pytorch/tuner/test_scale_batch_size.py
rohitgr7: reverted

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -650,6 +655,16 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
                 return self.strategy.launcher.launch(trainer_fn, *args, trainer=self, **kwargs)
             else:
                 return trainer_fn(*args, **kwargs)
+
+        except _TunerExitException as exception:
+            self.state.status = TrainerStatus.FINISHED
+            if distributed_available() and self.world_size > 1:
+                # try syncing remaing processes, kill otherwise
+                self.strategy.reconciliate_processes(traceback.format_exc())
```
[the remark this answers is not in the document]
rohitgr7: done!

src/pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,146 @@
...
+
+            batch_arg_name: name of the attribute that stores the batch size.
+                It is expected that the user has provided a model or datamodule that has a hyperparameter
+                with that name. We will look for this attribute name in the following places
+
+                - ``model``
+                - ``model.hparams``
+                - ``trainer.datamodule`` (the datamodule passed to the tune method)
+        """
+        # TODO: Add input validation.
```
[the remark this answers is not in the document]
rohitgr7: oh no! redundant TODO. I added this for me while working on this.

src/pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -237,19 +238,30 @@ def _attach_model_callbacks(self) -> None:
 
     @staticmethod
     def _reorder_callbacks(callbacks: List[Callback]) -> List[Callback]:
```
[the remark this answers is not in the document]
rohitgr7: done!

src/pytorch_lightning/callbacks/batch_size_finder.py
```diff
@@ -0,0 +1,145 @@
...
+
+class BatchSizeFinder(Callback):
+    SUPPORTED_MODES = ("power", "binsearch")
+
+    def __init__(
+        self,
+        mode: str = "power",
+        steps_per_trial: int = 3,
+        init_val: int = 2,
+        max_trials: int = 25,
```
Borda, line 39: Can we also set at what training stage it is called - step/epoch?
Which is the reason doe maximize performances for freeze/unfreeze
(then LR shall be able to do the same)

rohitgr7: sorry, didn't get you here.

otaj (approved):

src/pytorch_lightning/callbacks/batch_size_finder.py
Borda: as spoke, lets do it as follow-up PR :)

Borda (approved):

src/pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -229,19 +230,30 @@ def _attach_model_callbacks(self) -> None:
...
-        checkpoints: List[Callback] = [c for c in callbacks if isinstance(c, Checkpoint)]
-        not_checkpoints = [c for c in callbacks if not isinstance(c, Checkpoint)]
-        return not_checkpoints + checkpoints
+        tuner_callbacks: List[Callback] = []
+        other_callbacks: List[Callback] = []
+        checkpoint_callbacks: List[Callback] = []
+
+        for cb in callbacks:
+            if isinstance(cb, BatchSizeFinder):
+                tuner_callbacks.append(cb)
```
carmocca, line 250: Since you called this "tuner callbacks" does that mean that LRFinder will also need to be at the beginning?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1018,44 +1031,34 @@ def tune(
...
+            dataloaders: A :class:`torch.utils.data.DataLoader` or a sequence of them specifying val/test/predict
+                samples used for running tuner on validation/testing/prediction.
+
             datamodule: An instance of :class:`~pytorch_lightning.core.datamodule.LightningDataModule`.
 
             scale_batch_size_kwargs: Arguments for :func:`~pytorch_lightning.tuner.batch_size_scaling.scale_batch_size`
 
             lr_find_kwargs: Arguments for :func:`~pytorch_lightning.tuner.lr_finder.lr_find`
+
+            method: Method to run tuner on. It can be any of ``("fit", "validate", "test", "predict")``.
```
carmocca, line 1043: So tuner now supports the other methods? Can you mention this in the CHANGELOG?

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -69,94 +53,120 @@ def scale_batch_size(
...
+        "loggers": trainer.loggers,
         "callbacks": trainer.callbacks,
-        "auto_scale_batch_size": trainer.auto_scale_batch_size,
-        "auto_lr_find": trainer.auto_lr_find,
-        "limit_train_batches": trainer.limit_train_batches,
     }
+    if trainer.state.fn == "fit":
+        loop = trainer.fit_loop
+        dumped_params["max_steps"] = trainer.max_steps
+        dumped_params["limit_val_batches"] = trainer.limit_val_batches
```
carmocca, line 86: Isn't limit_train_batches missing now? It was there in the previous code

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -237,40 +246,80 @@ def _adjust_batch_size(
...
-def _is_valid_batch_size(batch_size: int, dataloader: DataLoader, trainer: "pl.Trainer"):
+def _is_valid_batch_size(batch_size: int, dataloader: DataLoader, trainer: "pl.Trainer") -> bool:
+    from pytorch_lightning.utilities.data import has_len_all_ranks
+
     module = trainer.lightning_module or trainer.datamodule
     return not has_len_all_ranks(dataloader, trainer.strategy, module) or batch_size <= len(dataloader)
 
 
+def _collect_garbage() -> None:
+    garbage_collection_cuda()
```
carmocca, line 299: What's the purpose of this extra function?

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -237,40 +246,80 @@ def _adjust_batch_size(
...
+
+        assert trainer.train_dataloader is not None
+        if not _is_valid_batch_size(new_size, trainer.train_dataloader, trainer):
+            new_size = min(new_size, len(trainer.train_dataloader.dataset))
+    else:
+        stage = trainer.state.stage
+        assert stage is not None
+        dataloaders = getattr(trainer, f"{stage.dataloader_prefix}_dataloaders")
+        if dataloaders is None:
+            getattr(trainer, f"reset_{stage.dataloader_prefix}_dataloader")()
```
carmocca, line 279: Shouldn't you use the `_reset_dataloaders` utility you added below?

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -62,14 +69,38 @@ def _tune(
...
                 scale_batch_size_kwargs.setdefault("mode", self.trainer.auto_scale_batch_size)
-            result["scale_batch_size"] = scale_batch_size(self.trainer, model, **scale_batch_size_kwargs)
+
+            result["scale_batch_size"] = self.scale_batch_size(
+                model, train_dataloaders, val_dataloaders, dataloaders, datamodule, method, **scale_batch_size_kwargs
+            )
 
         # Run learning rate finder:
         if self.trainer.auto_lr_find:
+            self.trainer.state.fn = TrainerFn.TUNING
```
carmocca, line 79: Should we think of removing `TrainerFn.TUNING`? Since it can call all of the other trainer functions internally

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -62,14 +69,38 @@ def _tune(
...
+            if isinstance(train_dataloaders, LightningDataModule):
+                datamodule = train_dataloaders
+                train_dataloaders = None
+
+            # If you supply a datamodule you can't supply train_dataloader or val_dataloaders
+            if (train_dataloaders is not None or val_dataloaders is not None) and datamodule is not None:
+                raise MisconfigurationException(
+                    "You cannot pass `train_dataloader` or `val_dataloaders` to `trainer.tune()`"
+                    " if datamodule is already passed."
+                )
```
carmocca, line 94: Can you add support for TUNING in attach_data instead? As in https://github.com/Lightning-AI/lightning/pull/14637

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -129,22 +167,34 @@ def scale_batch_size(
...
+            init_val=init_val,
+            max_trials=max_trials,
+            batch_arg_name=batch_arg_name,
         )
+        # do not continue with the loop in case trainer.tuner is used
+        batch_size_finder._early_exit = True
+        self.trainer.callbacks = [batch_size_finder] + self.trainer.callbacks
+
+        if method == "fit":
+            self.trainer.fit(model, train_dataloaders, val_dataloaders, datamodule)
```
carmocca, line 187: This will overwrite the `Tuning` TrainerFn. so might as well remove it

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -212,3 +262,34 @@ def lr_find(
...
+            raise MisconfigurationException(
+                f"In tuner with method={method!r}, `dataloaders` argument should be None,"
+                " please consider setting `train_dataloaders` and `val_dataloaders` instead."
+            )
+    else:
+        if train_dataloaders is not None or val_dataloaders is not None:
+            raise MisconfigurationException(
+                f"In tuner with `method`={method!r}, `train_dataloaders` and `val_dataloaders`"
+                " arguments should be None, please consider setting `dataloaders` instead."
+            )
```
carmocca, line 289: Wouldn't these validations be covered by the code added in https://github.com/Lightning-AI/lightning/pull/14637?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -637,6 +641,13 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
                 return self.strategy.launcher.launch(trainer_fn, *args, trainer=self, **kwargs)
             else:
                 return trainer_fn(*args, **kwargs)
+
+        except _TunerExitException as exception:
+            self.state.status = TrainerStatus.FINISHED
+            self._call_callback_hooks("on_exception", exception)
```
carmocca, line 647: I interpret `on_exception` as a hook to inject logic when an exception is raised that messes up your run.

But this `_TunerExitException` is not really an exception in the sense that something went wrong. You are just using it to bypass everything and stop.

Could we avoid using an exception and instead do `loop.should_stop`? Then training would stop normally

carmocca (changes requested):

src/pytorch_lightning/trainer/connectors/callback_connector.py
rohitgr7: yes. in the other PR I am adding it

src/pytorch_lightning/trainer/trainer.py
rohitgr7: done!

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: added back!

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: there's a TODO. Right now it's checked only for train, so no need to use that one.

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: ah! I was trying to add this utility for all types of accelerators, but couldn't find a way. reverted.

src/pytorch_lightning/tuner/tuning.py
rohitgr7: yes, in a follow-up. Since lr_finder wasn't merged, I didn't remove it here.

rohitgr7: added a TODO.

src/pytorch_lightning/tuner/tuning.py
rohitgr7: this is a little different. here we are accepting both `train/val_dataloaders` and `dataloaders`. So if a user passes all of them for some reason, this error is being raised.

src/pytorch_lightning/tuner/tuning.py
rohitgr7: in a followup once both the PR get's merged.

src/pytorch_lightning/tuner/tuning.py
rohitgr7: it's removed in the LRfinder PR. We won't need it.

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -637,6 +641,13 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
...
+            self._teardown()
+            self.state.stage = None
```
rohitgr7, line 649:
```suggestion
            self._call_teardown_hook()
            self._teardown()
            self.state.stage = None
            self.state.status = TrainerStatus.FINISHED
```
need to think more. For now can we do this? Even if we do `loop.shoud_stop`, it will update this state of the loop and during the next run of let's say `trainer.fit/validate`, loop might not start because the last exit will have `loop.shoud_stop=True`.

src/pytorch_lightning/trainer/trainer.py
rohitgr7: answered here: https://github.com/Lightning-AI/lightning/pull/11089/files#r972909982

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -212,3 +263,34 @@ def lr_find(
...
+def _check_tuner_configuration(
+    trainer: "pl.Trainer",
+    train_dataloaders: Optional[Union[TRAIN_DATALOADERS, "pl.LightningDataModule"]] = None,
+    val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+    dataloaders: Optional[EVAL_DATALOADERS] = None,
+    method: str = "fit",
+) -> None:
+    supported_methods = ("fit", "validate", "test", "predict")
+    if method not in supported_methods:
+        raise MisconfigurationException(f"method {method!r} is invalid. Should be one of {supported_methods}.")
```
carmocca, line 277: Let's avoid using MisconfigurationException and use more standard exceptions (ValueError for the ones that check arguments) instead: #10918

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -129,22 +167,35 @@ def scale_batch_size(
...
-            val_dataloaders=val_dataloaders,
-            datamodule=datamodule,
-            scale_batch_size_kwargs={
-                "mode": mode,
-                "steps_per_trial": steps_per_trial,
-                "init_val": init_val,
-                "max_trials": max_trials,
-                "batch_arg_name": batch_arg_name,
-            },
+        # TODO: TrainerFn.TUNING since we are now calling fit/validate/test/predict methods directly
```
carmocca, line 170:
```suggestion
        # TODO: Remove TrainerFn.TUNING since we are now calling fit/validate/test/predict methods directly
```

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -129,22 +167,35 @@ def scale_batch_size(
...
+        batch_size_finder: Callback = BatchSizeFinder(
+            mode=mode,
+            steps_per_trial=steps_per_trial,
+            init_val=init_val,
+            max_trials=max_trials,
+            batch_arg_name=batch_arg_name,
         )
+        # do not continue with the loop in case trainer.tuner is used
+        batch_size_finder._early_exit = True
+        self.trainer.callbacks = [batch_size_finder] + self.trainer.callbacks
```
carmocca, line 185: Setting it at the begginning here is not important, no? As it will be reordered later again

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -212,3 +263,34 @@ def lr_find(
...
         self.trainer.auto_lr_find = False
         return result["lr_find"]
+
+
+def _check_tuner_configuration(
+    trainer: "pl.Trainer",
+    train_dataloaders: Optional[Union[TRAIN_DATALOADERS, "pl.LightningDataModule"]] = None,
+    val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+    dataloaders: Optional[EVAL_DATALOADERS] = None,
+    method: str = "fit",
```
carmocca, line 273: You could use `Literal` here

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -85,7 +116,9 @@ def scale_batch_size(
         model: "pl.LightningModule",
         train_dataloaders: Optional[Union[TRAIN_DATALOADERS, "pl.LightningDataModule"]] = None,
         val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+        dataloaders: Optional[EVAL_DATALOADERS] = None,
         datamodule: Optional["pl.LightningDataModule"] = None,
+        method: str = "fit",
```
carmocca, line 121: You could use `Literal` here

src/pytorch_lightning/tuner/tuning.py
```diff
@@ -41,8 +43,13 @@ def on_trainer_init(self, auto_lr_find: Union[str, bool], auto_scale_batch_size:
     def _tune(
         self,
         model: "pl.LightningModule",
+        train_dataloaders: Optional[Union[TRAIN_DATALOADERS, LightningDataModule]] = None,
+        val_dataloaders: Optional[EVAL_DATALOADERS] = None,
+        dataloaders: Optional[EVAL_DATALOADERS] = None,
+        datamodule: Optional[LightningDataModule] = None,
         scale_batch_size_kwargs: Optional[Dict[str, Any]] = None,
         lr_find_kwargs: Optional[Dict[str, Any]] = None,
+        method: str = "fit",
```
carmocca, line 52: You could use `Literal` here

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -237,40 +249,76 @@ def _adjust_batch_size(
...
 
 
+def _reset_dataloaders(trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+    if trainer.state.fn == "fit":
+        trainer.reset_train_dataloader(pl_module)
+        trainer.reset_val_dataloader(pl_module)
+    else:
+        stage = trainer.state.stage
+        assert stage is not None
+        getattr(trainer, f"reset_{stage.dataloader_prefix}_dataloader")(pl_module)
```
carmocca, line 308: nit:

```suggestion
        reset_fn = getattr(trainer, f"reset_{stage.dataloader_prefix}_dataloader")
        reset_fn(pl_module)
```

src/pytorch_lightning/tuner/batch_size_scaling.py
carmocca: Which TODO? The one in L269?

What is only checked for train?

Couldn't this do:


```python
_reset_dataloaders(trainer, model)
if trainer.state.fn == "fit":
    assert trainer.train_dataloader is not None
    if not _is_valid_batch_size(new_size, trainer.train_dataloader, trainer):
        new_size = min(new_size, len(trainer.train_dataloader.dataset))
else:
    assert dataloaders is not None
    if not _is_valid_batch_size(new_size, dataloaders[0], trainer):
        new_size = min(new_size, len(dataloaders[0].dataset))
```

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -69,61 +53,95 @@ def scale_batch_size(
...
+        if hasattr(loop, "verbose"):
+            loop.verbose = False
 
 
 def __scale_batch_restore_params(trainer: "pl.Trainer", params: Dict[str, Any]) -> None:
-    trainer.auto_scale_batch_size = params["auto_scale_batch_size"]
-    trainer.auto_lr_find = params["auto_lr_find"]
-    trainer.fit_loop.max_steps = params["max_steps"]
-    trainer.logger = params["logger"]
+    # TODO: There are more states that needs to be reset (#4512 and #4870)
```
carmocca, line 122: What would those be?

src/pytorch_lightning/trainer/trainer.py
carmocca: Yes. I think the tuner callback should completely reset the loop states on teardown.

rohitgr7: rather I'd say it should restore the loop state which it does while restoring the params.

but if we do:
> Could we avoid using an exception and instead do loop.should_stop?

then this loop state `loop.should_stop` will be True and during the next run, it will exit without running anything
```
trainer.tune() -> sets `loop.should_stop = False`
trainer.fit() -> might not loop

```

src/pytorch_lightning/tuner/tuning.py
rohitgr7: done!

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: one is profiler. Not sure about the other one. Added it here to investigate when this TODO is WIP

src/pytorch_lightning/tuner/tuning.py
rohitgr7: yep, it isn't.

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: during fit, we can have both val and train dataloaders, but we can't tell whether users are tuning for val_Dataloaders too, or not. So, for now, I have not updated the logic as before, and we only track the total batch size for train dataloaders during fit, but this is a design ques, should we consider val dataloaders here or not?

also, it should not reload the dataloaders if they are already reloaded. For eg in the case of the fine-tuning stage.

rohitgr7: updated the TODOs

src/pytorch_lightning/tuner/tuning.py
carmocca: There are other instances that should be updated

src/pytorch_lightning/trainer/trainer.py
carmocca: Just as the Callback could signal to stop, it could also unsignal it on teardown, no?

carmocca (approved):
Synced outside of GitHub! LGTM! Great and challenging work

We can try to remove the exception in the future if we finish #8604

PR #11101, opened 2021-12-16
Remove `hpc_save`

jjenniferdai:
## What does this PR do?

Item 2. in https://github.com/PyTorchLightning/pytorch-lightning/issues/9407#issuecomment-991399437
Part of #9407 cleanup, remove `hpc_save`.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -299,39 +299,11 @@ def restore_lr_schedulers(self) -> None:
     # PRIVATE OPS
     # ----------------------------------
 
-    def hpc_save(self, folderpath: str, logger: Optional[LightningLoggerBase]) -> str:
-        # make sure the checkpoint folder exists
+    def hpc_save_path(self, folderpath: str) -> str:
```
ananthsub, line 302: super nit: you could make this, `max_ckpt_version_in_folder`, and `get_max_ckpt_path_from_folder` staticmethods. that's usually a sign that this class can be further streamlined by moving those out to utility functions. that could be for a separate issue/PR

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -61,7 +61,11 @@ def slurm_sigusr1_handler_fn(self, signum: Signals, frame: FrameType) -> None:
         if self.trainer.is_global_zero:
             # save weights
             log.info("handling SIGUSR1")
-            self.trainer.checkpoint_connector.hpc_save(self.trainer.weights_save_path, self.trainer.logger)
+            # save logger to make sure we get all the metrics
+            if self.trainer.logger:
+                self.trainer.logger.finalize("finished")
+            hpc_save_path = self.trainer.checkpoint_connector.hpc_save_path(self.trainer.weights_save_path)
+            self.trainer.save_checkpoint(hpc_save_path)
```
ananthsub, line 68: this can't be under a `self.trainer.is_global_zero` check from L61. `trainer.save_checkpoint` relies on collectives, meaning it must be called on all ranks simultaneously. otherwise the user will see hangs in a distributed environment. 

To minimize changes in this PR, you could encapsulate just the logger finalization call under the `is_global_zero` check. `trainer.save_checkpoint` already takes care of all the other rank-specific checks. 

it looks like we don't have a test with deepspeed-3 and SLURM signal handling. in the code today, this would crash, because we need to save on all ranks in that case. 

so this is an awesome unification you're doing so we don't have to worry about divergent code paths like this anymore 😄

ananthsub (changes requested):

pytorch_lightning/trainer/connectors/signal_connector.py
ananthsub: looking at what's done in the main trainer, the logger finalization is called on all ranks, and it's up to logger implementations to determine if they only want to be called on rank 0 or not.

https://github.com/PyTorchLightning/pytorch-lightning/blob/7ed3dbf1913351868835ac3c549a0fbf6c612e09/pytorch_lightning/trainer/trainer.py#L1429-L1446

So I don't think we need to guard this with the global zero check either.

jjenniferdai: I think* the second finalize call can be removed, but someone correct me if I'm missing something :)

tchaton: Hey @ananthsub. Great catch ! Yes, this should be covering only the requeing.

Yes, we should be call the logger finalize on all ranks, so let's remove the second one.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -489,3 +422,54 @@ def _get_loops_state_dict(self) -> Dict[str, Any]:
             "test_loop": self.trainer.test_loop.state_dict(),
             "predict_loop": self.trainer.predict_loop.state_dict(),
         }
+
+    # ----------------------------------
+    # UTILITIES
+    # ----------------------------------
+
+    @staticmethod
+    def max_ckpt_version_in_folder(dir_path: _PATH, name_key: str = "ckpt_") -> Optional[int]:
```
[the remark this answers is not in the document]
jjenniferdai: protected? they're sort of like utility functions, `hpc_save_path()` gets called in `slurm_sigusr1_handler_fn`

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -61,8 +61,14 @@ def slurm_sigusr1_handler_fn(self, signum: Signals, frame: FrameType) -> None:
         if self.trainer.is_global_zero:
             # save weights
             log.info("handling SIGUSR1")
```
ananthsub, line 67: super nit
```suggestion
        rank_zero_info("handling SIGUSR1")
```

pytorch_lightning/trainer/connectors/checkpoint_connector.py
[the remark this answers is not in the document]
ananthsub: @jjenniferdai you could make the utilities which are only called within this class private

pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -65,8 +65,14 @@ def slurm_sigusr1_handler_fn(self, signum: _SIGNUM, frame: FrameType) -> None:
...
-            self.trainer.checkpoint_connector.hpc_save(self.trainer.weights_save_path, self.trainer.logger)
 
+        # save logger to make sure we get all the metrics
+        if self.trainer.logger:
+            self.trainer.logger.finalize("finished")
+        hpc_save_path = self.trainer.checkpoint_connector.hpc_save_path(self.trainer.weights_save_path)
+        self.trainer.save_checkpoint(hpc_save_path)
```
ananthsub, line 73: this is so much cleaner!

ananthsub (approved):

pytorch_lightning/trainer/connectors/checkpoint_connector.py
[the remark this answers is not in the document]
jjenniferdai: some tests calling them though, should we just update those with e.g. `_CheckpointConnector__ max_ckpt_version_in_folder` ?

justusschock (approved):

ananthsub (approved):

awaelchli (approved):

carmocca (approved):

Borda (approved):

PR #11119, opened 2021-12-17
Reset the current progress tracking state during double evaluation

carmocca:
## What does this PR do?

With fault-tolerance disabled, the following runs twice through the test set, it's equivalent to `max_test_epochs=2`

```python
trainer = Trainer(...)
trainer.test(...)
trainer.test(...)
```

However, currently in master, if fault-tolerance is enabled, the second test exits early. This is because the `done` condition for the test loop has been already fulfilled:

```py
self.dataloader_progress.current.completed >= self.num_dataloaders
```

To avoid this, we need to check whether we are NOT fitting, and whether the previous state is done, then in that case reset the state.

This is different from how fit works where this is the expected behavior and the user is required to set a different `max_epochs` value. The equivalent would be that the user has to change `self.num_dataloaders` which does not make sense.

Predict also has this bug, but there we don't need the fit check.

### Test plan

This is covered by existing tests, for example: `test_checkpoint_path_input`

None

## Before submitting

## PR review

cc @carmocca @justusschock @awaelchli @ninginthecloud @borda

### DIFF
--- a/pytorch_lightning/loops/dataloader/evaluation_loop.py
+++ b/pytorch_lightning/loops/dataloader/evaluation_loop.py
@@ -84,6 +84,10 @@ class EvaluationLoop(DataLoaderLoop):
             self._max_batches = [self._max_batches] * len(self.dataloaders)
 
         super().reset()
+        # when restarting, if we are running `validate` or `test` twice, since there's no concept of `max_epochs` we
+        # need to reset the current state when the loop has finished running
+        if self.done and self.trainer.state.fn != TrainerFn.FITTING:
+            self.dataloader_progress.reset_on_run()
 
     def on_skip(self) -> List:
         return []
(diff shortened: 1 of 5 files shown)


pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -80,6 +81,10 @@ def reset(self) -> None:
             self._max_batches = [self._max_batches] * len(self.dataloaders)
 
         super().reset()
+        # when restarting, if we are running `validate` or `test` twice, since there's no concept of `max_epochs` we
+        # need to reset the current state when the loop has finished running
+        if self.done and self.trainer.state.fn != TrainerFn.FITTING:
+            self.dataloader_progress.reset_on_run()
```
carmocca, line 87: I considered putting this in the super implementation, but I don't think the `state.fn` check makes sense there.

justusschock (approved):

tchaton (approved):
LGTM !

pytorch_lightning/loops/dataloader/evaluation_loop.py
rohitgr7: quick ques: fault_tolerant doesn't restart/reset the states for `.validate/.test/.predict` call?

carmocca: It does, see the super implementation.

The bug described here is only for when it is called more than once

Borda (approved):

Issue #10549, opened 2021-11-15
[RFC] Introduce `Strategy` in favor of `TrainingTypePlugin`


kaushikb11:
## Proposed refactor

Part of #9932 
Related PR: #10548 

### Motivation

As we are introducing `StrategyPlugin` in favor of `TrainingTypePlugin` for this release.

@ananthsub  raised some really good points for discussion regarding `StrategyPlugin`. [Reference](https://github.com/PyTorchLightning/pytorch-lightning/pull/10548#issuecomment-969279546)

### Proposed solution

Three main points to be discussed here

* Drop `Plugin` from the name of the class
* It should be under `pytorch_lightning/strategy` directory
* Over time, these classes should be renamed as `DDPStrategy` and so on.

Would like your thoughts? @PyTorchLightning/core-contributors 

cc @borda @justusschock @awaelchli @akihironitta @kaushikb11 @ananthsub

awaelchli:
Yes, definitely. We agreed on this I think. This was already set in motion by the introduction of the strategy argument in the Trainer.

> Over time, these classes should be renamed as DDPStrategy and so on.

Has to be done for 1.6. We are targeting stable API for these components at 1.6.

awaelchli:
Should we name all classes without the "Plugin"/"Strategy" suffix? Here is how this would look like:

DataParallel, (or just DP)
DDP2,
DDP,
DDPSpawn,
DDPFullySharded,
DeepSpeed,
Horovod,
IPU,
SingleDevice,
SingleTPU,
TPUSpawn,
TrainingType,
Parallel,
DDPSharded,
DDPSpawnSharded,

ananthsub:
On naming, Loggers, Profilers, Accelerators, and ClusterEnvironments are suffixed with the component name (e.g. TensorboardLogger, SimpleProfiler, GPUAccelerator, SlurmEnvironment). 

If we drop the suffix for strategy:

- Would we also make this change for other components of Lightning so the naming scheme is unified across components (leading to one less thing to think about for users) ? 
- Do you think there's a risk with naming collisions? For instance, the IPU accelerator and IPU strategy could both be IPU classes. As a result, users would likely need to change the import for them to be:

```
from pytorch_lightning.accelerators import IPU as IPUAccelerator
from pytorch_lightning.strategy import IPU as IPUStrategy
```

awaelchli:
Proposed plan:

1. Rename classes XPlugin to XStrategy
[...]

PR #11120, opened 2021-12-17
Rename the `TrainingTypePlugin` base to `Strategy`

awaelchli:
## What does this PR do?

Renames the base class to `Strategy`. 
The next PRs will cover all related usages:
- Renaming subclasses, e.g., DDPPlugin -> DDPStrategy
- module training_type -> strategy
- Registry: #11122
- other variable names across the project

Part of #10549

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @akihironitta


carmocca (approved):

justusschock (approved):

docs/source/extensions/plugins.rst
```diff
@@ -29,7 +29,7 @@ We expose Accelerators and Plugins mainly for expert users that want to extend L
 
 There are two types of Plugins in Lightning with different responsibilities:
 
-TrainingTypePlugin
+Strategy
 ------------------
```
rohitgr7, line 33:
```suggestion
Strategy
--------
```

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -37,7 +37,7 @@
 TBroadcast = TypeVar("TBroadcast")
 
 
-class TrainingTypePlugin(ABC):
+class Strategy(ABC):
```
rohitgr7, line 40: the filename will be renamed later?

also, this isn't going to follow a deprecation process?

justusschock: This was still marked as experimental, not sure if we need one here

awaelchli: Yes, (my) plan is to rename components first, then the files. 
Happy to hear suggestions how to "deprecate" things, but as Justus said it is inevitable that 1.6 will break people's custom plugins unfortunately, considering all the changes we already did. 

But we can deprecate things on an import level for example. People importing from `plugins.training_type` should get a warning that it was moved to `strategy.x`, maybe?

rohitgr7: if it's experimental then it's fine.
is it targeted to stabilize by v1.6?

rohitgr7 (approved):

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: 100%

four4fish:
I marked this as breaking change as user who extend TrainingTypePlugin will need to update their class inheritance

pytorch_lightning/trainer/trainer.py
```diff
@@ -1681,7 +1681,7 @@ def accelerator(self) -> Accelerator:
         return self.training_type_plugin.accelerator
 
     @property
-    def training_type_plugin(self) -> TrainingTypePlugin:
+    def training_type_plugin(self) -> Strategy:
```
four4fish, line 1684: Are we planning on completely get rid of training_type_plugin this word in code base? It still heavily used in trainer.py and accelerator_connector.py

four4fish (approved):

pytorch_lightning/plugins/training_type/training_type_plugin.py
Borda: > Yes, our plan is to rename components first, then the files.

I would rather make this in single PR, as it trivial rename shall be fine

Borda (approved):

pytorch_lightning/trainer/trainer.py
awaelchli: Yes, everywhere. It only makes sense if it is consistent everywhere.

pytorch_lightning/plugins/training_type/training_type_plugin.py
awaelchli: Unfortunately this will result in 100+ files changed, nobody would review this. I would prefer to take it in smaller pieces.

tchaton (approved):
LGTM !

kaushikb11 (changes requested):
This is a heavy breaking change on the import level itself for users who extend TrainingTypePlugin. Can’t we prevent it by introducing `strategy.py` file in the PR?

justusschock:
@kaushikb11 But then we only delay the breaking changes. I think we could introduce a `training_type_plugin` file for BC, but I really think we should go with these changes as they were considered to be experimental. And since we want to have it stable for 1.6, we cannot delay the BC beyond this

awaelchli:
@kaushikb11 To prevent a huge PR, I would prefer to go step by step by renaming parts individually. Once we move the files we can still add a simple reroute/deprecation path for mapping the training_type module to strategy

kaushikb11 (approved):

PR #11132, opened 2021-12-17
Fix evaluation logging on epoch end with multiple dataloaders

### LINKED ISSUE #10904: Logging does not work on evaluation epoch end with multiple dataloaders

carmocca:
## What does this PR do?

Fixes #10904

None

## Before submitting

## PR review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -311,6 +311,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed bug where `Trainer(track_grad_norm=..., logger=False)' would fail ([#11114](https://github.com/PyTorchLightning/pytorch-lightning/pull/11114))
 
 
+- Fixed logging on `{test,validation}_epoch_end` with multiple dataloaders ([#11132](https://github.com/PyTorchLightning/pytorch-lightning/pull/11132))
+
+
 - Fixed double evaluation bug with fault-tolerance enabled where the second call was completely skipped ([#11119](https://github.com/PyTorchLightning/pytorch-lightning/pull/11119))
 
 ## [1.5.6] - 2021-12-15
--- a/pytorch_lightning/loops/dataloader/evaluation_loop.py
+++ b/pytorch_lightning/loops/dataloader/evaluation_loop.py
@@ -147,10 +147,9 @@ class EvaluationLoop(DataLoaderLoop):
 
         logged_outputs, self._logged_outputs = self._logged_outputs, []  # free memory
         # include any logged outputs on epoch_end
-        if self.num_dataloaders < 2:  # TODO: remove this check
-            epoch_end_logged_outputs = self.trainer.logger_connector.update_eval_epoch_metrics()
-            for dl_outputs in logged_outputs:
-                dl_outputs.update(epoch_end_logged_outputs)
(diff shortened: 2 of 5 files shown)


tchaton (approved):
LGTM !

rohitgr7 (approved):

awaelchli (approved):

PR #11144, opened 2021-12-18
Move cuda related benchmark code from accelerator_connector to GPU accelerator

### LINKED ISSUE #10536: Move cuda related benchmark code from accelerator connector to GPU accelerator

Move this line of code to GPU accelerator
https://github.com/PyTorchLightning/pytorch-lightning/blob/7a9a08c5d3ca4699fb439f691c40e1320b37507a/pytorch_lightning/trainer/connectors/accelerator_connector.py#L183-L184

Alternatively, can we deprecate the `benchmark` flag as mentioned by @justusschock in #9006?

### Motivation

Since this line of code is specific to GPU accelerator, it should live in that class, not in the generic accelerator_connector.

daniellepintz:
## What does this PR do?
Fixes #10536

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -30,6 +30,9 @@ _log = logging.getLogger(__name__)
 class GPUAccelerator(Accelerator):
     """Accelerator for GPU devices."""
 
+    def __init__(self, benchmark=False):
+        self._benchmark = benchmark
+
     def setup_environment(self, root_device: torch.device) -> None:
         """
         Raises:
@@ -46,7 +49,7 @@ class GPUAccelerator(Accelerator):
         # clear cache before training
         torch.cuda.empty_cache()
         # benchmarking
-        torch.backends.cudnn.benchmark = trainer.benchmark
+        torch.backends.cudnn.benchmark = self._benchmark
 
     @staticmethod
     def set_nvidia_flags(local_rank: int) -> None:
@@ -77,6 +80,9 @@ class GPUAccelerator(Accelerator):
     def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return torch.cuda.device_count()
+
+    def teardown(self) -> None:
+        torch.backends.cudnn.benchmark = False
 
 
 def get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -45,6 +45,8 @@ class GPUAccelerator(Accelerator):
         self.set_nvidia_flags(trainer.local_rank)
         # clear cache before training
         torch.cuda.empty_cache()
+        # benchmarking
+        torch.backends.cudnn.benchmark = trainer.benchmark
 
     @staticmethod
     def set_nvidia_flags(local_rank: int) -> None:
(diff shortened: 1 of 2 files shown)


pytorch_lightning/accelerators/gpu.py
```diff
@@ -45,6 +45,8 @@ def setup(self, trainer: "pl.Trainer") -> None:
         self.set_nvidia_flags(trainer.local_rank)
         # clear cache before training
         torch.cuda.empty_cache()
+        # benchmarking
+        torch.backends.cudnn.benchmark = trainer.benchmark
```
four4fish, line 49: I'm not sure it matters, but the accelerator.setup() get called way later than accelerator_connector.init, will move to setup_envirment() make more sense?

daniellepintz: I'm not sure how much of a difference that will make, it looks like they get called one after the other: https://github.com/PyTorchLightning/pytorch-lightning/blob/9826de21625b162dfef43f30864b5a6d9efab003/pytorch_lightning/trainer/trainer.py#L1106-L1107

I'm guessing that as long as it is set before training it should be fine, but someone please correct me if I'm wrong.

pytorch_lightning/accelerators/gpu.py
```diff
@@ -30,6 +30,9 @@
 class GPUAccelerator(Accelerator):
     """Accelerator for GPU devices."""
 
+    def __init__(self, benchmark=False):
+        self._benchmark = benchmark
```
ananthsub, line 34:
```suggestion
    def __init__(self, benchmark: bool = False):
        self._benchmark: bool = benchmark
```

ananthsub (approved):

ananthsub (commented):
The test here needs to be updated. Ideally this test should be pulled out to https://github.com/PyTorchLightning/pytorch-lightning/blob/0ee78e96efa71f962f4cfafa73588e4f6a0b8bce/tests/trainer/test_trainer.py#L641-L657

Let's also update the docstring for the trainer flag to specify that this flag is enabled during the course of trainer execution, given we now reset the value in the teardown: https://github.com/PyTorchLightning/pytorch-lightning/blob/787f41eff60691fe39ebc4082d75c69ea4b40e6c/pytorch_lightning/trainer/trainer.py#L226

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -791,15 +787,14 @@ def select_accelerator(self) -> Accelerator:
             return self.distributed_backend
 
         if self.use_gpu:
-            acc_cls = GPUAccelerator
+            accelerator = GPUAccelerator(benchmark=self.benchmark)
```
four4fish, line 790: This will break user who passed in a accelerator class. The line 407-411 handles when user pass in their own accelerator.  So this function is not guaranteed to be called

four4fish (changes requested):

pytorch_lightning/trainer/connectors/accelerator_connector.py
tchaton: Yes, the benchmark value won't be set in this case.

pytorch_lightning/accelerators/gpu.py
carmocca:
```suggestion
    def __init__(self, benchmark: bool = False):
        self._benchmark = benchmark
```

Only one annotation is required 😉

pytorch_lightning/accelerators/gpu.py
```diff
@@ -76,6 +81,9 @@ def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return torch.cuda.device_count()
 
+    def teardown(self) -> None:
+        torch.backends.cudnn.benchmark = False
```
carmocca, line 85: It sounds overkill but maybe this should restore the previous value before the Trainer

PR #11146, opened 2021-12-18
Add Bagua Strategy

### LINKED ISSUE #10455: Support Bagua Training Plugin
## 🚀 Feature

Bagua is a deep learning training acceleration framework for PyTorch.
You could read more the supported features and benchmarks [here](https://tutorials.baguasys.com/).
Mnist [Example](https://github.com/BaguaSys/bagua/blob/master/examples/mnist/main.py).

We could support it by introducing a new `BaguaPlugin` for training by passing it to the `strategy` flag

```python
trainer = Trainer(accelerator="gpu", devices=4, strategy="bagua")
```

cc @borda @akihironitta

wangraying:
## What does this PR do?

Fixes #10455
Fixes https://github.com/BaguaSys/bagua/issues/304

### Suggested follow ups:

- Have Bagua provide Enums for the option strings
- Add registry values for the different configurations
- Update their `bagua.distributed.run` to `bagua.distributed.launch` to match torch

None

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # move the model to the correct device
+        self.model_to_device()
+
+        model = LightningBaguaModule(self.model)
+        self.configure_bagua_ddp(model)
+
+        # start the background communication for async algorithm
+        if self._bagua_algorithm == "async":
+            self.model.bagua_algorithm.resume(self.model)
```
awaelchli, line 92: I moved this code from pre_dispatch to here (pre_dispatch removed on master). Now, some adjustments need to be discussed here. I recommend we look at super().setup(). There we already call model_to_device(). Instead of implementing `configure_bagua_ddp`, we could just implement `configure_ddp()` maybe? 

Does Bagua want to wrap the model also for inference, or should it only be done for training?

wangraying: Yes, we can implement `configure_ddp()` instead.

Bagua can optionally wrap the model for inference, they have the same behavior.

pytorch_lightning/plugins/environments/bagua_environment.py
```diff
@@ -0,0 +1,65 @@
...
+
+import logging
+import os
+
+from pytorch_lightning.plugins.environments.cluster_environment import ClusterEnvironment
+
+log = logging.getLogger(__name__)
+
+
+class BaguaEnvironment(ClusterEnvironment):
```
justusschock, line 23: does bagua come with its own launcher or why do you need a separate cluster environment for it? Ideally we could handle it similar to how horovod does it.

I'm asking because a cluster (like slurm, kubeflow, redis etc.) usually is separated from the strategy. So if we can get those terms sorted out, it might be easier to use this in the end (ideally also with lightning creating the processes and setting the necessary environment variables when no external launcher is strictly required).

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class LightningBaguaModule(_LightningModuleWrapperBase):
+    def __init__(self, pl_module: "pl.LightningModule") -> None:
+        super().__init__(pl_module)
+        # Bagua use `bagua_module_name` to distinguish different modules
+        self._bagua_module_name = pl_module._get_name() + str(id(pl_module))
+
+    def forward(self, *inputs, **kwargs):
+        return super().forward(*inputs, **kwargs)
```
justusschock, line 39:
```suggestion
```
No need when just forwarding super call

justusschock (commented):
Already looking pretty good. Comments below for clarity

pytorch_lightning/plugins/environments/bagua_environment.py
wangraying: Yes, Bagua has its own launcher, [here](https://tutorials.baguasys.com/getting-started/#launch-job) it the doc. Some optimizations in Bagua, like [Bagua-Net](https://tutorials.baguasys.com/more-optimizations/bagua-net) and [auto-tuning](https://tutorials.baguasys.com/performance-autotuning/), need to be enabled using Bagua launcher.

By design, Bagua is compatible with pytorch native distributed environment. We can launch a Bagua job using Pytorch launcher and start training using different [distributed algorithms](https://tutorials.baguasys.com/algorithms/) in Bagua, but the advanced optimizations mentioned above is off.

I have tested this strategy under Bagua env and Lightning env, but other env like slurm, kubeflow etc. has not tested yet ( It is not easy for me to have access to such envs).

And when using Bagua launcher, it can be automatically detected as a Bagua env. So that we have similar usage as horovod?

wangraying: I am not quite clear with how horovod handle this, it seems we can just launch the job using `horovodrun`.
As for bagua, we could do this similarly, by launching the job using `baguarun` or `bagua.distributed.launch`.

justusschock: I see, I was not sure, whether we need a environment plugin here, but since you also tested it with lightning env, that should be fine :)

docs/source/accelerators/gpu.rst
```diff
@@ -278,6 +278,7 @@ Lightning allows multiple ways of training
 - DistributedDataParallel 2 (``strategy='ddp2'``) (DP in a machine, DDP across machines).
 - Horovod (``strategy='horovod'``) (multi-machine, multi-gpu, configured at runtime)
 - TPUs (``tpu_cores=8|x``) (tpu or TPU pod)
+- Bagua (``strategy='bagua'``) (multiple-gpus across many machines (python script based))
```
SeanNaren, line 281:
```suggestion
- [Bagua](https://github.com/BaguaSys/bagua) (``strategy='bagua'``) (multiple-gpus across many machines)
- TPUs (``tpu_cores=8|x``) (tpu or TPU pod)
```

I'd swap this around just so that we group the GPU DDP components together! I'd also consider adding a different description to explain why Bagua is different to say Horovod or DDP (probably something our throughput and it being advanced?)

SeanNaren (approved):
This is really really awesome. I think eventually we'll need to give a nice one-shot explanation to users to explain what to use and why later, but having the ability to swap to this from the trainer is epic :)

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,235 @@
...
+            if reduce_op.lower() in ("avg", "mean"):
+                op = BaguaReduceOp.AVG
+            elif reduce_op.lower() == "sum":
+                op = BaguaReduceOp.SUM
+            else:
+                raise ValueError(f"unrecognized `reduce_op`: {reduce_op}")
+        elif isinstance(reduce_op, ReduceOp):
+            op = from_torch_reduce_op(reduce_op)
+            if op is None:
+                raise ValueError(f"unrecognized `reduce_op`: {reduce_op}")
```
rohitgr7, line 232:
```suggestion
                raise ValueError(f"Unrecognized `reduce_op`: {reduce_op}")
        elif isinstance(reduce_op, ReduceOp):
            op = from_torch_reduce_op(reduce_op)
            if op is None:
                raise ValueError(f"Unrecognized `reduce_op`: {reduce_op}")
```
Can we add better error msg? something like `BaguaStrategy doesn't support reduce_op: {reduce_op}`, just to let users know where the error is actually coming from.

tests/strategies/test_bagua_strategy.py
```diff
@@ -0,0 +1,40 @@
...
+from tests.helpers.runif import RunIf
+
+
+class TestModel(BoringModel):
+    def __init__(self):
+        super().__init__()
+        self.layer = torch.nn.Linear(32, 2)
+
+
+@RunIf(skip_windows=True, bagua=True, min_gpus=2, standalone=True)
```
rohitgr7, line 28: is it not supported on windows?

tests/strategies/test_bagua_strategy.py
```diff
@@ -0,0 +1,40 @@
...
+def test_bagua_algorithm():
+    model = TestModel()
+    bagua_strategy = BaguaStrategy(algorithm="gradient_allreduce")
+    trainer = Trainer(
+        max_epochs=1,
+        strategy=bagua_strategy,
+        gpus=2,
+    )
```
rohitgr7, line 36:
```suggestion
def test_bagua_algorithm(tmpdir):
    model = TestModel()
    bagua_strategy = BaguaStrategy(algorithm="gradient_allreduce")
    trainer = Trainer(
        default_root_dir=tmpdir,
        fast_dev_run=1,
        strategy=bagua_strategy,
        accelerator='gpu',
        devices=2,
    )
```
increase the value of `fast_dev_run` if it needs more optimization steps for the test to pass.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,235 @@
...
+        Arguments:
+            algorithm:  Distributed algorithm used to do the actual communication and update. Built-in algorithms
+                include "gradient_allreduce", "bytegrad", "decentralized", "low_precision_decentralized", "qadam" and
+                "async".
+            do_flatten: Whether to flatten the Bagua communication buckets. The flatten operation will reset data
+                pointer of bucket tensors so that they can use faster code paths. Default: ``True``.
+            kwargs: Additional arguments that will be passed to initialize the Bagua algorithm.
+        """
+
+        super().__init__(
```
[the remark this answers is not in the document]
rohitgr7: also a small unit test for it.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,235 @@
...
+            precision_plugin=precision_plugin,
+        )
+
+        self._bagua_algorithm = algorithm
+        self._bagua_do_flatten = do_flatten
+        self._bagua_kwargs = kwargs
+
+    def setup_environment(self) -> None:
+        # start the other scripts
+        if not self.cluster_environment.creates_processes_externally:  # type: ignore[union-attr]
```
SeanNaren, line 105: n00b question, this is only called when we do not use the bagua launcher right?

tests/strategies/test_bagua_strategy.py
wangraying: Yes, only on Linux now, and not support CPU as well.

rohitgr7: can you add an error for it? if the platform is windows and someone tries to use `BaguaStrategy`... or does it fail during bagua installation itself?

pytorch_lightning/strategies/bagua.py
wangraying: Bagua launcher will not call this. It seems other env like `TorchElasticEnvironment`, `KubeflowEnvironment`  and `LSFEnvironment` will not call it either, and I have not tested for those. Have tested for `LightningEnvironment`, this will be called with a condition.

tests/strategies/test_bagua_strategy.py
wangraying: It will fail on bagua installation.

docs/source/api_references.rst
```diff
@@ -43,6 +43,7 @@ Strategy API
     :nosignatures:
     :template: classtemplate.rst
 
+    BaguaStrategy
```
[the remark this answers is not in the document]
wangraying: updated

pytorch_lightning/strategies/bagua.py
wangraying: done

pytorch_lightning/strategies/bagua.py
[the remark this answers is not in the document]
wangraying: done

wangraying:
> This is really really awesome. I think eventually we'll need to give a nice one-shot explanation to users to explain what to use and why later, but having the ability to swap to this from the trainer is epic :)

Would `Bagua (strategy='bagua') (multiple-gpus across many machines with advanced training algorithms)` be OK?

wangraying:
> This is really really awesome. I think eventually we'll need to give a nice one-shot explanation to users to explain what to use and why later, but having the ability to swap to this from the trainer is epic :)

Thanks SeanNaren, looking forward to this being done :)

docs/source/accelerators/gpu.rst
wangraying: Would Bagua (strategy='bagua') (multiple-gpus across many machines with advanced training algorithms) be OK?

docs/source/accelerators/gpu.rst
```diff
@@ -484,6 +485,129 @@ number of worker processes:
...
+        devices=4,
+    )
+
+    # train on 4 GPUs, using Asynchronous Model Average algorithm, with a synchronization interval of 100ms
+    trainer = Trainer(
+        max_epochs=1,
+        strategy=BaguaStrategy(algorithm="async", sync_interval_ms=100),
+        accelerator="gpu",
+        devices=4,
+    )
```
carmocca, line 548: I would suggest instead of this block, just having some bullet points describing the algorithm options and their differences, preferably with a link to more detailed explanations

docs/source/accelerators/gpu.rst
```diff
@@ -484,6 +485,129 @@ number of worker processes:
...
+
+To use *QAdam*, we need to initialize
+`QAdamOptimizer <https://bagua.readthedocs.io/en/latest/autoapi/bagua/torch_api/algorithms/q_adam/index.html#bagua.torch_api.algorithms.q_adam.QAdamOptimizer>`_ first:
+
+.. code-block:: python
+
+    import pytorch_lightning as pl
+    from pytorch_lightning import Trainer
```
carmocca, line 556:
```suggestion
```

I believe these are not required.

docs/source/accelerators/gpu.rst
```diff
@@ -484,6 +485,129 @@ number of worker processes:
...
+    trainer = Trainer(
+        accelerator="gpu",
+        devices=4,
+        strategy=BaguaStrategy(algorithm="qadam"),
+    )
+    trainer.fit(model)
+
+It is recommend to start training through `Bagua built-in launch utilities <https://tutorials.baguasys.com/getting-started/#launch-job>`_.
+However, we can also use the same way as Distributed Data Parallel. Below are examples using ``bagua.distributed.launch``,
+whose usage is similar to ``torch.distributed.launch``:
```
carmocca, line 579: Why would the user ever choose these over their respective Trainer arguments?
Why is it recommended?

docs/source/accelerators/gpu.rst
```diff
@@ -484,6 +485,129 @@ number of worker processes:
...
+
+    model = MyModel()
+    trainer = Trainer(
+        accelerator="gpu",
+        devices=4,
+        strategy=BaguaStrategy(algorithm="qadam"),
+    )
+    trainer.fit(model)
+
+It is recommend to start training through `Bagua built-in launch utilities <https://tutorials.baguasys.com/getting-started/#launch-job>`_.
```
carmocca, line 577:
```suggestion
It is recommended to start training through `Bagua built-in launch utilities <https://tutorials.baguasys.com/getting-started/#launch-job>`_.
```

tests/strategies/test_bagua_strategy.py
```diff
@@ -0,0 +1,51 @@
...
+
+    for param in model.parameters():
+        assert torch.norm(param) < 3
+
+
+@RunIf(bagua=False, min_gpus=1)
+def test_bagua_not_available():
+
+    with pytest.raises(MisconfigurationException, match="you must have `Bagua` installed"):
+        Trainer(strategy="bagua", gpus=1)
```
carmocca, line 51: That test will never run because our only job with GPUs installs bagua, you need to do:

```python
@mock.patch("torch.cuda.device_count", return_value=1)
def test_bagua_not_available(monkeypatch, _):
    import pytorch_lightning.strategies.bagua as imports

    monkeypatch.setattr(imports, "_BAGUA_AVAILABLE", False)
    with pytest.raises(MisconfigurationException, match="you must have `Bagua` installed"):
        Trainer(strategy="bagua", gpus=1)
```

tests/strategies/test_bagua_strategy.py
```diff
@@ -0,0 +1,51 @@
...
+class TestModel(BoringModel):
+    def __init__(self):
+        super().__init__()
+        self.layer = torch.nn.Linear(32, 2)
+
+
+@RunIf(skip_windows=True, bagua=True, min_gpus=2, standalone=True)
+def test_bagua_algorithm(tmpdir):
+    model = TestModel()
+    bagua_strategy = BaguaStrategy(algorithm="gradient_allreduce")
```
carmocca, line 33: Can we add a test parametrizing all algorithms and testing that bagua is instantiated as expected? It does not need to `fit`. You could mock `BaguaDistributedDataParallel`

wangraying: Yes, We tested for all algorithms previously, but it turns out  it is quite time consuming. After discussed with @tchaton , we decided to move other tests to Bagua side and only keep a simple gut check in lightning.

docs/source/accelerators/gpu.rst
wangraying: It seems we used in the definition of `MyModel(pl.LightningModule)` and declaration of  `Trainer`. Should we remove these two or it just looks like a bit redundant?

docs/source/accelerators/gpu.rst
wangraying: Oh, I didn't make it clear. I mean users should use Bagua launch, but if they use other launch utilities provided by lightning or torchelastic, they can start training as well.

The difference is they can only Bagua-Net by bagua launch.

docs/source/accelerators/gpu.rst
wangraying: I added a brief explanation for each algorithm, does it look all right?

docs/source/accelerators/gpu.rst
wangraying: I moved `we can also use the same way as Distributed Data Parallel` this part to ``note``, does it look better?

tests/strategies/test_bagua_strategy.py
wangraying: done

tests/strategies/test_bagua_strategy.py
wangraying: Yes. Bagua has this kind of validation, it will throw an exception on invalid inputs.

pytorch_lightning/plugins/environments/bagua_environment.py
```diff
@@ -0,0 +1,65 @@
...
+    """Environment for distributed training with `Bagua <https://tutorials.baguasys.com/>`_"""
+
+    def __init__(self) -> None:
+        super().__init__()
+
```
ananthsub, line 28: this shouldn't be needed

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        os.environ["RANK"] = str(self.global_rank)
+        os.environ["NODE_RANK"] = str(self.node_rank)
+        os.environ["WORLD_SIZE"] = str(self.world_size)
+        os.environ["LOCAL_RANK"] = str(self.local_rank)
+
+    def _check_qadam_optimizer(self) -> None:
+
+        trainer = self.lightning_module.trainer
+        has_qadam_optimizer = any([isinstance(opt, QAdamOptimizer) for opt in trainer.optimizers])
+
```
ananthsub, line 149: we shouldn't need this indirection now that the Strategy class owns the optimizers

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        if not has_qadam_optimizer or len(trainer.optimizers) > 1 or len(trainer.lr_schedulers) > 1:
+            raise MisconfigurationException("Bagua QAdam can only accept one QAdamOptimizer and one LR Scheduler.")
+
+    def configure_ddp(self) -> None:
+        model = LightningBaguaModule(self.model)  # type: ignore[arg-type]
+        self._model = self._setup_model(model)
```
ananthsub, line 155: should this always be setup, even in the case of validation/testing/prediction where there's no gradient comms needed? 

see this handling with distributed data parallel: https://github.com/PyTorchLightning/pytorch-lightning/blob/f7f835fa0e22dc34631aa43d54bf73304c9bda22/pytorch_lightning/strategies/ddp.py#L168-L171

docs/source/accelerators/gpu.rst
```diff
@@ -484,6 +485,129 @@ number of worker processes:
...
+    trainer = Trainer(
+        max_epochs=1,
+        strategy=BaguaStrategy(algorithm="low_precision_decentralized"),
+        accelerator="gpu",
+        devices=4,
+    )
+
+    # train on 4 GPUs, using Asynchronous Model Average algorithm, with a synchronization interval of 100ms
+    trainer = Trainer(
+        max_epochs=1,
```
carmocca, line 544:
```suggestion
```

docs/source/accelerators/gpu.rst
```diff
@@ -484,6 +485,129 @@ number of worker processes:
...
+    # train on 4 GPUs, using Bagua Decentralized SGD
+    trainer = Trainer(
+        strategy=BaguaStrategy(algorithm="decentralized"),
+        accelerator="gpu",
+        devices=4,
+    )
+
+    # train on 4 GPUs, using Bagua Low Precision Decentralized SGD
+    trainer = Trainer(
+        max_epochs=1,
```
carmocca, line 536:
```suggestion
```

docs/source/accelerators/gpu.rst
carmocca: I would remove them so it showcases the notable imports (Bagua's) more

tests/strategies/test_bagua_strategy.py
carmocca: I am referring to tests that check that we pass the expected information to Bagua.

pytorch_lightning/strategies/bagua.py
wangraying: No, this should not always be setup. Bagua follows similar pattern with Torch DDP. But it is ok for us to call `configure_ddp` under validation/testing/prediction, in this case no communications will be initiated (can only be initiated under training mode).

tests/strategies/test_bagua_strategy.py
wangraying: I added a few tests to check this.

pytorch_lightning/plugins/environments/bagua_environment.py
wangraying: removed

pytorch_lightning/strategies/bagua.py
wangraying: done

tests/strategies/test_bagua_strategy.py
```diff
@@ -0,0 +1,55 @@
...
+from pytorch_lightning.strategies import BaguaStrategy
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from tests.helpers.boring_model import BoringModel
+from tests.helpers.runif import RunIf
+
+
+class TestModel(BoringModel):
+    def __init__(self):
+        super().__init__()
+        self.layer = torch.nn.Linear(32, 2)
```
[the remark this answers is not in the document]
wangraying: done

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+    def __init__(self, pl_module: "pl.LightningModule") -> None:
+        super().__init__(pl_module)
+        # Bagua use `bagua_module_name` to distinguish different modules
+        self._bagua_module_name = pl_module._get_name() + str(id(pl_module))
+
+
+_bagua_reduce_ops: Dict[ReduceOp, BaguaReduceOp] = {}
+
+
+def from_torch_reduce_op(op: ReduceOp) -> Optional[BaguaReduceOp]:
```
rohitgr7, line 45:
```suggestion
def _from_torch_reduce_op(op: ReduceOp) -> Optional[BaguaReduceOp]:
```

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+            op = BaguaReduceOp.AVG
+        elif isinstance(reduce_op, str):
+            if reduce_op.lower() in ("avg", "mean"):
+                op = BaguaReduceOp.AVG
+            elif reduce_op.lower() == "sum":
+                op = BaguaReduceOp.SUM
+            else:
+                raise ValueError(f"Unrecognized `reduce_op` for `BaguaStrategy`: {reduce_op}")
+        elif isinstance(reduce_op, ReduceOp):
+            op = from_torch_reduce_op(reduce_op)
```
rohitgr7, line 233:
```suggestion
            op = _from_torch_reduce_op(reduce_op)
```

rohitgr7 (approved):

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -703,6 +710,8 @@ def select_strategy(self) -> Strategy:
             plugin = DeepSpeedStrategy(
                 cluster_environment=self.select_cluster_environment(), parallel_devices=self.parallel_devices
             )
+        elif self.use_ddp and self.use_bagua:
```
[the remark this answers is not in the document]
wangraying: I followed the pattern with `DeepSpeedStrategy`, it seems`DeepSpeedStrategy` and `BaguaStrategy` both inherit from `DDPStrategy`.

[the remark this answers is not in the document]
rohitgr7: I think just `use_bagua` should work fine because `self._strategy_type` can have only one value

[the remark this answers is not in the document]
rohitgr7: but within lightning it will still fall under `self._strategy_type='bagua'` and user should get an error. But nvm, I think it will be resolved soon later after an ongoing refactor.

carmocca:
You should add the newly added files to https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CODEOWNERS and include yourself

wangraying:
> You should add the newly added files to https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CODEOWNERS and include yourself

Yep. I saw there are codeowners for the directories which newly added files reside in. Should I include myself in it (and hmm... I'm from a third-party team)? :)

pytorch_lightning/plugins/environments/bagua_environment.py
```diff
@@ -0,0 +1,62 @@
...
+    def main_port(self) -> int:
+        return int(os.environ.get("MASTER_PORT", -1))
+
+    @property
+    def service_port(self) -> int:
+        return int(os.environ.get("BAGUA_SERVICE_PORT", -1))
+
+    @staticmethod
+    def detect() -> bool:
+        return "BAGUA_SERVICE_PORT" in os.environ
```
tchaton, line 44: This would work only with your own launcher right ?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+    def __init__(
+        self,
+        algorithm: str = "gradient_allreduce",
+        do_flatten: bool = True,
+        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
+        parallel_devices: Optional[List[torch.device]] = None,
+        cluster_environment: Optional[ClusterEnvironment] = None,
+        checkpoint_io: Optional[CheckpointIO] = None,
+        precision_plugin: Optional[PrecisionPlugin] = None,
+        **kwargs: Union[Any, Dict[str, Any]],
```
tchaton, line 77:
```suggestion
        **bagua_kwargs: Union[Any, Dict[str, Any]],
```

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+        """Strategy for training using the `Bagua <https://github.com/BaguaSys/bagua>`_ library, with advanced
+        distributed training algorithms and system optimizations.
+
+        Arguments:
+            algorithm:  Distributed algorithm used to do the actual communication and update. Built-in algorithms
+                include "gradient_allreduce", "bytegrad", "decentralized", "low_precision_decentralized", "qadam" and
+                "async".
+            do_flatten: Whether to flatten the Bagua communication buckets. The flatten operation will reset data
+                pointer of bucket tensors so that they can use faster code paths.
+            kwargs: Additional arguments that will be passed to initialize the Bagua algorithm.
```
tchaton, line 88:
```suggestion
            bagua_kwargs: Additional keyword arguments that will be passed to initialize the Bagua algorithm.
```

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+        # determine which process we are and world size
+        self.set_world_ranks()
+
+        self._init_bagua_distributed()
+
+    def _init_bagua_distributed(self) -> None:
+
+        self._set_node_environment_variables()
+        log.info(
+            "initializing Bagua distributed: "
```
tchaton, line 126:
```suggestion
            "Initializing Bagua Distributed: "
```

pytorch_lightning/strategies/bagua.py
tchaton: Yes, this shouldn't be called during validation and testing as it would be an overhead.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+
+        if not has_qadam_optimizer or len(self.optimizers) > 1 or len(self.lr_schedulers) > 1:
+            raise MisconfigurationException("Bagua QAdam can only accept one QAdamOptimizer and one LR Scheduler.")
+
+    def configure_ddp(self) -> None:
+        model = LightningBaguaModule(self.model)  # type: ignore[arg-type]
+        self._model = self._setup_model(model)
+
+        # start the background communication for async algorithm
+        if self.lightning_module.trainer.training and self._bagua_algorithm == "async":
```
tchaton, line 157: Can Bagua provides an enum there ?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+            self.model.bagua_algorithm.resume(self.model)  # type: ignore
+
+    def _setup_model(self, model: Module) -> BaguaDistributedDataParallel:
+        """Wraps the model into a Bagua distributed module."""
+
+        if self._bagua_algorithm == "qadam":
```
tchaton, line 163: Same here ?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+            self._check_qadam_optimizer()
+            self._bagua_kwargs["q_adam_optimizer"] = self.optimizers[0]
+
+        algorithm = Algorithm.init(self._bagua_algorithm, **self._bagua_kwargs)
```
tchaton, line 167: Neat !

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,238 @@
...
+    @property
+    def lightning_module(self) -> "pl.LightningModule":
+        model = self._model
+        if isinstance(model, BaguaDistributedDataParallel):
+            model = model.module
+        return unwrap_lightning_module(model)  # type: ignore[arg-type]
+
+    @classmethod
+    def register_plugins(cls, plugin_registry: Dict) -> None:
+        plugin_registry.register("bagua", cls, description="Default Bagua Plugin")
```
tchaton, line 184: I would register "bagua_async', etc ...

tchaton (commented):
Great work !

wangraying:
can't find the link, I added a short explaination for `baguarun` @tchaton

tchaton:
Hey @wangraying,

You have some failing test with master.

Screenshot 2022-01-21 at 08 07 24

wangraying:
> Hey @wangraying,
> 
> You have some failing test with master.
> 
> Screenshot 2022-01-21 at 08 07 24

done. seems caused by previous refactor on removing this property.

docs/source/accelerators/gpu.rst
```diff
@@ -489,6 +490,127 @@ number of worker processes:
...
+.. code-block:: bash
+
+    # start training with 8 GPUs on a single node
+    python -m bagua.distributed.launch --nproc_per_node=8 train.py
+
+    # Run on node1 to start training on two nodes (node1 and node2), 8 GPUs per node
+    python -m bagua.distributed.launch --nproc_per_node=8 --nnodes=2 --node_rank=0 --master_addr=hostname1 --master_port=port1 train.py
+
+    # Run on node2 to start training on two nodes (node1 and node2), 8 GPUs per node
+    python -m bagua.distributed.launch --nproc_per_node=8 --nnodes=2 --node_rank=1 --master_addr=hostname1 --master_port=port1 train.py
```
carmocca, line 590: If the launcher link above already describes these commands, what's the benefit of adding them here too?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+    BaguaDistributedDataParallel = None
+
+log = logging.getLogger(__name__)
+
+
+class LightningBaguaModule(_LightningModuleWrapperBase):
+    def __init__(self, pl_module: "pl.LightningModule") -> None:
+        super().__init__(pl_module)
+        # Bagua use `bagua_module_name` to distinguish different modules
+        self._bagua_module_name = pl_module._get_name() + str(id(pl_module))
```
carmocca, line 39: Is there a benefit to using `_get_name`? It's protected so better if we avoid it.

```suggestion
        self._bagua_module_name = f"{pl_module.__class__.__name__}{id(pl_module)}"
```

tests/strategies/test_bagua_strategy.py
```diff
@@ -0,0 +1,121 @@
...
+class BoringModel4QAdam(BoringModel):
+    def configure_optimizers(self):
+        from bagua.torch_api.algorithms.q_adam import QAdamOptimizer
+
+        optimizer = QAdamOptimizer(self.layer.parameters(), lr=0.05, warmup_steps=20)
+        lr_scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=1)
+        return [optimizer], [lr_scheduler]
+
+
+@RunIf(skip_windows=True, bagua=True, min_gpus=1)
```
carmocca, line 36: Does Bagua support MacOs?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        self._bagua_kwargs = bagua_kwargs
+
+    def setup_environment(self) -> None:
+        # start the other scripts
+        if not self.cluster_environment.creates_processes_externally:  # type: ignore[union-attr]
+            self._call_children_scripts()
+
+        self.setup_distributed()
+
+    def setup_distributed(self) -> None:
```
carmocca, line 114: Side note: Needing to override a hook like this showcases paper-cuts in the current DDP design

cc @awaelchli

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+
+        self._set_node_environment_variables()
+        log.info(
+            "Initializing Bagua Distributed: "
+            f"GLOBAL_RANK: {self.global_rank}, "
+            f"MEMBER: {self.global_rank + 1}/{self.world_size}"
+        )
+
+        # need to set device first before initialize Bagua distributed environment
+        torch.cuda.set_device(self.local_rank)
```
carmocca, line 132: Doesn't the `GPUAccelerator` take care of this?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        algorithm: str = "gradient_allreduce",
+        do_flatten: bool = True,
+        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
+        parallel_devices: Optional[List[torch.device]] = None,
+        cluster_environment: Optional[ClusterEnvironment] = None,
+        checkpoint_io: Optional[CheckpointIO] = None,
+        precision_plugin: Optional[PrecisionPlugin] = None,
+        **bagua_kwargs: Union[Any, Dict[str, Any]],
+    ):
+        """Strategy for training using the `Bagua <https://github.com/BaguaSys/bagua>`_ library, with advanced
```
carmocca, line 79: Can Bagua run on CPU at all? For debugging for example.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        # start the background communication for async algorithm
+        if self.lightning_module.trainer.training and self._bagua_algorithm == "async":
+            self.model.bagua_algorithm.resume(self.model)  # type: ignore
+
+    def _setup_model(self, model: Module) -> BaguaDistributedDataParallel:
+        """Wraps the model into a Bagua distributed module."""
+
+        if self._bagua_algorithm == "qadam":
+            self._check_qadam_optimizer()
+            self._bagua_kwargs["q_adam_optimizer"] = self.optimizers[0]
```
carmocca, line 165: Can you move this block inside `_check_qadam_optimizer`? To group everything together

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+
+    @classmethod
+    def register_plugins(cls, plugin_registry: Dict) -> None:
+        plugin_registry.register("bagua", cls, description="Default Bagua Plugin")
+
+    def teardown(self) -> None:
+        # abort the background communication for async algorithm
+        if self.lightning_module.trainer.training and self._bagua_algorithm == "async":
+            self.model.bagua_algorithm.abort(self.model)  # type: ignore
+
```
carmocca, line 190: Shouldn't this also do the unwrapping of `DistributedDataParallel` as in the parent?

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        if group is not None:
+            raise ValueError("`Bagua` does not support allreduce using a subcommunicator at this time. Unset `group`.")
+
+        if reduce_op is None:
+            op = BaguaReduceOp.AVG
+        elif isinstance(reduce_op, str):
+            if reduce_op.lower() in ("avg", "mean"):
+                op = BaguaReduceOp.AVG
+            elif reduce_op.lower() == "sum":
+                op = BaguaReduceOp.SUM
```
carmocca, line 230: Shouldn't this use the mapping you created at the beginning of this file?

pytorch_lightning/plugins/environments/bagua_environment.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def world_size(self) -> int:
+        return int(os.environ["WORLD_SIZE"])
+
+    def set_world_size(self, size: int) -> None:
+        log.debug("`BaguaEnvironment.set_world_size` was called, but setting world size is not allowed. Ignored.")
+
+    def global_rank(self) -> int:
+        return int(os.environ["RANK"])
```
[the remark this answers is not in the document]
awaelchli: I would say this is fine given that there is a contract that the variable is always present when bagua is used. And if I understand correctly, that will always be true.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+            {
+                ReduceOp.SUM: BaguaReduceOp.SUM,
+                ReduceOp.PRODUCT: BaguaReduceOp.PRODUCT,
+                ReduceOp.MIN: BaguaReduceOp.MIN,
+                ReduceOp.MAX: BaguaReduceOp.MAX,
+                ReduceOp.BAND: BaguaReduceOp.BAND,
+                ReduceOp.BOR: BaguaReduceOp.BOR,
+                ReduceOp.BXOR: BaguaReduceOp.BXOR,
+            }
+        )
```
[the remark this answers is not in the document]
awaelchli: agreed, can be a constant :)

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        """Strategy for training using the `Bagua <https://github.com/BaguaSys/bagua>`_ library, with advanced
+        distributed training algorithms and system optimizations.
+
+        Arguments:
+            algorithm:  Distributed algorithm used to do the actual communication and update. Built-in algorithms
+                include "gradient_allreduce", "bytegrad", "decentralized", "low_precision_decentralized", "qadam" and
+                "async".
+            do_flatten: Whether to flatten the Bagua communication buckets. The flatten operation will reset data
+                pointer of bucket tensors so that they can use faster code paths.
+            bagua_kwargs: Additional keyword arguments that will be passed to initialize the Bagua algorithm.
```
awaelchli, line 88: potential confusion for the naming here.
do_flatten is a parameter for the BaguaDistributedParallel module, but bagua_kwargs contains args for the Algorithm. The documentation is clear though.

Other note, in Lightning we usually don't prefix arg names to indicate they are bool, i.e., here it would just be "flatten". I'm fine with either.

pytorch_lightning/strategies/bagua.py
awaelchli: this is what `setup_environment` would do on the accelerator, but the above override of `setup_environment` excluded it.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+        if not is_initialized():
+            bagua.init_process_group()
+
+    def _set_node_environment_variables(self) -> None:
+        os.environ["MASTER_ADDR"] = self.cluster_environment.main_address  # type: ignore[union-attr]
+        os.environ["MASTER_PORT"] = str(self.cluster_environment.main_port)  # type: ignore[union-attr]
+        os.environ["RANK"] = str(self.global_rank)
+        os.environ["NODE_RANK"] = str(self.node_rank)
+        os.environ["WORLD_SIZE"] = str(self.world_size)
+        os.environ["LOCAL_RANK"] = str(self.local_rank)
```
awaelchli, line 150: why is this necessary?` self.cluster_environment.main_address` already is equal to  os.environ["MASTER_ADDR"]. The same for the rest of these variables. 

The call to `_set_node_environment_variables` can be entirely omitted here. Let's remove it?

awaelchli (approved):

pytorch_lightning/strategies/bagua.py
awaelchli: Discussed on a call with @wangraying and added a docstring explaining why it was necessary to do this. Looks great!

pytorch_lightning/strategies/bagua.py
awaelchli: Discussed with @wangraying 
No, it does not support GPU. Added a sentence in the docstring.

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,239 @@
...
+            cluster_environment=cluster_environment,
+            checkpoint_io=checkpoint_io,
+            precision_plugin=precision_plugin,
+        )
+
+        self._bagua_algorithm = algorithm
+        self._bagua_do_flatten = do_flatten
+        self._bagua_kwargs = bagua_kwargs
+
+    def setup_environment(self) -> None:
```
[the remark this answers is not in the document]
awaelchli: Discussed with @wangraying 
We can remove this and let super() handle it.

tests/strategies/test_bagua_strategy.py
awaelchli: no it does not. 
it seems the bagua package is also not installable on macos so the guard bagua=True here seems to be enough, but we could also add a skip condition to the test here. As you wish.

docs/source/accelerators/gpu.rst
awaelchli: Fine for me to leave it like this. Alternatively we could also shorten it, just to show that you can also run with torch.distributed.launch. 

BTW, torch.distributed.launch is deprecated in favor of torch.distributed.run, should we update it here? @wangraying

awaelchli (approved):
Awesome!

tchaton (approved):
LGTM !

pytorch_lightning/strategies/bagua.py
carmocca: I went through the codebase and there aren't any. Resolving

carmocca (approved):

docs/source/accelerators/gpu.rst
carmocca: I think this suggestion should be an issue in their repo, we can update this here when they make the change

awaelchli:
🎉 🎉 🎉 🎉 🎉

tchaton:
Congrats ! Awesome work !

pytorch_lightning/strategies/bagua.py
```diff
@@ -0,0 +1,235 @@
...
+        algorithm = Algorithm.init(self._bagua_algorithm, **self._bagua_kwargs)
+        return BaguaDistributedDataParallel(
+            module=model,
+            optimizers=self.optimizers,
+            algorithm=algorithm,
+            gradient_as_bucket_view=self._bagua_flatten,
+        )
+
+    @classmethod
+    def register_plugins(cls, plugin_registry: Dict) -> None:
```
four4fish, line 183: This method name has to be register_strategies, otherwise won't be called 
https://github.com/PyTorchLightning/pytorch-lightning/blob/1203094a201bd38f0b8b77d93bc39fc95f06d8ae/pytorch_lightning/strategies/strategy_registry.py#L137

There is no test covers strategy="bagua", so this issue didn't get caught
I will fix it in #11448, or feel free to open a seperate PR to fix this

PR #1279, opened 2020-03-29
Remove default optimizer, add None optimizer option

ethanwharris:
# Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure to update the docs?   
- [x] Did you write any new necessary tests?  
- [x] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CHANGELOG.md)?

## What does this PR do?
- no longer automatically choose an optimizer
- will run without an optimizer and warn the user if `configure_optimizers` is not overriden or returns `None`
- refactored default optimizer to the init_optimizers method
- refactored `init_optimizers` and `configure_schedulers` to their own mixin in seperate file (optimizers.py)
- changed `init_optimizers` to just take model so that `configure_optimizers` is only called in one place

## PR review    
Anyone in the community is free to review the PR once the tests have passed.     
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #11155, opened 2021-12-18
Deprecate `TrainerOptimizersMixin` and move functionality to `core/optimizer.py`

### LINKED ISSUE #10681: Move code from `TrainerOptimizersMixin` into TrainingTypePlugin
## Proposed refactor
Move all the code from this class to TTP:
https://github.com/PyTorchLightning/pytorch-lightning/blob/2036dfb5df2a48056d1c043ee311aed793187070/pytorch_lightning/trainer/optimizers.py#L28

### Motivation

After #10596 the TTP will own the optimizer related logic previously owned by Accelerator. It makes sense for the TTP to also own the optimizer related code in `TrainerOptimizersMixin`. This would allow us to avoid this type of circuitous logic: https://github.com/PyTorchLightning/pytorch-lightning/blob/dcafc95f2b0fd3f176d425139ca99676ce943a12/pytorch_lightning/plugins/training_type/training_type_plugin.py#L244-L245

We in general want to reduce the number of mixins (#10417)

cc @justusschock @awaelchli @akihironitta

daniellepintz:
## What does this PR do?

Fixes #10681

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/core/lightning.py
```diff
@@ -1969,3 +1969,183 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
 
         self._register_state_dict_hook(state_dict_hook)
         self._register_load_state_dict_pre_hook(pre_load_state_dict_hook, True)
+
+    def init_optimizers_and_lr_schedulers(self) -> Tuple[List, List, List]:
```
rohitgr7, line 1973: should this be private?

also quick ques: why lightningModule is handling the parsing/validation logic?

daniellepintz: > should this be private?

Since it is used by other classes across our codebase I initially left it as public but since we don't want users to access it I have now made it private. This is a problem we have with many private methods in the codebase, where they are not truly private...

> why lightningModule is handling the parsing/validation logic?

I agree with @awaelchli 's reasoning in https://github.com/PyTorchLightning/pytorch-lightning/issues/10681#issuecomment-997146642: "where it's close to the hook that retrieves the optimizers and schedulers from the user (configure_optimizers)"

But I am open to other suggestions of where to put this logic!

pytorch_lightning/trainer/optimizers.py
```diff
@@ -1,250 +0,0 @@
...
-from torch import optim
-from torch.optim.optimizer import Optimizer
-
-import pytorch_lightning as pl
-from pytorch_lightning.core.optimizer import LightningOptimizer
-from pytorch_lightning.utilities import rank_zero_warn
-from pytorch_lightning.utilities.exceptions import MisconfigurationException
-
-
-class TrainerOptimizersMixin(ABC):
```
ananthsub, line 28: this can't be removed straight away. The Trainer inherits this, so it's part of the public trainer API. Therefore `TrainerOptimizersMixin.init_optimizers` needs to be marked as deprecated before we can remove it.

pytorch_lightning/core/lightning.py
```diff
@@ -1969,3 +1969,188 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
 
         self._register_state_dict_hook(state_dict_hook)
         self._register_load_state_dict_pre_hook(pre_load_state_dict_hook, True)
+
+    def _init_optimizers_and_lr_schedulers(self) -> Tuple[List, List, List]:
```
ananthsub, line 1973: on second thought, I think @awaelchli 's suggestion of moving this to a pure utilities function/file is much better. That avoids a ton of bloat in the LightningModule API.
```
init_optimizers_and_lr_schedulers(lightning_module: LightningModule) -> ???:
   ...
```

then we can also add nice unit tests for this to cover all of the return type options

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +163,30 @@ def closure_dis():
         assert trainer is not None
         with trainer.profiler.profile(profiler_action):
             trainer.training_type_plugin.optimizer_step(self._optimizer, self._optimizer_idx, closure, **kwargs)
+
+
+class _MockOptimizer(Optimizer):
+    """The `_MockOptimizer` will be used inplace of an optimizer in the event that `None` is returned from
+    `configure_optimizers`."""
```
ananthsub, line 170: (for a another issue/PR) @ethanwharris I saw this was added in #1279 - do you recall what the use case was? 

If users don't have optimizers, should they use manual optimization instead of returning None here?

this leaks out like this which could be very confusing for a user
```py
class MyLightningModule(LightningModule):
    def configure_optimizers():
        return None

    def training_step(batch, batch_idx):
        opt = self.optimizers()
        # opt is not None! what?!
```

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -105,9 +105,7 @@ def setup_optimizers(self, trainer: "pl.Trainer") -> None:
         """
         if trainer.state.fn not in (TrainerFn.FITTING, TrainerFn.TUNING):
             return
-        optimizers, lr_schedulers, optimizer_frequencies = self.init_optimizers(
-            trainer=trainer, model=self.lightning_module
-        )
+        optimizers, lr_schedulers, optimizer_frequencies = self.lightning_module._init_optimizers_and_lr_schedulers()
         self.optimizers = optimizers
         self.lr_schedulers = lr_schedulers
         self.optimizer_frequencies = optimizer_frequencies
```
ananthsub, line 111: nit: these lines could be combined

pytorch_lightning/trainer/trainer.py
```diff
@@ -118,7 +117,6 @@
 
 class Trainer(
     TrainerCallbackHookMixin,
-    TrainerOptimizersMixin,
```
ananthsub, line 121: this is really exciting to see the mixins being whittled down like this. my hope is that v1.6, all of the mixins are marked as deprecated, and the vestiges of the OG trainer design will eventually be retired (composition over inheritance!)

pytorch_lightning/core/optimizer.py
ethanwharris: This goes back a bit haha. Not sure about the use case, someone wanted to disable the optimizer. If it's possible to use manual optimization to acheive this then I think it would be fine to remove it and document it somewhere.

pytorch_lightning/core/lightning.py
awaelchli: I would only make it a method if we expect anyone to override the parsing, but at this point I don't think this is a good idea. Function is much better for unit testing as pointed out by @ananthsub. Regardless, I am still in favor of keeping the function in core, arguing that this is where the hook and return type is defined in the first place, and fighting against the current behavior of using the pl/utilities as a dumping ground for random functions :)

pytorch_lightning/core/lightning.py
awaelchli: A collection of possible places I can come up with:
- core/lightning.py
- core/optimizer.py
- pl/utilities/parsing??.py

rohitgr7: +1 for core/optimizers, maybe as a function?

pytorch_lightning/core/lightning.py
ananthsub: I think this makes sense to add as a set of utility functions under core/optimizers then

pytorch_lightning/trainer/optimizers.py
daniellepintz: Ok, will do. May I suggest we remove it in v1.7, for a shorter deprecation cycle, since this is something that is technically part of the public trainer API, but isn't really meant to be used by users. I see it as similar to the `AcceleratorConnector` methods deprecated in https://github.com/PyTorchLightning/pytorch-lightning/pull/10101. cc @awaelchli

pytorch_lightning/core/lightning.py
daniellepintz: Moved it to `core/optimizers` as functions!

pytorch_lightning/core/lightning.py
daniellepintz: > I think this makes sense to add as a set of utility functions under core/optimizers then

Done!

pytorch_lightning/plugins/training_type/training_type_plugin.py
```diff
@@ -105,12 +106,9 @@ def setup_optimizers(self, trainer: "pl.Trainer") -> None:
         """
         if trainer.state.fn not in (TrainerFn.FITTING, TrainerFn.TUNING):
             return
-        optimizers, lr_schedulers, optimizer_frequencies = self.init_optimizers(
-            trainer=trainer, model=self.lightning_module
+        self.optimizers, self.lr_schedulers, self.optimizer_frequencies = _init_optimizers_and_lr_schedulers(
```
tchaton, line 109: Neat !

pytorch_lightning/core/optimizer.py
four4fish: @ananthsub  @daniellepintz  Can we address this in a separate PR? We have use cases for None in configure_optimizers() and they depends on the configure_optimizers() could be None to achieve their implementations.

pytorch_lightning/trainer/optimizers.py
ananthsub: Since this has been around for longer, and since the stated API evolution is 2 minor releases, I think we should remove only for v1.8

pytorch_lightning/trainer/trainer.py
```diff
@@ -2292,6 +2291,17 @@ def terminate_on_nan(self, val: bool) -> None:
         )
         self._terminate_on_nan = val  # : 212
 
+    def convert_to_lightning_optimizers(self):
+        def _convert_to_lightning_optimizer(trainer, optimizer):
```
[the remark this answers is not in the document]
awaelchli: Yes agree, something needs to change here because the same method cannot be in the "mixin" and on the Trainer, otherwise one of the two gets ignored.

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -562,11 +560,21 @@ def distributed_sampler_kwargs(self):
         distributed_sampler_kwargs = dict(num_replicas=self.world_size, rank=self.global_rank)
         return distributed_sampler_kwargs
 
-    def init_optimizers(self, trainer: "pl.Trainer", model: "pl.LightningModule") -> Tuple[List, List, List]:
+    def setup_optimizers(self, trainer: "pl.Trainer") -> None:
+        """Creates optimizers and schedulers.
```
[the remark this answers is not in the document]
awaelchli: This is equivalent to how it was before (deepspeed had init_optimizers overridden), this should be fine.

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +164,216 @@ def closure_dis():
...
         with trainer.profiler.profile(profiler_action):
             trainer.training_type_plugin.optimizer_step(self._optimizer, self._optimizer_idx, closure, **kwargs)
+
+
+def _init_optimizers_and_lr_schedulers(model: "pl.LightningModule") -> Tuple[List, List, List]:
+    prev_fx_name = model._current_fx_name
+    model._current_fx_name = "configure_optimizers"
+    model.trainer._lightning_optimizers = None
+    optim_conf = model.configure_optimizers()
+    model._current_fx_name = prev_fx_name
```
awaelchli, line 174: this part looks like it was changed

awaelchli (approved):
FYI, we started thinking about typing the config return types in  #10802

pytorch_lightning/core/optimizer.py
ananthsub: @four4fish yes, I filed https://github.com/PyTorchLightning/pytorch-lightning/issues/11178 to discuss further.

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
daniellepintz: > Yes agree, something needs to change here because the same method cannot be in the "mixin" and on the Trainer, otherwise one of the two gets ignored.

Good point. Alternatively we can just mark this method as private. It might make sense for it to be in the trainer since it is modifying state on the trainer. But I don't feel too strongly either way

pytorch_lightning/core/optimizer.py
daniellepintz: Yeah, I can't remember why I thought that was necessary, but putting it back now to how it was before

pytorch_lightning/trainer/trainer.py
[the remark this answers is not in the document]
daniellepintz: > Yes agree, something needs to change here because the same method cannot be in the "mixin" and on the Trainer, otherwise one of the two gets ignored.

Good point. Alternatively we can just mark this method as private. It might make sense for it to be in the trainer since it is modifying state on the trainer. But I don't feel too strongly either way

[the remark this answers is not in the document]
awaelchli: sounds good. and would you move the inner function to the optimizer.py as well as @tchaton suggested?

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +164,220 @@ def closure_dis():
         assert trainer is not None
         with trainer.profiler.profile(profiler_action):
             trainer.training_type_plugin.optimizer_step(self._optimizer, self._optimizer_idx, closure, **kwargs)
+
+
+def _init_optimizers_and_lr_schedulers(model: "pl.LightningModule") -> Tuple[List, List, List]:
```
ananthsub, line 169:
```suggestion
def init_optimizers_and_lr_schedulers(model: "pl.LightningModule") -> Tuple[List, List, List]:
```

let's add a short docstring here as well

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +164,220 @@ def closure_dis():
...
+
+
+def _configure_schedulers(
+    schedulers: list, monitor: Optional[str], is_manual_optimization: bool
+) -> List[Dict[str, Any]]:
+    """Convert each scheduler into dict structure with relevant information."""
+    lr_schedulers = []
+    default_config = _get_default_scheduler_config()
+    for scheduler in schedulers:
+        if is_manual_optimization:
```
ananthsub, line 256: nit: `is_manual_optimization` is constant across all schedulers. there's no point in checking it once per scheduler in the loop like this. this could be a todo to speed this up in a separate PR later.

pytorch_lightning/trainer/optimizers.py
```diff
@@ -21,15 +21,26 @@
...
 from pytorch_lightning.core.optimizer import LightningOptimizer
-from pytorch_lightning.utilities import rank_zero_warn
+from pytorch_lightning.utilities import rank_zero_deprecation, rank_zero_warn
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 
 
 class TrainerOptimizersMixin(ABC):
+    r"""
+    .. deprecated:: v1.6
+        The `TrainerOptimizersMixin` was deprecated in v1.6 and will be removed in v1.7.
```
ananthsub, line 31:
```suggestion
        The `TrainerOptimizersMixin` was deprecated in v1.6 and will be removed in v1.8.
```

ananthsub (approved):

rohitgr7 (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2287,6 +2286,11 @@ def terminate_on_nan(self, val: bool) -> None:
         )
         self._terminate_on_nan = val  # : 212
 
+    def _convert_to_lightning_optimizers(self):
```
carmocca, line 2289: Add typing

pytorch_lightning/trainer/trainer.py
```diff
@@ -1852,7 +1851,7 @@ def slurm_job_id(self) -> Optional[int]:
     @property
     def lightning_optimizers(self) -> List[LightningOptimizer]:
         if self._lightning_optimizers is None:
-            self.convert_to_lightning_optimizers()
+            self._convert_to_lightning_optimizers()
```
carmocca, line 1854: This returns a list but this call converts to dict?

carmocca (commented):
I think this PR could be split into separate pieces, one for the actual intention of "deprecating and moving functionality" and other for all the strategy's optimizer refactor changes

pytorch_lightning/core/optimizer.py
daniellepintz: why should this be on the public API? do we expect end users to call it directly? @rohitgr7 previously wanted me to change this to private

ananthsub: I see, to start with we could make it private, and make sure the API is tested & stable. Since it's meant to be called externally by the strategy implementations (not end users), I think eventually this should be public. But it's minor if we want to make it private first and later make it public. That's always easier than the other way around 

@awaelchli @rohitgr7 what do you think?

rohitgr7: from the user's point, it should be private, but as @ananthsub mentioned that this is being accessed outside its module. tbh we have already broken this thing at a few other places already (accessing private functions outside their modules), so not sure if it's a common thing in FOSS. IMO keep it private.

awaelchli: Public or protected, fine with me. Private is a bit extreme considering it's getting imported/called from outside this module.

daniellepintz: I think we should keep it private since it's not meant to be called by end users, and as Rohit said, we already break the rule of accessing private functions outside their modules all over the codebase

pytorch_lightning/trainer/trainer.py
daniellepintz: Weird. Changed the return type to dict

pytorch_lightning/core/optimizer.py
daniellepintz: Oh, didn't realize private and protected are different things. Then we can make it protected (which is is now).

daniellepintz:
@carmocca I removed the strategy refactor, ready for final review and merge!

pytorch_lightning/trainer/optimizers.py
```diff
@@ -21,15 +21,26 @@
...
 
 class TrainerOptimizersMixin(ABC):
+    r"""
+    .. deprecated:: v1.6
+        The `TrainerOptimizersMixin` was deprecated in v1.6 and will be removed in v1.8.
+    """
 
     _lightning_optimizers: Optional[List[LightningOptimizer]]
 
     def init_optimizers(self, model: Optional["pl.LightningModule"]) -> Tuple[List, List, List]:
```
ananthsub, line 36: could you change the implementation of this to call what's in `core/optimizers` now so there's no divergence?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1920,9 +1919,9 @@ def slurm_job_id(self) -> Optional[int]:
         return SLURMEnvironment.job_id()
 
     @property
-    def lightning_optimizers(self) -> List[LightningOptimizer]:
+    def lightning_optimizers(self) -> Dict[int, LightningOptimizer]:
```
ananthsub, line 1922: was this a bug before that the typehint was wrong?

daniellepintz: Yes

pytorch_lightning/trainer/optimizers.py
```diff
@@ -20,29 +20,29 @@
...
-            optim_conf = _MockOptimizer()
-
-        optimizers, lr_schedulers, optimizer_frequencies, monitor = self._configure_optimizers(optim_conf)
-        lr_schedulers = self._configure_schedulers(lr_schedulers, monitor, not pl_module.automatic_optimization)
-        _validate_scheduler_optimizer(optimizers, lr_schedulers)
-        return optimizers, lr_schedulers, optimizer_frequencies
+        return _init_optimizers_and_lr_schedulers(pl_module)
 
     @staticmethod
     def _configure_optimizers(
```
ananthsub, line 48: this isn't used anymore and can be removed

pytorch_lightning/trainer/trainer.py
ananthsub: could you add this to the changelog under a fix?

daniellepintz: sure I will add it. But does this really qualify as a bug? It is just a typehint

daniellepintz: sure I will add it. I didn't really consider a wrong typehint a bug

pytorch_lightning/trainer/optimizers.py
daniellepintz: Good call. done

pytorch_lightning/trainer/optimizers.py
```diff
@@ -13,103 +13,42 @@
...
+        r"""
+        .. deprecated:: v1.6
+            `TrainerOptimizersMixin.convert_to_lightning_optimizers` was deprecated in v1.6 and will be removed in v1.8.
+        """
+        rank_zero_deprecation(
+            "`TrainerOptimizersMixin.convert_to_lightning_optimizers` was deprecated in v1.6 and will be removed in "
+            "v1.8."
+        )
+
         def _convert_to_lightning_optimizer(trainer, optimizer):
```
ananthsub, line 52: this should also use the new implementation of `_convert_to_lightning_optimizer` to avoid these falling out of sync

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +164,222 @@ def closure_dis():
...
+            f"Found unsupported keys in the optimizer configuration: {set(extra_keys)}", category=RuntimeWarning
+        )
+
+
+def _convert_to_lightning_optimizer(trainer: "pl.Trainer", optimizer: Optimizer) -> LightningOptimizer:
+    if not isinstance(optimizer, LightningOptimizer):
+        optimizer = LightningOptimizer(optimizer)  # type: ignore [assignment]
+    optimizer._on_trainer_init(trainer)
+    return optimizer  # type: ignore [return-value]
+
```
ananthsub, line 359: why not also add `_convert_to_lightning_optimizers` here? then all the logic is in one place

pytorch_lightning/trainer/optimizers.py
daniellepintz: Fixed

pytorch_lightning/core/optimizer.py
daniellepintz: Fixed

awaelchli (approved):

daniellepintz:
Would appreciate a final review here from code owners @Borda @carmocca @justusschock @kaushikb11 
(And feel free to merge if it's ready)

pytorch_lightning/core/optimizer.py
```diff
@@ -54,7 +57,10 @@ def optimizer(self) -> Optimizer:
         return self._optimizer
 
     def _on_trainer_init(self, trainer: "pl.Trainer") -> None:
-        self._trainer = proxy(trainer)
+        if isinstance(trainer, weakref.ProxyType):
```
ananthsub, line 60: Could you add a comment for this?

daniellepintz: sure. although I don't really understand why it's needed and it wasn't needed before 🙁

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +169,227 @@ def closure_dis():
...
+    # two lists, optimizer + lr schedulers
+    elif (
+        isinstance(optim_conf, (list, tuple))
+        and len(optim_conf) == 2
+        and isinstance(optim_conf[0], list)
+        and all(isinstance(opt, Optimizer) for opt in optim_conf[0])
+    ):
+        opt, sch = optim_conf
+        optimizers = opt
+        lr_schedulers = sch if isinstance(sch, list) else [sch]
```
Borda, line 209:
```suggestion
        lr_schedulers = sch if isinstance(sch, Sequence) else [sch]
```

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +169,227 @@ def closure_dis():
...
+    # unknown configuration
+    else:
+        raise MisconfigurationException(
+            "Unknown configuration for model optimizers."
+            " Output from `model.configure_optimizers()` should either be:\n"
+            " * `torch.optim.Optimizer`\n"
+            " * [`torch.optim.Optimizer`]\n"
+            " * ([`torch.optim.Optimizer`], [`torch.optim.lr_scheduler`])\n"
+            ' * {"optimizer": `torch.optim.Optimizer`, (optional) "lr_scheduler": `torch.optim.lr_scheduler`}\n'
+            ' * A list of the previously described dict format, with an optional "frequency" key (int)'
```
Borda, line 250: can be simplified and drop `torch.optim`?

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +169,227 @@ def closure_dis():
...
+    elif isinstance(optim_conf, (list, tuple)) and all(isinstance(opt, Optimizer) for opt in optim_conf):
+        optimizers = list(optim_conf)
+    # unknown configuration
+    else:
+        raise MisconfigurationException(
+            "Unknown configuration for model optimizers."
+            " Output from `model.configure_optimizers()` should either be:\n"
+            " * `torch.optim.Optimizer`\n"
+            " * [`torch.optim.Optimizer`]\n"
+            " * ([`torch.optim.Optimizer`], [`torch.optim.lr_scheduler`])\n"
```
Borda, line 248: is `torch.optim.lr_scheduler` is class?

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +169,227 @@ def closure_dis():
...
+        if optimizer_frequencies and len(optimizer_frequencies) != len(optimizers):
+            raise ValueError("A frequency must be given to each optimizer.")
+    # single list or tuple, multiple optimizer
+    elif isinstance(optim_conf, (list, tuple)) and all(isinstance(opt, Optimizer) for opt in optim_conf):
+        optimizers = list(optim_conf)
+    # unknown configuration
+    else:
+        raise MisconfigurationException(
+            "Unknown configuration for model optimizers."
+            " Output from `model.configure_optimizers()` should either be:\n"
```
Borda, line 245:
```suggestion
            " Output from `model.configure_optimizers()` should either be one of:\n"
```

pytorch_lightning/core/optimizer.py
```diff
@@ -162,3 +169,227 @@ def closure_dis():
...
+    def __init__(self) -> None:
+        super().__init__([torch.zeros(1)], {})
+
+    def add_param_group(self, param_group: Dict[Any, Any]) -> None:
+        pass  # Do Nothing
+
+    def load_state_dict(self, state_dict: Dict[Any, Any]) -> None:
+        pass  # Do Nothing
+
+    def state_dict(self) -> Dict[Any, Any]:
```
Borda, line 384: I guess that fist is a string...
```suggestion
    def state_dict(self) -> Dict[str, Any]:
```

Borda (approved):

pytorch_lightning/core/optimizer.py
daniellepintz: yes: https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py#L22

daniellepintz:
Thanks @Borda! Updated the PR with the changes

pytorch_lightning/core/optimizer.py
daniellepintz: I am getting these mypy errors when I make this change:
`pytorch_lightning/core/optimizer.py:191: error: Need type annotation for "lr_schedulers" (hint: "lr_schedulers: List[<type>] = ...")  [var-annotated]
pytorch_lightning/core/optimizer.py:206: error: Incompatible types in assignment (expression has type "Sequence[Any]", variable has type "List[Any]")  [assignment]
`

PR #11177, opened 2021-12-20
Remove explicit isinstance checks in strategies for checkpoint io

carmocca:
## What does this PR do?

Removes `isinstance` checks when setting the `checkpoint_io` class.

These are unpythonic as they block duck-typing. I don't think there's a strong argument for having them.

Fixes https://github.com/PyTorchLightning/pytorch-lightning/pull/11071#discussion_r770964490

None

## Before submitting

## PR review

cc @borda @akihironitta


tchaton (approved):
LGTM !

pytorch_lightning/plugins/training_type/deepspeed.py
```diff
@@ -857,12 +856,6 @@ def register_plugins(cls, plugin_registry: Dict) -> None:
             offload_optimizer_device="nvme",
         )
 
-    @DDPPlugin.checkpoint_io.setter
-    def checkpoint_io(self, io: Optional[CheckpointIO]) -> None:
-        if io is not None:
-            raise MisconfigurationException("DeepSpeed does not support custom `CheckpointIO` plugin.")
-        self._checkpoint_io = io
```
rohitgr7, line 864: keeping this would be a problem? it doesn't actually support any checkpoint_io and we don't have any isinstance check?

pytorch_lightning/plugins/training_type/single_tpu.py
```diff
@@ -86,9 +85,3 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], filepath: _PATH) -> None:
     def teardown(self) -> None:
         # TPU teardown
         os.environ.pop("PT_XLA_DEBUG", None)
-
-    @SingleDevicePlugin.checkpoint_io.setter
-    def checkpoint_io(self, io: Optional[XLACheckpointIO]) -> None:
-        if io is not None and not isinstance(io, XLACheckpointIO):
-            raise MisconfigurationException(f"{self.__class__.__name__}.checkpoint_io` must be a `XLACheckpointIO`.")
-        self._checkpoint_io = io
```
four4fish, line 94: After we remove this, if user pass in a wrong/customized checkpoint, will we override it or fail silently?
Do you think we should include this logic in accelerator_connector? as plugins choosing and misconfig check should be in accelerator_connector?

pytorch_lightning/plugins/training_type/deepspeed.py
carmocca: > it doesn't actually support any checkpoint_io 

Can you explain why not?

pytorch_lightning/plugins/training_type/single_tpu.py
carmocca: If the users chooses to do this, he should be allowed. It's his/her responsibility to make sure their implementation is correct.

Typing checks can help with what are the expected input types.

I don't think this validation should be moved.

pytorch_lightning/plugins/training_type/deepspeed.py
rohitgr7: it uses deepspeed engine
https://github.com/PyTorchLightning/pytorch-lightning/blob/9da78a94bd1e9c3088930b24133be2b11f1dbcef/pytorch_lightning/plugins/training_type/deepspeed.py#L720-L739

carmocca: This hook being overridden should not mean you cannot pass a different IO instance. Perhaps this new instance provides the user with other implementations useful for their deepspeed engine subclass

rohitgr7: ah! good point 👍

rohitgr7: if they are subclassing and overriding the default implementation, then they should be overriding the checkpoint_io setter too then? because for that case DeepspeedStrategy is the base and if something is not available in base and I am introducing it in the fork, I should also enable it?

or is this how things are done in such an unusual case?

ananthsub (approved):

pytorch_lightning/plugins/training_type/deepspeed.py
ananthsub: do you think it's misleading if users set this but don't see their component being used by the strategy?

awaelchli (approved):

pytorch_lightning/plugins/training_type/deepspeed.py
carmocca: I think it's okay. Save and load aren't used but remove is

PR #11199, opened 2021-12-21
Fix CLI race condition saving the config

### LINKED ISSUE #11158: Hang when using Lightning CLI from config file and DDP
## 🐛 Bug

When I use Lightning CLI to instantiate training from a YAML config file with DDP, I got the error below and the process will hang.

> ```RuntimeError: SaveConfigCallback expected xxxx/lightning_logs/version_5/config.yaml to NOT exist. Aborting to avoid overwriting results of a previous run. You can delete the previous config file, set `LightningCLI(save_config_callback=None)` to disable config saving, or set `LightningCLI(save_config_overwrite=True)` to overwrite the config file.```

Ctrl + C will not work. I have to kill the process manually.

I suspect this is because each DDP process tries to write the same config file to disk. Passing either `save_config_callback=None` or `save_config_overwrite=True` like suggested in the error message solves the issue.

### To Reproduce

```python3
# bug.py
import torch

from pytorch_lightning.utilities.cli import LightningCLI

class RandomDataset(Dataset):
    def __init__(self, size, num_samples):
        self.len = num_samples
        self.data = torch.randn(num_samples, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len

    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

    def train_dataloader(self):

    def val_dataloader(self):

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

if __name__ == "__main__":
    LightningCLI(BoringModel)
```

```yaml
# bug.yaml
trainer:
  gpus: 2
  strategy: ddp
  max_epochs: 10
```

Run Lightning CLI
```bash
python bug.py fit --config bug.yaml
```

### Expected behavior

There should be no error by default, without explicitly passing `save_config_callback=None` or `save_config_overwrite=True` to Lightning CLI.

### Environment

* CUDA:
        - GPU:
                - GeForce RTX 3090
                - GeForce RTX 3090
                - GeForce RTX 3090
                - GeForce RTX 3090
        - available:         True
        - version:           11.3
* Packages:
        - numpy:             1.21.2
        - pyTorch_debug:     False
        - pyTorch_version:   1.10.0
        - pytorch-lightning: 1.5.5
        - tqdm:              4.62.3
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.8.12
        - version:           #86~18.04.1-Ubuntu SMP Fri Jun 18 01:23:22 UTC 2021

### Additional context

cc @carmocca @mauvilsa

carmocca:
## What does this PR do?

Fixes #11158

### Test plan

No test as it's caused by a race condition. Testing it would require adding delays which are slow and flaky.

None

## Before submitting

## PR review

cc @borda @carmocca @mauvilsa


awaelchli (approved):

pytorch_lightning/utilities/cli.py
```diff
@@ -404,21 +404,25 @@ def __init__(
...
+        if not self.overwrite:
+            if fs.isfile(config_path):
+                raise RuntimeError(
+                    f"{self.__class__.__name__} expected {config_path} to NOT exist. Aborting to avoid overwriting"
+                    " results of a previous run. You can delete the previous config file,"
+                    " set `LightningCLI(save_config_callback=None)` to disable config saving,"
+                    " or set `LightningCLI(save_config_overwrite=True)` to overwrite the config file."
+                )
+            # make sure all processes finished `isfile` before letting rank 0 write
+            trainer.training_type_plugin.barrier()
```
carmocca, line 420:
```suggestion
            trainer.strategy.barrier()
```

CHANGELOG.md
```diff
@@ -342,8 +342,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed double evaluation bug with fault-tolerance enabled where the second call was completely skipped ([#11119](https://github.com/PyTorchLightning/pytorch-lightning/pull/11119))
 
 
-- Fixed an issue with the `TPUSpawnPlugin` handling the `XLA_USE_BF16` environment variable incorrectly ([#10990](https://github.com/PyTorchLightning/pytorch-lightning/pull/10990))
+- Fixed `LightningCLI` race condition saving the config ([#11199](https://github.com/PyTorchLightning/pytorch-lightning/pull/11199))
```
s-rog, line 345:
```suggestion
- Fixed `LightningCLI` race condition while saving the config ([#11199](https://github.com/PyTorchLightning/pytorch-lightning/pull/11199))
```

s-rog (approved):

pytorch_lightning/utilities/cli.py
```diff
@@ -404,21 +404,25 @@ def __init__(
...
-        if not self.overwrite and os.path.isfile(config_path):
-            raise RuntimeError(
-                f"{self.__class__.__name__} expected {config_path} to NOT exist. Aborting to avoid overwriting"
-                " results of a previous run. You can delete the previous config file,"
-                " set `LightningCLI(save_config_callback=None)` to disable config saving,"
-                " or set `LightningCLI(save_config_overwrite=True)` to overwrite the config file."
-            )
+        if not self.overwrite:
+            if fs.isfile(config_path):
+                raise RuntimeError(
```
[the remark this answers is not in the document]
carmocca: > I guess so that all processes terminate when this happens.

Yes

> if it needs to raise it also triggers the termination of the other processes?

This would also work. We'd need to broadcast the boolean decision.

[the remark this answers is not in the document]
carmocca: Done

pytorch_lightning/utilities/cli.py
```diff
@@ -404,21 +404,31 @@ def __init__(
...
-                " results of a previous run. You can delete the previous config file,"
-                " set `LightningCLI(save_config_callback=None)` to disable config saving,"
-                " or set `LightningCLI(save_config_overwrite=True)` to overwrite the config file."
-            )
-        if trainer.is_global_zero:
+        fs = get_filesystem(log_dir)
+
+        if not self.overwrite:
+            # check if the file exists on rank 0
+            file_exists = fs.isfile(config_path) if is_global_zero else False
```
awaelchli, line 415: @carmocca in ModelCheckpoint we have the same code except that we use `fs.exists`. Should we make it consistent? 

This special way of checking file existence with broadcast could even become a function, as we may reuse it in the future.

awaelchli (approved):

pytorch_lightning/utilities/cli.py
carmocca: Perhaps, but for a follow-up.

Do you think it should be part of checkpoint io?

pytorch_lightning/utilities/cli.py
```diff
@@ -404,21 +404,31 @@ def __init__(
     def setup(self, trainer: Trainer, pl_module: LightningModule, stage: Optional[str] = None) -> None:
         # save the config in `setup` because (1) we want it to save regardless of the trainer function run
         # and we want to save before processes are spawned
-        log_dir = trainer.log_dir
+        log_dir = trainer.log_dir  # this broadcasts the directory
         assert log_dir is not None
+        is_global_zero = trainer.is_global_zero
```
Borda, line 409: is this needed while it is trivial mapping?

Borda (approved):

PR #11201, opened 2021-12-21
Move `epoch_{start,end}` hooks from `TrainingEpochLoop` to `FitLoop`

carmocca:
## What does this PR do?

The `TrainingEpochLoop` (a loop over batches) currently calls the epoch start/end hooks.

These should be called by `FitLoop` instead (a loop over epochs). Note how we currently access the parent progress with `self.trainer.fit_loop.epoch_progress...` which is a clear indicator that the structure is not correct.

This PR fixes it.

None

## Before submitting

## PR review

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7 @carmocca @ninginthecloud @ananthsub


pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -261,14 +261,14 @@ def on_validation_start(self, trainer, pl_module):
     def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx, dataloader_idx):
         if self._should_update(self.val_batch_idx):
             _update_n(self.val_progress_bar, self.val_batch_idx)
-            if trainer.state.fn == pl.trainer.states.TrainerFn.FITTING:
+            if trainer.state.fn == "fit":
```
kaushikb11, line 264: Curious. Why change this?

kaushikb11 (approved):

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: This was missed in the previous PR, the change is consistent with the reasoning here: https://github.com/PyTorchLightning/pytorch-lightning/pull/11213#discussion_r779514176

pytorch_lightning/loops/fit_loop.py
```diff
@@ -221,27 +228,76 @@ def on_advance_start(self) -> None:  # type: ignore[override]
...
-            self.global_step += 1
+            self._outputs = self.epoch_loop.run(data_fetcher)
 
     def on_advance_end(self) -> None:
+        # inform logger the batch loop has finished
+        self.trainer.logger_connector.epoch_end_reached()
+
+        # get the model and call model.training_epoch_end
+        model = self.trainer.lightning_module
+        if is_overridden("training_epoch_end", model) and self._outputs:
```
rohitgr7, line 256: quick ques: what if the user stores the outputs in some state variables within `training_step` and returns only the loss for some reason. Should it avoid calling `training_epoch_end` in such a case too? maybe they could still use `training_epoch_end` hook and clean up their cache manually.

carmocca: There's `on_train_epoch_end` for that. Doing that in `training_epoch_end` would be inefficient because then we save all losses for nothing.

Note also that this condition is copied verbatim from where it was.

justusschock (approved):

rohitgr7 (approved):

PR #9310, opened 2021-09-03
fix progress bar restart with fault-tolerant training enabled

awaelchli:
## What does this PR do?

Fixes a TODO in #9130 

The progress bar callback did not resume correctly, because of two reasons:
- `on_train_start()` was resuming the correct progress value but `on_train_epoch_start()` would overwrite it back to 0
- the `on_*_epoch_start` hooks reset the tqdm counter to 0 instead of the value we would resume from

This came up in the context of fault tolerant training. 

<details>
  <summary>Repro script</summary>

```python
import os
from time import sleep

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        sleep(0.5)
        print(batch_idx, self.trainer.fit_loop.batch_idx)
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)


def run():
    train_data = DataLoader(RandomDataset(32, 64), batch_size=2)

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        weights_summary=None,
        # resume_from_checkpoint=".pl_auto_save.ckpt",
    )
    trainer.fit(model, train_dataloaders=train_data)


if __name__ == "__main__":
    run()

```

</details>

**The added test fails on master.**

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
[...]

PR #11213, opened 2021-12-22
Integrate progress tracking into the progress bar

carmocca:
## What does this PR do?

Removes the index increments from the progress bar implementation, instead, relying on the loops' progress tracking.

None

## Before submitting

## PR review

cc @justusschock @awaelchli @akihironitta @rohitgr7 @carmocca @ninginthecloud @SeanNaren @kaushikb11 @borda


tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -568,36 +565,6 @@ def get_metrics(self, trainer: Trainer, model: LightningModule):
     assert "v_num" not in standard_metrics.keys()
 
 
-def test_tqdm_progress_bar_main_bar_resume():
-    """Test that the progress bar can resume its counters based on the Trainer state."""
```
carmocca, line 572: I'm not sure about what the intention of this test was. If anybody knows, I can put it back.

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -172,13 +172,8 @@ class CurrentProgressBar(TQDMProgressBar):
         val_batches_seen = 0
         test_batches_seen = 0
 
-        def on_train_batch_start(self, trainer, pl_module, batch, batch_idx):
-            super().on_train_batch_start(trainer, pl_module, batch, batch_idx)
-            assert self.train_batch_idx == trainer.fit_loop.batch_idx
```
carmocca, line 177: Both of these come from progress tracking now, so no need to test the equality. This is supposed to test the functionality of the progress bar.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -47,52 +47,57 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
     def test_batch_idx(self) -> int:
-        """The current batch index being processed during testing.
+        """The number of batches processed during testing.
 
         Use this to update your progress bar.
         """
-        return self._test_batch_idx
+        if self.trainer is None:
+            return 0
+        return self.trainer.test_loop.epoch_loop.batch_progress.current.processed
```
tchaton, line 86: Did you choose this over `self.trainer.fit_loop.batch_idx` as you are only want to display the processed value?

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -216,87 +225,74 @@ def on_sanity_check_end(self, trainer, pl_module):
...
-        super().on_train_batch_end(trainer, pl_module, outputs, batch, batch_idx)
         if self._should_update(self.train_batch_idx):
-            self._update_bar(self.main_progress_bar)
+            _update_n(self.main_progress_bar, self.train_batch_idx + self._val_processed)
             self.main_progress_bar.set_postfix(self.get_metrics(trainer, pl_module))
 
     def on_train_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        if self.is_enabled:
-            self._update_bar(self.main_progress_bar)
+        _update_n(self.main_progress_bar, self.train_batch_idx + self._val_processed)
```
tchaton, line 247: Why updating if the progress bar can disabled ?

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -282,48 +277,48 @@ def test_tqdm_progress_bar_value_on_colab(tmpdir):
...
+    @n.setter
+    def n(self, value):
+        self.__n = value
+        # track the changes in the `n` value
+        if not len(self.n_values) or value != self.n_values[-1]:
+            self.n_values.append(value)
 
 
 @pytest.mark.parametrize(
-    "train_batches,val_batches,refresh_rate,train_deltas,val_deltas",
```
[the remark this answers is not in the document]
tchaton: cleaner !

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -282,48 +271,48 @@ def test_tqdm_progress_bar_value_on_colab(tmpdir):
...
+        [2, 3, 1, [1, 2, 3, 4, 5], [1, 2, 3]],
+        [0, 0, 3, None, None],
+        [1, 0, 3, [1], None],
         [1, 1, 3, [2], [1]],
-        [5, 0, 3, [3, 2], []],
-        [5, 2, 3, [3, 3, 1], [2]],
-        [5, 2, 6, [6, 1], [2]],
+        [5, 0, 3, [3, 5], None],
+        [5, 2, 3, [3, 7], [2]],
+        [5, 2, 6, [7], [2]],
```
tchaton, line 303: Can you explain me the values there ? Not sure to understand where the 7 is coming from.

pytorch_lightning/callbacks/progress/base.py
carmocca: This is the `test_batch_idx` so we wouldn't access the `fit_loop.batch_idx`

And `processed` is what matches the previous manually-managed value when the hooks are called.

tests/callbacks/test_tqdm_progress_bar.py
carmocca: We had 6 + 1 before because deltas were used. Deltas as in the difference between the previous and current call.

Now it's one update with the value 7 (6 + 1)

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: `_update_n` checks for disabled

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -282,48 +277,48 @@ def test_tqdm_progress_bar_value_on_colab(tmpdir):
...
     [
-        [2, 3, 1, [1, 1, 1, 1, 1], [1, 1, 1]],
-        [0, 0, 3, [], []],
-        [1, 0, 3, [1], []],
+        [2, 3, 1, [1, 2, 3, 4, 5], [1, 2, 3]],
+        [0, 0, 3, None, None],
+        [1, 0, 3, [1], None],
         [1, 1, 3, [2], [1]],
-        [5, 0, 3, [3, 2], []],
-        [5, 2, 3, [3, 3, 1], [2]],
```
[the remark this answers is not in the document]
tchaton: Yes same here. Seems like a bug.

tests/callbacks/test_tqdm_progress_bar.py
tchaton: Hey @carmocca.

I am not sure to fully understand your explanation.

Here is what I understand.

train_updates = [all the values divisible by train_batches] + [train_batches + val_batches]

[5, 2, 3] -> [3] + [5 + 2]
[5, 2, 6] -> [] (none is divisble) + [5 + 2]

tchaton, line 303:
```suggestion
        [5, 3, 6, [5 + 3], [3]],
```

tchaton: IMO, the refresh should take both train and val as a single unit which would give `[5, 3, 6] -> [8], [1]` assuming no `check_val_interval`.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -47,52 +47,57 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
-        self._predict_batch_idx = 0
+        self._trainer: Optional["pl.Trainer"] = None
 
     @property
-    def trainer(self):
+    def trainer(self) -> "pl.Trainer":
         return self._trainer
 
     @property
     def train_batch_idx(self) -> int:
```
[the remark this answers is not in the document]
awaelchli: That is a matter of interpretation. The original intention was to express the progress bar showing the current batch being processed, and this makes sense if you look at the empty bar showing 0/n, it does NOT mean batch 0 has completed, it means batch 0 is being processed/trained. You can then always re-interpret this as saying 0 batches have been completed. Both views are valid.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -216,87 +225,74 @@ def on_sanity_check_end(self, trainer, pl_module):
...
     def on_test_batch_end(self, trainer, pl_module, outputs, batch, batch_idx, dataloader_idx):
-        super().on_test_batch_end(trainer, pl_module, outputs, batch, batch_idx, dataloader_idx)
         if self._should_update(self.test_batch_idx):
-            self._update_bar(self.test_progress_bar)
+            _update_n(self.test_progress_bar, self.test_batch_idx)
 
     def on_test_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        if self.is_enabled:
-            self._update_bar(self.test_progress_bar)
+        _update_n(self.test_progress_bar, self.test_batch_idx)
```
awaelchli, line 284: this will update the bar when it is disabled?

tests/callbacks/test_tqdm_progress_bar.py
awaelchli: It was part of the fix #9310 as git blame shows. Resuming the progress bar needed to be taken care of separately which is redundant now that it relies on progress tracking. Given progress tracking is tested.

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: `_update_n` checks for disabled

pytorch_lightning/callbacks/progress/base.py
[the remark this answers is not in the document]
carmocca: Well sure, but generally in programming indices is 0-based where the 0-th occurrence is the first one.

Here 0-th is the lack of an occurrence (no batch ran yet), that's what I refer to as a count.

In any case this is irrelevant to the changes in this PR. Resolving

tests/callbacks/test_tqdm_progress_bar.py
carmocca: > suggestion

I don't think it's worth keeping `5 + 3`, as it only makes sense in this review because you can see the previous change. When somebody revisits in a month, the person will think "why is this left as 5 + 3"?

> IMO, the refresh should take both train and val as a single unit which would give [5, 3, 6] -> [8], [1] assuming no check_val_interval.

It could, but that is unrelated to this PR. Here I haven't changed the refresh logic

carmocca: I don't get your point entirely, but this PR does not touch the refresh rate logic, it's still:

`(idx % self.refresh_rate == 0)` where the same `idx` values are passed.

This 7 here comes from the update `on_train_epoch_end`, where both training and validation have run (5 + 2 = 7)

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -135,6 +135,15 @@ def is_enabled(self) -> bool:
     def is_disabled(self) -> bool:
         return not self.is_enabled
 
+    @property
+    def _val_processed(self):
+        if self.trainer is None:
+            return 0
+        if self.trainer.state.fn == "fit":
+            # use total in case validation runs more than once per training epoch
+            return self.trainer.fit_loop.epoch_loop.val_loop.epoch_loop.batch_progress.total.processed
```
awaelchli, line 144: does that mean we reset the val_loop batch_progress.total every training epoch? I couldn't find that in the code. That would be a requirement to make this work with the current way the bar is supposed to report progress on mutiple val runs perepoch

awaelchli: If we take this example below and run it with your branch vs. master, we see there is a different behavior when running into the second epoch. 
Let's think about how to rectify this tomorrow


```py

import os
import time

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        time.sleep(0.5)
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        time.sleep(0.5)
        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

def run():
    train_data = DataLoader(RandomDataset(32, 10), batch_size=1)
    val_data = DataLoader(RandomDataset(32, 5), batch_size=1)

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        num_sanity_val_steps=0,
        max_epochs=2,
        val_check_interval=0.5
    )


if __name__ == "__main__":
    run()

```

carmocca: I wondered if somebody would notice this :P

I opened this PR at the same time than #11244. That one fixes the bug you noticed. This branch wasn't updated after #11244 got merged.

You'll see it working properly now that I've merged master.

tests/callbacks/test_tqdm_progress_bar.py
tchaton: > I don't think it's worth keeping 5 + 3, as it only makes sense in this review because you can see the previous change. When somebody revisits in a month, the person will think "why is this left as 5 + 3"?

Yes, that's entirely fine.

> It could, but that is unrelated to this PR. Here I haven't changed the refresh logic
Yes, let me open an issue.

tchaton (approved):
LGTM !

pytorch_lightning/callbacks/progress/tqdm_progress.py
awaelchli: Yes seems to work now!

awaelchli (approved):

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -135,6 +135,15 @@ def is_enabled(self) -> bool:
     def is_disabled(self) -> bool:
         return not self.is_enabled
 
+    @property
+    def _val_processed(self):
+        if self.trainer is None:
+            return 0
+        if self.trainer.state.fn == "fit":
```
rohitgr7, line 142: use `TrainerFn`?

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -47,52 +47,57 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
     def val_batch_idx(self) -> int:
-        """The current batch index being processed during validation.
+        """The number of batches processed during validation.
 
         Use this to update your progress bar.
         """
-        return self._val_batch_idx
+        if self.trainer is None:
+            return 0
+        if self.trainer.state.fn == "fit":
```
rohitgr7, line 74: use `TrainerFn`?

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -47,52 +47,57 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
     def train_batch_idx(self) -> int:
-        """The current batch index being processed during training.
+        """The number of batches processed during training.
 
         Use this to update your progress bar.
         """
-        return self._train_batch_idx
+        if self.trainer is None:
+            return 0
+        return self.trainer.fit_loop.epoch_loop.batch_progress.current.processed
```
rohitgr7, line 64: I wonder if this will break the kfold example.

rohitgr7: https://github.com/PyTorchLightning/pytorch-lightning/blob/9b873dcfcc9f060a351d559c30a0250efbd8a6b8/pl_examples/loop_examples/kfold.py#L279-L280
https://github.com/PyTorchLightning/pytorch-lightning/blob/9b873dcfcc9f060a351d559c30a0250efbd8a6b8/pl_examples/loop_examples/kfold.py#L173-L176

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: Since this is an "optional" callback and not an internal component, I prefer to use "fit" as a user would.

carmocca: Also, making the import will increase the chances of a circular import

pytorch_lightning/callbacks/progress/base.py
carmocca: It doesn't because of

https://github.com/PyTorchLightning/pytorch-lightning/blob/3a2df4f75d104b0ecf0af9b35e8ebd535a99a591/pl_examples/loop_examples/kfold.py#L236-L240

rohitgr7: ah! that's quite neat!

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: hmm. seeing that issue a lot lately regarding circular imports. I wonder if we can generalize it better somehow using some sort of trainer property.

rohitgr7 (approved):

PR #11217, opened 2021-12-22
Skip testing with PyTorch 1.7 and Python 3.9 on Ubuntu

### LINKED ISSUE #11184: `cpu (ubuntu-18.04, 3.9, oldest, stable)` fails with segmentation fault
## 🐛 Bug

Observed that `cpu (ubuntu-18.04, 3.9, oldest, stable)` quite often fails with segmentation fault after the actual tests had passed as the following:

```
=== 2218 passed, 532 skipped, 7 xfailed, 1470 warnings in 634.06s (0:10:34) ====
/home/runner/work/_temp/50029d2b-e6ee-470c-bd34-b5acf8834c42.sh: line 2:  3747 Segmentation fault      (core dumped) coverage run --source pytorch_lightning -m pytest pytorch_lightning tests -v --durations=50 --junitxml=junit/test-results-Linux-py3.9-oldest-stable.xml
```

### To Reproduce

- https://github.com/PyTorchLightning/pytorch-lightning/runs/4587925567?check_suite_focus=true#step:16:2825
- https://github.com/PyTorchLightning/pytorch-lightning/runs/4587519066?check_suite_focus=true#step:16:2825
- https://github.com/PyTorchLightning/pytorch-lightning/runs/4587417079?check_suite_focus=true#step:16:2825
- https://github.com/PyTorchLightning/pytorch-lightning/runs/4559402636?check_suite_focus=true#step:16:2821

### Expected behavior
No segmentation fault :green_apple: 

### Environment

- PyTorch Lightning Version (e.g., 1.5.0):
- PyTorch Version (e.g., 1.10):
- Python version (e.g., 3.9):
- OS (e.g., Linux):
- CUDA/cuDNN version:

### Additional context

akihironitta:
## What does this PR do?
Fixes #11184.

Skips testing with PyTorch 1.7 and Python 3.9 on Ubuntu due to a known issue with the setting: https://github.com/PyTorchLightning/pytorch-lightning/pull/11217#issuecomment-1006806605

## Before submitting

## PR review

## Did you have fun?


akihironitta:
wip

carmocca:
Looks like it's not the coverage as it still segfaulted.

Reading online, this could be useful

https://docs.python.org/3/library/faulthandler.html

```python
python -X faulthandler -m pytest tests ...
```

carmocca:
Hey! Snooped around the commits to check the progress.

Looks like `faulthandler` wasn't useful right?

gdb is also a good shot, checking the logs, I can see a really promising suspect

```python
2021-12-28T09:31:15.7410651Z #3  horovod::common::TimelineWriter::WriterLoop (this=0x7fffce5bf3c0 <horovod::common::(anonymous namespace)::horovod_global+32>) at /tmp/pip-install-kby_a7sr/horovod_d3effaaf9d4c4fc5afb4c4709d3b8444/horovod/common/timeline.cc:323
2021-12-28T09:31:15.7413575Z #4  0x00007fffe6edd68f in execute_native_thread_routine () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libc10.so
2021-12-28T09:31:15.7414965Z #5  0x00007ffff71f76db in start_thread (arg=0x7fff71ffd700) at pthread_create.c:463
2021-12-28T09:31:15.7415945Z #6  0x00007ffff753071f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
2021-12-28T09:31:15.7416476Z 
2021-12-28T09:31:16.7765898Z Thread 1076 (Thread 0x7fff72ffe700 (LWP 5570) "python"):
2021-12-28T09:31:16.7767171Z #0  0x00007ffff7530a47 in epoll_wait (epfd=16, events=0x7fff72ffdba0, maxevents=64, timeout=10) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
2021-12-28T09:31:16.7770120Z #1  0x00007fffdaf1bbd5 in gloo::transport::tcp::Loop::run() () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so
2021-12-28T09:31:16.7772120Z #2  0x00007fffe731f6df in execute_native_thread_routine () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch.so
2021-12-28T09:31:16.7773645Z #3  0x00007ffff71f76db in start_thread (arg=0x7fff72ffe700) at pthread_create.c:463
2021-12-28T09:31:16.7774607Z #4  0x00007ffff753071f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
2021-12-28T09:31:16.7775142Z 
2021-12-28T09:31:16.7775731Z Thread 1075 (Thread 0x7fff81ffd700 (LWP 5569) "python"):
2021-12-28T09:31:16.7777088Z #0  0x00007ffff74f37a0 in __GI___nanosleep (requested_time=requested_time@entry=0x7fff81ffcbc0, remaining=remaining@entry=0x7fff81ffcbc0) at ../sysdeps/unix/sysv/linux/nanosleep.c:28
2021-12-28T09:31:16.7778775Z #1  0x00007fffce31254d in std::this_thread::sleep_for<long, std::ratio<1l, 1000000000l> > (__rtime=<synthetic pointer>...) at /usr/include/c++/7/thread:373
2021-12-28T09:31:16.7781088Z #2  horovod::common::(anonymous namespace)::RunLoopOnce (state=...) at /tmp/pip-install-kby_a7sr/horovod_d3effaaf9d4c4fc5afb4c4709d3b8444/horovod/common/operations.cc:597
2021-12-28T09:31:16.7784258Z #3  horovod::common::(anonymous namespace)::BackgroundThreadLoop (state=...) at /tmp/pip-install-kby_a7sr/horovod_d3effaaf9d4c4fc5afb4c4709d3b8444/horovod/common/operations.cc:546
```

Horovod!

So let's try skipping all horovod tests and see if it hangs.

akihironitta:
Thank you for having a look, Carlos!

> Looks like faulthandler wasn't useful right?

It seemed so as observed in 73f78f0. pytest actually runs the handler by default since v5.0 (https://docs.pytest.org/en/latest/how-to/failures.html#fault-handler), so there's basically no need to add `python -X faulthandler ...`, but even the option is added, it didn't provide any information for unknown reasons... 

---

Just for the record, things I tried are:
- At 3494f10, tried not installing horovod and skipping all horovod tests in favour of [Carlos's comment](https://github.com/PyTorchLightning/pytorch-lightning/pull/11217#issuecomment-1006212305), but it didn't resolve the issue.
- At 7ca993c, tried updating all testing packages (e.g. pytest) to the latest ones in the hope of resolving it somehow, but it didn't resolve the issue.

---

I'll now have a closer look at the gdb output again... When I've got no ideas for further possible investigation/solutions, I'll ask for some possible ideas to try here.

.github/workflows/ci_test-full.yml
```diff
@@ -108,58 +115,77 @@ jobs:
...
-      run: |
-        HOROVOD_BUILT=$(python -c "import horovod.torch; horovod.torch.nccl_built(); print('SUCCESS')" || true)
-        if [[ $HOROVOD_BUILT != "SUCCESS" ]]; then
-          pip uninstall -y horovod
-          echo $(grep "horovod" requirements/extra.txt) > requirements/horovod.txt
-          pip install --no-cache-dir -r requirements/horovod.txt
-        fi
-        horovodrun --check-build
-      shell: bash
+    # - name: Reinstall Horovod if necessary
```
carmocca, line 118: This is not a good way to disable horovod, because it's already available in the docker image used by the job.

For this to work, you'd need to explicitly uninstall, not to stop installing.

I'd try hard-coding `_HOROVOD_AVAILABLE = False` so the tests get all skipped.

akihironitta: https://github.com/PyTorchLightning/pytorch-lightning/blob/cdc1bb1a6874ce0106766f8a862715b8b6b90564/.github/workflows/ci_test-full.yml#L64

Is it really? I thought excluding horovod installation above should skip all tests needing horovod as the docker image used by this job is normal ubuntu.

I'll try hard-coding it and uninstall it as you suggest anyway now :]

carmocca: NVM you are entirely right :)

.github/workflows/ci_test-full.yml
```diff
@@ -106,60 +113,81 @@ jobs:
...
-      if: always()
-      # see: https://github.com/actions/toolkit/issues/399
-      continue-on-error: true
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        file: coverage.xml
-        flags: cpu,pytest,python${{ matrix.python-version }}
-        name: CPU-coverage
-        fail_ci_if_error: false
+        python -m pytest tests/accelerators -v
```
carmocca, line 148: Interesting, so just accelerators still segfaults.

What if you disable accelerators and just test callbacks? (for example)

.github/workflows/ci_test-full.yml
```diff
@@ -106,60 +113,82 @@ jobs:
...
-      continue-on-error: true
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        file: coverage.xml
-        flags: cpu,pytest,python${{ matrix.python-version }}
-        name: CPU-coverage
-        fail_ci_if_error: false
+        python -m pytest tests/foo -v
+        python -m pytest tests/callbacks -v
+        python -m pytest tests/accelerators -v
```
carmocca, line 150:
```suggestion
        # python -m pytest tests/callbacks -v
        # python -m pytest tests/accelerators -v
```

.github/workflows/ci_test-full.yml
```diff
@@ -76,90 +80,57 @@ jobs:
...
-        fail_ci_if_error: false
+        cp pl_examples/bug_report/bug_report_model.py .
+        python -m pytest bug_report_model.py
+
+        # we know this fails
+        python -m pytest tests/callbacks/test_callback_hook_outputs.py::test_free_memory_on_eval_outputs -v
+
+    - name: Setup tmate session
+      if: ${{ failure() }}
+      uses: mxschmitt/action-tmate@v3
```
carmocca, line 107: With this, we can ssh and try things inside the image. I am able to reproduce the segfault running this command consistently.

.github/workflows/ci_test-full.yml
```diff
@@ -76,90 +80,57 @@ jobs:
...
-        token: ${{ secrets.CODECOV_TOKEN }}
-        file: coverage.xml
-        flags: cpu,pytest,python${{ matrix.python-version }}
-        name: CPU-coverage
-        fail_ci_if_error: false
+        cp pl_examples/bug_report/bug_report_model.py .
+        python -m pytest bug_report_model.py
+
+        # we know this fails
+        python -m pytest tests/callbacks/test_callback_hook_outputs.py::test_free_memory_on_eval_outputs -v
```
carmocca, line 103: I tired removing the callback, logger, modelcheckpoint and progress bar from this test and still segfaults. We need to start skipping things from training and narrow down what's the cause.

.github/workflows/ci_test-full.yml
```diff
@@ -50,12 +48,18 @@ jobs:
...
 
     - name: Setup Windows
       if: runner.os == 'windows'
       run: |
         python .github/prune-packages.py requirements/extra.txt "horovod"
 
+    - name: Setup Ubuntu
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt install gdb
```
carmocca, line 61: A smaller gdb stacktrace:

```python
$ gdb --return-child-result --batch -q -ex r -ex "thread apply all bt" --args "$(readlink -f $(which python))" -m pytest tests/callbacks/test_callback_hook_outputs.py::test_free_memory_on_eval_outputs
```
```python
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffe1405700 (LWP 4790)]
==================================================================== test session starts =====================================================================
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/runner/work/pytorch-lightning/pytorch-lightning, configfile: setup.cfg
plugins: rerunfailures-10.2
collected 1 item                                                                                                                                             

tests/callbacks/test_callback_hook_outputs.py [New Thread 0x7fffdc2a6700 (LWP 4791)]
.                                                                                                        [100%]

=============================================================== 1 passed, 8 warnings in 0.06s ================================================================

Thread 3 "python3.9" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffdc2a6700 (LWP 4791)]
0x00007ffff7e42323 in PyThreadState_Clear (tstate=0x7fffd4000b60) at Python/pystate.c:785
785     Python/pystate.c: No such file or directory.

Thread 3 (Thread 0x7fffdc2a6700 (LWP 4791) "python3.9"):
#0  0x00007ffff7e42323 in PyThreadState_Clear (tstate=0x7fffd4000b60) at Python/pystate.c:785
#1  0x00007ffff4fa022c in pybind11::gil_scoped_acquire::dec_ref() () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_python.so
#2  0x00007ffff4fa0269 in pybind11::gil_scoped_acquire::~gil_scoped_acquire() () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_python.so
#3  0x00007ffff52b7049 in torch::autograd::python::PythonEngine::thread_init(int, std::shared_ptr<torch::autograd::ReadyQueue> const&, bool) () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_python.so
#4  0x00007ffff49016df in execute_native_thread_routine () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch.so
#5  0x00007ffff79de609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#6  0x00007ffff7b1a293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 2 (Thread 0x7fffe1405700 (LWP 4790) "python3.9"):
#0  futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x7fffe388f6e0 <thread_status+96>) at ../sysdeps/nptl/futex-internal.h:183
#1  __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x7fffe388f690 <thread_status+16>, cond=0x7fffe388f6b8 <thread_status+56>) at pthread_cond_wait.c:508
#2  __pthread_cond_wait (cond=0x7fffe388f6b8 <thread_status+56>, mutex=0x7fffe388f690 <thread_status+16>) at pthread_cond_wait.c:638
#3  0x00007fffe1e175ab in blas_thread_server () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/numpy/core/../../numpy.libs/libopenblas64_p-r0-2f7c42d4.3.18.so
#4  0x00007ffff79de609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#5  0x00007ffff7b1a293 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Thread 1 (Thread 0x7ffff7876740 (LWP 4785) "python3.9"):
#0  0x00007ffff7a91d55 in _int_free (av=0x7ffff7be3b80 <main_arena>, p=<optimized out>, have_lock=<optimized out>) at malloc.c:4419
#1  0x00007fffe4c22a01 in c10::impl::AnnotatedKernel::~AnnotatedKernel() () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so
#2  0x00007fffe4c28f9f in c10::impl::OperatorEntry::deregisterKernel_(c10::Dispatcher const&, c10::optional<c10::DispatchKey>, std::_List_iterator<c10::impl::AnnotatedKernel>) () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so
#3  0x00007fffe4c1f16b in c10::Dispatcher::deregisterImpl_(c10::OperatorHandle const&, c10::OperatorName const&, c10::optional<c10::DispatchKey>, std::_List_iterator<c10::impl::AnnotatedKernel>) () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so
#4  0x00007fffe4c1f4fd in std::_Function_handler<void (), c10::Dispatcher::registerImpl(c10::OperatorName, c10::optional<c10::DispatchKey>, c10::KernelFunction, c10::optional<c10::impl::CppSignature>, std::unique_ptr<c10::FunctionSchema, std::default_delete<c10::FunctionSchema> >, std::string)::{lambda()#1}>::_M_invoke(std::_Any_data const&) () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so
#5  0x00007fffe4b58825 in torch::detail::TorchLibraryInit::~TorchLibraryInit() () from /opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/torch/lib/libtorch_cpu.so
#6  0x00007ffff7a41a27 in __run_exit_handlers (status=0, listp=0x7ffff7be3718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#7  0x00007ffff7a41be0 in __GI_exit (status=<optimized out>) at exit.c:139
#8  0x00007ffff7a1f0ba in __libc_start_main (main=0x555555555060 <main>, argc=4, argv=0x7fffffffcc08, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffcbf8) at ../csu/libc-start.c:342
#9  0x000055555555509e in _start ()
```

.github/workflows/ci_test-full.yml
carmocca: It's backward

carmocca:
Okay, it's not our problem


```python

$ python -m pytest kk.py 
==================================================================== test session starts =====================================================================
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/runner/work/pytorch-lightning/pytorch-lightning, configfile: setup.cfg
plugins: rerunfailures-10.2
collected 1 item                                                                                                                                             

kk.py .                                                                                                                                                [100%]

=============================================================== 1 passed, 8 warnings in 0.01s ================================================================
Segmentation fault (core dumped)
runner@fv-az199-142:~/work/pytorch-lightning/pytorch-lightning$ cat kk.py 
import torch


def test_segfault():
    # does not segfault always
    # with <32, it does not segfault???
    model = torch.nn.Linear(32, 1)
    batch = torch.randn(1, 32)
    x = model(batch)
    loss = x.sum()
    loss.backward()
```

Works without `pytest` though.

Let's reproduce it in a clean docker image and bring this to upstream torch

carmocca:
So the problem [was reported already](https://github.com/pytest-dev/pytest/issues/8841#issuecomment-874155081) but the fix only made it to 1.8, we'll need to exclude this job parametrization

.github/workflows/ci_test-full.yml
akihironitta: That's something I've been trying to find!!! (circle ci has the feature built-in, but couldn't find one for GitHub Actions...)

akihironitta:
@carmocca Thank you very much for taking your time and investigating this! :tada: 

> So the problem [was reported already](https://github.com/pytest-dev/pytest/issues/8841#issuecomment-874155081) but the fix only made it to 1.8, we'll need to exclude this job parametrization

I had a look at the GitHub issue before and had it in my mind, but gdb didn't provide such info (i.e. `pybind11::gil_scoped_acquire...`). I'm curious, for future debugging, about how you've got to see the above core dump. I'm assuming you've done via ssh to the session, correct?

akihironitta:
Just for the record, this issue seems to have been resolved in [#50998](https://github.com/pytorch/pytorch/pull/50998) which is included in v1.8.0 as seen in their release note: https://github.com/pytorch/pytorch/releases/tag/v1.8.0

kaushikb11 (approved):

rohitgr7 (approved):

carmocca (approved):

.github/workflows/ci_test-full.yml
```diff
@@ -21,6 +21,9 @@ jobs:
         python-version: ["3.7", "3.9"]  # minimum, maximum
         requires: ["oldest", "latest"]
         release: ["stable"]
+        exclude:
+          # Skip if torch<1.8 and py3.9 on Linux: https://github.com/pytorch/pytorch/issues/50014
+          - {os: ubuntu-18.04, python-version: "3.9", requires: "oldest", release: "stable"}
```
Borda, line 26: btw, I would not run any oldest for py3.9
```yaml
- {python-version: "3.9", requires: "oldest"}
```

akihironitta: @Borda Could I ask why you wouldn't? This problem shouldn't apply to other oldest settings.

Borda: use `release: "stable"` is redundant since we do not have other release values then stable now
and in general, if someone is using almost the latest python I don't think he would assist on the oldest packages...

akihironitta: > use `release: "stable"` is redundant since we do not have other release values then stable now

:+1: 

> in general, if someone is using almost the latest python I don't think he would assist on the oldest packages...

I see your point. I've just created an issue #11367 to gain some visibility on this.

PR #11268, opened 2021-12-27
Refine the pytorch profiler

guyang3532:
This pr include three changes:

1. Remove the white list for record function and start record function in pytorch profiler. Then we can get more information for pytorch profiler.
2. Remove the default activities, because they will be set in pytorch.
3. Move the `setup_profiler` to an early location. Because currently the `setup_profiler` is called too late and the hooks such as `on_fit_start` will not be recorded. Further more, the `trainer.lightning_module` is set to `profiler` in `setup_profiler`, and currently  when `profiler.start` is firstly called, the `setup_profiler` has not been called and the `profiler._lightning_module` is still None then the `profiler._register` will not work.


pytorch_lightning/profiler/pytorch.py
```diff
@@ -402,14 +391,11 @@ def start(self, action_name: str) -> None:
 
         if (
             self.profiler is not None
-            and (
-                any(action_name.endswith(func) for func in self._record_functions)
```
guotuofeng, line 406: do we still need self._record_functions and RECORD_FUNCTIONS/RECORD_FUNCTION_PREFIX?

guyang3532: Removed. Thanks

guotuofeng: RECORD_FUNCTION_PREFIX seems still no used?

pytorch_lightning/profiler/pytorch.py
```diff
@@ -194,14 +195,6 @@ def __call__(self, num_step: int) -> "ProfilerAction":
...
 
-    RECORD_FUNCTIONS = {
-        "training_step_and_backward",
-        "training_step",
-        "backward",
-        "validation_step",
-        "test_step",
-        "predict_step",
-    }
     RECORD_FUNCTION_PREFIX = "optimizer_step_with_closure_"
```
guotuofeng, line 198: RECORD_FUNCTION_PREFIX seems still no used?

pytorch_lightning/profiler/pytorch.py
```diff
@@ -287,9 +276,7 @@ def __init__(
         self._export_to_chrome = export_to_chrome
         self._row_limit = row_limit
         self._sort_by_key = sort_by_key or f"{'cuda' if profiler_kwargs.get('use_cuda', False) else 'cpu'}_time_total"
-        self._user_record_functions = record_functions or set()
-        self._record_functions_start = self._user_record_functions | self.START_RECORD_FUNCTIONS
-        self._record_functions = self._user_record_functions | self.RECORD_FUNCTIONS
+        self._record_functions_start = self.START_RECORD_FUNCTIONS
```
guotuofeng, line 279: do we still need it?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1246,8 +1247,8 @@ def run_stage(self) -> None:
         return self._run_stage()
 
     def _run_stage(self):
-        self.strategy.barrier("run-stage")
-        self.strategy.dispatch(self)
+        self.training_type_plugin.barrier("run-stage")
```
guotuofeng, line 1250: what's the consideration for this change?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1246,8 +1247,8 @@ def run_stage(self) -> None:
...
+        self.training_type_plugin.dispatch(self)
         self.__setup_profiler()
```
guotuofeng, line 1252: do we need it?

pytorch_lightning/trainer/trainer.py
guyang3532: this change is from rebase with the master code.

pytorch_lightning/trainer/trainer.py
guyang3532: this change is from rebase with the master code.

pytorch_lightning/profiler/pytorch.py
```diff
@@ -180,6 +182,11 @@ def __call__(self, num_step: int) -> "ProfilerAction":
 
         self._step()
         action = self._schedule(max(self.num_step, 0))
+        if self._prev_schedule_action == ProfilerAction.RECORD and action == ProfilerAction.WARMUP:
+            # Work around the corner case when validation starts before train.
+            # In this case, the action is RECORD in validation loop, and then call into the train
+            # and the action is still WARMUP in train and the pytorch will recognize this as error.
```
carmocca, line 188:
```suggestion
            # and the action is still WARMUP in train and pytorch will recognize this as error.
```

pytorch_lightning/profiler/pytorch.py
```diff
@@ -521,10 +494,10 @@ def _delete_profilers(self) -> None:
             self._register = None
 
     def teardown(self, stage: Optional[str] = None) -> None:
-        self._delete_profilers()
+        for _, v in self._recording_map.items():
+            v.__exit__(None, None, None)
+        self._recording_map.clear()
 
-        for k in self._recording_map:
-            self.stop(k)
```
carmocca, line 527: Why not use `stop` anymore?

At least you can extract the `__exit__` logic inside of it to a separate protected function that `teardown` also calls.

pytorch_lightning/profiler/pytorch.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, model: nn.Module) -> None:
         self._handles: Dict[str, List["RemovableHandle"]] = {}
 
     def _start_recording_forward(self, _: nn.Module, input: Tensor, record_name: str) -> Tensor:
-        record = record_function(record_name)
+        # Add [pytorch-lightning][module] in name for pytorch profiler to recognize
```
carmocca, line 72: Should it be something like `[pl]` instead of `[pytorch-lightning]`? The full name might decrease the readability too much. We could also not have it at all as using the lightning profiler and doing additional profiling without lightning seems really niche.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1103,14 +1103,16 @@ def _run(
...
-
         # ----------------------------
         # SET UP TRAINING
         # ----------------------------
         self._call_callback_hooks("on_before_accelerator_backend_setup")
         self.strategy.setup_environment()
+
+        self.__setup_profiler()
+        # hook
+        self._data_connector.prepare_data()
```
carmocca, line 1114: Is this change strictly necessary for the other changes in this PR to work? If not, then let's leave this for a separate PR. Hook order changes are important changes as they silently change the behaviour.

pytorch_lightning/profiler/pytorch.py
guyang3532: Because the `stop`  call the `del self._recording_map[action_name]` which can not be called when we iterate self._recording_map.
Yep, I extract the `__exit__` logit to a separate function.

pytorch_lightning/profiler/pytorch.py
guyang3532: Fixed. thanks!

pytorch_lightning/profiler/pytorch.py
guyang3532: fixed. thanks!

pytorch_lightning/trainer/trainer.py
guyang3532: Thanks for your suggestion. I remove this change and will submit it in another PR.

pytorch_lightning/profiler/pytorch.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, model: nn.Module) -> None:
         self._handles: Dict[str, List["RemovableHandle"]] = {}
 
     def _start_recording_forward(self, _: nn.Module, input: Tensor, record_name: str) -> Tensor:
-        record = record_function(record_name)
+        # Add [pl][module] in name for pytorch profiler to recognize
```
tchaton, line 72: Won't this make the trace harder to read?

pytorch_lightning/profiler/pytorch.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, model: nn.Module) -> None:
...
+        record = record_function("[pl][module]" + record_name)
```
tchaton, line 73: Note for below. Would be great to add decoration for the backward pass too. As we can access the loss, we can record backward.

pytorch_lightning/profiler/pytorch.py
```diff
@@ -180,6 +182,11 @@ def __call__(self, num_step: int) -> "ProfilerAction":
 
         self._step()
         action = self._schedule(max(self.num_step, 0))
+        if self._prev_schedule_action == ProfilerAction.RECORD and action == ProfilerAction.WARMUP:
+            # Work around the corner case when validation starts before train.
```
tchaton, line 186: Neat!

pytorch_lightning/profiler/pytorch.py
```diff
@@ -327,8 +319,6 @@ def _init_kineto(self, profiler_kwargs: Any) -> None:
         self._schedule = ScheduleWrapper(schedule) if schedule is not None else schedule
         self._profiler_kwargs["schedule"] = self._schedule
 
-        activities = profiler_kwargs.get("activities", None)
```
tchaton, line 330: I would prefer to keep those. It is important users have control over the activities to be used.

pytorch_lightning/profiler/pytorch.py
tchaton: Let's rename to `_exit_and_clear_recording_map` then for readability.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1109,6 +1109,9 @@ def _run(
         # ----------------------------
         self._call_callback_hooks("on_before_accelerator_backend_setup")
         self.strategy.setup_environment()
+
+        self.__setup_profiler()
```
tchaton, line 1113: We could move this even higher, at the beginning of the _run method as ddp_spawn and ddp have been aligned. @awaelchli

tchaton (commented):
Really nice work !

pytorch_lightning/profiler/pytorch.py
guyang3532: we will organize pytorch-lightning tags in a separate view and show without this header.

pytorch_lightning/profiler/pytorch.py
guyang3532: Yes, I just move the default value and user can also control over it though 'activities' argument.

pytorch_lightning/trainer/trainer.py
guyang3532: Yes, I tried it but found the `self._setup_profiler` must called after `self.strategy.setup_environment` and `self.strategy.setup_environment` must called after `self._call_callback_hooks("on_before_accelerator_backend_setup")`, if want move `self._setup_profiler` higher, must move all them three higher. I'm not sure if this will have any problem.

pytorch_lightning/profiler/pytorch.py
ananthsub: this is confusing that the action name being profiled isn't the exact same as what shows up in the trace. could you describe how this is intended to be post-processed/viewed?

would it make sense for `record_function` to accept metadata that allows the profiler to post-process these records?

pytorch_lightning/profiler/pytorch.py
guyang3532: Good suggestion! Maybe we can add it in another pr.

pytorch_lightning/profiler/pytorch.py
guyang3532: Currently the `record_function` doesn't support metadata. What I mean is we will set up a separate view for pytorch-lightning in pytroch profiler (maybe called pytorch-lightning view) which shows the action name without pre-tags. The pytorch-lightning view is for user to easily understand profiling info in pytorch-lightning. In the older trace view the showed name is still exactly same as the action name.

carmocca (approved):

pytorch_lightning/profiler/pytorch.py
```diff
@@ -519,11 +491,14 @@ def _delete_profilers(self) -> None:
             self._register.__exit__(None, None, None)
             self._register = None
 
+    def _clear_recording_map(self) -> None:
```
guotuofeng, line 494: how about https://stackoverflow.com/questions/5384914/how-to-delete-items-from-a-dictionary-while-iterating-over-it?

pytorch_lightning/profiler/pytorch.py
```diff
@@ -372,7 +365,7 @@ def _default_activities(self) -> List["ProfilerActivity"]:
         return activities
 
     def start(self, action_name: str) -> None:
-        if self.profiler is None and any(action_name.endswith(func) for func in self._record_functions_start):
+        if self.profiler is None:
```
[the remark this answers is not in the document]
guyang3532: Yes, we can do it  in later pr.

pytorch_lightning/profiler/pytorch.py
```diff
@@ -522,7 +507,7 @@ def _delete_profilers(self) -> None:
     def teardown(self, stage: Optional[str] = None) -> None:
         self._delete_profilers()
 
-        for k in self._recording_map:
+        for k in list(self._recording_map.keys()):
```
carmocca, line 510: nit:

```suggestion
        for k in self._recording_map:
```

tests/strategies/test_deepspeed_strategy.py
```diff
@@ -232,7 +232,8 @@ def backward(self, loss: Tensor, optimizer: Optimizer, optimizer_idx: int, *args
     [(RandomDataset, "auto"), (RandomDataset, 10), (RandomIterableDataset, "auto"), (RandomIterableDataset, 10)],
 )
 @mock.patch("deepspeed.init_distributed", autospec=True)
-def test_deepspeed_auto_batch_size_config_select(mock_deepspeed_distributed, tmpdir, dataset_cls, value):
+@mock.patch("pytorch_lightning.Trainer.log_dir", new_callable=mock.PropertyMock, return_value="abc")
```
carmocca, line 235: Why did you need to add this?

carmocca (approved):

pytorch_lightning/profiler/pytorch.py
guyang3532: Because `self.stop(k)`  calls `del self._recording_map[action_name]`, if iterate using `for k in self._recording_map:` , will raise "RuntimeError: dictionary changed size during iteration". This code is refer to https://stackoverflow.com/questions/5384914/how-to-delete-items-from-a-dictionary-while-iterating-over-it

tests/strategies/test_deepspeed_strategy.py
guyang3532: It is because profiler need to access `log_dir` and broadcast `log_dir` through `DeepSpeedStrategy`. And will raise "distributed not init" error because the  `deepspeed.init_distributed` is mocked in this test. So we mock the log_dir too.

guyang3532:
Hi @tchaton @carmocca, all test has passed and this pr is ready to merge : )

tchaton (approved):
LGTM ! Neat !

guotuofeng (approved):

guyang3532:
@tchaton could you help to merge this pr? Or do we need anyone else to review?

pytorch_lightning/profiler/pytorch.py
```diff
@@ -522,7 +509,7 @@ def _delete_profilers(self) -> None:
...
```
rohitgr7, line 512:
```suggestion
        for k in list(self._recording_map):
```

carmocca: https://github.com/PyTorchLightning/pytorch-lightning/pull/11268#discussion_r788540655

carmocca: You are right. :)

tests/profiler/test_profiler.py
```diff
@@ -387,11 +387,11 @@ def test_pytorch_profiler_trainer_ddp(tmpdir, pytorch_profiler):
         devices=2,
     )
     trainer.fit(model)
-    expected = {"[Strategy]DDPStrategy.validation_step"}
+    expected = {"[pl][profile][Strategy]DDPStrategy.validation_step"}
```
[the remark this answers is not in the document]
guyang3532: This is just for pytorch profiler and will not impact other profilers.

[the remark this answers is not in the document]
guyang3532: Because we need to recognize which tags is from Lightning (There are still other tags from pytorch internal) and we will remove these prefixes in the final view which showed to user in Pytorch Profiller.

PR #11270, opened 2021-12-27
Update logic to make sure logged_metrics always contain tensors

rohitgr7:
## What does this PR do?

Fixes #11251
Discussion for the issue: #11251
in short logged metrics contain mixed tensors and non-tensors values because we update them with metrics after converting them to scalars when we log the metrics. 

## Before submitting

## PR review

## Did you have fun?

Part of #1

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `DeviceStatsMonitor` to group metrics based on the logger's `group_separator` ([#11254](https://github.com/PyTorchLightning/pytorch-lightning/pull/11254))
 
 
+- Updated the logic to make sure logged_metrics are contain tensors ([#11270](https://github.com/PyTorchLightning/pytorch-lightning/pull/11270))
+
+
 ### Deprecated
 
 - Deprecated `ClusterEnvironment.master_{address,port}` in favor of `ClusterEnvironment.main_{address,port}` ([#10103](https://github.com/PyTorchLightning/pytorch-lightning/issues/10103))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,7 +179,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
 - `Trainer.logged_metrics` now always contains scalar tensors, even when a Python scalar was logged ([#11270](https://github.com/PyTorchLightning/pytorch-lightning/pull/11270))
->>>>>>> 260bfe93b (update chlog)
 
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,7 +178,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `DeviceStatsMonitor` to group metrics based on the logger's `group_separator` ([#11254](https://github.com/PyTorchLightning/pytorch-lightning/pull/11254))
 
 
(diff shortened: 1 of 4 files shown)


tshu-w:
Hi, thanks for the quick fix, just one question, is there any reason that log_metrics is saved as tensor instead of scalars.

rohitgr7:
> Hi, thanks for the quick fix, just one question, is there any reason that log_metrics is saved as tensor instead of scalars.

I think for consistency it would be good to have them as tensors since we have similar behavior with `callback_metrics` and is not very difficult to extract scalar values though.

CHANGELOG.md
```diff
@@ -175,6 +175,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Renamed `training_type_plugin` file to `strategy` ([#11239](https://github.com/PyTorchLightning/pytorch-lightning/pull/11239))
 
 
+- Updated the logic to make sure logged_metrics are contain tensors ([#11270](https://github.com/PyTorchLightning/pytorch-lightning/pull/11270))
```
awaelchli, line 178:
```suggestion
- `Trainer.logged_metrics` now always contains scalar tensors, even when a Python scalar was logged ([#11270](https://github.com/PyTorchLightning/pytorch-lightning/pull/11270))
```
is this description more clear?

awaelchli (approved):

CHANGELOG.md
rohitgr7: yep, thanks.

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -102,13 +102,14 @@ def log_metrics(self, metrics: _OUT_DICT, step: Optional[int] = None) -> None:
         if self.trainer.logger is None or not metrics:
             return
 
-        # turn all tensors to scalars
-        scalar_metrics = metrics_to_scalars(metrics)
+        if step is None and "step" in metrics:
```
carmocca, line 105: I don't get why the step changes. The step is supposed to be an int in all cases.

This should only change what we update with:

```diff
        # turn all tensors to scalars
        scalar_metrics = metrics_to_scalars(metrics)

        if step is None:
            step = scalar_metrics.pop("step", None)

-        self._logged_metrics.update(scalar_metrics)
+        self._logged_metrics.update(metrics)
```

rohitgr7: it's typed with optional[int], and we don't keep step inside logged_metrics so we need to extract it from metrics itself before updating the logged_metrics.
was getting the mypy error: https://github.com/PyTorchLightning/pytorch-lightning/runs/4645001501?check_suite_focus=true

carmocca: My suggested change does not output any mypy errors locally.

And we don't need to remove `step` from `metrics` as it would be there only if added by the user.

rohitgr7: okay.. will update. it will a minor change in the behavior then.

carmocca (approved):

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -108,7 +108,7 @@ def log_metrics(self, metrics: _OUT_DICT, step: Optional[int] = None) -> None:
         if step is None:
             step = scalar_metrics.pop("step", None)
 
-        self._logged_metrics.update(scalar_metrics)
+        self._logged_metrics.update(metrics)
```
tchaton, line 111: Let's move this before converting to metrics to scalar. The code is confusing.

justusschock (approved):

Issue #11248, opened 2021-12-24
Deprecate `TrainerDataLoadingMixin` and move logic to `DataConnector`


daniellepintz:
## Proposed refactor

Move everything from this class:
https://github.com/PyTorchLightning/pytorch-lightning/blob/ca9b25db80f08a3b9a3c448048949ec1adb845ba/pytorch_lightning/trainer/data_loading.py#L46

to this class:
https://github.com/PyTorchLightning/pytorch-lightning/blob/ca9b25db80f08a3b9a3c448048949ec1adb845ba/pytorch_lightning/trainer/connectors/data_connector.py#L36

### Motivation
The `TrainerDataLoadingMixin` is the last remaining Mixin after `TrainerCallbackHookMixin` and `TrainerOptimizersMixin` have been deprecated. As part of our move towards composition over inheritance, it makes sense to deprecate this last one as well and have no mixins on the Trainer class. 

cc @justusschock @awaelchli @akihironitta @rohitgr7 @ninginthecloud @tchaton

PR #11282, opened 2021-12-31
Deprecate `TrainerDataLoadingMixin` and move logic to `DataConnector`

daniellepintz:
## What does this PR do?
Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/11248

## Before submitting

## PR review

## Did you have fun?


kaushikb11 (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -646,7 +647,7 @@ def _setup_on_init(self, num_sanity_val_steps: int) -> None:
         self.should_stop = False
         self.state = TrainerState()
         self.num_training_batches = float("inf")
-        self.train_dataloader = None
+        self.train_dataloader: DataLoader = None
```
akihironitta, line 650: Should it be `Optional` because `None` is assigned to it?
```suggestion
        self.train_dataloader: Optional[DataLoader] = None
```

pytorch_lightning/trainer/data_loading.py
```diff
@@ -11,483 +11,137 @@
...
-        kwargs.setdefault("seed", int(os.getenv("PL_GLOBAL_SEED", 0)))
-        cls = UnrepeatedDistributedSampler if mode == RunningStage.PREDICTING else DistributedSampler
-        sampler = cls(dataloader.dataset, **kwargs)
-        return sampler
+        rank_zero_deprecation(
+            "`TrainerDataLoadingMixin.prepare_dataloader` was deprecated in v1.6 and will be removed in v1.8."
+        )
+        return self._data_connector._prepare_dataloader(dataloader, shuffle, mode)
 
     def reset_train_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
```
awaelchli, line 46: these should just be moved to the Trainer, right? 
`trainer.reset_x_dataloader()` will then call into the data connector. Otherwise users can't call the reset manually.

pl_examples/loop_examples/kfold.py
```diff
@@ -216,13 +216,13 @@ def on_load_checkpoint(self, state_dict: Dict) -> None:
         self.current_fold = state_dict["current_fold"]
 
     def _reset_fitting(self) -> None:
-        self.trainer.reset_train_dataloader()
```
rohitgr7, line 219: not that it's private, is there a way for the user to manually reset/load the dataloader?
One case I was using this one is to prepare an lr_scheduler in configure_optimizers since dataloaders aren't loaded at that point.

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -242,6 +264,441 @@ def attach_datamodule(
...
+            )
+
+    def _reset_train_val_dataloaders(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets train and val dataloaders if none are attached to the trainer.
+
+        The val dataloader must be initialized before training loop starts, as the training loop
+        inspects the val dataloader to determine whether to run the evaluation loop.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 635:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -242,6 +264,441 @@ def attach_datamodule(
...
+        if source.is_defined() and has_step:
+            self.trainer.num_test_batches, self.trainer.test_dataloaders = self._reset_eval_dataloader(
+                RunningStage.TESTING, model=pl_module
+            )
+
+    def _reset_predict_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the predict dataloader and determines the number of batches.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 619:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -242,6 +264,441 @@ def attach_datamodule(
...
+            )
+
+            # store epoch of dataloader reset for reload_dataloaders_every_n_epochs
+            self.trainer._last_val_dl_reload_epoch = self.trainer.current_epoch
+
+    def _reset_test_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the test dataloader and determines the number of batches.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 605:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -242,6 +264,441 @@ def attach_datamodule(
...
+
+                loader_num_batches.append(num_batches)
+
+        return loader_num_batches, dataloaders
+
+    def _reset_val_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the validation dataloader and determines the number of batches.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 588:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -242,6 +264,441 @@ def attach_datamodule(
...
+        cls = UnrepeatedDistributedSampler if mode == RunningStage.PREDICTING else DistributedSampler
+        sampler = cls(dataloader.dataset, **kwargs)
+        return sampler
+
+    def _reset_train_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the train dataloader and initialises required variables (number of batches, when to validate,
+        etc.).
+
+        Args:
+            model: The `LightningModule` if calling this outside of the trainer scope.
```
rohitgr7, line 416:
```suggestion
            model: The ``LightningModule`` if calling this outside of the trainer scope.
```

pl_examples/loop_examples/kfold.py
daniellepintz: Using grep.app I only found two cases of users using `reset_train/val/test/predict_dataloader`, so I thought it might be okay to deprecate and make private. But if we still want to users to have this functionality, I can move them to the Trainer

pytorch_lightning/trainer/data_loading.py
daniellepintz: Using grep.app I only found two cases of users using `reset_train/val/test/predict_dataloader`, so I thought it might be okay to deprecate and make private. But if we still want to users to have this functionality, I can move them to the Trainer

pytorch_lightning/trainer/trainer.py
daniellepintz: @akihironitta this leads to the mypy error 
`pytorch_lightning/loops/fit_loop.py:226: error: Argument 1 to "process_dataloader" of "Strategy" has incompatible type "Optional[DataLoader[Any]]"; expected "DataLoader[Any]"  [arg-type]`

Not sure if I should change the types in signature of `process_dataloader`, or just put this line back to: `self.train_dataloader: DataLoader = None`

pl_examples/loop_examples/kfold.py
rohitgr7: yeah! I think it's pretty useful.

pytorch_lightning/trainer/data_loading.py
daniellepintz: Done

pytorch_lightning/trainer/trainer.py
akihironitta: @daniellepintz `self.train_dataloader: DataLoader = None` is the wrong type for sure, but I think we can revert it here and handle it in #7037 because this PR is not about typing, and there are heaps of mypy errors in the file anyway (which is currently ignored as configured in `pyproject.toml`).

daniellepintz: sounds good! thanks

pytorch_lightning/trainer/trainer.py
```diff
@@ -2364,6 +2369,151 @@ def terminate_on_nan(self, val: bool) -> None:
         )
         self._terminate_on_nan = val  # : 212
 
+    def reset_train_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
```
rohitgr7, line 2372: how about moving the reload logic to data_connector and `trainer.reset_train_dataloader` just calls `trainer._data_connector._reset_train_dataloader`? not sure if it's a good idea, but just trying to avoid all the logic being kept inside the single Trainer module.

CHANGELOG.md
```diff
@@ -225,6 +225,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated `TrainerCallbackHookMixin` ([#11148](https://github.com/PyTorchLightning/pytorch-lightning/pull/11148))
 
 
+- Deprecated `TrainerDataLoadingMixin` and moved functionality to `DataConnector` ([#11282](https://github.com/PyTorchLightning/pytorch-lightning/pull/11282))
```
rohitgr7, line 228:
```suggestion
- Deprecated `TrainerDataLoadingMixin` and moved functionality to `Trainer` and `DataConnector` ([#11282](https://github.com/PyTorchLightning/pytorch-lightning/pull/11282))
```

pytorch_lightning/strategies/ipu.py
```diff
@@ -121,8 +121,9 @@ def setup(self, trainer: "pl.Trainer") -> None:
         # patch the dataloader creation function with the custom `poptorch.DataLoader`.
         # this violates the intended control flow for the plugins, but since this is experimental, we have chosen
         # to use the simpler solution before adding abstractions to override the `DataLoader` class
-        self._update_dataloader_original = pl.trainer.data_loading._update_dataloader
-        pl.trainer.data_loading._update_dataloader = self._convert_to_poptorch_loader
+        self._update_dataloader_original = pl.trainer.connectors.data_connector._update_dataloader
+
```
carmocca, line 125: nit: there wasn't a space here before

```suggestion
```

pytorch_lightning/trainer/trainer.py
carmocca: I'm not convinced by this. This annotation is wrong. The correct thing would be to add a check or assertion in the fit_loop.

#7037 is way too large so this annotation will likely get forgotten and can give us a false sense of security because mypy will no longer know it's optional.

tests/deprecated_api/test_remove_1-8.py
```diff
@@ -271,6 +272,22 @@ def test_v1_8_0_deprecated_training_type_plugin_property():
...
+    trainer.fit(model, datamodule=dm)
+
+    with pytest.deprecated_call(
+        match=r"`TrainerDataLoadingMixin.prepare_dataloader` was deprecated in v1.6 and will be removed in v1.8.",
+    ):
+        trainer.prepare_dataloader(dataloader=model.train_dataloader, shuffle=False)
+    with pytest.deprecated_call(
+        match=r"`TrainerDataLoadingMixin.request_dataloader` was deprecated in v1.6 and will be removed in v1.8.",
+    ):
+        trainer.request_dataloader(stage=RunningStage.TRAINING)
```
carmocca, line 288: offtopic comment: when this is removed we can get rid of the optional `model=...` argument from call_hook

pytorch_lightning/trainer/trainer.py
daniellepintz: I don't feel strongly either way. I probably have a slight preference for keeping it as it is now, since I feel like having a function that just calls the private function adds a bit of unnecessary abstraction/complexity. If others feel strongly we can change it

pytorch_lightning/trainer/trainer.py
daniellepintz: That makes sense, thanks. Updated

tests/deprecated_api/test_remove_1-8.py
daniellepintz: Yes! looking forward to it :)

pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -242,6 +263,289 @@ def attach_datamodule(
         if hasattr(datamodule, "data_pipeline"):
             model.data_pipeline = datamodule.data_pipeline
 
+    @property
```
carmocca, line 266: General python design is to put properties before methods. cc @awaelchli

pytorch_lightning/trainer/trainer.py
carmocca: I'd say it's fine as it is. I'd only do that if it aids in testing, but we don't have unit tests for each of these separate methods.

carmocca (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2364,6 +2369,151 @@ def terminate_on_nan(self, val: bool) -> None:
...
+        """Resets the train dataloader and initialises required variables (number of batches, when to validate,
+        etc.).
+
+        Args:
+            model: The `LightningModule` if calling this outside of the trainer scope.
```
rohitgr7, line 2377:
```suggestion
            model: The ``LightningModule`` if calling this outside of the trainer scope.
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2364,6 +2369,151 @@ def terminate_on_nan(self, val: bool) -> None:
...
+            )
+
+        # store epoch of dataloader reset for reload_dataloaders_every_n_epochs
+        self._last_train_dl_reload_epoch = self.current_epoch
+
+    def reset_val_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the validation dataloader and determines the number of batches.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 2464:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2364,6 +2369,151 @@ def terminate_on_nan(self, val: bool) -> None:
...
+            )
+
+            # store epoch of dataloader reset for reload_dataloaders_every_n_epochs
+            self._last_val_dl_reload_epoch = self.current_epoch
+
+    def reset_test_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the test dataloader and determines the number of batches.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 2481:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2364,6 +2369,151 @@ def terminate_on_nan(self, val: bool) -> None:
...
+        if source.is_defined() and has_step:
+            self.num_test_batches, self.test_dataloaders = self._data_connector._reset_eval_dataloader(
+                RunningStage.TESTING, model=pl_module
+            )
+
+    def reset_predict_dataloader(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets the predict dataloader and determines the number of batches.
+
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 2495:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2364,6 +2369,151 @@ def terminate_on_nan(self, val: bool) -> None:
...
+                RunningStage.PREDICTING, model=pl_module
+            )
+
+    def reset_train_val_dataloaders(self, model: Optional["pl.LightningModule"] = None) -> None:
+        """Resets train and val dataloaders if none are attached to the trainer.
+
+        The val dataloader must be initialized before training loop starts, as the training loop
+        inspects the val dataloader to determine whether to run the evaluation loop.
+        Args:
+            model: The `LightningModule` if called outside of the trainer scope.
```
rohitgr7, line 2510:
```suggestion
            model: The ``LightningModule`` if called outside of the trainer scope.
```

rohitgr7 (approved):
nit:

pytorch_lightning/trainer/trainer.py
awaelchli: It is fine because the trainer owns the dataloaders, so resetting is her responsibility. But this won't be the case in the future, pretty sure. Let's just please not mix properties and methods.

awaelchli (approved):

daniellepintz:
Ready to be merged!

pytorch_lightning/trainer/trainer.py
awaelchli: > n any case, let's just please not mix properties and methods like this.


@daniellepintz I believe my comment was missed here. Could you send a follow up PR? I believe reviewers did not catch this properly

daniellepintz: @awaelchli sorry I am confused by that line. Could you clarify which property you are talking about?

awaelchli: I mean the disorganization that this has caused. Previously, methods and properties were organized, but now new methods have been added and now it's mixed up:

method
method
property
property
new method
new method

It should be

method
method
new method
new method
property
property

given that Trainer wants to have all properties at the bottom.

daniellepintz: Ah I see good point. Will send a follow up PR

PR #11288, opened 2021-12-31
Avoid non-blocking GPU->CPU copies.

### LINKED ISSUE #11287: GPU prediction race condition results in BasePredictionWriter observing incorrect (zero) values
## 🐛 Bug

When writing prediction results using a trivial BasePredictionWriter subclass with write_interval "epoch", the final batch is written as zeros when the following conditions are met:

* GPU accelerator (haven't tested with other non-CPU accelerators)
* Sufficiently expensive model that the CUDA stream has not completed by the time the prediction writer is called
  * (This is why BoringModel is not appropriate to reproduce here.)

I believe that this is a race condition resulting from non-blocking copies from GPU to CPU without explicit CUDA synchronization. This results in CPU tensors being incorrectly observed as zeros before the GPU computation completes. Details below.

### To Reproduce

```
import argparse
import logging
import os

from torchvision.datasets import FakeData
from torchvision.models import resnet50
from torchvision.transforms import ToTensor

from pytorch_lightning.plugins import DDPSpawnPlugin
from pytorch_lightning.callbacks import BasePredictionWriter
import torch


logging.basicConfig(
    format='%(asctime)s %(levelname)-8s %(message)s',
    level=logging.INFO,
    datefmt='%Y-%m-%d %H:%M:%S')


class Model(pl.LightningModule):

    def __init__(self):
        super().__init__()
        # Arbitrary model, but this is a race, and the race needs the model to
        # be expensive enough to cause the CUDA stream to queue.
        self.model = resnet50(pretrained=False)

    def forward(self, x):
        x = self.model(x)
        # Return scalar values (easier to print), and ensure that 0s are not possible.
        # It's important that the result depends on the slow self.model(x) step.
        return (x.abs() + 0.1).sum(dim=1)

    def predict_step(self, batch, batch_idx, dataloader_idx=None):
        x, _ = batch
        y = self(x)
        return y


class InferenceWriter(BasePredictionWriter):
    """Writes predictions on epoch end."""

    def __init__(self, output_path: str):
        super().__init__("epoch")
        self.output_file = os.path.join(output_path, "predictions.pt")

    def write_on_epoch_end(
        self, trainer, module, predictions, batch_indices
    ):
        assert len(predictions) == 1
        predictions = predictions[0]
        outputs = torch.cat(predictions)
        logging.info("Saving output to %s.", self.output_file)
        torch.save(outputs, self.output_file)

    def read(self):
        return torch.load(self.output_file)


parser.add_argument("--output_path", required=True)
parser.add_argument("--batch_size", default=256, type=int)
parser.add_argument("--image_size", default=288, type=int, help="Image size for inference")


def main(args):
    dataset = FakeData(
        args.batch_size * 2,
        image_size=(3, args.image_size, args.image_size),
        transform=ToTensor(),
    )
    model = Model()
    dataloader = DataLoader(
        dataset,
        batch_size=args.batch_size,
        num_workers=2,
        persistent_workers=True,  # unnecessary here, but silences warning
    )
    writer = InferenceWriter(args.output_path)
    trainer = pl.Trainer(
        devices=1,
        num_nodes=1,
        accelerator="gpu",
        default_root_dir=args.output_path,
        strategy=DDPSpawnPlugin(find_unused_parameters=False),
        callbacks=[writer],
    )
    logging.info("Starting inference")
    trainer.predict(model, dataloaders=dataloader)
    logging.info("Loading features")
    outputs = writer.read()

    logging.info("First batch values: %s", outputs[:args.batch_size])
    logging.info("Second batch values: %s", outputs[args.batch_size:])


if __name__ == '__main__':
    args = parser.parse_args()
    main(args)
```

Output:
```
2021-12-31 05:27:38 INFO     First batch values: tensor([6145.3164, 6103.9824, 6139.4873, 6153.1626, 6133.7827, 6134.3999, ...
2021-12-31 05:27:39 INFO     Second batch values: tensor([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., ...
```

### Expected behavior

Reproduce script: Both the first and the second batch (logged in the script above) should contain nonzero values, from the same distribution. The computation is such that zeroes are not possible outputs.

In general: Tensors returned to the user on API surfaces should contain valid values when their contents are observed.

### Cause (I think)

If I understand the semantics of tensors copied from GPU to CPU with non_blocking=True, these tensors are unsafe without explicit CUDA synchronization to ensure that the CPU tensors have valid values (assuming the copy to CPU is covered by something like torch.cuda.synchronize()).

A non_blocking GPU -> CPU transfer creates a CPU tensor that is effectively a future, similar to CUDA tensors whose values have not resolved. But unlike CUDA tensors, the CPU tensor does not act like a future, and does not know that the value has not been written. As a result, operations that observe the tensor values do not wait for the tensor to be valid, and incorrect (zero) values are observed.

Since these futures are sharp edges, I suggest that we don't expose them to users unless we synchronize appropriately.

Similar bugs resulting from non-blocking GPU -> CPU copies have been reported elsewhere, for instance there are two similar reports on [this thread](https://discuss.pytorch.org/t/should-we-set-non-blocking-to-true/38234/17).

### Possible fixes

I think this could be done using cuda synchronization before callbacks that might see outputs that have been non-blocking moved to CPU. Calling `torch.cuda.synchronize()` before exposing affected tensors (CPU tensors from non-blocking copies from GPU operations) to the user might fix this.

However I think it would be better to not set non_blocking=True when copying tensors to CPU to avoid this case. It will be hard to find all the possible cases above. I don't see a strong argument to using non-blocking copies from GPU, and avoiding incorrect results due to the surprising behavior of these tensors seems like a compelling argument against. I expect that changing `move_data_to_device` to not set non_blocking when device is CPU should work.

### Fixes I've tested

Copying tensors to CPU in `predict_step`: `y = y.cpu()`. This uses a blocking transfer, making Lightning's non-blocking `.to()` call a no-op.

Inspecting the result of GPU operations also fixes this, by forcing us to wait for the GPU. This has to be done on the GPU tensors, before the CPU copy. (eg. `y.mean().item()` in `predict_step` in the reproduce example).

Using CPU compute also fixes this, but is slow, since the examples need enough computation to expose the race with GPUs.

Additionally, various things to "win" the race condition here also work:
* A simple sleep in write_on_epoch_end "fixes" this
* Using a smaller inference size (reducing GPU compute) "fixes" this

### Environment

* CUDA:
        - GPU:
                - A100-SXM4-40GB
                - A100-SXM4-40GB
                - A100-SXM4-40GB
                - A100-SXM4-40GB
                - A100-SXM4-40GB
                - A100-SXM4-40GB
                - A100-SXM4-40GB
                - A100-SXM4-40GB
        - available:         True
        - version:           11.2
* Packages:
        - numpy:             1.21.5
        - pyTorch_debug:     False
        - pyTorch_version:   1.10.0
        - pytorch-lightning: 1.5.7
        - tqdm:              4.62.3
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.9.9

cc @tchaton @rohitgr7

edpizzi:
Non-blocking GPU->CPU transfers can create race windows where tensor contents are observed to have incorrect values.

Fixes #11287. See #11287 for detailed mechanics of the issue this resolves.

Verified that this change fixes the behavior in the reproduce script from #11287.

No

## Before submitting

   - see discussion on the issue thread for challenges with contbuild tests for this. I did test that this resolves the reproduce script in #11287

## PR review

  - can't seem to do this, and the review guidelines wiki page no longer loads for me

## Did you have fun?

Fixing: yes. Root causing the bug: no. It's subtle, and many log statements you add to debug the issue fix the issue (by observing the value of GPU tensors, forcing a CUDA sync), which is frustrating if you don't know that this is the nature of the issue.

I also had some trouble getting tests to run in my environment. It looks to me like some tests fail in GPU environments but pass in non-GPU environments. Others fail when run on slurm environments, which is the only way I can easily access GPUs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -385,6 +385,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed data fetcher selection ([#11294](https://github.com/PyTorchLightning/pytorch-lightning/pull/11294))
 
 
+- Fixed a race that could result in incorrect (zero) values being observed in prediction writer callbacks. ([#11287](https://github.com/PyTorchLightning/pytorch-lightning/pull/11287))
+
+
 ## [1.5.7] - 2021-12-21
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -385,7 +385,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed data fetcher selection ([#11294](https://github.com/PyTorchLightning/pytorch-lightning/pull/11294))
 
 
-- Fixed a race that could result in incorrect (zero) values being observed in prediction writer callbacks. ([#11287](https://github.com/PyTorchLightning/pytorch-lightning/pull/11287))
+- Fixed a race condition that could result in incorrect (zero) values being observed in prediction writer callbacks ([#11288](https://github.com/PyTorchLightning/pytorch-lightning/pull/11288))
 
 
 ## [1.5.7] - 2021-12-21
--- a/pytorch_lightning/utilities/apply_func.py
+++ b/pytorch_lightning/utilities/apply_func.py
@@ -35,6 +35,9 @@ else:
     Batch = type(None)
 
 
+CPU_DEVICES = [None, "cpu", torch.device("cpu")]
+
+
 def to_dtype_tensor(
     value: Union[int, float, List[Union[int, float]]], dtype: torch.dtype, device: Union[str, torch.device]
 ) -> torch.Tensor:
@@ -274,7 +277,10 @@ def move_data_to_device(batch: Any, device: Union[str, torch.device]) -> Any:
                 setattr(device_data, field, device_field)
             return device_data
 
-        kwargs = dict(non_blocking=True) if isinstance(data, torch.Tensor) else {}
+        kwargs = {}
+        # Don't issue non-blocking transfers to CPU. #11287
+        if isinstance(data, torch.Tensor) and device not in CPU_DEVICES:
+            kwargs["non_blocking"] = True
         data_output = data.to(device, **kwargs)
         if data_output is not None:
             return data_output
--- a/pytorch_lightning/utilities/apply_func.py
+++ b/pytorch_lightning/utilities/apply_func.py
@@ -35,7 +35,7 @@ else:
     Batch = type(None)
 
 
-CPU_DEVICES = (None, "cpu", torch.device("cpu"))
+CPU_DEVICES = ("cpu", torch.device("cpu"))
 
 
 def to_dtype_tensor(
--- a/pytorch_lightning/utilities/apply_func.py
+++ b/pytorch_lightning/utilities/apply_func.py
@@ -35,7 +35,7 @@ else:
     Batch = type(None)
 
 
-CPU_DEVICES = ("cpu", torch.device("cpu"))
+_CPU_DEVICES = ("cpu", torch.device("cpu"))
 
 
 def to_dtype_tensor(
@@ -278,8 +278,8 @@ def move_data_to_device(batch: Any, device: Union[str, torch.device]) -> Any:
             return device_data
 
         kwargs = {}
-        # Don't issue non-blocking transfers to CPU. #11287
-        if isinstance(data, torch.Tensor) and device not in CPU_DEVICES:
+        # Don't issue non-blocking transfers to CPU
+        if isinstance(data, torch.Tensor) and device not in _CPU_DEVICES:
             kwargs["non_blocking"] = True
         data_output = data.to(device, **kwargs)
         if data_output is not None:
--- a/pytorch_lightning/utilities/apply_func.py
+++ b/pytorch_lightning/utilities/apply_func.py
@@ -35,7 +35,7 @@ else:
     Batch = type(None)
 
 
-CPU_DEVICES = [None, "cpu", torch.device("cpu")]
+CPU_DEVICES = (None, "cpu", torch.device("cpu"))
 
 
 def to_dtype_tensor(


pytorch_lightning/utilities/apply_func.py
```diff
@@ -35,6 +35,9 @@
     Batch = type(None)
 
 
+CPU_DEVICES = [None, "cpu", torch.device("cpu")]
```
justusschock, line 38:
```suggestion
CPU_DEVICES = (None, "cpu", torch.device("cpu"))
```

Constants should not be mutable :)

justusschock (approved):

CHANGELOG.md
```diff
@@ -371,6 +371,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Fixed dataloaders not getting reloaded the correct amount of times when setting `reload_dataloaders_every_n_epochs` and `check_val_every_n_epoch` ([#10948](https://github.com/PyTorchLightning/pytorch-lightning/pull/10948))
 
+- Fixed a race that could result in incorrect (zero) values being observed in prediction writer callbacks. ([#11287](https://github.com/PyTorchLightning/pytorch-lightning/pull/11287))
```
rohitgr7, line 374:
```suggestion
- Fixed a race that could result in incorrect (zero) values being observed in prediction writer callbacks. ([#11288](https://github.com/PyTorchLightning/pytorch-lightning/pull/11288))
```

rohitgr7 (approved):

Borda (approved):

pytorch_lightning/utilities/apply_func.py
```diff
@@ -35,6 +35,9 @@
     Batch = type(None)
 
 
+CPU_DEVICES = (None, "cpu", torch.device("cpu"))
```
awaelchli, line 38: the device can't be None, see the type annotation below.

awaelchli (approved):

pytorch_lightning/utilities/apply_func.py
edpizzi: My first attempt at fixing this was basically `torch.device(device) == torch.device("cpu")`, and it failed unit tests because we get None here, contrary to the type annotation.

That said, we can remove `None` here, since that case is a no-op (`torch.to(None, non_blocking=True)` is safe, and does nothing).

edpizzi: I've removed `None`, so I think this can be resolved.

pytorch_lightning/utilities/apply_func.py
```diff
@@ -35,6 +35,9 @@
     Batch = type(None)
 
 
+CPU_DEVICES = ("cpu", torch.device("cpu"))
```
carmocca, line 38: This can be protected

```suggestion
_CPU_DEVICES = ("cpu", torch.device("cpu"))
```

pytorch_lightning/utilities/apply_func.py
```diff
@@ -274,7 +277,10 @@ def batch_to(data: Any) -> Any:
                 setattr(device_data, field, device_field)
             return device_data
 
-        kwargs = dict(non_blocking=True) if isinstance(data, torch.Tensor) else {}
+        kwargs = {}
+        # Don't issue non-blocking transfers to CPU. #11287
+        if isinstance(data, torch.Tensor) and device not in CPU_DEVICES:
```
carmocca, line 282:
```suggestion
        if isinstance(data, torch.Tensor) and device not in _CPU_DEVICES:
```

pytorch_lightning/utilities/apply_func.py
```diff
@@ -274,7 +277,10 @@ def batch_to(data: Any) -> Any:
                 setattr(device_data, field, device_field)
             return device_data
 
-        kwargs = dict(non_blocking=True) if isinstance(data, torch.Tensor) else {}
+        kwargs = {}
+        # Don't issue non-blocking transfers to CPU. #11287
```
carmocca, line 281: There's git blame to find the associated PR number

```suggestion
        # Don't issue non-blocking transfers to CPU
```

carmocca (approved):

PR #11307, opened 2022-01-04
Fix lr scheduler state not being dumped to checkpoint in deepspeed strategy

### LINKED ISSUE #11188: lr scheduler state not saved with DeepSpeed plugin
## 🐛 Bug

When using deepspeed plugin, the learning rate scheduler state is not saved. The key `lr_scheduler` in `mp_rank_00_model_states.pt` model file is None. The bug causes the resuming from checkpoint to behave unexpected, where it should continue from previous lr scheduler state but instead lr scheduler start from scratch. This issue happens for most pytorch_lightning release versions, but the issue does not happen for some commits, e.g. commit 86c033572da084fd85d4d4c2d637db18d2d431ec.

### To Reproduce

```
import os
import torch
from pytorch_lightning.plugins import DeepSpeedPlugin
from deepspeed.ops.adam import FusedAdam
from transformers.optimization import AdamW, get_scheduler

class RandomDataset(Dataset):

        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

    def train_dataloader(self):
    
    def test_dataloader(self):

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return loss

        loss = self(batch).sum()
        self.log("valid_loss", loss)
        return loss

        loss = self(batch).sum()
        self.log("test_loss", loss)
        return loss

        no_decay = ["bias"]
        params_decay = [
            p for n, p in self.named_parameters() if not any(nd in n for nd in no_decay)
        ]
        params_nodecay = [
            p for n, p in self.named_parameters() if any(nd in n for nd in no_decay)
        ]
        optim_groups = [
            {
                "params": params_decay,
                "weight_decay": 0.01,
            },
            {"params": params_nodecay, "weight_decay": 0.0},
        ]
        optimizer = torch.optim.SGD(optim_groups, lr=0.1)
        scheduler = get_scheduler("linear", optimizer, num_warmup_steps=20, num_training_steps=200)
        return ([optimizer], [{"scheduler": scheduler, "interval": "step", "frequency": 1}])

def run():
    train_data = DataLoader(RandomDataset(32, 64), batch_size=4)
    val_data = DataLoader(RandomDataset(32, 64), batch_size=4)
    test_data = DataLoader(RandomDataset(32, 64), batch_size=4)

    model = BoringModel()
    checkpoint_callback = ModelCheckpoint(
        dirpath='tests4/checkpoints',
        save_last=True,
        every_n_train_steps=10,
    )
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        gpus=-1,
        limit_train_batches=1.0,
        limit_val_batches=1,
        num_sanity_val_steps=0,
        precision=16,
        accelerator='deepspeed',
        max_steps=100,
        log_every_n_steps=1,
        plugins=[DeepSpeedPlugin(stage=2)],
        weights_summary=None,
        callbacks=[checkpoint_callback, pl.callbacks.LearningRateMonitor(logging_interval="step")],
        #resume_from_checkpoint='tests4/checkpoints/epoch=3-step=99.ckpt',
    )
    trainer.fit(model)
    trainer.test(model)

    # test lr_scheduler state
    model_state_path = "tests4/checkpoints/epoch=3-step=99.ckpt/global_step100/mp_rank_00_model_states.pt"
    model_state = torch.load(model_state_path, map_location="cpu")
    print(model_state["lr_schedulers"])


if __name__ == '__main__':
    run()
```

Just run `python run.py`. For latest release version 1.5.6 and many other versions including current master branch, it produces None, meaning lr_scheduler state is not saved. For commit id 86c033572da084fd85d4d4c2d637db18d2d431ec, it produces the correct state:
```
{'base_lrs': [0.1, 0.1], 'last_epoch': 47, '_step_count': 48, 'verbose': False, '_get_lr_called_within_step': False, '_last_lr': [0.05888888888888889, 0.05888888888888889], 'lr_lambdas': [None, None]}
```

Do first run with the code exactly as above. Do second run by changing max_steps from 100 to 200, and uncomment `#resume_from_checkpoint` line. The two runs gives the tensorboard curves of resuming training. For pytorch lightning version 1.5.6 and many others, it produces wrong lr curve:
Screen Shot 2021-12-20 at 5 19 20 PM

However, the correct resuming curve (like the one produced with commit id 86c033572da084fd85d4d4c2d637db18d2d431ec) should be
Screen Shot 2021-12-20 at 5 16 57 PM

### Expected behavior

The lr_scheduler state should be saved properly in order to resume from checkpoint correctly. Could you clarify if we should handle lr_scheduler state loading or saving by ourselves, or it should be properly handled by pytorch_lightning? If it should be handled by pytorch_lightning, do you know how to resolve the bug?

cc @tchaton @SeanNaren @awaelchli

awaelchli:
## What does this PR do?

Part of #11188

The added assertion fails on master.

## Before submitting

## PR review

## Did you have fun?

cc @borda @SeanNaren @awaelchli @rohitgr7


SeanNaren (approved):
Thanks for fixing this @awaelchli. I assume based on the logic, the scheduler will be loaded if found in the state dict elsewhere in the code

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -741,7 +741,7 @@ def save_checkpoint(self, checkpoint: Dict, filepath: _PATH) -> None:
             )
         # Use deepspeed's internal checkpointing function to handle partitioned weights across processes
         # dump states as a checkpoint dictionary object
-        _exclude_keys = ["state_dict", "optimizer_states", "lr_schedulers"]
+        _exclude_keys = ["state_dict", "optimizer_states"]
```
carmocca, line 744: If the `lr_schedulers` weren't getting saved, why didn't this test fail? It checks that the `lr_schedulers` key exists in the saved checkpoint

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/models/test_hooks.py#L496

carmocca (approved):

justusschock (approved):

rohitgr7 (approved):

pytorch_lightning/strategies/deepspeed.py
awaelchli: Ah, I didn't know this hook test uses deepspeed. Thanks! 

It's because the entry is a list and it is empty if no scheduler is defined/saved. And so this matches ANY.

awaelchli: and `on_save_checkpoint` where the test asserts the contents gets called before the `save_checkpoint` call (where deepspeed dropped the key)

tridao:
> Thanks for fixing this @awaelchli. I assume based on the logic, the scheduler will be loaded if found in the state dict elsewhere in the code

I'm not sure that's the case, since the Deepspeed plugin has `lightning_restore_optimizer_and_schedulers` set to False.
Would the [checkpoint_connector](https://github.com/PyTorchLightning/pytorch-lightning/blob/7eab379da2fdca542849ed4ad313d0851c2271e3/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L261) ignore the lr_schedulers key when restoring?
```
def restore_optimizers_and_schedulers(self) -> None:
        """Restores the optimizers and learning rate scheduler states from the pre-loaded checkpoint."""
        if not self._loaded_checkpoint or not self.trainer.strategy.lightning_restore_optimizer_and_schedulers:
            return
```

awaelchli:
Yes you are right, this needs to be taken care of too.

eelxpeng:
Right. I tried with current commits. Though `lr_schedulers` are saved, the `lr_scheduler` state will not be restored.

Issue #11205, opened 2021-12-21
Adopt PEP 585, PEP 604, and non-quoted annotations


carmocca:
## Proposed refactor

We've dropped support for Python 3.6, which means we can use the new annotations format.

### Motivation

New shiny things are nice

### Pitch

Adopt [PEP 585](PEP 585), [PEP 604](https://www.python.org/dev/peps/pep-0604/), and non-quoted annotations.

We can do this automatically with the following command:

```console
isort -a "from __future__ import annotations" --append-only pytorch_lightning; pre-commit run pyupgrade --all-files; autoflake -ir --remove-unused-variables pytorch_lightning; pre-commit run black --all-files
```

### Additional context

There's a bit of discussion of replacing [PEP 563](https://www.python.org/dev/peps/pep-0563/) with [PEP 649](https://www.python.org/dev/peps/pep-0649/)

I don't think it impacts our project, but depending on the resolution of the latter, the import statement might need to change.

Also, we might want to delay this until 1.6 will be released, to avoid conflicts with the bug-fix branch.

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

carmocca:
Looks like this would break the `LightningCLI` so will have to wait for a while

mauvilsa:
[...]

PR #11316, opened 2022-01-04
Add typing to some utility files

carmocca:
## What does this PR do?

Part of #7037
Part of #11205

I'm going to try and always have one typing PR open for reviews :)

None

## Before submitting

## PR review

cc @borda

### DIFF
--- a/pytorch_lightning/utilities/deepspeed.py
+++ b/pytorch_lightning/utilities/deepspeed.py
@@ -32,12 +32,14 @@ Example usage within the Lightning checkpoint directory where 'latest' is found:
 >>> convert_zero_checkpoint_to_fp32_state_dict(save_path, output_path) # doctest: +SKIP
 Saving fp32 state dict to lightning_model.pt
 """
+from __future__ import annotations
 
 import os
 
 import torch
 
 from pytorch_lightning.utilities import _DEEPSPEED_AVAILABLE
+from pytorch_lightning.utilities.types import _PATH
 
 if _DEEPSPEED_AVAILABLE:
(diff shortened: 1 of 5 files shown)


pytorch_lightning/utilities/deepspeed.py
```diff
@@ -49,7 +51,7 @@
 CPU_DEVICE = torch.device("cpu")
 
 
-def ds_checkpoint_dir(checkpoint_dir: str, tag: str = None):
+def ds_checkpoint_dir(checkpoint_dir: _PATH, tag: str | None = None) -> str:
```
ananthsub, line 54: n00b q: is this the new convention in favor of `Optional[str]` ? if so, should we seek to migrate these all at once to avoid having multiple conventions existing in the codebase at the same time?

carmocca: Yes, see [PEP 604](https://www.python.org/dev/peps/pep-0604/)

pre-commit automatically does the update if the file has `from __future__ import annotations`.

> should we seek to migrate these all at once

Yes. This is the plan described in #11205, but won't do it until the bug-fix releases stop since this would be a huge source of conflicts.

Since these are small files, I think it's fine to update them already, less work for later.

> avoid having multiple conventions existing in the codebase at the same time?

This is not a big problem because everything will be updated automatically later.
I personally won't mention this in code review for any new code unless the author wants to use this new notation directly.

justusschock (approved):

awaelchli (approved):

Borda (approved):

PR #11332, opened 2022-01-05
Update print_results

### LINKED ISSUE #11139: Improve the evaluation results verbose look
## 🚀 Feature

When evaluation `trainer.validate(verbose=True)` (or `test`) finishes, we print a dictionary with the results obtained

```python
--------------------------------------------------------------------------------
DATALOADER:0 TEST RESULTS
{'test_loss': -3.4134674072265625}
--------------------------------------------------------------------------------
```

https://github.com/PyTorchLightning/pytorch-lightning/blob/860959fb3fc9d54c3d65984d64bfa391f68bc51b/pytorch_lightning/loops/dataloader/evaluation_loop.py#L275-L288

### Pitch

Propose and implement a new and nicer format. For example, you can check if https://github.com/willmcgugan/rich is installed and then use it.

Please, comment in this issue first about what's your idea is before jumping on a PR.

### Motivation

The current output is very barebones, both ugly and not easy to parse.

### Alternatives

Don't do it

cc @borda @awaelchli

AndresAlgaba:
## What does this PR do?
Fixes #11139 

### Possible follow-ups:

> Should we get the console from the rich progress bar if available as it would contain optionally a user thema?

> I wonder if we could merge in ModelSummary, simple profiler, and this all in one and create a single utility that builds the table for us. We just provide the headers and data and it returns the table. Not in this one. In a follow-up, if approved.

None

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +273,59 @@ def _on_evaluation_epoch_end(self) -> None:
...
+
+                for j, metric in enumerate(rows):
+                    v = metrics_dict.get(metric[0])
+                    v = (v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v
+                    rows[j].append(f"{v}")
+
+            for row in rows:
+                table.add_row(*row)
+
+            console.print("", table)
```
carmocca, line 306: Is this empty space on purpose?

```suggestion
            console.print(table)
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +273,59 @@ def _on_evaluation_epoch_end(self) -> None:
...
+            for i, metrics_dict in enumerate(new_results):
+                for j in range(len(rows)):
+                    v = metrics_dict.get(unique_keys[j])
+                    v = (v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v
+                    rows[j].append(f"{v}")
+
+            cols = [f"DATALOADER {i}" for i in range(len(results))]
+
+            row_format = "{:^25}" * (len(cols) + 1)
+            print("\n", row_format.format("", *cols))
```
carmocca, line 328: Same comment here

```suggestion
            print(row_format.format("", *cols))
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +273,59 @@ def _on_evaluation_epoch_end(self) -> None:
...
+
+            rows = [[] for x in range(len(unique_keys))]
+
+            for i, metrics_dict in enumerate(new_results):
+                for j in range(len(rows)):
+                    v = metrics_dict.get(unique_keys[j])
+                    v = (v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v
+                    rows[j].append(f"{v}")
+
+            cols = [f"DATALOADER {i}" for i in range(len(results))]
```
carmocca, line 325: This is missing the "Metric" column name

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +273,59 @@ def _on_evaluation_epoch_end(self) -> None:
...
+
+            row_format = "{:^25}" * (len(cols) + 1)
+            print("\n", row_format.format("", *cols))
+
+            for col, row in zip(unique_keys, rows):
```
carmocca, line 330: For the native version, let's keep the before and after separators

```python
-------------------------------------------------------------------------
                               DATALOADER 0             DATALOADER 1                                                                                                                                                                                                                                                                                                               
           foo               1.028334140777588       -0.9873988032341003   
       valid_loss         -0.0073436894454061985    0.018566807731986046  
------------------------------------------------------------------------- 
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +273,59 @@ def _on_evaluation_epoch_end(self) -> None:
...
+
+            for i, metrics_dict in enumerate(new_results):
+                for j in range(len(rows)):
+                    v = metrics_dict.get(unique_keys[j])
+                    v = (v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v
+                    rows[j].append(f"{v}")
+
+            cols = [f"DATALOADER {i}" for i in range(len(results))]
+
+            row_format = "{:^25}" * (len(cols) + 1)
```
carmocca, line 327: Do you think we could improve this to support long metric names?

Current output is:
```python
                               DATALOADER 0             DATALOADER 1                                                                                                                                                                                                                                                                                                               
a really really really long metric name   0.20651429891586304      0.4664147198200226    
           foo              0.10347846895456314      0.13967294991016388   
       valid_loss           -0.1181236281991005       2.235642671585083    
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +273,59 @@ def _on_evaluation_epoch_end(self) -> None:
         self.trainer.logger_connector.on_epoch_end()
 
     def _print_results(self, results: List[_OUT_DICT], stage: RunningStage) -> None:
```
carmocca, line 275: Perhaps not yet since you are still making changes, but this will need some tests.

carmocca (commented):
This is looking awesome!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: The empty line is intentional as otherwise the table or horizontal line gets messed up.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Yes, again, the empty line is intentional as otherwise the table or horizontal line gets messed up.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Added it in the latest commit!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Ok, added those again!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I've made the column width adaptable in the native version (in rich table this is done automatically) with a minimum width of 25.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Yes, agree! What kind of tests did you have in mind and where should I add them? For now, I am still fixing some mypy issues, and thinking about a fix for when multiple metrics are logged under a single name, e.g., 

```python
self.log("performance", {"acc": loss, "recall": loss})
```

because right now the output looks like

```
┏━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃   Metric    ┃    DATALOADER 0     ┃                    DATALOADER 1                     ┃
┡━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ performance │        None         │ {'acc': tensor(-0.5408), 'recall': tensor(-0.5408)} │
│  test_acc   │ -0.8093445301055908 │                 -0.5408046841621399                 │
│  test_loss  │ -0.8093445301055908 │                        None                         │
│   test_x    │ -0.8093445301055908 │                        None                         │
│   test_y    │        None         │                 -0.5408046841621399                 │
│   test_z    │        None         │                 -0.5408046841621399                 │
└─────────────┴─────────────────────┴─────────────────────────────────────────────────────┘
```

Any suggestions here?

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +272,67 @@ def _on_evaluation_epoch_end(self) -> None:
...
-            print(f"DATALOADER:{i} {stage.upper()} RESULTS")
-            pprint(
-                {
-                    k: (v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v
-                    for k, v in metrics_dict.items()
-                }
-            )
-            print("-" * 80)
+        if _RICH_AVAILABLE:
+            from rich.console import Console
```
tchaton, line 276: Let's move those methods into utlity.

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -272,16 +272,67 @@ def _on_evaluation_epoch_end(self) -> None:
...
+            from rich.table import Table
+
+            console = Console()
```
tchaton, line 279: Should we get the console from the rich progress bar if available as it would contain optionally a user thema?

pytorch_lightning/loops/dataloader/evaluation_loop.py
tchaton: I think it would be better as follow.

```py
-------------------------------------------------------------------------                                                                                                                                                                                                                                                                                                          
         Metric              DATALOADER 0             DATALOADER 1     
-------------------------------------------------------------------------                                                                                                                                                                                                                                                                                                          
           foo               1.028334140777588       -0.9873988032341003   
       valid_loss         -0.0073436894454061985    0.018566807731986046  
-------------------------------------------------------------------------                                                                                                                                                                                                                                                                                                          
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
tchaton: How does it look if you have 10 dataloaders ?

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I agree and have added the additional line in my latest commit.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: In the latest commit I have modified this bit. Now, in the case of a nested dict, I remove the "performance" as a metric and replace it with all the metrics which it contained (i.e. "acc" and "recall" in the example above). Instead of None I also put a blank right now. Do you agree with these choices?

AndresAlgaba: > How does it look if you have 10 dataloaders ?

I have tried this for both the rich and native implementation. For rich, the columns get adjusted to be smaller automatically (until they become unreadable). I will look into whether they can start on a new line. For native, they get really messed up once they do not longer fit the screen, so I will definitely need to fix this.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: What do you mean exactly with this?

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I believe this is a nice extension!

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -271,17 +271,86 @@ def _on_evaluation_epoch_end(self) -> None:
...
+            for row in rows:
+                table.add_row(*row)
+
+            console.print("", table)
+
+        else:
+            import os
+
+            term_size = os.get_terminal_size()
+            print("\n", "\u2500" * (term_size.columns - 1))
```
carmocca, line 328: In another conversation, you said about the leading newline:

> Yes, again, the empty line is intentional as otherwise the table or horizontal line gets messed up.

Can you share the output?

For me locally, with the newline I see:

```python
Testing: 100%|███████████████████████████████████████| 1/1 [00:00<00:00, 59.16it/s]

X──────────────────────────────────────────────────────────────────────────────────
         Metric                DATALOADER 0       
──────────────────────────────────────────────────────────────────────────────────X
        test_loss           -0.3928866386413574   
───────────────────────────────────────────────────────────────────────────────────
```

Where I've added "X"s to indicate spaces that shouldn't be there

And without:

```python
Testing: 100%|████████████████████████████████████████| 1/1 [00:00<00:00, 85.22it/s]
───────────────────────────────────────────────────────────────────────────────────
         Metric                DATALOADER 0       
───────────────────────────────────────────────────────────────────────────────────
        test_loss           0.1391649842262268    
────────────────────────────────────────────────────────────────────────────────────
```

So still, I suggest

```suggestion
            print("\u2500" * (term_size.columns - 1))
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -271,17 +271,86 @@ def _on_evaluation_epoch_end(self) -> None:
...
+
+            for row in rows:
+                table.add_row(*row)
+
+            console.print("", table)
+
+        else:
+            import os
+
+            term_size = os.get_terminal_size()
```
carmocca, line 327: As a 4k screen user, this seems way too much, especially on pychar, where the terminal is anchored to the bottom.

I think something like max(120, actual values width) would be a better choice

120 because because our code is formatted to 120 columns max. Could also be 80 or 88

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: He means that after we are set on the design, we should refactor as much unto utility methods we use to test, and move them out of this file.

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -271,17 +271,86 @@ def _on_evaluation_epoch_end(self) -> None:
         self.trainer._call_lightning_module_hook("on_epoch_end")
         self.trainer.logger_connector.on_epoch_end()
 
+    def _get_keys(self, iterable: Union[List, dict]) -> Iterable[str]:
```
carmocca, line 274: You might find this function interesting to apply a function to a collection, filtering by type.

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/utilities/apply_func.py#L71

I didn't look to deep into the actual implementation yet but looks like you could use it instead of these 2 you wrote here

tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -769,4 +769,4 @@ def test_dataloader(self):
     trainer = Trainer(default_root_dir=tmpdir, fast_dev_run=1)
     results = trainer.test(model)
     # what's logged in `test_epoch_end` gets included in the results of each dataloader
-    assert results == [{"foo/dataloader_idx_0": 1, "foobar": 3}, {"foo/dataloader_idx_1": 2, "foobar": 3}]
```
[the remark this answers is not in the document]
AndresAlgaba: Okay, in my latest commit I've created a new object to prevent introducing a breaking change!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I always get an expected result when printing in the ```trainer.test()``` (without the need for additional spacing), but not when I use ```trainer.validate(verbose=True)```.

When I remove the space in the rich implementation (i.e. ```console.print(table)``` instead of ```console.print("", table)```), I get the following result (scroll also to the right to see that the line actually starts at the end of the previous line):
```
                                                                                                                            ┏
━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━┓
┃   Metric   ┃    DATALOADER 0     ┃
┡━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━┩
│ valid_acc  │ -1.5097321271896362 │
│ valid_loss │ -1.5097321271896362 │
│  valid_x   │ -1.5097321271896362 │
└────────────┴─────────────────────┘
```

And for the native implementation, when I remove the space (i.e. ```print("\u2500" * term_size)``` instead of ```print("\n", "\u2500" * (term_size.columns - 1))```), I get the following result:
```
                                                                                                                                                                      ─
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
         Metric                DATALOADER 0
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
        valid_acc           0.5556910634040833
       valid_loss           0.5556910634040833
         valid_x            0.5556910634040833
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
Note that in the native implementation to account for the additional jump, I have to use ```term_size - 1```.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I've added the max 120 in my latest commit to have a minimum width of 120 if this is what was intended?

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Will take a look at this!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I've removed the spacing, for ```trainer.test(()``` this is completely fine, however for ```trainer.validate(verbose=True)``` I get the unexpected results as shown above. Maybe you can check whether you get something similar?

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I've given it a first try as follows:
```
apply_to_collection(new_results, Dict, dict.keys)
```
to get the keys, and:
```
apply_to_collection(metrics, Dict, dict.get, metric[0])
```
to get the values, but there seems to be a problem with the recursion as in the following example:
```
self.log("performance", {"acc": loss, "recall": loss})
```
I don't get 'acc' and 'recall' back, but only 'performance' via dict.keys, and the values also do not get returned via dict.get.

I will look further into this, but if you have any ideas, please share them!

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: Sorry! This is what I meant: f78f4bb

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -271,17 +271,80 @@ def _on_evaluation_epoch_end(self) -> None:
...
+                    v = list(self._find_value(metrics, unique_keys[j]))
+                    rows[j].append(f"{v[0]}" if v else " ")
+
+            cols = [f"{stage.capitalize()} Metric"] + [f"DataLoader {i}" for i in range(len(results))]
+            max_length = max(len(max(unique_keys + cols, key=len)), 25)
+            row_format = "{:^{max_length}}" * len(cols)
+            # cap wide terminals to 120 columns, but go over if there are many columns
+            term_size = max(min(120, os.get_terminal_size().columns), max_length * len(cols))
+
+            print("\u2500" * term_size)
```
carmocca, line 345: Is there a downside to using the symbol instead of the unicode directly?

Did you use unicode to clarify that it's not a normal dash?

```suggestion
            print("─" * term_size)
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: I cannot reproduce. Looks perfect to me as it is currently for both validate and test.

Resolving, but tests should check this.

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: Let's get started on the tests now that the terminal design is basically finalized.

For the tests, you can mock `sys.stdout` and then call `_print_results` with custom inputs and validate that the `stdout` text is the expected.

```python    
from contextlib import redirect_stdout

out = StringIO()
with redirect_stdout(out):
    loop._print_results(...)

expected = ...
assert out.getvalue() == expected
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -271,17 +271,80 @@ def _on_evaluation_epoch_end(self) -> None:
...
+                    yield k
+
+    def _find_value(self, data: dict, target: str) -> Iterable[Any]:
+        for k, v in data.items():
+            if k == target:
+                yield v
+            elif isinstance(v, dict):
+                yield from self._find_value(v, target)
+
     def _print_results(self, results: List[_OUT_DICT], stage: RunningStage) -> None:
```
carmocca, line 292: Let's make it a staticmethod so we don't need to instantiate the class to run it during testing

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -271,17 +271,65 @@ def _on_evaluation_epoch_end(self) -> None:
...
+            row_format = "{:^{max_length}}" * len(cols)
+            # cap wide terminals to 120 columns, but go over if there are many table columns
+            term_size = max(min(120, os.get_terminal_size().columns), max_length * len(cols))
+
+            print("\u2500" * term_size)
+            print(row_format.format(*cols, max_length=max_length))
+            print("\u2500" * term_size)
+            for col, row in zip(unique_keys, rows):
+                print(row_format.format(col, *row, max_length=max_length))
+            print("\u2500" * term_size)
```
carmocca, line 336: Should we construct the full string and then do a single `print` call with it?

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Ok, done!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: No downside on using the symbol, indeed it was for clarification but will commit the suggestion.

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: You missed `_print_results` itself :P

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Thanks for the template, will start on it :)

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Oh I see, fixed it!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I've added a first test in my latest commit, will continue to work on this.

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: I believe that is a cleaner solution, yes!

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: > > How does it look if you have 10 dataloaders ?
> 
> I have tried this for both the rich and native implementation. For rich, the columns get adjusted to be smaller automatically (until they become unreadable). I will look into whether they can start on a new line. For native, they get really messed up once they do not longer fit the screen, so I will definitely need to fix this.

In my latest commit I added functionality for a big native table. For rich, I'm not sure whether it's possible to 'overflow' the table, see: https://rich.readthedocs.io/en/stable/console.html#overflow and https://github.com/Textualize/rich/issues/1357.

AndresAlgaba: Not sure why the new tests are failing now, as I believe I have not changed anything which should affect the tests in my latest commit and they passed earlier.

carmocca: We had some CI issues that should be fixed in master. I merged master into your branch, let's see...

AndresAlgaba: Still fails, will take a look at it!

AndresAlgaba: Was just a small fix of what was expected :) Will take a further look at rich (with 10+ dataloaders) and additional testingv (also mainly for rich). If there are some "exotic" metrics of which you can think, please mention them here? Also should tests be created for the two static methods _get_keys and _find_values? The apply to collection doesnt seem to work (https://github.com/PyTorchLightning/pytorch-lightning/pull/11332#discussion_r780672963).

AndresAlgaba: Added logic for large rich tables, I feel like we may shorten the code by not checking whether the table is larger than the terminal. However, then we need to make some decisions on whether we want to add empty cells in the table, and whether to always use the terminal size to draw horizontal lines,... What do you think?

carmocca: > Also should tests be created for the two static methods _get_keys and _find_values? The apply to collection doesnt seem to work (#11332 (comment)).

No need, I'll take a look when reviewing.

> not sure whether this is the best way to suggest alternatives within PRs?

You can comment on your own PR to create separate discussions.

If you want to discuss alternatives, include examples of how it would look like.

carmocca:
The current output looks pretty good, what other design changes do you want to do?

This is the script I'm using to check

```python
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def validation_step(self, batch, batch_idx, dataloader_idx=0):
        loss = self(batch).sum()
        self.log("valid_loss", loss)
        self.log("performance", {"acc": loss, "recall": loss})
        self.log("a really long metric name", loss)
        self.log("a really really really really really really long metric name", loss)
        self.log("a really really really really really really really really really really really long metric name", loss)

    def test_step(self, batch, batch_idx, dataloader_idx=0):
        loss = self(batch).sum()
        self.log("valid_loss", loss)
        self.log("performance", {"acc": loss, "recall": loss})
        self.log("a really long metric name", loss)
        self.log("a really really really really really really long metric name", loss)
        self.log("a really really really really really really really really really really really long metric name", loss)

def run():
    test_data = [DataLoader(RandomDataset(32, 64), batch_size=2) for _ in range(5)]

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        #limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        enable_model_summary=False,
    )
    import pytorch_lightning.loops.dataloader.evaluation_loop as imports
    imports._RICH_AVAILABLE = False
    trainer.validate(model, dataloaders=test_data)


if __name__ == "__main__":
    run()
```

pytorch_lightning/loops/dataloader/evaluation_loop.py
AndresAlgaba: Okay, thanks!

AndresAlgaba:
Thanks for sharing the script you're using to check! I've added some functionality to account for smaller terminals. Will take a final look at the testing of the console output, especially for rich. Also, this: ```(v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v``` used to be in the initial code, but I'm not sure whether this is still relevant as logged tensors seem to be squashed into a single number? I cannot think of any further changes, do you have any suggestions or should I mark the PR as ready for review after finishing this stuff?

carmocca:
> but I'm not sure whether this is still relevant as logged tensors seem to be squashed into a single number?

Not necessarily:

```python
f"{torch.tensor([1])}"
Out[23]: 'tensor([1])'
f"{torch.tensor(1)}"
Out[24]: '1'
```
where both have 1 element `numel() == 1`

> should I mark the PR as ready for review after finishing this stuff?

Sure!

AndresAlgaba:
> > but I'm not sure whether this is still relevant as logged tensors seem to be squashed into a single number?
> 
> Not necessarily:
> 
> ```python
> f"{torch.tensor([1])}"
> Out[23]: 'tensor([1])'
> f"{torch.tensor(1)}"
> Out[24]: '1'
> ```
> 
> where both have 1 element `numel() == 1`
> 
> > should I mark the PR as ready for review after finishing this stuff?
> 
> Sure!

I have added some additional tests, maybe they should be merged into one/two tests, or should we test everything separately? 
I have also added this again: ```(v.item() if v.numel() == 1 else v.tolist()) if isinstance(v, torch.Tensor) else v```.
For rich testing, I was not able to find a good solution though... Looking into it I found: https://rich.readthedocs.io/en/stable/console.html#capturing-output and it seems like we may have to return the console output from ```_print_results``` to be able to test it?

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: Done

carmocca:
> For rich testing, I was not able to find a good solution though... Looking into it I found: https://rich.readthedocs.io/en/stable/console.html#capturing-output and it seems like we may have to return the console output from _print_results to be able to test it?

I used your suggestion here to expose the `file` argument and simplify testing. I added the rich tests too!

carmocca (approved):
Really nice. I think this is basically finished. We can see what other reviewers have to say

tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -770,3 +773,155 @@ def test_dataloader(self):
...
+"""
+
+
+@pytest.mark.parametrize(
+    ["inputs", "expected"],
+    [
+        pytest.param(inputs0, expected0, id="case0"),
+        pytest.param(inputs1, expected1, id="case1"),
+        pytest.param(inputs2, expected2, id="case2"),
+        pytest.param(inputs3, expected3, id="case3"),
```
akihironitta, line 849: Just pushed a commit naming these test cases to suppress the raw inputs in the test log:
```
tests/trainer/logging_/test_eval_loop_logging.py::test_native_print_results[inputs0-\n\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\n       Test metric             DataLoader 0             DataLoader 1\n\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\n           log                       5\n         no_log                                               6\n\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\u2500\n] PASSED [ 74%]
```

akihironitta (approved):
Tried this enhancement, and it looks very nice!
image

tests/trainer/logging_/test_eval_loop_logging.py
carmocca: Good shout

awaelchli (approved):

tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -770,3 +773,155 @@ def test_dataloader(self):
...
+    ],
+)
+def test_native_print_results(monkeypatch, inputs, expected):
```
rohitgr7, line 852: looks really cool!

I wonder if we could merge in ModelSummary, simple profiler, and this all in one and create a single utility that builds the table for us. We just provide the headers and data and it returns the table. Not in this one. In a follow-up, if approved.

rohitgr7 (approved):

tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -770,3 +773,155 @@ def test_dataloader(self):
...
+@pytest.mark.parametrize(
+    ["inputs", "expected"],
+    [
+        pytest.param(inputs0, expected0, id="case0"),
+        pytest.param(inputs1, expected1, id="case1"),
+        pytest.param(inputs2, expected2, id="case2"),
+        pytest.param(inputs3, expected3, id="case3"),
+    ],
+)
+@RunIf(rich=True)
```
carmocca, line 922: @AndresAlgaba or @akihironitta, can you look into the Windows failures?
We might need to replace a character with another in the tests or just skip them for windows.

AndresAlgaba: I believe the problem is in the "\r\n vs \n newlines in Python on Mac vs Windows": https://stackoverflow.com/questions/4599936/handling-r-n-vs-n-newlines-in-python-on-mac-vs-windows. What is the preferred solution in this case?

carmocca: Would doing

`assert out.getvalue().replace(os.linesep, "\n") == expected.lstrip()`

work?

AndresAlgaba: Cool! I believe this works. Pushed a commit :)

carmocca: `rich` tests need it too

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -273,17 +281,81 @@ def _on_evaluation_epoch_end(self) -> None:
...
+                if v:
+                    val = v[0]
+                    if isinstance(val, torch.Tensor):
+                        val = val.item() if val.numel() == 1 else val.tolist()
+                    row.append(f"{val}")
+                else:
+                    row.append(" ")
+
+        # keep one column with max length for metrics
+        num_cols = int((term_size - max_length) / max_length)
```
carmocca, line 325: One job is failing:

```python
[1,0]<stderr>:  File "/Users/runner/work/pytorch-lightning/pytorch-lightning/pytorch_lightning/loops/dataloader/evaluation_loop.py", line 325, in _print_results
[1,0]<stderr>:    num_cols = int((term_size - max_length) / max_length)
[1,0]<stderr>:ZeroDivisionError: division by zero
```

AndresAlgaba: Where do you get this error? I believe it implies that terminal column size is zero:
```python
term_size = shutil.get_terminal_size(fallback=(120, 30)).columns
max_length = int(min(max(len(max(metrics + headers, key=len)), 25), term_size / 2))
```
We can solve this by adding that ```max_length``` should be at least equal to one, but not sure whether this makes sense with terminal column size being equal to zero?

carmocca: I'm not sure either. It's caused by Horovod tests: https://github.com/PyTorchLightning/pytorch-lightning/runs/5046435233?check_suite_focus=true#step:16:3236

It might be a problem with it and `shutil`. We can fall back to a 120 default in this case

akihironitta: A quick search led me to:
- this issue https://bugs.python.org/issue42174 and
- this PR which includes a fix https://github.com/python/cpython/pull/29046

>    .. versionchanged:: 3.11
>       The ``fallback`` values are also used if :func:`os.get_terminal_size`
>       returns zeroes.

The fix/update will be shipped in 3.11, so we need to handle it in our side for the time being.

@AndresAlgaba Shall we use the default 120 if term_size==0 as @carmocca suggests?

AndresAlgaba: Thanks for looking into it @akihironitta! Seems good to me, added the ```or 120``` in the latest commit..

PR #11117, opened 2021-12-17
Drop Python 3.6 support

carmocca:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/9981

### Does your PR introduce any breaking changes?

We no longer support Python 3.6, this follows the schedule in https://endoflife.date/python

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [n/a] Did you write any **new necessary tests**? (not for typos and docs)
- [n/a] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

cc @carmocca @akihironitta @borda

PR #11333, opened 2022-01-05
Add typing to accelerators/gpu.py

carmocca:
## What does this PR do?

Part of #7037
Part of #11205

I'm going to try and always have one typing PR open for reviews :)

None

## Before submitting

## PR review

cc @borda


pytorch_lightning/accelerators/gpu.py
```diff
@@ -106,7 +109,8 @@ def get_nvidia_gpu_stats(device: torch.device) -> Dict[str, float]:
     gpu_stat_keys = [k for k, _ in gpu_stat_metrics]
     gpu_query = ",".join(gpu_stat_keys)
 
-    gpu_id = _get_gpu_id(device.index)
+    index = torch._utils._get_device_index(device)
```
carmocca, line 112: This is a very simple function, we can copy it locally too.

I used this so we support all device types.

pytorch_lightning/strategies/single_device.py
```diff
@@ -20,20 +22,21 @@
...
 class SingleDeviceStrategy(Strategy):
     """Strategy that handles communication on a single device."""
 
     def __init__(
         self,
-        device: torch.device,
-        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
-        checkpoint_io: Optional[CheckpointIO] = None,
-        precision_plugin: Optional[PrecisionPlugin] = None,
+        device: _DEVICE,
```
carmocca, line 33: Now this supports all device types.

pytorch_lightning/strategies/single_device.py
```diff
@@ -20,20 +22,21 @@
...
+        accelerator: pl.accelerators.accelerator.Accelerator | None = None,
+        checkpoint_io: CheckpointIO | None = None,
+        precision_plugin: PrecisionPlugin | None = None,
     ):
         super().__init__(accelerator=accelerator, checkpoint_io=checkpoint_io, precision_plugin=precision_plugin)
-        self.device: torch.device = device
+        self._root_device = torch.device(device)
```
carmocca, line 39: There's no reason to have read-only `root_device` and a writable `device`. Hiding the attribute.

tchaton (approved):
LGTM !

pytorch_lightning/strategies/single_device.py
awaelchli: yes but this is now very limiting, because subclasses can't restrict the type because it is a union of 3 types now. mypy does not allow it.

justusschock: I agree, I think we should omit raw strings and ints here. So far we said that this is not an API each user will use and therefore we can expect people to stick to the API of `torch.device` and I'd advocate to keep it that way.

carmocca: > mypy does not allow to choose only a subset.

@awaelchli can you illustrate the problem exactly? this works:

https://mypy-play.net/?mypy=latest&python=3.10&gist=3491fba8a2c0c70f4ecaab22a2da919a

> So far we said that this is not an API each user will use and therefore we can expect people to stick to the API of torch.device and I'd advocate to keep it that way.

Since the goal is for all this to become stable soon (hence fully public), I'd say it is also expected that we minimize boilerplate on these APIs

pytorch_lightning/strategies/single_device.py
```diff
@@ -46,7 +49,7 @@ def on_tpu(self) -> bool:
     def on_gpu(self) -> bool:
         return self.root_device.type == "cuda" and torch.cuda.is_available()
 
-    def reduce(self, tensor: Union[Any, torch.Tensor], *args: Any, **kwargs: Any) -> Union[Any, torch.Tensor]:
+    def reduce(self, tensor: Any | torch.Tensor, *args: Any, **kwargs: Any) -> Any | torch.Tensor:
```
akihironitta, line 52: noob question: how is it different from `-> Any` (without `torch.Tensor`)?

carmocca: I believe @justusschock advocated for this in the past.

tchaton (approved):
LGTM !

pytorch_lightning/utilities/types.py
```diff
@@ -44,6 +44,7 @@
     Dict[str, Sequence[DataLoader]],
 ]
 EVAL_DATALOADERS = Union[DataLoader, Sequence[DataLoader]]
+_DEVICE = Union[torch.device, str, int]
```
four4fish, line 47: In which case the devices will be int? I feel this losses typing restrict for some classes/method. For example, in GPUAccelerator, before the device has to be torch.device or str, now int will pass the type check too.

pytorch_lightning/accelerators/gpu.py
```diff
@@ -54,7 +57,7 @@ def set_nvidia_flags(local_rank: int) -> None:
         devices = os.getenv("CUDA_VISIBLE_DEVICES", all_gpu_ids)
         _log.info(f"LOCAL_RANK: {local_rank} - CUDA_VISIBLE_DEVICES: [{devices}]")
 
-    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
+    def get_device_stats(self, device: _DEVICE) -> dict[str, Any]:
```
four4fish, line 60: Should we have same device typing for `get_nvidia_gpu_stats` and `get_nvidia_gpu_stats`. 
If pass in a str device to `get_device_stats`, we will fail silently as `get_nvidia_gpu_stats` couldn't handle it.

pytorch_lightning/accelerators/gpu.py
```diff
@@ -130,5 +134,5 @@ def _get_gpu_id(device_id: int) -> str:
     """Get the unmasked real GPU IDs."""
     # All devices if `CUDA_VISIBLE_DEVICES` unset
     default = ",".join(str(i) for i in range(torch.cuda.device_count()))
-    cuda_visible_devices: List[str] = os.getenv("CUDA_VISIBLE_DEVICES", default=default).split(",")
+    cuda_visible_devices = os.getenv("CUDA_VISIBLE_DEVICES", default=default).split(",")
```
four4fish, line 137: Unrelated to this PR, but seems we could consolidate/unify the device related functions a bit. 
We have similar logics in https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/utilities/device_parser.py too

carmocca: Want to open an issue with the precise proposal?

pytorch_lightning/utilities/types.py
carmocca: `pytorch` defines `int` as a valid `device` in their functions:

https://github.com/pytorch/pytorch/blob/62441157e3b5ee96e9cb99f7621422b8bf394d09/torch/cuda/memory.py#L117

four4fish: We can either add support for our device or not align with `pytorch` at this time, should we have an issue support int device across our code base?

pytorch_lightning/accelerators/gpu.py
carmocca: If i'm not mistaken, with the changes in this PR, all types are supported for both.

four4fish: in line 109 we are calling "gpu_id = _get_gpu_id(device.index)" if device is str or int, it will fail?

pytorch_lightning/utilities/types.py
carmocca: Right now this only impacts `SingleDeviceStrategy`, hence why I added

`self._root_device = torch.device(device)`

as `self.root_device` is always a `torch.device`.

I don't think `self.root_device` needs to support `str` or `int`.

pytorch_lightning/accelerators/gpu.py
carmocca: It was changed too: https://github.com/PyTorchLightning/pytorch-lightning/pull/11333/files#diff-83cb1466a7f054dee61f11ba07f6be97615216d8eb399dd468e17fd89425ec56R112

four4fish: Oh sorry, my bad, saw it now!

pytorch_lightning/utilities/types.py
four4fish: Sure, we can discuss if this type been used somewhere else or this topic come up again

pytorch_lightning/strategies/single_device.py
awaelchli: I illustrate:

class SingleDeviceStrategy:

    def __init__(self, device: Union[int, str, device]):
        pass

class LimitedSingleDeviceStrategy:
    def __init__(self, device: device):
        pass

Now mypy will complain that the super type does not match the type of this class.

awaelchli (approved):

pytorch_lightning/strategies/single_device.py
awaelchli: I was a bit mislead by a related issue with *args **kwargs not working that way, but the above sample actually passes mypy checks.

pytorch_lightning/accelerators/gpu.py
four4fish: Issue created https://github.com/PyTorchLightning/pytorch-lightning/issues/11427

pytorch_lightning/strategies/single_device.py
carmocca: To elaborate myself, this is useful when it's `Any` but you want to include the information that it can also be those union types.

For example, for this snippet

```python
from typing import Union, Any

x: Union[Any, int, str]

reveal_type(x)
```

The revealed type is

```python
main.py:5: note: Revealed type is "Union[Any, builtins.int, builtins.str]"
```

So mypy does not throw away this extra information

I guess this is also why `Optional[Any] == Union[Any, None]` is preferred over just `Any`

akihironitta (approved):
Love the new annotations!

pytorch_lightning/strategies/single_device.py
```diff
@@ -20,20 +22,21 @@
...
 
     def __init__(
         self,
-        device: torch.device,
-        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
-        checkpoint_io: Optional[CheckpointIO] = None,
-        precision_plugin: Optional[PrecisionPlugin] = None,
+        device: _DEVICE,
+        accelerator: pl.accelerators.accelerator.Accelerator | None = None,
+        checkpoint_io: CheckpointIO | None = None,
```
rohitgr7, line 35: quick ques: is this new API supported on 3.6? since we test that on the release branch.

rohitgr7 (approved):

pytorch_lightning/strategies/single_device.py
akihironitta: @rohitgr7 I don't think it's supported on 3.6 as PEP 585 which is one of the PEPs we're addressing mentions: https://www.python.org/dev/peps/pep-0585/#backwards-compatibility
> Python files with the "annotations" future-import (available since Python 3.7) can parameterize...

---

@carmocca #11117 should be cherry picked into the weekly release branch, right? since Python 3.6 ended its life.

carmocca: @akihironitta we plan to still support it during all `1.5.x` releases

PR #11377, opened 2022-01-09
Move progress bar disabling out of the Trainer

### LINKED ISSUE #10982: Move progress bar disabling out of the Trainer
## Proposed refactor

Move this logic to individual progress bar callback implementations: 
https://github.com/PyTorchLightning/pytorch-lightning/blob/6369e3b77fa3f38613b661517f6361f842f611c9/pytorch_lightning/trainer/trainer.py#L1273-L1275

### Motivation

1. Simplifies the trainer
2. Avoid duplication of this logic in between trainer & spawning plugins. For example, this logic is replicated in the TPU Spawn strategy: https://github.com/PyTorchLightning/pytorch-lightning/blob/6369e3b77fa3f38613b661517f6361f842f611c9/pytorch_lightning/plugins/training_type/tpu_spawn.py#L158-L159

3. Avoid duplication of this logic across different running stages:

- Training: listed above
- Evaluation: https://github.com/PyTorchLightning/pytorch-lightning/blob/6369e3b77fa3f38613b661517f6361f842f611c9/pytorch_lightning/trainer/trainer.py#L1287-L1289
- Prediction: ???

4. Allows custom progress bars to collect information from all ranks.

We underwent a very similar refactor for loggers here to remove rank 0 restrictions: 
https://github.com/PyTorchLightning/pytorch-lightning/issues/8589
https://github.com/PyTorchLightning/pytorch-lightning/pull/8608

#7740 

### Pitch

`enable` and `disable` can be internal implementations of the progress bar callback. This flags can be set anytime after the `setup` hook runs.

```python
def setup(self, trainer, pl_module, stage) -> None:
    if not trainer.is_global_zero:
        self.disable()
    ...
```

This way the trainer doesn't have to do any special checks for progress bars in the middle of the training control flow

### Additional context

cc @borda @justusschock @awaelchli @akihironitta @SeanNaren @kaushikb11

daniellepintz:
## What does this PR do?
Fixes #10982

## Before submitting

## PR review

## Did you have fun?


kaushikb11 (approved):

akihironitta (approved):

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -159,6 +159,8 @@ def print(self, *args, **kwargs):
 
     def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
         self._trainer = trainer
+        if not trainer.is_global_zero:
```
rohitgr7, line 162: quick ques: this will now disable with `predict` too? since this is not the case on master. was this a bug before?

rohitgr7 (approved):

pytorch_lightning/callbacks/progress/base.py
daniellepintz: yes. I believe this was a bug before

rohitgr7: mind update the changelog? with a test?

ananthsub, line 162: let's do this to be explicit about enable/disable
```suggestion
        if trainer.is_global_zero:
            self.enable()
        else:
            self.disable()
```

ananthsub (approved):
As a follow up, it no longer seems like `enable` or `disable` *need* to be part of the progress bar callback API, as the trainer no longer directly calls them.

https://github.com/PyTorchLightning/pytorch-lightning/blob/3e0569fccc9808b1017ace63951485e2f76563eb/pytorch_lightning/callbacks/progress/base.py#L139-L154

If that is the case, then we could further streamline the API by not making these required. All progress bar implementations would be required to resolve this themselves, and we can move thie disabling logic out of the base progress bar implementation. 

it's similar to what we did with loggers allowing them to log from all ranks, and particular implementations can choose to log from rank 0 only if they wish to.

pytorch_lightning/callbacks/progress/base.py
daniellepintz: I had this initially, but this will break the case where we manually disable the progress bar before the setup hook gets called, like in this test: https://github.com/PyTorchLightning/pytorch-lightning/blob/3e0569fccc9808b1017ace63951485e2f76563eb/tests/callbacks/test_tqdm_progress_bar.py#L465-L480

If we remove `enable` and `disable` from the progress bar callback API this will no longer be an issue, but we would have to wait for the dep. cycle

daniellepintz: Done!

rohitgr7 (approved):

daniellepintz:
`tests/checkpointing/test_model_checkpoint.py::test_model_checkpoint_score_and_ckpt[True-False-True-val_log]` is failing for some reason on ubuntu, but works in windows environment. Does not seem related to this PR. has anyone else seen this issue before?

daniellepintz:
@carmocca did you mean to revert the docstring changes here? https://github.com/PyTorchLightning/pytorch-lightning/pull/11377/commits/578ea1a8a857adba601e9ed7ef59ebacc6bd3bad

carmocca:
Nope, my bad!

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -153,6 +149,8 @@ def print(self, *args: Any, **kwargs: Any) -> None:
 
     def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
         self._trainer = trainer
+        # if not trainer.is_global_zero:
+        #     self.disable()
```
awaelchli, line 153: Are you having trouble here? Is there an issue with tests failing?

awaelchli (approved):

pytorch_lightning/callbacks/progress/base.py
daniellepintz: yes :/ the issue is https://github.com/PyTorchLightning/pytorch-lightning/pull/11377#issuecomment-1009489813

test failing with https://gist.github.com/daniellepintz/205b9b05750fa212534b66a5e8ac1f08

daniellepintz: I am quite confused because this test shouldnt even interact with the progress bar at all: https://github.com/PyTorchLightning/pytorch-lightning/blob/18bbb39eef77c8c3d19789b1067164e962640c93/tests/checkpointing/test_model_checkpoint.py#L154

daniellepintz: I have identified the problem is due to the test I added `test_tqdm_progress_bar_disabled_when_not_rank_zero`, but I am still trying to figure out why

daniellepintz:
This PR is ready to be merged. I had to make `test_tqdm_progress_bar_disabled_when_not_rank_zero` a standalone test, to avoid the very strange issue with `test_model_checkpoint_score_and_ckpt[True-False-True-val_log]` described in the comments above

rohitgr7:
> This PR is ready to be merged. I had to make `test_tqdm_progress_bar_disabled_when_not_rank_zero` a standalone test, to avoid the very strange issue with `test_model_checkpoint_score_and_ckpt[True-False-True-val_log]` described in the comments above

is it failing randomly sometimes or always? if random, maybe try seeding the test.

daniellepintz:
> is it failing randomly sometimes or always?

always

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -618,3 +618,32 @@ def test_step(self, batch, batch_idx):
 
     trainer.test(model, verbose=False)
     assert pbar.calls["test"] == []
+
+
+@RunIf(standalone=True)
```
carmocca, line 623: I tried running 
`python -m pytest tests/callbacks/test_tqdm_progress_bar.py::test_tqdm_progress_bar_disabled_when_not_rank_zero tests/checkpointing/test_model_checkpoint.py::test_model_checkpoint_score_and_ckpt`
(after removing the standalone decorator)

And had no issues. If standalone helps it's probably because there's an unwanted interaction between tests.

This should be investigated before merging this PR

daniellepintz: yes, the tests work locally. the problem is only in the CI. It does seem like there's an unwanted interaction between tests. I have already spent many hours investigating this, and narrowed the problem down to the`@mock.patch` decorator. I tried changing it to `with mock.patch ...` inside the test, but that didn't help either. however when I remove the `mock.patch` completely, the checkpointing test passes (in the CI).

At this point I don't know what to do next. One option is we could merge this without the test, since this change would still be a strict improvement from the current situation.

Borda (approved):

tests/callbacks/test_tqdm_progress_bar.py
carmocca: If it's consistent in the CI, we can ssh into it and debug there

Just needs to add this

```yaml
    - name: Setup tmate session
      if: always()
      uses: mxschmitt/action-tmate@v3
```

To the problematic job

awaelchli:
Some investigation has to be done here #11590. I am convinced this PR is affected by this. 
To unblock here, we could add pytest-random-order and find a seed that makes the tests pass. That would be fine with me, but on the other hand this PR is not a very high priority.

Issue #7666, opened 2021-05-23
Restructure Lightning Docs


edenlightning:
## 🚀 Feature
Restructure Lightning documentation to make it easier to read, find, and use.

### Motivation
Lightning has grown a lot, and so did the docs. We should do a massive clean up.

oplatek:
Can you please links to github source code?

For example in here: https://pytorch-lightning.readthedocs.io/en/latest/extensions/generated/pytorch_lightning.callbacks.Callback.html#pytorch_lightning.callbacks.Callback

I cannot jump to Github page exploring the source code. I just jump to another docs page.
<img width="855" alt="Screenshot 2021-05-31 at 16 43 56" src="https://user-images.githubusercontent.com/229266/120209952-883eca80-c22f-11eb-9a2c-d88cd0c547c7.png">

akihironitta:
@oplatek Jumping to the source code in the docs (which is the same as the PyTorch docs) isn't enough for you? e.g. https://pytorch-lightning.readthedocs.io/en/latest/_modules/pytorch_lightning/callbacks/base.html#Callback

oplatek:
@akihironitta do not take it as a criticism ... it is a nice feature for sure! Just want more always :) 

However, linking directly to GitHub would be really awesome for several use-cases I use very frequently:
1. exploring history on GitHub and comparing it with my version to debug the differences
2. GitHub supports jumps to references / jump to definitions - one can use a web browser almost like an IDE.

When I am thinking about it would be enough to put a single link to a python module/file on GitHub from the top of each page in the docs:

e.g. for https://pytorch-lightning.readthedocs.io/en/latest/_modules/pytorch_lightning/callbacks/base.html
put a link to https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/callbacks/base.py

I always search for the Github code anyway based on the docs.

PR #10440, opened 2021-11-09
Replacing latest tag with stable for lightning-bolts links

ThomVett:
Links to example tutorials use the latest tag:
- the tag works for pytorch-lightning links
- latest tag does not work for lightning-bolts links

Replacing `latest` with `0.4.0` for lightning-bolts links as they bring up the correct file.

Resolves #10427

PR #11394, opened 2022-01-10
Refine style guide

rohitgr7:
## What does this PR do?

Part of docs revamp #7666

## Before submitting

## PR review

## Did you have fun?

cc @borda @rohitgr7


docs/source/starter/style_guide.rst
```diff
@@ -24,7 +26,7 @@ In Lightning we differentiate between a system and a model.
 
 A model is something like a resnet18, RNN, etc.
 
-A system defines how a collection of models interact with each other. Examples of this are:
+A system defines how a collection of models interact with each other with user-defined training/evaluation logic. Examples of this are:
 
 * GANs
 * Seq2Seq
```
akihironitta, line 32: The below bullet could be simply deleted.
```
* etc
```

docs/source/starter/style_guide.rst
```diff
@@ -198,7 +211,20 @@ because all these questions need to be answered:
...
+------------
+
+********
+Examples
+********
+
+Checkout the live examples to get your hands dirty:
+
+- `Introduction to PyTorch Lightning <https://pytorch-lightning.readthedocs.io/en/latest/notebooks/lightning_examples/mnist-hello-world.html>`_
+- `Introduction to DataModules <https://pytorch-lightning.readthedocs.io/en/latest/notebooks/lightning_examples/datamodules.html>`_
```
akihironitta, line 230: Following Jirka's comment in https://github.com/PyTorchLightning/pytorch-lightning/pull/10440#discussion_r773263578, let's use `stable` instead of `latest`.
```suggestion
- `Introduction to PyTorch Lightning <https://pytorch-lightning.readthedocs.io/en/stable/notebooks/lightning_examples/mnist-hello-world.html>`_
- `Introduction to DataModules <https://pytorch-lightning.readthedocs.io/en/stable/notebooks/lightning_examples/datamodules.html>`_
```

docs/source/starter/style_guide.rst
rohitgr7: why ? 🤔

docs/source/starter/style_guide.rst
```diff
@@ -49,7 +51,7 @@ Here's a LightningModule that defines a system:
 .. testcode::
 
     class LitModel(LightningModule):
-        def __init__(self, encoder: nn.Module = None, decoder: nn.Module = None):
```
ananthsub, line 52: n00b question: why remove the typehint?

docs/source/starter/style_guide.rst
```diff
@@ -147,26 +153,27 @@ In practice, this code looks like:
...
 
 .. code-block:: python
 
-    def training_step(self):
-        x, y = ...
+    def training_step(self, batch, batch_idx):
+        x, y = batch
         z = self(x)  # < ---------- instead of self.encoder(x)
         pred = self.decoder(z)
         ...
```
ananthsub, line 179: should we recommend users to implement forward then to avoid a special carveout for DataParallel? it's not intuitive why DataParallel requires this

docs/source/starter/style_guide.rst
```diff
@@ -176,19 +183,25 @@ However, when using DataParallel, you will need to call forward manually
...
 
-.. warning:: Make sure not to use ddp_spawn with num_workers > 0 or you will bottleneck your code.
+.. warning:: Make sure not to use ``Trainer(strategy="ddp_spawn")`` with ``num_workers>0`` in a DataLoader or you will bottleneck your code.
 
 DataModules
 ===========
-Lightning introduced datamodules. The problem with dataloaders is that sharing full datasets is often still challenging
-because all these questions need to be answered:
+
+Lightning introduced :class:`~pytorch_lightning.core.datamodule.LightningDataModule` in v0.9 release. They are designed as a way of decoupling data-related
```
ananthsub, line 200: the version it was introduced in is not so useful imo

docs/source/starter/style_guide.rst
```diff
@@ -1,7 +1,8 @@
 ###########
-Style guide
+Style Guide
 ###########
-A main goal of Lightning is to improve readability and reproducibility. Imagine looking into any GitHub repo,
+
+A main goal of Lightning is to improve readability and reproducibility. Imagine looking into any GitHub repo or a research project,
 finding a lightning module and knowing exactly where to look to find the things you care about.
```
ananthsub, line 7:
```suggestion
A main goal of Lightning is to improve readability and reproducibility. Imagine looking into any GitHub repo or a research project,
finding a lightning module, and knowing exactly where to look to find the things you care about.

```

docs/source/starter/style_guide.rst
akihironitta: It just looks unnecessary to me, but if you think it looks nicer as is, let's keep it :)

docs/source/starter/style_guide.rst
```diff
@@ -38,7 +40,7 @@ Here's a LightningModule that defines a model:
 .. testcode::
 
     class LitModel(LightningModule):
-        def __init__(self, num_layers: int = 3):
+        def __init__(self):
```
ananthsub, line 43: "A LightningModule can define both a system and a model" defeats the point above. 
Users should define pure nn.Modules *outside* of the LightningModule. Including the examples with the layers like this isn't how I've seen the vast majority of users interact with Lightning

docs/source/starter/style_guide.rst
rohitgr7: will update!

docs/source/starter/style_guide.rst
rohitgr7: will update!

docs/source/starter/style_guide.rst
rohitgr7: we don't usually have typehints in doc examples.

docs/source/starter/style_guide.rst
rohitgr7: It's a choice I'd say... although a system is what we recommend but if it's a very simple model, then we don't enforce it.

docs/source/starter/style_guide.rst
rohitgr7: DP is a special case worth highlighting. Advising implementing a forward method always might not be good for this section.

docs/source/starter/style_guide.rst
awaelchli: yes agreed

docs/source/starter/style_guide.rst
```diff
@@ -176,19 +189,25 @@ However, when using DataParallel, you will need to call forward manually
...
 DataModules
 ===========
-Lightning introduced datamodules. The problem with dataloaders is that sharing full datasets is often still challenging
-because all these questions need to be answered:
+
+The :class:`~pytorch_lightning.core.datamodule.LightningDataModule` is designed as a way of decoupling data-related
+hooks from the :class:`~pytorch_lightning.core.lightning.LightningModule` so you can develop dataset agnostic models. It makes it easy to hot swap different
+datasets with your model, so you can test it and benchmark it across domains. It also makes sharing and reusing the exact data splits and transforms across projects possible.
+
+The problem with dataloaders is that sharing full datasets is often still challenging because all these questions need to be answered:
```
awaelchli, line 210: This does not longer fit well since you have introduced the concept already above. So the impact of asking these questions is lost. Let's make sure the flow of the intro does not break here.

docs/source/starter/style_guide.rst
```diff
@@ -198,7 +217,20 @@ because all these questions need to be answered:
 It's for this reason that we recommend you use datamodules. This is specially important when collaborating because
 it will save your team a lot of time as well.
 
-All they need to do is drop a datamodule into a lightning trainer and not worry about what was done to the data.
+All they need to do is drop a datamodule into a :class:`~pytorch_lightning.trainer.trainer.Trainer` and not worry about what was done to the data.
 
 This is true for both academic and corporate settings where data cleaning and ad-hoc instructions slow down the progress
 of iterating through ideas.
+
+Checkout :ref:`data` doc to understand data management within Lightning and it's best practices.
```
awaelchli, line 225:
```suggestion
Checkout :ref:`data` doc to understand data management within Lightning and its best practices.
```

docs/source/starter/style_guide.rst
rohitgr7: removed the long intro.. I guess the reference is enough.

docs/source/starter/style_guide.rst
```diff
@@ -176,29 +189,33 @@ However, when using DataParallel, you will need to call forward manually
...
 
-* What splits were used?
-* How many samples does this dataset have?
-* What transforms were used?
-* etc...
+The :class:`~pytorch_lightning.core.datamodule.LightningDataModule` is designed as a way of decoupling data-related
+hooks from the :class:`~pytorch_lightning.core.lightning.LightningModule` so you can develop dataset agnostic models. It makes it easy to hot swap different
+datasets with your model, so you can test it and benchmark it across domains. It also makes sharing and reusing the exact data splits and transforms across projects possible.
+
+Checkout :ref:`data` document to understand data management within Lightning and its best practices.
```
awaelchli, line 210:
```suggestion
Check out :ref:`data` document to understand data management within Lightning and its best practices.
```
Quiz:
Which of the following is a valid git command?

A) git check out
B) git checkout
C) git chickenout

awaelchli (approved):

docs/source/starter/style_guide.rst
rohitgr7: C) :smile: 

oh! man. I swear I checked it on all PRs when you suggested it on one of them :p
I'll include this as a check for the final doc revamp PR.

awaelchli: No worries. I'm just making a little fun of git 😄

docs/source/starter/style_guide.rst
```diff
@@ -147,26 +153,33 @@ In practice, this code looks like:
...
-        x, y = ...
+    def forward(self, x):
+        embeddings = self.encoder(x)
+        return embeddings
+
+
+    def training_step(self, batch, batch_idx):
+        x, y = batch
         z = self(x)  # < ---------- instead of self.encoder(x)
         pred = self.decoder(z)
```
justusschock, line 184: a question in general (even though you did not change this): If we need to call `self(x)` instead of `self.encoder(x)`, can we still call `self.decoder(z)` or would this also have to go to `forward`? I don't see a difference there. If there is one, it probably has to be explained here...

rohitgr7: yes.. this isn't required anymore. I think.
since batch is already splitter at this point, so calling self(), or self.encoder(x) won't be a problem of any sort.

for DP model is warpped around `model=DataParallel(LightningParallelModule(LightningModule)))`

`LightningParallelModule` takes care of which method to call, while `DataParallel` splits the batch and calls LightningParallelModule's forward which then calls `training_step` while training with the splitted batch.

cc: @akihironitta

rohitgr7: this is what I could find from my investigation. But if anyone can confirm this, I'll update this.

justusschock: I think so as well :)

rohitgr7: thanks @justusschock for bringing this up :)

docs/source/starter/style_guide.rst
akihironitta: I agree with @ananthsub.

@rohitgr7 How about explicitly mentioning here that users are discouraged to define a "model" like the example (although users can still write that way)?

docs/source/starter/style_guide.rst
```diff
@@ -49,7 +51,7 @@ Here's a LightningModule that defines a system:
 .. testcode::
 
     class LitModel(LightningModule):
-        def __init__(self, encoder: nn.Module = None, decoder: nn.Module = None):
+        def __init__(self, encoder, decoder=None):
```
akihironitta, line 54: nit
```suggestion
        def __init__(self, encoder, decoder):
```

akihironitta (approved):

docs/source/starter/style_guide.rst
```diff
@@ -24,41 +26,60 @@ In Lightning we differentiate between a system and a model.
...
     class LitModel(LightningModule):
-        def __init__(self, num_layers: int = 3):
+        def __init__(self):
             super().__init__()
             self.layer_1 = nn.Linear()
             self.layer_2 = nn.Linear()
             self.layer_3 = nn.Linear()
 
-Here's a LightningModule that defines a system:
+Here's a LightningModule that defines a system. This is how we recommend.
```
carmocca, line 49:
```suggestion
Here's a LightningModule that defines a system. This structure is what we recommend as a best practice.
```

carmocca (approved):

docs/source/starter/style_guide.rst
```diff
@@ -1,8 +1,9 @@
 ###########
-Style guide
+Style Guide
 ###########
-A main goal of Lightning is to improve readability and reproducibility. Imagine looking into any GitHub repo,
-finding a lightning module and knowing exactly where to look to find the things you care about.
+
+A main goal of Lightning is to improve readability and reproducibility. Imagine looking into any GitHub repo or a research project,
+finding a lightning module, and knowing exactly where to look to find the things you care about.
```
justusschock, line 6:
```suggestion
finding a :class:`~pytorch_lightning.core.lightning.LightningModule`, and knowing exactly where to look to find the things you care about.
```

justusschock (approved):

akihironitta (approved):

docs/source/starter/style_guide.rst
```diff
@@ -24,41 +26,60 @@ In Lightning we differentiate between a system and a model.
...
     class LitModel(LightningModule):
-        def __init__(self, num_layers: int = 3):
+        def __init__(self):
             super().__init__()
             self.layer_1 = nn.Linear()
             self.layer_2 = nn.Linear()
             self.layer_3 = nn.Linear()
 
-Here's a LightningModule that defines a system:
+Here's a LightningModule that defines a system. This structure is what we recommend as a best practice.
```
ananthsub, line 49: This should come first in the docs since it's the recommended path to follow 

Somewhere here we should emphasize that this distinction improves modularity, which has a slew of benefits (testability, reduced dependencies, easier to refactor)

docs/source/starter/style_guide.rst
```diff
@@ -24,41 +26,62 @@ In Lightning we differentiate between a system and a model.
...
 A model is something like a resnet18, RNN, etc.
 
-A system defines how a collection of models interact with each other. Examples of this are:
+A system defines how a collection of models interact with each other with user-defined training/evaluation logic. Examples of this are:
 
 * GANs
 * Seq2Seq
 * BERT
-* etc
+* etc...
```
Borda, line 34:
```suggestion
* etc.
```
I guess it is "etc." or "..." but not both

PR #11442, opened 2022-01-12
Set `Loop.restarting` recursively

carmocca:
## What does this PR do?

Part of #8578

The PR requires this to fix failing tests. It also makes sense as a standalone feature.

None

## Before submitting

## PR review

cc @borda @carmocca @justusschock @awaelchli @ninginthecloud @ananthsub @rohitgr7


pytorch_lightning/loops/base.py
```diff
@@ -301,6 +314,7 @@ def load_state_dict(
         for k, v in self.__dict__.items():
             if isinstance(v, Loop):
                 v.load_state_dict(state_dict.copy(), prefix + k + ".")
+        self.restarting = True
```
rohitgr7, line 317: quick ques: should it set restarting=True only if FT is enabled? I think loop save/reload can be used for other cases but shouldn't always enable FT partially by setting it to True.

carmocca: restarting is necessary to reset the state of the loops when reloading, which is part of fault tolerance.

Fault tolerance also includes the dataloader state, rng, ...

Setting restarting=True only if fault-tolerance is enabled seems limiting.

rohitgr7: but this is set once the loop state is loaded already and if dataloader state needs to be loaded next and depends upon this flag then it will eventually reload even if we limit it just for fault tolerance.

Since these loops are pretty easy to use, then users can rely on them to save and reload whenever they want by simply saving/loading these saved states without having to worry about any of these flags being enabled in non-fault tolerant cases.

if we limit this under the fault-tolerant condition, will it be a problem somewhere?

carmocca: I think I'm missing your argument a bit, but let me clarify mine.

`restarting` is part of the Loop's base API by itself, meaning, it fits in the design even if fault-tolerance did not exist.
The fact that we check `_FaultTolerantMode.detect_current_mode().is_enabled` in this function is totally a hack to avoid restoring the progress tracking state since we are not ready for it yet (I'm working on removing this hack in #8578, which motivates this PR).

So from this point of view, setting `restarting=True` only when fault-tolerance is enabled would be extending this hack further. We could do it, but it would be removed anyways with #8578.

Did that clear it up?

rohitgr7: sorry, I was just playing around with loops a little. That clarifies it. Thank you :)

carmocca: No need to be sorry at all! Are questions are welcome 🎉

awaelchli (approved):

rohitgr7 (approved):

akihironitta (approved):

PR #11443, opened 2022-01-12
Use a dataclass as the scheduler config

carmocca:
## What does this PR do?

Closes #11211
Part of #7037 

- Deprecates `trainer.lr_schedulrs` which still needs to return a dict for compatibility.
- Adds `trainer.lr_scheduler_configs` which does return a list of `LRSchedulerConfig` dataclasses.

If anybody has a better name than `lr_scheduler_config`, let me know.

### Follow-ups:

Resolve:
```
# TODO: rename to `lr_scheduler_configs` to match the property in the `Trainer`
```

None

## Before submitting

## PR review

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7 @tchaton


pytorch_lightning/strategies/strategy.py
```diff
@@ -52,6 +52,7 @@ def __init__(
         self.checkpoint_io = checkpoint_io
         self.precision_plugin = precision_plugin
         self.optimizers: List[Optimizer] = []
+        # FIXME: rename to _config?
```
carmocca, line 55: There's no need to rename this in this PR, but do we want to?

akihironitta: +1 for renaming it to `lr_scheduler_configs`. Should we have consistent variable names in the codebase because currently, variable name `lr_scheduler(s)` is used for both configs and actual lr schedulers.

carmocca: Yes. Would you like to help with that after this PR is merged?

akihironitta: Yes!

pytorch_lightning/core/optimizer.py
```diff
@@ -350,31 +356,18 @@ def _validate_scheduler_api(lr_schedulers: List[Dict[str, Any]], model: "pl.Ligh
             )
 
 
-def _get_default_scheduler_config() -> Dict[str, Any]:
```
awaelchli, line 353: Good that flash no longer depends on it https://github.com/PyTorchLightning/lightning-flash/pull/1099
We should be good here 😃

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -461,28 +461,33 @@ def zero_stage_3(self) -> bool:
         return self.config.get("zero_optimization") and self.config.get("zero_optimization").get("stage") == 3
 
     def _initialize_deepspeed_train(self, model):
+        optimizer, scheduler = None, None
         if "optimizer" in self.config:
             rank_zero_info(
                 "You have specified an optimizer and/or scheduler within the DeepSpeed config."
                 " It is recommended to define it in `LightningModule.configure_optimizers`."
             )
-            optimizer, lr_scheduler = None, _get_default_scheduler_config()
```
awaelchli, line 469: you can use the dataclass here with defaults, then this avoids the None checks below and the code is as simple as before.

pytorch_lightning/strategies/sharded_spawn.py
```diff
@@ -39,11 +39,10 @@ class DDPSpawnShardedStrategy(DDPSpawnStrategy):
 
     def configure_ddp(self) -> None:
         trainer = self.lightning_module.trainer
-        self.model, optimizers = self._setup_model_and_optimizers(
+        self.model, self.optimizers = self._setup_model_and_optimizers(
```
awaelchli, line 42: why is this relevant to this PR?

awaelchli: this should only be done after the lightning optimizer logic has been cleaned up.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2022,19 +2017,26 @@ def optimizers(self, new_optims: Optional[List[Optimizer]]) -> None:
...
         return self.strategy.lr_schedulers
 
-    @lr_schedulers.setter
-    def lr_schedulers(self, new_schedulers: List[LRSchedulerConfig]) -> None:
-        self.strategy.lr_schedulers = new_schedulers
+    @property
+    def lr_schedulers(self) -> List[Dict[str, Any]]:
+        rank_zero_deprecation(
+            "`Trainer.lr_schedulers` is deprecated in v1.6 and will be removed in v1.8."
+            " You can use `trainer.lr_scheduler_configs` instead which contains dataclasses instead of dictionaries.",
```
awaelchli, line 2027: do we need a way for people to access schedulers directly outside the LightningModule? If so, I strongly recommend to change the deprecation to just a warning that mentions that the output changes to the list of schedulers.

awaelchli (approved):

pytorch_lightning/core/optimizer.py
carmocca: Another side comment is that `LRSchedulerConfig` is not protected so I guess we'll have to keep compatibility for it from now on.

pytorch_lightning/strategies/deepspeed.py
carmocca: The scheduler instance is a required argument of `LRSchedulerConfig` and I didn't want to change that and make it Optional. Otherwise, we open the typing can of worms of checking `assert is not None` everywhere.

Considering this, do you see a different way of writing this?

pytorch_lightning/trainer/trainer.py
carmocca: I didn't get the suggestion completely. Are you suggestion we keep both

`lr_scheduler_configs(self) -> List[LRSchedulerConfig]:`
and `lr_schedulers(self) -> List[Dict[str, Any]]`?

What would be the benefit?

Note that the previous annotation

`lr_schedulers(self) -> List[LRSchedulerConfig]:`

was incorrect

> do we need a way for people to access schedulers directly outside the LightningModule

Yes. Notice how many places I had to rename because of its use outside of the trainer/strategy.

awaelchli: It is a pretty simple question. either we answer yes or no. If yes, something like this:

```py
def lr_schedulers(self) -> List[LRScheduler]:
    if version >= 1.8:
        return schedulers
    else: 
        # TODO: remove this warning in 1.8
        warn("Warning: behavior will change in 1.8 ...")
        return configs
```
the schedulers can currently be accessed either in the LM and through filtering them out of the config.

carmocca: Ah I see. I think we can open an issue proposing this change, but this doesn't block this PR.

We would still have a deprecation message, not for the name but for the output.

pytorch_lightning/strategies/deepspeed.py
awaelchli: I tried to make it optional so we a can do `LRSchedulerConfig(None)` here but then the problem is that elsewhere we would need to `assert scheduler_config.scheduler is not None` to make typing checks pass when accessing that attribute.

So no, I guess it can't go much simpler ..

pytorch_lightning/core/optimizer.py
```diff
@@ -251,18 +258,21 @@ def _configure_optimizers(
...
         if isinstance(scheduler, dict):
             # check provided keys
-            extra_keys = [k for k in scheduler.keys() if k not in default_config.keys()]
+            supported_keys = {field.name for field in fields(LRSchedulerConfig)}
+            extra_keys = scheduler.keys() - supported_keys
             if extra_keys:
                 rank_zero_warn(
-                    f"Found unsupported keys in the lr scheduler dict: {extra_keys}", category=RuntimeWarning
+                    f"Found unsupported keys in the lr scheduler dict: {extra_keys}. "
+                    " HINT: remove them from the output of `configure_optimizers`.",
```
rohitgr7, line 272:
```suggestion
                    f"Found unsupported keys in the lr scheduler dict: {extra_keys}."
                    " HINT: remove them from the output of `configure_optimizers`.",
```

pytorch_lightning/core/lightning.py
```diff
@@ -158,19 +164,19 @@ def optimizers(
         # multiple opts
         return opts
 
-    def lr_schedulers(self) -> Optional[Union[Any, List[Any]]]:
+    def lr_schedulers(self) -> Optional[Union[_LRScheduler, List[_LRScheduler]]]:
```
rohitgr7, line 167:
```suggestion
    def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRSchedulerTypeUnion]]]:
```
 can be ReduceLROnPlateau?

pytorch_lightning/trainer/trainer.py
```diff
@@ -2014,19 +2014,26 @@ def lightning_optimizers(self) -> Dict[int, LightningOptimizer]:
...
```
rohitgr7, line 2024:
```suggestion
            " You can use `trainer.lr_scheduler_configs` instead.",
```
it will return dictionaries only right? since you have used `asdict` down below.

rohitgr7 (approved):
LGTM!
small comments

pytorch_lightning/trainer/trainer.py
carmocca: `lr_scheduler_configs` returns dataclasses
`lr_schedulers` returns dictionaries

Here I'm pointing the difference to the user.

pytorch_lightning/strategies/strategy.py
carmocca: @akihironitta You are good to start this now. No need to deprecate because strategies are currently experimental so should be a simple rename

PR #9047, opened 2021-08-23
2/n inter batch parallelism

tchaton:
## What does this PR do?

This PR adds DataFetcher to `data_connector` and loops. 

Parts #8867
Parts #8316

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #11466, opened 2022-01-13
Simplify data fetching

carmocca:
## What does this PR do?

Refactors code introduced in #9047

- Remove (some) unnecessary abstractions and methods
- Removed some dead code
- Remove profiling, it's not necessary at the fetching level for now, `batch_to_device` is still profiled by `call_hook`.
- Removes hack to move the batches to device caused by an esoteric bug(?) where `enumerate` was keeping a reference to the batch alive after `training_step`. See https://github.com/PyTorchLightning/pytorch-lightning/pull/11466#discussion_r785254597

Possible follow-ups:
- Each loop should construct its own `step_kwargs`
- We can simplify further, there's really no need for an abstract interface at the moment.

Simplification is good just to reduce complexity, have faster code, and improve unit-testability which this area lacks.

Some average times over 3 runs using a dumb benchmark:

| Branch | Regular training (s) | InterBatchParallelism (s) | Speedup      |
|--------|----------------------|---------------------------|--------------|
| Master | 1.8212               | 1.3283                    | 1.3711       |
| #11466 | 1.8087               | 1.2795                    | 1.4136       |

which makes it 4.24% faster. Code:

<details>

```python
import os
from time import time
from typing import Any
from unittest import mock

import torch

def get_cycles_per_ms() -> float:
    def measure() -> float:
        start = torch.cuda.Event(enable_timing=True)
        end = torch.cuda.Event(enable_timing=True)
        start.record()
        torch.cuda._sleep(1000000)
        end.record()
        end.synchronize()
        cycles_per_ms = 1000000 / start.elapsed_time(end)
        return cycles_per_ms

    num = 10
    vals = []
    for _ in range(num):
        vals.append(measure())
    vals = sorted(vals)
    stats = vals[2 : num - 2]
    return sum(stats) / len(stats)


BATCH_SIZE = 32
DATASET_LEN = 64
EMB_SZ = 100
EMB_DIM = 64


class RandomIndicesDataset(Dataset):
    def __getitem__(self, index):
        return torch.randint(EMB_DIM, [BATCH_SIZE])

    def __len__(self):
        return 16


class RecommenderModel(BoringModel):
    def __init__(self):
        super().__init__()
        self.layer = None
        self.local_embedding = torch.nn.Embedding(EMB_SZ, EMB_DIM)
        self.CYCLES_PER_MS = int(get_cycles_per_ms())

    def forward(self, indices: torch.Tensor):
        result = self.local_embedding(indices)
        return result

    def on_after_batch_transfer(self, batch: Any, dataloader_idx: int) -> Any:
        # emulate heavy routine
        torch.cuda._sleep(self.CYCLES_PER_MS * 50)
        return batch

    def training_step_end(self, training_step_outputs):
        # emulate heavy routine
        torch.cuda._sleep(self.CYCLES_PER_MS * 50)
        return training_step_outputs

        return torch.optim.SGD(self.parameters(), lr=0.1)

    def train_dataloader(self):
        return DataLoader(RandomIndicesDataset(), batch_size=4)

    def val_dataloader(self):
        return DataLoader(RandomIndicesDataset(), batch_size=4)

    def test_dataloader(self):
        return DataLoader(RandomIndicesDataset(), batch_size=4)


model = RecommenderModel()

trainer_kwargs = dict(
    max_epochs=3,
    accelerator="gpu",
    devices=1,
    num_sanity_val_steps=0,
    enable_progress_bar=0,
    enable_checkpointing=False,
    logger=False,
    enable_model_summary=False,
)

trainer = Trainer(**trainer_kwargs)
with mock.patch.dict(os.environ, {"PL_INTER_BATCH_PARALLELISM": "1"}):
    t0 = time()
    trainer.fit(model)
    t1 = time()
inter_batch_duration = t1 - t0
global_step = trainer.global_step

torch.cuda.synchronize()

trainer = Trainer(**trainer_kwargs)
t2 = time()
trainer.fit(model)
t3 = time()
regular_duration = t3 - t2

print(regular_duration, inter_batch_duration, regular_duration / inter_batch_duration)
```

</details>

All the data fetching implementations are experimental, so no.
But if you were using them, I definitely broke your code 🙈

## Before submitting

## PR review

cc @justusschock @awaelchli @akihironitta @rohitgr7 @ninginthecloud @borda

### DIFF
--- a/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
@@ -20,7 +20,6 @@ from typing import Any, Dict, Iterator, Optional, Union
 from deprecate import void
 
 from pytorch_lightning.loops.base import Loop
-from pytorch_lightning.loops.utilities import _update_dataloader_iter
 from pytorch_lightning.trainer.progress import BatchProgress
 from pytorch_lightning.trainer.states import TrainerFn
 from pytorch_lightning.trainer.supporters import CombinedLoader
@@ -88,7 +87,7 @@ class EvaluationEpochLoop(Loop):
         self._data_fetcher = data_fetcher
 
         self._reload_dataloader_state_dict(data_fetcher)
-        self._dataloader_iter = _update_dataloader_iter(data_fetcher, self.batch_progress.current.ready)
+        self._dataloader_iter = iter(data_fetcher)
 
     def advance(  # type: ignore[override]
         self, data_fetcher: AbstractDataFetcher, dataloader_idx: Optional[int], dl_max_batches: int
@@ -106,7 +105,8 @@ class EvaluationEpochLoop(Loop):
         void(dl_max_batches)
 
         assert self._dataloader_iter is not None
-        batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
+        batch_idx = self.batch_progress.current.ready
+        batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
 
         if batch is None:
             raise StopIteration
--- a/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
@@ -111,12 +111,10 @@ class EvaluationEpochLoop(Loop):
         if batch is None:
             raise StopIteration
 
-        if not data_fetcher.store_on_device:
-            batch = self.trainer._call_strategy_hook("batch_to_device", batch, dataloader_idx=(dataloader_idx or 0))
-
         self.batch_progress.increment_ready()
 
         # configure step_kwargs
+        # TODO: each loop should construct its own kwargs, so we avoid the dataloader_idx reference here
         kwargs = self._build_kwargs(batch, batch_idx, dataloader_idx)
 
         # hook
--- a/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
@@ -105,17 +105,15 @@ class EvaluationEpochLoop(Loop):
         void(dl_max_batches)
 
         assert self._dataloader_iter is not None
-        batch_idx = self.batch_progress.current.ready
         batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
-
         if batch is None:
             raise StopIteration
 
-        self.batch_progress.increment_ready()
-
         # configure step_kwargs
         # TODO: each loop should construct its own kwargs, so we avoid the dataloader_idx reference here
-        kwargs = self._build_kwargs(batch, batch_idx, dataloader_idx)
+        kwargs = self._build_kwargs(batch, self.batch_progress.current.ready, dataloader_idx)
+
+        self.batch_progress.increment_ready()
 
         # hook
         self._on_evaluation_batch_start(**kwargs)
--- a/pytorch_lightning/loops/epoch/training_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/training_epoch_loop.py
@@ -149,9 +149,6 @@ class TrainingEpochLoop(loops.Loop[_OUTPUTS_TYPE]):
         assert self._dataloader_iter is not None
         batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
 
-        if not data_fetcher.store_on_device:
-            batch = self.trainer._call_strategy_hook("batch_to_device", batch)
-
         self.batch_progress.increment_ready()
 
         self.trainer.logger_connector.on_batch_start(batch, batch_idx)
--- a/pytorch_lightning/loops/epoch/training_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/training_epoch_loop.py
@@ -226,9 +226,8 @@ class TrainingEpochLoop(loops.Loop[_OUTPUTS_TYPE]):
         self.trainer.logger_connector.update_train_step_metrics()
 
     def on_advance_end(self) -> None:
-        """Runs validation and Checkpointing if necessary."""
         # -----------------------------------------
-        # VALIDATE IF NEEDED + CHECKPOINT CALLBACK
+        # VALIDATE IF NEEDED
         # -----------------------------------------
         should_check_val = self._should_check_val_fx(self.batch_idx, self.batch_progress.is_last_batch)
         if should_check_val:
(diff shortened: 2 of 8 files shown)


pytorch_lightning/utilities/fetching.py
```diff
@@ -278,16 +241,14 @@ def fetching_function(self) -> Optional[Tuple[Any, bool]]:
             raise StopIteration
         else:
             try:
-                yield_batch = self.pop_batch()
+                yield_batch = self.batches.pop(0)
                 self._fetch_next_batch()
 
+                # TODO: move `wait` into `move_to_device`?
```
carmocca, line 247: Should this TODO be TODONE?

awaelchli: I don't understand your question. I guess it won't matter since move_to_device is only used in two places and together with wait(). So either way should be good.

awaelchli (approved):
impressive 🤩

pytorch_lightning/utilities/fetching.py
carmocca: Sorry, it was a bad joke. Yeah, the question was if it's worth moving it inside

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -147,10 +147,11 @@ def advance(self, data_fetcher: AbstractDataFetcher) -> None:  # type: ignore[ov
...
         assert self._dataloader_iter is not None
-        batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
-
-        if not data_fetcher.store_on_device:
-            batch = self.trainer._call_strategy_hook("batch_to_device", batch)
+        if not isinstance(data_fetcher, DataLoaderIterDataFetcher):
+            batch_idx = self.batch_idx + 1
+            batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
+        else:
+            batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
```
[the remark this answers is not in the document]
carmocca: The code before did exactly that to consume both iterators the same way:

```python
def _update_dataloader_iter(data_fetcher: AbstractDataFetcher, batch_idx: int) -> Iterator:
    """Attach the dataloader."""
    if not isinstance(data_fetcher, DataLoaderIterDataFetcher):
        # restore iteration
        return enumerate(data_fetcher, batch_idx)
    else:
        return iter(data_fetcher)
```

But it turns out that `enumerate` holds an owning reference to the training batch, probably until the next `next` is called.
So the batch's CUDA memory wasn't getting freed as expected.

@tchaton already observed this back when he worked on this code and added this comment:

```python
                # FIXME: Why does this count as a python `referrers` ?
                # return (self.move_data_to_device(yield_batch) if not self.store_on_device else yield_batch, False)
```

when in reality, the referrer was `enumerate`. This took me a while to notice but it's just how enumerate must be implemented in C. Have a look at this snippet

```python
import gc


class C:
    def __init__(self):
        print("__init__")

    def __del__(self):
        print("Destructed!")


class Generator:
    def __iter__(self):
        for _ in range(3):
            yield C()


class Main:
    def run(self):
        iterator = enumerate(Generator())

        print("Consuming generator")
        # C gets created, `__init__` printed
        batch = next(iterator)

        # see who is referencing the batch other than this variable
        refs = gc.get_referrers(batch)
        # enumerate!
        print(refs)

        # let's dereference it and force collection
        batch = None
        gc.collect()

        # this gets printed before the destructor, which means that the object has not been deleted
        # if the object had cuda memory, it wouldn't have been freed.
        print("main finished")


Main().run()
```

prints

```python
Consuming generator
__init__
[]
main finished
Destructed!
```

Now do this change

```diff
-        iterator = enumerate(Generator())
+        iterator = iter(Generator())
```

Now the printed order is

```python
Consuming generator
__init__
[]
Destructed!
main finished
```

Discovering this behaviour was really interesting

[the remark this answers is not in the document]
awaelchli: That is some deep digging you did there 😮 
Where would be a good place to put a comment summarizing this?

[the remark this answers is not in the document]
carmocca: Nowhere really since it would reference code that does not exist (removed in this pr)

However, we have memory tests that fail if this is violated.

pytorch_lightning/utilities/fetching.py
```diff
@@ -57,48 +55,30 @@ def fetching_function(self) -> Any:
...
-    ) -> None:
+    def setup(self, dataloader: Iterable, batch_to_device: Optional[Callable] = None) -> None:
         self._add_capture_metadata_collate(dataloader)
 
         self.dataloader = dataloader
-        self.stage = stage
         self.batch_to_device = batch_to_device
-        self.profiler = profiler
-
-        if self.profiler is not None and stage is None:
```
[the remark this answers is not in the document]
carmocca: Because `call_hook` already integrates the profiler and I don't think this needs to also profile each `next`. If this is requested by users, we can think about what's the cleanest way to add it back.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -106,19 +105,16 @@ def advance(  # type: ignore[override]
...
         if batch is None:
             raise StopIteration
 
-        if not data_fetcher.store_on_device:
-            batch = self.trainer._call_strategy_hook("batch_to_device", batch, dataloader_idx=(dataloader_idx or 0))
+        # configure step_kwargs
+        # TODO: each loop should construct its own kwargs, so we avoid the dataloader_idx reference here
+        kwargs = self._build_kwargs(batch, self.batch_progress.current.ready, dataloader_idx)
 
         self.batch_progress.increment_ready()
```
rohitgr7, line 116:
```suggestion
        self.batch_progress.increment_ready()

        # configure step_kwargs
        # TODO: each loop should construct its own kwargs, so we avoid the dataloader_idx reference here
        kwargs = self._build_kwargs(batch, self.batch_idx, dataloader_idx)
```
just for better code-readability. will this break anything?

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -225,11 +226,7 @@ def advance(self, data_fetcher: AbstractDataFetcher) -> None:  # type: ignore[ov
         self.trainer.logger_connector.update_train_step_metrics()
 
     def on_advance_end(self) -> None:
-        """Runs validation and Checkpointing if necessary.
-
-        Raises:
-            StopIteration: if :attr:`done` evaluates to ``True`` to finish this epoch
-        """
+        """Runs validation and Checkpointing if necessary."""
```
rohitgr7, line 229:
```suggestion
        """Runs validation and checkpointing if necessary."""
```
nit:

justusschock (approved):

rohitgr7 (approved):

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
carmocca: > will this break anything?

Yes. two issues there:

- EvaluationEpochLoop does not define `self.batch_idx`
- We need to `increment_ready` after accessing it.

Issue #11429, opened 2022-01-11
[RFC] Standardize all stateful components on `state_dict`/`load_state_dict`


jjenniferdai:
## Proposed refactor
Standardize all stateful components on `state_dict`/`load_state_dict`.

### Background

PyTorch convention uses `state_dict`/`load_state_dict` for gathering and loading object state. In lightning, some components follow this convention, while other components do not (see Appedix: current state gathering section)

### Motivation
 
Each component should contribute saving/loading their own state with the same APIs.

1. independently contributing/loading one’s own local component state (aligning with PyTorch primitives: `state_dict`/`load_state_dict`)
2. operating and depending on global component state (Lightning `CheckpointHooks`: `on_save/load_checkpoint`)

**This issue is focused on aligning all components on 1**. Following this convention will allow consistency across Lightning components and consistency with PyTorch conventions. Any stateful component can simply implement their own `state_dict`/`load_state_dict` methods to contribute their own state.
For now **2** is only adjusted as needed (Callbacks).

### Pitch

Lightning already has this Stateful Protocol in `auto_restart.py`. We can move this `_SupportsStateDict` out to the more central `core/hooks.py` file:

https://github.com/PyTorchLightning/pytorch-lightning/blob/34c62da37dc1ed9a1de7e023c690c7528ee56c60/pytorch_lightning/utilities/auto_restart.py#L638-L646

```py
@runtime_checkable
class Stateful(Protocol):
    """This class is used to detect if an object is stateful using `isinstance(obj, Stateful)`."""

    def state_dict(self) -> Dict[str, Any]:
        ...

    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
        ...
```
### Additional context

Part of #7740

### Appendix: current state gathering:
Specifically, current components contribute state in the following different ways:
1. Some components contribute saving/loading their own state with only `on_save/load_checkpoint` [DataModule, PrecisionPlugin, Callbacks]
a. [DataModule, PrecisionPlugin] use `on_save/load_checkpoint` from `CheckpointHooks` https://github.com/PyTorchLightning/pytorch-lightning/blob/948cfd24de4f64a2980395581f15544e5e37eab0/pytorch_lightning/core/hooks.py#L765-L807
[...]

PR #11469, opened 2022-01-13
Rename `_SupportsStateDict` --> `_Stateful` Protocol

jjenniferdai:
## What does this PR do?

Part of #11429: rename `_SupportsStateDict` --> `Stateful` Protocol
## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/utilities/types.py
```diff
@@ -23,7 +23,9 @@
 from torch.optim import Optimizer
 from torch.utils.data import DataLoader
 from torchmetrics import Metric
-from typing_extensions import Protocol, runtime_checkable, TypedDict
+from typing_extensions import TypedDict
+
+from pytorch_lightning.core.hooks import Stateful
```
carmocca, line 28: I strongly suggest that `Stateful` stays in this file. You'll notice there was no `from pytorch_lightning` import here and adding one opens the door for circular dependency hell, as many other source files import `types.py`

carmocca: You can see this already with the red CI

ananthsub: @carmocca @jjenniferdai  what if we added this to a new file: core/protocols.py ?

carmocca: I don't think that would work because of the imports in `pytorch_lightning/core/__init__.py`

What's the downside of keeping it here?

What could work too is `pytorch_lightning/protocols.py`. But then we'd have to move this file too for consistency?

ananthsub: I see, then moving the code around / mapping these dependencies should be handled in a separate issue/PR. The important thing now is establishing this protocol and updating the various lightning components to adopt it

jjenniferdai:
just rename first?

tests/utilities/test_stateful_protocol.py
```diff
@@ -0,0 +1,37 @@
+# Copyright The PyTorch Lightning team.
```
carmocca, line 1: We should use a more general name, for example: `test_types.py`

carmocca (approved):

ananthsub (approved):

rohitgr7 (approved):

pytorch_lightning/utilities/types.py
```diff
@@ -49,8 +49,8 @@
 
 
 @runtime_checkable
-class _SupportsStateDict(Protocol):
-    """This class is used to detect if an object is stateful using `isinstance(obj, _SupportsStateDict)`."""
+class Stateful(Protocol):
```
rohitgr7, line 52: any reason why is it not protected anymore?

awaelchli (approved):

pytorch_lightning/utilities/types.py
jjenniferdai: I was thinking `Stateful` protocol should be considered "public" API, same level as `CheckpointHooks`

rohitgr7: can't say about `CheckpointHooks` and other hook classes now, possibly they will be made protected too since they are not used externally and are used only for the inheritance for public components like LM or LDM. But if they are targeted to be made protected too, we should make this protected now. Just to avoid 1 more possible deprecation.

cc @daniellepintz

Borda (approved):

pytorch_lightning/utilities/types.py
jjenniferdai: ah agree it's easier to elevate than deprecate so let's keep it protected for now

> else I think it will start appearing in API references in docs.

would we not want this?

PR #11505, opened 2022-01-17
Fix Black version conflict

chunyang-wen:
## What does this PR do?

Address the version confict for black in pre-commit configuration file.


carmocca (approved):
Good catch

akihironitta (approved):

ananthsub (approved):

PR #11515, opened 2022-01-17
Add typing to data fetching

carmocca:
## What does this PR do?

Part of #7037

None

## Before submitting

## PR review

### DIFF
--- a/pytorch_lightning/utilities/fetching.py
+++ b/pytorch_lightning/utilities/fetching.py
@@ -15,7 +15,7 @@
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Iterator
 from copy import deepcopy
-from typing import Any, Callable, Generator, List, Optional, Tuple, Union
+from typing import Any, Callable, List, Optional, Tuple
 
 import torch
 from torch.utils.data.dataloader import DataLoader
@@ -58,28 +58,28 @@ class AbstractDataFetcher(ABC):
     def prefetching(self) -> None:
         """Override with your own pre-fetching logic."""
 
+    def on_fetch_start(self) -> Any:
+        """Hook to override to handle the logic before fetching a batch."""
+
+    def on_fetch_end(self, batch: Any, start_output: Any) -> None:
+        """Hook to extend which handles the logic after fetching a batch."""
+
+    def wait(self) -> None:
+        """Hook to override to indicate the `DataFetcher` to wait for an event."""
+
     def __init__(self, prefetch_batches: int = 0) -> None:
         if prefetch_batches < 0:
             raise MisconfigurationException("`prefetch_batches` should at least be 0.")
         self.prefetch_batches = prefetch_batches
-
-        self.dataloader: Optional[Union[DataLoader, CombinedLoader]] = None
+        self.dataloader: Optional[Iterable] = None
         self.dataloader_iter: Optional[Iterator] = None
-
-        self.batch_to_device: Optional[Callable]
-
-        self.batches: List
-        self.fetched: int
-        self.done: bool
-
+        self.batch_to_device: Optional[Callable] = None
         self.reset()
 
     def setup(self, dataloader: Iterable, batch_to_device: Optional[Callable] = None) -> None:
         self._add_capture_metadata_collate(dataloader)
-
         self.dataloader = dataloader
         self.batch_to_device = batch_to_device
-
         self._attach_data_fetcher()
 
     @staticmethod
@@ -92,8 +92,8 @@ class AbstractDataFetcher(ABC):
 
         apply_to_collection(dataloader, DataLoader, _add_capture_metadata_collate)
 
-    def _apply_patch(self):
-        def _apply_patch_fn(loader: DataLoader, iterator: Iterator):
+    def _apply_patch(self) -> None:
+        def _apply_patch_fn(loader: DataLoader, iterator: Iterator) -> None:
             if isinstance(loader, CycleIterator):
                 loader = loader.loader
                 # cycle_iterator = iterator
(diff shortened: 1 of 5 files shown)


pytorch_lightning/utilities/fetching.py
```diff
@@ -158,13 +154,13 @@ def loader_iters(self) -> List[Iterator]:
 
     @property
     def state(self) -> Any:
-        def collect_state(iterator: Iterator):
+        def collect_state(iterator: Iterator) -> Any:
```
carmocca, line 157: I'm not sure about the correct return type here: Would it be the one of `MergedIteratorState.state`?

pytorch_lightning/utilities/fetching.py
```diff
@@ -173,9 +169,7 @@ def _attach_data_fetcher_fn(loader: DataLoader):
 
         apply_to_collection(self.loaders, (DataLoader, CycleIterator), _attach_data_fetcher_fn)
 
-    def __iter__(self) -> Generator[Tuple[Any, bool], None, None]:
-        if self.dataloader is None:
-            raise MisconfigurationException("The iterate hasn't been provided. HINT: Did you call setup function ?.")
+    def __iter__(self) -> "AbstractDataFetcher":
```
justusschock, line 172: What about annotating with `Self` here and only falling back to this for older python versions?

pytorch_lightning/utilities/fetching.py
```diff
@@ -282,23 +288,21 @@ class InterBatchParallelDataFetcher(DataFetcher):
         batch 2:             [HtoD]                          [forward][backward]
     """
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
justusschock, line 291: Just out of curiosity? Why do we go with *args and **kwargs here if we know the signature of `super().__init__`?

pytorch_lightning/utilities/fetching.py
```diff
@@ -257,10 +259,14 @@ def _fetch_next_batch(self, iterator: Iterator) -> None:
...
 
     def move_to_device(self, batch: Any) -> Any:
-        if self.store_on_device and self.batch_to_device is not None:
+        if self.store_on_device:
             batch = self.batch_to_device(batch)
         return batch
 
+    def reset(self) -> None:
+        super().reset()
+        self.batches: List[Any] = []
```
justusschock, line 268: Can we add `batches: List[Any]` to the stub (like static attributes) if defined outside `__init__`? Then no need for typing here anymore :)

pytorch_lightning/utilities/fetching.py
```diff
@@ -308,26 +312,23 @@ def wait(self) -> None:
...
 
 
-class StepFuncDataLoaderIter:
+class StepFuncDataLoaderIter(Iterator):
 
     """This class is a wrapper to keep track of dataloader iterator fetching event while left entirely to user
     control."""
 
-    def __init__(self, iterator: Iterator, data_fetcher: "AbstractDataFetcher"):
+    def __init__(self, iterator: Iterator, data_fetcher: "AbstractDataFetcher") -> None:
```
justusschock, line 320: Why can't we annotate the `data_fetcher` with the class directly? It's the same file, so shouldn't have cyclic import issues...

pytorch_lightning/utilities/fetching.py
```diff
@@ -308,26 +312,23 @@ def wait(self) -> None:
...
-
     def __next__(self) -> Any:
         try:
             data = next(self.iterator)
             self.data_fetcher.fetched += 1
             return data
         except StopIteration:
             self.data_fetcher.done = True
-            raise StopIteration
+            raise
```
justusschock, line 331: very nit: I find it more explicit and easier to read to do something like:

```python
try:
    ...
except StopIteration as e:
    ...
    raise e
```

justusschock (approved):

pytorch_lightning/utilities/fetching.py
carmocca: Just for simplicity, we could add them too and have them in both places

pytorch_lightning/utilities/fetching.py
carmocca: I tried this trick:

```python
try:
    from typing_extensions import Self
except ImportError:
    # workaround for Python 3.7.
    # see https://www.python.org/dev/peps/pep-0673/
    from typing import TypeVar

    Self = TypeVar("AbstractDataFetcher", bound="AbstractDataFetcher")
```

But getting:

```python
pytorch_lightning/utilities/fetching.py:43: error: String argument 1 "AbstractDataFetcher" to TypeVar(...) does not match variable name "Self"  [misc]
pytorch_lightning/utilities/fetching.py:43: error: Incompatible types in assignment (expression has type "TypeVar", variable has type "_SpecialForm")  [assignment]
pytorch_lightning/utilities/fetching.py:182: error: Variable "typing_extensions.Self" is not valid as a type  [valid-type]
pytorch_lightning/utilities/fetching.py:182: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

```

justusschock: I see. Unfortunately this is a very cryptic error message. So I also have no idea how to resolve it

akihironitta (approved):

pytorch_lightning/utilities/fetching.py
```diff
@@ -308,26 +314,23 @@ def wait(self) -> None:
...
     def __next__(self) -> Any:
         try:
             data = next(self.iterator)
             self.data_fetcher.fetched += 1
             return data
-        except StopIteration:
+        except StopIteration as e:
             self.data_fetcher.done = True
-            raise StopIteration
+            raise e
```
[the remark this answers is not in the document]
carmocca: Battle of the nits: https://github.com/PyTorchLightning/pytorch-lightning/pull/11515#discussion_r789646065

Haha

awaelchli (approved):

pytorch_lightning/utilities/fetching.py
```diff
@@ -58,29 +58,35 @@ def fetching_function(self) -> Any:
...
+    def setup(self, dataloader: Iterable, **kwargs: Any) -> None:
         self._add_capture_metadata_collate(dataloader)
+        self._dataloader = dataloader
 
-        self.dataloader = dataloader
-        self.batch_to_device = batch_to_device
-
-        self._attach_data_fetcher()
+    @property
+    def dataloader(self) -> Iterable:
```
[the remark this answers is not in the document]
carmocca: The order is already very messed. Better to leave it for a PR which moves everything where it should be.

pytorch_lightning/utilities/fetching.py
tchaton: I believe it should be Collection of `MergedIteratorState`

tchaton (approved):
LGTM !

PR #11516, opened 2022-01-17
Refactor fetching function

carmocca:
## What does this PR do?

The previous implementations had some code smells. This PR simplifies it.

Part of #11515

None user-facing

## Before submitting

## PR review

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7 @ninginthecloud

### DIFF
--- a/pytorch_lightning/utilities/fetching.py
+++ b/pytorch_lightning/utilities/fetching.py
@@ -228,44 +228,33 @@ class DataFetcher(AbstractDataFetcher):
         """Hook to override to indicate the `DataFetcher` to wait for an event."""
 
     def prefetching(self) -> None:
+        iterator = self.dataloader_iter
         for _ in range(self.prefetch_batches):
             try:
-                self._fetch_next_batch()
+                self._fetch_next_batch(iterator)
             except StopIteration:
                 break
 
-    def fetching_function(self) -> Optional[Tuple[Any, bool]]:
-        if self.done:
-            while self.batches:
-                return self._get_queued_batch()
-            raise StopIteration
+    def fetching_function(self) -> Tuple[Any, bool]:
+        if self.batches:
+            batch = self.batches.pop(0)
         else:
+            # empty iterator, no prefetching done
+            raise StopIteration
+        if not self.done:
+            assert self.dataloader_iter is not None
             try:
-                yield_batch = self.batches.pop(0)
-                self._fetch_next_batch()
-                # wait for batch to be available.
-                self.wait()
-                # yield last and has next
-                return self.move_to_device(yield_batch), False
+                self._fetch_next_batch(self.dataloader_iter)
             except StopIteration:
-                self.batches.insert(0, yield_batch)
                 self.done = True
-                return self._get_queued_batch()
-
-            except IndexError:
-                raise StopIteration
-
-    def _fetch_next_batch(self):
-        data = self.on_fetch_start()
-        batch = next(self.dataloader_iter)
-        self.fetched += 1
-        self.on_fetch_end(batch, data)
-
-    def _get_queued_batch(self) -> Tuple[Any, bool]:
-        batch = self.batches.pop(0)
-        is_last = len(self.batches) == 0
         self.wait()
-        return self.move_to_device(batch), is_last
+        return self.move_to_device(batch), len(self.batches) == 0
+
+    def _fetch_next_batch(self, iterator: Iterator) -> None:
+        start_output = self.on_fetch_start()
+        batch = next(iterator)
+        self.fetched += 1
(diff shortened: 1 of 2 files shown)


pytorch_lightning/utilities/fetching.py
```diff
@@ -228,44 +228,33 @@ def wait(self) -> None:
         """Hook to override to indicate the `DataFetcher` to wait for an event."""
 
     def prefetching(self) -> None:
+        iterator = self.dataloader_iter
         for _ in range(self.prefetch_batches):
             try:
-                self._fetch_next_batch()
+                self._fetch_next_batch(iterator)
```
justusschock, line 234:
```suggestion
                self._fetch_next_batch(self.dataloader_iter)
```
No need for a separate variable :)

pytorch_lightning/utilities/fetching.py
```diff
@@ -228,44 +228,33 @@ def wait(self) -> None:
         """Hook to override to indicate the `DataFetcher` to wait for an event."""
 
     def prefetching(self) -> None:
+        iterator = self.dataloader_iter
```
justusschock, line 231:
```suggestion
```

pytorch_lightning/utilities/fetching.py
```diff
@@ -228,44 +228,33 @@ def wait(self) -> None:
...
-        if self.done:
-            while self.batches:
-                return self._get_queued_batch()
-            raise StopIteration
+    def fetching_function(self) -> Tuple[Any, bool]:
+        if self.batches:
+            batch = self.batches.pop(0)
         else:
+            # empty iterator, no prefetching done
+            raise StopIteration
```
justusschock, line 243: This means, you won't get a batch here at all right? If no pre-fetching was done, we should still wait for a batch to be normally loaded, no? How can this work if you early-exit the function here? The `__iter__` imoplementation does not catch this and therefore it would result in end of iterating at all or am I missing something?

pytorch_lightning/utilities/fetching.py
```diff
@@ -228,44 +228,33 @@ def wait(self) -> None:
...
+                self._fetch_next_batch(self.dataloader_iter)
             except StopIteration:
-                self.batches.insert(0, yield_batch)
                 self.done = True
-                return self._get_queued_batch()
-
-            except IndexError:
-                raise StopIteration
+        self.wait()
+        return self.move_to_device(batch), len(self.batches) == 0
```
justusschock, line 251: How do you get this batch in case of no pre-fetching?

carmocca: I'll follow-up in your other comment which I think asks the same thing

pytorch_lightning/utilities/fetching.py
carmocca: pre-fetching=0 is not supported. See the original implementation before I started refactoring: https://github.com/PyTorchLightning/pytorch-lightning/blob/0066ff012962ffa920e9bbd750f4edbea05d4505/pytorch_lightning/utilities/fetching.py#L65-L71

This should be explicitly checked for, but needs further refactoring of the inheritance structure.

The only class that uses `prefetch_batches=0` is `DataLoaderIterDataFetcher` but it overrides `fetching_function`.

tchaton (approved):
Nice !

justusschock (approved):

pytorch_lightning/utilities/fetching.py
```diff
@@ -228,44 +228,33 @@ def wait(self) -> None:
...
-                return self._get_queued_batch()
-            raise StopIteration
+    def fetching_function(self) -> Tuple[Any, bool]:
+        if self.batches:
+            batch = self.batches.pop(0)
         else:
+            # empty iterator, no prefetching done
+            raise StopIteration
+        if not self.done:
+            assert self.dataloader_iter is not None
```
ethanwharris, line 245: Maybe a `RuntimeError` or `AssertionError` would be better here, with a message saying to open an issue on github or somethingn (since presumably we don't expect this to ever fail)

ethanwharris (approved):
LGTM, small comment

pytorch_lightning/utilities/fetching.py
carmocca: `assert ...` will raise an `AssertionError`, no?

Note that I'm doing this just because of typing (#11515), we are forced to add many of these assertions throughout the codebase because optional set via callbacks are un-checkable with mypy.

ethanwharris: Assert statements can be optimized away (not guaranteed to be run) so they're not generally recommend if there's any real chance they could fail. Since this is just to cheer up mypy then I think it's fine :smiley:

PR #11519, opened 2022-01-18
Remove ignore F401

chunyang-wen:
## What does this PR do?

Remove `F401` Ignoring part in `__init__.py`. Instead we use `__all__` to avoid it.

Relates to #11592


chunyang-wen:
@carmocca

tchaton (approved):
LGTM !

PR #11521, opened 2022-01-18
Let Accelerator inherit from ABC to make sure abstractmethod takes effect

chunyang-wen:
## What does this PR do?

+ We have to inherit from `ABC` in order to make `abstractmethod` to take effect. Make `Accelerator` inherit from `ABC`
+ Update docstring to better describe the function

### DIFF
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from abc import abstractmethod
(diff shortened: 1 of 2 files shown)


chunyang-wen:
The failed test case means that `ABC` takes effect.

carmocca (approved):

justusschock (approved):

ananthsub (commented):
thanks for working on this @chunyang-wen . Could you add an entry to the CHANGELOG since this may be a breaking change for people who write customized accelerator classes and who hadn't implemented the abstract methods?

tchaton (approved):
LGTM !

akihironitta (approved):

awaelchli (approved):

awaelchli:
Thank you for the quick fix @chunyang-wen <3

PR #11522, opened 2022-01-18
Remove useless pass and abc

chunyang-wen:
## What does this PR do?

+ Remove uselesss `pass` statement
+ Callback has no abstract methods, so inherit from `abc.ABC` is not used when we have no `abstractmethod` or `abstractproperty` which we require subclass to implement. For callback, actually all the methods is optional.


chunyang-wen:
It seems that the pipelines are a little fragile with unknown broken reason.

carmocca (approved):

justusschock (approved):

Borda (approved):

chunyang-wen:
Can anyway merge this one?

Issue #10687, opened 2021-11-23
[RFC] Support passing pluggable Accelerators to Trainer


kaushikb11:
## 🚀 Feature

With the present Lightning accelerator design, new accelerators cannot be passed to Trainer unless they are part of the Lightning accelerators.

This is not possible.

```python
trainer = Trainer(accelerator=NewSOTAAccelerator(), devices=4)
```

There is a lot of innovation happening in the space of ML Accelerators, and the list will continue to grow. We should enable support for this functionality and make it easier for users to experiment with new accelerators using Lightning.

This proposal also aims at cleaning up and moving hardware specific logic from the accelerator connector to the accelerators.

For example, the `HPUAccelerator` [PR](https://github.com/PyTorchLightning/pytorch-lightning/pull/10404), which is still in development, adds support for Habana's Gaudi Accelerator. Based on the above points, the Accelerator interface would look like this.

```python
class HPUAccelerator(Accelerator):
    """Accelerator for HPU devices."""
    
    @property
    def accelerator_type(self) -> str:
        """Accelerator type."""
        return "hpu"
	
    @staticmethod
    def parse_devices(devices) -> int:
        # Include the HPU device parsing logic here
        return devices
    
    @staticmethod
    def auto_device_count() -> int:
        """Get the HPU devices when set to auto."""
        return habana.device_count()
    
    @staticmethod
    def get_parallel_devices(devices: int) -> List[torch.device]:
        """Gets parallel devices for the given HPU devices."""
        # Moved the logic from accelerator connector
        return [torch.device("hpu")] * devices
    
    def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
        """Gets stats for the given HPU device."""
        return {}
```

After defining `HPUAccelerator`, the user could provide it to the Trainer without it being part of the Lightning accelerators.

```python
trainer = Trainer(accelerator=HPUAccelerator(), devices=4, strategy=HPUPlugin())
```

cc @borda @tchaton @rohitgr7 @akihironitta

SeanNaren:
I love this! However a double edge sword; we should not detract from the importance of bringing these accelerators/strategies in core Lightning. 

I.e, If we have supporters/maintainers for the Habana accelerator, we should bring this into Lightning.

kaushikb11:
Yes, agreed! Habana accelerator was an example. The aim of the proposal is to add flexibility for the users, and not be limited by our options.

[...]

PR #11536, opened 2022-01-19
Remove `Strategy.on_tpu` property

ananthsub:
## What does this PR do?

Part of #10416 

Reasons for removal:
- This can be deduced from the `root_device` property. The property `on_tpu` is redundant yet it's also an `abstractmethod`. Requiring every strategy class to implement this is boilerplate.
- We don't offer `on_ipu` or `on_X` for existing accelerator implementations
- Given the Accelerator is also customizable/pluggable, the pattern of offering `on_gpu` and `on_tpu` flags doesn't scale.

Yes: removes the `on_tpu` property from the Strategy base class.
Reason: The strategy is still an experimental API, and this is part of making it stable

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -413,7 +413,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed `Strategy.on_gpu` ([#11537](https://github.com/PyTorchLightning/pytorch-lightning/pull/11537))
 
 
-- Removed `Strategy.on_tpu` property ([]())
+- Removed `Strategy.on_tpu` property ([#11536](https://github.com/PyTorchLightning/pytorch-lightning/pull/11536))
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -413,6 +413,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed `Strategy.on_gpu` ([#11537](https://github.com/PyTorchLightning/pytorch-lightning/pull/11537))
 
 
+- Removed `Strategy.on_tpu` property ([]())
+
+
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


kaushikb11 (approved):

carmocca (approved):

four4fish:
+1 for this change.

four4fish:
We have

four4fish:
+1 to this change. I realized the on_tpu is not actually used anywhere, only in tests. Other than on_tpu(), the _XLA_AVAILABLE check is only used at import time. Should we have device_availablity check at X_accelerator initialization time?

ananthsub:
> +1 to this change. I realized the on_tpu is not actually used anywhere, only in tests. Other than on_tpu(), the _XLA_AVAILABLE check is only used at import time. Should we have device_availablity check at X_accelerator initialization time?

@kaushikb11 what do you think^?

In general I am in favor of those checks running as early as possible to avoid wasting users' time. In the GPU accelerator, we do check this here: https://github.com/PyTorchLightning/pytorch-lightning/blob/f41d1e5e5ebb7040a39d137695e818cada9a9234/pytorch_lightning/accelerators/gpu.py#L36-L44

But we don't assert that cuda is available - should we?
And should we do that at the init time vs in `setup_environment` ? 

@four4fish Would this also simplify the accelerator connector initialization logic?

four4fish:
@ananthsub @kaushikb11 I think we can introduce a new function called _device_check() in X_Accelerator, which will be called by accelerator_connector right after accelerator initialization. 
We don't have consistent device availibliies check in neither accelerator class, strategies class nor accelerator_connector at this point, so I introduced it in the [new accelerator_connector rewrite ](https://github.com/PyTorchLightning/pytorch-lightning/blob/4ee369dcc90f57ec6746f9f4f24dc5380ebe9896/pytorch_lightning/trainer/connectors/accelerator_connector_new.py#L160-L162). But it's not ideal, I think it's a better fit in accelerator API.

kaushikb11:
@four4fish @ananthsub I agree. We could add `device_check` functionality in the Accelerator Class. I had planned to add it after starting to work on #10687

tchaton (approved):
LGTM !

Issue #7166, opened 2021-04-22
Log epoch metrics before the `on_evaluation_end` hook


dcfidalgo:
## 🚀 Feature
I would like the epoch metrics to get logged before the `on_evaluation_end` hook.

### Motivation

When doing HPO with the [Ray Tune](https://docs.ray.io/en/master/tune/index.html) library, you usually report the metrics back to Tune `on_validation_end`. If you use a trial scheduler such as the ASHA algorithm, it will terminate the trial if the metric indicates low performance for this specific trial. If i want to visualize the HPO in WandB, for example, i would like to know for which values of the metric the trials got terminated. At the moment this is not possible, since the epoch metrics are not logged before the trial gets terminated.

### Pitch

Right now the order of hooks and logging is this one (taken from the `Trainer.run_evaluation()` method, version 1.3.0rc1):
```python
        # hook
        self.evaluation_loop.on_evaluation_epoch_end()

        # update epoch-level lr_schedulers
        if on_epoch:
            self.optimizer_connector.update_learning_rates(interval='epoch')

        # hook
        self.evaluation_loop.on_evaluation_end()

        # log epoch metrics
        eval_loop_results = self.evaluation_loop.log_epoch_metrics_on_evaluation_end()
```
Simply switching the order of the `log_epoch_metrics_on_evaluation_end` and the `on_evaluation_end` hook, would allow to inspect the epoch metrics in, for example, WandB, even if the trial gets terminated by Ray Tune. If one wants to create/modify the epoch metrics before they get logged, you still can use the `on_evaluation_epoch_end` hook.

### Alternatives

Introduce a new hook that is fired after the epoch metrics are logged. This hook could then be used by the tune reporter callback.

### Additional context

Not sure if there is an obvious reason for the current order, if so, please excuse my misunderstanding. 
If you think the change is desirable i would be more than happy to provide a pull request.

Keep up the great work, lightning is a really amazing library!

justusschock:
Hi @dcfidalgo 

From what I saw, this should be possible to do.
@carmocca Is it possible to also log values in `on_evaluation_end`? I don't think so, right? So would you agree that these two things can be swapped?

cc @PyTorchLightning/core-contributors

carmocca:
> @carmocca Is it possible to also log values in `on_evaluation_end`? I don't think so, right?

It is not:

[...]

PR #11543, opened 2022-01-19
Support logging on evaluation end

### LINKED ISSUE #5285: Allow self.log from on_validation_end
## 🚀 Feature

Currently, `self.log` is [not supported](https://github.com/PyTorchLightning/pytorch-lightning/blob/9c8701f2e2c09a984ac0322942c8f742a7cc6bc9/pytorch_lightning/trainer/connectors/logger_connector/callback_hook_validator.py#L194) in `on_validation_end`. I would like it to be.

### Motivation

I want to compute some metric each time the validation loop runs, but I don't want it to be tied to `on_validation_batch_end`: i.e I want to do my own batching. I also want to `self.log` this so I can then use it as a key to monitor for checkpointing. Finally, the documentation claims:

```
Use the log() method to log from anywhere in a LightningModule and Callback except functions with batch_start in their names.
```

I don't see why we should prevent logging in this method, it just adds a surprising friction.

### Pitch

Don't prevent logging in `on_validation_end`.

carmocca:
## What does this PR do?

Fixes #5285

None

## Before submitting

## PR review

cc @borda @carmocca @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy @awaelchli @ninginthecloud @daniellepintz

### DIFF
--- a/pytorch_lightning/loops/dataloader/evaluation_loop.py
+++ b/pytorch_lightning/loops/dataloader/evaluation_loop.py
@@ -145,6 +145,9 @@ class EvaluationLoop(DataLoaderLoop):
         # hook
         self._on_evaluation_epoch_end()
 
+        # hook
+        self._on_evaluation_end()
+
         logged_outputs, self._logged_outputs = self._logged_outputs, []  # free memory
         # include any logged outputs on epoch_end
         epoch_end_logged_outputs = self.trainer.logger_connector.update_eval_epoch_metrics()
@@ -154,8 +157,8 @@ class EvaluationLoop(DataLoaderLoop):
         # log metrics
         self.trainer.logger_connector.log_eval_end_metrics()
 
-        # hook
-        self._on_evaluation_end()
+        # reset the logger connector state
+        self.trainer.logger_connector.reset_results()
 
         # enable train mode again
         self._on_evaluation_model_train()
@@ -229,9 +232,6 @@ class EvaluationLoop(DataLoaderLoop):
             self.trainer._call_callback_hooks("on_validation_end", *args, **kwargs)
             self.trainer._call_lightning_module_hook("on_validation_end", *args, **kwargs)
             self.trainer._call_strategy_hook("on_validation_end", *args, **kwargs)
-
-        # reset the logger connector state
-        self.trainer.logger_connector.reset_results()
 
     def _on_evaluation_epoch_start(self, *args: Any, **kwargs: Any) -> None:
         """Runs ``on_epoch_start`` and ``on_{validation/test}_epoch_start`` hooks."""
--- a/tests/loops/test_evaluation_loop.py
+++ b/tests/loops/test_evaluation_loop.py
@@ -42,25 +42,6 @@ def test_on_evaluation_epoch_end(eval_epoch_end_mock, tmpdir):
     assert eval_epoch_end_mock.call_count == 4
 
 
-@mock.patch(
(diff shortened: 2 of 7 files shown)


tests/loops/test_evaluation_loop.py
```diff
@@ -42,25 +42,6 @@ def test_on_evaluation_epoch_end(eval_epoch_end_mock, tmpdir):
     assert eval_epoch_end_mock.call_count == 4
 
 
-@mock.patch(
-    "pytorch_lightning.trainer.connectors.logger_connector.logger_connector.LoggerConnector.log_eval_end_metrics"
-)
-def test_log_epoch_metrics_before_on_evaluation_end(update_eval_epoch_metrics_mock, tmpdir):
```
carmocca, line 48: This test is not useful anymore because logging results can be computed dynamically. This wasn't the case when it was added: https://github.com/PyTorchLightning/pytorch-lightning/issues/7166

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -154,8 +157,8 @@ def on_run_end(self) -> List[_OUT_DICT]:
         # log metrics
         self.trainer.logger_connector.log_eval_end_metrics()
 
-        # hook
-        self._on_evaluation_end()
```
tchaton, line 158: Good catch !

tchaton (approved):
LGTM !

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -145,6 +145,9 @@ def on_run_end(self) -> List[_OUT_DICT]:
         # hook
         self._on_evaluation_epoch_end()
 
+        # hook
+        self._on_evaluation_end()
```
rohitgr7, line 149: just concerned with an edge case where a user logs inside this hook and expect `ModelCheckpoint` or `EarlyStopping` to monitor it, or some other crazy callback using `callback_metrics/logged_metrics`.

where do we draw a line and have a hook where we expect all the logs will be present?

justusschock (approved):

pytorch_lightning/loops/dataloader/evaluation_loop.py
carmocca: This is not a problem after the logger redesign because `ModelCheckpoint` saves (for validation) also during `on_validation_end` and it's the last callback to run.

rohitgr7: but we run callbacks before lightningmodule:
https://github.com/PyTorchLightning/pytorch-lightning/blob/1a2536361749d13e3f30cec11c7da11747f425ab/pytorch_lightning/loops/dataloader/evaluation_loop.py#L229-L231

carmocca: Mmm great point.

Yeah then this won't work well for the `LightningModule`.

So we have these options:
1. Drop this feature
1. Go forward knowing this
1. Invert the order for this hook in particular (not a great idea as it would be a breaking change).

What do you think is the best? We should probably mention this in the issue.

rohitgr7: personally 1. since we have a workaround of using `on_validation_epoch_end`. Also, the user seems to be happy with it: https://github.com/PyTorchLightning/pytorch-lightning/issues/5285#issuecomment-1017022560

2. IMO, will leak an issue even though many users maybe not be using it, but we can't tell for sure.
3. same reason.

PR #11552, opened 2022-01-20
Fix val_loop run on restart

### LINKED ISSUE #11504: Redundant Validation When Resuming Training
## 🐛 Bug

When training is resumed from a checkpoint, the following happens for the first epoch of the resumed run:
1. Validation
2. Training
3. Validation

### To Reproduce
https://colab.research.google.com/drive/1UxXoTVFusy8xnFW-ZhodLbjzewSdKsHq?usp=sharing
The model in the notebook is trained for 2 epochs.
In the prints, you can see that for the original training both epochs are run correctly with one validation per epoch, after the training epoch is completed.
When training is resumed from the checkpoint of the first epoch, it can be seen that the epoch has two validation runs, before and after the training.

### Expected behavior
There should be only one validation run per epoch, and it should be after the training.

### Environment

* CUDA:
	- GPU:
		- Tesla T4
	- available:         True
	- version:           11.1
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     False
	- pyTorch_version:   1.10.0+cu111
	- pytorch-lightning: 1.6.0dev
	- tqdm:              4.62.3
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.7.12

cc @borda @carmocca @justusschock @ananthsub @ninginthecloud @rohitgr7

rohitgr7:
## What does this PR do?

Fixes #11504

## Before submitting

## PR review

## Did you have fun?

cc @borda @carmocca @justusschock @ananthsub @ninginthecloud @rohitgr7

### DIFF
--- a/tests/loops/epoch/test_training_epoch_loop.py
+++ b/tests/loops/epoch/test_training_epoch_loop.py
@@ -168,6 +168,10 @@ def test_no_val_on_train_epoch_loop_restart(tmpdir):
     with patch.object(
         trainer.fit_loop.epoch_loop.val_loop,
         "run",
-    ) as mocked:
+        wraps=trainer.fit_loop.epoch_loop.val_loop.run,
+    ) as run_mocked, patch.object(
+        trainer.fit_loop.epoch_loop.val_loop, "advance", wraps=trainer.fit_loop.epoch_loop.val_loop.advance
+    ) as advance_mocked:
         trainer.fit(model, ckpt_path=ckpt_path)
-        assert mocked.call_count == 1
+        assert run_mocked.call_count == 1
+        assert advance_mocked.call_count == 1
--- a/tests/loops/epoch/test_training_epoch_loop.py
+++ b/tests/loops/epoch/test_training_epoch_loop.py
@@ -149,12 +149,12 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
 
 def test_no_val_on_train_epoch_loop_restart(tmpdir):
     """Test that training validation loop doesn't get triggered at the beginning of a restart."""
-    trainer_kwargs = {"max_epochs": 1, "limit_train_batches": 1, "limit_val_batches": 1, "enable_checkpointing": False}
+    trainer_kwargs = {"max_epochs": 1, "limit_train_batches": 1, "limit_val_batches": 1, "num_sanity_val_steps": 0, "enable_checkpointing": False}
     trainer = Trainer(**trainer_kwargs)
     model = BoringModel()
     trainer.fit(model)
-    ckpt_path = tmpdir / "last.ckpt"
-    trainer.save_checkpoint(str(ckpt_path))
+    ckpt_path = str(tmpdir / "last.ckpt")
+    trainer.save_checkpoint(ckpt_path)
 
     trainer_kwargs["max_epochs"] = 2
     trainer = Trainer(**trainer_kwargs)
@@ -163,5 +163,5 @@ def test_no_val_on_train_epoch_loop_restart(tmpdir):
         trainer.fit_loop.epoch_loop.val_loop,
         "run",
     ) as mocked:
-        trainer.fit(model, ckpt_path=str(ckpt_path))
+        trainer.fit(model, ckpt_path=ckpt_path)
         assert mocked.call_count == 1
--- a/tests/loops/epoch/test_training_epoch_loop.py
+++ b/tests/loops/epoch/test_training_epoch_loop.py
@@ -149,7 +149,13 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
 
 def test_no_val_on_train_epoch_loop_restart(tmpdir):
     """Test that training validation loop doesn't get triggered at the beginning of a restart."""
-    trainer_kwargs = {"max_epochs": 1, "limit_train_batches": 1, "limit_val_batches": 1, "num_sanity_val_steps": 0, "enable_checkpointing": False}
+    trainer_kwargs = {
+        "max_epochs": 1,
+        "limit_train_batches": 1,
+        "limit_val_batches": 1,
+        "num_sanity_val_steps": 0,
+        "enable_checkpointing": False,
+    }
     trainer = Trainer(**trainer_kwargs)
     model = BoringModel()
     trainer.fit(model)
--- a/tests/loops/epoch/test_training_epoch_loop.py
+++ b/tests/loops/epoch/test_training_epoch_loop.py
@@ -11,6 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from unittest.mock import patch
+
 import pytest
 
 from pytorch_lightning.loops import TrainingEpochLoop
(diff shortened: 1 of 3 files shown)


tests/loops/epoch/test_training_epoch_loop.py
```diff
@@ -141,3 +143,17 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
         num_optimizers=-1,  # does not matter for manual optimization
     )
     assert prepared == expected
+
+
+def test_no_val_on_train_epoch_loop_restart(tmpdir):
```
carmocca, line 148: Although I usually suggest unit-tests, can you write more of an integration test here?

The solution you propose here works but it's not ideal, it should get properly fixed with #8578: particularly this commit https://github.com/PyTorchLightning/pytorch-lightning/pull/8578/commits/e60215c05d2ae67a730d9126322d62fb1978e457

Since that branch will not need to change `_should_check_val_fx`, this test will become stale, so I would rather have a broader test which works for both PRs.

Thank you!

rohitgr7: if it's not ideal, then I'll make this draft for now and see once #8578 get's merged, if this issue still exists.
Since there are no more patch releases, are you planning to release it in v1.6?

carmocca: If it works for the bug-fix release, I'm totally fine with fixing it this way. #8578 cannot be applied to the bug-fix branch.

Also the extra test coverage is welcome

rohitgr7: since there are no more bugfix releases, will the patch in this PR still exists after your PR gets merged?

carmocca: In my PR I would remove the change you are doing here to `_should_check_val_fx`, but as I said, the test you add here will be helpful because it's something not tested in master.

> since there are no more bugfix releases

Where did you get this from? If there are bugs to fix and we can do the fix to the bug-fix release, we should. This would be one example

rohitgr7: > Where did you get this from?

since there wasn't any bugfix PR, so I assumed.

will update the test then :)

tests/loops/epoch/test_training_epoch_loop.py
```diff
@@ -141,3 +145,23 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
...
+    """Test that training validation loop doesn't get triggered at the beginning of a restart."""
+    trainer_kwargs = {"max_epochs": 1, "limit_train_batches": 1, "limit_val_batches": 1, "enable_checkpointing": False}
```
carmocca, line 152: This should set 

```suggestion
    trainer_kwargs = {"max_epochs": 1, "limit_train_batches": 1, "limit_val_batches": 1, "num_sanity_val_steps": 0, "enable_checkpointing": False}
```

tests/loops/epoch/test_training_epoch_loop.py
```diff
@@ -141,3 +145,23 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
...
+    trainer.save_checkpoint(str(ckpt_path))
+
+    trainer_kwargs["max_epochs"] = 2
+    trainer = Trainer(**trainer_kwargs)
+
+    with patch.object(
+        trainer.fit_loop.epoch_loop.val_loop,
+        "run",
+    ) as mocked:
+        trainer.fit(model, ckpt_path=str(ckpt_path))
```
carmocca, line 166:
```suggestion
        trainer.fit(model, ckpt_path=ckpt_path)
```

tests/loops/epoch/test_training_epoch_loop.py
```diff
@@ -141,3 +145,23 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
...
+
+
+def test_no_val_on_train_epoch_loop_restart(tmpdir):
+    """Test that training validation loop doesn't get triggered at the beginning of a restart."""
+    trainer_kwargs = {"max_epochs": 1, "limit_train_batches": 1, "limit_val_batches": 1, "enable_checkpointing": False}
+    trainer = Trainer(**trainer_kwargs)
+    model = BoringModel()
+    trainer.fit(model)
+    ckpt_path = tmpdir / "last.ckpt"
+    trainer.save_checkpoint(str(ckpt_path))
```
carmocca, line 157:
```suggestion
    ckpt_path = str(tmpdir / "last.ckpt")
    trainer.save_checkpoint(ckpt_path)
```

carmocca (approved):

tests/loops/epoch/test_training_epoch_loop.py
rohitgr7: we don't run a sanity check on a restart, so it would be fine?

carmocca: This doesn't matter for this PR but it will for #8578.

The reason is because here we are not "restarting" since the epoch was saved after the trainer completely finished (aka after `on_train_end`)

rohitgr7: we are restarting with more epochs... is it different?
quick ques: will restarting be set to True if I choose to train with more epochs after your PR?

ananthsub (approved):

ninginthecloud (approved):
Looks good to me! Thanks for fixing this issue!

tests/loops/epoch/test_training_epoch_loop.py
carmocca: No. restarting only matters to know whether we are restarting from within an epoch or not.

tests/loops/epoch/test_training_epoch_loop.py
```diff
@@ -141,3 +145,29 @@ def test_prepare_outputs_training_batch_end_manual(batch_end_outputs, expected):
...
+    trainer = Trainer(**trainer_kwargs)
+    model = BoringModel()
+    trainer.fit(model)
+    ckpt_path = str(tmpdir / "last.ckpt")
+    trainer.save_checkpoint(ckpt_path)
+
+    trainer_kwargs["max_epochs"] = 2
+    trainer = Trainer(**trainer_kwargs)
+
+    with patch.object(
```
ninginthecloud, line 168: Hi, @rohitgr7, could we also check if `val_loop.advance()` is called as well? One of our internal users is trying to patch your fix to their codebase, but encounter `val_loop` runs with  zero data reading. 
Because `val_loop.run()` does not mean we indeed have evaluation, but we may just skip the loop within `val_loop` due to `self.done=True`. Thanks a lot.

rohitgr7: sure!

rohitgr7: sure!

carmocca: Having added the `advance` mock, the `run` mock becomes redundant

rohitgr7: we can't patch 2 methods together?

rohitgr7: should I remove it?

ninginthecloud: We can remove `run` mock. If we mock `advance`, then there's no need to check `run`. because if `advance` is called, so is `run`.

kaushikb11 (approved):

Borda (approved):

Issue #7406, opened 2021-05-06
`global_step`/`current_epoch` issues


carmocca:
There are several critical design issues surrounding both of the `global_step` and `current_epoch` values and their interaction with the training loop and `ModelCheckpoint`.

I'll use this issue to track and summarize some of the challenges and design decisions.

### Global step issues

`global_step` refers to the number of optimizer steps applied. It takes the initial value of 0 (because 0 steps have been applied to the optimizer).

Most people expect `global_step` to be updated like this:

```python
def training_step(data)
    loss = model(data)
    loss.backward()
    return loss   

global_step = 0
for x in range(epochs):
    on_train_epoch_start()
    for data in loader:
        on_train_batch_start()
        loss = training_step(data)
        on_train_batch_end()
        optimizer.step(loss)
        global_step += 1
    on_train_epoch_end()
```

Any of the previous hooks can `self.log()`. These `self.log()` calls use the `global_step` value as the x-axis (if you were to plot it)

But the previous is not the case. If you print the `global_step` inside each hook this is the output:
(`max_epochs=1, limit_train_batches=1`)

```
on_train_epoch_start: 0
on_train_batch_start: 0
on_train_batch_end: 0
on_train_epoch_end: 0 <-- Should be 1!
```

<details>

```python
import os

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)

    def on_train_epoch_start(self) -> None:
        print("on_train_epoch_start", self.trainer.global_step)

    def on_train_batch_start(self, *_, **__):
        print("on_train_batch_start", self.trainer.global_step)

    def on_train_batch_end(self, *_, **__):
        print("on_train_batch_end", self.trainer.global_step)

    def on_train_epoch_end(self):
[...]

PR #11562, opened 2022-01-21
Changes in preparation to #8578

carmocca:
## What does this PR do?

This PR includes an assortment of changes which are not relevant by themselves but are supposed to make reviewing #8578 easier.

Part of #8578

Removes `FitLoop.current_epoch`. The property will have no purpose after #8578. Also having a setter there is dangerous.

## Before submitting

## PR review

cc @borda @justusschock @awaelchli @akihironitta @rohitgr7 @carmocca @ananthsub @ninginthecloud

### DIFF
--- a/pytorch_lightning/loops/fit_loop.py
+++ b/pytorch_lightning/loops/fit_loop.py
@@ -57,16 +57,6 @@ class FitLoop(Loop[None]):
         self._outputs: _EPOCH_OUTPUTS_TYPE = []
 
     @property
-    def current_epoch(self) -> int:
-        """Return the current epoch."""
-        return self.epoch_progress.current.completed
-
-    @current_epoch.setter
-    def current_epoch(self, value: int) -> None:
-        """Setter for the current epoch."""
-        self.epoch_progress.current.completed = value
-
-    @property
     def global_step(self) -> int:
         """Returns the global step."""
         return self.epoch_loop.global_step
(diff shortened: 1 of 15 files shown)


tchaton (approved):
LGTM !

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -967,15 +966,13 @@ def assert_checkpoint_log_dir(idx):
         assert_checkpoint_content(ckpt_dir)
 
         # load from checkpoint
-        trainer_config["callbacks"] = [ModelCheckpoint(dirpath=ckpt_dir, save_top_k=-1)]
```
carmocca, line 970: Need to remove this because creating a new instance means we lose the existing value for
`ModelCheckpoint._last_global_step_saved`

which avoids saving a duplicate checkpoint on the second fit call. Perhaps this value should be saved and reloaded so that creating a new instance also works.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -218,7 +218,9 @@ def restore_loops(self) -> None:
             return
 
         self.trainer.fit_loop.global_step = self._loaded_checkpoint["global_step"]
```
rohitgr7, line 220: are you planning to track global_step in the progress too in the future? since this is being updated manually everywhere.

rohitgr7 (approved):

pytorch_lightning/trainer/connectors/checkpoint_connector.py
carmocca: Yes!

pytorch_lightning/loops/fit_loop.py
```diff
@@ -149,19 +139,15 @@ def _results(self) -> _ResultCollection:
...
-        """Evaluates when to leave the loop.
-
-        Returns True if trainer.should_stop was set (e.g. by early stopping) or if the maximum number of steps or epochs
-        is reached.
-        """
+        """Evaluates when to leave the loop."""
         # TODO(@awaelchli): Move track steps inside training loop and move part of these condition inside training loop
         stop_steps = _is_max_limit_reached(self.global_step, self.max_steps)
-        stop_epochs = _is_max_limit_reached(self.current_epoch, self.max_epochs)
+        stop_epochs = _is_max_limit_reached(self.epoch_progress.current.completed, self.max_epochs)
```
awaelchli, line 145: the drawback is that it becomes less readable.

awaelchli (approved):

pytorch_lightning/loops/fit_loop.py
rohitgr7: can be just `self.trainer.current_epoch` since it's just a property now?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
carmocca: We can discuss in #7406

pytorch_lightning/loops/fit_loop.py
carmocca: This is on purpose because the follow-up will use `processed`

Issue #11449, opened 2022-01-12
Rewrite Accelerator_connector and follow up tasks


four4fish:
## Proposed refactor

We have been discussing this for a while. There are issues related to this topic like:

### Motivation

Moving towards sable strategy version
The current logic is not clear and hard to maintain 
There are a lot of simplification we can do after the rewrite

### Pitch

**The new logic can be divided to 3 parts** (Details in the [PR](https://github.com/PyTorchLightning/pytorch-lightning/pull/11448/))

Part1 : Check mis config set by user - conflict between flags, duplication between flags. And set final flag

Part 2: Choose Strategy, Accelerator, Precision, cluster_envirment and set up parallel devices

Part 3: Initialized Strategy, set up Strategy's Accelerator, Precision, Checkpoint_IO, Cluster environment and Parallel_devices (all require lazy initialization)

### **Follow up items from #11448** 
1. [ ] Move error messages to precisionPlugin, strategy and accelerator __init__ method if possible. 
    eg: move this check to IPUPrecision plugin. from @carmocca
```python
if self._precision_flag not in (16, 32):
                raise MisconfigurationException(
                    f"`Trainer(accelerator='ipu', precision={self._precision_flag!r})` is not supported."
                )
 ```
move this check to strategy. from  @ananthsub 
```python
if self._precision_flag in (16, "bf16") and self._amp_type_flag == AMPType.APEX:
            if isinstance(self.strategy, (DDPShardedStrategy, DDPSpawnShardedStrategy, DDPFullyShardedStrategy)):
                raise MisconfigurationException(
                    "Sharded plugins are not supported with apex, please switch to `amp_backend='native'`."
```
         
2. [x] Add typing to accel_connector. Can we do this as a separate PR after unused properties deprecation? from @kaushikb11  @awaelchli @ananthsub 
3. [ ] Reduce duplicated strategy registry code: Classmethod inheritance doesn't work with current strategy registry logic, cls is the base class not the inheritance class. To reduce duplicated `register_strategies` method, we need redo the strategy registry logic. @kaushikb11 @awaelchli @tchaton 
4. [ ] Flag conflict and fallback logic revisit:
       - different flag set to the same thing: should be error (from @tchaton )
       - dp/ddp2 on cpu fallback to ddp: should be error instead of silent fallback (from @ananthsub )
[...]

PR #11448, opened 2022-01-12
[NOT ready for review] Rewrite accelerator_connector

four4fish:
## What does this PR do?

Fixes #\<issue_number>

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

Issue #7650, opened 2021-05-21
Make lazy initialization in plugins more robust


awaelchli:
## 🚀 Feature

We currently set some attributes like sync_batch_norm, num_nodes, devices etc. lazily in the accelerator connector, so the user does not have to provide them at instantiation. 

Example:

Trainer(gpus=4, plugins=DDPPlugin(find_unused_parameters=True))  # plugin may require gpus, num_nodes etc.

# AcceleratorConnector does this:
training_type_plugin.num_nodes = ...

This is fragile. Some attributes may have to be recomputed based on the order in which others are set.

### Pitch

Provide one single lazy init method that takes all arguments required. The plugin is responsible for making sure dependencies are resolved in one place: 

class DDPPlugin(...):

    def lazy_init(self, **kwargs):
         self.num_nodes = kwargs.get("num_nodes")
         self.num_processes = ...

tchaton:
Enforcing a lazy mechanism for setting the parameters could reduce friction for customisation of the plugins as less logic would like within the AcceleratorConnector.

carmocca:
So is this about having something similar to the dataclass `__post_init__`? Called by the Trainer? When would it be called exactly?

awaelchli:
Yes exactly. It would be called once the accelerator connector has finished setup and instantiating all objects needed.

joanovallesr:
Hi! I'd like to take a crack at implementing this. Since plugins have evolved into Strategies since this issue was opened, I want to confirm if the plan is still to implement a `lazy_init(**kwargs)` directly on the `DDPStrategy` class and refactor the `_AcceleratorConnector` to use it. If so, I have my environment set up and can start drafting a PR!

PR #11572, opened 2022-01-22
Lazy initialize Strategy.parallel_devices

### LINKED ISSUE #11097: Lazy initialize properties in Strategies

four4fish:
## What does this PR do?

Part of #11449
This is will simplify accelerator_connector refactor

Fixes #11097

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -222,6 +222,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `MisconfigurationException` to `ModuleNotFoundError` when `rich` isn't available ([#11360](https://github.com/PyTorchLightning/pytorch-lightning/pull/11360))
 
 
+- Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#](https://github.com/PyTorchLightning/pytorch-lightning/pull/))
+
+
 ### Deprecated
 
 - Deprecated `ClusterEnvironment.master_{address,port}` in favor of `ClusterEnvironment.main_{address,port}` ([#10103](https://github.com/PyTorchLightning/pytorch-lightning/pull/10103))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,7 +236,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#11572](https://github.com/PyTorchLightning/pytorch-lightning/pull/11572))
 
-  
+
 - Sorted `SimpleProfiler(extended=False)` summary based on mean duration for each hook ([#11671](https://github.com/PyTorchLightning/pytorch-lightning/pull/11671))
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -222,7 +222,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `MisconfigurationException` to `ModuleNotFoundError` when `rich` isn't available ([#11360](https://github.com/PyTorchLightning/pytorch-lightning/pull/11360))
 
 
-- Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#](https://github.com/PyTorchLightning/pytorch-lightning/pull/))
+- Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#11572](https://github.com/PyTorchLightning/pytorch-lightning/pull/11572))
 
 
 ### Deprecated
--- a/pytorch_lightning/strategies/ddp.py
+++ b/pytorch_lightning/strategies/ddp.py
@@ -106,7 +106,6 @@ class DDPStrategy(ParallelStrategy):
         self.interactive_ddp_procs = []
         self._num_nodes = 1
         self.sync_batchnorm = False
-        self.num_processes = len(self.parallel_devices) if self.parallel_devices is not None else 0
         self._ddp_kwargs = kwargs
         self._ddp_comm_state = ddp_comm_state
         self._ddp_comm_hook = ddp_comm_hook
@@ -134,6 +133,10 @@ class DDPStrategy(ParallelStrategy):
         # note that world ranks is related to num_nodes, when resetting it, need to reset world ranks
         self._num_nodes = num_nodes
         self.set_world_ranks()
+
+    @property
+    def num_processes(self):
+        return len(self.parallel_devices) if self.parallel_devices is not None else 0
 
     @property
     def distributed_sampler_kwargs(self):
--- a/pytorch_lightning/strategies/ddp_spawn.py
+++ b/pytorch_lightning/strategies/ddp_spawn.py
@@ -82,7 +82,6 @@ class DDPSpawnStrategy(ParallelStrategy):
         self._num_nodes = 1
         self.sync_batchnorm = False
         self._ddp_kwargs = kwargs
-        self.num_processes = len(parallel_devices) if parallel_devices is not None else 0
         self._ddp_comm_state = ddp_comm_state
         self._ddp_comm_hook = ddp_comm_hook
         self._ddp_comm_wrapper = ddp_comm_wrapper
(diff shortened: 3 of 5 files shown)


pytorch_lightning/strategies/parallel.py
```diff
@@ -72,6 +72,14 @@ def world_size(self) -> int:
     def is_global_zero(self) -> bool:
         return self.global_rank == 0
 
+    @property
+    def parallel_devices(self):
+        return self._parallel_devices
```
carmocca, line 77: Did you forget to rename the attribute in the `__init__`? L43

four4fish: No, it's same style as 'accelerator' and etc properties. 
Like https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/strategies/strategy.py#L44-L85

I also feel this way of lazy initialization is confusing. Should we change all of them to 
```
def __init__(accelerator):
      self._accelerator = accelerator
     
@property 
def accelerator(self):
      return self._accelerator
```
?

carmocca: Duh!

I think it's fine as it is.

carmocca (approved):

pytorch_lightning/strategies/ddp.py
```diff
@@ -135,6 +134,10 @@ def num_nodes(self, num_nodes: int) -> None:
         self._num_nodes = num_nodes
         self.set_world_ranks()
 
+    @property
+    def num_processes(self):
+        return len(self.parallel_devices) if self.parallel_devices is not None else 0
```
rohitgr7, line 139: there are some places in accelerator_connector where this is set manually? is that required?
for eg L898.

pytorch_lightning/strategies/ddp.py
```diff
@@ -135,6 +134,10 @@ def num_nodes(self, num_nodes: int) -> None:
         self._num_nodes = num_nodes
         self.set_world_ranks()
 
+    @property
+    def num_processes(self):
```
rohitgr7, line 138: will it work for ddp2? since ddp2 inherits DDP strategy, and num_processes is num_nodes in that case set manually in L900 of `accelerator_connector`.

pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -107,6 +106,10 @@ def local_rank(self) -> int:
     def root_device(self):
         return self.parallel_devices[self.local_rank]
 
+    @property
+    def num_processes(self):
```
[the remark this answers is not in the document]
four4fish: The previous version has self.num_processes set in __init__(), so TPUSpawnStrategy has `num_processes` property too.

pytorch_lightning/strategies/ddp.py
four4fish: accelerator_connector has self.num_processes too, but it's not equal to strategy.num_processes. They are not necessarily the same value. 

Only place accelerator_connector set strategy.num_processes is the L762-L763 before(modified below). 

The` accelerator_connector.num_processes` could be num_nodes, `parallel_devices` or `num_processes`, which equal to `trainer.num_processes` and been used in logging and profilings. 

The `Strategies.num_processes` is calculated from `strategies.parallel_devices`, which is unrelated to `accelerator_connector.num_proceses`. accelerator_connector need to set it manually base on "parallel_devices" in L762-L763 is because the `strategies.parallel_devices` is not lazy initialized

This is a pre-step for #11448, we should definitely revisit all the device related logics after #11448

pytorch_lightning/strategies/ddp.py
four4fish: same as above

pytorch_lightning/strategies/ddp.py
rohitgr7: ah! you are right... got a little confused with the same name.
Thanks for clarifying :)

rohitgr7 (approved):

pytorch_lightning/strategies/ddp.py
four4fish: Yeah, it's really confusing, and I think the devices related logic probably is wrong somewhere too. We can revisit them after the accelerator_connector rewrite

CHANGELOG.md
```diff
@@ -222,6 +222,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `MisconfigurationException` to `ModuleNotFoundError` when `rich` isn't available ([#11360](https://github.com/PyTorchLightning/pytorch-lightning/pull/11360))
 
 
+- Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#](https://github.com/PyTorchLightning/pytorch-lightning/pull/))
+
+
```
akihironitta, line 227:
```suggestion
- Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#11572](https://github.com/PyTorchLightning/pytorch-lightning/pull/11572))


```

four4fish: Thank you @akihironitta !!!

pytorch_lightning/strategies/parallel.py
```diff
@@ -72,6 +72,14 @@ def world_size(self) -> int:
...
+
+    @parallel_devices.setter
+    def parallel_devices(self, parallel_devices):
+        self._parallel_devices = parallel_devices
```
awaelchli, line 81: here we see a hint at the difference between the proposals #11097 and #7650. 

If a setter like this one gets introduced, and other values depend on this value here, these would have to be re-computed everytime this setter is invoked (depending on the requirements of the plugin, error handling would be needed to). If multiple of these setters are necessary, things get complicated and logic gets duplicated. #7650 would solve this, because it would set all attributes in one shot together. I haven't found the time to investigate in this direction further.

awaelchli (approved):

pytorch_lightning/strategies/parallel.py
four4fish: +1 for #7650. The accelerator_connector should be the only place allowed to set/modify these parameters (other than to pass in when creating the object). With #7650 proposal, we can call the lazy_init at the end of the accelerator_connector __init__, which will be way less error-prone and the behavior for properties will be more consistent and robust

This PR take the  #11097 easy way for now to unblock accelerator_connector rewrite..

tchaton (approved):
LGTM !

justusschock (approved):

PR #11575, opened 2022-01-22
Avoid enforcing `shuffle=False` for eval dataloaders

### LINKED ISSUE #11054: Setting `shuffle=True` for `val_dataloader` doesn't work when `ddp` enabled.
## 🐛 Bug

Setting `shuffle=True` for `val_dataloader` doesn't work when `ddp` enabled.

### To Reproduce

minimum snippet:

```python
from torch import nn
from torch.optim import Adam


class PlModule(LightningModule):
    def __init__(self):
        super().__init__()
        self.model = nn.Linear(1, 1)
        self.train_dataset = list(range(1000))
        self.val_dataset = list(range(1000))

        return Adam(self.model.parameters(), lr=0.1)

    def training_step(self, batch, *args, **kwargs):
        return self.model(batch)

    def validation_step(self, batch, *args, **kwargs):
        print(f'{batch=}')
        print(f'{self.trainer.val_dataloaders[0].sampler.shuffle=}')
        exit(0)

    def train_dataloader(self):
        return DataLoader(self.train_dataset, shuffle=True, batch_size=5)

    def val_dataloader(self):
        return DataLoader(self.val_dataset, shuffle=True, batch_size=5)



if __name__ == '__main__':
    module = PlModule()
    trainer = Trainer(
        limit_train_batches=0.0,
        limit_val_batches=1.0,
        num_sanity_val_steps=-1,
        gpus=2,
        strategy='ddp',
        enable_model_summary=False
    )
    trainer.fit(module)
```

By execute the snippet above, we get the stdout:

```
GPU available: True, used: True
initializing distributed: GLOBAL_RANK: 1, MEMBER: 2/2
initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/2
----------------------------------------------------------------------------------------------------
distributed_backend=nccl
All distributed processes registered. Starting with 2 processes
----------------------------------------------------------------------------------------------------

LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0,1,2,3,4,5,6,7]
LOCAL_RANK: 1 - CUDA_VISIBLE_DEVICES: [0,1,2,3,4,5,6,7]
Validation sanity check: 0it [00:00, ?it/s]/home/balabala/miniconda3/lib/python3.8/site-packages/pytorch_lightning/trainer/data_loading.py:452: UserWarning: Your `val_dataloader` has `shuffle=True`,it is strongly recommended that you turn this off for val/test/predict dataloaders.
  rank_zero_warn(
/home/balabala/miniconda3/lib/python3.8/site-packages/pytorch_lightning/trainer/data_loading.py:110: UserWarning: The dataloader, val_dataloader 0, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 96 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  rank_zero_warn(
Validation sanity check:   0%|                          | 0/100 [00:00<?, ?it/s]batch=tensor([1, 3, 5, 7, 9], device='cuda:1')
self.trainer.val_dataloaders[0].sampler.shuffle=False
batch=tensor([0, 2, 4, 6, 8], device='cuda:0')
self.trainer.val_dataloaders[0].sampler.shuffle=False
```

### Expected behavior

the stdout should contains the follow:

```
batch= # some random batch
self.trainer.val_dataloaders[0].sampler.shuffle=True
```

### Environment

- PyTorch Lightning Version 1.5.2
- PyTorch Version 1.10.0
- Python version 3.8.10

cc @borda @tchaton @rohitgr7 @justusschock @awaelchli @ninginthecloud @kaushikb11 @akihironitta

rohitgr7:
## What does this PR do?

Fixes #11054

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/connectors/data_connector.py
+++ b/pytorch_lightning/trainer/connectors/data_connector.py
@@ -435,7 +435,13 @@ class DataConnector:
             )
 
         # add samplers
-        dataloaders = [self._prepare_dataloader(dl, False, mode=mode) for dl in dataloaders if dl is not None]
+        dataloaders = [
+            self._prepare_dataloader(
+                dl, shuffle=hasattr(dl, "sampler") and isinstance(dl.sampler, RandomSampler), mode=mode
+            )
+            for dl in dataloaders
+            if dl is not None
+        ]
 
         # add worker_init_fn for correct seeding in worker processes
         apply_to_collection(
--- a/pytorch_lightning/trainer/connectors/data_connector.py
+++ b/pytorch_lightning/trainer/connectors/data_connector.py
@@ -19,7 +19,6 @@ from typing import Any, Collection, Iterable, List, Optional, Tuple, Union
 from weakref import proxy
 
 from torch.utils.data import DataLoader, RandomSampler, Sampler, SequentialSampler
-from torch.utils.data.dataset import IterableDataset
 from torch.utils.data.distributed import DistributedSampler
 
 import pytorch_lightning as pl
@@ -35,6 +34,7 @@ from pytorch_lightning.utilities.auto_restart import (
 )
 from pytorch_lightning.utilities.data import (
     _auto_add_worker_init_fn,
+    _is_dataloader_shuffled,
     _replace_dataloader_init_method,
     _update_dataloader,
     has_iterable_dataset,
@@ -329,7 +329,10 @@ class DataConnector:
             and not has_iterable_dataset(dataloader)
         )
 
-    def _prepare_dataloader(self, dataloader: Any, shuffle: bool, mode: Optional[RunningStage] = None) -> Any:
+    # TODO: shuffle here is kept for BC. Remove it once data_loading.py is removed (#11248)
+    def _prepare_dataloader(
+        self, dataloader: Any, shuffle: Optional[bool] = None, mode: Optional[RunningStage] = None
+    ) -> Any:
         """This function handles to following functionalities:
 
         - Injecting a `DistributedDataSampler` into the `DataLoader` if on a distributed environment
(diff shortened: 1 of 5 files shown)


pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -435,7 +435,13 @@ def _reset_eval_dataloader(
             )
 
         # add samplers
-        dataloaders = [self._prepare_dataloader(dl, False, mode=mode) for dl in dataloaders if dl is not None]
+        dataloaders = [
+            self._prepare_dataloader(
+                dl, shuffle=hasattr(dl, "sampler") and isinstance(dl.sampler, RandomSampler), mode=mode
```
carmocca, line 440: Shouldn't this use the same conditions as `_check_eval_shuffling`?

rohitgr7: yes, though so but if its a custom sampler, then it would set shuffle=True, which might be incorrect.

rohitgr7: ok, looks like we raise an error in the case of the custom sampler already. will update.

pytorch_lightning/trainer/trainer.py
```diff
@@ -1773,7 +1773,6 @@ def reset_train_dataloader(self, model: Optional["pl.LightningModule"] = None) -
             self.train_dataloader,
             DataLoader,
             self._data_connector._prepare_dataloader,
-            shuffle=True,
```
rohitgr7, line 1776: reviewer: this is removed from here and is being set inside `_prepare_dataloader`, because we can't check for its value here in case of a `CombinedLoader`.

carmocca (approved):

ananthsub (approved):

tchaton (approved):
LGTM !

PR #11594, opened 2022-01-24
reduce only loss with dp

### LINKED ISSUE #11539: Validation_step_batch_end return empty list in 'dp' mode
### To Reproduce

The following tutorial is used and cloned in my machine to test the functionality of Distributed Data-Parallel and Data Parallel.

https://colab.research.google.com/github/wandb/examples/blob/master/colabs/pytorch-lightning/Optimize_Pytorch_Lightning_models_with_Weights_%26_Biases.ipynb

The validation step will return preds for a customized callback LogPredictionsCallback() that saves images.
The results between DDP and DP are completely different. In DDP mode, the sanity checking pass.
However, in DP mode, the outputs in LogPredictionsCallback() are just an empty list and fail the sanity checking. 

As 2 GPU is required. Colab will not provide multiple free GPUs. The bug cannot reproduce in the Colab environment. 

### Expected behavior

The preds from validation_step() should be accumulated and can be called "outputs" in Callback, just like DDP mode.

### Environment
* CUDA:
        - GPU:
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
        - available:         True
        - version:           11.2
* Packages:
        - numpy:             1.21.5
        - pyTorch_debug:     False
        - pyTorch_version:   1.10.0
        - pytorch-lightning: 1.5.8
        - tqdm:              4.62.3
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.9.7
        - version:           #202201071026-Ubuntu SMP Fri Jan 7 16:52:09 UTC 2022

rohitgr7:
## What does this PR do?

Fixes #11539
also: https://github.com/PyTorchLightning/pytorch-lightning/discussions/9914

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/strategies/dp.py
+++ b/pytorch_lightning/strategies/dp.py
@@ -137,18 +137,15 @@ class DataParallelStrategy(ParallelStrategy):
             return self.model(*args, **kwargs)
 
     def training_step_end(self, output):
-        if not is_overridden("training_step_end", self.lightning_module):
-            return self.reduce(output)
-        return output
+        if is_overridden("training_step_end", self.lightning_module):
+            return output
 
-    def validation_step_end(self, output):
-        if not is_overridden("validation_step_end", self.lightning_module):
-            return self.reduce(output)
-        return output
+        if isinstance(output, dict) and "loss" in output:
+            output["loss"] = self.reduce(output["loss"])
 
-    def test_step_end(self, output):
-        if not is_overridden("test_step_end", self.lightning_module):
-            return self.reduce(output)
+        elif isinstance(output, torch.Tensor):
+            output = self.reduce(output)
+
         return output
 
     def teardown(self) -> None:
--- a/tests/accelerators/test_dp.py
+++ b/tests/accelerators/test_dp.py
@@ -134,8 +134,24 @@ class ReductionTestModel(BoringModel):
 
     def training_epoch_end(self, outputs):
         assert outputs[0]["loss"].shape == torch.Size([])
-        assert outputs[0]["reduce_int"].item() == 0  # mean([0, 1]) = 0
-        assert outputs[0]["reduce_float"].item() == 0.5  # mean([0., 1.]) = 0.5
+        self._assert_extra_outputs(outputs)
+
+    def validation_epoch_end(self, outputs):
+        assert outputs[0]["x"].shape == torch.Size([2])
+        self._assert_extra_outputs(outputs)
+
+    def test_epoch_end(self, outputs):
+        assert outputs[0]["y"].shape == torch.Size([2])
+        self._assert_extra_outputs(outputs)
+
+    def _assert_extra_outputs(self, outputs):
+        out = outputs[0]["reduce_int"]
+        assert torch.eq(out, torch.tensor([0, 1], device="cuda:0")).all()
+        assert out.dtype is torch.int
+
+        out = outputs[0]["reduce_float"]
+        assert torch.eq(out, torch.tensor([0.0, 1.0], device="cuda:0")).all()
+        assert out.dtype is torch.float
 
 
 def test_dp_raise_exception_with_batch_transfer_hooks(tmpdir, monkeypatch):
@@ -188,11 +204,9 @@ def test_dp_training_step_dict(tmpdir):
 
     trainer = pl.Trainer(
         default_root_dir=tmpdir,
-        max_epochs=1,
-        limit_train_batches=1,
-        limit_val_batches=1,
-        limit_test_batches=1,
+        fast_dev_run=True,
         gpus=2,
         strategy="dp",
     )
     trainer.fit(model)
+    trainer.test(model)


justusschock (approved):

pytorch_lightning/strategies/dp.py
```diff
@@ -137,18 +137,15 @@ def predict_step(self, *args, **kwargs) -> STEP_OUTPUT:
             return self.model(*args, **kwargs)
 
     def training_step_end(self, output):
-        if not is_overridden("training_step_end", self.lightning_module):
-            return self.reduce(output)
-        return output
+        if is_overridden("training_step_end", self.lightning_module):
+            return output
 
-    def validation_step_end(self, output):
```
carmocca, line 144: Why not reduce the evaluation outputs?

tests/accelerators/test_dp.py
```diff
@@ -134,8 +134,24 @@ def test_step(self, batch, batch_idx):
 
     def training_epoch_end(self, outputs):
         assert outputs[0]["loss"].shape == torch.Size([])
-        assert outputs[0]["reduce_int"].item() == 0  # mean([0, 1]) = 0
-        assert outputs[0]["reduce_float"].item() == 0.5  # mean([0., 1.]) = 0.5
```
carmocca, line 138: Shouldn't these two lines stay for `training_epoch_end`? You are not testing anymore the `.item()` equalities

pytorch_lightning/strategies/dp.py
rohitgr7: the PR updates the logic to reduce only the loss.
reduction for loss is required to make sure `loss.backward` works perfectly for dp.
during evaluation we don't need a loss to compute gradients thus we don't reduce anything.

tests/accelerators/test_dp.py
rohitgr7: they are no longer 0-dimensional tensor since during training we don't reduce anything except the loss. Ther outputs will stay as it is and assertions for these are present in `_assert_extra_outputs` down below.

carmocca (approved):
This is technically a breaking change not? Somebody could be relying on this specific behaviour.

pytorch_lightning/strategies/dp.py
justusschock: one other way to do this would be I think to change `loss.backward()` to `loss.backward(torch.ones_like(loss))` (which should be mathematically equal) and not to reduce anything.

rohitgr7: TIL.

we can do this.. but just wondering
is this a common practice?
since we expose backward hook.

justusschock: I am not sure. But this is the way to handle non-scalar loss-values. We could also add some logic beforehand to check if `gradients` is passed in kwargs and add it there if not and if the loss is not scalar.

CHANGELOG.md
```diff
@@ -222,6 +222,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `MisconfigurationException` to `ModuleNotFoundError` when `rich` isn't available ([#11360](https://github.com/PyTorchLightning/pytorch-lightning/pull/11360))
 
 
+- Enable reducing only loss when using DP ([#11594](https://github.com/PyTorchLightning/pytorch-lightning/pull/11594))
```
akihironitta, line 225: Could be written like this for people relying on the reduction.
```suggestion
- Disabled reducing training_step output other than loss when using DP ([#11594](https://github.com/PyTorchLightning/pytorch-lightning/pull/11594))
```

akihironitta (approved):

pytorch_lightning/strategies/dp.py
rohitgr7: Or let's keep it like this. Maybe not super convinient for every user to do it this way, since we ourselves have doubts of it being mathematically equivalent. If someone wants do configure this differently then can still choose to override `training_step_end`.

CHANGELOG.md
```diff
@@ -222,6 +222,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `MisconfigurationException` to `ModuleNotFoundError` when `rich` isn't available ([#11360](https://github.com/PyTorchLightning/pytorch-lightning/pull/11360))
 
 
+- Disabled reducing training_step output other than loss when using DP ([#11594](https://github.com/PyTorchLightning/pytorch-lightning/pull/11594))
```
awaelchli, line 225:
```suggestion
- When using DP (data-parallel), Lightning will no longer automatically reduce all tensors returned in training_step; it will only reduce the loss unless `training_step_end` is overridden ([#11594](https://github.com/PyTorchLightning/pytorch-lightning/pull/11594))
```

CHANGELOG.md
```diff
@@ -222,6 +222,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
+
+
```
awaelchli, line 227:
```suggestion
- When using DP (data-parallel), the `training_epoch_end` hook will no longer receive reduced outputs from `training_step` and instead get the full tensor of results from all GPUs.
```

Some implementations of training_epoch_end will break, because they may assume a particular shape of tensors. I would add this extra changelog entry to make it absolutely clear that these hooks are affected.

pytorch_lightning/strategies/dp.py
awaelchli: Yes it is equivalent and as @justusschock said the key is the chain rule. 

Detailed:


<details>
  <summary>Details</summary>
  
Let x = f(w) be the unreduced output of a neural net f with weights w, where x = [x0, x1] has size 2 for this this illustration. Let L = x0 + x1 be the reduced loss. 

The derivative of L w.r.t x0 is 

dL / dx0 = 1

The derivative of L w.r.t. x1 is 

dL / dx1 = 1

So together: dL / dx = [1, 1]

The chain rule says:
dL / dw = (dL / dx) * (dx / dw)

With `x.backward([1, 1])` we tell pytorch that we have computed the value of dL/dx ourselves, i.e. [1, 1], and we would like pytorch to compute the rest, i.e., dx/dw so that finally we obtain the derivative dL/dw. 

So, L.backward() is the same as x.backward([1, 1]).
In DP, x0 gets computed by GPU 0, and x1 gets computed by GPU 1, but that doesn't change the math. 

</details>

I don't know if for sure we should do this but if it leads to a simplification, I think it would be fine to explore it (perhaps after this PR).

tchaton: IMO, I would go for Justus's proposal. I believe it is cleaner to leave the tensors exactly as they are and let the users deal with them.

rohitgr7: the reduced loss in this case will be: L = (x0 + x1) / 2, since batches are split across devices.

do we have to divide the loss tensor [x1, x2] by `num_devices` in such a case then?

PR #11599, opened 2022-01-24
Add `estimated_stepping_batches` property to `Trainer`

### LINKED ISSUE #10760: Add total training steps as a property to trainer
## 🚀 Feature

See title:
Similar issues: #5449, #10430.. I'll keep linking more.

### Motivation

This is a highly requested feature from the lightning community. Total training steps is being used by some of the lr schedulers, especially when using transformer models, but since there are a lot of arguments/flags involved for computing it, it's not easy for a user to create one that can work on all the possible edge cases with no-code change. Also we make updates to some of these flags and accessibility to some of its components (for eg. train_datalaoders in v1.5), so there is a possibility that a custom one created by a user might get outdated soon, and one has to write a new one which is only possible if they are well aware with the codebase internals. But we as core-contributors and maintain it with some tests of course.

### Pitch
would like to credit @SeanNaren for helping out :) 

```py
@property
def num_training_steps(self) -> int:
    """Total training steps inferred from datamodule and devices."""
    if self.trainer.num_training_batches != float('inf'):
        dataset_size = self.trainer.num_training_batches
    else:
        rank_zero_warn('Requesting dataloader...')
        dataset_size = len(self.trainer._data_connector._train_dataloader_source.dataloader())

    if isinstance(self.trainer.limit_train_batches, int):
        dataset_size = min(dataset_size, self.trainer.limit_train_batches)
    else:
        dataset_size = int(dataset_size * self.trainer.limit_train_batches)

    accelerator_connector = self.trainer._accelerator_connector
    if accelerator_connector.use_ddp2 or accelerator_connector.use_dp:
        effective_devices = 1
    else:
        effective_devices = self.trainer.devices

    effective_devices = effective_devices * self.trainer.num_nodes
    effective_batch_size = self.trainer.accumulate_grad_batches * effective_devices
    max_estimated_steps = math.ceil(dataset_size // effective_batch_size) * self.trainer.max_epochs

    max_estimated_steps = min(max_estimated_steps, self.trainer.max_steps) if self.trainer.max_steps != -1 else max_estimated_steps
    return max_estimated_steps
```

### Additional context

cc @borda
cc @PyTorchLightning/core-contributors

rohitgr7:
## What does this PR do?

Fixes #10760

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @kaushikb11 @awaelchli @ananthsub @ninginthecloud @jjenniferdai @rohitgr7 @akihironitta


SeanNaren:
Overall looks clean to me!

carmocca:
@rohitgr7 Is this waiting on something?

rohitgr7:
hey @carmocca !
just need to add 1 test for TPU and IPU each. will try to get it done today.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2612,6 +2613,36 @@ def weights_summary(self, val: Optional[str]) -> None:
...
+            rank_zero_warn("Loading `train_dataloader` to estimate number of training steps.")
+            self.reset_train_dataloader()
+
+        total_batches = self.num_training_batches
+
+        # iterable dataset
+        if total_batches == float("inf"):
+            return self.max_steps
+
+        self.accumulate_grad_batches = self.accumulation_scheduler.get_accumulate_grad_batches(self.current_epoch)
```
awaelchli, line 2637: I'm worried about this. If someone uses an actual scheduler, the value of `estimated_num_training_steps` will change every epoch. Is that meaningful? Since it's a property and not a method, it may not be obvious that this value gets recomputed every epoch (if I access it every epoch). 

Three options 
1) document that behavior 
2) since the scheduling is finite, one could give a better estimate by integrating the scheduling into the computation here (so we essentially wouldn't depend on current_epoch)
3) Only allow accumulation with a constant

pytorch_lightning/trainer/trainer.py
```diff
@@ -2612,6 +2613,36 @@ def weights_summary(self, val: Optional[str]) -> None:
         rank_zero_deprecation("Setting `Trainer.weights_summary` is deprecated in v1.5 and will be removed in v1.7.")
         self._weights_summary = val
 
+    """
+    Estimator Properties
+    """
+
+    @property
+    def estimated_num_training_steps(self) -> int:
```
awaelchli, line 2621: since accumulation is factored into the value here, the name "training_steps" is not really accurate here, right? When accumulating, we do indeed perform a training step, just no optimizer step.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2612,6 +2613,36 @@ def weights_summary(self, val: Optional[str]) -> None:
         rank_zero_deprecation("Setting `Trainer.weights_summary` is deprecated in v1.5 and will be removed in v1.7.")
         self._weights_summary = val
 
+    """
+    Estimator Properties
+    """
```
awaelchli, line 2618: Is there a plan to add others? If not let's just move it to an existing section?

pytorch_lightning/trainer/trainer.py
rohitgr7: I'd say 3rd one? since the only use-case I currently know of is that this is being used within a scheduler and the accumulation factor is assumed to be constant in such a case.

pytorch_lightning/trainer/trainer.py
rohitgr7: how about `estimated_num_optimization_steps`?

pytorch_lightning/trainer/trainer.py
rohitgr7: not right now. will update!

pytorch_lightning/trainer/trainer.py
```diff
@@ -2616,6 +2617,41 @@ def weights_summary(self, val: Optional[str]) -> None:
...
+                "Estimated optimization steps cannot be computed with different"
+                " `accumulate_grad_batches` at different epochs."
+            )
+
+        # infinite training
+        if self.max_epochs == -1 and self.max_steps == -1:
+            return float("inf")
+
+        if self.train_dataloader is None:
+            rank_zero_warn("Loading `train_dataloader` to estimate number of optimization steps.")
```
carmocca, line 2637: Why a warning?

```suggestion
            rank_zero_info("Loading `train_dataloader` to estimate number of optimization steps.")
```

tests/trainer/properties/test_estimated_num_training_steps.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def test_num_optimization_steps_with_diff_multiple_grad_accum_factor():
+    """Test that an error is raised if `Trainer` is configured with different gradient accumulation factors at
+    different epochs."""
+    grad_scheduler = GradientAccumulationScheduler(scheduling={7: 2})
+    trainer = Trainer(callbacks=[grad_scheduler])
+    model = BoringModel()
+    trainer._data_connector.attach_data(model)
+    trainer.strategy.connect(model)
```
carmocca, line 44: Not required for the exception

```suggestion
```

carmocca (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2616,6 +2617,41 @@ def weights_summary(self, val: Optional[str]) -> None:
     Other
     """
 
+    @property
+    def estimated_num_optimization_steps(self) -> int:
```
carmocca, line 2621: If this does not account for multiple optimizers or TBPTT splits, then I don't think this is an appropriate name.

I had to make a similar choice in #11805 where I used `batches_that_stepped`. I think we should have a similar name (estimated_stepping_batches?)

Also remember to update the CHANGELOG entry and PR description after we decide!

pytorch_lightning/trainer/trainer.py
awaelchli: yeah agreed, it's nothing that the user can do to prevent this, so info is more appropriate imo.

pytorch_lightning/trainer/trainer.py
awaelchli: estimated_num_optimization_steps: I like it more than training steps. I think it is as close as we can get.

pytorch_lightning/trainer/trainer.py
awaelchli: I can't actually find any other name that would come closer, so I think I'm fine with the optimization steps. We can't go for "estimated_batches_steps" or similar because it takes into account accumulation (which in turn is both connected to the number of batches and the optimization steps).  

maybe a warning/info for when multiple optimizers are present?

pytorch_lightning/trainer/trainer.py
awaelchli: we wouldn't see an info message multiple times, right? only one time at the beginning of the training if we call it too early?

pytorch_lightning/trainer/trainer.py
awaelchli: Yes, that's acceptable imo.

awaelchli (approved):
great to see we are exploring this utility. 
my comments are nits

pytorch_lightning/trainer/trainer.py
rohitgr7: yes..

pytorch_lightning/trainer/trainer.py
rohitgr7: well, when using multiple optimizers, IMO, a warning won't make sense because it will still be optimization steps for each optimizer. I guess the problem here is with the name since by `optimization_steps` do we mean global optimization_steps or total optimization_steps for all the optimizers?

and I can't think of a better name..

tests/trainer/properties/test_estimated_num_training_steps.py
```diff
@@ -0,0 +1,167 @@
...
+    monkeypatch.setattr(torch.cuda, "is_available", lambda: True)
+    monkeypatch.setattr(torch.cuda, "device_count", lambda: 7)
+    trainer = Trainer(max_epochs=1, devices=7, accelerator="gpu", strategy="dp")
+    model = BoringModel()
+    trainer._data_connector.attach_data(model)
+    trainer.strategy.connect(model)
+    assert trainer.estimated_num_optimization_steps == 64
+
+
+@RunIf(tpu=True)
```
kaushikb11, line 144: Why not mock TPU Availability?

tests/trainer/properties/test_estimated_num_training_steps.py
```diff
@@ -0,0 +1,167 @@
...
+    trainer._data_connector.attach_data(model)
+    trainer.strategy.connect(model)
+    assert trainer.estimated_num_optimization_steps == estimated_steps
+
+
+@pytest.mark.parametrize("num_nodes,estimated_steps", [(1, 64), (2, 32), (3, 22)])
+def test_num_optimization_steps_ddp2(num_nodes, estimated_steps, monkeypatch):
+    """Test optimization steps with DDP2."""
+    monkeypatch.setattr(torch.cuda, "is_available", lambda: True)
+    monkeypatch.setattr(torch.cuda, "device_count", lambda: 7)
```
kaushikb11, line 125: 7! Wild :p

tests/trainer/properties/test_estimated_num_training_steps.py
```diff
@@ -0,0 +1,167 @@
...
+    trainer = Trainer(max_epochs=1, num_nodes=num_nodes, devices=7, accelerator="gpu", strategy="ddp")
+    model = BoringModel()
+    trainer._data_connector.attach_data(model)
+    trainer.strategy.connect(model)
+    assert trainer.estimated_num_optimization_steps == estimated_steps
+
+
+@pytest.mark.parametrize("num_nodes,estimated_steps", [(1, 64), (2, 32), (3, 22)])
+def test_num_optimization_steps_ddp2(num_nodes, estimated_steps, monkeypatch):
```
kaushikb11, line 122: Could parameterize strategies `ddp`, `dp` and `ddp2` into one test.

tests/trainer/properties/test_estimated_num_training_steps.py
rohitgr7: I tried.. couldn't get it to work. Wanna git it a shot?

tests/trainer/properties/test_estimated_num_training_steps.py
rohitgr7: yeah.. +1

tests/trainer/properties/test_estimated_num_optimization_steps.py
```diff
@@ -0,0 +1,160 @@
...
+def test_num_optimization_steps_with_diff_multiple_grad_accum_factor():
+    """Test that an error is raised if `Trainer` is configured with different gradient accumulation factors at
+    different epochs."""
+    grad_scheduler = GradientAccumulationScheduler(scheduling={7: 2})
+    trainer = Trainer(callbacks=[grad_scheduler])
+    with pytest.raises(MisconfigurationException, match="cannot be computed with different"):
+        assert trainer.estimated_num_optimization_steps
+
+
+def test_num_optimization_steps_raises_warning_with_no_dataloaders_loaded(caplog):
```
kaushikb11, line 48: Isn't it an info message instead of warning?

tests/trainer/properties/test_estimated_num_training_steps.py
kaushikb11: Sure

kaushikb11 (approved):

pytorch_lightning/trainer/trainer.py
carmocca: > We can't go for "estimated_batches_steps" or similar because it takes into account accumulation (which in turn is both connected to the number of batches and the optimization steps).

"batches that stepped" refers to "batches where an optimizer will step". That takes into account accumulation.

If this estimate is called `*_optimization_steps` then it will be confusing after #11805  because it will be different than the global step at the end of training in the case of multiple optimizers or tbptt.

Would you like to discuss this off github?

tests/trainer/properties/test_estimated_num_training_steps.py
kaushikb11: So it's hard to mock, as it calls `trainer.strategy.connect(model)`, which calls XLA's xm api.

rohitgr7: yep!

tests/trainer/properties/test_estimated_num_optimization_steps.py
```diff
@@ -0,0 +1,160 @@
...
+    """Test that a warning is raised when dataloaders are loaded explicitly if they are not already configured."""
```
rohitgr7, line 49:
```suggestion
def test_num_optimization_steps_raises_info_with_no_dataloaders_loaded(caplog):
    """Test that an info message is generated when dataloaders are loaded explicitly if they are not already configured."""
```

pytorch_lightning/trainer/trainer.py
rohitgr7: did you guys discuss this offline?

kaushikb11: @carmocca @awaelchli

rohitgr7: hey  @kaushikb11 !
we discussed this .. will be updated the PR soon :)

docs/source/common/trainer.rst
```diff
@@ -1808,3 +1809,19 @@ The metrics sent to the progress bar.
...
     assert progress_bar_metrics["a_val"] == 2
+
+
+estimated_num_optimization_steps
+********************************
+
+Calculates the estimated number of optimization steps for the complete training. The value is inferred from DataLoaders, gradient
+accumulation factor and distributed setting.
+
+.. code-block:: python
```
Borda, line 1820: this seems the same as docstring later, so can we rather refer to the `estimated_num_optimization_steps` API?

pytorch_lightning/trainer/trainer.py
```diff
@@ -2661,6 +2662,53 @@ def weights_summary(self, val: Optional[str]) -> None:
...
+
+        # iterable dataset
+        if total_batches == float("inf"):
+            return self.max_steps
+
+        self.accumulate_grad_batches = accumulation_scheduler.get_accumulate_grad_batches(self.current_epoch)
+        effective_batch_size = self.accumulate_grad_batches
+        max_estimated_steps = math.ceil(total_batches / effective_batch_size) * (
+            self.max_epochs if self.max_epochs != -1 else 1
+        )
```
Borda, line 2707:
```suggestion
        max_estimated_steps = math.ceil(total_batches / effective_batch_size) * max(self.max_epochs, 1)
```
or is there a case with `self.max_epochs` is 0?

tchaton (approved):
Awesome ! @rohitgr7  Might be worth considering to introduce an `estimated_progress` too.

pytorch_lightning/trainer/trainer.py
rohitgr7: I don't think so.. but just kept to avoid zero-division error in any case.

rohitgr7: nope..

docs/source/common/optimizers.rst
```diff
@@ -592,3 +593,21 @@ Here we configure gradient clipping differently for optimizer B.
             self.clip_gradients(
                 optimizer, gradient_clip_val=gradient_clip_val * 2, gradient_clip_algorithm=gradient_clip_algorithm
             )
+
+
+Total Stepping Batches
```
ananthsub, line 598: @rohitgr7 I had a question on this name

Why not total optimization steps? People often get confused about the terminology in Lightning between steps and batches, and this property combines both of them. Is the concept of a batch even relevant here, since it's really related to model weight updates?

rohitgr7: It is resolved? Or you still need some clarification?

ananthsub: i had missed the discussion in the PR which answered my questions

docs/source/common/optimizers.rst
```diff
@@ -592,3 +593,21 @@ Here we configure gradient clipping differently for optimizer B.
...
+:class:`~torch.optim.lr_scheduler.OneCycleLR` scheduler, which requires pre-computed ``total_steps`` during initialization.
+
+.. code-block:: python
+
+    def configure_optimizers(self):
+        optimizer = ...
+        scheduler = torch.optim.lr_scheduler.OneCycleLR(
+            optimizer, max_lr=1e-3, total_steps=self.trainer.estimated_stepping_batches
+        )
+        return [optimizer], [scheduler]
```
Reason239, line 613: Hi!

I think this code example may be misleading, bacause (if I get it right) the default scheduler interval is 'epoch', so here the scheduler would only update every epoch (and wouldn't complete the cycle, because the total_steps is set to estimated_stepping_batches, not number of epochs). I tested it in [Colab](https://colab.research.google.com/drive/1fDL6d1whBex9qf4tfELBCdLgkmPf6IXZ?usp=sharing)

Should it be like this?
```
    return [optimizer], [{"scheduler": scheduler, "interval": "step"}]
```

awaelchli: Sounds good to me. Would you like to send a PR updating this example?

Issue #11538, opened 2022-01-19
enable ffcv with PL


williamFalcon:
A team introduced ffcv which does optimizations in the dataloader level.
As long as it can be a drop in replacement for dataloader it should add the benefits to any PyTorch and Lightning script.

ie:
```python
# current
data = Dataloader(...)

trainer.fit(..., data)

# 10x faster with ffcv
data = FFCVDataloader

trainer.fit(..., data)
```

Let's make sure when users want to get the benefits of both FFCV and PL they can do that!
cc @borda @rohitgr7 @Felonious-Spellfire @justusschock @awaelchli @ninginthecloud @otaj @tchaton @carmocca

tchaton:
@ethanwharris This could be built-in within Flash.

paantya:
https://github.com/libffcv/ffcv/issues/72

justusschock:
This is related to #10696

siddk:
Any update on this?

carmocca:
FFCV should be perfectly usable with no modifications to the PTL source code after the 1.6 release.

Before, some patching to the loops was necessary to disable PTL's pre-fetching as it conflicted with the pre-fetching done by FFCV. Now, PTL avoids pre-fetching whenever it is possible.

From my own benchmarks, the speed up from FFCV applies perfectly when used with PTL. We'll be releasing public benchmarks in the future.

We'll also explore making or internal data processing more flexible so that maybe we can provide an opinionated `LightningDataModule` integration that makes switching existing `LightningDataModule`s easier.

A caveat I've noticed are that you'll want to remove the `ToDevice(device)` from the label pipeline if it's the last operation in your transformations ([example](https://github.com/libffcv/ffcv-imagenet/blob/main/train_imagenet.py#L236)) as PTL already takes care of moving the tensors to the correct devices.

Features like fault-tolerance do not support FFCV.

Note that there's a lot of surface for bugs here and I think most of the battle-testing has been done on map-style datasets. If you encounter any memory or speed issues, please open a separate issue and you can ping me on it.

Cheers!

siddk:
Ok - awesome! Thanks so much!

siddk:
Second question (sorry for the delayed follow-up); is the recommended pattern to couple FFCV dataloading **within** a LightningDataModule, or should I just pass the FFCV loaders directly to the Trainer?

carmocca:
It's up to you, but both will work.

[...]

PR #11606, opened 2022-01-25
Support no pre-fetching

carmocca:
## What does this PR do?

Adds support for no-prefetching.
This was proposed by @justusschock in https://github.com/PyTorchLightning/pytorch-lightning/pull/11516#discussion_r786558553.
Also means users do not need to patch the Loops to disable it as it is done by some `ffcv` users.

One drawback is that `is_last_batch` will not be correctly set in this case.

Part of #11538 

None

## Before submitting

## PR review

### DIFF
--- a/pytorch_lightning/utilities/fetching.py
+++ b/pytorch_lightning/utilities/fetching.py
@@ -237,18 +237,28 @@ class DataFetcher(AbstractDataFetcher):
 
     def fetching_function(self) -> Tuple[Any, bool]:
         if self.batches:
+            # we pre-fetched, consume one
             batch = self.batches.pop(0)
-        else:
-            # empty iterator, no prefetching done
-            raise StopIteration
-        if not self.done:
-            assert self.dataloader_iter is not None
             try:
+                # and replace it
                 self._fetch_next_batch(self.dataloader_iter)
             except StopIteration:
+                self.done = not self.batches
+
+        elif not self.done:
+            # we did not prefetch at all
+            try:
+                self._fetch_next_batch(self.dataloader_iter)
+                batch = self.batches.pop(0)
+            except StopIteration as e:
                 self.done = True
+                raise e
+
+        else:
+            raise StopIteration
+
         self.wait()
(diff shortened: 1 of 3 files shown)


tests/utilities/test_fetching.py
```diff
@@ -40,28 +40,32 @@ def __iter__(self):
             yield 2
             yield 3
 
-    for prefetch_batches in range(1, 5):
```
carmocca, line 43: The only relevant change in this test is that range now starts from 0.

pytorch_lightning/utilities/fetching.py
```diff
@@ -236,19 +234,27 @@ def prefetching(self) -> None:
                 break
 
     def fetching_function(self) -> Tuple[Any, bool]:
+        assert self.dataloader_iter is not None
```
tchaton, line 237: Mind adding more comments on this method. This is quite core to the fetching and would benefit from line comments.

pytorch_lightning/utilities/fetching.py
```diff
@@ -236,19 +234,27 @@ def prefetching(self) -> None:
...
-            assert self.dataloader_iter is not None
             try:
+                # and replace it
                 self._fetch_next_batch(self.dataloader_iter)
             except StopIteration:
+                self.done = not self.batches
+        elif not self.done:
+            # we did not prefetch at all
+            try:
+                self._fetch_next_batch(self.dataloader_iter)
```
tchaton, line 249: This comment is confusing as it is not fully accurate. You might have actually pre-fetched.

tchaton (approved):
Super super neat !

pytorch_lightning/utilities/fetching.py
carmocca: If prefetching was done (in `prefetching`) then the `if self.batches` block above will be the one to always run as it will also replace the batch consumed.

This block is an elif so it will not execute when the above block does.

rohitgr7 (approved):

ananthsub (approved):

PR #11614, opened 2022-01-25
Add default args in Trainer methods doc

rasbt:
## What does this PR do?

Added the default values in the Trainer method parameter docstrings where it was missing. I think this makes reading the docs easier.

Btw. maybe it would be better to have a consistent rule/style for this? E.g., I like how scikit-learn is doing it

```    
    Parameters
    ----------
    estimator : estimator instance
        An unfitted estimator.

    n_features_to_select : "auto", int or float, default='warn'
        If `"auto"`, the behaviour depends on the `tol` parameter:

        - if `tol` is not `None`, then features are selected until the score
          improvement does not exceed `tol`.
        - otherwise, half of the features are selected.

        If integer, the parameter is the absolute number of features to select.
        If float between 0 and 1, it is the fraction of features to select.

        .. versionadded:: 1.1
           The option `"auto"` was added in version 1.1.

        .. deprecated:: 1.1
           The default changed from `None` to `"warn"` in 1.1 and will become
           `"auto"` in 1.3. `None` and `'warn'` will be removed in 1.3.
           To keep the same behaviour as `None`, set
           `n_features_to_select="auto" and `tol=None`.

    tol : float, default=None
        If the score is not incremented by at least `tol` between two
        consecutive feature additions or removals, stop adding or removing.
        `tol` is enabled only when `n_features_to_select` is `"auto"`.
```

but maybe we can discuss this in a separate issue/PR

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -230,7 +230,7 @@ class Trainer(
                 that only one process at a time can access them.
                 Default: ``False``.
 
-            benchmark: If ``True`` enables cudnn.benchmark.
+            benchmark: If ``True``, enables cudnn.benchmark.
                 Default: ``False``.
 
             callbacks: Add a callback or list of callbacks.
@@ -454,7 +454,7 @@ class Trainer(
                 In 'max_size_cycle' mode, the trainer ends one epoch when the largest dataset is traversed,
                 and smaller datasets reload when running out of their data. In 'min_size' mode, all the datasets
                 reload when reaching the minimum length of datasets.
-                Default: ``max_size_cycle``.
+                Default: ``"max_size_cycle"``.
 
             stochastic_weight_avg: Whether to use `Stochastic Weight Averaging (SWA)
                 <https://pytorch.org/blog/pytorch-1.6-now-includes-stochastic-weight-averaging/>`_.
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -205,8 +205,10 @@ class Trainer(
                     and will be removed in v1.7.0. Please use the ``strategy`` argument instead.
 
             accumulate_grad_batches: Accumulates grads every k batches or as set up in the dict.
+                Default: ``None``.
 
             amp_backend: The mixed precision backend to use ("native" or "apex").
+                Default: ``'native''``.
 
             amp_level: The optimization level to use (O1, O2, etc...). By default it will be set to "O2"
                 if ``amp_backend`` is set to "apex".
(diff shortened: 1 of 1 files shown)


awaelchli:
Currently, the sphinx doc generator adds the default values automatically to the rendered docs page based on the default values specified in the signature. We have discussed manually adding these for each parameter in the past, like you did here. The maintenance effort is higher: when a default value changes, it needs to be updated manually in the docstring too. But considering that the trainer defaults don't change very often, and it is sometimes hard to read the docs, I am fine with either way :)

For the rest of the code base, I would let sphinx handle the defaults.

rasbt:
Thanks for the note @awaelchli 

I totally agree with you regarding the maintenance efforts. If it can be added automatically, this would be awesome. I was wondering if this is possible with Sphinx (I know that scikit-learn for example still adds the defaults manually, and I thought there was maybe a limitation in Sphinx)?

Right now it looks like that the default don't show up in the parameter docstrings:

Screen Shot 2022-01-26 at 7 03 41 AM

They are only in the call sig above: 
Screen Shot 2022-01-26 at 7 05 02 AM

Or is this what you mean? As a user, I find it very painful to scroll up and down and look for the defaults. It would be nice to have the defaults mentioned in the parameter descriptions. But maybe this shouldn't be done manually like you said, and we could think about a more maintainable solution automating this.

tchaton (approved):
Nice addition! Thanks. Even if the maintenance effort is higher, I believe the Trainer should be fully documented.

rasbt:
Just to bump this up. Any ideas how we should best go about this? I think ideally we want some automatic solution. The question is, does it exist? Maybe not in Sphix, but it might be possible to write a parser doing this. But maybe more a long- rather than short-term solution. 

Looking at other well maintained packages like scikit-learn, it looks like people still do this manually.

awaelchli (approved):

awaelchli:
@rasbt Automatic solutions won't work well because defaults are often chosen at runtime because they depend on the setting of other arguments. This is especially true for the Trainer which has so many arguments. Another example is #11950 where an argument gets set to default None simply to detect internally whether the user has explicitly set a value or not, but the value None is never used in the end. We will always have to describe such cases in words :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -213,23 +215,29 @@ def __init__(
...
+                Default: ``False``.
 
             auto_select_gpus: If enabled and ``gpus`` is an integer, pick available
                 gpus automatically. This is especially useful when
                 GPUs are configured to be in "exclusive mode", such
                 that only one process at a time can access them.
+                Default: ``False``.
 
-            benchmark: If true enables cudnn.benchmark.
+            benchmark: If ``True`` enables cudnn.benchmark.
```
akihironitta, line 233:
```suggestion
            benchmark: If ``True``, enables cudnn.benchmark.
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -410,14 +448,17 @@ def __init__(
 
             move_metrics_to_cpu: Whether to force internal logged metrics to be moved to cpu.
                 This can save some gpu memory, but can make training slower. Use with attention.
+                Default: ``False``.
 
             multiple_trainloader_mode: How to loop over the datasets when there are multiple train loaders.
                 In 'max_size_cycle' mode, the trainer ends one epoch when the largest dataset is traversed,
                 and smaller datasets reload when running out of their data. In 'min_size' mode, all the datasets
                 reload when reaching the minimum length of datasets.
+                Default: ``max_size_cycle``.
```
akihironitta, line 457:
```suggestion
                Default: ``"max_size_cycle"``.
```

akihironitta (approved):

rasbt (commented):
Haha, was just about to apply those. LGTM. Thanks, Adrian.

Issue #2727, opened 2020-07-27
Hydra configs with multi GPU DDP training in Pytorch Lightning


topshik:
As far as I understand DDP backend runs my training script from beginning for each GPU that I use. Is there a way to avoid creating different hydra `output` directories in each of the scripts? Should I block somehow every process except one with local rank 0? In my case I'm saving model checkpoints and .yaml file to default hydra output directory, but config file is copied twice and checkpoints are saved once. Anyways, spawning too many of directories is not convenient.

What can I do?

#### Code

```
@hydra.main(config_path="train-config.yaml", strict=False)
def train(config: DictConfig) -> None:
    config.hydra_base_dir = os.getcwd()
    original_wd = hydra.utils.get_original_cwd()
    os.chdir(original_wd)

    checkpoint_callback = ModelCheckpoint(
        filepath=config.hydra_base_dir,
        save_top_k=3,
        verbose=True,
        monitor="val_loss",
        mode="min",
    )
    shutil.copy2("train-config.yaml", os.path.join(config.hydra_base_dir, "train-config.yaml"))

    wandb_logger = WandbLogger(
        offline=False,
    )

    model = MyModel(config)

    trainer = pl.Trainer(
        max_epochs=config.train.max_epochs,
        gpus=config.train.n_gpu,
        auto_select_gpus=True,
        distributed_backend="ddp",
        checkpoint_callback=checkpoint_callback,
        logger=wandb_logger,
    )

    trainer.fit(model)
```

#### What's your environment?

 - OS: Ubuntu 18.04
- Conda, Python 3.7.7
- hydra-core==0.11.3
- pytorch-lightning==0.8.5
- wandb==0.9.3

topshik:
I've just found out that process local rank can be accessed via `local_rank = os.environ.get("LOCAL_RANK", 0)`, because this is how lightning handles it under cover. Seems like there is needed some clarification on how to work with different DDP processes in the docs.

Borda:
cc: @yukw777 @omry

omry:
You could configure Hydra run dir via the command line or your config file to be whatever you want it to be, see [this](https://hydra.cc/docs/next/configure_hydra/workdir).

However, I think the right approach to do DDP training with Hydra is to use multirun. With multirun - each running script gets it's own subdirectory under the primary working directory by design and not by accident.
I didn't see anyone doing it yet and I do not know how it will work with PL.
[...]

PR #11617, opened 2022-01-25
Improving Hydra+DDP support

### LINKED ISSUE #11300: DDP with Hydra multirun doesn't work when dirpath in checkpoint callback is specified
## 🐛 Bug

Running DDP with Hydra multirun ends up with "Killed" error message when launching the second task:
```
Epoch 0    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0/939 0:00:00 • -:--:-- 0.00it/s [W reducer.cpp:1158] Warning: find_unused_parameters=True was specified in DDP constructor, but did not find any unused parameters in the forward pass. This flag results in an extra traversal of the autograd graph every iteration,  which can adversely affect performance. If your model indeed never has any unused parameters in the forward pass, consider turning this flag off. Note that this warning may be a false positive if your model has flow control causing later iterations to have unused parameters. (function operator())
[W reducer.cpp:1158] Warning: find_unused_parameters=True was specified in DDP constructor, but did not find any unused parameters in the forward pass. This flag results in an extra traversal of the autograd graph every iteration,  which can adversely affect performance. If your model indeed never has any unused parameters in the forward pass, consider turning this flag off. Note that this warning may be a false positive if your model has flow control causing later iterations to have unused parameters. (function operator())
Epoch 0    ━━━━━━━━━━━━━━━━ 939/939 0:00:13 •        70.53it/s loss: 0.142      
                                    0:00:00                    v_num:           
[2022-01-03 15:21:38,513][src.train][INFO] - Starting testing!
[2022-01-03 15:21:38,514][pytorch_lightning.utilities.distributed][INFO] - Restoring states from the checkpoint path at /home/user/lightning-hydra-template/logs/multiruns/2022-01-03/15-21-17/0/checkpoints/epoch_000.ckpt
[2022-01-03 15:21:38,535][pytorch_lightning.accelerators.gpu][INFO] - LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0,1]
[2022-01-03 15:21:41,523][HYDRA]        #1 : trainer.max_epochs=1 datamodule.batch_size=64 trainer.gpus=2 +trainer.strategy=ddp
Killed
```

I experience this ONLY when passing the `dirpath` parameter to checkpoint callback:
```python
ModelCheckpoint(dirpath="checkpoints/")
```

Tested for lightning v1.5.7. I believe this issue wasn't around in one of the previous releases. 

This probably has something to do with the way hydra changes working directory for each new run - the directory for storing checkpoints also gets changed. If I remember correctly, there was some workaround implemented in lightning which made DDP possible despite that.

jgbos:
## What does this PR do?

There are documented issues for Hydra+DDP  in #11300 and #2727.   This PR attempts to fix these issues by redefining how a process is spawned when Hydra is used.

Fixes #11300

### Problem

Current approach is defined here: https://github.com/PyTorchLightning/pytorch-lightning/blob/fe34bf2a653ebd50e6a3a00be829e3611f820c3c/pytorch_lightning/strategies/ddp.py#L233

This PR addresses the issue of running with Hydra `multirun`.  For example, lets say we have the following Hydra app:

```python
@hydra.main(config_path=None, config_name="myconfig", version_base="1.1")
def task_fn(cfg):
    trainer = Trainer(gpus=2, strategy="ddp")
    model = BoringModel()
    trainer.fit(model)

if __name__ == "__main__":
    task_fn()
```

We can execute a `multirun` job with this app using the following command:
```bash
python script.py foo=1,2 --multirun
```

This command will attempt to launch 2 jobs sequentially: one with `foo=1` and one with `foo=2`. For the first job, `foo=1`, PL launchers a normal job that begins execution while the second job is spawned with a subprocess using the following command derived from `sys.argv`:
```bash
python script.py foo=1,2 --multirun hydra.run.dir=<os_cwd> hydra.job.name=train_ddp_process_<local_rank>
```
This will spawn a new `mutlirun` job instead of running a normal job with `foo=2`. The command should be
```bash
python script.py foo=2 hydra.run.dir=<os_cwd>
```

### Solution

Every Hydra process will save the reproducible configuration of the job in a `config.yaml` file located in the `hydra.output_subdir`  experiment directory.  Using Hydra's CLI, we can execute the app with same configuration as the current experiment by defining `--config_path, -cp` and `--config_name, -cn`:
```bash
python script.py -cp <path to hydra output dir> -cn config.yaml hydra.run.dir=<os_cwd> 
```

Here `config.yaml` contains the value for `foo` appropriate for the current `multirun` job.  I've outlined the support for `multirun`, but this should support any Hydra application launched from the command line.

### Lingering Issue Not Solved

In order to run `multirun` on a local machine the user must add additional code to their task function before launching the next job.  This code, shown below, will destroy all distributed processes and remove PL related environment variables.  Without this code the `multirun` job will hang after the first job.

```python
@hydra.main(config_path=None, config_name="myconfig", version_base="1.1")
def task_fn(cfg):
    trainer = Trainer(gpus=2, strategy="ddp")
    model = BoringModel()
    trainer.fit(model)
    trainer.test(model)

    # Need to do this in addition to Lightning shutting down the
    # distributed processes in order to run a multirun loop with hydra
    if dist.is_initialized():
        dist.destroy_process_group()

    os.environ.pop("LOCAL_RANK", None)
```

### Thoughts

This solution should help most people, but IMHO the current method of spawning using `sys.argv` is not robust.  It would be nice to be able to execute a PL application similar to how I implemented the Hydra solution.  This would require dynamically creating a configurations for the application. The use of `save_hyperparameters` is nice but is only done inside the creation of a `LightningModule`, it would be nice to create a description of the model at a higher level — outside of the module. By defining the description outside of the module, one could use a similar approach as the Hydra solution above for PL applications.   I would recommend taking a look at [hydra-zen](https://github.com/mit-ll-responsible-ai/hydra-zen) (FYI, I'm a contributor). 

None

## Before submitting

Personal TODO:
- [x] Add test checking processes use the correct parameters when spawned

PL TODO:


pytorch_lightning/strategies/ddp.py
```diff
@@ -232,9 +232,27 @@ def _call_children_scripts(self):
             cwd: Optional[str] = None
             if _HYDRA_AVAILABLE:
                 if HydraConfig.initialized():
-                    cwd = get_original_cwd()
-                    os_cwd = f'"{os.getcwd()}"'
-                    command += [f"hydra.run.dir={os_cwd}", f"hydra.job.name=train_ddp_process_{local_rank}"]
+                    orig_cwd = get_original_cwd()
```
tchaton, line 235: Can we move this code to its own method ?

tests/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,157 @@
...
+from pytorch_lightning.utilities.imports import _HYDRA_AVAILABLE
+from tests.helpers.runif import RunIf
+
+if _HYDRA_AVAILABLE:
+    from omegaconf import OmegaConf
+
+
+# fixture to run in a clean temporary directory
+@pytest.fixture()
+def cleandir():
```
tchaton, line 19: You could use tmpdir provided by pytest there.

tests/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,157 @@
...
+
+@hydra.main()
+def task_fn(cfg):
+    trainer = Trainer(gpus=cfg.gpus, strategy=cfg.strategy, fast_dev_run=True)
+    model = BoringModelGPU()
+    trainer.fit(model)
+
+    # Need to do this in addition to Lightning shutting down the
+    # distributed processes in order to run a multirun loop with hydra
+    if dist.is_initialized():
```
tchaton, line 77: Can we detect the user is using multirun with Hydra and do this automatically within the DDP Plugin ?

tests/strategies/test_ddp_hydra_support.py
jgbos: oh cool, I wasn't aware of that, I'll update.

tests/strategies/test_ddp_hydra_support.py
jgbos: I'll look into it, but I think you cannot.

jgbos: I cannot find any mechanism to detect multirun.  Each job is a standard run job and only the sweeper is aware of the multirun.  Maybe PL could have an option to fully shutdown distributed and clean the environment?

rsokl: @jgbos  could you inspect `HydraConf.sweep`? I think `HydraConf.sweep.dir is not MISSING` might be a reliable signal that multirun is happening

jgbos: Unfortunately, as far as I can tell both `run.dir` and `sweep.dir` are set no matter the run mode.

jgbos: Oh, I may have found it.  We can check `OmegaConf.is_missing(hydra_cfg.job, "num")`.

jgbos: @tchaton is the `teardown` the appropriate place to do this?

pytorch_lightning/strategies/ddp.py
```diff
@@ -520,3 +548,27 @@ def teardown(self) -> None:
             self.lightning_module.cpu()
             # clean up memory
             torch.cuda.empty_cache()
+
+        if _HYDRA_AVAILABLE:
+            if HydraConfig.initialized():
```
tchaton, line 553: Would you mind adding this within its own method ?

pytorch_lightning/strategies/ddp.py
```diff
@@ -245,6 +244,35 @@ def _call_children_scripts(self):
 
         self._rank_0_has_called_call_children_scripts = True
 
+    def _get_spawn_command_for_hydra(self, command: List[str], local_rank: str) -> Tuple[str, str]:
```
tchaton, line 247: Might be worth moving those utilities under /utilities/hydra.py

tchaton (commented):
Looks really neat! Mind moving the hydra-specific logic to its own utility file?

jgbos:
@tchaton I'm working through some issues with `teardown`.  It appears there are issues with the process hanging on `torch.distributed.destroy_process_group()`, but if I do it in my task function outside of `trainer.fit` I do not have any issues.  Is there a different spot to put the code currently in `teardown`?

pytorch_lightning/utilities/hydra.py
```diff
@@ -0,0 +1,71 @@
...
+
+from pytorch_lightning.utilities import _HYDRA_AVAILABLE
+
+if _HYDRA_AVAILABLE:
+    from hydra.core.hydra_config import HydraConfig
+    from hydra.utils import get_original_cwd
+    from omegaconf import OmegaConf
+
+
+def get_ddp_spawn_command_for_hydra(command: List[str], local_rank: str) -> Tuple[str, str]:
```
carmocca, line 15: Local rank is an int

```suggestion
def get_ddp_spawn_command_for_hydra(command: List[str], local_rank: int) -> Tuple[str, str]:
```

carmocca, line 15: The return annotation does not look right

```suggestion
def get_ddp_spawn_command_for_hydra(command: List[str], local_rank: int) -> Tuple[List[str], Optional[str]]:
```

pytorch_lightning/utilities/hydra.py
```diff
@@ -0,0 +1,71 @@
...
+    """Modifies the DDP spawn command to support Hydra initiated processes.
+
+    If Hydra is initialized:   1) Set `cwd` to the hydra working directory   2) Use the stored configuration in
+    `hydra_cfg.output_subdir / config.yaml` to spawn a new child
```
carmocca, line 19: These can be comments in the code directly

pytorch_lightning/utilities/hydra.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    cwd: Optional[str] = None
+    if HydraConfig.initialized():
+        orig_cwd = get_original_cwd()
+        cwd = os.getcwd()
+        os_cwd = f'"{cwd}"'  # this is needed to handle characters like `=` in the directory name
+
+        hydra_cfg = HydraConfig.get()
+        hydra_output = os.path.join(os.path.relpath(cwd, orig_cwd), hydra_cfg.output_subdir)
```
carmocca, line 29: Why is `relpath` necessary?

pytorch_lightning/utilities/hydra.py
```diff
@@ -0,0 +1,71 @@
...
+                "NODE_RANK",
+                "WORLD_SIZE",
+                "MASTER_ADDR",
+                "MASTER_PORT",
+                "PL_GLOBAL_SEED",
+                "PL_SEED_WORKERS",
+            )
+
+            for name in envs:
+                os.environ.pop(name, None)
```
carmocca, line 71: I guess you added these because of:

https://github.com/PyTorchLightning/pytorch-lightning/blob/32e7d32956e1685d36f2ab0ca3770baa2f76ce10/tests/conftest.py#L46-L82

But I don't think you need to pop them if Hydra did not set them.

tests/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,210 @@
...
+@pytest.mark.usefixtures("cleandir")
+@pytest.mark.parametrize("gpus", [1, 2])
+@pytest.mark.parametrize("subdir", [None, "dksa", ".hello"])
+def test_ddp_with_hydra_runjob(gpus, subdir):
+    # Save script locally
+    with open("temp.py", "w") as fn:
+        fn.write(script)
+
+    # Run CLI
+    cmd = [sys.executable, "temp.py", f"+gpus={gpus}", '+strategy="ddp"']
```
carmocca, line 100: Can you use the future format?

instead of `gpus=2` -> `accelerator="gpu", devices=2`

or even

`accelerator="auto", devices=2`

so the test also runs on CPU

tests/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,210 @@
...
+
+if __name__ == "__main__":
+    task_fn()
+"""
+
+
+@RunIf(skip_windows=True, min_gpus=2)
+@pytest.mark.skipif(not _HYDRA_AVAILABLE, reason="Hydra not Available")
+@pytest.mark.usefixtures("cleandir")
+@pytest.mark.parametrize("gpus", [1, 2])
```
carmocca, line 92: Note that setting `gpus=1` with the DDP strategy will already run distributed. I think we can get away with just testing that (it's faster!)

pytorch_lightning/utilities/hydra.py
jgbos: ah yes, it comes from the loop iterator.  for some reason thought it was coming from the environment variable

pytorch_lightning/utilities/hydra.py
jgbos: I'm looking into that, I don't think it will be needed.

tests/strategies/test_ddp_hydra_support.py
jgbos: hmm, technically I shouldn't do `gpus=1` because this logic only gets executed for `gpus > 1`.

awaelchli:
@rohitgr7 We could potentially have a hydra specific launcher after #11643 is finalized. In that sense, I like the approach of this PR of creating a utility function encapsulating the hydra command.

rohitgr7:
> @rohitgr7 We could potentially have a hydra specific launcher after #11643 is finalized. In that sense, I like the approach of this PR of creating a utility function encapsulating the hydra command.

@awaelchli pushed hydra launcher here. is this something you had in mind?

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
```diff
@@ -0,0 +1,82 @@
...
+
+from pytorch_lightning.strategies.launchers.subprocess_script import _SubprocessScriptLauncher
+from pytorch_lightning.utilities import _HYDRA_AVAILABLE
+
+if _HYDRA_AVAILABLE:
+    from hydra.core.hydra_config import HydraConfig
+    from hydra.utils import to_absolute_path
+
+
+class _SubprocessScriptHydraLauncher(_SubprocessScriptLauncher):
```
carmocca, line 15: I would put this at the beginning

```suggestion
class _HydraSubprocessScriptLauncher(_SubprocessScriptLauncher):
```

Or even

```suggestion
class _HydraLauncher(_SubprocessScriptLauncher):
```

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
```diff
@@ -0,0 +1,82 @@
...
+    """Hydra Launcher subclasses `_SubprocessScriptLauncher to manage Hydra specific commands when running under
+    Hydra environment."""
```
carmocca, line 17: Simpler description, it seems redundant, and I wouldn't mention the subclass necessarily:
```suggestion
    """Hydra Launcher to support Hydra commands."""
```

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
```diff
@@ -0,0 +1,82 @@
...
+    def _get_launch_command(self, command: List[str], local_rank: int) -> Tuple[List[str], Optional[str]]:
+        """Modifies the command to support Hydra initiated processes."""
+        if not HydraConfig.initialized():
+            return command, None
+
+        # If Hydra is initialized:
+        #   1) Set `cwd` to the hydra working directory
+        #   2) Use the stored configuration in `hydra_cfg.output_subdir / config.yaml` to spawn a new child
+
+        cwd = os.getcwd()
```
carmocca, line 31: Does the `cwd` need to be part of this?

AFAIK neither of the implementations do anything special with it.

Isn't returning `None` equal to returning `os.getcwd()`?

At least, in the previous code, this was `hydra.utils.get_original_cwd()`, not sure if this has been removed by mistake.

carmocca (commented):
This looks good to me.

cc @jgbos to double check the launcher changes.

pytorch_lightning/utilities/hydra.py
carmocca: Any update on this?

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
rohitgr7: I noticed, this too. This was done before I integrated the launcher, so need to confirm with @jgbos .

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
rohitgr7: `_HydraSubprocessScriptLauncher` looks better.

awaelchli: I like the idea of a hydra specific launcher. 
However, my suggestion is not to subclass the regular process launcher. We are moving away from that pattern now in the strategies too (#11863). 

I suggest to put both launchers into the same file and simply share code through logical pieces of utility function in that same file.

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
```diff
@@ -0,0 +1,82 @@
+import os
```
awaelchli, line 1: I propose to move this to subprocess_script.py for the reason outlined in my other comment.

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
```diff
@@ -0,0 +1,82 @@
...
+            f"hydra.run.dir={os_cwd}",
+            f"hydra.job.name=train_ddp_process_{local_rank}",
+        ]
+        return command, cwd
+
+    def teardown(self) -> None:
+        if HydraConfig.initialized():
+            # shutdown any distributed process groups
+            if torch.distributed.is_initialized():
+                torch.distributed.destroy_process_group()
```
awaelchli, line 72: I don't understand this. 

teardown runs when the fit() ends, not when the process finishes. If we do this, any call to trainer.fit/validate/test afterwards would fail to work in a distributed setting. 

this is the reason we don't do this for the regular script launcher.

pytorch_lightning/strategies/strategy.py
```diff
@@ -442,6 +442,9 @@ def teardown(self) -> None:
         optimizers_to_device(self.optimizers, torch.device("cpu"))
         self.precision_plugin.teardown()
 
+        if self.launcher:
```
awaelchli, line 445: careful, above in the docstring: "teardown the training process"
Bad wording, we don't mean the process as in "subprocess" here, rather the "training procedure/run"

pytorch_lightning/strategies/launchers/base.py
```diff
@@ -34,3 +34,6 @@ def is_interactive_compatible(self) -> bool:
     @abstractmethod
     def launch(self, function: Callable, *args: Any, **kwargs: Any) -> Any:
         """Launches the processes."""
+
+    def teardown(self) -> None:
+        """This method is called to teardown the launcher."""
```
awaelchli, line 39: we need to be carful. how do we precisely define a teardown for a launcher? 
If you look at where the process boundaries are when comparing ddp vs ddp spawn, it's vastly different. 

Can we just remove it from the base class for now and call it at the end of the launch() method internally here on the specific implementation?

tests/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,143 @@
...
+@hydra.main()
+def task_fn(cfg):
+    trainer = Trainer(accelerator="auto", devices=cfg.devices, strategy=cfg.strategy, fast_dev_run=True)
+    model = BoringModelGPU()
+    trainer.fit(model)
+    # make sure teardown executed
+    assert not torch.distributed.is_initialized()
+    assert "LOCAL_RANK" not in os.environ
+
+
```
awaelchli, line 75: can you check what happens here if you insert a trainer.validate() here after fit? 
I would be surprised if that works

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
rohitgr7: yes, you are right. the current process group is destroyed and not process further.

tests/strategies/test_ddp_hydra_support.py
rohitgr7: yep, doesn't work since process groups are destroyed after fit

jgbos:
Just a heads up to all, I've been on vacation for the past week.  I'll look at all the changes and try to respond to the comments this week.

pytorch_lightning/utilities/hydra.py
rohitgr7: maybe we can delete the progress group and remove the env variables when the object is deleted inside `__del__`, so make it work with two consecutive `trainer.method` calls?

rohitgr7: but I think it will be a problem again with such an example in the same script:
```py
trainer = Trainer(ddp)
trainer.fit()
trainer.validate()

trainer = Trainer(ddp)
trainer.fit()
```

jgbos: I'll take another look if this is really required.  I'm still trying to figure out how to make this PR robust for Hydra multirun.

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
jgbos: So it looks like this is in the wrong place to destroy the process groups.  I've actually started putting this piece of code in the `setup` method.  Without it, Lightning hangs (deadlocks?) on initiating the process group for Hydra multiruns.  It's been frustrating to debug this issue and determine why we need it.

pytorch_lightning/strategies/launchers/subprocess_script.py
```diff
@@ -156,3 +153,87 @@ def _check_can_spawn_children(self) -> None:
...
+            # this fails for `python -m pdb -m a.b.c <args>`
+            command_no_args = command[:3]
+
+        command = command_no_args
+
+        # run the Hydra job using the current job configuration
+        # - typically located in:
+        #        RUN MODE: hydra.run.dir/.hydra/config.ayml
+        #        MULTIRUN MODE: hydra.sweep.dir/hydra.sweep.subdir/.hydra/config.yaml
+        command += ["-cp", hydra_output, "-cn", "config.yaml"]
```
Jasha10, line 191: Using `-cn config.yaml` will produce an untyped config, i.e. a config (DictConfig or ListConfig) that is not backed by a dataclass.
This may or may not be a problem, depending on the use case. For example, using an untyped config means that `Enum` values will be treated by OmegaConf as strings (rather than as values of the given Enum subclass), and structured-config methods such as `OmegaConf.to_object` will not produce the expected result.

This is related to [hydra#1805](https://github.com/facebookresearch/hydra/issues/1805).

jgbos:
So I'm attempting to revive this PR.  I got everything running but there are still some outstanding issues to address.  The two main issues:
  - Hydra now comes with `version_base`, we need to handle `1.1` and `1.2`
  - The Hydra task functions needs to clean up the environment and shutdown distributed in order to execute multirun (at least using basic sweeper and launchers)

Here is the basic task function (see `tests/tests_pytorch/strategies/test_ddp_hydra_support.py`):
```python
@hydra.main(config_path=None, version_base="1.1")
def task_fn(cfg):
    trainer = Trainer(accelerator="auto", devices=cfg.devices, strategy=cfg.strategy, fast_dev_run=True)
    model = BoringModelGPU()
    trainer.fit(model)
    trainer.test(model)

    # the code below is needed for multirun only
    if torch.distributed.is_initialized():
        torch.distributed.destroy_process_group()

    os.environ.pop("LOCAL_RANK", None)
```

pytorch_lightning/strategies/launchers/subprocess_script_hydra.py
jgbos: much delayed, but you are correct on this

tests/strategies/test_ddp_hydra_support.py
jgbos: yeah, I removed that teardown logic for now.  It may have to be left to the user's task function for multirun

jgbos:
I added support for Hydra Re-Run (https://hydra.cc/docs/1.2/experimental/rerun/).

I'm trying to figure out why tests are failing in actions but do fine on my machine.  Not sure I'm understanding how to setup tests with the latest updates.

pytorch_lightning/strategies/launchers/subprocess_script.py
jgbos: I added support to use `experimental-rerun`

jgbos:
I'm thinking now that `_SubprocessScriptLauncher` should have a method to compute the subprocess command:

```python
def _subprocess_cmd(self, local_rank: int) -> Sequence[str]:
    ...
```

Then we could have the base subprocess launcher (no hydra) and the hydra specific launcher.  We could then make a separate strategy `hydra_ddp`.

Any thoughts on this?

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,146 @@
...
+def task_fn(cfg):
+    trainer = Trainer(accelerator="auto", devices=cfg.devices, strategy=cfg.strategy, fast_dev_run=True)
+    model = BoringModelGPU()
+    trainer.fit(model)
+    trainer.test(model)
+
+    if torch.distributed.is_initialized():
+        torch.distributed.destroy_process_group()
+
+    os.environ.pop("LOCAL_RANK", None)
```
carmocca, line 77: This are the hard limitations you haven't been able to solve, correct?

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,146 @@
...
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+from pytorch_lightning.utilities.imports import _RequirementAvailable
+from tests_pytorch.helpers.runif import RunIf
+
+_HYDRA_AVAILABLE = bool(_RequirementAvailable("hydra-core"))
```
carmocca, line 12: Why not use the variable from `src/pytorch_lightning/strategies/launchers/subprocess_script.py`?

carmocca:
@jgbos You chose to get rid of the launcher subclass, right?

> We could then make a separate strategy hydra_ddp.

Why would we want this? This should be invisible to users

jgbos:
Hi @carmocca , thanks for the comments.  wrt to the sublcassing, let me clarify. I wondering if we should update `_SubprocessScriptLauncher` to have a `_subprocess_cmd` method:

```python

class _SubprocessScriptLauncher(_Launcher):
    ...

    def _subprocess_cmd(self, local_rank: int) -> Sequence[str]:
        if __main__.__spec__ is None:  # pragma: no-cover
            return [sys.executable, os.path.abspath(sys.argv[0])] + sys.argv[1:]
        else:
            return [sys.executable, "-m", __main__.__spec__.name] + sys.argv[1:]

    def _call_children_scripts(self) -> None:
        ...
        for local_rank in range(1, self.num_processes):
            ...
            cmd = self.subprocess_cmd(local_rank)
            subprocess.Popen(command, env=env_copy)
            ....

```

Then we could make specific subprocess launchers that can be easily customized by the user, for instance Hydra:

```python
class _HydraSubprocessScriptLauncher(_SubprocessScriptLauncher):

    def _subprocess_cmd(self, local_rank: int) -> Sequence[str]:
        # return the subprocess command
        ...
```

Since this is a new launcher we'd have to make a new `DDPStrategy`:

```python
class HydraDDPStrategy(DDPStrategy):
    strategy_name = "hydra_ddp"

            self._launcher = _HydraSubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
            self._rank_0_will_call_children_scripts = True

    @classmethod
    def register_strategies(cls, strategy_registry: Dict) -> None:
        strategy_registry.register(
            "hydra_ddp_find_unused_parameters_false",
            cls,
            description="Hydra DDP Strategy with `find_unused_parameters` as False",
            find_unused_parameters=False,
        )
        strategy_registry.register(
            cls.strategy_name,
            cls,
            description=f"{cls.__class__.__name__}",
        )
```

I like the approach of explicitly defining the strategy that supports Hydra (and other customizable launchers).  I haven't pushed this change because I was unsure of PL's opinion on this.

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
jgbos: No I have not solved this issue due to the dynamics of `torch.distributed`, PL, and Hydra `multirun`.  I cannot put those commands in `teardown` because that would negate doing `fit` then `test`.

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
jgbos: Are you suggesting I use `from pytorch_lightning.strategies.launchers.subprocess_script import _HYDRA_AVAILABLE`?

carmocca: Exactly!

carmocca:
The `subprocess_cmd` refactor makes sense to me. But I would definitely avoid the need to introduce a separate strategy. The fact that Hydra needs a custom launch command does not mean it needs a custom strategy. We should select the correct launcher automatically.

cc @awaelchli

jgbos:
Sounds good @carmocca .  I extracted the subprocess command but I still have it automatically selecting based on whether Hydra was initialized.

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,192 @@
...
+    # Run CLI
+    run_process(
+        [
+            sys.executable,
+            "temp.py",
+            "-cp",
+            ".",
+            "-cn",
+            "config.yaml",
+            f"+devices={devices}",
```
carmocca, line 184: Devices does not change

```suggestion
            "+devices=2",
```

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,192 @@
...
+    with open("config.yaml", "w") as fn:
+        fn.write(yaml_file)
+
+    # create fake multirun params based on `num_jobs`
+    fake_param = "+foo="
+    devices = 2
+    for i in range(num_jobs):
+        fake_param += f"{i}"
+        if i < num_jobs - 1:
+            fake_param += ","
```
carmocca, line 173:
```suggestion
    fake_param = "+foo=" + ",".join(str(i) for i in range(num_jobs))
```

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,192 @@
...
+@pytest.mark.skipif(not _HYDRA_AVAILABLE, reason="Hydra not Available")
+@pytest.mark.usefixtures("cleandir")
+@pytest.mark.parametrize("num_jobs", [1, 2])
+def test_ddp_with_hydra_multirunjob(num_jobs):
+    # Save script locally
+    with open("temp.py", "w") as fn:
+        fn.write(script)
+
+    # create fake multirun params based on `num_jobs`
+    fake_param = "+foo="
```
carmocca, line 122: Same suggestion

src/pytorch_lightning/strategies/launchers/subprocess_script.py
```diff
@@ -162,3 +136,45 @@ def _check_can_spawn_children(self) -> None:
...
+        command += ["--experimental-rerun", pickled_config]
+
+    else:
+        command += ["-cp", hydra_output, "-cn", "config.yaml"]
+        command += [
+            f"hydra.output_subdir=.pl_ddp_hydra_{local_rank}",
+            f"hydra.run.dir={hydra_cfg.runtime.output_dir}",
+        ]
+
+    print("*********************", command)
```
carmocca, line 179: Aren't these `*`s a bit spammy? Can you comment here an example of how it looks?

carmocca (approved):

src/pytorch_lightning/strategies/launchers/subprocess_script.py
jgbos: whoops.  I thought I deleted that.

jgbos:
@awaelchli Thanks.  I agree we need more info from the crash as I'm not sure why it is crashing either.  I'll have to figure out how to reproduce this on my machine.

turian:
@jgbos Looks like there are merge conflicts. Could you merge to master and we can see the CI results again?

turian:
Are hydra-core 1.1 and 1.2 both necessary to support?

Perhaps consider pinning [hydra-core to 1.2.0 for now](https://github.com/Lightning-AI/lightning/issues/7110)?

See also:
https://github.com/Lightning-AI/lightning/pull/13309/files

jgbos:
@turian I'll double check but the requirement, `hydra-core>=1.0.5, <1.3.0` should work in general.  I'll look into updating the `rerun` tests as they require version >=1.2

rohitgr7 (approved):

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,168 @@
...
+# fixture to run hydra jobs in a clean temporary directory
+# Hydra creates its own output directories and logs
+@pytest.fixture
+def cleandir(tmp_path):
+    """Run function in a temporary directory."""
+    old_dir = os.getcwd()  # get current working directory (cwd)
+    os.chdir(tmp_path)  # change cwd to the temp-directory
+    yield tmp_path  # yields control to the test to be run
+    os.chdir(old_dir)
+    logging.shutdown()
```
carmocca, line 28: @rohitgr7 I see you changed the tests to use tmpdir. Then this should be removed

tests/tests_pytorch/strategies/test_ddp_hydra_support.py
```diff
@@ -0,0 +1,168 @@
+import logging
```
awaelchli, line 1: If we could move the file to `.../strategies/launchers/test_subprocess_script.py` that would be helpful, thanks!

awaelchli (approved):
We will follow up on this. Created #14799

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
```diff
@@ -0,0 +1,164 @@
...
+
+from pytorch_lightning import Trainer
+
+from tests.tests_pytorch.helpers import BoringModel
+
+class BoringModelGPU(BoringModel):
+    def on_train_start(self) -> None:
+        # make sure that the model is on GPU when training
+        assert self.device == torch.device(f"cuda:{self.trainer.strategy.local_rank}")
+        self.start_cuda_memory = torch.cuda.memory_allocated()
```
carmocca: Why do this?

```suggestion
```

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
```diff
@@ -0,0 +1,164 @@
...
+
+yaml_file = """
+hydra:
+  callbacks:
+    save_job_info:
+      _target_: hydra.experimental.callbacks.PickleJobInfoCallback
+"""
+
+
+@RunIf(min_cuda_gpus=2, skip_windows=True, standalone=True)
```
carmocca, line 131: Wouldn't the test also be valid with 1 device?

Also, does it need to be standalone?

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
```diff
@@ -0,0 +1,164 @@
...
+    # Make sure the parameter was set and used
+    cfg = OmegaConf.load(logs[0])
+    assert cfg.devices == devices
+
+    # Make sure PL spawned a job that is logged by Hydra
+    logs = list(Path.cwd().glob("**/*.log"))
+    assert len(logs) == 1
+
+
+@RunIf(min_cuda_gpus=2, skip_windows=True, standalone=True)
```
carmocca, line 94: Wouldn't the test also be valid with 1 device?

Also, does it need to be standalone?

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
```diff
@@ -0,0 +1,164 @@
...
+        torch.distributed.destroy_process_group()
+
+    os.environ.pop("LOCAL_RANK", None)
+
+if __name__ == "__main__":
+    task_fn()
+"""
+
+
+@RunIf(min_cuda_gpus=2, skip_windows=True, standalone=True)
```
carmocca, line 65: Wouldn't the test also be valid with 1 device?

Also, does it need to be standalone?

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
```diff
@@ -0,0 +1,162 @@
...
+def task_fn(cfg):
+    trainer = Trainer(accelerator="auto", devices=cfg.devices, strategy=cfg.strategy, fast_dev_run=True)
+    model = BoringModelGPU()
+    trainer.fit(model)
+    trainer.test(model)
+
+    if torch.distributed.is_initialized():
+        torch.distributed.destroy_process_group()
+
+    os.environ.pop("LOCAL_RANK", None)
```
awaelchli, line 58: Just saw this. Why is this needed? If real processes get launched in the tests, they need to run as "standalone". The processes should then end normally and terminate, so the process group and env variable won't survive anywhere. 

```suggestion
```

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
awaelchli: @carmocca When the subprocess script launcher is involved, always needs to be standalone.

awaelchli: It has to be two devices, otherwise the subprocess launcher wouldn't create a new process, which is where the hydra changes in this PR take place.

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
jgbos: I think that was a copy and paste from another test.

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
jgbos: This is needed for Hydra multirun to run correctly (at least with basic sweepers).  Without it the process hangs.  I couldn't come up with a solution that would work in Lightning.

tests/tests_pytorch/strategies/launchers/test_subprocess_script.py
```diff
@@ -0,0 +1,162 @@
...
+            "config.yaml",
+            "+devices=2",
+            '+strategy="ddp"',
+            fake_param,
+            "--multirun",
+        ]
+    )
+
+    pickles = sorted(Path.cwd().glob("**/.hydra/config.pickle"))
+    assert len(pickles) == num_jobs
```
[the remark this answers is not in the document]
jgbos: Ah, the test is missing the `cleandir` argument.

carmocca:
Finally merged! Thank you so much for your time @jgbos

Liangtaiwan:
@jgbos @awaelchli @carmocca @rohitgr7 
It seems this PR cause some issues.

1. PYTHONPATH issue.
When I use ddp with 2 devices (gpu), the second process cannot access my own module.
There is no issue with pl==1.7.7. 
PYTHONPATH=.:$PYTHONPATH #my cmd
`Traceback (most recent call last):
  File "/home/$username/diskey-scale/diskey/main.py", line 11, in <module>
    from diskey.conf.config import MainConfig
ModuleNotFoundError: No module named 'diskey'

2. ddp deadlock
after changing PYTHONPATH to
`PYTHONPATH=$HOME/diskey-scale:$PYTHONPATH `
ModuleNotFoundError is solved.
However, the following error will be triggered.
`
hydra.errors.ConfigCompositionException: 'config.yaml' is validated against ConfigStore schema with the same name.
This behavior is deprecated in Hydra 1.1 and will be removed in Hydra 1.2.
In addition, the automatically matched schema contains a defaults list.
This combination is no longer supported.
See https://hydra.cc/docs/next/upgrades/1.0_to_1.1/automatic_schema_matching for migration instructions.
`

These issues was not happended with hydra-submitit-launcher (with slurm).

awaelchli:
@Liangtaiwan Not sure what is going on there. Could you report the exact issue, steps to reproduce, and environment you run in a [GitHub issue](https://github.com/Lightning-AI/lightning/issues/new/choose)? Perhaps @jgbos will be able to help too?
Thanks!

jgbos:
@Liangtaiwan sorry, I would need to know more about the error. Definitely open an issue and I will try to help.  Would be good to know Hydra version and how you executed the script.  Also, this error should not occur if launching on slurm as each gpu task is launched via the submission script (i.e., submitit launcher) instead of this script.

Issue #3544, opened 2020-09-18
Support DataModule on_save_checkpoint and on_load_checkpoint hooks


ananthsub:
## 🚀 Feature
LightningModules have the ability to augment the checkpoint dict with custom information relevant to the module. The motivation for this issue is to provide the same functionality for DataModules, as they too might have custom information they need to read from or write to the checkpoint dict. Importantly, there will only be one checkpoint saved during training. This checkpoint will comprise the default state generated by the trainer, the lightning module custom data (optional), and the data module custom data (optional)
```
Ckpt = Trainer State + Lightning module custom data (optional) + Data module custom data (optional)
```

cc @nateraw 

### Motivation
Custom data we may want to include in the checkpoint:
- Hyperparams
- NLP: Path to a vocabulary file file (or even the vocabulary if it's small enough)
- Image Text Recognition: Storing the alphabet character set of the dataset
- Serialized dataloader state
....

These are all pieces of state that could live in advanced data modules. Importantly, this state isn't static: it can consist of arguments passed into the datamodule, or state generated during training

### Pitch

- Offer the `on_{save/load}_checkpoint` functions for the DataModule
- Give the DataModule (if it exists) a chance to add things to the checkpoint before we save the checkpoint in a callback. This is an additional line [here](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L286-L287)

```
if self.trainer.datamodule is not None:
    self.trainer.datamodule.on_save_checkpoint(checkpoint)
```

- Give the DataModule (if it exists) a chance to load things from the checkpoint whenever the trainer loads a checkpoint. This is an additional line [here]( https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L106-L107)

```
if self.trainer.datamodule is not None:
    self.trainer.datamodule.on_load_checkpoint(checkpoint)
```

**Sharing code:**
We can refactor the existing LightningModule hooks [here](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/core/lightning.py#L1258-L1299) into a separate `CheckpointHook` mixin: 

Then both the LightningModule and DataModule can inherit from this mixin (similar to what was done for the DataHook), meaning both the LightningModule and DataModule will share the function declarations. 

[...]

PR #11637, opened 2022-01-26
Introduce `Stateful` DataModule

jjenniferdai:
## What does this PR do?

part of #11429: introduce `Stateful` DataModule (see **Appendix** code changes outlined in issue)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/core/datamodule.py
+++ b/pytorch_lightning/core/datamodule.py
@@ -259,7 +259,17 @@ class LightningDataModule(CheckpointHooks, DataHooks, HyperparametersMixin):
         return datamodule
 
     def state_dict(self) -> Dict[str, Any]:
+        """Called when saving a checkpoint, implement to generate and save datamodule state.
+
+        Returns:
+            A dictionary containing datamodule state.
+        """
         return {}
 
     def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        """Called when loading a checkpoint, implement to reload datamodule state given datamodule state_dict.
+
+        Args:
+            state_dict: the datamodule state returned by ``state_dict``.
+        """
         pass
--- a/pytorch_lightning/core/datamodule.py
+++ b/pytorch_lightning/core/datamodule.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 """LightningDataModule for loading DataLoaders with ease."""
 from argparse import ArgumentParser, Namespace
-from typing import Any, List, Mapping, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, Union
 
 from torch.utils.data import DataLoader, Dataset, IterableDataset
 
@@ -257,3 +257,9 @@ class LightningDataModule(CheckpointHooks, DataHooks, HyperparametersMixin):
         if test_dataset is not None:
             datamodule.test_dataloader = test_dataloader
         return datamodule
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {}
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        pass
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -333,7 +333,7 @@ class CheckpointConnector:
                 CHECKPOINT_HYPER_PARAMS_KEY:
                 CHECKPOINT_HYPER_PARAMS_TYPE:
                 something_cool_i_want_to_save: anything you define through model.on_save_checkpoint
-                LightningDataModule.__class__.__name__: pl DataModule's state
+                LightningDataModule.__class__.__qualname__: pl DataModule's state
             }
         """
 
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -154,8 +154,8 @@ class CheckpointConnector:
         datamodule = self.trainer.datamodule
         if datamodule is not None:
             datamodule.on_load_checkpoint(self._loaded_checkpoint)
-            if datamodule.__class__.__name__ in self._loaded_checkpoint:
-                datamodule.load_state_dict(self._loaded_checkpoint[datamodule.__class__.__name__])
+            if datamodule.__class__.__qualname__ in self._loaded_checkpoint:
+                datamodule.load_state_dict(self._loaded_checkpoint[datamodule.__class__.__qualname__])
 
     def restore_model(self) -> None:
         """Restores a model's weights from a PyTorch Lightning checkpoint.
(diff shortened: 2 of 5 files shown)


tests/core/test_datamodules.py
```diff
@@ -221,13 +276,14 @@ def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
...
+    assert "my_dm_state" in checkpoint
+    assert checkpoint["my_dm_state"] == {"my": "state_dict"}
 
     for trainer_fn in TrainerFn:
         trainer.state.fn = trainer_fn
-        with mock.patch.object(dm, "on_load_checkpoint") as dm_mock:
-            trainer._restore_modules_and_callbacks(checkpoint_path)
-            dm_mock.assert_called_once()
+        trainer._restore_modules_and_callbacks(checkpoint_path)
+        assert dm.my_state_dict == {"my": "state_dict", "on_load": "update"}
```
jjenniferdai, line 286: example failure - because we introduce `load_state_dict` call after `on_load_checkpoint` - if a user does already have a `Stateful` datamodule (and does not expect lightning to call `load_state_dict`), changes made in `on_load_checkpoint` could be overridden by lightning's `load_state_dict` call after it.

What's the smoothest way to resolve/introduce this change?

pytorch_lightning/core/datamodule.py
```diff
@@ -257,3 +257,9 @@ def test_dataloader():
         if test_dataset is not None:
             datamodule.test_dataloader = test_dataloader
         return datamodule
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {}
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        pass
```
ananthsub, line 265: please add docstrings

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -385,10 +387,15 @@ def dump_checkpoint(self, weights_only: bool = False) -> dict:
             else:
                 checkpoint[pl.LightningModule.CHECKPOINT_HYPER_PARAMS_KEY] = dict(model.hparams)
 
-        # give the model a chance to dump a few things
+        # dump stateful datamodule
+        datamodule = self.trainer.datamodule
+        if datamodule is not None:
+            checkpoint[datamodule.__class__.__name__] = datamodule.state_dict()
```
ananthsub, line 393: all `state_dict()` calls should happen before any `on_save_checkpoint` hooks are called

tests/core/test_datamodules.py
ananthsub: taken a step further, the user may have defined a load_state_dict/state_dict with different arguments, and the lightning trainer calling this function will result in an error because of a mismatched signature. 

I don't think there's any smooth way to get around this because it's a risk of adding to the superclass and relying on inheritance/OOP

however i think the chance of this happening is very, very small.

ananthsub (commented):
as a follow up, please deprecate CheckpointHooks off the PrecisionPlugin interface: https://github.com/PyTorchLightning/pytorch-lightning/blob/86b177ebe5427725b35fde1a8808a7b59b8a277a/pytorch_lightning/core/datamodule.py#L26

It is no longer needed following the Stateful change and should be removed in v1.8

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -154,6 +154,8 @@ def restore_datamodule(self) -> None:
         datamodule = self.trainer.datamodule
         if datamodule is not None:
             datamodule.on_load_checkpoint(self._loaded_checkpoint)
+            if datamodule.__class__.__name__ in self._loaded_checkpoint:
```
ananthsub, line 157: https://github.com/PyTorchLightning/pytorch-lightning/blob/86b177ebe5427725b35fde1a8808a7b59b8a277a/pytorch_lightning/callbacks/base.py#L43 

let's use qualname here too for consistency

jjenniferdai:
Did we want to deprecate `CheckpointHooks` off `DataModule`? The original issue #11429 discussion was more focused only on adding the `Stateful` alignment for all components (Motivation point #1) without editing `CheckpointHooks` (Motivation point #2) yet

ananthsub:
> Did we want to deprecate `CheckpointHooks` off `DataModule`? The original issue #11429 discussion was more focused only on adding the `Stateful` alignment for all components (Motivation point 1) without editing `CheckpointHooks` (Motivation point 2) yet

I should have clarified that we can follow it up in a separate issue/PR. It is not a blocker for this one.

I initially proposed using the CheckpointHooks to DataModule back in https://github.com/PyTorchLightning/pytorch-lightning/issues/3544 . It was to capture dataloader state that had proprietary APIs. I wish at the time I had recognized the `Stateful` protocol lurking in pytorch and the rest of the lightning codebase as the contract for saving/loading state, but I didn't. I don't think there's a use case for the datamodule to be augmenting/changing other components state through the `on_save/load_checkpoint` hooks.

ananthsub (approved):
nice!

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -331,7 +333,7 @@ def dump_checkpoint(self, weights_only: bool = False) -> dict:
                 CHECKPOINT_HYPER_PARAMS_KEY:
                 CHECKPOINT_HYPER_PARAMS_TYPE:
                 something_cool_i_want_to_save: anything you define through model.on_save_checkpoint
-                LightningDataModule.__class__.__name__: pl DataModule's state
+                LightningDataModule.__class__.__qualname__: pl DataModule's state
```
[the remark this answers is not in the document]
justusschock: I strongly prefer the `__qualname__` with a deprecation-phase where we support both. This is way more unique. And unless that class is not defined in the global scope of a file (which would be a problem with spawn etc. anyways, they should be identical

justusschock (approved):

Borda (approved):

pytorch_lightning/core/datamodule.py
```diff
@@ -257,3 +257,19 @@ def test_dataloader():
...
+
+    def state_dict(self) -> Dict[str, Any]:
+        """Called when saving a checkpoint, implement to generate and save datamodule state.
+
+        Returns:
+            A dictionary containing datamodule state.
+        """
+        return {}
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
```
rohitgr7, line 269:
```suggestion
    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
```
quick ques: does `Dist` supports `OrderedDict`? since, for PT Module, it is OrderedDict. Or should we use `Mapping`

rohitgr7 (approved):

pytorch_lightning/trainer/connectors/checkpoint_connector.py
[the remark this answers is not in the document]
jjenniferdai: I don't think the old checkpoint format was actually `LightningDataModule.__class__.__name__` ?`LightningDataModule` could just update/add to the dict with `on_save_checkpoint` but lightning never dictated `LightningDataModule.__class__.__name__` key

pytorch_lightning/core/datamodule.py
jjenniferdai: yep checked mypy has no issues with this:

```
def testingmethod() -> Dict[Any, Any]:
    return OrderedDict()
```

Issue #9396, opened 2021-09-09
checkpoint migration


awaelchli:
## 🚀 Feature

Add upgrade functions to utilities for internal use. Checkpoints get upgraded automatically (when possible) when a user loads the checkpoint using `Trainer(resume_from_checkpoint)` or `Model.load_from_checkpoint`. 

### Motivation

Lightning changes over time with removals and additions, this includes checkpoint[ contents and structure](https://github.com/PyTorchLightning/pytorch-lightning/blob/3070a9ea6edb9648152218a79620f035d9bf719c/pytorch_lightning/core/saving.py#L167). When changes happen, we bake the upgrade logic into the code base at the appropriate place, but the danger is that the information why and when these changes were made [gets lost over time](https://github.com/PyTorchLightning/pytorch-lightning/blob/3070a9ea6edb9648152218a79620f035d9bf719c/pytorch_lightning/utilities/argparse_utils.py#L7). 

### Pitch

For each BC change we create an upgrade function that gets applied at the appropriate place.

```py
def upgrade_xyz_v1_2_0(checkpoint):
    # upgrades the checkpoint from a previous version to 1.2.0
    return checkpoint

def upgrade_abc_v1_3_8(checkpoint):
    # upgrades the checkpoint from previous version to 1.3.8
    return checkpoint


def upgrade(checkpoint)
    checkpoint = upgrade_xyz_v1_2_0(checkpoint)
    checkpoint = upgrade_abc_v1_3_8(checkpoint)
    return

# in Lightning:
ckpt = upgrade(pl_load(path))

```

Benefits with this approach are:
- each upgrade is documented individually
- central location for all upgrades, the order in which they are applied is fully transparent
- can unit test each upgrade individually! 

### Alternatives

keep as is

### Additional context

PRs that started this work: 
- #9166: legacy load context manager to patch Lightning for unpickling 
- #8558: upgrade functions

PRs that added checkpoint back-compatibility logic that can be avoided by this proposal:
- #11638
______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning

[...]

PR #11638, opened 2022-01-27
Introduce `Stateful` PrecisionPlugin

jjenniferdai:
## What does this PR do?

part of #11429: introduce `Stateful` PrecisionPlugin (see Appendix code changes outlined in issue)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/plugins/precision/apex_amp.py
+++ b/pytorch_lightning/plugins/precision/apex_amp.py
@@ -21,6 +21,7 @@ import pytorch_lightning as pl
 from pytorch_lightning.plugins.precision.mixed import MixedPrecisionPlugin
 from pytorch_lightning.utilities import _APEX_AVAILABLE, AMPType
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.rank_zero import rank_zero_deprecation
 from pytorch_lightning.utilities.types import _PARAMETERS
 
 if _APEX_AVAILABLE:
@@ -97,3 +98,15 @@ class ApexMixedPrecisionPlugin(MixedPrecisionPlugin):
 
     def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
         amp.load_state_dict(state_dict)
+
+    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
+        rank_zero_deprecation(
+            "`ApexMixedPrecisionPlugin.on_load_checkpoint` is deprecated in v1.6 and will be removed in v1.8."
+            " Use `ApexMixedPrecisionPlugin.load_state_dict` instead."
+        )
+
+    def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
+        rank_zero_deprecation(
+            "`ApexMixedPrecisionPlugin.on_save_checkpoint` is deprecated in v1.6 and will be removed in v1.8."
+            " Use `ApexMixedPrecisionPlugin.state_dict` instead."
+        )
--- a/pytorch_lightning/plugins/precision/apex_amp.py
+++ b/pytorch_lightning/plugins/precision/apex_amp.py
@@ -101,9 +101,14 @@ class ApexMixedPrecisionPlugin(MixedPrecisionPlugin):
         amp.load_state_dict(state_dict)
 
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        """"``ApexMixedPrecisionPlugin.on_load_checkpoint`` is deprecated in v1.6.
-        Lightning will auto-restore ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.load_state_dict`` instead"""
+        """ "``ApexMixedPrecisionPlugin.on_load_checkpoint`` is deprecated in v1.6.
+
+        Lightning will auto-restore ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.load_state_dict``
+        instead
+        """
 
     def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        """"``ApexMixedPrecisionPlugin.on_save_checkpoint`` is deprecated in v1.6.
-        Lightning will auto-save ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.state_dict`` instead"""
+        """ "``ApexMixedPrecisionPlugin.on_save_checkpoint`` is deprecated in v1.6.
+
+        Lightning will auto-save ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.state_dict`` instead
+        """
--- a/pytorch_lightning/plugins/precision/apex_amp.py
+++ b/pytorch_lightning/plugins/precision/apex_amp.py
@@ -100,14 +100,14 @@ class ApexMixedPrecisionPlugin(MixedPrecisionPlugin):
         amp.load_state_dict(state_dict)
 
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        """ "``ApexMixedPrecisionPlugin.on_load_checkpoint`` is deprecated in v1.6.
+        """``ApexMixedPrecisionPlugin.on_load_checkpoint`` is deprecated in v1.6.
 
         Lightning will auto-restore ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.load_state_dict``
         instead
         """
 
     def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        """ "``ApexMixedPrecisionPlugin.on_save_checkpoint`` is deprecated in v1.6.
+        """``ApexMixedPrecisionPlugin.on_save_checkpoint`` is deprecated in v1.6.
 
         Lightning will auto-save ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.state_dict`` instead
         """
--- a/pytorch_lightning/plugins/precision/apex_amp.py
+++ b/pytorch_lightning/plugins/precision/apex_amp.py
@@ -96,5 +96,4 @@ class ApexMixedPrecisionPlugin(MixedPrecisionPlugin):
         return amp.state_dict()
 
     def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
-        if state_dict:
-            amp.load_state_dict(state_dict)
+        amp.load_state_dict(state_dict)
--- a/pytorch_lightning/plugins/precision/apex_amp.py
+++ b/pytorch_lightning/plugins/precision/apex_amp.py
@@ -21,7 +21,6 @@ import pytorch_lightning as pl
 from pytorch_lightning.plugins.precision.mixed import MixedPrecisionPlugin
 from pytorch_lightning.utilities import _APEX_AVAILABLE, AMPType
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
-from pytorch_lightning.utilities.rank_zero import rank_zero_deprecation
 from pytorch_lightning.utilities.types import _PARAMETERS
 
 if _APEX_AVAILABLE:
--- a/pytorch_lightning/plugins/precision/apex_amp.py
+++ b/pytorch_lightning/plugins/precision/apex_amp.py
@@ -101,13 +101,9 @@ class ApexMixedPrecisionPlugin(MixedPrecisionPlugin):
         amp.load_state_dict(state_dict)
 
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        rank_zero_deprecation(
-            "`ApexMixedPrecisionPlugin.on_load_checkpoint` is deprecated in v1.6 and will be removed in v1.8."
-            " Use `ApexMixedPrecisionPlugin.load_state_dict` instead."
-        )
+        """"``ApexMixedPrecisionPlugin.on_load_checkpoint`` is deprecated in v1.6.
+        Lightning will auto-restore ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.load_state_dict`` instead"""
 
     def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        rank_zero_deprecation(
-            "`ApexMixedPrecisionPlugin.on_save_checkpoint` is deprecated in v1.6 and will be removed in v1.8."
-            " Use `ApexMixedPrecisionPlugin.state_dict` instead."
-        )
+        """"``ApexMixedPrecisionPlugin.on_save_checkpoint`` is deprecated in v1.6.
+        Lightning will auto-save ApexMixedPrecisionPlugin state with ``ApexMixedPrecisionPlugin.state_dict`` instead"""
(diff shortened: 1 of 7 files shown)


pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -192,7 +192,17 @@ def restore_training_state(self) -> None:
...
 
         # restore precision plugin (scaler etc.)
-        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
+        prec_plugin = self.trainer.precision_plugin
+        prec_plugin.on_load_checkpoint(self._loaded_checkpoint)
+        if prec_plugin.__class__.__name__ in self._loaded_checkpoint:
+            prec_plugin.load_state_dict(self._loaded_checkpoint[prec_plugin.__class__.__name__])
+
+        # old checkpoints compatibility
+        # should we raise error and force user to run utilities/upgrade_checkpoint instead?
```
jjenniferdai, line 201: see comment

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -389,6 +401,8 @@ def dump_checkpoint(self, weights_only: bool = False) -> dict:
         model.on_save_checkpoint(checkpoint)
         if self.trainer.datamodule is not None:
             self.trainer.datamodule.on_save_checkpoint(checkpoint)
+        if not weights_only:
+            self.trainer.precision_plugin.on_save_checkpoint(checkpoint)
```
ananthsub, line 405: why does this happen after the other on_save_checkpoint hooks? let's keep the same order as before

pytorch_lightning/trainer/connectors/checkpoint_connector.py
ananthsub: cc @awaelchli - we must keep backwards compatibility with old checkpoints. if we can do this best effort with comments about including new keys & falling back to old keys for precision-related logic, we should do that.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -192,7 +192,17 @@ def restore_training_state(self) -> None:
...
+        prec_plugin.on_load_checkpoint(self._loaded_checkpoint)
+        if prec_plugin.__class__.__name__ in self._loaded_checkpoint:
+            prec_plugin.load_state_dict(self._loaded_checkpoint[prec_plugin.__class__.__name__])
+
+        # old checkpoints compatibility
+        # should we raise error and force user to run utilities/upgrade_checkpoint instead?
+        if "amp_scaling_state" in self._loaded_checkpoint:
+            prec_plugin.load_state_dict(self._loaded_checkpoint["amp_scaling_state"])
+        if "native_amp_scaling_state" in self._loaded_checkpoint:
+            prec_plugin.load_state_dict(self._loaded_checkpoint["native_amp_scaling_state"])
```
ananthsub, line 205: this is now sufficiently complicated enough that it can be its own function within this class

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -372,7 +382,9 @@ def dump_checkpoint(self, weights_only: bool = False) -> dict:
                 lr_schedulers.append(config.scheduler.state_dict())
             checkpoint["lr_schedulers"] = lr_schedulers
 
-            self.trainer.precision_plugin.on_save_checkpoint(checkpoint)
+            # precision plugin
+            prec_plugin = self.trainer.precision_plugin
+            checkpoint[prec_plugin.__class__.__name__] = self.trainer.precision_plugin.state_dict()
```
ananthsub, line 387: note: now if the precision plugin/module ever changes, loading old checkpoints will fail

ananthsub (commented):
as a follow up, please deprecate CheckpointHooks off the PrecisionPlugin interface: https://github.com/PyTorchLightning/pytorch-lightning/blob/86b177ebe5427725b35fde1a8808a7b59b8a277a/pytorch_lightning/plugins/precision/precision_plugin.py#L29

It is no longer needed following the Stateful change and should be removed in v1.8

pytorch_lightning/trainer/connectors/checkpoint_connector.py
ananthsub: https://github.com/PyTorchLightning/pytorch-lightning/blob/86b177ebe5427725b35fde1a8808a7b59b8a277a/pytorch_lightning/callbacks/base.py#L43 

let's use qualname here too for consistency

pytorch_lightning/trainer/connectors/checkpoint_connector.py
carmocca: Making users upgrade checkpoints is tedious. Let's avoid it for as long as possible.

awaelchli: >  should we raise error and force user to run utilities/upgrade_checkpoint instead?

Neither IMO. Lightning should upgrade the checkpoint directly when loading the file, and before sending the dict to the internal components. The development of this feature is currently stale: #9396
Here for example, an upgrade would consist of detecting the keys amp_scaling_state, native_amp_scaling_state, and merging them into the new format saved under the new key `NativeMixedPrecisionPlugin.__class.__name__` or whatever it ends up being :)

pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -92,9 +92,9 @@ def optimizer_step(
...
-    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        if "amp_scaling_state" in checkpoint:
-            amp.load_state_dict(checkpoint["amp_scaling_state"])
+    def state_dict(self) -> Dict[str, Any]:
+        return amp.state_dict()
 
-    def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        checkpoint["amp_scaling_state"] = amp.state_dict()
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        if state_dict:
```
tchaton, line 99: Is this required ? Would it fail if empty ?

pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -106,10 +106,10 @@ def forward_context(self) -> Generator[None, None, None]:
...
 
-    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        if self.scaler is not None and "native_amp_scaling_state" in checkpoint:
-            self.scaler.load_state_dict(checkpoint["native_amp_scaling_state"])
-
-    def on_save_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
+    def state_dict(self) -> Dict[str, Any]:
         if self.scaler is not None:
-            checkpoint["native_amp_scaling_state"] = self.scaler.state_dict()
+            return self.scaler.state_dict()
```
ananthsub, line 111: this returns None if scaler isn't set. should this return an empty dict instead? if so, does the checkpoint connector filter out empty dicts from being saved or loaded?

pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -109,7 +109,9 @@ def forward_context(self) -> Generator[None, None, None]:
     def state_dict(self) -> Dict[str, Any]:
         if self.scaler is not None:
             return self.scaler.state_dict()
+        else:
+            return {}
```
ananthsub:
```suggestion
        if self.scaler is not None:
            return self.scaler.state_dict()
        return {}
```

ananthsub (approved):

pytorch_lightning/plugins/precision/apex_amp.py
jjenniferdai: (moved the empty check to just not save if empty)

pytorch_lightning/trainer/connectors/checkpoint_connector.py
carmocca: I agree. For now, we can keep it as it is and refactor with #9396.

```suggestion
        # old checkpoints compatibility
```

pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -92,9 +92,8 @@ def optimizer_step(
         if not isinstance(model, pl.LightningModule) or not model.automatic_optimization or not skipped_backward:
             optimizer.step(**kwargs)
 
-    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
```
carmocca, line 95: This will be a breaking change if a subclass is calling the super implementation, consider keeping an empty shell during deprecation

pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -242,3 +242,20 @@ def teardown(self) -> None:
 
         It is the right place to release memory and free other resources.
         """
+
+    def state_dict(self) -> Dict[str, Any]:
```
carmocca, line 246: Did you consider making the class use `Stateful` as a mixin?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -192,7 +192,17 @@ def restore_training_state(self) -> None:
             return
 
         # restore precision plugin (scaler etc.)
-        self.trainer.precision_plugin.on_load_checkpoint(self._loaded_checkpoint)
+        prec_plugin = self.trainer.precision_plugin
+        prec_plugin.on_load_checkpoint(self._loaded_checkpoint)
+        if prec_plugin.__class__.__qualname__ in self._loaded_checkpoint:
```
carmocca, line 197: Here you do `on_load_checkpoint` -> `load_state_dict`

but below: `state_dict` -> `on_save_checkpoint`.

Is this on purpose?

pytorch_lightning/plugins/precision/apex_amp.py
justusschock: I think it shouldn't be an empty-shell, but it should raise a warning and call either `state_dict` or `load_state_dict` for convenience. Usually that's what people want to do here :)

pytorch_lightning/trainer/connectors/checkpoint_connector.py
justusschock: I would always have the old api falling back to the new one with a warning and *never* the other way around :)

pytorch_lightning/plugins/precision/precision_plugin.py
Borda: just was thinking the same as I see it already a few times...

Borda (approved):
@carmocca let's merge this as it is and create an issue bout unifying the stateful mixin?

carmocca:
> let's merge this as it is and create an issue bout unifying the stateful mixin?

We can wait for @jjenniferdai to address comments.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
jjenniferdai: > note: now if the precision plugin/module ever changes, loading old checkpoints will fail

if I understand this comment correctly - for compatibility with old checkpoint keys, added an `isinstance` check for the precision plugin before calling `load_state_dict(ckpt["old_precplugin_key"])`

pytorch_lightning/plugins/precision/apex_amp.py
jjenniferdai: good point, I think an empty shell will do bc lightning takes care of saving/loading states in `state_dict/load_state_dict` now.

The only need to keep the actual old `on_save/load_checkpoint` contents is if users manually check `ckpt["old_precplugin_key"]` somewhere in their code... do we need a deprecation process to support this? e.g. write precision plugin state_dict under both `"old_precplugin_key"` and under `qualname`

pytorch_lightning/trainer/connectors/checkpoint_connector.py
jjenniferdai: save: `state_dict` needs to come before `on_save_checkpoint` in order for `on_save_checkpoint` to receive the "full" ckpt dict argument.

load: not as definitive as save, but I was thinking `on_load_checkpoint` gets the loaded ckpt dict to do any necessary logic before passing it to `load_state_dict` calls - thoughts?

pytorch_lightning/plugins/precision/precision_plugin.py
jjenniferdai: originally proposed `Protocol` so that users could simply implement `state_dict/load_state_dict` without having to extend anything. but now that we're adding them to lightning base classes, I think mixin could be better - can change this in another PR

pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -242,3 +242,20 @@ def teardown(self) -> None:
...
+        """Called when saving a checkpoint, implement to generate and save precision plugin state.
```
rohitgr7, line 247:
```suggestion
        """Called when saving a checkpoint, implement to generate precision plugin state_dict.
```
the hook specifically doesn't save it.

rohitgr7 (approved):

ananthsub (approved):

pytorch_lightning/trainer/connectors/checkpoint_connector.py
ananthsub: @jjenniferdai please add that as a comment directly in the code

pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -106,10 +107,23 @@ def forward_context(self) -> Generator[None, None, None]:
...
+        return {}
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        if self.scaler is not None:
+            self.scaler.load_state_dict(state_dict)
+
     def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
-        if self.scaler is not None and "native_amp_scaling_state" in checkpoint:
-            self.scaler.load_state_dict(checkpoint["native_amp_scaling_state"])
+        rank_zero_deprecation(
```
carmocca, line 120: Missing deprecation tests for these 2 here and in the apex plugin

carmocca (approved):

pytorch_lightning/plugins/precision/native_amp.py
awaelchli: so these are  hooks called by Lightning, not the user. Only if the user overrides these hooks AND calls super will they every see this warning, which is less likely. 

I would put the warning into the docstring instead. 
Optionally one can use the is_overridden utility in the place where this hook gets called.

jjenniferdai: - protecting against `super.on_save/load_checkpoint()` calls: presumably this would only happen inside an overriden `on_save/load_checkpoint`. 
- protecting against overriding `on_save/load_checkpoint()`: this itself actually does not need any warnings - until issue #11824

we can either do
1. `if is_overridden: deprecation("if you are calling super.on_save/load_checkpoint, super method is deprecated.)`, then after issue #11824 the warning message just becomes `"checkpointhooks are deprecated"`
2. or we can just do docstring for now until  `"checkpointhooks are deprecated"` entirely

I'm thinking 2

ananthsub: I think docstring for now and deprecation in a follow up is reasonable

awaelchli: yes

awaelchli (approved):
Unfortunately we never had any tests around amp checkpointing (I couldn't find any?). I created this issue #11885 because we need some form of checks to be confident with refactors like these.

Borda (approved):
lgtm

PR #11073, opened 2021-12-15
Align ddp and ddp-spawn strategies in setting up the environment

awaelchli:
## What does this PR do?

Aligns the regular ddp strategy and the ddp spawn strategy in setting up the environment (through the method `setup_environement` which is called in the Trainer._run).

**Before:**

DDP: Setup environment in `setup_environment` method/hook.
DDPSpawn: At spawn time, in worker_setup function (slightly earlier than `setup_environment` method/hook

**Now:**

Both DDP and DDPSpawn set the same variables in `setup_environment` method.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @borda @justusschock @awaelchli @rohitgr7 @kaushikb11 @akihironitta

Issue #10895, opened 2021-12-02
Deprecate or re-purpose the `get_mp_spawn_kwargs`  hook in spawn plugins


awaelchli:
## Proposed refactor

Discuss and refactor `TPU/DDPSpawnPlugin.get_mp_spawn_kwargs`.

### Motivation

As part of our effort to converge to a stable and well organized Strategy+Accelerator API #10416  we are looking for opportunities to simplify a couple of hooks. One is the get_mp_spawn_kwargs in the Spawn-plugins. 

```py
    def get_mp_spawn_kwargs(self, trainer: Optional["pl.Trainer"] = None) -> Dict[str, Any]:
        return {
            "nprocs": len(self.parallel_devices),
            "start_method": self.start_method,
        }
```
### Pitch

**Proposal 1:** 
- Remove/deprecate the method
- Directly use `start_method` as an attribute and/or constructor argument
- `nprocs` can be set already in the constructor
- A custom implementation that needs to pass additional arguments would instead override the new `spawn()` method.

**Proposal 2:**
- Keep it
- Remove the Trainer argument
- Mark it as protected, since the method is only used internally by the Plugin

**Proposal 3:**
- Keep as is

### Additional context

Original discussion started in #10034 by @four4fish  

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @borda @justusschock @awaelchli @akihironitta @tchaton

ananthsub:
[...]

PR #11643, opened 2022-01-27
Add process launchers

### LINKED ISSUE #10985: Interface for Process Creation (DDPSpawn vs. DDP)
## 🚀 Feature

Extrude an interface that allows us to abstract and disentangle the process creation / spawning from plugins.

### Motivation

The simplifications that #10059 introduced brought the DDPSpawnPlugin and DDPPlugin closer together in their function, execution order and API. The fundamental difference between the two however remains in how the processes are created. 

**DDPSpawnPlugin**
The spawning logic in DDPSpawn comprises mainly these three methods:
https://github.com/PyTorchLightning/pytorch-lightning/blob/aeb0b5595fd73d086f4ae0f99d3f1f112f6a4c29/pytorch_lightning/plugins/training_type/ddp_spawn.py#L152
https://github.com/PyTorchLightning/pytorch-lightning/blob/aeb0b5595fd73d086f4ae0f99d3f1f112f6a4c29/pytorch_lightning/plugins/training_type/ddp_spawn.py#L245
https://github.com/PyTorchLightning/pytorch-lightning/blob/aeb0b5595fd73d086f4ae0f99d3f1f112f6a4c29/pytorch_lightning/plugins/training_type/ddp_spawn.py#L271

**DDPPlugin**
As with the spawn plugin, the creation of subprocesses is quite strongly decoupled in a single method in the DDPPlugin:

https://github.com/PyTorchLightning/pytorch-lightning/blob/aeb0b5595fd73d086f4ae0f99d3f1f112f6a4c29/pytorch_lightning/plugins/training_type/ddp.py#L155

The Trainer today (after #10896) has to differentiate between the two and call them differently:

```py
    if isinstance(self.training_type_plugin, DDPSpawnPlugin):
        spawn_output = self.training_type_plugin.spawn(trainer_fn, *args, **kwargs)
        self.training_type_plugin._recover_results_in_main_process(spawn_output, self)
        return spawn_output.trainer_results
    else:
```

Here, the plugin type check leaks into the trainer. This and the fact that the spawning logic is quite isolated inside the respective plugins motivates a refactor that separates them. Two designs have been proposed so far.

### Pitch

**Proposal 1 (@ananthsub):**

```py
class AbstractSpawn(ABC):

   @abstractmethod
    def spawn(self, ...)

   @abstractmethod
    def collect_rank_zero_results(...):
        pass

   @abstractmethod
    def recover_results_in_main_process(...):
        pass


class DDPSpawnPlugin(ParallelPlugin, AbstractSpawn):
    def spawn(self, function, *args, **kwargs):
        ...

    def recover_results_in_main_process(...):
        pass

    def collect_rank_zero_results(...):
        pass

```

In this proposal, the Trainer call reduces to:

```py
if isinstance(self.training_type_plugin, AbstractSpawn):
    ...
else:
    ...
```

**Proposal 2 (@awaelchli):**
```py

    class Executor(ABC):
        def create_processes(...):
            ...
    
    class ScriptExecutor(Executor):
        # calls script in subprocesses like in current DDPPlugin

    class SpawnExecutor(Executor):
        # spawns processes from Trainer function like in DDPSpawnPlugin

        # draft implementation
        def create_processes(fn):
            # trainer reference up for debate
            output = self._spawn(self._wrap(fn))
            return self.recover_results_in_main_process(trainer, output)
        
        def _wrap(fn):
            fn()  # run it
            return self.collect_rank_zero_results()
```
The plugins would then own an instance of this executor. The DDPPlugin and DDPSpawnPlugin would collapse to a single class, for the sake of demonstration call it DDPNew, and it owns either a ScriptExecutor or a SpawnExecutor:

```py
class DDPNew(ParallelPlugin):
    def __init__(..., executor: Executor)
        self.checkpoint_io = ...
        self.executor = executor
```

### Additional context

At this point a very open discussion. The proposal may be updated depending on the feedback and discussions.

https://github.com/PyTorchLightning/pytorch-lightning/pull/10896#discussion_r761357218
Thanks @ananthsub for kicking off the discussion.

cc @borda @tchaton @justusschock @awaelchli @kaushikb11 @akihironitta

rohitgr7:
## What does this PR do?

Fixes #10985 

## Before submitting

## PR review

## Did you have fun?

cc @borda @kaushikb11 @rohitgr7 @justusschock @awaelchli @akihironitta


awaelchli (commented):
partial review, great work so far

pytorch_lightning/strategies/executors/ddp.py
```diff
@@ -0,0 +1,117 @@
...
+    def execute(self, trainer, function, *args, **kwargs):
+        self._call_children_scripts()
+        return function(*args, **kwargs)
+
+    def _call_children_scripts(self):
+        # bookkeeping of spawned processes
+        self._check_can_spawn_children()
+
+        # DDP Environment variables
+        os.environ["MASTER_ADDR"] = self.strategy.cluster_environment.main_address
```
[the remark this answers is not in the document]
rohitgr7: yes, but the issue here is that for DDP spawn, we need strategy since it's being used at a lot of places in the spawn function. Since we pass in the trainer to recover results, we can rely on the trainer argument to extract strategy, cluster environment, almost everything and remove any argument inside executor init?

pytorch_lightning/strategies/launchers/ddp.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, cluster_environment: ClusterEnvironment, num_processes: int, num_nodes: int) -> None:
+        super().__init__()
+        self.cluster_environment = cluster_environment
+        self.num_processes = num_processes
+        self.num_nodes = num_nodes
+        self.interactive_ddp_procs: List[Popen] = []
+
+    def launch(self, function: Callable, *args: Any, **kwargs: Any) -> Any:
+        kwargs.pop("trainer", None)
+        kwargs.pop("strategy")
```
ananthsub, line 43: if strategy is required to be present, why isn't it an argument to `launch` ?

pytorch_lightning/strategies/launchers/ddp_spawn.py
```diff
@@ -0,0 +1,164 @@
...
+from pytorch_lightning.strategies.launchers.base import Launcher
+from pytorch_lightning.strategies.strategy import Strategy
+from pytorch_lightning.trainer.states import TrainerFn, TrainerState
+from pytorch_lightning.utilities.apply_func import apply_to_collection, move_data_to_device
+from pytorch_lightning.utilities.distributed import rank_zero_debug
+from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.types import _PATH
+
+
+class DDPSpawnLauncher(Launcher):
```
ananthsub, line 33: naming wise, this has nothing to do with DDP/DistributedDataParallel. It's a general torch multi-processing spawn launcher

pytorch_lightning/strategies/launchers/single_process.py
```diff
@@ -0,0 +1,23 @@
...
+from typing import Any, Callable
+
+from pytorch_lightning.strategies.launchers.base import Launcher
+
+
+class SingleProcessLauncher(Launcher):
+    def launch(self, function: Callable, *args: Any, **kwargs: Any) -> Any:
+        kwargs.pop("trainer", None)
+        kwargs.pop("strategy")
+        return function(*args, **kwargs)
```
ananthsub, line 23: why do we need this?

pytorch_lightning/strategies/launchers/tpu_spawn.py
```diff
@@ -0,0 +1,106 @@
...
+from pytorch_lightning.utilities.distributed import rank_zero_debug
+from pytorch_lightning.utilities.model_helpers import is_overridden
+
+if _TPU_AVAILABLE:
+    import torch_xla.distributed.xla_multiprocessing as xmp
+else:
+    xm, xmp, MpDeviceLoader, rendezvous = [None] * 4
+
+
+class TPUSpawnLauncher(DDPSpawnLauncher):
```
ananthsub, line 36: let's avoid inheritance wherever possible

pytorch_lightning/strategies/parallel.py
```diff
@@ -125,6 +126,10 @@ def block_backward_sync(self):
         else:
             yield None
 
+    def launch(self, function: Callable, *args: Any, **kwargs: Any) -> Any:
+        launcher = SingleProcessLauncher()
+        return launcher.launch(function, *args, **kwargs)
```
ananthsub, line 131: i don't think we should have this. this is unnecessary redirection for something that's supposed to execute inline.

pytorch_lightning/strategies/single_device.py
```diff
@@ -40,6 +41,10 @@ def __init__(
         self.local_rank = 0
         self.world_size = 1
 
+    def launch(self, function: Callable, *args: Any, **kwargs: Any) -> Any:
+        launcher = SingleProcessLauncher()
+        return launcher.launch(function, *args, **kwargs)
```
ananthsub, line 46: same here

pytorch_lightning/strategies/strategy.py
```diff
@@ -139,6 +139,10 @@ def setup_precision_plugin(self) -> None:
         self.optimizers = optimizers
         self.lr_schedulers = schedulers
 
+    @abstractmethod
+    def launch(self, function: Any, *args: Any, **kwargs: Any) -> Any:
+        """Launch the proceses using a Launcher."""
+
```
ananthsub, line 145: i don't think this makes sense for all strategies

pytorch_lightning/trainer/trainer.py
```diff
@@ -668,12 +668,9 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
...
         try:
-            if isinstance(self.strategy, DDPSpawnStrategy):
-                spawn_output: _SpawnOutput = self.strategy.spawn(trainer_fn, *args, **kwargs)
-                self.strategy._recover_results_in_main_process(spawn_output, self)
-                return spawn_output.trainer_results
-            else:
-                return trainer_fn(*args, **kwargs)
+            kwargs["trainer"] = self
+            kwargs["strategy"] = self.strategy
+            return self.strategy.launch(trainer_fn, *args, **kwargs)
```
ananthsub, line 673: this should be opt-in

why couldn't it be designed as something like
```py
if self.strategy.launcher:
    return self.strategy.launch(trainer_fn, *args, **kwargs)
else:
    trainer_fn(*args, **kwargs)
```

then we have no need for the singleprocess launcher. single-device strategies and parallel plugins don't need it at all. 

wouldn't that make it easier to merge the DDP and DDP spawn strategy classes over time?

pytorch_lightning/strategies/launchers/ddp_spawn.py
rohitgr7: yeah! will update this... did this initially to check how many launchers we actually needed so kept that prefix.

pytorch_lightning/strategies/launchers/ddp.py
rohitgr7: will make this an argument.

pytorch_lightning/trainer/trainer.py
ananthsub: another concern is on the cyclic dependency between a launcher and a strategy. if the launcher is part of the strategy, but the launcher also requires the strategy, doesn't this point to something we can clear up?

rohitgr7: since the launcher is not a class member variable, we can't really tell still this point whether it will execute on a single process or on a multi-device for DDP. Do you think it should be an argument? Then SingleProcessLauncher can be cleaned up.

pytorch_lightning/strategies/launchers/tpu_spawn.py
rohitgr7: but there are some methods from spawn launcher required here. Else it would need to replicate those here.

pytorch_lightning/lite/lite.py
```diff
@@ -399,16 +399,19 @@ def seed_everything(seed: Optional[int] = None, workers: Optional[bool] = None)
         return seed_everything(seed=seed, workers=workers)
 
     def _run_impl(self, run_method: Callable, *args: Any, **kwargs: Any) -> Any:
-        self._strategy.setup_environment()
-
         # apply sharded context to prevent OOM
         run_method = partial(self._run_with_sharded_context, run_method)
+        run_method = partial(self._wrapped_run_method, run_method)
```
awaelchli, line 404: i don't think we need this, it can be merged with run_with_sharded_context if necessary

pytorch_lightning/strategies/launchers/multi_process.py
```diff
@@ -0,0 +1,112 @@
...
+from pytorch_lightning.plugins.environments.cluster_environment import ClusterEnvironment
+from pytorch_lightning.strategies.launchers.base import Launcher
+from pytorch_lightning.utilities import _HYDRA_AVAILABLE
+
+if _HYDRA_AVAILABLE:
+    from hydra.core.hydra_config import HydraConfig
+    from hydra.utils import get_original_cwd, to_absolute_path
+
+
+class MultiProcessLauncher(Launcher):
```
awaelchli, line 33: The name is too close to `multiprocessing.Process`, but we actually use the subprocess module. I would suggest to choose a different name. 

- SubprocessLauncher
- SubprocessScriptLauncher (to emphasize this calls the script directly)

pytorch_lightning/strategies/ddp.py
```diff
@@ -144,18 +135,19 @@ def distributed_sampler_kwargs(self):
     def _is_single_process_single_device(self) -> bool:
         return True
 
-    def setup_environment(self) -> None:
-        # start the other scripts
-        if not self.cluster_environment.creates_processes_externally:
-            self._call_children_scripts()
+    def configure_multi_process_launcher(self):
```
awaelchli, line 138: I'd like to avoid this, need to brainstorm.

pytorch_lightning/strategies/launchers/multi_process.py
justusschock: I vote for the second one although it is quite long :D

pytorch_lightning/strategies/ddp.py
rohitgr7: Yes, its annoying, I know. But we can't make the decision for it during init. One workaround can be to construct it with strategy and make the decision whether to create children scripts inside Launcher.launch.

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,165 @@
...
+import numpy as np
+import torch
+import torch.multiprocessing as mp
+
+import pytorch_lightning as pl
+from pytorch_lightning.strategies.launchers.base import Launcher
+from pytorch_lightning.strategies.strategy import Strategy
+from pytorch_lightning.trainer.states import TrainerFn, TrainerState
+from pytorch_lightning.utilities.apply_func import apply_to_collection, move_data_to_device
+from pytorch_lightning.utilities.distributed import rank_zero_debug
```
rohitgr7, line 28:
```suggestion
from pytorch_lightning.utilities.rank_zero import rank_zero_debug
```

pytorch_lightning/strategies/launchers/xla_spawn.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import Any, Callable, Optional
+
+import torch.multiprocessing as mp
+
+import pytorch_lightning as pl
+from pytorch_lightning.strategies.launchers.spawn import _FakeQueue, _SpawnOutput, SpawnLauncher
+from pytorch_lightning.trainer.states import TrainerFn
+from pytorch_lightning.utilities import _TPU_AVAILABLE
+from pytorch_lightning.utilities.apply_func import move_data_to_device
+from pytorch_lightning.utilities.distributed import rank_zero_debug
```
rohitgr7, line 26:
```suggestion
from pytorch_lightning.utilities.rank_zero import rank_zero_debug
```

pytorch_lightning/strategies/ddp.py
tchaton: Do you need to make a decision at init? It could be done when the launch is triggered no?

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,165 @@
...
+from pytorch_lightning.strategies.launchers.base import Launcher
+from pytorch_lightning.strategies.strategy import Strategy
+from pytorch_lightning.trainer.states import TrainerFn, TrainerState
+from pytorch_lightning.utilities.apply_func import apply_to_collection, move_data_to_device
+from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.rank_zero import rank_zero_debug
+from pytorch_lightning.utilities.types import _PATH
+
+
+class SpawnLauncher(Launcher):
```
tchaton, line 33: docstring.

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,165 @@
...
+        os.environ["MASTER_PORT"] = str(self._strategy.cluster_environment.main_port)
+        context = mp.get_context("spawn")
+        return_queue = context.SimpleQueue()
+        mp.spawn(
+            self._wrapped_function,
+            args=(trainer, function, args, kwargs, return_queue),
+            nprocs=self._strategy.num_processes,
+        )
+        spawn_output = return_queue.get()
+        if trainer is None:
```
tchaton, line 48: Neat !

pytorch_lightning/trainer/trainer.py
```diff
@@ -669,10 +669,9 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
             **kwargs: keyword arguments to be passed to `trainer_fn`
         """
         try:
-            if isinstance(self.strategy, DDPSpawnStrategy):
-                spawn_output: _SpawnOutput = self.strategy.spawn(trainer_fn, *args, **kwargs)
-                self.strategy._recover_results_in_main_process(spawn_output, self)
-                return spawn_output.trainer_results
+            if self.strategy.launcher is not None:
+                kwargs["trainer"] = self
+                return self.strategy.launcher.launch(trainer_fn, *args, **kwargs)
```
tchaton, line 674:
```suggestion
                return self.strategy.launcher.launch(trainer_fn, *args, trainer=self, **kwargs)
```

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,165 @@
...
+            return spawn_output
+
+        self._recover_results_in_main_process(spawn_output, trainer)
+        return spawn_output.trainer_results
+
+    def _wrapped_function(
```
tchaton, line 54:
```suggestion
    def _wrapping_function(
```

pytorch_lightning/strategies/ddp.py
rohitgr7: yes, but for that, we have to put the logic inside the launcher and construct `strategy` inside that instead of num_processes, nodes and cluster environment. Also if the decision is made inside and for the cases where processes are created externally, subprocess scripts won't be created and calling it `SubprocessScriptLauncher` might not be completely correct.

pytorch_lightning/strategies/launchers/multi_process.py
rohitgr7: done!

setup.cfg
```diff
@@ -25,13 +25,13 @@ addopts =
...
-    # TODO: remove in 1.7
-    ignore::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning:pytorch_lightning.core.decorators
-    ignore::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning:pytorch_lightning.core.memory
+   # error out on our deprecation warnings - ensures the code and tests are kept up-to-date
+   error::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning
+   error::FutureWarning
+   # warnings from deprecated modules on import
+   # TODO: remove in 1.7
+   ignore::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning:pytorch_lightning.core.decorators
+   ignore::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning:pytorch_lightning.core.memory
```
rohitgr7, line 34:
```suggestion
    # error out on our deprecation warnings - ensures the code and tests are kept up-to-date
    error::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning
    error::FutureWarning
    # warnings from deprecated modules on import
    # TODO: remove in 1.7
    ignore::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning:pytorch_lightning.core.decorators
    ignore::pytorch_lightning.utilities.rank_zero.LightningDeprecationWarning:pytorch_lightning.core.memory
```

pytorch_lightning/lite/lite.py
```diff
@@ -399,17 +399,16 @@ def seed_everything(seed: Optional[int] = None, workers: Optional[bool] = None)
         return seed_everything(seed=seed, workers=workers)
 
     def _run_impl(self, run_method: Callable, *args: Any, **kwargs: Any) -> Any:
-        self._strategy.setup_environment()
-
```
carmocca, line 403: Why is this gone?

pytorch_lightning/strategies/launchers/base.py
```diff
@@ -0,0 +1,27 @@
...
+class _Launcher(ABC):
+    r"""
+    Abstract base class used to build new Launchers.
+
+    Subclass this class and override any of the relevant methods
+    """
+
+    @abstractmethod
+    def launch(self, function: Callable, *args: Any, **kwargs: Any) -> Any:
+        """Launches the proceses."""
```
carmocca, line 27:
```suggestion
        """Launches the processes."""
```

pytorch_lightning/strategies/strategy.py
```diff
@@ -100,6 +106,9 @@ def connect(self, model: Module) -> None:
         """Called by the accelerator to connect the accelerator and the model with this plugin."""
         self.model = model
 
+    def _configure_launcher(self):
+        """Attach the launcher based on Strategy."""
+
```
carmocca, line 111: Why are we adding this to the base Strategy? Shouldn't this be added to ParallelStrategy instead?

rohitgr7: inside the accelerator_connector, we just make a call for it, no matter what the strategy is, if you tink, I can add a condition there to make a call for it only if it is a ParallelStrategy.

pytorch_lightning/lite/lite.py
rohitgr7: moved to L411. Tests would have failed if I'd have removed it.

tchaton (approved):
LGTM !

pytorch_lightning/strategies/launchers/subprocess_script.py
```diff
@@ -0,0 +1,118 @@
...
+    r"""
+    Creates and launches subprocess scripts on each device.
+    """
+
+    def __init__(self, cluster_environment: ClusterEnvironment, num_processes: int, num_nodes: int) -> None:
+        super().__init__()
+        self.cluster_environment = cluster_environment
+        self.num_processes = num_processes
+        self.num_nodes = num_nodes
+        self.interactive_ddp_procs: List[Popen] = []
```
ananthsub, line 43: nit: no longer need to keep ddp in the name

rohitgr7: Good catch! Btw do you know why we need to store process in it? I don't see it being used anywhere.

pytorch_lightning/strategies/ddp.py
```diff
@@ -108,7 +99,7 @@ def __init__(
         self._model_averaging_period = model_averaging_period
         self._pids: Optional[List[int]] = None
         self._sync_dir: Optional[str] = None
-        self._rank_0_has_called_call_children_scripts: bool = False
+        self._rank_0_will_call_children_scripts: bool = False
```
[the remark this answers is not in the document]
rohitgr7: Yes, in case if user sets the local rank to the environment, we don't create children scripts and the decision is made inside the launcher. And if it doesn't create the children scripts then we don't need to to set this flag to True.

Another case is when someone calls fit and then test, using the same trainer instance, the second time it launches, we need to keep this as false.

pytorch_lightning/strategies/launchers/subprocess_script.py
ananthsub: I'm not sure of its history (n00b git question: how do you track the git blame across renames?)

was the intention for the strategy to terminate the subprocess at the end?

pytorch_lightning/trainer/trainer.py
```diff
@@ -1201,9 +1199,6 @@ def _run(
         self.state.status = TrainerStatus.FINISHED
         self.state.stage = None
 
-        if isinstance(self.strategy, DDPSpawnStrategy):
-            results = self.strategy._collect_rank_zero_results(self, results)
-
```
awaelchli, line 1206: Beautiful. The plan is working out!

awaelchli:
@rohitgr7 As promised I added extensive docs to the new launcher classes. Also fixed the 5 remaining mypy typing errors.

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,191 @@
...
+    def _wrapping_function(
+        self,
+        process_idx: int,
+        trainer: Optional["pl.Trainer"],
+        function: Callable,
+        args: Any,
+        kwargs: Any,
+        return_queue: SimpleQueue,
+    ) -> None:
+        self._strategy._worker_setup(process_idx)
```
awaelchli, line 91: here we are currently calling into the `strategy._worker_setup`.  We will be able to clean this up in a follow up here #11073 and `setup_environment` in the trainer will be called as early as possible.
cc @ananthsub

pytorch_lightning/strategies/strategy.py
carmocca: IMO yes. Thoughts about adding this just to `ParallelStrategy` @awaelchli?

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,191 @@
...
+        Args:
+            trainer: reference to the Trainer.
+            queue: the instance of the queue to append the data.
+        """
+        callback_metrics: dict = apply_to_collection(
+            trainer.callback_metrics, torch.Tensor, lambda x: x.cpu().numpy()
+        )  # send as numpy to avoid issues with memory sharing
+        queue.put(callback_metrics)
+
+    def get_from_queue(self, trainer: "pl.Trainer", queue: "_FakeQueue") -> None:
```
awaelchli, line 160: in lightning_module.py we have deprecation messages that still reference the DDPSpawnStrategy.add_to_queue and get_from_queue. We need to update these mentions

tests/strategies/test_ddp_spawn_strategy.py
```diff
@@ -82,16 +83,24 @@ def test_ddp_spawn_extra_parameters(tmpdir):
...
     def get_from_queue(self, trainer: Trainer, queue) -> None:
-        self.new_test_val = queue.get()
+        trainer.strategy.new_test_val = queue.get()
         return super().get_from_queue(trainer, queue)
 
 
+class TestDDPSpawnStrategy(DDPSpawnStrategy):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
```
awaelchli, line 99:
```suggestion
```

pytorch_lightning/strategies/strategy.py
awaelchli: It makes more sense to me to add it to the base class. Every strategy should have the possibility to choose a launcher, there is conceptually no restriction necessary for it to be exclusively on the ParallelStrategy.

But we we should also keep in mind:

The only reason why we have this method (correct me if I'm wrong @rohitgr7) is because we need to create the launcher lazily. Hence, this `_configure_launcher` method will become entirely obsolete with #7650. Until then, I suggest to keep it protected on the strategy base class.

rohitgr7: yes! exactly @awaelchli 
let's keep this in base for now

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,191 @@
...
+
+    The main process in which this launcher is invoked creates N so-called worker processes (using
+    :func:`torch.multiprocessing.spawn`) that run the given function.
+    Worker processes have a rank that ranges from 0 to N - 1.
+
+    Note:
+        - This launcher requires all objects to be pickleable.
+        - It is important that the entry point to the program/script is guarded by ``if __name__ == "__main__"``.
+
+    Arguments:
```
carmocca, line 44: nit: (not sure if `Arguments:` works too)

```suggestion
    Args:
```

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,191 @@
...
+        Arguments:
+            function: The entry point for all spawned processes.
+            *args: Optional positional arguments to be passed to the given function.
+            **kwargs: Optional keyword arguments to be passed to the given function.
+                If a keyword argument named `trainer` is present and is an instance of
+                :class:`~pytorch_lightning.trainer.trainer.Trainer`, a selected set of attributes from the trainer get
+                restored in the main process after processes join. The `trainer` keyword argument will NOT be passed
+                into the function.
+        """
+        trainer = kwargs.pop("trainer", None)
```
carmocca, line 66: Why not have the signature as

```py
def launch(self, function: Callable, *args: Any, trainer: Optional[Trainer] = None, **kwargs: Any) -> Any:
```

So you can remove
```suggestion
```

And the kwargs docstring is simplified?

Same comment would apply to the other launcher implementations

pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -0,0 +1,191 @@
...
+        os.environ["MASTER_PORT"] = str(self._strategy.cluster_environment.main_port)
```
carmocca, line 67: Why does this set MASTER_PORT if the `cluster_environment.main_port` implementations already get the value from the environment?

pytorch_lightning/strategies/strategy.py
carmocca: Fine with me at the moment.

> Every strategy should have the possibility to choose a launcher, there is conceptually no restriction necessary for it to be exclusively on the ParallelStrategy.

While there's no technical restriction, I still think it adds confusion that non-parallel strategies could configure process launchers.

If the `ParallelStrategy` is not the right abstraction for you then perhaps we should think of a middle step like

`Strategy -> LaunchingStrategy -> ParallelStrategy`

(not for this PR, you can resolve).

carmocca (approved):
Awesome job! @rohitgr7 @awaelchli 

Looking forward to those follow-ups

pytorch_lightning/strategies/launchers/spawn.py
awaelchli: both work, one is an alias for the other

pytorch_lightning/strategies/launchers/spawn.py
awaelchli: Created an issue  #11965, but just when I submitted it I remember now the real reason for it. No, it cannot be removed so easily. See answer on the issue.

ananthsub (approved):

awaelchli (approved):
I will follow up with clean up for #10895

pytorch_lightning/strategies/launchers/spawn.py
carmocca: I'm confused. My proposal should work with Lite and Lightning.

Both are just passing `*args, **kwargs`, the difference is that the "pop" happens with the signature instead of manually inside the method

pytorch_lightning/strategies/launchers/spawn.py
awaelchli: Just realized, it is there to fix the network port before spawning the processes. Otherwise, we would end up using a different port in each spawned process because of this:
https://github.com/PyTorchLightning/pytorch-lightning/blob/4dba492fb55e2862bb20f304e6bef43f1f151fc7/pytorch_lightning/plugins/environments/lightning_environment.py#L59

`find_free_network_port()` chooses a random free port. All processes would end up trying to connect on a different port. That's why we set the env variable BEFORE spawning processes so all of them end up using the same port.

pytorch_lightning/strategies/launchers/spawn.py
awaelchli: Ah yes, you are right, if it's an optional kwarg then yes

rohitgr7:
thanks, @awaelchli for the discussions for this PR :)

PR #11657, opened 2022-01-28
Fix progress bar tracking with multiple dataloaders

### LINKED ISSUE #11615: tqdm progress not working correctly with multiple dataloaders
## 🐛 Bug

TQDM progress is progressing from the beginning for each dataloader on master but on patch release, it's starting from where it left for the previous dataloader. Thus causing the issue in the attached image for multiple dataloaders.

### To Reproduce

<p>

```py
import torch

from pytorch_lightning import LightningDataModule, LightningModule, Trainer
from pytorch_lightning.callbacks import RichProgressBar


class RandomDataset(Dataset):
    def __init__(self, size: int, length: int):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

    def validation_step(self, batch, batch_idx, dataloader_idx=None):
        loss = self(batch).sum

    def val_dataloader(self):
        return [DataLoader(RandomDataset(32, 64)), DataLoader(RandomDataset(32, 64))]
```
</p>
</details>

```py
trainer = Trainer()
trainer.validate(BoringModel())
```

Output:
Screenshot 2022-01-25 at 6 56 09 PM

Same issue with validation under `trainer.fit()`, `trainer.test()` and `trainer.predict()`

Also if someone does this:
```py
def val_dataloader(self):
    return [DataLoader(RandomIterableDataset(32, 2000)), DataLoader(RandomDataset(32, 2000))]
```
then it shows total batches as inf, even though the second dataloader has a length, but it doesn't show up since we probably might be assigning it to sum(all_dataloader_batches). what I think might look good is that it should set the total individually for each dataloader, and for every new iteration starts from the beginning.

### Environment

- OS (e.g., Linux): mac
- CUDA/cuDNN version:

### Additional context

working fine on patch releases. RichProgressBar is also working fine.

cc @borda @awaelchli @rohitgr7 @carmocca

rohitgr7:
## What does this PR do?

Fixes #11615
More specifically, the PR now enables that a new bar is being used for each eval dataloaders. It won't destroy the old one after the first dataloader is completer, rather it resets it.

This is how it will look like: https://github.com/PyTorchLightning/pytorch-lightning/issues/11615#issuecomment-1021530958

## Before submitting

## PR review

## Did you have fun?

cc @awaelchli @rohitgr7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -457,6 +457,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Disbled sampler replacement when using `IterableDataset` ([#11507](https://github.com/PyTorchLightning/pytorch-lightning/pull/11507))
 
 
+- Fixed `TQDMProgressBar` counter when using multple validation dataloaders ([#11657](https://github.com/PyTorchLightning/pytorch-lightning/pull/11657))
+
+
 ## [1.5.8] - 2022-01-05
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -222,6 +222,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `MisconfigurationException` to `ModuleNotFoundError` when `rich` isn't available ([#11360](https://github.com/PyTorchLightning/pytorch-lightning/pull/11360))
 
 
+- Update `TQDMProgressBar` to run a separate progress bar for each eval dataloader ([#11657](https://github.com/PyTorchLightning/pytorch-lightning/pull/11657))
+
+
 ### Deprecated
 
 - Deprecated `ClusterEnvironment.master_{address,port}` in favor of `ClusterEnvironment.main_{address,port}` ([#10103](https://github.com/PyTorchLightning/pytorch-lightning/pull/10103))
@@ -455,9 +458,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
 - Disbled sampler replacement when using `IterableDataset` ([#11507](https://github.com/PyTorchLightning/pytorch-lightning/pull/11507))
-
-
-- Fixed `TQDMProgressBar` counter when using multple validation dataloaders ([#11657](https://github.com/PyTorchLightning/pytorch-lightning/pull/11657))
 
 
 ## [1.5.8] - 2022-01-05
--- a/pytorch_lightning/callbacks/progress/base.py
+++ b/pytorch_lightning/callbacks/progress/base.py
@@ -48,6 +48,9 @@ class ProgressBarBase(Callback):
 
     def __init__(self) -> None:
         self._trainer: Optional["pl.Trainer"] = None
+        self._val_progress: Optional[int] = None
+        self._test_progress: Optional[int] = None
+        self._predict_progress: Optional[int] = None
 
     @property
     def trainer(self) -> "pl.Trainer":
@@ -63,6 +66,34 @@ class ProgressBarBase(Callback):
         """
         return self.trainer.fit_loop.epoch_loop.batch_progress.current.processed
 
+    def on_validation_batch_start(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
+    ):
+        if self._val_progress is None or batch_idx == 0:
+            max_batches = trainer.num_sanity_val_batches if trainer.sanity_checking else trainer.num_val_batches
+            self._val_progress = sum(max_batches[:dataloader_idx])
+
+    def on_test_batch_start(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
+    ):
+        if self._test_progress is None or batch_idx == 0:
+            self._test_progress = sum(trainer.num_test_batches[:dataloader_idx])
+
+    def on_predict_batch_start(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
+    ):
+        if self._predict_progress is None or batch_idx == 0:
+            self._predict_progress = sum(trainer.num_predict_batches[:dataloader_idx])
+
+    def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._val_progress = None
+
+    def on_test_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._test_progress = None
+
+    def on_predict_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._predict_progress = None
+
     @property
     def val_batch_idx(self) -> int:
         """The number of batches processed during validation.
@@ -70,8 +101,13 @@ class ProgressBarBase(Callback):
         Use this to update your progress bar.
         """
         if self.trainer.state.fn == "fit":
-            return self.trainer.fit_loop.epoch_loop.val_loop.epoch_loop.batch_progress.current.processed
-        return self.trainer.validate_loop.epoch_loop.batch_progress.current.processed
+            loop = self.trainer.fit_loop.epoch_loop.val_loop
+        else:
+            loop = self.trainer.validate_loop
+
+        current_batch_idx = loop.epoch_loop.batch_progress.current.processed
+        batch_idx = self._val_progress + current_batch_idx
+        return batch_idx
 
     @property
     def test_batch_idx(self) -> int:
@@ -79,7 +115,10 @@ class ProgressBarBase(Callback):
 
         Use this to update your progress bar.
         """
-        return self.trainer.test_loop.epoch_loop.batch_progress.current.processed
+        loop = self.trainer.test_loop
+        current_batch_idx = loop.epoch_loop.batch_progress.current.processed
+        batch_idx = self._test_progress + current_batch_idx
+        return batch_idx
 
     @property
     def predict_batch_idx(self) -> int:
@@ -87,7 +126,10 @@ class ProgressBarBase(Callback):
 
         Use this to update your progress bar.
         """
-        return self.trainer.predict_loop.epoch_loop.batch_progress.current.processed
+        loop = self.trainer.predict_loop
+        current_batch_idx = loop.epoch_loop.batch_progress.current.processed
+        batch_idx = self._predict_progress + current_batch_idx
+        return batch_idx
 
     @property
     def total_train_batches(self) -> Union[int, float]:
@@ -105,6 +147,9 @@ class ProgressBarBase(Callback):
         Use this to set the total number of iterations in the progress bar. Can return ``inf`` if the validation
         dataloader is of infinite size.
         """
+        if self.trainer.sanity_checking:
+            return sum(self.trainer.num_sanity_val_batches)
+
         total_val_batches = 0
         if self.trainer.enable_validation:
             is_val_epoch = (self.trainer.current_epoch + 1) % self.trainer.check_val_every_n_epoch == 0
--- a/pytorch_lightning/callbacks/progress/base.py
+++ b/pytorch_lightning/callbacks/progress/base.py
@@ -48,15 +48,27 @@ class ProgressBarBase(Callback):
 
     def __init__(self) -> None:
         self._trainer: Optional["pl.Trainer"] = None
-        self._val_progress: Optional[int] = None
-        self._test_progress: Optional[int] = None
-        self._predict_progress: Optional[int] = None
+        self._current_eval_dataloader_idx: Optional[int] = None
 
     @property
     def trainer(self) -> "pl.Trainer":
         if self._trainer is None:
             raise TypeError(f"The `{self.__class__.__name__}._trainer` reference has not been set yet.")
         return self._trainer
+
+    def is_dataloader_changed(self, dataloader_idx: int) -> bool:
+        old_dataloader_idx = self._current_eval_dataloader_idx
+        self._current_eval_dataloader_idx = dataloader_idx
+        return old_dataloader_idx != dataloader_idx
+
+    def on_validation_end(self, *_: Any) -> None:
+        self._current_eval_dataloader_idx = None
+
+    def on_test_end(self, *_: Any) -> None:
+        self._current_eval_dataloader_idx = None
+
+    def on_predict_end(self, *_: Any) -> None:
+        self._current_eval_dataloader_idx = None
 
     @property
     def train_batch_idx(self) -> int:
@@ -65,34 +77,6 @@ class ProgressBarBase(Callback):
         Use this to update your progress bar.
         """
         return self.trainer.fit_loop.epoch_loop.batch_progress.current.processed
-
-    def on_validation_batch_start(
-        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
-    ) -> None:
-        if self._val_progress is None or batch_idx == 0:
-            max_batches = trainer.num_sanity_val_batches if trainer.sanity_checking else trainer.num_val_batches
-            self._val_progress = sum(max_batches[:dataloader_idx])
-
-    def on_test_batch_start(
-        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
-    ) -> None:
-        if self._test_progress is None or batch_idx == 0:
-            self._test_progress = sum(trainer.num_test_batches[:dataloader_idx])
-
-    def on_predict_batch_start(
-        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
-    ) -> None:
-        if self._predict_progress is None or batch_idx == 0:
-            self._predict_progress = sum(trainer.num_predict_batches[:dataloader_idx])
-
-    def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        self._val_progress = None
-
-    def on_test_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        self._test_progress = None
-
-    def on_predict_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        self._predict_progress = None
 
     @property
     def val_batch_idx(self) -> int:
@@ -106,8 +90,7 @@ class ProgressBarBase(Callback):
             loop = self.trainer.validate_loop
 
         current_batch_idx = loop.epoch_loop.batch_progress.current.processed
-        batch_idx = self._val_progress + current_batch_idx
-        return batch_idx
+        return current_batch_idx
 
     @property
     def test_batch_idx(self) -> int:
@@ -115,10 +98,7 @@ class ProgressBarBase(Callback):
 
         Use this to update your progress bar.
         """
-        loop = self.trainer.test_loop
-        current_batch_idx = loop.epoch_loop.batch_progress.current.processed
-        batch_idx = self._test_progress + current_batch_idx
-        return batch_idx
+        return self.trainer.test_loop.epoch_loop.batch_progress.current.processed
 
     @property
     def predict_batch_idx(self) -> int:
@@ -126,10 +106,7 @@ class ProgressBarBase(Callback):
 
         Use this to update your progress bar.
         """
-        loop = self.trainer.predict_loop
-        current_batch_idx = loop.epoch_loop.batch_progress.current.processed
-        batch_idx = self._predict_progress + current_batch_idx
-        return batch_idx
+        return self.trainer.predict_loop.epoch_loop.batch_progress.current.processed
 
     @property
     def total_train_batches(self) -> Union[int, float]:
@@ -148,14 +125,9 @@ class ProgressBarBase(Callback):
         dataloader is of infinite size.
         """
         if self.trainer.sanity_checking:
-            return sum(self.trainer.num_sanity_val_batches)
+            return self.trainer.num_sanity_val_batches[self._current_eval_dataloader_idx]
 
-        total_val_batches = 0
-        if self.trainer.enable_validation:
-            is_val_epoch = (self.trainer.current_epoch + 1) % self.trainer.check_val_every_n_epoch == 0
-            total_val_batches = sum(self.trainer.num_val_batches) if is_val_epoch else 0
-
-        return total_val_batches
+        return self.trainer.num_val_batches[self._current_eval_dataloader_idx]
 
     @property
     def total_test_batches(self) -> Union[int, float]:
@@ -164,7 +136,7 @@ class ProgressBarBase(Callback):
         Use this to set the total number of iterations in the progress bar. Can return ``inf`` if the test dataloader is
         of infinite size.
         """
-        return sum(self.trainer.num_test_batches)
+        return self.trainer.num_test_batches[self._current_eval_dataloader_idx]
 
     @property
     def total_predict_batches(self) -> Union[int, float]:
@@ -173,7 +145,7 @@ class ProgressBarBase(Callback):
         Use this to set the total number of iterations in the progress bar. Can return ``inf`` if the predict dataloader
         is of infinite size.
         """
-        return sum(self.trainer.num_predict_batches)
+        return self.trainer.num_predict_batches[self._current_eval_dataloader_idx]
 
     def disable(self) -> None:
         """You should provide a way to disable the progress bar.
--- a/pytorch_lightning/callbacks/progress/base.py
+++ b/pytorch_lightning/callbacks/progress/base.py
@@ -68,20 +68,20 @@ class ProgressBarBase(Callback):
 
     def on_validation_batch_start(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
-    ):
+    ) -> None:
         if self._val_progress is None or batch_idx == 0:
             max_batches = trainer.num_sanity_val_batches if trainer.sanity_checking else trainer.num_val_batches
             self._val_progress = sum(max_batches[:dataloader_idx])
 
     def on_test_batch_start(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
-    ):
+    ) -> None:
         if self._test_progress is None or batch_idx == 0:
             self._test_progress = sum(trainer.num_test_batches[:dataloader_idx])
 
     def on_predict_batch_start(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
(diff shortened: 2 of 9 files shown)


carmocca:
Does this implement the separate progress bar idea described in the linked issue?

rohitgr7:
> Does this implement the separate progress bar idea described in the linked issue?

no, I have mentioned the reason here why I didn't go in that direction for now. For now, it just fixes the bug. To support the idea for a different progress bar, I need some more time to look at whether it will be possible with the rich progress bar as well. Will keep that open to discuss possible solutions.

carmocca:
Just synced with Rohit, we are going to try to pursue the separate bar idea first and if that does not work out, continue this PR. That feature replaces this fix because this is not a problem in the bug-fix branch.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -48,13 +48,28 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
         self._trainer: Optional["pl.Trainer"] = None
+        self._current_eval_dataloader_idx: Optional[int] = None
 
     @property
     def trainer(self) -> "pl.Trainer":
         if self._trainer is None:
             raise TypeError(f"The `{self.__class__.__name__}._trainer` reference has not been set yet.")
         return self._trainer
 
+    def is_dataloader_changed(self, dataloader_idx: int) -> bool:
```
rohitgr7, line 59: here we track and re-assing the current eval dataloader on each batch start.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -48,13 +48,28 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
+        old_dataloader_idx = self._current_eval_dataloader_idx
+        self._current_eval_dataloader_idx = dataloader_idx
+        return old_dataloader_idx != dataloader_idx
+
+    def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._current_eval_dataloader_idx = None
```
rohitgr7, line 65: here it is set to None to avoid issues with sequential trainer_fn calls.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -105,12 +124,11 @@ def total_val_batches(self) -> Union[int, float]:
...
-        total_val_batches = 0
-        if self.trainer.enable_validation:
-            is_val_epoch = (self.trainer.current_epoch + 1) % self.trainer.check_val_every_n_epoch == 0
-            total_val_batches = sum(self.trainer.num_val_batches) if is_val_epoch else 0
+        assert self._current_eval_dataloader_idx is not None
+        if self.trainer.sanity_checking:
+            return self.trainer.num_sanity_val_batches[self._current_eval_dataloader_idx]
 
-        return total_val_batches
+        return self.trainer.num_val_batches[self._current_eval_dataloader_idx]
```
rohitgr7, line 131: total_meth_batches returns the num_batches for the current dataloader since we now create a separate bar for each dataloader.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -128,7 +147,18 @@ def total_predict_batches(self) -> Union[int, float]:
         Use this to set the total number of iterations in the progress bar. Can return ``inf`` if the predict dataloader
         is of infinite size.
         """
-        return sum(self.trainer.num_predict_batches)
+        assert self._current_eval_dataloader_idx is not None
+        return self.trainer.num_predict_batches[self._current_eval_dataloader_idx]
+
+    @property
+    def num_val_batches(self) -> Union[int, float]:
```
rohitgr7, line 154: this is used for main progress bar since for that it needs total val batches in all the dataloaders under specific condition only.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -227,12 +244,12 @@ def init_predict_tqdm(self) -> Tqdm:
...
         # The main progress bar doesn't exist in `trainer.validate()`
-        has_main_bar = self._main_progress_bar is not None
+        has_main_bar = not self.trainer.state.fn == "validate"
         bar = Tqdm(
-            desc="Validating",
+            desc=self.validation_description,
             position=(2 * self.process_position + has_main_bar),
             disable=self.is_disabled,
-            leave=False,
+            leave=not has_main_bar,
```
rohitgr7, line 252: this is done, to make sure the progress bar doesn't disappear when calling `trainer.validate`

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -273,62 +291,75 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", *_: Any) -> None:
...
 
-    def on_validation_start(self, trainer: "pl.Trainer", *_: Any) -> None:
-        if trainer.sanity_checking:
-            self.val_progress_bar.total = sum(trainer.num_sanity_val_batches)
-        else:
+    def on_validation_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        if not trainer.sanity_checking:
             self.val_progress_bar = self.init_validation_tqdm()
+
+    def on_validation_batch_start(
```
rohitgr7, line 309: this is the place where we assign the total counter and description whenever a new dataloader comes in.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -273,62 +291,75 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", *_: Any) -> None:
...
             self.main_progress_bar.set_postfix(self.get_metrics(trainer, pl_module))
         self.val_progress_bar.close()
+        super().on_validation_end(trainer, pl_module)
 
-    def on_test_start(self, *_: Any) -> None:
+    def on_test_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         self.test_progress_bar = self.init_test_tqdm()
-        self.test_progress_bar.total = convert_inf(self.total_test_batches)
+
+    def on_test_batch_start(
```
rohitgr7, line 332: same for test/predict.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -347,8 +378,8 @@ def print(self, *args: Any, sep: str = " ", **kwargs: Any) -> None:
             s = sep.join(map(str, args))
             active_progress_bar.write(s, **kwargs)
 
-    def _should_update(self, idx: int) -> bool:
-        return self.refresh_rate > 0 and idx % self.refresh_rate == 0
+    def _should_update(self, current: int, total: Union[int, float]) -> bool:
+        return self.refresh_rate > 0 and (current % self.refresh_rate == 0 or current == total)
```
rohitgr7, line 382: this is done to make sure the progress bar ends correctly when refresh rate isn't aligned with total counter.

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -345,8 +346,13 @@ def on_train_epoch_start(self, trainer, pl_module):
             )
         self.refresh()
 
-    def on_validation_epoch_start(self, trainer, pl_module):
-        if self.total_val_batches > 0:
+    def on_validation_batch_start(
+        self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
+    ) -> None:
+        if self.is_dataloader_changed(dataloader_idx):
+            if self.val_progress_bar_id is not None:
```
carmocca, line 353: Should these 2 be an `or`?

If it does not exist OR the dataloader changed, we want to change it

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -115,6 +115,26 @@ def __getstate__(self) -> Dict:
         # can't pickle the tqdm objects
         return {k: v if not isinstance(v, _tqdm) else None for k, v in vars(self).items()}
 
+    @property
+    def sanity_check_description(self) -> str:
+        return "Validation Sanity Check"
```
carmocca, line 120: This would be more consistent

```suggestion
        return "Sanity Checking"
```

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -227,12 +244,12 @@ def init_predict_tqdm(self) -> Tqdm:
     def init_validation_tqdm(self) -> Tqdm:
         """Override this to customize the tqdm bar for validation."""
         # The main progress bar doesn't exist in `trainer.validate()`
-        has_main_bar = self._main_progress_bar is not None
```
carmocca, line 230: What was the problem with this check?

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca, line 252: Why this change?

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -115,6 +115,26 @@ def __getstate__(self) -> Dict:
         # can't pickle the tqdm objects
         return {k: v if not isinstance(v, _tqdm) else None for k, v in vars(self).items()}
 
+    @property
+    def sanity_check_description(self) -> str:
+        return "Validation Sanity Check"
+
+    @property
+    def train_description(self) -> str:
```
carmocca, line 123: This customization could be left for a separate PR

pytorch_lightning/callbacks/progress/base.py
carmocca, line 154: Why do we need this distinction?

The name is too similar to `total_val_batches`

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -273,62 +291,75 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", *_: Any) -> None:
         self.main_progress_bar.set_description(f"Epoch {trainer.current_epoch}")
 
     def on_train_batch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", *_: Any) -> None:
-        if self._should_update(self.train_batch_idx):
+        if self._should_update(self.train_batch_idx, self.total_train_batches):
             _update_n(self.main_progress_bar, self.train_batch_idx + self._val_processed)
             self.main_progress_bar.set_postfix(self.get_metrics(trainer, pl_module))
 
     def on_train_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        _update_n(self.main_progress_bar, self.train_batch_idx + self._val_processed)
```
carmocca, line 281: Why remove this?

We did not need to pass the `total` to `_update_n` before because the epoch end hooks did update unconditionally.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -48,13 +48,28 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
 
     def __init__(self) -> None:
         self._trainer: Optional["pl.Trainer"] = None
+        self._current_eval_dataloader_idx: Optional[int] = None
```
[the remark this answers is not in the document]
rohitgr7: well, for each dataloader, even the first one, it needs to be assigned inside batch_start hook, now if I make it as 0, then I would have to keep an additional initialization inside `on_validation_epoch_start` hook, and we will have the bar init at 2 different places. If I keep it None then `None!=0` condition will satisfy and it will create the bar accordingly.

Also if FT is initialized with dataloader_idx=1, then default 0 won't work.

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -178,9 +178,6 @@ def _get_max_batches(self) -> List[int]:
             max_batches = self.trainer.num_test_batches
         else:
             if self.trainer.sanity_checking:
-                self.trainer.num_sanity_val_batches = [
```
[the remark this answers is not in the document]
rohitgr7: yeah, that would be better. I'll revert this change.

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: with trainer.test/predict, we would need to display bar once it is finished, with `trainer.validate` this wasn't the case since this was the common hook for fit_val and just val. I have added tests for it.

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: just trying to avoid duplicate manual text at 2 different places.

pytorch_lightning/callbacks/progress/base.py
rohitgr7: yeah! for main progress bar, we need the sum of all val batches, but for the individual val progress bar, we just need the current one. We can move this to tqdm progress bar module directly, but then we would need to do the same for rich as well, so thought maybe keep a common one.

or maybe a better name for this property?

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: the progress bar is updated only in batch-level hooks. check the update in `_should_update`.
I am trying to bring rich and tqdm progress bar API closer. This was done just to make sure that main progress bar exits with proper counter.

pytorch_lightning/callbacks/progress/rich_progress.py
rohitgr7: the condition checks if it does exist, then remove the old one from the display.

pytorch_lightning/callbacks/progress/base.py
carmocca: `total_val_batches_current_epoch`?

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -128,7 +147,18 @@ def total_predict_batches(self) -> Union[int, float]:
...
         """
-        return sum(self.trainer.num_predict_batches)
+        assert self._current_eval_dataloader_idx is not None
+        return self.trainer.num_predict_batches[self._current_eval_dataloader_idx]
+
+    @property
+    def num_val_batches(self) -> Union[int, float]:
+        if (
+            self.trainer.enable_validation
+            and (self.trainer.current_epoch + 1) % self.trainer.check_val_every_n_epoch == 0
```
carmocca, line 157: This could use the logic at the beginning of `TrainingEpochLoop._should_check_val_fx` if we pull it out into a separate method

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: Fine

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: Good catch. Can you fix this in a separate PR with its own test?

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: Good catch. Can you fix this in a separate PR with its own test?

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: This was done on purpose: https://github.com/PyTorchLightning/pytorch-lightning/pull/11069/files#r769739944 because the logger connector state will be different in these two hooks.

> I am trying to bring rich and tqdm progress bar API closer

In any case, I think rich should also update in `epoch_end`.
But there's no need to make both implementations exactly equal, they will have their natural differences as they use different libraries.

pytorch_lightning/callbacks/progress/rich_progress.py
carmocca: Oh I didn't notice the hook change

pytorch_lightning/callbacks/progress/base.py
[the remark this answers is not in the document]
carmocca: If you initialize to 0, what's the problem with the following in a batch start hook?

```python
    if self.bar is None or self.current_dl_idx != dl_idx:
        self.bar = Bar()
```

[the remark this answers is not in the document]
rohitgr7: won't work if someone calls `trainer.fit` and then `trainer.validate`.

else we will have to go with 
```py
if self.bar is None or self.bar.is_closed_or_something or self.current_dl_idx != dl_idx
```
but doing this adds more responsibility on a single hook.
my idea here was to initialize the bar in epoch_start hooks and initialize the counters in batch_start hooks.

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: I tried a few examples.. seems to be working fine.. but maybe I missed something?
can you share a trainer config where it will fail?

pytorch_lightning/callbacks/progress/tqdm_progress.py
carmocca: I know it sucks. It took me a while to get used to doing it too.

But it makes life much easier for reviewers, the person cherry-picking commits for a release and any community person auditing the commits

rohitgr7: good point.. but it's a pain since CI is blocked, and once it opens up.. incoming merge conflicts 😭

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -256,14 +273,15 @@ def on_sanity_check_start(self, *_: Any) -> None:
 
     def on_sanity_check_end(self, *_: Any) -> None:
         self.main_progress_bar.close()
+        self.main_progress_bar = None
```
carmocca, line 276: Can you add a comment for why we dereference the main_progress_bar?

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -475,13 +475,15 @@ def _get_monitor_value(self, key: str) -> Any:
         # this is a separate method to aid in testing
         return self.trainer.callback_metrics.get(key)
 
+    def _is_check_val_epoch(self):
```
carmocca, line 478:
```suggestion
    def _should_check_val_epoch(self):
```

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -475,13 +475,15 @@ def _get_monitor_value(self, key: str) -> Any:
...
+        return (
+            self.trainer.enable_validation
+            and (self.trainer.current_epoch + 1) % self.trainer.check_val_every_n_epoch == 0
+        )
+
     def _should_check_val_fx(self, batch_idx: int, is_last_batch: bool) -> bool:
```
carmocca, line 484: We could rename this for consistency

```suggestion
    def _should_check_val_batch(self, batch_idx: int, is_last_batch: bool) -> bool:
```

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -402,6 +426,7 @@ def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx,
         elif self.val_progress_bar_id is not None:
             # check to see if we should update the main training progress bar
             if self.main_progress_bar_id is not None:
+                # TODO: Fix this in a follow-up
```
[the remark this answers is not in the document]
rohitgr7: I'd prefer to do it in a follow-up, since some of it is related.
Update the TODO.

pytorch_lightning/loops/epoch/training_epoch_loop.py
rohitgr7: this checks for val_check_epoch too internally.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -131,7 +133,27 @@ def total_predict_batches(self) -> Union[int, float]:
...
+        return old_dataloader_idx != dataloader_idx
+
+    def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._current_eval_dataloader_idx = None
+
+    def on_test_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._current_eval_dataloader_idx = None
+
+    def on_predict_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        self._current_eval_dataloader_idx = None
```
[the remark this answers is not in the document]
rohitgr7: since we are using them to reset the dataloader_idx tracker, I can create a simple method which will remove these.

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: removed the dereference.

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -276,36 +354,16 @@ def test_tqdm_progress_bar_value_on_colab(tmpdir):
...
     [
         [2, 3, 1, [1, 2, 3, 4, 5], [1, 2, 3]],
         [0, 0, 3, None, None],
         [1, 0, 3, [1], None],
-        [1, 1, 3, [2], [1]],
+        [1, 1, 3, [1, 2], [1]],
         [5, 0, 3, [3, 5], None],
-        [5, 2, 3, [3, 7], [2]],
-        [5, 2, 6, [7], [2]],
+        [5, 2, 3, [3, 5, 7], [2]],
```
awaelchli, line 365: I don't see how these numbers are changing again #11213
we're not using multiple loaders here

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -75,59 +113,99 @@ def test_tqdm_progress_bar_misconfiguration():
...
+            return self._get_dataloaders()
 
-    # check val progress bar total
-    assert sum(len(loader) for loader in trainer.val_dataloaders) == m
-    assert bar.val_progress_bar.total == m
-    assert not bar.val_progress_bar.leave
+        def validation_step(self, batch, batch_idx, dataloader_idx=None):
+            return
 
-    # main progress bar should have reached the end (train batches + val batches)
```
awaelchli, line 100: why remove these comments? has anything changed?

awaelchli (approved):

tests/callbacks/test_tqdm_progress_bar.py
rohitgr7: yes, we are not but the way the progress is being tracked now is changed to support multiple-dataloaders.
here:
refresh_rate is 3, total train batches = 5
now the first update happens at 3, the second one should happen at 6, but since train batches are limited at 5 (2 more batches), we need to update at 5 too and start with validation after that.

tests/callbacks/test_tqdm_progress_bar.py
rohitgr7: nope.. added back.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -115,6 +115,26 @@ def __getstate__(self) -> Dict:
...
+    def validation_description(self) -> str:
+        return "Validation"
+
+    @property
+    def test_description(self) -> str:
+        return "Testing"
+
+    @property
+    def predict_description(self) -> str:
+        return "Predicting"
```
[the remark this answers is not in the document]
rohitgr7: not really. This is how it was done with rich progress bar. I can change it in a follow-up for both tqdm and rich.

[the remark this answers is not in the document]
kaushikb11: Wdyt is the best way for the users to override if they subclass any of both bars?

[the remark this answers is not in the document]
rohitgr7: for model checkpoint, we have enabled this behavior using class attributes.

CHANGELOG.md
```diff
@@ -302,6 +302,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed `parallel_devices` property in `ParallelStrategy` to be lazy initialized ([#11572](https://github.com/PyTorchLightning/pytorch-lightning/pull/11572))
 
 
+- Update `TQDMProgressBar` to run a separate progress bar for each eval dataloader ([#11657](https://github.com/PyTorchLightning/pytorch-lightning/pull/11657))
```
kaushikb11, line 305:
```suggestion
- Updated `TQDMProgressBar` to run a separate progress bar for each eval dataloader ([#11657](https://github.com/PyTorchLightning/pytorch-lightning/pull/11657))
```

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -100,39 +125,55 @@ def total_train_batches(self) -> Union[int, float]:
...
         of infinite size.
         """
-        return sum(self.trainer.num_test_batches)
+        assert self._current_eval_dataloader_idx is not None
+        return self.trainer.num_test_batches[self._current_eval_dataloader_idx]
 
     @property
-    def total_predict_batches(self) -> Union[int, float]:
-        """The total number of prediction batches, which may change from epoch to epoch.
+    def total_predict_batches_current_dataloader(self) -> Union[int, float]:
```
kaushikb11, line 151: As discussed a while back, it would be considered a breaking change, right?

rohitgr7: yes. I have added the label.

rohitgr7: do you think it should be deprecated first?
even if we do so, it will still break for someone who has implemented custom progress bars.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -49,13 +49,34 @@ def on_train_batch_end(self, trainer, pl_module, outputs, batch_idx):
...
+        self._current_eval_dataloader_idx: Optional[int] = None
 
     @property
     def trainer(self) -> "pl.Trainer":
         if self._trainer is None:
             raise TypeError(f"The `{self.__class__.__name__}._trainer` reference has not been set yet.")
         return self._trainer
 
+    @property
+    def sanity_check_description(self) -> str:
```
Borda, line 61: could be also static?

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -100,39 +125,55 @@ def total_train_batches(self) -> Union[int, float]:
...
 
         Use this to set the total number of iterations in the progress bar. Can return ``inf`` if the predict dataloader
         is of infinite size.
         """
-        return sum(self.trainer.num_predict_batches)
+        assert self._current_eval_dataloader_idx is not None
+        return self.trainer.num_predict_batches[self._current_eval_dataloader_idx]
+
+    @property
+    def total_val_batches(self) -> Union[int, float]:
```
Borda, line 161: how it can be float?

Borda (approved):

pytorch_lightning/callbacks/progress/base.py
rohitgr7: can properties be static?

PR #11689, opened 2022-02-01
Fix rich progress bar metric render on epoch end

### LINKED ISSUE #10653: Progressbar showed different value vs tensorboard and myself output

rohitgr7:
## What does this PR do?

Fixes #10653 (for rich)

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -396,6 +400,10 @@ def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx,
             self._update(self.val_progress_bar_id, self.val_batch_idx, self.total_val_batches)
         self.refresh()
 
+    def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
```
carmocca, line 403: Should we use epoch_end instead? TQDM does

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -396,6 +400,10 @@ def on_validation_batch_end(self, trainer, pl_module, outputs, batch, batch_idx,
...
+        if trainer.state.fn == "fit":
```
carmocca, line 404: What's the problem if we remove this `if`?

pytorch_lightning/callbacks/progress/rich_progress.py
rohitgr7: done!

pytorch_lightning/callbacks/progress/rich_progress.py
rohitgr7: else during `.validate`, it will compute and display the val logs. For now, this is an issue with rich. Fixed it here: https://github.com/PyTorchLightning/pytorch-lightning/pull/11690

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -302,7 +302,7 @@ def on_validation_epoch_end(self, *_: Any) -> None:
         _update_n(self.val_progress_bar, self._val_processed)
 
     def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        if self._main_progress_bar is not None and trainer.state.fn == "fit":
+        if self._main_progress_bar is not None and trainer.state.fn == "fit" and not trainer.sanity_checking:
```
rohitgr7, line 305: removing computation for sanity check since for that we use an invisible main_progress_bar that does not display any logs.

carmocca: I think we should keep it because that's the point of sanity checking, trying the code before even though it does not display anything valuable

pytorch_lightning/callbacks/progress/rich_progress.py
carmocca: Because my observation was wrong.

TQDM:
- Update n on `on_eval_epoch_end`
- Update metrics on `on_eval_end`

So we can do the same for rich (as you did initially!)

pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -373,6 +374,10 @@ def on_validation_epoch_end(self, trainer, pl_module):
         if self.val_progress_bar_id is not None:
             self._update(self.val_progress_bar_id, self.val_batch_idx, self.total_val_batches, visible=False)
 
+    def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
+        if trainer.state.fn == "fit":
```
kaushikb11, line 378: Why not use the Enum?

kaushikb11 (approved):

Borda (approved):

pytorch_lightning/callbacks/progress/rich_progress.py
rohitgr7: it is sort of like a public hook, and I see we don't use the Enum in other callbacks as well.

PR #11711, opened 2022-02-03
Return the output of the optimizer step

### LINKED ISSUE #5235: Lighting wrapped optimizers do not allow return values
This is both a bug report /feature request:

Any optimizer.step function might have a return value. However, NO return values are currently handled by the optimizer.LightningOptimizer class in the "step" function. Optimizers with return values  in their step function therefore do not operate properly.

BugFix: Add return value to to the respective line, something like:
```
return_val=None
if make_optimizer_step:
      return_val=self.__optimizer_step(*args, closure=closure, profiler_name=profiler_name, **kwargs)
....

return return_val
```
Also the `__optimizer_step` function must be able to pass on the return value

cc @borda @rohitgr7

carmocca:
## What does this PR do?

Fixes #5235

None

## Before submitting

## PR review

cc @borda @rohitgr7

### DIFF
--- a/pytorch_lightning/core/lightning.py
+++ b/pytorch_lightning/core/lightning.py
@@ -1544,24 +1544,28 @@ class LightningModule(
         on_tpu: bool = False,
         using_native_amp: bool = False,
         using_lbfgs: bool = False,
-    ) -> None:
+    ) -> Any:
         r"""
-        Override this method to adjust the default way the
-        :class:`~pytorch_lightning.trainer.trainer.Trainer` calls each optimizer.
-        By default, Lightning calls ``step()`` and ``zero_grad()`` as shown in the example
-        once per optimizer. This method (and ``zero_grad()``) won't be called during the
-        accumulation phase when ``Trainer(accumulate_grad_batches != 1)``.
+        Override this method to adjust the default way the :class:`~pytorch_lightning.trainer.trainer.Trainer` calls
+        each optimizer.
+
+        By default, Lightning calls ``step()`` and ``zero_grad()`` as shown in the example once per optimizer.
+        This method (and ``zero_grad()``) won't be called during the accumulation phase when
+        ``Trainer(accumulate_grad_batches != 1)``. Overriding this hook has no benefit with manual optimization.
 
         Args:
             epoch: Current epoch
             batch_idx: Index of current batch
             optimizer: A PyTorch optimizer
             optimizer_idx: If you used multiple optimizers, this indexes into that list.
-            optimizer_closure: Closure for all optimizers. This closure must be executed as it includes the
+            optimizer_closure: The optimizer closure. This closure must be executed as it includes the
                 calls to ``training_step()``, ``optimizer.zero_grad()``, and ``backward()``.
             on_tpu: ``True`` if TPU backward is required
             using_native_amp: ``True`` if using native amp
             using_lbfgs: True if the matching optimizer is :class:`torch.optim.LBFGS`
+
+        Returns:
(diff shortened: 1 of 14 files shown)


pytorch_lightning/core/lightning.py
```diff
@@ -1546,18 +1546,19 @@ def optimizer_step(
...
+        This method (and ``zero_grad()``) won't be called during the accumulation phase when
+        ``Trainer(accumulate_grad_batches != 1)``. Overriding this hook has no benefit with manual optimization.
 
         Args:
             epoch: Current epoch
             batch_idx: Index of current batch
             optimizer: A PyTorch optimizer
             optimizer_idx: If you used multiple optimizers, this indexes into that list.
-            optimizer_closure: Closure for all optimizers. This closure must be executed as it includes the
+            optimizer_closure: The optimizer closure. This closure must be executed as it includes the
```
rohitgr7, line 1561: can we add a small example down below representing the value returned from optimizer.step()?

rohitgr7, line 1561: slightly offtopic, but important IMO. Can you add a note that passing the closure into the optimizer.step() is required to make sure training_step, backward... is called? I have seen a few issues over slack/forums regarding that.

pytorch_lightning/lite/wrappers.py
```diff
@@ -56,7 +56,7 @@ def state_dict(self) -> Dict[str, Tensor]:
 
     def step(self, closure: Optional[Callable] = None) -> None:
```
rohitgr7, line 57:
```suggestion
    def step(self, closure: Optional[Callable] = None) -> Any:
```

rohitgr7 (approved):

pytorch_lightning/core/lightning.py
carmocca: The examples below and `optimizer_closure` description already mention this.

Technically passing it into the step is not required if you don't use `LBFGS`, what's required is that it's always executed.

carmocca: I don't think we need to because the `LightningOptimizer` is invisible to the user. After this PR it works exactly as if it did not exist. So it's more on PyTorch to describe any uses for this behavior.

pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -90,8 +90,10 @@ def optimizer_step(
         # in manual optimization, the closure does not return a value
```
akihironitta, line 90: If I understand it correctly, this comment is no longer true.
```suggestion
```

akihironitta (approved):

pytorch_lightning/plugins/precision/native_amp.py
carmocca: It is true. Why do you think it isnt?

ananthsub (approved):

awaelchli (approved):

tests/core/test_lightning_optimizer.py
```diff
@@ -314,3 +317,31 @@ def test_lightning_optimizer_keeps_hooks(tmpdir):
...
     del lightning_optimizer
     assert len(optimizer._fwd_handles) == 1
+
+
+def test_params_groups_and_state_are_accessible(tmpdir):
+    class TestModel(BoringModel):
+        def training_step(self, batch, batch_idx, optimizer_idx):
+            output = self.layer(batch)
+            loss = self.loss(batch, output)
+            self.__loss = loss
```
tchaton, line 327:
```suggestion
            self._loss = loss
```

PR #11714, opened 2022-02-03
Do not expose return_predictions in LightningCLI predict subcommand

mauvilsa:
## What does this PR do?

For the `predict` subcommand, the `return_predictions` parameter is now skipped. From what I understand, this parameter makes no sense to be an argument of the CLI.

Fixes #10509

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/utilities/cli.py
+++ b/pytorch_lightning/utilities/cli.py
@@ -633,7 +633,7 @@ class LightningCLI:
             "fit": {"model", "train_dataloaders", "val_dataloaders", "datamodule"},
             "validate": {"model", "dataloaders", "datamodule"},
             "test": {"model", "dataloaders", "datamodule"},
-            "predict": {"model", "dataloaders", "datamodule"},
+            "predict": {"model", "dataloaders", "datamodule", "return_predictions"},
             "tune": {"model", "train_dataloaders", "val_dataloaders", "datamodule"},
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,6 +244,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Sorted `SimpleProfiler(extended=False)` summary based on mean duration for each hook ([#11671](https://github.com/PyTorchLightning/pytorch-lightning/pull/11671))
 
 
+- `predict` subcommand in `LightningCLI` no longer exposes `return_predictions` ([#11714](https://github.com/PyTorchLightning/pytorch-lightning/pull/11714))
+
+
 ### Deprecated
 
 - Deprecated `ClusterEnvironment.master_{address,port}` in favor of `ClusterEnvironment.main_{address,port}` ([#10103](https://github.com/PyTorchLightning/pytorch-lightning/pull/10103))


pytorch_lightning/utilities/cli.py
```diff
@@ -633,7 +633,7 @@ def subcommands() -> Dict[str, Set[str]]:
             "fit": {"model", "train_dataloaders", "val_dataloaders", "datamodule"},
             "validate": {"model", "dataloaders", "datamodule"},
             "test": {"model", "dataloaders", "datamodule"},
-            "predict": {"model", "dataloaders", "datamodule"},
+            "predict": {"model", "dataloaders", "datamodule", "return_predictions"},
```
carmocca, line 636: Why does it not make sense?

It could be useful with `LightningCLI(run=False)` or a custom CLI that overrides `LightningCLI.predict`

mauvilsa: `run=False` disables subcommands and is not affected by this change. From what I understand `return_predictions` affects the return value of `Trainer.predict`. And this return value is ignored in the code, so it does nothing whatever the value:

https://github.com/PyTorchLightning/pytorch-lightning/blob/d132a9c3b73bad3e893f809a53401722087bec5e/pytorch_lightning/utilities/cli.py#L823

Maybe I am wrong and not understanding how it works.

carmocca: This works:

```python
import torch

from pytorch_lightning.utilities.cli import LightningCLI


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def predict_step(self, batch, batch_idx: int, dataloader_idx: int = 0):
        return 123

    def predict_dataloader(self):


class MyLightningCLI(LightningCLI):
    def predict(self, **kwargs):
        predictions = self.trainer.predict(**kwargs)
        print(predictions)


if __name__ == "__main__":
    MyLightningCLI(BoringModel)
```

```python
$ python script.py predict --return_predictions=True
```

carmocca: I agree this should be included in the documentation. Would you like to use this PR for it?

`return_predictions` is `True` by default which is useless with the CLI unless you do the above. But I wouldn't have the CLI changing it.

mauvilsa: But I still don't understand what is the benefit of exposing it. When would someone ever want to do `--return_predictions=False`? If I run predict is because I want the predictions. The only other use case I can think of is I want to know how long the predictions take but don't care about the output. But that can be achieved by redirecting to `/dev/null`. On the other hand, keeping it exposed creates confusion.

rohitgr7: > When would someone ever want to do `--return_predictions=False`

possibly saving predictions to the disk rather than returning them due to memory overhead?

mauvilsa: Saving to disk would mean that the `predict` method is implemented and there you save to a file. But this requires `return_predictions=True`, no?

rohitgr7: I can save inside `predict_step` itself to avoid accumulating predictions for the whole dataset and just set `return_predictions=False`

```py
def predict_step(...):
    preds = ...
    # save preds
```

mauvilsa: Okay, now I think I understand the entire picture. Anyway I think some things can be done to support all use cases and be more intuitive. Also the method `LightningCLI.predict` imposes a limitation so I would have thought of doing it a bit different. I would prefer to not document this just yet. For the moment I will close this pull request and will come back with a proposal in an new issue.

PR #11754, opened 2022-02-05
Add `SyncBatchNormPlugin`

### LINKED ISSUE #4349: Support sync batch norm as a plugin
## 🚀 Feature
Support sync batch norm as a plugin so we don't duplicate code across accelerators

### Motivation

It'll look very similar to this: https://github.com/PyTorchLightning/pytorch-lightning/pull/4285

### Pitch

```
import torch
from pytorch_lightning.core.lightning import LightningModule


class SyncBatchNormPlugin(object):
    """
    Plugin to link a custom sync batchnorm implementation to any arbitrary accelerator.
    Example::
        class MySyncBN(SyncBatchNormPlugin):

            def configure_sync_batchnorm(self, model):
                model = MySyncBNWrapper(model)
                return model
        my_sync_bn = MySyncBN()
        trainer = Trainer(sync_batchnorm=True, plugins=[my_sync_bn])
    """

    def configure_sync_batchnorm(self, model: LightningModule) -> LightningModule:
        """
        By default: adds global batchnorm for a model spread across multiple GPUs and nodes.
        Override this to synchronize batchnorm between specific process groups instead
        of the whole world or use a different sync_bn like `apex`'s version.
        Args:
            model: pointer to current :class:`LightningModule`.
        Return:
            LightningModule with batchnorm layers synchronized between process groups
        """
        model = torch.nn.SyncBatchNorm.convert_sync_batchnorm(model, process_group=None)

        return model
```

Do we even need the `sync_batchnorm` flag on the `Trainer` with this plugin?

awaelchli:
## What does this PR do?

Fixes #4349

- Added a simple base class for plugins that wrap layers for syncing during forward/backward
- Added a implementation for native sync batch norm
- Added tests for all cases of configuration

This decouples the baked-in support for sync batchnorm in ddp- and ddp spawn strategies. This would for example enable adding a plugin for Apex-specific sync batch norm in the future.

TODO: 
- [x] Tests
- [x] Rebase onto #11448 

## Before submitting

## PR review

## Did you have fun?

cc @borda @akihironitta


ananthsub (commented):
Back when I filed the issue, my reasoning to make this a plugin was to support both pytorch native & Apex: https://nvidia.github.io/apex/_modules/apex/parallel.html#convert_syncbn_model

Almost all instances I've seen have moved to the native version. But i'm sure there are a handful of users who still might use the apex version. Making this an interface will make it easier to support apex later. That could be a good first issue

awaelchli:
Ah interesting! Then we should probably aim for an abstract interface and a concrete default torch implementation. I'll explore it, thanks!

pytorch_lightning/plugins/sync_batchnorm.py
```diff
@@ -19,7 +20,23 @@
...
+    @abstractmethod
+    def apply(model: "pl.LightningModule") -> "pl.LightningModule":
+        """Override this static method to apply synchronization to the layers of this model."""
+
+    @staticmethod
+    @abstractmethod
+    def revert(model: "pl.LightningModule") -> "pl.LightningModule":
+        """Override this static method to undo all modifications made in :meth:`apply`."""
+
+
```
ananthsub: - i don't think these have to be static methods. someone could provide specialized instances. the strategy would still call self.layer_sync.apply/revert

- making these not nn.Module based means these could be applied to Lite
```suggestion
    @abstractmethod
    def apply(self, model: nn.Module) -> nn.Module:
        """Override this static method to apply synchronization to the layers of this model."""

    @abstractmethod
    def revert(self, nn.Module) -> nn.Module:
        """Override this static method to undo all modifications made in :meth:`apply`."""


```

pytorch_lightning/plugins/sync_batchnorm.py
```diff
@@ -19,7 +20,23 @@
...
+class NativeSyncBatchNorm(ABC):
```
ananthsub:
```suggestion
class NativeSyncBatchNorm(LayerSync):
```

tests/models/test_sync_batchnorm.py
```diff
@@ -1,131 +0,0 @@
...
-        loss = F.cross_entropy(y_hat, y)
-
-        return loss
-
-    def configure_optimizers(self):
-        return torch.optim.Adam(self.linear.parameters(), lr=0.02)
-
-
-# TODO: Fatal Python error: Bus error
-@pytest.mark.skip(reason="Fatal Python error: Bus error")
```
awaelchli, line 72: this test was skipped and outdated (failing for other reasons)
I rewrote it and moved it to parity tests

pytorch_lightning/utilities/distributed.py
```diff
@@ -367,40 +366,6 @@ def _collect_states_on_rank_zero(state: Dict[str, Any]) -> Dict[int, Any]:
     return {rank: _broadcast_object_list(state, rank) for rank in range(torch.distributed.get_world_size())}
 
 
-class _BatchNormXd(torch.nn.modules.batchnorm._BatchNorm):
```
awaelchli, line 370: moved these utilities without modification

pytorch_lightning/plugins/sync_batchnorm.py
awaelchli: done

tests/strategies/test_custom_plugin.py
```diff
@@ -18,27 +18,8 @@
...
-
-class CustomParallelStrategy(DDPStrategy):
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
-        # Set to None so it will be overwritten by the accelerator connector.
-        self.sync_batchnorm = None
-
-
-@RunIf(skip_windows=True)
-def test_sync_batchnorm_set(tmpdir):
```
awaelchli, line 34: moved to `test_accelerator_connector` where it belongs

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -291,6 +294,13 @@ def _check_config_and_set_final_flags(
                     self.checkpoint_io = plugin
                 elif isinstance(plugin, ClusterEnvironment):
                     self._cluster_environment_flag = plugin
+                elif isinstance(plugin, LayerSync):
+                    if sync_batchnorm and not isinstance(plugin, NativeSyncBatchNorm):
```
carmocca, line 298: Shouldn't this check the base interface?

```suggestion
                    if sync_batchnorm and not isinstance(plugin, LayerSync):
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -291,6 +294,13 @@ def _check_config_and_set_final_flags(
...
+                        raise MisconfigurationException(
+                            "You set `Trainer(sync_batchnorm=True)` and provided a `LayerSync` plugin, but this"
```
carmocca, line 300:
```suggestion
                            "You set `Trainer(sync_batchnorm=True)` and provided a `{plugin.__class__.__name__}` plugin, but this"
```

pytorch_lightning/plugins/layer_sync.py
```diff
@@ -0,0 +1,98 @@
...
+    def revert(self, model: Module) -> Module:
+        """Convert the wrapped batchnorm layers back to regular batchnorm layers.
+
+        Args:
+            model: Reference to the current LightningModule
+
+        Return:
+            LightningModule with regular batchnorm layers that will no longer sync across processes.
+        """
+        return _revert_sync_batchnorm(model)
```
carmocca, line 64: I saw that you moved them but why not move the `_revert_sync_batchnorm` directly inside of this method?

pytorch_lightning/strategies/ddp.py
```diff
@@ -86,7 +87,7 @@ def __init__(
         )
         log.detail(f"{self.__class__.__name__}: initializing DDP plugin")
         self._num_nodes = 1
-        self.sync_batchnorm = False
+        self._layer_sync: Optional[LayerSync] = None
```
carmocca, line 90: Should this be defined instead in `ParallelStrategy.__init__`?

carmocca (approved):
Looking great!

pytorch_lightning/plugins/layer_sync.py
```diff
@@ -0,0 +1,98 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+
+import torch
+from torch.nn import Module
+
+
+class LayerSync(ABC):
```
carmocca, line 21: Have you considered making this a contextmanager even if we don't use it with a simple `with` statement?

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: no because I want to express here that `Trainer(sync_batchnorm=True, plugins= NativeSyncBatchNorm()) `is valid and should not error, while any other conflicting choice would create an error.

pytorch_lightning/plugins/layer_sync.py
awaelchli: Because we will have the apex version for sync batchnorm and will probably want to share code.

pytorch_lightning/plugins/layer_sync.py
awaelchli: it requires the model as input. i don't see how a context manager could possibly access the model(s) instances under it.

carmocca: Something like this?

```python
class LayerSync:
    def __init__(self, model):
        self.model = model  # weakref.proxy?

    def __enter__(self) -> Module:
        pass

    def __exit__(self, exc_type, exc_value, exc_traceback) -> Module:
        pass

    def apply(self) -> Module:
        return self.__enter__()

    def revert(self) -> Module:
        return self.__exit__(None, None, None)
```

pytorch_lightning/plugins/layer_sync.py
carmocca: I suggest that we move them outside when we want to share it, just in case it doesn't work out later.

pytorch_lightning/plugins/layer_sync.py
awaelchli: I still don't see this. It has no effect on the code under the context manager. What is the meaning of this?

```py
with LayerSync(model):
    model(x)
    model2(x)
```

Let's not do this. Even if it somehow made sense, nobody can't take advantage of this "feature". Also, we don't want people using plugins outside the trainer. That will only limit us later when we want to develop the plugin further.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -317,6 +320,13 @@ def _check_config_and_set_final_flags(
                     self.checkpoint_io = plugin
                 elif isinstance(plugin, ClusterEnvironment):
                     self._cluster_environment_flag = plugin
+                elif isinstance(plugin, LayerSync):
+                    if sync_batchnorm and not isinstance(plugin, NativeSyncBatchNorm):
+                        raise MisconfigurationException(
+                            f"You set `Trainer(sync_batchnorm=True)` and provided a `{plugin.__class__.__name__}`"
+                            f" plugin, but this is not allowed. Choose one or the other."
```
kaushikb11, line 327:
```suggestion
                            " plugin, but this is not allowed. Choose one or the other."
```

kaushikb11 (approved):
Awesome!

ananthsub (approved):

pytorch_lightning/plugins/layer_sync.py
```diff
@@ -0,0 +1,94 @@
...
+    @abstractmethod
+    def apply(self, model: Module) -> Module:
+        """Override this method to apply synchronization to the layers of this model."""
+
+    @abstractmethod
+    def revert(self, model: Module) -> Module:
+        """Override this method to undo all modifications made in :meth:`apply`."""
+
+
+class NativeSyncBatchNorm(LayerSync):
```
justusschock, line 34: So we don't support the Apex variant of it? I know that we didn't so far, but now that we have an interface for this, we could :)

pytorch_lightning/strategies/parallel.py
```diff
@@ -42,6 +43,7 @@ def __init__(
         super().__init__(accelerator=accelerator, checkpoint_io=checkpoint_io, precision_plugin=precision_plugin)
         self.parallel_devices = parallel_devices
         self.cluster_environment = cluster_environment
+        self._layer_sync: Optional[LayerSync] = None
```
justusschock, line 46: Just a question: Why is this plugin private if other plugins (like cluster env and precision) are not?

justusschock (approved):
LGTM, 2 very minor questions :)

pytorch_lightning/plugins/layer_sync.py
awaelchli: Yes, it's a motivation of this feature. I would be opening a feature issue after we merge this.

pytorch_lightning/strategies/parallel.py
awaelchli: I think mainly as a security because we set it from the accelerator connector. We may also want to expose it as a constructor arg in the future because the others are too. Just trying to give us max flexibility after 1.6 when we declare it as stable. We can always add new optional args and properties :)

tchaton (approved):
LGTM !

PR #11757, opened 2022-02-05
bug fix: restore_optimizers correctly handles non-mapping values in optimizer.state.values()

### LINKED ISSUE #11741: Bug report: This line causes error when using MADGRAD optimizer & resume training
https://github.com/PyTorchLightning/pytorch-lightning/blob/8c07d8bf905e395bbd2142b5df7b185b8e936c41/pytorch_lightning/trainer/connectors/checkpoint_connector.py#L284

The reported error is:
`  File "/home/tiger/.local/lib/python3.7/site-packages/pytorch_lightning/trainer/connectors/checkpoint_connector.py", line 283, in restore_optimizers
    for k, v in state.items():
AttributeError: 'Tensor' object has no attribute 'items'`

For MADGRAD, optimizer.state["k"] is tensor-type instead of mapping type. Therefore, we need to fix this problem for making Pytorch Lightning compatible with such kind of optimizers.

The correct way of fixing this issue is changing it to:
Screen Shot 2022-02-05 at 4 57 59 AM

cc @awaelchli @ananthsub @ninginthecloud @rohitgr7

circlecrystal:
## What does this PR do?

* This PR fixed the bug reported in https://github.com/PyTorchLightning/pytorch-lightning/issues/11741
* To reproduce this bug (and this is exactly what the test code below does): 1) use MADGRAD optmizer 2) use multiple GPUs for training, then resume training
* The main root cause for this bug to exist is because the original restore_optimizers() method cannot correctly handles non-mapping values in optmizer.state.values().

The test code is provided below:

```python
import os

import torch
from torch import nn
from torch.nn import functional as F
from torchvision import transforms
from torchvision.datasets import MNIST
from pytorch_lightning.core.lightning import LightningModule

from madgrad import MADGRAD


class LitMNIST(LightningModule):
    def __init__(self):
        super().__init__()

        # mnist images are (1, 28, 28) (channels, height, width)
        self.layer_1 = nn.Linear(28 * 28, 128)
        self.layer_2 = nn.Linear(128, 256)
        self.layer_3 = nn.Linear(256, 10)

    def forward(self, x):
        batch_size, channels, height, width = x.size()

        # (b, 1, 28, 28) -> (b, 1*28*28)
        x = x.view(batch_size, -1)
        x = self.layer_1(x)
        x = F.relu(x)
        x = self.layer_2(x)
        x = F.relu(x)
        x = self.layer_3(x)

        x = F.log_softmax(x, dim=1)
        return x

        x, y = batch
        logits = self(x)
        loss = F.nll_loss(logits, y)
        return loss 

        return MADGRAD(self.parameters(), lr=1e-6)

    def train_dataloader(self):
        # transforms
        # prepare transforms standard to MNIST
        transform=transforms.Compose([transforms.ToTensor(), transforms.Normalize((0.1307,), (0.3081,))])
        # data
        mnist_train = MNIST(os.getcwd(), train=True, download=True, transform=transform)
        return DataLoader(mnist_train, batch_size=64)


model = LitMNIST()
checkpoint_callback = ModelCheckpoint(save_last=True, save_top_k=1, monitor="loss", mode="min", filename="best")
trainer1 = Trainer(gpus=-1, strategy="dp", max_epochs=1, callbacks=[checkpoint_callback])
trainer1.fit(model)
trainer2 = Trainer(gpus=-1, strategy="dp", max_epochs=2, callbacks=[checkpoint_callback])
trainer2.fit(model, ckpt_path=checkpoint_callback.last_model_path)
```

(Notice that the madgrad.py being used in the test code can be freely downloaded from the official repo: https://github.com/facebookresearch/madgrad/blob/main/madgrad/madgrad.py)

Without this bug fix, the test code will fail and report error message as:
image

Fixes #11741

No.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import collections
 import logging
 import os
 import re
@@ -282,7 +281,7 @@ class CheckpointConnector:
             # avoids OOM
             if self.trainer.root_gpu is not None:
                 for param, state in optimizer.state.items():
-                    if isinstance(state, collections.Mapping):
+                    if isinstance(state, dict):
                         for k, v in state.items():
                             if isinstance(v, torch.Tensor):
                                 state[k] = v.cuda(self.trainer.root_gpu)
--- a/pytorch_lightning/trainer/connectors/checkpoint_connector.py
+++ b/pytorch_lightning/trainer/connectors/checkpoint_connector.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
 import logging
 import os
 import re
@@ -280,10 +281,13 @@ class CheckpointConnector:
             # move optimizer to GPU 1 weight at a time
             # avoids OOM
             if self.trainer.root_gpu is not None:
-                for state in optimizer.state.values():
-                    for k, v in state.items():
-                        if isinstance(v, torch.Tensor):
-                            state[k] = v.cuda(self.trainer.root_gpu)
+                for param, state in optimizer.state.items():
+                    if isinstance(state, collections.Mapping):
+                        for k, v in state.items():
+                            if isinstance(v, torch.Tensor):
+                                state[k] = v.cuda(self.trainer.root_gpu)
+                    elif isinstance(state, torch.Tensor):
+                        optimizer.state[param] = state.cuda(self.trainer.root_gpu)
 
     def restore_lr_schedulers(self) -> None:
         """Restores the learning rate scheduler states from the pre-loaded checkpoint."""
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -532,6 +532,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `RichProgressBar` now correctly shows the `on_epoch` logged values on train epoch end ([#11689](https://github.com/PyTorchLightning/pytorch-lightning/pull/11689))
 
 
+- Fixed `restore_optimizers` for mapping states ([#11757](https://github.com/PyTorchLightning/pytorch-lightning/pull/11757))
+
+
 - Fixed check for available modules ([#11526](https://github.com/PyTorchLightning/pytorch-lightning/pull/11526))
 
 


pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -280,10 +281,13 @@ def restore_optimizers(self) -> None:
...
-                    for k, v in state.items():
-                        if isinstance(v, torch.Tensor):
-                            state[k] = v.cuda(self.trainer.root_gpu)
+                for param, state in optimizer.state.items():
+                    if isinstance(state, collections.Mapping):
+                        for k, v in state.items():
+                            if isinstance(v, torch.Tensor):
+                                state[k] = v.cuda(self.trainer.root_gpu)
+                    elif isinstance(state, torch.Tensor):
+                        optimizer.state[param] = state.cuda(self.trainer.root_gpu)
```
ananthsub, line 290: thanks for working on this!
This logic doesn't need to be in the checkpoint connector anymore. It can go here: 
https://github.com/PyTorchLightning/pytorch-lightning/blob/8ddf9f996d4a98e7be5fc591b7734b6f78313771/pytorch_lightning/strategies/strategy.py#L320-L323

then we can delete lines 280-290 from here

Inside the strategy class, you can replace `self.trainer.root_gpu` with `self.root_device`

this is what it would look like after:
```py
    for param, state in optimizer.state.items():
        if isinstance(state, collections.Mapping):
            for k, v in state.items():
                if isinstance(v, torch.Tensor):
                    state[k] = v.to(self.root_device)
        elif isinstance(state, torch.Tensor):
            optimizer.state[param] = state.to(self.root_device)
```

ananthsub: Actually I wonder if we could directly use this utility: https://github.com/PyTorchLightning/pytorch-lightning/blob/8ddf9f996d4a98e7be5fc591b7734b6f78313771/pytorch_lightning/strategies/strategy.py#L143-L149

circlecrystal: > thanks for working on this! This logic doesn't need to be in the checkpoint connector anymore. It can go here:
> 
> https://github.com/PyTorchLightning/pytorch-lightning/blob/8ddf9f996d4a98e7be5fc591b7734b6f78313771/pytorch_lightning/strategies/strategy.py#L320-L323
> 
> then we can delete lines 280-290 from here
> 
> Inside the strategy class, you can replace `self.trainer.root_gpu` with `self.root_device`
> 
> this is what it would look like after:
> 
> ```python
> if self.root_device.type == "cuda":
>     for param, state in optimizer.state.items():
>         if isinstance(state, collections.Mapping):
>             for k, v in state.items():
>                 if isinstance(v, torch.Tensor):
>                     state[k] = v.to(self.root_device)
>         elif isinstance(state, torch.Tensor):
>             optimizer.state[param] = state.to(self.root_device)
> ```
> 
> fyi @four4fish as this will simpliy device properties further by removing a dependency on `root_gpu`

I'm wondering which is the better way:
1) create a new PR for the code refactoring, and keep this PR solely for the bug fixing purpose
2) fix the bug with the code refactoring

Please tell me which way is preferred as a better practice for Pytorch Lightning. Thanks.

ananthsub: Keep this PR solely for the bugfix. I created #11758 for the refactoring

ananthsub (approved):
Please update the changelog to document this fix

rohitgr7 (approved):

awaelchli:
We didn't add a test for this? We could have easily mocked the behavior of e.g. MADGRAD optimizer. Can we follow up on this to avoid regressions?

PR #11758, opened 2022-02-05
Refactor `Strategy._move_optimizer_states` as utility functions

ananthsub:
## What does this PR do?

Refactoring to make the code easier to test by splitting this out as a separate function. I didn't see any existing tests for this. This will simplify https://github.com/PyTorchLightning/pytorch-lightning/pull/11757

Yes, if custom Strategy implementations overrode `_move_optimizer_to_device`, they will break after this PR

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/strategies/strategy.py
+++ b/pytorch_lightning/strategies/strategy.py
@@ -30,9 +30,10 @@ from pytorch_lightning.plugins.precision import PrecisionPlugin
 from pytorch_lightning.strategies.launchers.base import _Launcher
 from pytorch_lightning.trainer.states import TrainerFn
 from pytorch_lightning.utilities import rank_zero_deprecation
-from pytorch_lightning.utilities.apply_func import apply_to_collection, move_data_to_device
+from pytorch_lightning.utilities.apply_func import move_data_to_device
 from pytorch_lightning.utilities.distributed import ReduceOp
 from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.optimizer import optimizers_to_device
 from pytorch_lightning.utilities.types import _PATH, LRSchedulerConfig, STEP_OUTPUT
 
 TBroadcast = TypeVar("TBroadcast")
@@ -138,7 +139,7 @@ class Strategy(ABC):
         self.accelerator.setup(trainer)
         self.setup_optimizers(trainer)
         self.setup_precision_plugin()
-        self._move_optimizer_state()
+        optimizers_to_device(self.optimizers, self.root_device)
 
     def setup_precision_plugin(self) -> None:
         """Attaches the precision plugin to the accelerator."""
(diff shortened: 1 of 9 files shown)


awaelchli:
@ananthsub do you think we could unify this with some similar lines of code we have in `CheckpointConnector.restore_optimizers`?

ananthsub:
> @ananthsub do you think we could unify this with some similar lines of code we have in `CheckpointConnector.restore_optimizers`?

Yes exactly. that was the motivation after seeing the bug here: https://github.com/PyTorchLightning/pytorch-lightning/issues/11741

See my comment here: https://github.com/PyTorchLightning/pytorch-lightning/pull/11757#discussion_r800075578

pytorch_lightning/utilities/optimizer.py
```diff
@@ -0,0 +1,33 @@
+# Copyright The PyTorch Lightning team.
```
awaelchli, line 1: would it perhaps better fit into the already existing apply_func.py, where we have other utilities to transfer collections of data from device to device?

tests/utilities/test_optimizer.py
```diff
@@ -0,0 +1,26 @@
+import collections
+
+import torch
+
+from pytorch_lightning.utilities.optimizer import optimizer_to_device
+
+
+def test_optimizer_to_device():
+    """Ensure that after the initial seed everything, the seed stays the same for the same run."""
```
awaelchli, line 9: the docstring appears to be from a copy-paste of an unrelated test

pytorch_lightning/utilities/optimizer.py
ananthsub: the reason i added a new file with optimizers in the name was that it's not obvious `apply_func` would contain this sort of utility, which makes it harder for new users to find this.

i could move it to `apply_func` if you feel strongly about it.

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -279,17 +279,6 @@ def restore_optimizers(self) -> None:
...
-            # move optimizer to GPU 1 weight at a time
-            # avoids OOM
-            if self.trainer.root_gpu is not None:
-                for param, state in optimizer.state.items():
-                    if isinstance(state, dict):
-                        for k, v in state.items():
-                            if isinstance(v, torch.Tensor):
-                                state[k] = v.cuda(self.trainer.root_gpu)
-                    elif isinstance(state, torch.Tensor):
-                        optimizer.state[param] = state.cuda(self.trainer.root_gpu)
```
ananthsub, line 292: @SeanNaren I see the DeepSpeed strategy overrides `load_optimizer_state_dict` to do nothing

After this change, do the optimizer states now need to be moved to the root_device within the strategy? Or does  the DeepSpeed engine internally already handle this?

tchaton: @ananthsub The test `test_deepspeed_multigpu_stage_3_warns_resume_training` should fail if the optimizer states where not moved to the root_device ? Can you make sure of this ?

SeanNaren: I had a brief look at the deepspeed code and didn't see anything that would break with these changes, however as @tchaton suggested I think the tests will verify this.

pytorch_lightning/strategies/strategy.py
```diff
@@ -140,14 +141,6 @@ def setup_precision_plugin(self) -> None:
         self.optimizers = optimizers
         self.lr_scheduler_configs = lr_scheduler_configs
 
-    def _move_optimizer_state(self, device: Optional[torch.device] = None) -> None:
-        """Moves the state of the optimizers to the appropriate device if needed."""
-        for opt in self.optimizers:
-            for p, v in opt.state.items():
-                # `self.root_device` would raise error if called outside the spawn process
-                # while training on 8 and more cores.
```
awaelchli, line 148: this is solved I guess, since we pulled up the process spawning to the beginning of Trainer.fit()

awaelchli (approved):

PR #11795, opened 2022-02-07
Track CPU stats with DeviceStatsMonitor

### LINKED ISSUE #11253: CPU System Metrics collection
## 🚀 Feature
Provide CPU profiling similar to GPU and XLA profiling provided by DeviceStatsMonitor. It would be nice if you could specify which device you wanted to profile with DevcieStatsMonitor vs. the profiling defaulting to whatever accelerator you are using.

### Motivation
I am running out of CPU memory and I need to figure out where this is occurring. It would be nice if I could easily monitor CPU stats (memory usage, percent utilization, etc).

### Pitch
Modify `DevcieStatsMonitor` to take a `device` arg that allows you to specify which device to profile. You can then pass multiple `DeviceStatsMonitor` callbacks to `Trainer`. The CPU Monitor can use `psutil` to track common memory attributes.

### Alternatives
N/A

### Additional context

Also discussed here: https://github.com/PyTorchLightning/pytorch-lightning/issues/9032#issuecomment-943743996

cc @borda @kaushikb11 @awaelchli @justusschock @akihironitta @rohitgr7

EricWiener:
## What does this PR do?
Adds CPU monitor tracking to `CPUAccelerator` + adds `cpu_stats` argument to `DeviceStatsMonitor` to add CPU stats tracking if another accelerator is being used.

Fixes #11253

```
        cpu_stats: if ``None``, it will log CPU stats only if the accelerator is CPU.
            It will raise a warning if `psutil` is not installed till version 1.9.0.
            if ``True``, it will log CPU stats regardless of the accelerator, and it will
            raise an exception if `psutil` is not installed.
            if ``False``, it will not log CPU stats regardless of the accelerator.
```

N

## Before submitting

## PR review

## Did you have fun?
Always 🎉


pytorch_lightning/accelerators/cpu.py
```diff
@@ -13,8 +13,9 @@
 # limitations under the License.
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Dict
 
+import psutil
```
ananthsub, line 18: `psutil` is an external dependency right? does this need to be guarded by an import like `if _PSUTIL_AVAILABLE` ? https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/utilities/imports.py

pytorch_lightning/accelerators/cpu.py
```diff
@@ -35,10 +36,19 @@ def setup_environment(self, root_device: torch.device) -> None:
...
     def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return 1
+
+
+def get_cpu_process_metrics() -> dict[str, float]:
+    metrics = {
+        f"vm_percent": psutil.virtual_memory().percent,
+        f"cpu_percent": psutil.cpu_percent(),
+        f"swap_percent": psutil.swap_memory().percent,
```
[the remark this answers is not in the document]
kaushikb11: Would it work on all OSs?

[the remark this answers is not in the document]
EricWiener: @kaushikb11 psutil supports:
Linux
Windows
macOS
FreeBSD, OpenBSD, NetBSD
Sun Solaris
AIX

[Source](https://psutil.readthedocs.io/en/latest/#:~:text=psutil%20currently%20supports%20the%20following,macOS)

Is that sufficient?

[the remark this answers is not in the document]
ananthsub: we can use psutil directly then. i went through the psutil docs and didn't see an advantage to using process vs psutil directly

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,17 +47,21 @@ class DeviceStatsMonitor(Callback):
...
+    def __init__(self, cpu_stats: bool = True) -> None:
+        super().__init__()
+        self.cpu_stats = cpu_stats
+
     def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
         if not trainer.logger:
             raise MisconfigurationException("Cannot use DeviceStatsMonitor callback with Trainer that has no logger.")
 
-    def on_train_batch_start(
+    def _on_train_batch_common(
```
daniellepintz, line 58: Why not keep on_train_batch_start and on_train_batch_end?

EricWiener: The functions we're doing the same thing. Seemed like a lot of duplicated code

daniellepintz: I see. Personally I prefer it how it was previously for clarity and readability. What do others thinks?

EricWiener: @daniellepintz I'm marking this as resolved after making @tchaton 's suggested change, but open to discussing further if you wish

pytorch_lightning/accelerators/cpu.py
```diff
@@ -35,10 +39,25 @@ def setup_environment(self, root_device: torch.device) -> None:
...
+
+def get_cpu_process_metrics() -> dict[str, float]:
+    if not _PSUTIL_AVAILABLE:
+        raise ModuleNotFoundError(
+            "`get_cpu_process_metrics` requires `psutil` to be installed. "
+            + "Install it by running `pip install -U psutil`."
+        )
+
+    metrics = {
+        f"cpu_vm_percent": psutil.virtual_memory().percent,
```
tchaton, line 59: Should we make those keys an enum ?

tchaton (approved):
LGTM ! Great work !

pytorch_lightning/accelerators/cpu.py
EricWiener: Done

EricWiener:
@ananthsub @daniellepintz could I please ask you for further feedback/approval. Thanks!

pytorch_lightning/accelerators/cpu.py
```diff
@@ -48,3 +53,27 @@ def auto_device_count() -> int:
...
+
+class CPUDeviceStatsEnum(LightningEnum):
+    """Enum for CPU device stats."""
+
+    CPU_VM_PERCENT = "cpu_vm_percent"
+    CPU_PERCENT = "cpu_percent"
+    CPU_SWAP_PERCENT = "cpu_swap_percent"
+
+
+def get_cpu_process_metrics() -> dict[str, float]:
```
ananthsub, line 66:
```suggestion
def get_cpu_process_metrics() -> Dict[str, float]:
```

pytorch_lightning/accelerators/cpu.py
```diff
@@ -48,3 +53,27 @@ def auto_device_count() -> int:
...
         return True
+
+
+class CPUDeviceStatsEnum(LightningEnum):
+    """Enum for CPU device stats."""
+
+    CPU_VM_PERCENT = "cpu_vm_percent"
+    CPU_PERCENT = "cpu_percent"
+    CPU_SWAP_PERCENT = "cpu_swap_percent"
+
```
ananthsub, line 64: i prefer defining these as constants within the file if they're used outside the module instead of defining an enum

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -69,6 +76,44 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+
+
+@RunIf(max_torch="1.7")
+@RunIf(min_gpus=1)
+def test_device_stats_gpu_from_nvidia_and_cpu(tmpdir):
+    """Test GPU stats + CPU stats are logged using a logger with Pytorch < 1.8.0."""
+    model = BoringModel()
+    device_stats = DeviceStatsMonitor(cpu_stats=True)
+
+    class DebugLogger(CSVLogger):
```
ananthsub, line 105: this isn't blocking, but we don't need a CSV logger here. we ought to minimize possibilities of writing out to disk. we only need an "in-memory" or dummy logger to examine the contents of metrics

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -24,6 +25,12 @@
 from tests.helpers import BoringModel
 from tests.helpers.runif import RunIf
 
+CPU_METRIC_KEYS = [
+    "cpu_vm_percent",
+    "cpu_percent",
+    "cpu_swap_percent",
+]
```
ananthsub, line 32: this risks falling out of sync with what's defined in the cpu accelerator

pytorch_lightning/accelerators/cpu.py
EricWiener: Pre-commit bot auto changes `Dict` to `dict`: https://pytorch-lightning.slack.com/archives/C01296PBDQV/p1644417099319789

pytorch_lightning/accelerators/cpu.py
```diff
@@ -18,14 +18,18 @@
...
 from pytorch_lightning.utilities.types import _DEVICE
 
 if _PSUTIL_AVAILABLE:
     import psutil
 
+# CPU device metrics
+CPU_VM_PERCENT = "cpu_vm_percent"
+CPU_PERCENT = "cpu_percent"
+CPU_SWAP_PERCENT = "cpu_swap_percent"
+
```
ananthsub, line 32:
```suggestion
# CPU device metrics
_CPU_VM_PERCENT = "cpu_vm_percent"
_CPU_PERCENT = "cpu_percent"
_CPU_SWAP_PERCENT = "cpu_swap_percent"

```

pytorch_lightning/accelerators/cpu.py
```diff
@@ -71,9 +69,8 @@ def get_cpu_process_metrics() -> dict[str, float]:
         )
 
     metrics = {
-        CPUDeviceStatsEnum.CPU_VM_PERCENT: psutil.virtual_memory().percent,
-        CPUDeviceStatsEnum.CPU_PERCENT: psutil.cpu_percent(),
-        CPUDeviceStatsEnum.CPU_SWAP_PERCENT: psutil.swap_memory().percent,
+        CPU_VM_PERCENT: psutil.virtual_memory().percent,
+        CPU_PERCENT: psutil.cpu_percent(),
+        CPU_SWAP_PERCENT: psutil.swap_memory().percent,
```
ananthsub, line 72:
```suggestion
        _CPU_VM_PERCENT: psutil.virtual_memory().percent,
        _CPU_PERCENT: psutil.cpu_percent(),
        _CPU_SWAP_PERCENT: psutil.swap_memory().percent,
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -17,6 +17,11 @@
 import pytest
 
 from pytorch_lightning import Trainer
+from pytorch_lightning.accelerators.cpu import (
+    CPU_PERCENT,
+    CPU_SWAP_PERCENT,
+    CPU_VM_PERCENT,
+)
```
ananthsub:
```suggestion
from pytorch_lightning.accelerators.cpu import (
    _CPU_PERCENT,
    _CPU_SWAP_PERCENT,
    _CPU_VM_PERCENT,
)
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -26,9 +31,9 @@
 from tests.helpers.runif import RunIf
 
 CPU_METRIC_KEYS = [
-    "cpu_vm_percent",
-    "cpu_percent",
-    "cpu_swap_percent",
+    CPU_VM_PERCENT,
+    CPU_SWAP_PERCENT,
+    CPU_PERCENT,
```
ananthsub, line 32:
```suggestion
    _CPU_VM_PERCENT,
    _CPU_SWAP_PERCENT,
    _CPU_PERCENT,
```

ananthsub (approved):

krshrimali (approved):
Nice work, LGTM! 🎉 TPU failures don't look relevant to me, but cc-ing @carmocca to confirm and merge this for you if all looks good to him!

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -32,6 +33,9 @@ class DeviceStatsMonitor(Callback):
     Automatically monitors and logs device stats during training stage. ``DeviceStatsMonitor``
     is a special callback as it requires a ``logger`` to passed as argument to the ``Trainer``.
 
+    Args:
+        cpu_stats: if True, will also record CPU stats in addition to accelerator stats (default: True).
```
rohitgr7, line 37:
```suggestion
        cpu_stats: if True, will also record CPU stats in addition to accelerator stats.
```

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,17 +47,21 @@ class DeviceStatsMonitor(Callback):
         >>> trainer = Trainer(callbacks=[device_stats]) # doctest: +SKIP
     """
 
+    def __init__(self, cpu_stats: bool = True) -> None:
+        super().__init__()
```
rohitgr7, line 51:
```suggestion
```

rohitgr7 (commented):
reviewing more!

docs/source/common/debugging.rst
```diff
@@ -159,6 +159,8 @@ Monitor and log device stats during training with the :class:`~pytorch_lightning
 
     trainer = Trainer(callbacks=[DeviceStatsMonitor()])
 
+Additionally, if using an accelerator besides the CPU, you can add tracking CPU metrics by specifying:
+`DeviceStatsMonitor(cpu_stats=True)`.
```
rohitgr7, line 163:
```suggestion
``DeviceStatsMonitor(cpu_stats=True)``.
```
`cpu_stats` is True by default, so they will always be logged no?

pytorch_lightning/accelerators/cpu.py
```diff
@@ -19,8 +19,17 @@
...
 from pytorch_lightning.accelerators.accelerator import Accelerator
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.imports import _PSUTIL_AVAILABLE
 from pytorch_lightning.utilities.types import _DEVICE
 
+if _PSUTIL_AVAILABLE:
+    import psutil
+
+# CPU device metrics
+_CPU_VM_PERCENT = "cpu_vm_percent"
```
rohitgr7, line 29: why define them as global variables? I don't see them used anywhere else.

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -69,6 +77,44 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
             for f in fields:
                 assert any(f in h for h in metrics.keys())
 
+            for f in CPU_METRIC_KEYS:
+                assert not any(f in h for h in metrics.keys()), "CPU Stats should not be included"
```
rohitgr7, line 81:
```suggestion
                assert not any(f in h for h in metrics), "CPU Stats should not be included"
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -69,6 +77,44 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
             for f in fields:
                 assert any(f in h for h in metrics.keys())
```
rohitgr7, line 78:
```suggestion
                assert any(f in h for h in metrics)
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -69,6 +77,44 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        @rank_zero_only
+        def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
+            fields = [
+                "utilization.gpu",
+                "memory.used",
+                "memory.free",
+                "utilization.memory",
+            ] + CPU_METRIC_KEYS
+            for f in fields:
+                assert any(f in h for h in metrics.keys())
```
rohitgr7, line 116:
```suggestion
                assert any(f in h for h in metrics)
```
also, shouldn't this be `all(...)` everywhere?

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +131,59 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+def test_device_stats_cpu(tmpdir):
+    """Test CPU stats are logged when no accelerator is used."""
+    model = BoringModel()
+
+    class DebugLogger(CSVLogger):
+        @rank_zero_only
+        def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
+            fields = CPU_METRIC_KEYS
+            for f in fields:
+                assert any(f in h for h in metrics.keys())
```
rohitgr7, line 143:
```suggestion
                assert any(f in h for h in metrics)
```

tests/callbacks/test_device_stats_monitor.py
rohitgr7: also, shouldn't this be `all(...)`?

tests/callbacks/test_device_stats_monitor.py
rohitgr7: also, shouldn't this be `all(...)` at L78?

tests/callbacks/test_device_stats_monitor.py
rohitgr7: also, shouldn't this be `all(...)`?

tests/callbacks/test_device_stats_monitor.py
rohitgr7: maybe I am missing something here.
cpu_stats is True by default, so here they should be present?

pytorch_lightning/accelerators/cpu.py
EricWiener: They are used by the tests

tests/callbacks/test_device_stats_monitor.py
EricWiener: No. We are looping through all the keys we want to check with `for f in fields:`. We just want to make sure the key appears in the logged metrics.

Ex: `f` could be `cpu_percent` and the metrics could be:
```
metrics = {'on_train_batch_start/cpu_percent': 18.2, 'on_train_batch_start/cpu_swap_percent': 47.0, 'on_train_batch_start/cpu_vm_percent': 66.3}
```

If we did `all()` when `f = "cpu_percent"` that would be checked that all logged metrics were for `cpu_percent`, which is not what we want. We just want to make sure that all metric keys appear in the logged metrics.

rohitgr7: oh! okay.. I misinterpreted it.

tests/callbacks/test_device_stats_monitor.py
EricWiener: Good call. This test wasn't running before, so it was passing with this mistake

pytorch_lightning/callbacks/device_stats_monitor.py
EricWiener: I moved to match the formatting in `trainer.py`. Is that okay?

pytorch_lightning/accelerators/cpu.py
```diff
@@ -48,3 +57,18 @@ def auto_device_count() -> int:
     def is_available() -> bool:
         """CPU is always available for execution."""
         return True
+
+
+def get_cpu_process_metrics() -> dict[str, float]:
+    if not _PSUTIL_AVAILABLE:
+        raise ModuleNotFoundError(
+            "`get_cpu_process_metrics` requires `psutil` to be installed. "
+            + "Install it by running `pip install -U psutil`."
```
Borda, line 66:
```suggestion
            "`get_cpu_process_metrics` requires `psutil` to be installed."
            " Install it by running `pip install -U psutil`."
```

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -32,6 +33,10 @@ class DeviceStatsMonitor(Callback):
     Automatically monitors and logs device stats during training stage. ``DeviceStatsMonitor``
     is a special callback as it requires a ``logger`` to passed as argument to the ``Trainer``.
 
+    Args:
+        cpu_stats: if True, will also record CPU stats in addition to accelerator stats.
+            Default: ``True``.
```
Borda, line 38:
```suggestion
```

EricWiener: For future reference, how come this change? I saw `trainer.py` use similar formatting

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,17 +47,25 @@ class DeviceStatsMonitor(Callback):
...
+        self.cpu_stats = cpu_stats
+
+    def setup(
+        self,
+        trainer: "pl.Trainer",
+        pl_module: "pl.LightningModule",
+        stage: Optional[str] = None,
+    ) -> None:
         if not trainer.loggers:
             raise MisconfigurationException("Cannot use DeviceStatsMonitor callback with Trainer that has no logger.")
```
rohitgr7, line 60:
```suggestion
            raise MisconfigurationException("Cannot use `DeviceStatsMonitor` callback with `Trainer` that has no logger.")
```
un-related to this PR, but just for consistency.

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +102,128 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
     trainer.fit(model)
 
 
+@RunIf(min_gpus=1)
```
rohitgr7, line 105:
```suggestion
@RunIf(min_gpus=1, min_torch="1.8")
```
although, we run with 1.8 on ci, but this should not fail if someone's running it on their personal machine.

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +102,128 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+def test_device_stats_gpu_from_nvidia_and_cpu(tmpdir):
```
rohitgr7, line 106:
```suggestion
def test_device_stats_gpu_from_torch_and_cpu(tmpdir):
```
can we merge and parameterize with `cpu_tests`. This test and `test_device_stats_gpu_from_nvidia_no_cpu`. They are quite similar I think.

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +102,128 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        logger=DebugLogger(tmpdir),
+        enable_checkpointing=False,
+        enable_progress_bar=False,
+    )
+
+    trainer.fit(model)
+
+
+@RunIf(min_gpus=1)
+def test_device_stats_gpu_from_nvidia_no_cpu(tmpdir):
```
rohitgr7, line 139:
```suggestion
def test_device_stats_gpu_from_torch_no_cpu(tmpdir):
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +102,128 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+    """Test only GPU stat stats are logged using a logger."""
```
rohitgr7, line 140:
```suggestion
    """Test only GPU stat stats are logged using a logger."""
```
also update the docstrings here if you merge them

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -32,6 +33,9 @@ class DeviceStatsMonitor(Callback):
     Automatically monitors and logs device stats during training stage. ``DeviceStatsMonitor``
     is a special callback as it requires a ``logger`` to passed as argument to the ``Trainer``.
 
+    Args:
+        cpu_stats: if True, will also record CPU stats in addition to accelerator stats.
```
rohitgr7, line 37:
```suggestion
        cpu_stats: if ``True``, will also record CPU stats in addition to accelerator stats.
```

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -63,11 +75,26 @@ def on_train_batch_start(
 
         device = trainer.strategy.root_device
         device_stats = trainer.accelerator.get_device_stats(device)
+
+        if self.cpu_stats and device.type != "cpu":
+            # Don't query CPU stats twice if CPU is accelerator
+            device_stats.update(get_cpu_process_metrics())
```
rohitgr7, line 81: can you add somethings info about it in the docs?
if you are not sure where, let me know. I'll push a commit for that :)

rohitgr7 (commented):
quick note everyone:
now CPU stats are logged too by default if running on a different accelerator. I hope this is the expected behavior, if not we can re-think the logic here :)

almost close! just a quick few more comments.

rohitgr7 (approved):
nice work!

pytorch_lightning/accelerators/cpu.py
```diff
@@ -48,3 +57,18 @@ def auto_device_count() -> int:
     def is_available() -> bool:
         """CPU is always available for execution."""
         return True
+
+
+def get_cpu_process_metrics() -> dict[str, float]:
+    if not _PSUTIL_AVAILABLE:
+        raise ModuleNotFoundError(
```
awaelchli, line 64: This is a breaking change for existing device stats users who upgrade lightning. 
I propose alternative options:
1) Make it a warning, skip logging device stats unless they have the package installed
2) Turn it off by default
3) Add psutil to base requirements (not preferred)

In any case, the PR would need a changelog

awaelchli (commented):
Looks great

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +102,100 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        logger=True,
+        enable_checkpointing=False,
+        enable_progress_bar=False,
+    )
+
+    trainer.fit(model)
+
+    # Note that you need to mock where the function is imported
+    # (not where it is defined). Please see the following for
+    # an explanation: https://docs.python.org/3/library/unittest.mock.html#where-to-patch
```
awaelchli, line 194: While I see that error made often by people, let's remove this comment. This should be known by everybody who uses unittest mocks, and if not, a comment in a random test is not the place to educate :)

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -85,6 +102,100 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
+            fields = CPU_METRIC_KEYS
+            for f in fields:
+                assert any(f in h for h in metrics)
+
+    device_stats = DeviceStatsMonitor()
+    trainer = Trainer(
+        default_root_dir=tmpdir,
+        max_epochs=2,
+        limit_train_batches=7,
```
awaelchli, line 161: can we reduce the numbers?
let's set limit_val_batches?

pytorch_lightning/accelerators/cpu.py
EricWiener: I'd prefer to go with 4) if that's good with you. If someone wants CPU metrics and uses `DeviceStatsMonitor` while running with only CPU, it would be a pain if no logging occurred and the user needed to look through the logs to figure out why. Additionally, if someone calls `get_cpu_process_metrics` they likely wouldn't expect to receive an empty dictionary.

rohitgr7: personally, I'd like to have 2) if possible.

maybe we can turn-if off by default, by either:
1. Changing `cpu_stats` to `add_cpu_stats`, which by default is False, but adds CPU stats for CPU accelerator and not for others.
2. default `cpu_stats=None`. If None or True and device is CPU then CPU stats are logged, else with other accelerators CPU stats are not logged. IF False, no CPU stats are logged in either case.

EricWiener: I'll go with 2

EricWiener: Actually, I don't see how this solves the issue of this being a breaking change. If `cpu_stats=None` and the CPU is used and `psutil` isn't installed, then this is a breaking change still

rohitgr7: yes, only if they add a callback, else it won't break other code-bases. not a breaking change but a feature since CPU accelerator never supported logging device stats. It would be a breaking change for other accelerators where `DeviceStatsMonitor` was working fine up until now but would break due to this.

EricWiener: This would still be a breaking change if the user previously used the `DeviceStatsMonitor` callback and had a training script they used both with/without a GPU (or another accelerator). Previously, if they ran their code on CPU only, the `DeviceStatsMonitor` would just do nothing. However, now if they run their code on CPU only (leaving in `DeviceStatsMonitor`) and don't have `psutil` installed, they will get an exception

rohitgr7: okay. it should have raised some error before in that case, but we can't do anything for that now. but at least it should be optional now for other accelerators IMO and should not fail there.

carmocca:
Moving the milestone to 1.7 so this change is done in the same release as #11796

pytorch_lightning/accelerators/cpu.py
kaushikb11: @awaelchli @rohitgr7 I am leaning towards option 2) here.

> Turn it off by default

If the user intends to have `Trainer(callbacks=DeviceStatsMonitor())` for the CPU Accelerator, we should raise an exception and request to install psutil.

kaushikb11: > what about logging cpu_stats with other accelerators?

`cpu_stats` will be False by default. If it's required, user could do `DeviceStatsMonitor(cpu_stats=True)`.

kaushikb11: > cpu_stats: if ``True``, will also record CPU stats in addition to accelerator stats.

"extra emphasis on addition to accelerator stats"

It would be bizarre behavior if user is doing `Trainer(DeviceStatsMonitor(cpu_stats=False))` because it would do absolutely nothing. 

If the `DeviceStatsMonitor` callback is passed to Trainer, the accelerator stats should be logged.

rohitgr7: yeah, that's why I want to confirm, should it be false by default as you mentioned?
```py
cpu_stats will be False by default
```

kaushikb11: yup, will push the changes once we come to the decision

carmocca (approved):
LGTM! Pushed some commits directly to the PR with small improvements

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -23,15 +23,24 @@
...
     r"""
     Automatically monitors and logs device stats during training stage. ``DeviceStatsMonitor``
     is a special callback as it requires a ``logger`` to passed as argument to the ``Trainer``.
 
+    Args:
+        cpu_stats: if ``None``, it will log CPU stats only if the accelerator is CPU.
+            It will raise a warning if `psutil` is not installed till version 1.9.0.
+            if ``True``, it will log CPU stats regardless of the accelerator, and it will
+            raise an exception if `psutil` is not installed.
+            if ``False``, it will not log CPU stats regardless of the accelerator.
```
awaelchli, line 42:
```suggestion
        cpu_stats: if ``None``, it will log CPU stats only if the accelerator is CPU.
            It will raise a warning if `psutil` is not installed till version 1.9.0.
            If ``True``, it will log CPU stats regardless of the accelerator, and it will
            raise an exception if `psutil` is not installed.
            If ``False``, it will not log CPU stats regardless of the accelerator.
```

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,31 +52,63 @@ class DeviceStatsMonitor(Callback):
...
-        batch_idx: int,
-        unused: int = 0,
+        stage: Optional[str] = None,
     ) -> None:
         if not trainer.loggers:
-            raise MisconfigurationException("Cannot use `DeviceStatsMonitor` callback with `Trainer(logger=False)`.")
-
+            raise MisconfigurationException(
+                "Cannot use `DeviceStatsMonitor` callback with `Trainer` that has no logger."
+            )
```
awaelchli, line 67: can we keep the message? 
so it's the same everywhere

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,31 +52,63 @@ class DeviceStatsMonitor(Callback):
...
+        # warn in setup to warn once
+        device = trainer.strategy.root_device
+        if self.cpu_stats is None and device.type == "cpu" and not _PSUTIL_AVAILABLE:
+            rank_zero_warn(
```
awaelchli, line 71: add a todo?

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -110,3 +176,15 @@ def test_prefix_metric_keys(tmpdir):
     separator = "."
     converted_metrics = _prefix_metric_keys(metrics, prefix, separator)
     assert converted_metrics == {"foo.1": 1.0, "foo.2": 2.0, "foo.3": 3.0}
+
+
+def test_warning_being_raised_when_psutil_is_not_available_with_device_stats(monkeypatch):
```
awaelchli, line 181:
```suggestion
def test_device_stats_monitor_warning_when_psutil_not_available(monkeypatch):
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -110,3 +176,15 @@ def test_prefix_metric_keys(tmpdir):
...
+    """Test that warning is raised when psutil is not available."""
+    import pytorch_lightning.callbacks.device_stats_monitor as imports
+
+    monkeypatch.setattr(imports, "_PSUTIL_AVAILABLE", False)
+    monitor = DeviceStatsMonitor()
+    trainer = Trainer()
+    assert trainer.strategy.root_device == torch.device("cpu")
+    with pytest.warns(UserWarning, match="psutil` is not installed"):
```
awaelchli, line 189: add todo?

awaelchli (approved):

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,33 +52,62 @@ class DeviceStatsMonitor(Callback):
...
+            )
 
+    def _get_and_log_device_stats(self, trainer: "pl.Trainer", key: str) -> None:
         if not trainer._logger_connector.should_update_logs:
             return
+        if not trainer.loggers:
+            raise MisconfigurationException("Cannot use `DeviceStatsMonitor` callback with `Trainer(logger=False)`.")
 
         device = trainer.strategy.root_device
+        if self.cpu_stats is None and device.type == "cpu" and not _PSUTIL_AVAILABLE:
```
rohitgr7, line 83: how about making `cpu_stats` a protected and marking it as False when the warning is raised so that you don't have to check the condition again here?

pytorch_lightning/utilities/imports.py
```diff
@@ -114,7 +115,7 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
 _NEPTUNE_GREATER_EQUAL_0_9 = _NEPTUNE_AVAILABLE and _compare_version("neptune", operator.ge, "0.9.0")
 _OMEGACONF_AVAILABLE = _package_available("omegaconf")
 _POPTORCH_AVAILABLE = _package_available("poptorch")
-_HABANA_FRAMEWORK_AVAILABLE = _package_available("habana_frameworks")
+_PSUTIL_AVAILABLE = _package_available("psutil")
```
rohitgr7, line 118: let's add the min version too? `1.9.0`. Just like `_RICH_AVAILABLE`.

pytorch_lightning/utilities/memory.py
```diff
@@ -141,3 +142,24 @@ def get_model_size_mb(model: Module) -> float:
...
     return size_mb
+
+
+# CPU device metrics
+_CPU_VM_PERCENT = "cpu_vm_percent"
+_CPU_PERCENT = "cpu_percent"
+_CPU_SWAP_PERCENT = "cpu_swap_percent"
+
+
+def get_cpu_process_metrics() -> Dict[str, float]:
```
rohitgr7, line 153: keep this inside accelerator/cpu.py?
since we have nvidia stat collection utility inside `accelerator/gpu`.

pytorch_lightning/utilities/memory.py
```diff
@@ -141,3 +142,24 @@ def get_model_size_mb(model: Module) -> float:
...
+    if not _PSUTIL_AVAILABLE:
+        raise ModuleNotFoundError(
+            "Fetching CPU device stats requires `psutil` to be installed."
+            " Install it by running `pip install -U psutil`."
```
rohitgr7, line 157:
```suggestion
            " Install it by running `pip install -U psutil>=1.9.0`."
```

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -54,6 +62,64 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        enable_progress_bar=False,
+        accelerator="cpu",
+    )
+    trainer.fit(model)
+
+
+@RunIf(psutil=True)
+@mock.patch("pytorch_lightning.accelerators.cpu.get_cpu_process_metrics")
+@mock.patch("pytorch_lightning.callbacks.device_stats_monitor.get_cpu_process_metrics")
+def test_device_stats_cpu_queried_once(cpu_metrics_device_stats_mock, cpu_metrics_cpu_accelerator_mock, tmpdir):
```
rohitgr7, line 102: I think this one and the above one can be merged.

pytorch_lightning/utilities/imports.py
kaushikb11: Why 1.9.0?

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,33 +52,62 @@ class DeviceStatsMonitor(Callback):
...
         if not trainer.loggers:
             raise MisconfigurationException("Cannot use `DeviceStatsMonitor` callback with `Trainer(logger=False)`.")
+        # warn in setup to warn once
+        device = trainer.strategy.root_device
+        if self.cpu_stats is None and device.type == "cpu" and not _PSUTIL_AVAILABLE:
+            # TODO: raise an exception from v1.9
+            rank_zero_warn(
+                "`DeviceStatsMonitor` will not log CPU stats as `psutil` is not installed."
+                " To install `psutil`, run `pip install psutil`."
+                " It will raise an exception if `psutil` is not installed post v1.9.0."
```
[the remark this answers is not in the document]
kaushikb11: Here, we meant Lightning v1.9

pytorch_lightning/utilities/memory.py
kaushikb11: Circular imports issue

tests/callbacks/test_device_stats_monitor.py
carmocca: Done

pytorch_lightning/utilities/memory.py
carmocca: We can use a local import. Done

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,33 +52,62 @@ class DeviceStatsMonitor(Callback):
...
     """
 
-    def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
-        if not trainer.loggers:
-            raise MisconfigurationException("Cannot use DeviceStatsMonitor callback with Trainer that has no logger.")
+    def __init__(self, cpu_stats: Optional[bool] = None) -> None:
+        self.cpu_stats = cpu_stats
 
-    def on_train_batch_start(
+    def setup(
```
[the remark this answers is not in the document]
carmocca: IMO yes, in case the user modifies the trainer in between.

tests/callbacks/test_device_stats_monitor.py
```diff
@@ -54,6 +62,40 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
+        default_root_dir=tmpdir,
+        max_epochs=1,
+        limit_train_batches=2,
+        limit_val_batches=0,
+        log_every_n_steps=1,
+        callbacks=device_stats,
+        logger=DebugLogger(tmpdir),
+        enable_checkpointing=False,
+        enable_progress_bar=False,
+        accelerator="cpu",
```
carmocca, line 91: What was the issue with having this test also be run on a GPU environment by setting

`accelerator="auto"`?

kaushikb11: It will fail as CPU Metrics won't be logged for GPU Accelerator when `DeviceStatsMonitor(cpu_stats=None)`

pytorch_lightning/callbacks/device_stats_monitor.py
kaushikb11: Good suggestion. Updated!

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -43,33 +52,62 @@ class DeviceStatsMonitor(Callback):
...
+            rank_zero_warn(
+                "`DeviceStatsMonitor` will not log CPU stats as `psutil` is not installed."
+                " To install `psutil`, run `pip install psutil`."
+                " It will raise an exception if `psutil` is not installed post v1.9.0."
+            )
 
+    def _get_and_log_device_stats(self, trainer: "pl.Trainer", key: str) -> None:
         if not trainer._logger_connector.should_update_logs:
             return
+        if not trainer.loggers:
```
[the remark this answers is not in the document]
kaushikb11: Updated!

pytorch_lightning/callbacks/device_stats_monitor.py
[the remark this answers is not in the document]
rohitgr7: what do you mean modifies? It doesn't log anything if it's not training.

pytorch_lightning/accelerators/cpu.py
```diff
@@ -35,8 +36,8 @@ def setup_environment(self, root_device: torch.device) -> None:
             raise MisconfigurationException(f"Device should be CPU, got {root_device} instead.")
 
     def get_device_stats(self, device: _DEVICE) -> Dict[str, Any]:
-        """CPU device stats aren't supported yet."""
-        return {}
+        """Get CPU stats from psutil."""
```
rohitgr7, line 39:
```suggestion
        """Get CPU stats from ``psutil`` package."""
```

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -23,15 +23,23 @@
...
 
 
 class DeviceStatsMonitor(Callback):
     r"""
     Automatically monitors and logs device stats during training stage. ``DeviceStatsMonitor``
     is a special callback as it requires a ``logger`` to passed as argument to the ``Trainer``.
 
+    Args:
+        cpu_stats: if ``None``, it will log CPU stats only if the accelerator is CPU.
+            It will raise a warning if ``psutil`` is not installed till version 1.9.0.
```
rohitgr7, line 38:
```suggestion
            It will raise a warning if ``psutil`` is not installed till v1.9.0.
```

rohitgr7 (approved):

pytorch_lightning/callbacks/device_stats_monitor.py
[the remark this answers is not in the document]
rohitgr7: If I do:
```py
trainer = Trainer(callbacks=[DeviceStatsMonitor()], logger=False)
trainer.validate(model)
```
IMO it should raise any exception since DeviceStatsMonitor doesn't do anything during `validate/test/predict`.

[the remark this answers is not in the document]
carmocca: Mmm true. We should add support for those in the future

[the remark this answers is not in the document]
rohitgr7: then let's update this too later? @kaushikb11

kaushikb11:
do

carmocca (approved):

PR #11797, opened 2022-02-07
Add `Accelerator.is_available()` interface requirement

### LINKED ISSUE #11818: [RFC] `Add Accelerator.is_available()` interface requirement
### Motivation

Such functionality on the Accelerator abstraction would:
- Enable automatic hardware selection without duplicating code across Trainer & individual accelerator implementations.
- Simplify the accelerator connector logic and rewrite effort: https://github.com/PyTorchLightning/pytorch-lightning/pull/11448
- Enable automatic runtime checking of hardware availability during execution
- Provide consistency with how the Trainer auto-detects the cluster environments natively supported by the framework. The corollary here is `ClusterEnvironment.detect`

https://github.com/PyTorchLightning/pytorch-lightning/blob/9e63281a4c4a62f32cad9801a23b63454f8311be/pytorch_lightning/plugins/environments/cluster_environment.py#L43-L46

https://github.com/PyTorchLightning/pytorch-lightning/blob/9e63281a4c4a62f32cad9801a23b63454f8311be/pytorch_lightning/trainer/connectors/accelerator_connector.py#L810-L812

### Pitch

```py
class Accelerator(ABC):

    @staticmethod
    @abstractmethod
    def is_available() -> bool:
    
    def setup_environment(self, root_device: torch.device) -> None:
        """Setup any processes or distributed connections.
        This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
        environment before setup is complete.
        Raises:
            RuntimeError:
                If corresponding hardware is not found.
        """
        if not self.is_available():
            raise RuntimeError(f"{self.__class__.__qualname__} is not configured to run on this hardware.")
```

```py
class CPUAccelerator(Accelerator):
    @staticmethod
    def is_available() -> bool:
        return True

```

```py
class GPUAccelerator(Accelerator):
   @staticmethod
    def is_available() -> bool:
        return torch.cuda.is_available() and torch.cuda.device_count() > 0 
```

and so on

See a more-detailed implementation here: https://github.com/PyTorchLightning/pytorch-lightning/pull/11797 for what this looks like in practice

To support `Trainer(accelerator="auto")` this is what the logic simplifies to:

```py
for acc_cls in (GPUAccelerator, TPUAccelerator, IPUAccelerator, CPUAccelerator):
    if acc_cls.is_available():
        return acc_cls()
return CPUAccelerator() # fallback to CPU
```

this could be even further simplified if we offered an AcceleratorRegistry, such that the Trainer/AcceleratorConnector didn't need to hardcode the list of accelerators to detect:
```py
for acc_cls in AcceleratorRegistry.impls:
    if acc_cls.is_available():
        return acc_cls()
return CPUAccelerator() # fallback to CPU
```

### Alternatives

Some other alternatives exist here: 
https://github.com/PyTorchLightning/pytorch-lightning/pull/11799
https://github.com/PyTorchLightning/pytorch-lightning/pull/11798

Issues with these approaches:
- Also breaking changes: simply instantiating the accelerator could raise a runtime error if the device isn't available.
- The bigger issue to me is that it does not ease support for `Trainer(accelerator="auto")`. The accelerator connector needs to hardcode & re-implement each of the device checks to determine which Accelerator to even instantiate.

### Additional context

cc @borda @tchaton @justusschock @awaelchli @akihironitta @rohitgr7

ananthsub:
## What does this PR do?

- Enables automatic hardware selection without duplicating code across Trainer & individual accelerator implementations. This can be further enhanced by the addition of an AcceleratorRegistry. The accelerator connector could iterate through all of the registered accelerators, call `acc_cls.is_available()` and determine the hardware immediately.

This aims to simplify the accelerator connector logic and rewrite effort in #11448
This would move the hardcoded, duplicated assertion logic from Trainer constructor checks to a single runtime check at `trainer.fit/validate/test/predict` calls

Given discussion on this PR, we can decide where the assertion on device availability should happen. Either in the accelerator init, setup_environment, or left up to individual accelerators to decide

Fixes #11818

Yes, this now:
- raises a TypeError if a custom accelerator has not implemented the new abstract method
- raises a RuntimeError if the configured hardware is not available during Trainer execution

## Before submitting

- [n/a] Did you list all the **breaking changes** introduced by this pull request?

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -35,13 +35,7 @@ class Accelerator(ABC):
 
         This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
         environment before setup is complete.
-
-        Raises:
-            RuntimeError:
-                If corresponding hardware is not found.
         """
-        if not self.is_available():
-            raise RuntimeError(f"{self.__class__.__qualname__} is not configured to run on this hardware.")
 
     def setup(self, trainer: "pl.Trainer") -> None:
         """Setup plugins for the trainer fit and creates optimizers.
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -35,7 +35,13 @@ class Accelerator(ABC):
 
         This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
         environment before setup is complete.
+
+        Raises:
+            RuntimeError:
+                If corresponding hardware is not found.
         """
+        if not self.is_available():
+            raise RuntimeError(f"{self.__class__.__qualname__} is not configured to run on this hardware.")
 
     def setup(self, trainer: "pl.Trainer") -> None:
         """Setup plugins for the trainer fit and creates optimizers.
@@ -59,3 +65,8 @@ class Accelerator(ABC):
     @abstractmethod
     def auto_device_count() -> int:
         """Get the device count when set to auto."""
+
+    @staticmethod
+    @abstractmethod
+    def is_available() -> bool:
+        """Detect if the hardware is available."""
--- a/pytorch_lightning/accelerators/cpu.py
+++ b/pytorch_lightning/accelerators/cpu.py
@@ -31,6 +31,7 @@ class CPUAccelerator(Accelerator):
             MisconfigurationException:
                 If the selected device is not CPU.
         """
+        super().setup_environment(root_device)
         if root_device.type != "cpu":
             raise MisconfigurationException(f"Device should be CPU, got {root_device} instead.")
 
@@ -42,3 +43,8 @@ class CPUAccelerator(Accelerator):
     def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return 1
+
+    @staticmethod
+    def is_available() -> bool:
+        """CPU is always available for execution."""
+        return True
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -82,7 +82,7 @@ class GPUAccelerator(Accelerator):
 
     @staticmethod
     def is_available() -> bool:
-        return torch.cuda.is_available() and torch.cuda.device_count() > 0
+        return torch.cuda.device_count() > 0
 
 
 def get_nvidia_gpu_stats(device: _DEVICE) -> dict[str, float]:
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -37,8 +37,14 @@ class GPUAccelerator(Accelerator):
         """
         Raises:
             MisconfigurationException:
+                If torch.cuda isn't available
+                If no CUDA devices are found
                 If the selected device is not GPU.
         """
+        if not torch.cuda.is_available():
+            raise MisconfigurationException("GPU Accelerator used, but CUDA isn't available.")
+        if torch.cuda.device_count() == 0:
+            raise MisconfigurationException("GPU Accelerator used, but found no CUDA devices available.")
         if root_device.type != "cuda":
             raise MisconfigurationException(f"Device should be GPU, got {root_device} instead")
         torch.cuda.set_device(root_device)
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -37,8 +37,8 @@ class GPUAccelerator(Accelerator):
         """
         Raises:
             MisconfigurationException:
-                If torch.cuda isn't available
-                If no CUDA devices are found
+                If torch.cuda isn't available.
+                If no CUDA devices are found.
                 If the selected device is not GPU.
         """
         if not torch.cuda.is_available():
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -38,7 +38,7 @@ class GPUAccelerator(Accelerator):
         Raises:
             MisconfigurationException:
                 If torch.cuda isn't available.
-                If no CUDA deices are found.
+                If no CUDA devices are found.
         """
         if not torch.cuda.is_available():
             raise MisconfigurationException("GPU Accelerator used, but CUDA isn't available.")
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -33,18 +33,24 @@ _log = logging.getLogger(__name__)
 class GPUAccelerator(Accelerator):
     """Accelerator for GPU devices."""
 
-    def setup_environment(self, root_device: torch.device) -> None:
+    def __init__(self) -> None:
         """
         Raises:
             MisconfigurationException:
                 If torch.cuda isn't available.
-                If no CUDA devices are found.
-                If the selected device is not GPU.
+                If no CUDA deices are found.
         """
         if not torch.cuda.is_available():
             raise MisconfigurationException("GPU Accelerator used, but CUDA isn't available.")
         if torch.cuda.device_count() == 0:
             raise MisconfigurationException("GPU Accelerator used, but found no CUDA devices available.")
+
+    def setup_environment(self, root_device: torch.device) -> None:
+        """
+        Raises:
+            MisconfigurationException:
+                If the selected device is not GPU.
+        """
         if root_device.type != "cuda":
             raise MisconfigurationException(f"Device should be GPU, got {root_device} instead")
         torch.cuda.set_device(root_device)
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -82,7 +82,7 @@ class GPUAccelerator(Accelerator):
 
     @staticmethod
     def is_available() -> bool:
-        return torch.cuda.is_available() and torch.cuda.device_count > 0
+        return torch.cuda.is_available() and torch.cuda.device_count() > 0
 
 
 def get_nvidia_gpu_stats(device: _DEVICE) -> dict[str, float]:
--- a/pytorch_lightning/accelerators/gpu.py
+++ b/pytorch_lightning/accelerators/gpu.py
@@ -33,24 +33,13 @@ _log = logging.getLogger(__name__)
 class GPUAccelerator(Accelerator):
     """Accelerator for GPU devices."""
 
-    def __init__(self) -> None:
-        """
-        Raises:
-            MisconfigurationException:
-                If torch.cuda isn't available.
-                If no CUDA devices are found.
-        """
-        if not torch.cuda.is_available():
-            raise MisconfigurationException("GPU Accelerator used, but CUDA isn't available.")
-        if torch.cuda.device_count() == 0:
-            raise MisconfigurationException("GPU Accelerator used, but found no CUDA devices available.")
-
     def setup_environment(self, root_device: torch.device) -> None:
         """
         Raises:
             MisconfigurationException:
                 If the selected device is not GPU.
         """
+        super().setup_environment(root_device)
         if root_device.type != "cuda":
             raise MisconfigurationException(f"Device should be GPU, got {root_device} instead")
         torch.cuda.set_device(root_device)
@@ -90,6 +79,10 @@ class GPUAccelerator(Accelerator):
     def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return torch.cuda.device_count()
+
+    @staticmethod
+    def is_available() -> bool:
+        return torch.cuda.is_available() and torch.cuda.device_count > 0
 
 
 def get_nvidia_gpu_stats(device: _DEVICE) -> dict[str, float]:
--- a/pytorch_lightning/accelerators/ipu.py
+++ b/pytorch_lightning/accelerators/ipu.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, Union
 import torch
 
 from pytorch_lightning.accelerators.accelerator import Accelerator
+from pytorch_lightning.utilities import _IPU_AVAILABLE
 
 
 class IPUAccelerator(Accelerator):
@@ -31,3 +32,7 @@ class IPUAccelerator(Accelerator):
         # TODO (@kaushikb11): 4 is the minimal unit they are shipped in.
         # Update this when api is exposed by the Graphcore team.
         return 4
+
+    @staticmethod
+    def is_available() -> bool:
+        return _IPU_AVAILABLE
--- a/pytorch_lightning/accelerators/tpu.py
+++ b/pytorch_lightning/accelerators/tpu.py
@@ -16,7 +16,7 @@ from typing import Any, Dict, Union
 import torch
 
 from pytorch_lightning.accelerators.accelerator import Accelerator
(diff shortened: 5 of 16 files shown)


pytorch_lightning/accelerators/gpu.py
```diff
@@ -37,8 +37,14 @@ def setup_environment(self, root_device: torch.device) -> None:
         """
         Raises:
             MisconfigurationException:
+                If torch.cuda isn't available
+                If no CUDA devices are found
```
carmocca, line 41:
```suggestion
                If torch.cuda isn't available.
                If no CUDA devices are found.
```

carmocca (approved):

tchaton (approved):

pytorch_lightning/accelerators/gpu.py
```diff
@@ -37,8 +37,14 @@ def setup_environment(self, root_device: torch.device) -> None:
...
         Raises:
             MisconfigurationException:
+                If torch.cuda isn't available.
+                If no CUDA devices are found.
                 If the selected device is not GPU.
         """
+        if not torch.cuda.is_available():
+            raise MisconfigurationException("GPU Accelerator used, but CUDA isn't available.")
+        if torch.cuda.device_count() == 0:
+            raise MisconfigurationException("GPU Accelerator used, but found no CUDA devices available.")
```
rohitgr7, line 47: quick ques: why not inside accelerator init?

four4fish: User should be able to create a accelerator object in any environment, we only want to check device availability when call trainer.fit/test/eval, and setup_evn is the first hook called in trainer._run

four4fish (approved):

pytorch_lightning/accelerators/gpu.py
carmocca: I would also prefer the `__init__`, it's generally where strategies run their checks:

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/strategies/ipu.py#L98-L102

https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/strategies/deepspeed.py#L263-L266

rohitgr7: > User should be able to create a accelerator object in any environment, we only want to check device availability when call trainer.fit/test/eval, and setup_evn is the first hook called in trainer._run

quick ques: can you explain more on this? do you mean creating obj in non-GPU env and using it in GPU env? is that possible?

four4fish: @rohitgr7 I think it's possible in production/complex system, for example: User needs to create obj in a config script, and then send the config file to a scheduler to trigger/schedule the training job.

four4fish: >But to the best of my knowledge, the config files define the structure of the object constructor but don't actually initialize them until the job is executed.

Agree. :) 
I saw our internal users/team have config wrappers on top of lightning (for reusability), where defined accelerator obj. I'm not sure whether it will fail if we add check in `__init__`, but feel better to be safe.

ananthsub: I can't recall any instance where the GPU accelerator is specified on a host which does not have GPU devices. In almost every instance I've seen, the entire training script is packaged and run on a GPU host.

I moved the assertion to happen back during `__init__` to fail faster

pytorch_lightning/accelerators/gpu.py
```diff
@@ -33,6 +33,18 @@
 class GPUAccelerator(Accelerator):
     """Accelerator for GPU devices."""
 
+    def __init__(self) -> None:
+        """
+        Raises:
+            MisconfigurationException:
+                If torch.cuda isn't available.
+                If no CUDA deices are found.
```
rohitgr7, line 41:
```suggestion
                If no CUDA devices are found.
```

rohitgr7 (approved):

pytorch_lightning/accelerators/gpu.py
```diff
@@ -33,6 +33,18 @@
...
+        """
+        Raises:
+            MisconfigurationException:
+                If torch.cuda isn't available.
+                If no CUDA devices are found.
+        """
+        if not torch.cuda.is_available():
+            raise MisconfigurationException("GPU Accelerator used, but CUDA isn't available.")
+        if torch.cuda.device_count() == 0:
+            raise MisconfigurationException("GPU Accelerator used, but found no CUDA devices available.")
```
ananthsub, line 46: @carmocca @awaelchli going through #11448, this logic is still replicated across the trainer & each accelerator to support `accelerator="auto"`

What do you think of this instead?
```py
Accelerator(ABC):

@staticmethod
@abstractmethod
def is_available() -> bool:

def setup_environment(self):
    if not self.is_available():
        raise MisconfigurationException("{self.__class__.__qualname__} is not configured to run in this environment")
```

This would function pretty much the same as how `ClusterEnvironment.detect()` works, which means we don't need to duplicate logic across individual accelerators & the accelerator connector.

And the runtime check during trainer execution is automatically applied without being a breaking change that this PR currently introduces.

cc @four4fish

tchaton (approved):
LGTM !

pytorch_lightning/accelerators/gpu.py
```diff
@@ -79,6 +80,10 @@ def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return torch.cuda.device_count()
 
+    @staticmethod
+    def is_available() -> bool:
+        return torch.cuda.is_available() and torch.cuda.device_count() > 0
```
justusschock, line 85:
```suggestion
        return torch.cuda.is_available() and torch.cuda.device_count() > 0
```
 Isn't that redundant? Doesn't `torch.cuda.is_available()` only return `True` if there are also some devices?

justusschock (approved):

pytorch_lightning/accelerators/gpu.py
ananthsub: you're right, this is redundant. device_count internally calls `is_available`

rohitgr7 (dismissed):

pytorch_lightning/accelerators/gpu.py
```diff
@@ -39,6 +39,7 @@ def setup_environment(self, root_device: torch.device) -> None:
             MisconfigurationException:
                 If the selected device is not GPU.
         """
+        super().setup_environment(root_device)
```
rohitgr7, line 42: with DDP it will now fail after `init_dist_connection`

pytorch_lightning/accelerators/tpu.py
```diff
@@ -47,3 +47,7 @@ def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
     def auto_device_count() -> int:
         """Get the devices when set to auto."""
         return 8
+
+    @staticmethod
+    def is_available() -> bool:
```
rohitgr7, line 52: for TPU I think it will fail early if no TPU will be available because we call `connect` before `setup_environment`.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -35,7 +35,13 @@ def setup_environment(self, root_device: torch.device) -> None:
 
         This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
         environment before setup is complete.
+
+        Raises:
+            RuntimeError:
+                If corresponding hardware is not found.
         """
+        if not self.is_available():
+            raise RuntimeError(f"{self.__class__.__qualname__} is not configured to run on this hardware.")
```
carmocca, line 44: If I dont' have a GPU, this says "GPUAccelerator is not configured to run on this hardware"

This makes it sounds like there's a setting for GPUAccelerator that would make it work, when in reality I don't even have the hardware.

I would suggest rewording to "GPUAccelerator cannot run on this hardware" or something similar

pytorch_lightning/accelerators/cpu.py
```diff
@@ -31,6 +31,7 @@ def setup_environment(self, root_device: torch.device) -> None:
             MisconfigurationException:
                 If the selected device is not CPU.
         """
+        super().setup_environment(root_device)
         if root_device.type != "cpu":
```
carmocca, line 35: Would a check like this be moved to `is_available`?

pytorch_lightning/accelerators/ipu.py
```diff
@@ -31,3 +32,7 @@ def auto_device_count() -> int:
         # TODO (@kaushikb11): 4 is the minimal unit they are shipped in.
         # Update this when api is exposed by the Graphcore team.
         return 4
+
+    @staticmethod
+    def is_available() -> bool:
+        return _IPU_AVAILABLE
```
carmocca, line 38: You don't need to do this here but I think it would be nicer to move the _IPU_AVAILABLE implementation here and remove this global variable.

Same comment for the other classes

tests/plugins/test_amp_plugins.py
```diff
@@ -45,6 +45,7 @@ class MyApexPlugin(ApexMixedPrecisionPlugin):
         "SLURM_LOCALID": "0",
     },
 )
+@mock.patch("torch.cuda.is_available", return_value=True)
```
carmocca, line 48: I wonder if we could get away with patching the newly introduced staticmethods instead of torch's. No need to do this in this PR

pytorch_lightning/accelerators/gpu.py
awaelchli: This is currently a no-op. The super implementation is empty.

carmocca:
Does this close #11799 and #11798?

tests/accelerators/test_cpu.py
```diff
@@ -22,6 +22,10 @@ def test_restore_checkpoint_after_pre_setup_default():
     assert not plugin.restore_checkpoint_after_setup
 
 
+def test_availability():
+    assert CPUAccelerator.is_available
```
awaelchli, line 26:
```suggestion
    assert CPUAccelerator.is_available()
```
otherwise it's tautological :)

tests/accelerators/test_ddp.py
```diff
@@ -90,11 +91,14 @@ def test_torch_distributed_backend_env_variables(tmpdir):
...
-@mock.patch("torch.cuda.device_count", return_value=1)
-@mock.patch("torch.cuda.is_available", return_value=True)
 @mock.patch("torch.cuda.set_device")
+@mock.patch("torch.cuda.is_available", return_value=True)
+@mock.patch("torch.cuda.device_count", return_value=1)
+@mock.patch("pytorch_lightning.accelerators.gpu.GPUAccelerator.is_available", return_value=True)
 @mock.patch.dict(os.environ, {"PL_TORCH_DISTRIBUTED_BACKEND": "gloo"}, clear=True)
-def test_ddp_torch_dist_is_available_in_setup(mock_set_device, mock_is_available, mock_device_count, tmpdir):
+def test_ddp_torch_dist_is_available_in_setup(
+    mock_set_device, mock_cuda_available, mock_device_count, mock_gpu_is_available, tmpdir
```
awaelchli, line 100: the order of arguments is incorrect. the decorator closest to the test function corresponds to the first argument, and the one on the very top corresponds to the right-most argument. 

it does not matter here because these are not being used, just as a note.

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -35,7 +35,13 @@ def setup_environment(self, root_device: torch.device) -> None:
 
         This is called before the LightningModule/DataModule setup hook which allows the user to access the accelerator
         environment before setup is complete.
+
+        Raises:
+            RuntimeError:
+                If corresponding hardware is not found.
         """
+        if not self.is_available():
```
carmocca, line 43: I think this should either be called in the init or not called anywhere and we let each accelerator choose.

We cannot be sure the device won't be accessed before `setup_environment`.

tests/accelerators/test_ddp.py
awaelchli: this applies to other test changes below as well.

awaelchli (approved):

pytorch_lightning/accelerators/ipu.py
ananthsub: I agree, I don't think the IPU available and TPU available make sense under imports. That's a good follow up for a separate PR

pytorch_lightning/accelerators/accelerator.py
awaelchli: I agree, we should first make sure setup_env() is the first method to be called in the run method and omit this check until then. 
Also, I prefer not to have this in init, as instantiating an accelerator should be allowed and running it is a different concern.

pytorch_lightning/accelerators/gpu.py
rohitgr7: I am actually pointing to this: https://github.com/PyTorchLightning/pytorch-lightning/blob/8394770d4afa5480f881229b150ac44eaa8c41b0/pytorch_lightning/strategies/ddp.py#L145-L151

pytorch_lightning/accelerators/accelerator.py
ananthsub: > We cannot be sure the device won't be accessed before setup_environment. Examples of this are shown in the comments by @rohitgr7

My aim with adding this to `setup_environment` by default is this represents the *latest* time by when the device check should run. This should not preclude earlier possible checks from running

pytorch_lightning/accelerators/cpu.py
ananthsub: is_available doesn't take an input arguments

accepting a device makes it more like `is_compatible` which is a different intention than a global availability check

pytorch_lightning/accelerators/tpu.py
ananthsub: For TPU and IPU these checks might feel redundant because there's only a limited set of strategies the accelerator is  paired with. Someone may write a custom TPU training strategy, forget to add that check, but get covered by this one.

pytorch_lightning/accelerators/ipu.py
ananthsub: @carmocca there's also a whole XLADeviceUtils which probably can be deduped further for TPUs @kaushikb11 https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/utilities/xla_device.py

pytorch_lightning/accelerators/accelerator.py
ananthsub: @rohitgr7 @awaelchli @carmocca to separate concerns, I can move the check for is_available into a separate PR to unblock this one. Even without this, the rest of the PR is still helpful for the accelerator connector rewrite

ananthsub:
> Does this close #11799 and #11798?

Yes, I will close them after this is merged

pytorch_lightning/accelerators/cpu.py
```diff
@@ -31,6 +31,7 @@ def setup_environment(self, root_device: torch.device) -> None:
             MisconfigurationException:
                 If the selected device is not CPU.
         """
+        super().setup_environment(root_device)
```
rohitgr7, line 34: is this required now?

rohitgr7 (approved):

pytorch_lightning/accelerators/cpu.py
ananthsub: i added it since it overrides the super class, just in case the super class changes its current no-op implementation to something more meaningful later

PR #11805, opened 2022-02-08
Integrate global step with progress tracking

### LINKED ISSUE #7406: `global_step`/`current_epoch` issues

carmocca:
## What does this PR do?

- `trainer.global_step` now returns the progress tracking's optimizer step count.
- `ModelCheckpoint` and other checkpoint logic now use this attribute.
- The previous `trainer.fit_loop.epoch_loop.global_step` counter is now renamed to `trainer.fit_loop.epoch_loop._batches_that_stepped`, a name that matches its actual behaviour.
  - Loggers still use this value as the x-axis step value.

Fixes #7406

- Any access to `trainer.global_step` during an intra-training validation hook will now correctly return the number of optimizer steps taken already. That would be `new_global_step == master_global_step + 1`. In pseudocode

Before:

```python
training_step()
validation_if_necessary()
global_step++
```

Now:
```python
training_step()
global_step++
validation_if_necessary()
```

- Saved checkpoints that use the global step value as part of the filename are now increased by 1 for the reason in the bullet before.

- If users were using TBPTT or multiple optimizes, the `trainer.global_step` value will account for those and be different from the value in current master.

- The `Trainer` arguments `{min,max}_steps` compare with the new `global_step` value so they suffer from the same breaking changes. In the case of multiple optimizers or TBPTT users will need to adjust them.

## Before submitting

## PR review

## Did you have fun?

cc @tchaton @rohitgr7 @akihironitta @awaelchli @ananthsub @ninginthecloud @borda @carmocca


pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -368,12 +348,8 @@ def save_checkpoint(self, trainer: "pl.Trainer") -> None:
         """
         self._validate_monitor_key(trainer)
 
-        # track epoch when ckpt was last checked
-        global_step = trainer.global_step
-        self._last_global_step_saved = global_step
```
carmocca, line 373: This should not be set until we are sure we saved

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -821,9 +800,7 @@ def test_model_checkpoint_save_last_checkpoint_contents(tmpdir):
     ckpt_last_epoch = torch.load(path_last_epoch)
     ckpt_last = torch.load(path_last)
 
-    # `-1` because this checkpoint is saved `on_train_epoch_end` which is considered part of the epoch so the
-    # `current_epoch` count has not been increased yet
```
carmocca, line 825: Both the "last" checkpoint and "top-k" checkpoint are now saved together.
We had a bug where we saved a duplicate "last" checkpoint `on_train_end`

tests/checkpointing/test_trainer_checkpoint.py
```diff
@@ -71,19 +71,3 @@ def validation_step(self, batch, batch_idx):
             assert best_model_path.endswith(f"epoch=0{idx}.ckpt")
         else:
             assert f"epoch={idx + 1}" in best_model_path
-
-
-def test_accumulated_gradient_batches_with_ckpt_path(tmpdir):
```
carmocca, line 76: I'm unsure about what this test intended to test, there's no assertions

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -319,22 +315,6 @@ def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModul
             return
         self.save_checkpoint(trainer)
 
-    def on_train_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
```
carmocca, line 322: This is an ancient artifact of the days we had a manual `_check_checkpoint_callback` call in the `Trainer`.

It's not necessary anymore because we will have already saved for this global step in either `on_validation_end` or in `on_train_epoch_end`, so `should_skip_saving_checkpoint = True` and this never runs.

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -246,17 +253,14 @@ def on_advance_end(self) -> None:
             self._run_validation()
             self.trainer.training = True
 
-        # -----------------------------------------
-        # SAVE LOGGERS (ie: Tensorboard, etc...)
-        # -----------------------------------------
-        self._save_loggers_on_train_batch_end()
```
awaelchli, line 252: I don't understand why we change the order if loggers depend on the fake counter. I would not expect the behavior to change for loggers. For them, what once was named global step is now called `_batches_that_stepped`.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -312,14 +299,12 @@ def on_advance_end(self) -> None:
...
         self.epoch_progress.increment_completed()
 
-        # the global step is manually decreased here due to backwards compatibility with existing loggers
-        # as they expect that the same step is used when logging epoch end metrics even when the batch loop has
-        # finished. this means the attribute does not exactly track the number of optimizer steps applied.
-        # TODO(@carmocca): deprecate and rename so users don't get confused
-        self.global_step -= 1
+        # we manually decrease here because loggers expect that the same step is used when logging epoch-end metrics
+        # even when the batch loop has finished
+        self.epoch_loop._batches_that_stepped -= 1
```
awaelchli, line 304: great!

tests/loggers/test_mlflow.py
```diff
@@ -136,7 +136,7 @@ def test_mlflow_log_dir(client, mlflow, tmpdir):
     assert trainer.log_dir == logger.save_dir
     trainer.fit(model)
     assert trainer.checkpoint_callback.dirpath == (tmpdir / "exp-id" / "run-id" / "checkpoints")
-    assert set(os.listdir(trainer.checkpoint_callback.dirpath)) == {"epoch=0-step=0.ckpt"}
+    assert set(os.listdir(trainer.checkpoint_callback.dirpath)) == {"epoch=0-step=1.ckpt"}
```
awaelchli, line 139: so basically now, we will never see a step=0 in the checkpointed files unless you manually save there?

tests/models/test_amp.py
```diff
@@ -199,7 +199,9 @@ def configure_optimizers(self):
     assert str(trainer.amp_backend) == "AMPType.APEX"
     trainer.fit(model)
     assert trainer.state.finished, f"Training failed with {trainer.state}"
-    assert bwd_mock.call_count == 10
+    # `max_steps` is fulfilled in the third batch first optimizer, but we don't check the loop
+    # `done` condition until all optimizers have run, so the number of backwards is higher than `max_steps`
+    assert bwd_mock.call_count == 6
```
awaelchli, line 204: `assert _batches_that_stepped == 10` would hold?

tests/plugins/test_checkpoint_io_plugin.py
```diff
@@ -52,7 +52,7 @@ def test_checkpoint_plugin_called(tmpdir):
     )
     trainer.fit(model)
 
-    assert checkpoint_plugin.save_checkpoint.call_count == 5
+    assert checkpoint_plugin.save_checkpoint.call_count == 4
```
awaelchli, line 55: is this decrease due to the save_last / topk bug mentioned above? but save_last is False by default

awaelchli: or because you removed `on_train_end`?

pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -231,16 +231,20 @@ def restore_loops(self) -> None:
         if not self._loaded_checkpoint:
             return
 
-        self.trainer.fit_loop.global_step = self._loaded_checkpoint["global_step"]
+        fit_loop = self.trainer.fit_loop
+        # set the `global_step` value for old checkpoints without the progress tracking state.
```
awaelchli, line 235: "for old checkpoints" 
can we mention the version?

awaelchli (approved):
mama mia

tests/loggers/test_mlflow.py
carmocca: Correct

pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: This change is not strictly necessary, but I found it more intuitive to put it alongside the increase, also having it after avoids the `+ 1` inside `_save_loggers_on_train_batch_end`

tests/plugins/test_checkpoint_io_plugin.py
carmocca: The latter if I'm not mistaken.

tests/models/test_amp.py
carmocca: No, `assert _batches_that_stepped == 3`.

As described in the comment, `max_steps=5` is fulfilled in the third batch, first optimizer.

So we do 1 extra backward because the loop done is not checked yet.
(does not impact batches_that_stepped)

This is a breaking change as described in the PR intro:

> The Trainer arguments {min,max}_steps compare with the new global_step value so they suffer from the same breaking changes. In the case of multiple optimizers or TBPTT users will need to adjust them.

CHANGELOG.md
```diff
@@ -288,6 +288,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `trainer.current_epoch` value is now increased by 1 during and after `on_train_end` ([#8578](https://github.com/PyTorchLightning/pytorch-lightning/pull/8578))
 
 
+- The `trainer.global_step` value now accounts for multiple optimizers and TBPTT splits ([#11805](https://github.com/PyTorchLightning/pytorch-lightning/pull/11805))
+
+
+- The `trainer.global_step` value is now increased right after the `optimizer.step()` call which will impact users who access it during an intra-training validation hook ([#11805](https://github.com/PyTorchLightning/pytorch-lightning/pull/11805))
+
+
+- The `trainer.global_step` value is now increased by one when included as part of the checkpoint's filename ([#11805](https://github.com/PyTorchLightning/pytorch-lightning/pull/11805))
```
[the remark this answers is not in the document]
carmocca: Made it more explicit

pytorch_lightning/loops/fit_loop.py
```diff
@@ -186,7 +173,7 @@ def done(self) -> bool:
         if self.trainer.should_stop:
             # early stopping
             met_min_epochs = self.epoch_progress.current.processed >= self.min_epochs if self.min_epochs else True
-            met_min_steps = self.global_step >= self.min_steps if self.min_steps else True
+            met_min_steps = self.epoch_loop.global_step >= self.min_steps if self.min_steps else True
```
rohitgr7, line 176: just a little skeptical about this behavior change.
If you go with this, should we mention this in docs?

rohitgr7 (approved):

pytorch_lightning/loops/fit_loop.py
carmocca: Done. `max_step` should link now to `global_step` where we mention multiple optimizers and tbptt

tchaton (approved):
Great work !

yifuwang:
Hi @carmocca, it seems that `last.ckpt` is not always generated anymore (due to the removal of `ModelCheckpoint.on_train_end`)? Some of Meta's internal test broke because they expected `last.ckpt` to exist but it wasn't produced anymore. Can you help us understand under what circumstances `last.ckpt` was generate before this change, but not anymore after this change?

@jjenniferdai @ananthsub

jjenniferdai:
I think `on_train_end` should be brought back. An example of when `last.ckpt` used to be generated and now is not - if the `ModelCheckpoint` trigger qualifications do not pass (e.g. `every_n_epochs = 10` but trainer `max_epochs=1`). Before we would still save `last.ckpt` from `on_train_end`, now we save nothing. If `save_last` is set users would expect `last.ckpt` to get saved.

yifuwang:
Hi @carmocca, this change broke many Meta's internal use cases' tests. To summarize how it's affecting us:

- `last.ckpt` is not always generated anymore even when `save_last` is set (due to the removal of `ModelCheckpoint.on_train_end`). This seems to have broken the contract "save_last: When ``True``, always saves the model at the end of the epoch to a file `last.ckpt`".
- As you mentioned, the step count in checkpoint filenames is incremented by 1.
- The steps at which loggers are flushed are shifted by 1.
- `trainer.fit_loop.global_step` is not set-able anymore (I know this might be considered an internal implementation detail, but we have use cases set it for both testing as well as customized checkpoint load. I wonder if we can add a setter for it.)

I wonder:
- Are all these changes are intended?
- If yes, are they all necessary? (e.g. can we keep the checkpointing/logging behavior intact by compensating the filename/flushing condition?)
- If all these are intended and necessary, can we introduce individual behavior changes separately over time?

awaelchli:
@yifuwang After reading your quote:

> This seems to have broken the contract "save_last: When True, always saves the model at the end of the epoch to a file last.ckpt".

I'm updating the docs for that argument. The description is wrong, the behavior is different in general. With default checkpointing behavior, the statement is still true though. Have you found a case where it is not?

yifuwang:
Thanks for the prompt response @awaelchli! Please see the example provided by @jjenniferdai above for how the behavior has differed after this change:

> An example of when last.ckpt used to be generated and now is not - if the ModelCheckpoint trigger qualifications do not pass (e.g. every_n_epochs = 10 but trainer max_epochs=1). Before we would still save last.ckpt from on_train_end, now we save nothing. If save_last is set users would expect last.ckpt to get saved.

carmocca:
`ModelCheckpoint` checks whether it has already saved at the current global step with https://github.com/PyTorchLightning/pytorch-lightning/blob/7ee690758ccad7f702460d056f6369c1d4371a46/pytorch_lightning/callbacks/model_checkpoint.py#L387

Before this PR, with all the manual decrements and increments to `global_step` we had a "bug" where we were saving a "duplicate" last checkpoint even though we had already saved at this global step. After fixing it, the `on_train_end` override became redundant: https://github.com/PyTorchLightning/pytorch-lightning/pull/11805#discussion_r812868075

As Adrian mentioned, the existing implementation meant the doc were not strictly correct.

One could avoid the issues you are seeing by instantiating a separate `ModelCheckpoint` that only saves "last" checkpoints. Then, the state of each "saving mode" will not impact the others: https://github.com/PyTorchLightning/pytorch-lightning/blob/7ee690758ccad7f702460d056f6369c1d4371a46/pytorch_lightning/callbacks/model_checkpoint.py#L366-L373

Following this structure would also make sense if one wants to checkpoint `every_n_epochs=N` and `save_last=True` but wants `save_last` to check every epoch. If setting both makes "last" checkpoints check for both conditions during training (the case before this PR), then the same logic should apply at the end of training.

> The steps at which loggers are flushed are shifted by 1.

This might not have been intended. Can you elaborate on what's the new observed and expected behaviour?

jjenniferdai:
> An example of when last.ckpt used to be generated and now is not - if the ModelCheckpoint trigger qualifications do not pass (e.g. every_n_epochs = 10 but trainer max_epochs=1). Before we would still save last.ckpt from on_train_end, now we save nothing. If save_last is set users would expect last.ckpt to get saved.

I think users would still expect to save `last.ckpt` in this example - right? It used to, but no longer does - because `ModelCheckpoint.save_checkpoint()` never gets called (@carmocca - i.e. all the logic in your above comment) both before and after, but before `last.ckpt` was still correctly generated in `on_train_end`

yifuwang:
@carmocca for the following snippet, a `last.ckpt` is generated before this PR, but not anymore after:

```
import uuid

import torch
from pytorch_lightning import Trainer, LightningModule


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

if __name__ == "__main__":
    tmpdir = f"/tmp/{uuid.uuid4()}"
    print(tmpdir)

    trainer = Trainer(
        default_root_dir=tmpdir,
        max_epochs=1,
        callbacks=[ModelCheckpoint(dirpath=tmpdir, every_n_epochs=10, save_last=True)],
        enable_checkpointing=True,
    )
    model = BoringModel()
    trainer.fit(
        model, train_dataloaders=DataLoader(RandomDataset(32, 64), batch_size=2)
    )
```

The following contract was respected prior to this PR, but not anymore after:

> save_last: When True, always saves the model at the end of the epoch to a file last.ckpt

It is a BC-breaking change that changes a behavior some users rely on, regardless of whether it is believed to be a "bug".

carmocca:
Yes, I'm fine restoring the previous behaviour for `every_n_epochs` as it wasn't an intended change of this PR. Will be done before the release.

zhong-yy:
I think it can cause confusion, especially when using the default filename ModelCheckpoint(filename="{epoch}-{step}"), where epoch numbers from 0 and step numbers from 1. See https://github.com/Lightning-AI/lightning/issues/16636

 Also, it is too counter-intuitive to number something from 1 instead of 0 in python. I can't find any explicit statement about this numbering stragtegy in the documentation [ModelCheckpoint](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.callbacks.ModelCheckpoint.html#pytorch_lightning.callbacks.ModelCheckpoint).

PR #11811, opened 2022-02-08
Fix horovod installation `base-cuda` Dockerfile

akihironitta:
## What does this PR do?
Fixes a similar problem to #11596, but the issue is horovod this time.

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [n/a] Did you make sure to **update the documentation** with your changes? (if necessary)
- [n/a] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [n/a] Did you list all the **breaking changes** introduced by this pull request?
- [n/a] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @tchaton @rohitgr7 @akihironitta @carmocca @borda

PR #11819, opened 2022-02-09
CI: sanity check for req. pkgs

Borda:
## What does this PR do?

it shall prevent situations like #11811 when we were silently skipping some tests which shall be running

## Before submitting

## PR review

## Did you have fun?


requirements/_check_avail_extras.py
```diff
@@ -0,0 +1,16 @@
...
+import torchtext
+
+print(matplotlib.__version__)
+print(torchtext.__version__)
+print(omegaconf.__version__)
+print(hydra.__version__)
+print(jsonargparse.__version__)
+# todo
+# print(gcsfs.__version__)
+print(rich.__name__)
```
rohitgr7, line 16: rich people and rich packages go by name and other common silly packages by version 😆

rohitgr7 (approved):

.azure-pipelines/ipu-tests.yml
```diff
@@ -53,8 +53,8 @@ jobs:
         export GIT_TERMINAL_PROMPT=1
         python -c "fname = 'requirements/extra.txt' ; lines = [line for line in open(fname).readlines() if 'fairscale' not in line] ; open(fname, 'w').writelines(lines)"
         python -c "fname = 'requirements/extra.txt' ; lines = [line for line in open(fname).readlines() if 'horovod' not in line] ; open(fname, 'w').writelines(lines)"
-        python ./requirements/adjust_versions.py requirements/extra.txt
-        python ./requirements/adjust_versions.py requirements/examples.txt
+        python ./requirements/_adjust_versions.py requirements/extra.txt
+        python ./requirements/_adjust_versions.py requirements/examples.txt
```
carmocca, line 57: Can we keep the non-prefix name? This looks ugly and this code is not importable, we don't need to signify that this is "protected"

.github/workflows/ci_test-conda.yml
```diff
@@ -11,6 +11,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.head_ref }}
   cancel-in-progress: ${{ ! (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release/')) }}
 
+defaults:
+  run:
+    shell: bash -l {0}
```
carmocca, line 16: q: What does this do?

.azure-pipelines/gpu-tests.yml
```diff
@@ -60,6 +60,8 @@ jobs:
     - bash: |
         python requirements/collect_env_details.py
         python -c "import torch ; mgpu = torch.cuda.device_count() ; assert mgpu >= 2, f'GPU: {mgpu}'"
+        python requirements/_check_avail_strategies.py
+        python requirements/_check_avail_extras.py
```
carmocca, line 64: These are already printed with the `pip list` in the previous step (L57)

.github/workflows/ci_test-full.yml
```diff
@@ -135,6 +129,10 @@ jobs:
         path: Datasets
         key: pl-dataset
 
+    - name: Sanity check
+      run: |
+        python requirements/_check_avail_extras.py
```
carmocca, line 134: Same comment about pip list (L106)

carmocca (commented):
I don't think we need this. All this information is already printed

.azure-pipelines/ipu-tests.yml
Borda: ok, then lets use - instead of _ in the name to be sure it is not importable

.github/workflows/ci_test-conda.yml
Borda: it is replacement for having `shell: bash -l {0}` in each `run` also recommended the action page we are using here

.azure-pipelines/gpu-tests.yml
Borda: yes, but the `pip list` is never verified, the print here is just to have qazi testing that the import really works...

.github/workflows/ci_test-full.yml
Borda: the same as above, we need to try to import these packages, if we make mistake in installation, related test are just skipped and we don't know about it...

akihironitta (approved):

.azure-pipelines/gpu-tests.yml
```diff
@@ -60,6 +60,8 @@ jobs:
     - bash: |
         python requirements/collect_env_details.py
         python -c "import torch ; mgpu = torch.cuda.device_count() ; assert mgpu >= 2, f'GPU: {mgpu}'"
+        python requirements/check-avail-strategies.py
+        python requirements/check-avail-extras.py
```
akihironitta, line 64: Even `horovod` is not found, the workflow doesn't fail here in CI:
https://dev.azure.com/PytorchLightning/pytorch-lightning/_build/results?buildId=56115&view=logs&j=3afc50db-e620-5b81-6016-870a6976ad29&t=059bb337-5de1-5e6a-4c56-42f55a328c0e&l=37

Borda: VERY good catch, thx, let's see if [7219995](https://github.com/PyTorchLightning/pytorch-lightning/pull/11819/commits/7219995ff038096c37f7412f3312ed1f434467a1) helps

akihironitta: It seems to be working now :) https://dev.azure.com/PytorchLightning/pytorch-lightning/_build/results?buildId=56189&view=logs&j=3afc50db-e620-5b81-6016-870a6976ad29&t=059bb337-5de1-5e6a-4c56-42f55a328c0e

akihironitta:
As #11811 is merged into master, all running jobs here should be green now 👀

requirements/check-avail-extras.py
```diff
@@ -0,0 +1,16 @@
...
+import torchtext
+
+print(matplotlib.__version__)
+print(torchtext.__version__)
+print(omegaconf.__version__)
+print(hydra.__version__)
+print(jsonargparse.__version__)
+# todo
+# print(gcsfs.__version__)
+print(rich.__name__)
```
carmocca, line 16: > yes, but the pip list is never verified, the print here is just to have qazi testing that the import really works...

In that case, what we care about here is importing them. Not printing the version because this will be more reliably done by `pip list`.

```suggestion
```

requirements/check-avail-strategies.py
```diff
@@ -0,0 +1,9 @@
+import bagua
+import deepspeed
+import fairscale
+import horovod.torch
+
+print(fairscale.__version__)
+print(deepspeed.__version__)
+print(bagua.__version__)
+print(horovod.torch.nccl_built())
```
carmocca, line 9:
```suggestion
```

requirements/check-avail-strategies.py
```diff
@@ -0,0 +1,9 @@
+import bagua
+import deepspeed
+import fairscale
+import horovod.torch
```
carmocca, line 4:
```suggestion
import horovod.torch; horovod.torch.nccl_built()
```

Also, does this need an `assert`? Doesn't `nccl_built` return a bool?

.github/workflows/ci_test-full.yml
```diff
@@ -127,6 +120,7 @@ jobs:
           pip install --no-cache-dir -r requirements/horovod.txt
         fi
         horovodrun --check-build
+        python -c "import horovod.torch; horovod.torch.nccl_built()"
```
carmocca, line 123: Might as well do it in the newly added file with the import

```suggestion
```

Borda: here we assume only this `horovod` strategy so don't think it needs to make a special file for it...

carmocca (approved):

.github/workflows/ci_test-full.yml
carmocca, line 123: `horovod.torch.nccl_built()` returns an integer so without an assertion this does nothing

And anyways we don't want to check this because this workflow is CPU

```suggestion
        python -c "import horovod.torch"
```

.github/workflows/release-docker.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Publish Latest to Docker
         uses: docker/build-push-action@v1.1.0
         # only on releases and latest Python and PyTorch
-        if: matrix.python_version == "3.9" && matrix.pytorch_version == "1.10"
+        if: matrix.python_version == '3.9' && matrix.pytorch_version == '1.10'
```
carmocca, line 42: Using this PR to fix this error:

https://github.com/PyTorchLightning/pytorch-lightning/actions/runs/1829634458/workflow#L42

.github/BECOMING_A_CORE_CONTRIBUTOR.md
```diff
@@ -38,7 +38,7 @@ Here, we describe general expectations from core contributors:
 
 ### Pull Requests (PRs)
 
-- Pull requests are the evolutionary mechanism of Lightning, so quality is extremely important. Make sure contributors adhere to the guidelines described in the [contributing section](CONTRIBUTING.md#Pull-request).
+- Pull requests are the evolutionary mechanism of Lightning, so quality is extremely important. Make sure contributors adhere to the guidelines described in the [contributing section](CONTRIBUTING.md#Pull-Request).
```
carmocca, line 41: Thought this would fix the docs failures but didn't :(

I don't know what causes them

akihironitta: @carmocca For the docs failures, #11894 should temporarily fix the error (warning).

edenlightning (approved):

requirements/check-avail-strategies.py
```diff
@@ -0,0 +1,7 @@
+import bagua  # noqa: F401
+import deepspeed  # noqa: F401
+import fairscale  # noqa: F401
+import horovod.torch
+
+# returns an error code
+assert horovod.torch.nccl_built()
```
carmocca, line 7: This assertion is failing and blocking master:

```python
Extension horovod.torch has not been built: /usr/local/lib/python3.7/dist-packages/horovod/torch/mpi_lib/_mpi_lib.cpython-37m-x86_64-linux-gnu.so not found
If this is not expected, reinstall Horovod with HOROVOD_WITH_PYTORCH=1 to debug the build error.
Warning! MPI libs are missing, but python applications are still available.
  File "requirements/check-avail-strategies.py", line 7, in <module>
    assert horovod.torch.nccl_built()
AttributeError: module 'horovod.torch' has no attribute 'nccl_built'
```

Borda: ok, so it seems it is working as expected :)
what kind of jobs is failing?

PR #11854, opened 2022-02-10
Flatten inheritance for FSDP strategy

edward-io:
## What does this PR do?

DDPFullyShardedStrategy now inherits from ParallelStrategy instead of DDPStrategy.

Part of #[11863](https://github.com/PyTorchLightning/pytorch-lightning/issues/11863)

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/strategies/fully_sharded.py
+++ b/pytorch_lightning/strategies/fully_sharded.py
@@ -132,8 +132,14 @@ class DDPFullyShardedStrategy(ParallelStrategy):
         self.min_num_params = min_num_params
         self.state_dict_device = torch.device("cpu") if state_dict_to_cpu else None
         self._process_group = None
+        self._pids: Optional[List[int]] = None
+        self._sync_dir: Optional[str] = None
         self._rank_0_will_call_children_scripts: bool = False
         self.set_world_ranks()
+
+    @property
+    def is_distributed(self) -> bool:
+        return True
 
     @property
     def process_group(self):
@@ -168,8 +174,6 @@ class DDPFullyShardedStrategy(ParallelStrategy):
         log.detail(f"{self.__class__.__name__}: setting up distributed...")
         reset_seed()
 
-        if self._layer_sync:
-            self.model = self._layer_sync.apply(self.model)
         # determine which process we are and world size
         self.set_world_ranks()
 
--- a/pytorch_lightning/strategies/fully_sharded.py
+++ b/pytorch_lightning/strategies/fully_sharded.py
@@ -13,19 +13,31 @@
 # limitations under the License.
 import contextlib
 import logging
-from typing import Dict, Generator, List, Optional
+import os
+import shutil
+import signal
+import time
+from pathlib import Path
+from typing import Any, Dict, Generator, List, Optional, Union
 
 import torch
+import torch.distributed
 
 import pytorch_lightning as pl
 from pytorch_lightning.plugins.environments.cluster_environment import ClusterEnvironment
 from pytorch_lightning.plugins.io.checkpoint_plugin import CheckpointIO
 from pytorch_lightning.plugins.precision import PrecisionPlugin
-from pytorch_lightning.strategies.ddp import DDPStrategy
-from pytorch_lightning.utilities import _FAIRSCALE_FULLY_SHARDED_AVAILABLE
+from pytorch_lightning.strategies.launchers.subprocess_script import _SubprocessScriptLauncher
+from pytorch_lightning.strategies.parallel import ParallelStrategy
+from pytorch_lightning.utilities import _FAIRSCALE_FULLY_SHARDED_AVAILABLE, _TORCH_GREATER_EQUAL_1_8
+from pytorch_lightning.utilities.distributed import distributed_available
+from pytorch_lightning.utilities.distributed import group as _group
+from pytorch_lightning.utilities.distributed import init_dist_connection, ReduceOp, sync_ddp_if_available
 from pytorch_lightning.utilities.enums import PrecisionType
-from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.exceptions import DeadlockDetectedException, MisconfigurationException
 from pytorch_lightning.utilities.optimizer import optimizers_to_device
+from pytorch_lightning.utilities.rank_zero import rank_zero_only, rank_zero_warn
+from pytorch_lightning.utilities.seed import reset_seed
 from pytorch_lightning.utilities.types import STEP_OUTPUT
 
 if _FAIRSCALE_FULLY_SHARDED_AVAILABLE:
@@ -35,7 +47,7 @@ if _FAIRSCALE_FULLY_SHARDED_AVAILABLE:
 log = logging.getLogger(__name__)
 
 
-class DDPFullyShardedStrategy(DDPStrategy):
+class DDPFullyShardedStrategy(ParallelStrategy):
 
     strategy_name = "ddp_fully_sharded"
 
@@ -109,6 +121,7 @@ class DDPFullyShardedStrategy(DDPStrategy):
             checkpoint_io=checkpoint_io,
             precision_plugin=precision_plugin,
         )
+        self._num_nodes = 1
         self.cpu_offload = cpu_offload
         self.move_grads_to_cpu = move_grads_to_cpu
         self.flatten_parameters = flatten_parameters
@@ -119,6 +132,8 @@ class DDPFullyShardedStrategy(DDPStrategy):
         self.min_num_params = min_num_params
         self.state_dict_device = torch.device("cpu") if state_dict_to_cpu else None
         self._process_group = None
+        self._rank_0_will_call_children_scripts: bool = False
+        self.set_world_ranks()
 
     @property
     def process_group(self):
@@ -126,25 +141,44 @@ class DDPFullyShardedStrategy(DDPStrategy):
             self._process_group = torch.distributed.new_group()
         return self._process_group
 
-    def setup_distributed(self) -> None:
+    @property
+    def root_device(self) -> torch.device:
+        return self.parallel_devices[self.local_rank]
+
+    @property
+    def num_nodes(self) -> int:
+        return self._num_nodes
+
+    @num_nodes.setter
+    def num_nodes(self, num_nodes: int) -> None:
+        # note that world ranks is related to num_nodes, when resetting it, need to reset world ranks
+        self._num_nodes = num_nodes
+        self.set_world_ranks()
+
+    @property
+    def num_processes(self) -> int:
+        return len(self.parallel_devices) if self.parallel_devices is not None else 0
+
+    def setup_environment(self) -> None:
         if not self.root_device.type == "cuda":
             raise MisconfigurationException(
                 "You selected strategy to be `ddp_fully_sharded`, but GPU is not available."
             )
-        super().setup_distributed()
 
-    def setup(self, trainer: "pl.Trainer") -> None:
-        self.accelerator.setup(trainer)
-        self.setup_optimizers(trainer)
-        self.setup_precision_plugin()
-        optimizers_to_device(self.optimizers, self.root_device)
+        log.detail(f"{self.__class__.__name__}: setting up distributed...")
+        reset_seed()
 
         if self._layer_sync:
             self.model = self._layer_sync.apply(self.model)
+        # determine which process we are and world size
+        self.set_world_ranks()
 
-        self.configure_ddp()
-        self.barrier()
-        self.setup_optimizers(trainer)
+        # set warning rank
+        rank_zero_only.rank = self.global_rank
+
+        init_dist_connection(self.cluster_environment, self.torch_distributed_backend)
+
+        super().setup_environment()
 
     @contextlib.contextmanager
     def model_sharded_context(self) -> Generator:
@@ -172,8 +206,16 @@ class DDPFullyShardedStrategy(DDPStrategy):
 
         log.detail(f"{self.__class__.__name__}: exiting model_sharded_context.")
 
-    def configure_ddp(self) -> None:
-        log.detail(f"{self.__class__.__name__}: configuring DDP... (cpu_offload: [{self.cpu_offload}])")
+    def setup(self, trainer: "pl.Trainer") -> None:
+        self.accelerator.setup(trainer)
+        self.setup_optimizers(trainer)
+        self.setup_precision_plugin()
+        optimizers_to_device(self.optimizers, self.root_device)
+
+        if self._layer_sync:
+            self.model = self._layer_sync.apply(self.model)
+
+        log.detail(f"{self.__class__.__name__}: configuring FSDP... (cpu_offload: [{self.cpu_offload}])")
         if not self.cpu_offload:
             # When using CPU Offload, FSDP will manage the CUDA movement for us.
             # Note: this would be problematic for large model (which could not fit in one GPU)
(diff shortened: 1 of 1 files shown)


ananthsub:
@edward-io does this depend on https://github.com/PyTorchLightning/pytorch-lightning/pull/11643 ? Or should we merge this, and then use #11643 to remove the duplication?

fyi @rohitgr7 @awaelchli

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -118,19 +136,66 @@ def __init__(
...
+        return self._num_nodes
+
+    @num_nodes.setter
+    def num_nodes(self, num_nodes: int) -> None:
+        # note that world ranks is related to num_nodes, when resetting it, need to reset world ranks
+        self._num_nodes = num_nodes
+        self.set_world_ranks()
+
+    @property
+    def num_processes(self):
```
ananthsub, line 162:
```suggestion
    def num_processes(self) -> int:
```

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -145,6 +210,76 @@ def setup(self, trainer: "pl.Trainer") -> None:
         self.barrier()
         self.setup_optimizers(trainer)
 
+    def _check_can_spawn_children(self):
```
ananthsub, line 213:
```suggestion
    def _check_can_spawn_children(self) -> None:
```

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -145,6 +210,76 @@ def setup(self, trainer: "pl.Trainer") -> None:
...
+        if self.local_rank != 0:
+            raise RuntimeError(
+                "Lightning attempted to launch new distributed processes with `local_rank > 0`. This should not happen."
+                " Possible reasons: 1) LOCAL_RANK environment variable was incorrectly modified by the user,"
+                " 2) `ClusterEnvironment.creates_processes_externally` incorrectly implemented."
+            )
+
+    def _call_children_scripts(self):
```
ananthsub, line 221:
```suggestion
    def _call_children_scripts(self) -> None:
```

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -171,6 +306,42 @@ def wrap_policy(*args, **kwargs):
 
         log.detail(f"{self.__class__.__name__}: exiting model_sharded_context.")
 
+    def determine_ddp_device_ids(self):
```
ananthsub, line 309:
```suggestion
    def determine_ddp_device_ids(self) -> Optional[int]:
```

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -171,6 +306,42 @@ def wrap_policy(*args, **kwargs):
...
+            torch.distributed.barrier(device_ids=self.determine_ddp_device_ids())
+        else:
+            torch.distributed.barrier()
+
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        obj = [obj]
+        if self.global_rank != src:
+            obj = [None]
+        torch.distributed.broadcast_object_list(obj, src, group=_group.WORLD)
+        return obj[0]
```
ananthsub, line 327: https://github.com/PyTorchLightning/pytorch-lightning/blob/79c4e5de60685dbc895641b0139ffc6180d069aa/pytorch_lightning/strategies/strategy.py#L37

```suggestion
    def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
        obj = [obj]
        if self.global_rank != src:
            obj = [None]
        torch.distributed.broadcast_object_list(obj, src, group=_group.WORLD)
        return obj[0]
```

pytorch_lightning/strategies/fully_sharded.py
four4fish: optional[int] i think

pytorch_lightning/strategies/fully_sharded.py
four4fish: This is not necessarily true
- for pytorch version< 1.8, torch collective using broadcast, which returns Async work handle, if async_op is set to True. None, if not async_op or if not part of the group
- for pytorch version .=1.8 torch collective using broadcast, which returns None. If rank is part of the group, object_list will contain the broadcasted objects from src rank.

I think change obj to TBroadcast will cause issues

edward-io:
> @edward-io does this depend on https://github.com/PyTorchLightning/pytorch-lightning/pull/11643 ? Or should we merge this, and then use https://github.com/PyTorchLightning/pytorch-lightning/pull/11643 to remove the duplication?

I believe we're doing the latter, @four4fish can confirm.

edward-io:
Addressed typing comments.

four4fish:
> @edward-io does this depend on https://github.com/PyTorchLightning/pytorch-lightning/pull/11643 ? Or should we merge this, and then use https://github.com/PyTorchLightning/pytorch-lightning/pull/11643 to remove the duplication?

I remember Adrian suggested merge 11643 first, I think it make sense to merge bigger PRs first too.

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -183,12 +224,106 @@ def configure_ddp(self) -> None:
 
         # setup optimizers after fully sharded has wrapped the lightning module
         self.setup_optimizers(self.lightning_module.trainer)
+        self.barrier()
+        self.setup_optimizers(trainer)
```
four4fish, line 228: Here is duplicated.
Let's keep the previous ones, the setup_optimizer should happen before model_to_device

four4fish: @edward-io is exactly reproducing the logic here. 
@ananthsub @awaelchli  does move_to_device for FSDP happens before set optimizer as well? Was these part of the duplication necessary? 
The comments above mentioned that `FSDP module.to(device) would first summon all parameters`, but we are not using FSDP.apply or module.to. Am I missing something here?

edward-io: > Here is duplicated. Let's keep the previous ones, the setup_optimizer should happen before model_to_device

oops. sorry, this was a bad merge

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -172,8 +205,16 @@ def wrap_policy(*args, **kwargs):
...
-    def configure_ddp(self) -> None:
-        log.detail(f"{self.__class__.__name__}: configuring DDP... (cpu_offload: [{self.cpu_offload}])")
+    def setup(self, trainer: "pl.Trainer") -> None:
+        self.accelerator.setup(trainer)
+        self.setup_optimizers(trainer)
+        self.setup_precision_plugin()
+        optimizers_to_device(self.optimizers, self.root_device)
+
+        if self.sync_batchnorm:
+            self.model = self.configure_sync_batchnorm(self.model)
```
ananthsub, line 215: note: might run into merge conflicts with #11754

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -182,13 +223,106 @@ def configure_ddp(self) -> None:
...
+        # todo (tchaton) Add support for non-shared file-system which will fail.
+        if len(os.listdir(sync_dir)) == (self.world_size // self.num_nodes):
+            return
+
+        for pid in self._pids:
+            if pid != os.getpid():
+                os.kill(pid, signal.SIGKILL)
+        shutil.rmtree(sync_dir)
+        raise DeadlockDetectedException(f"DeadLock detected from rank: {self.global_rank} \n {trace}")
+
```
ananthsub, line 325: you can remove the override for `post_training_step` in L342 now

ananthsub (approved):

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -109,6 +121,8 @@ def __init__(
             checkpoint_io=checkpoint_io,
             precision_plugin=precision_plugin,
         )
+        self._num_nodes = 1
+        self.sync_batchnorm = False
```
awaelchli, line 125: needs rebase onto master (same in other prs) #11754

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -205,9 +340,6 @@ def predict_step(self, *args, **kwargs) -> STEP_OUTPUT:
         with self.precision_plugin.predict_step_context():
             return self.model.predict_step(*args, **kwargs)
 
-    def post_training_step(self):
-        pass
-
```
awaelchli, line 210: pre_backward is missing too. it was implemented in ddp but the regular sharded strategy also ignores it. 
asking for confirmation that this is intentional

```py
    def pre_backward(self, closure_loss: torch.Tensor) -> None:
        """Run before precision plugin executes backward."""
        if not self.lightning_module.automatic_optimization:
            prepare_for_backward(self.model, closure_loss)
```

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -205,9 +340,6 @@ def predict_step(self, *args, **kwargs) -> STEP_OUTPUT:
...
     @classmethod
     def register_strategies(cls, strategy_registry: Dict) -> None:
```
awaelchli, line 344: where is the teardown? I think we need it?

```py

    def teardown(self) -> None:
        log.detail(f"{self.__class__.__name__}: tearing down DDP plugin")
        super().teardown()
        if isinstance(self.model, DistributedDataParallel):
            self.model = self.lightning_module

        if self._layer_sync:
            self.model = self._layer_sync.revert(self.model)

            # GPU teardown
            log.detail(f"{self.__class__.__name__}: moving model to CPU")
            self.lightning_module.cpu()
            # clean up memory
            torch.cuda.empty_cache()

```

awaelchli (changes requested):

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -119,32 +132,53 @@ def __init__(
...
+
+    @property
+    def num_nodes(self) -> int:
+        return self._num_nodes
+
+    @num_nodes.setter
+    def num_nodes(self, num_nodes: int) -> None:
+        # note that world ranks is related to num_nodes, when resetting it, need to reset world ranks
+        self._num_nodes = num_nodes
+        self.set_world_ranks()
```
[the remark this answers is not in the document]
four4fish: I don't think this is needed anywhere. Accl_conn calls set_world_ranks() at the end, when every properties is final. https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L739

Unless user do trainer.strategy.num_nodes=X, the num_nodes shouldn't change after trainer initialized. 
We shouldn't support runtime device change anyway, so I think we don't need self.set_world_ranks()

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -119,32 +132,53 @@ def __init__(
...
+
+    @property
+    def num_processes(self) -> int:
```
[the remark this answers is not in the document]
four4fish: ParallelStrategy has distributed_sampler_kwargs method, I think len(self.parallel_devices) == num_processes*num_nodes?

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -119,32 +132,53 @@ def __init__(
...
-    def setup(self, trainer: "pl.Trainer") -> None:
-        self.accelerator.setup(trainer)
-        self.setup_optimizers(trainer)
-        self.setup_precision_plugin()
-        optimizers_to_device(self.optimizers, self.root_device)
+        log.detail(f"{self.__class__.__name__}: setting up distributed...")
+        reset_seed()
 
         if self._layer_sync:
             self.model = self._layer_sync.apply(self.model)
```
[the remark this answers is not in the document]
four4fish: the sync batch norm was in setup() for both ddp and fsdp
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/strategies/fully_sharded.py#L142-L143

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -119,32 +132,53 @@ def __init__(
         self.min_num_params = min_num_params
         self.state_dict_device = torch.device("cpu") if state_dict_to_cpu else None
         self._process_group = None
+        self._rank_0_will_call_children_scripts: bool = False
```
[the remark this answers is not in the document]
edward-io: > Where did _share_information_to_prevent_deadlock go?

That's right, since its not called in the setup method, its not needed in this file AFAICT.

pytorch_lightning/strategies/fully_sharded.py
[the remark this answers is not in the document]
ananthsub: `_is_single_process_single_device` is useful only for the DDP communication hooks, which don't take effect here. That's part of the motivation of flattening the inheritance is to see what's actually used.

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -172,8 +210,16 @@ def wrap_policy(*args, **kwargs):
...
         log.detail(f"{self.__class__.__name__}: exiting model_sharded_context.")
 
-    def configure_ddp(self) -> None:
-        log.detail(f"{self.__class__.__name__}: configuring DDP... (cpu_offload: [{self.cpu_offload}])")
+    def setup(self, trainer: "pl.Trainer") -> None:
+        self.accelerator.setup(trainer)
+        self.setup_optimizers(trainer)
+        self.setup_precision_plugin()
+        optimizers_to_device(self.optimizers, self.root_device)
+
```
tchaton, line 218: Seems to be missing some logic from DDP there. `_rank_0_will_call_children_scripts`.

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -172,8 +210,16 @@ def wrap_policy(*args, **kwargs):
 
         log.detail(f"{self.__class__.__name__}: exiting model_sharded_context.")
 
-    def configure_ddp(self) -> None:
-        log.detail(f"{self.__class__.__name__}: configuring DDP... (cpu_offload: [{self.cpu_offload}])")
+    def setup(self, trainer: "pl.Trainer") -> None:
+        self.accelerator.setup(trainer)
```
tchaton, line 214: should be super().setup(trainer) instead.

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -172,8 +210,16 @@ def wrap_policy(*args, **kwargs):
...
+        self.setup_optimizers(trainer)
```
tchaton, line 215: Do you know why setup_optimizers is called 2 times?

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -119,32 +132,57 @@ def __init__(
...
+    def num_nodes(self, num_nodes: int) -> None:
+        # note that world ranks is related to num_nodes, when resetting it, need to reset world ranks
+        self._num_nodes = num_nodes
+        self.set_world_ranks()
+
+    @property
+    def num_processes(self) -> int:
+        return len(self.parallel_devices) if self.parallel_devices is not None else 0
+
+    def setup_environment(self) -> None:
```
tchaton, line 168: Let's move this code into `setup_distributed` as DDP. Could maybe be moved to the Parallel class.

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -182,13 +228,106 @@ def configure_ddp(self) -> None:
...
+            return
+        self.cluster_environment.set_global_rank(self.node_rank * self.num_processes + self.local_rank)
+        self.cluster_environment.set_world_size(self.num_nodes * self.num_processes)
+        rank_zero_only.rank = self.cluster_environment.global_rank()
+
+    def barrier(self, *args, **kwargs) -> None:
+        if not distributed_available():
+            return
+        if _TORCH_GREATER_EQUAL_1_8 and torch.distributed.get_backend() == "nccl":
+            if self.root_device.type == "cpu":
```
tchaton, line 255: Why it is different from master? Should be its own PR.

Borda:
@awaelchli, is this still actual regarding the ongoing Lite refactor? :otter:

PR #11860, opened 2022-02-10
Flatten inheritance for DeepSpeed strategy

edward-io:
## What does this PR do?

DeepSpeedStrategy now inherits from ParallelStrategy instead of DDPStrategy.

Fixes #[11863](https://github.com/PyTorchLightning/pytorch-lightning/issues/11863)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/strategies/deepspeed.py
+++ b/pytorch_lightning/strategies/deepspeed.py
@@ -326,6 +326,8 @@ class DeepSpeedStrategy(ParallelStrategy):
         self.remote_device = remote_device
         self.load_full_weights = load_full_weights
         self._num_nodes: int = 1
+        self._pids: Optional[List[int]] = None
+        self._sync_dir: Optional[str] = None
         self._rank_0_will_call_children_scripts: bool = False
 
         # default FP16 parameters.
@@ -956,6 +958,16 @@ class DeepSpeedStrategy(ParallelStrategy):
             offload_optimizer_device="nvme",
         )
 
+    def _should_run_deadlock_detection(self) -> bool:
+        """Determines whether the plugin will perform process reconciliation in case of errors.
+
+        If the environment variable `PL_RECONCILE_PROCESS` is set, run detection regardless of the cluster environment.
+        By default this is disabled. Otherwise, if the cluster environment creates the processes, allow the scheduler /
+        parent process to perform the process termination, external to Lightning.
+        """
+        return os.getenv("PL_RECONCILE_PROCESS", "0") == "1" or self._rank_0_will_call_children_scripts
+
+
     def validation_step(self, *args, **kwargs) -> Optional[STEP_OUTPUT]:
         with self.precision_plugin.val_step_context():
             return self.model(*args, **kwargs)
@@ -967,3 +979,38 @@ class DeepSpeedStrategy(ParallelStrategy):
     def predict_step(self, *args, **kwargs) -> STEP_OUTPUT:
         with self.precision_plugin.predict_step_context():
             return self.model(*args, **kwargs)
+
+    def reconciliate_processes(self, trace: str) -> None:
+        if self.world_size < 2:
+            return
+
+        if not self._should_run_deadlock_detection():
+            return
+
+        sync_dir = self._sync_dir
+
+        if not sync_dir:
+            rank_zero_warn("Error handling mechanism for deadlock detection is uninitialized. Skipping check.")
+            return
+
+        # The cluster may be configured to periodically purge the `/tmp`
+        # directory, in which case `sync_dir` may not exist anymore at this
+        # point. Idempotently create it to ensure its existence.
+        Path(sync_dir).mkdir(parents=True, exist_ok=True)
+
+        # save a file locally.
+        torch.save(True, os.path.join(sync_dir, f"{self.global_rank}.pl"))
+
+        # sleep for a short time
+        time.sleep(3)
+
+        # return if all processes wrote a file in the `sync_dir`.
+        # todo (tchaton) Add support for non-shared file-system which will fail.
+        if len(os.listdir(sync_dir)) == (self.world_size // self.num_nodes):
+            return
+
+        for pid in self._pids:
+            if pid != os.getpid():
+                os.kill(pid, signal.SIGKILL)
+        shutil.rmtree(sync_dir)
+        raise DeadlockDetectedException(f"DeadLock detected from rank: {self.global_rank} \n {trace}")
--- a/pytorch_lightning/strategies/deepspeed.py
+++ b/pytorch_lightning/strategies/deepspeed.py
@@ -22,6 +22,7 @@ from pathlib import Path
 from typing import Any, Dict, Generator, List, Mapping, Optional, Tuple, Union
 
 import torch
+import torch.distributed
 from torch.nn import Module
 from torch.optim import Optimizer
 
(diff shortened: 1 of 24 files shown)


edward-io:
Update typing.

edward-io:
Rebase, reorder methods.

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -334,17 +425,39 @@ def _load_config(self, config):
                 config = json.load(f)
         return config
 
-    def setup_distributed(self):
-        reset_seed()
+    def set_world_ranks(self) -> None:
```
ananthsub, line 428: same comment as elsewhere, we can make this protected

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -354,6 +467,26 @@ def setup(self, trainer: "pl.Trainer") -> None:
...
 
+    def barrier(self, *args, **kwargs) -> None:
+        if not distributed_available():
+            return
+        if _TORCH_GREATER_EQUAL_1_8 and torch.distributed.get_backend() == "nccl":
+            torch.distributed.barrier(device_ids=self.determine_ddp_device_ids())
+        else:
+            torch.distributed.barrier()
+
+    def setup_distributed(self):
```
ananthsub, line 478: can make this protected

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -771,16 +838,32 @@ def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
...
-            )
-        return False
+    def broadcast(self, obj: object, src: int = 0) -> object:
+        obj = [obj]
+        if self.global_rank != src:
+            obj = [None]
+        torch.distributed.broadcast_object_list(obj, src, group=_group.WORLD)
+        return obj[0]
+
+    def model_to_device(self) -> None:
```
ananthsub, line 848: super nit: same as elsewhere, let's move this next to `setup`

ananthsub (approved):

pytorch_lightning/strategies/deepspeed.py
edward-io: @ananthsub i don't think we can because Accelerator Connector calls it in the _lazy_init_strategy.

ananthsub: ah I didn't realize that!

tchaton (commented):
@edward-io @ananthsub  I am still failing to understand the full reasoning behind this initiative.

Inheritance is a strong mechanism to enforce the right abstraction design and code sync between classes and the fact it might break is a great sign. It means the abstraction isn't just right yet and work is left to be done. 

Now, we are back to square one. If I improve the deadlock mechanism, I need to change it in all DDP-based strategies and am likely to forget one. This happened countless times in the past. 
 
I know the long-term goal is to move everything into plugins and mixin to remove code duplication, but bugs are likely to come in the meantime as this is a long refactor.

Lightning is meant to have the most stable DDP out there, any DDP bugs will damage the PyTorch Lightning brand.

I would prefer for us to abstract the duplicated code first (DDP deadlock reconciliation, collectives, launchers, etc...) before doing the flattening. This seems quite a high-risk initiative to me and done in the wrong order, especially right before the v1.6 release.

IMO, I would prefer to de-scope this for 1.6 to avoid any bad surprises.

@carmocca

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -321,6 +337,83 @@ def __init__(
         self.hysteresis = hysteresis
         self.min_loss_scale = min_loss_scale
 
+    @property
+    def num_nodes(self) -> int:
+        return self._num_nodes
+
+    @num_nodes.setter
+    def num_nodes(self, num_nodes: int) -> None:
```
tchaton, line 345: Seems to be missing `is_distributed` property. This means DistributedSampler won't be applied.

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -321,6 +337,83 @@ def __init__(
         self.hysteresis = hysteresis
         self.min_loss_scale = min_loss_scale
 
+    @property
```
tchaton, line 340: Let's be consistent and order property as DDP. This would make debugging and readability simpler.

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -334,7 +427,19 @@ def _load_config(self, config):
...
+        self.cluster_environment.set_global_rank(self.node_rank * self.num_processes + self.local_rank)
+        self.cluster_environment.set_world_size(self.num_nodes * self.num_processes)
+        rank_zero_only.rank = self.cluster_environment.global_rank()
+
+    def _configure_launcher(self) -> None:
+        self._launcher = _SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
+        if not self.cluster_environment.creates_processes_externally:
+            self._rank_0_will_call_children_scripts = True
+
+    def setup_environment(self) -> None:
```
tchaton, line 442: Let's follow DDP structure with `setup_environment` and `setup_distributed` for consistency.

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -492,24 +627,6 @@ def _initialize_deepspeed_train(self, model):
             self.lr_scheduler_configs = [lr_scheduler]
         self.model = model
 
-    @contextlib.contextmanager
-    def model_sharded_context(self) -> Generator[None, None, None]:
```
tchaton, line 496: Why was it moved?

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -354,6 +479,24 @@ def setup(self, trainer: "pl.Trainer") -> None:
...
 
+    def model_to_device(self) -> None:
+        log.detail(f"{self.__class__.__name__}: moving model to device [{self.root_device}]...")
+        self.model.to(self.root_device)
+
+    def barrier(self, *args, **kwargs) -> None:
+        if not distributed_available():
+            return
+
+        if self.root_device.type == "cpu":
```
tchaton, line 490: Why it is different from master? Let's have this change into its own Pr

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -372,10 +515,6 @@ def _set_node_environment_variables(self) -> None:
         os.environ["WORLD_SIZE"] = str(self.world_size)
         os.environ["LOCAL_RANK"] = str(self.local_rank)
 
-    @property
-    def restore_checkpoint_after_setup(self) -> bool:
```
tchaton, line 376: Why was it moved?

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -458,10 +597,6 @@ def _init_optimizers(self) -> Tuple[Optimizer, Optional[LRSchedulerConfig], Opti
             optimizer_frequencies[0] if optimizer_frequencies else None,
         )
 
-    @property
-    def zero_stage_3(self) -> bool:
```
tchaton, line 462: Why was it moved ?

pytorch_lightning/strategies/deepspeed.py
tchaton: Note: Moving code makes it harder for the reviewer to understand what is added and just moved. I would advise keeping the code re-organization to another PR.

pytorch_lightning/strategies/deepspeed.py
```diff
@@ -771,16 +848,28 @@ def load_checkpoint(self, checkpoint_path: _PATH) -> Dict[str, Any]:
             )
         return client_state
 
-    @property
-    def lightning_restore_optimizer(self) -> bool:
```
tchaton, line 775: Same here.

four4fish:
@tchaton The flatten inheritance for Strategy was voted to be one of the top projects in grid&meta roadmap https://www.notion.so/gridai/1f24be0d26e84adaa064ade0c39e50cc?v=4bab7db846714ff98e72b8a9c62b5050 (we don't have permission anymore, you guys still could see this?)

I feel flatten the inheritance is super helpful for stable strategy support and avoiding bugs. Silent overrides or call order (where to call super.xfunction) in the overrides function is always error prone. More important, it's very hard to trace what logic exactly happening for each strategies with this inheritance structure.

For example: 
1. DDP has configure_ddp() function (which supports localSGD) has been inherited by deepspeed and horovod, which could be misleading
2. A while ago, @ananthsub found it’s very not obvious to tell whether we are move_model_to_device first or setup_optimizer first for some strategies, because the inheritance structure and the inconsistent place where each strategy calls their super.xfunction. If all the logic clearly listed in each strategies, we may found [this bug](https://github.com/PyTorchLightning/pytorch-lightning/pull/11886) way earlier

Borda:
@carmocca @awaelchli, how do we this about this one?

PR #7014, opened 2021-04-14
Fixed `num_sanity_val_steps` affecting reproducibility of training data shuffling

awaelchli:
## What does this PR do?

Fixes #6988

**Test fails on master**

**Important note**
This fix will impact reproducibility from one version to the next as the trainer runs a number of sanity validation steps by default.

## Before submitting
- [ ] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #11870, opened 2022-02-11
Improve mechanism to reset the seed after sanity check

### LINKED ISSUE #11840: Reproducibility issue vs vanilla pytorch on multiple training runs
## 🐛 Bug

**Pytorch Lightning (PL):**

The first run of a model in PL gives one set of results. Subsequent runs (new model instance, same data) all give the _same results_ **but** _different to that of the first run_. There is no reseeding in between these runs. If I reseed, then I am able to reproduce the original results.

**Vanilla pytorch (PT):**

With the same model and data reproduced in PT, on the first run I'm able to get the same results as the first run of the PL model. So I know the models are doing the same thing. However on subsequent runs without reseeding in between, I get a _different set of results_ that are **different every time** both with respect to each other and with respect to the results in PL. Again, if I reseed, then I am able to reproduce the original results.

On the basis of these results, I think we can tease out a few distinct (but obviously related) issues:

1. It appears something in that first run in PL is modifying the RNG state unexpectedly which is _not happening on subsequent runs_
2. This modification is different to whatever is done in vanilla PT which appears to modify the RNG state every time we fit the model and in a different way too because the second set of results in PL are different to the second set of results in PT
3. Why is there a difference in behaviour between subsequent runs in lightning vs subsequent runs in pytorch? Specifically why do we always get the same results in pytorch lightning but not in vanilla pytorch
4. I don't think we should need to reseed between runs. If we do, should this be documented?

I'm not sure what is going on here e.g. is this is a bug in PT, PL, my code or just a documentation issue, but my inclination is that PL should follow PT and we should always get the same results as each other.

### To Reproduce

Please see the colab notebook below with the issue reproduced with the Boring model.

https://colab.research.google.com/drive/1rhR6k7bN5cKwiJLHz7szVR2T14Wyxe3B?usp=sharing

### Expected behavior

We should get _exactly_ the same results in PT as we do in PL. If it turns out I'm doing something wrong (which may well be the case!) then this should probably become a documentation ticket rather than a bug in order to more openly highlight the issue to users.

### Environment

* CUDA:
	- GPU:
	- available:         False
	- version:           11.1
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     False
	- pyTorch_version:   1.10.0+cu111
	- pytorch-lightning: 1.5.10
	- tqdm:              4.62.3
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.7.12
	- version:           # 1 SMP Tue Dec 7 09:58:10 PST 2021

### Additional context

For PL, I followed the guidance in the documentation [here](https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#reproducibility).

For PT, I followed the guidance in the documentation [here](https://pytorch.org/docs/stable/notes/randomness.html).

cc @awaelchli @akihironitta

awaelchli:
## What does this PR do?

An improvement to #7014
Fixes #11840

The previous approach was to reset the seed to the global seed value that was previously set with `seed_everything`.
https://github.com/PyTorchLightning/pytorch-lightning/blob/6f22b3623c28028026b3cb8bb534c1ebca9c5ac8/pytorch_lightning/trainer/trainer.py#L1375-L1377

However, this would lead to multiple subsequent fits to produce identical results. The expected behavior however is:

```py

# Scenario 1
seed_everything(1)
trainer.fit()
trainer.fit()  # different results expected, since we didn't explicitly re-seed

# Scenario 2
seed_everything(2)
trainer.fit()
seed_everything(2)
trainer.fit()  # same results as first fit expected, since we explicitly re-seeded

```

This PR enables both scenarios to work as expected. An existing test was modified to enable scenario 2.
In addition, there is a test `test_multiple_lr_find_calls_gives_same_results` for the LR finder that was relying on the previous behavior. To preserve that, we now also properly reset the random state after tuning.

Theoretically yes. Anyone who relied on the previous behavior will get different results now.

## Before submitting

## PR review

## Did you have fun?

cc @borda


pytorch_lightning/utilities/auto_restart.py
```diff
@@ -280,6 +281,26 @@ def set_rng_states(rng_state_dict: Dict[str, Any]) -> None:
     python_set_rng_state((version, tuple(state), gauss))
 
 
+@contextmanager
+def isolate_rng() -> Generator[None, None, None]:
```
awaelchli, line 285: I'm open to other name suggestions of course :) 

The definition is here because also the other functions `collect_rng_states` and `set_rng_states` are defined here. Another option is to move them to the utilities/seed.py file but it's also not 100% fitting the semantics.

tests/utilities/test_auto_restart.py
```diff
@@ -1574,3 +1575,24 @@ def configure_optimizers(self):
     torch.testing.assert_allclose(total_batches, failed_batches + restart_batches)
     assert not torch.equal(total_weight, failed_weight)
     assert torch.equal(total_weight, model.layer.weight)
+
+
+def test_isolate_rng():
+    """Test that the isolate_rng context manager isolates the random state form the outer scope."""
```
awaelchli, line 1581:
```suggestion
    """Test that the isolate_rng context manager isolates the random state from the outer scope."""
```

pytorch_lightning/utilities/auto_restart.py
Borda: How about `isolate_randomize` as you can also run shuffle inside, right?

carmocca: I like `isolate_rng`.

But I would move it to `seed.py`, to me, `auto_restart` is for fault-tolerance support

carmocca (approved):

tchaton (approved):
LGTM !

pytorch_lightning/utilities/seed.py
```diff
@@ -117,3 +120,36 @@ def pl_worker_init_function(worker_id: int, rank: Optional[int] = None) -> None:
...
     stdlib_seed = (stdlib_ss.generate_state(2, dtype=np.uint64).astype(object) * [1 << 64, 1]).sum()
     random.seed(stdlib_seed)
+
+
+def collect_rng_states() -> Dict[str, Any]:
+    """Collect the global random state of :mod:`torch`, :mod:`numpy` and Python."""
+    return {"torch": torch.get_rng_state(), "numpy": np.random.get_state(), "python": python_get_rng_state()}
+
+
+def set_rng_states(rng_state_dict: Dict[str, Any]) -> None:
```
carmocca, line 130: I would make these 2 protected.

rohitgr7 (approved):

Issue #11232, opened 2021-12-23
[RFC] Better support using multiple loggers simultaneously by deprecating `LoggerCollection`


ananthsub:
## Proposed refactor

Deprecate `LoggerCollection`:
https://github.com/PyTorchLightning/pytorch-lightning/blob/a6a28e08d22f59f2468ff1049c67507df7519e7c/pytorch_lightning/loggers/base.py#L370-L464

## Motivation

The logger collection implements the base logger interface and wraps a sequence of loggers. This way, the trainer always behaves as though a single logger is being used. Such a collection is somewhat handy for writing data, but is much less helpful when reading state from loggers, especially when individual loggers have different state/property values. For instance, It's not clear at all what `LoggerCollection.name` *should* be. The concatenation of all its instance loggers names? The first logger's name? This also came up here: https://github.com/PyTorchLightning/pytorch-lightning/issues/10954

This leads to inconsistencies across components. Namely, Lightning offers no such LoggerCollection-equivalent for Callbacks. The trainer treats them as a flat list, and when hooks need to be called, the Trainer iterates over this list, and calls the callbacks' hooks successively. Why do we not do the same for loggers? 

This leads to leaks in the user-facing API. If users pass a list of loggers to the `Trainer`'s `logger` flag, accessing `trainer.logger` returns a single instance of type `LoggerCollection`, which is different than what users specified, which means users have to learn about this abstraction and likely update their code to handle cases where a single logger is used vs multiple.

## Pitch

1. Introduce a `loggers` property on the Trainer which returns a list of the passed in Loggers objects, so users have an option to not use the LoggerCollection
```py
@property
def loggers(self) -> List[LightningLoggerBase]:
```
2. Deprecate `LoggerCollection` in favor of this `loggers` property. Wherever we call the Logger APIs from within the Trainer, iterate over the `loggers` list and call the hooks instead. 

Example:

Current
https://github.com/PyTorchLightning/pytorch-lightning/blob/a6a28e08d22f59f2468ff1049c67507df7519e7c/pytorch_lightning/trainer/trainer.py#L1444-L1445

New
```py
for logger in self.loggers:
    logger.finalize("success")
```

3. After `LoggerCollection` is removed, we could redefine `logger `on the Trainer to be something like this. Or we could deprecate `logger` entirely in favor of `loggers` to avoid any potential misuse/missed expectations around supporting multiple loggers. 

```py
@property
[...]

PR #10976, opened 2021-12-07
Return only unique names/versions for LoggerCollection

twsl:
## What does this PR do?

Fixes #10954

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #11920, opened 2022-02-14
Refactor codebase to use `trainer.loggers` over `trainer.logger` when needed

akashkw:
## What does this PR do?
Refactors codebase to not use `trainer.logger` and `LoggerCollection` wherever multiple loggers may appear. Instead use `trainer.loggers`.

Part of #11232

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -591,6 +592,7 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
                 if isinstance(trainer.logger.version, str)
                 else f"version_{trainer.logger.version}"
             )
+            # TODO: Find out what ckpt_path should be with multiple loggers
```
akashkw, line 595: What do we want `ckpt_path` to be in the case of multiple loggers? Currently `LoggerCollection` runs a concatenation in `logger.name` and `logger.version`, but this might not be the best approach here.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -213,6 +213,7 @@ def get_standard_metrics(trainer: "pl.Trainer", pl_module: "pl.LightningModule")
     if pl_module.truncated_bptt_steps > 0:
         items_dict["split_idx"] = trainer.fit_loop.split_idx
 
+    # TODO: Adapt for trainer.loggers
```
akashkw, line 216: `LoggerCollection` concatenates `logger.version` here, but does it make sense to do that here? I don't think it really makes sense to include `logger.version` at all in the progress bar.

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -213,8 +213,9 @@ def get_standard_metrics(trainer: "pl.Trainer", pl_module: "pl.LightningModule")
     if pl_module.truncated_bptt_steps > 0:
         items_dict["split_idx"] = trainer.fit_loop.split_idx
 
-    if trainer.logger is not None and trainer.logger.version is not None:
-        version = trainer.logger.version
```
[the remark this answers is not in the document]
daniellepintz: Question: why do we even want the logger version on the progress bar?

pytorch_lightning/callbacks/gpu_stats_monitor.py
```diff
@@ -162,8 +162,9 @@ def on_train_batch_start(
             # First log at beginning of second step
             logs["batch_time/inter_step (ms)"] = (time.time() - self._snap_inter_step_time) * 1000
 
-        assert trainer.logger is not None
-        trainer.logger.log_metrics(logs, step=trainer.global_step)
+        assert trainer.loggers
```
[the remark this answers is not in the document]
daniellepintz: True, it's not required, but do we want to keep it for debugging purposes? Rather than silently not logging anything?

pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -155,7 +155,14 @@ def optimizer_step(
...
         if trainer.track_grad_norm == -1:
             return
-        kwargs = {"group_separator": trainer.logger.group_separator} if trainer.logger is not None else {}
+
+        if not trainer.loggers:
+            kwargs = {}
+        elif len(trainer.loggers) > 1:
+            kwargs = {"group_separator": trainer.loggers[0].group_separator}
+        else:
+            kwargs = {"group_separator": "/"}
```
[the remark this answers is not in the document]
daniellepintz: @akashkw I think you meant to flip lines 162 and 164?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,26 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
         """
         if self.dirpath is not None:
             return  # short circuit
-
-        if trainer.logger is not None:
+        if trainer.loggers:
```
daniellepintz, line 582: I think we have the chance to make a decision here. Where do we want to store the checkpoints if there are multiple loggers, and no dirpath is passed to ModelCheckpoint, and no weights_save_path is passed to Trainer?

The options I see are:
1. (Current behavior) save checkpoints to default_root_dir
2. Save checkpoints to first logger in list
3. Save checkpoints to all loggers in list

pytorch_lightning/callbacks/gpu_stats_monitor.py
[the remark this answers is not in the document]
akashkw: I would assume debugging was the original intention, because otherwise the line would be `if trainer.logger is not None` rather than `assert trainer.logger is not None`, in which case it would still makes sense to keep this.

pytorch_lightning/callbacks/lr_monitor.py
```diff
@@ -149,7 +149,7 @@ def _check_no_key(key: str) -> bool:
         self.last_momentum_values = {name + "-momentum": None for name in names_flatten}
 
     def on_train_batch_start(self, trainer: "pl.Trainer", *args: Any, **kwargs: Any) -> None:
-        assert trainer.logger is not None
+        assert trainer.loggers
```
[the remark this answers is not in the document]
akashkw: https://github.com/PyTorchLightning/pytorch-lightning/pull/11920#discussion_r808465119

pytorch_lightning/callbacks/lr_monitor.py
```diff
@@ -158,16 +158,18 @@ def on_train_batch_start(self, trainer: "pl.Trainer", *args: Any, **kwargs: Any)
             latest_stat = self._extract_stats(trainer, interval)
 
             if latest_stat:
-                trainer.logger.log_metrics(latest_stat, step=trainer.global_step)
+                for logger in trainer.loggers:
+                    logger.log_metrics(latest_stat, step=trainer.global_step)
 
     def on_train_epoch_start(self, trainer: "pl.Trainer", *args: Any, **kwargs: Any) -> None:
-        assert trainer.logger is not None
+        assert trainer.loggers
```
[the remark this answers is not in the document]
akashkw: https://github.com/PyTorchLightning/pytorch-lightning/pull/11920#discussion_r808465119

pytorch_lightning/plugins/precision/precision_plugin.py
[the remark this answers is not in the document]
akashkw: > @akashkw I think you meant to flip lines 162 and 164?

Whoops, that is correct. And the `"/"` is just the return value from `LightningLoggerBase.group_separator`, which would be called if `trainer.logger` were a `LoggerCollection`. Since we are deprecating `LoggerCollection` I just copied that return value here.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,26 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
         """
         if self.dirpath is not None:
             return  # short circuit
-
-        if trainer.logger is not None:
```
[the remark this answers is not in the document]
akashkw: The current plan is to discourage use of `trainer.logger` whenever multiple loggers are configured. Right now trainer.logger gives a warning message when there are multiple loggers, but we will change that to a deprecation warning when `LoggerCollection` is deprecated. In that case, we should only be calling `trainer.logger` internally if we know there is only one logger.
This approach is still up for discussion though.

pytorch_lightning/callbacks/gpu_stats_monitor.py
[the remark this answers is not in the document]
carmocca: Any assertions inside the code (not tests) are there just to notify mypy that we don't expect that the value is `None`, never for debugging

pytorch_lightning/callbacks/model_checkpoint.py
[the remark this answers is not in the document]
carmocca: > but we will change that to a deprecation warning when LoggerCollection is deprecated.

From what I understood today, that won't be the case

pytorch_lightning/callbacks/progress/base.py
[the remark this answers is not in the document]
carmocca: I've always dropped the `v_num` in my experiments, but I guess others might like having it.

pytorch_lightning/callbacks/gpu_stats_monitor.py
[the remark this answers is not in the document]
akashkw: > Any assertions inside the code (not tests) are there just to notify mypy that we don't expect that the value is `None`, never for debugging

Oh that's good to know, thank you. There are some other places in the code where mypy was giving me issues so this will be useful.

pytorch_lightning/callbacks/model_checkpoint.py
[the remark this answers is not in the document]
daniellepintz: Could you clarify?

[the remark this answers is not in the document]
daniellepintz: @carmocca could you clarify?

Akash is referring to this:
https://github.com/PyTorchLightning/pytorch-lightning/blob/d61371922bf89a2b810711e67eee47f224be60cc/pytorch_lightning/trainer/trainer.py#L2562-L2566

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,26 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
+                else:
+                    save_dir = trainer.default_root_dir
+
+            if len(trainer.loggers) == 1:
+                version = (
+                    trainer.logger.version
+                    if isinstance(trainer.logger.version, str)
+                    else f"version_{trainer.logger.version}"
+                )
+                # TODO: Find out what ckpt_path should be with multiple loggers
```
daniellepintz, line 598: let's remove this TODO since we decided to keep it in default_root_dir for now

pytorch_lightning/callbacks/device_stats_monitor.py
```diff
@@ -44,7 +44,7 @@ class DeviceStatsMonitor(Callback):
     """
 
     def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
-        if not trainer.logger:
+        if not trainer.loggers:
```
awaelchli, line 47: `trainer.py` L1846 could also be changed to plural for consistency.

pytorch_lightning/callbacks/model_checkpoint.py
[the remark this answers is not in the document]
awaelchli: we should avoid calling loggers to avoid the warning internally, i agree

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,25 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
             else:
-                save_dir = trainer.logger.save_dir or trainer.default_root_dir
-
-            version = (
-                trainer.logger.version
-                if isinstance(trainer.logger.version, str)
-                else f"version_{trainer.logger.version}"
-            )
-            ckpt_path = os.path.join(save_dir, str(trainer.logger.name), version, "checkpoints")
+                if len(trainer.loggers) == 1:
```
awaelchli, line 587: doesn't this connect to the above else so that we would write elif?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,25 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
+                    save_dir = trainer.logger.save_dir or trainer.default_root_dir
+                else:
+                    save_dir = trainer.default_root_dir
```
awaelchli, line 590: As discussed offline, we need to be a little creative here.
Here is an trainer configuration with two loggers:
```py
trainer = Trainer(
        logger=[TensorBoardLogger("here1"), TensorBoardLogger("here2")],
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        max_epochs=2,
    )
```
Running this on master will create a directory structure like this (bug_report is the cwd here):

image


On this branch, the structure will look like this:
image


The main difference here is just that the checkpoints are no longer under a version folder. This is because we want to move away from the LoggerCollection and now we have to make a hard decision where we want to save the checkpoints. We discussed options:

1. Save in default_root dir (current implementation in this PR)
2. Save to the log folder of the first logger
3. Save a copy of the checkpoint to all logger folders

We will create an RFC to discuss these options. To proceed with this PR, I propose two options:

A) Keep it as a pure refactor: We would need to introduce a utility function that produces the same path here that the LoggerCollection did for backward compatibility.
B) Make a breaking change in this PR: Go with the current implementation and just go with default_root_dir.

@daniellepintz @akashkw @carmocca We discussed B) in our meeting but I think I prefer A) for now. What are your thoughts?

pytorch_lightning/callbacks/progress/base.py
```diff
@@ -213,11 +213,13 @@ def get_standard_metrics(trainer: "pl.Trainer", pl_module: "pl.LightningModule")
...
         version = trainer.logger.version
-        if isinstance(version, str):
-            # show last 4 places of long version strings
-            version = version[-4:]
-        items_dict["v_num"] = version
+        if version is not None:
+            if isinstance(version, str):
+                # show last 4 places of long version strings
+                version = version[-4:]
+            items_dict["v_num"] = version
```
awaelchli, line 223: we drop the version number for multiple loggers, which I find very reasonable considering we anyway only showed the ending characters.
not sure if it's worth mentioning in the changlog

pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -155,7 +155,15 @@ def optimizer_step(
     def _track_grad_norm(self, trainer: "pl.Trainer") -> None:
         if trainer.track_grad_norm == -1:
             return
-        kwargs = {"group_separator": trainer.logger.group_separator} if trainer.logger is not None else {}
+
+        if not trainer.loggers:
+            kwargs = {}
+        elif len(trainer.loggers) == 1:
+            assert trainer.logger
```
awaelchli, line 162: nit, you could save the assertion and access loggers[0] instead

awaelchli (approved):

pytorch_lightning/callbacks/model_checkpoint.py
akashkw: I feel like option A makes more sense to me. Changing this behavior should have its own separate discussion with its own PR. We could make a separate utility function but I don't think that will be strictly necessary. Replicating this behavior should only require 2 extra lines to concatenate the names and versions.

daniellepintz: @awaelchli thanks for bringing this up! Thinking about it more, I agree that option A makes the most sense! I will create a separate RFC to discuss making any changes here

pytorch_lightning/callbacks/progress/base.py
daniellepintz: We could consider doing something similar to the suggestion [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/11920#discussion_r810399237) for this. Keep the exact same behavior as a pure refactor, and then create an issue to discuss, which I am planning to do any way for logger v_num.

pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -155,7 +155,14 @@ def optimizer_step(
...
         if trainer.track_grad_norm == -1:
             return
-        kwargs = {"group_separator": trainer.logger.group_separator} if trainer.logger is not None else {}
+
+        if not trainer.loggers:
+            kwargs = {}
+        elif len(trainer.loggers) == 1:
+            kwargs = {"group_separator": trainer.loggers[0].group_separator}
+        else:
+            kwargs = {"group_separator": "/"}
```
carmocca, line 164: grad_norm already has `group_separator="/"` as default

```suggestion
        kwargs = {}
        if len(trainer.loggers) == 1:
            kwargs["group_separator"] = trainer.loggers[0].group_separator
```

carmocca (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -1248,10 +1248,10 @@ def _log_hyperparams(self) -> None:
         # log hyper-parameters
         hparams_initial = None
 
-        if self.logger is not None:
-            # save exp to get started (this is where the first experiment logs are written)
-            datamodule_log_hyperparams = self.datamodule._log_hyperparams if self.datamodule is not None else False
+        # save exp to get started (this is where the first experiment logs are written)
+        datamodule_log_hyperparams = self.datamodule._log_hyperparams if self.datamodule is not None else False
 
+        if self.loggers:
```
daniellepintz, line 1254: why not move this check to the top of the method and return early if there are no loggers?

daniellepintz (approved):
LGTM!

pytorch_lightning/trainer/trainer.py
akashkw: good suggestion!

pl_examples/domain_templates/generative_adversarial_net.py
```diff
@@ -206,7 +206,8 @@ def on_train_epoch_end(self):
         # log sampled images
         sample_imgs = self(z)
         grid = torchvision.utils.make_grid(sample_imgs)
-        self.logger.experiment.add_image("generated_images", grid, self.current_epoch)
+        for logger in self.loggers:
+            self.logger.experiment.add_image("generated_images", grid, self.current_epoch)
```
rohitgr7, line 210:
```suggestion
            logger.experiment.add_image("generated_images", grid, self.current_epoch)
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -1245,41 +1245,44 @@ def _run(
...
+                    "in both the LightningModule's and LightningDataModule's hparams "
+                    "but have different values."
+                )
+            hparams_initial = {**lightning_hparams, **datamodule_hparams}
+        elif self.lightning_module._log_hyperparams:
+            hparams_initial = self.lightning_module.hparams_initial
+        elif datamodule_log_hyperparams:
+            hparams_initial = self.datamodule.hparams_initial
 
+        for logger in self.loggers:
```
rohitgr7, line 1281: quick ques: are we deprecating LoggerCollection?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,28 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
+                if len(trainer.loggers) == 1
+                else "_".join(dict.fromkeys(str(logger.name) for logger in trainer.loggers))
+            )
             version = (
-                trainer.logger.version
-                if isinstance(trainer.logger.version, str)
-                else f"version_{trainer.logger.version}"
+                trainer.loggers[0].version
+                if len(trainer.loggers) == 1
+                else "_".join(dict.fromkeys(str(logger.version) for logger in trainer.loggers))
```
rohitgr7, line 599:
```suggestion
                "_".join(str(logger.version) for logger in trainer.loggers)
```
is this different?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -578,20 +579,28 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
-            else:
+            elif len(trainer.loggers) == 1:
                 save_dir = trainer.logger.save_dir or trainer.default_root_dir
+            else:
+                save_dir = trainer.default_root_dir
 
+            name = (
+                trainer.loggers[0].name
+                if len(trainer.loggers) == 1
+                else "_".join(dict.fromkeys(str(logger.name) for logger in trainer.loggers))
```
rohitgr7, line 594:
```suggestion
                "_".join(str(logger.name) for logger in trainer.loggers)
```
can you add a test for both?

pytorch_lightning/trainer/trainer.py
carmocca: Yes: #11232

pytorch_lightning/callbacks/model_checkpoint.py
akashkw: `dict.fromkeys` was added to prevent duplicate names and versions. You can see the original PR here https://github.com/PyTorchLightning/pytorch-lightning/pull/10976. Since this is the original behavior in `LoggerCollection`, I think we should leave this as is to keep it a pure refactor.

rohitgr7: how about `set` then? just trying to avoid creating a dict since we are only concerned with keys. works for both single and multiple loggers.

daniellepintz:
Can we put back the revert early if no loggers change, since that doesnt seem like it caused the failures?

pytorch_lightning/callbacks/model_checkpoint.py
akashkw: I tested it out and this actually results in a different behavior. It looks like `dict.fromkeys()` will remove duplicates and preserve the order, while `set` only removes duplicates. This is because dictionaries are ordered as of Python 3.7.

rohitgr7: Oh! Yeah that's true. Totally forgot about it.

akashkw: Yeah this behavior is not very transparent so I added a comment to explain it

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -572,20 +574,20 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
+            else:
+                save_dir = trainer.default_root_dir
 
-            version = (
-                trainer.logger.version
-                if isinstance(trainer.logger.version, str)
-                else f"version_{trainer.logger.version}"
-            )
-            ckpt_path = os.path.join(save_dir, str(trainer.logger.name), version, "checkpoints")
+            name = _name(trainer.loggers)
```
daniellepintz, line 586: Beautiful!

tests/utilities/test_logger.py
```diff
@@ -172,3 +175,37 @@ def test_add_prefix():
...
+    name = _name([logger3])
+    assert name == "foo"
+    name = _name(loggers)
+    assert name == "foo_bar_baz"
+    name = _name(loggers, "-")
+    assert name == "foo-bar-baz"
+
+
+def test_version(tmpdir):
+    """Verify names of loggers are concatenated properly."""
```
rohitgr7, line 198:
```suggestion
    """Verify versions of loggers are concatenated properly."""
```
this now looks like the most trivial unit test.. haha

rohitgr7 (approved):

ananthsub (approved):

tests/utilities/test_logger.py
akashkw: Trivial... but it does make you think about the edge cases :)

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1273,7 +1273,7 @@ def test_none_monitor_saves_correct_best_model_path(tmpdir):
 def test_last_global_step_saved():
     # this should not save anything
     model_checkpoint = ModelCheckpoint(save_top_k=0, save_last=False, monitor="foo")
-    trainer = Mock()
+    trainer = MagicMock()
```
akashkw, line 1276: I believe this shouldn't be an issue. This test was failing because `trainer.loggers` didn't have `__iter__` implemented as a `Mock`. Please let me know I should be doing something else.

PR #11935, opened 2022-02-15
Add accelerator method `teardown()`

### LINKED ISSUE #11091: Strategies and Accelerator teardown logic clean up
## Proposed refactor

After #11022 Move accelerator to strategies, the teardown logic can be simplified and cleaned up a bit. 

### Motivation

Reduce redundant code and make code more clean

### Pitch

In ddp, ddp_spawning and hovorod plugins, teardown() all has device specific logic like:
https://github.com/four4fish/pytorch-lightning/blob/move-accelerator/pytorch_lightning/plugins/training_type/ddp.py#L507-L511

with accelerator moved into strategy, device related teardown should move to X_accelerator
```
class GPUAccelerator()
        def teardown(self):
               # self.lightning_module.cpu()  suggested by @ananthsub module movement should be handled by strategies
               torch.cuda.empty_cache()
```

Some teardown() logic are shared between different strategy and same device type, like single_tpu and tpu_spawn has the same teardown() logic:
https://github.com/four4fish/pytorch-lightning/blob/move-accelerator/pytorch_lightning/plugins/training_type/single_tpu.py#L90-L92

we can move it to TPUAccelerator
```
class TPUAccelerator()
          def teardown(self) -> None:
```

### Additional context

edward-io:
## What does this PR do?

Addresses #11091

## Before submitting

## PR review

## Did you have fun?


ananthsub (approved):

ananthsub (commented):
Please also add a unit test for each of the accelerator teardown implementations

edward-io:
add tests - it doesn't seem like there's a good way to verify that cuda cache is emptied, so I'm only checking that teardown is called.

tests/accelerators/test_ddp.py
```diff
@@ -145,3 +146,19 @@ def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule")
         callbacks=CustomCallback(),
     )
     trainer.fit(model)
+
+
+@RunIf(min_gpus=2, min_torch="1.8.1", standalone=True)
+def test_ddp_wrapper(tmpdir):
+    model = BoringModel()
+    strategy = DDPStrategy()
+    with patch.object(DDPStrategy, "teardown", wraps=strategy.teardown) as mock:
```
carmocca, line 155: This context manager should wrap just the `trainer.fit` call

tests/accelerators/test_ddp_spawn.py
```diff
@@ -81,3 +84,19 @@ def test_ddp_all_dataloaders_passed_to_fit(tmpdir):
     )
     trainer.fit(model, train_dataloaders=model.train_dataloader(), val_dataloaders=model.val_dataloader())
     assert trainer.state.finished, "DDP doesn't work with dataloaders passed to fit()."
+
+
+@RunIf(min_gpus=2, standalone=True)
+def test_ddp_wrapper(tmpdir):
+    model = BoringModel()
+    strategy = DDPSpawnStrategy()
+    with patch.object(DDPSpawnStrategy, "teardown", wraps=strategy.teardown) as mock:
```
carmocca, line 93: Same comment

tests/accelerators/test_ddp.py
```diff
@@ -145,3 +146,19 @@ def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule")
...
+        trainer = Trainer(
+            default_root_dir=tmpdir,
+            fast_dev_run=True,
+            strategy="ddp",
+            gpus=2,
```
carmocca, line 160: This test does not need 2 gpus

tests/accelerators/test_tpu.py
```diff
@@ -332,3 +333,15 @@ def test_devices_auto_choice_tpu():
     trainer = Trainer(accelerator="auto", devices="auto")
     assert trainer.devices == 8
     assert trainer.tpu_cores == 8
+
+
+@RunIf(tpu=True)
+def test_teardown(tmpdir):
+    model = BoringModel()
+    strategy = TPUSpawnStrategy()
+    with patch.object(TPUSpawnStrategy, "teardown", wraps=strategy.teardown) as mock:
```
carmocca, line 342: Same

tests/accelerators/test_ddp_spawn.py
```diff
@@ -81,3 +84,19 @@ def test_ddp_all_dataloaders_passed_to_fit(tmpdir):
...
+        trainer = Trainer(
+            default_root_dir=tmpdir,
+            fast_dev_run=True,
+            strategy="ddp",
+            gpus=2,
```
carmocca, line 98: Same

tests/accelerators/test_tpu.py
```diff
@@ -332,3 +333,15 @@ def test_devices_auto_choice_tpu():
...
+        trainer = Trainer(strategy=strategy, max_epochs=1, tpu_cores=8, default_root_dir=tmpdir, fast_dev_run=True)
```
carmocca, line 343: redundant with fast_dev_run
```suggestion
        trainer = Trainer(strategy=strategy, tpu_cores=8, default_root_dir=tmpdir, fast_dev_run=True)
```

tests/accelerators/test_ddp.py
```diff
@@ -145,3 +146,19 @@ def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule")
         callbacks=CustomCallback(),
     )
     trainer.fit(model)
+
+
+@RunIf(min_gpus=2, min_torch="1.8.1", standalone=True)
```
carmocca, line 151: This should work in 1.7 too

```suggestion
@RunIf(min_gpus=2, standalone=True)
```

pytorch_lightning/strategies/single_tpu.py
```diff
@@ -83,5 +83,4 @@ def register_strategies(cls, strategy_registry: Dict) -> None:
 
     def teardown(self) -> None:
         super().teardown()
-        # TPU teardown
-        os.environ.pop("PT_XLA_DEBUG", None)
+        self.accelerator.teardown()
```
four4fish, line 86: I think we can remove this override? The TPU tear down logic is in accelerator, single device and tpu single device shares the same teardown logic now

four4fish (approved):

rohitgr7 (approved):

tests/strategies/test_ddp_spawn.py
```diff
@@ -81,3 +84,19 @@ def test_ddp_all_dataloaders_passed_to_fit(tmpdir):
     )
     trainer.fit(model, train_dataloaders=model.train_dataloader(), val_dataloaders=model.val_dataloader())
     assert trainer.state.finished, "DDP doesn't work with dataloaders passed to fit()."
+
+
+@RunIf(min_gpus=1, standalone=True)
+def test_ddp_wrapper(tmpdir):
```
akihironitta, line 90:
```suggestion
def test_ddp_spawn_teardown(tmpdir):
```

tests/strategies/test_ddp.py
```diff
@@ -145,3 +146,19 @@ def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule")
         callbacks=CustomCallback(),
     )
     trainer.fit(model)
+
+
+@RunIf(min_gpus=1, standalone=True)
+def test_ddp_wrapper_teardown(tmpdir):
```
akihironitta, line 152:
```suggestion
def test_ddp_teardown(tmpdir):
```

akihironitta (approved):
nit

ananthsub (approved):

Borda (approved):

pytorch_lightning/accelerators/tpu.py
```diff
@@ -43,6 +44,9 @@ def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
         }
         return device_stats
 
+    def teardown(self) -> None:
+        os.environ.pop("PT_XLA_DEBUG", None)
```
awaelchli, line 48: a teardown should not remove environment variables that it doesn't add itself.
This is done in the strategy, let's keep it there?

tests/strategies/test_ddp.py
```diff
@@ -145,3 +146,19 @@ def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule")
         callbacks=CustomCallback(),
     )
     trainer.fit(model)
+
+
+@RunIf(min_gpus=1, standalone=True)
+def test_ddp_teardown(tmpdir):
```
awaelchli, line 152: Thanks for thinking of a test! But in this case, let's do a unit test instead. 

We can't afford doing a full training run to check each individual hook call separately. We need to move away from this pattern. 

IMO there are two categories of tests

1) check that hooks get called and called in the right order
2) check that the implementation of a hook does what it should

To address 1) we don't need one test for each single hook. Instead, we can do what we have in `tests/models/test_hooks.py`
To address 2) we can instantiate the class and call the hook directly and assert the side-effects. [Random example](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/plugins/precision/test_tpu_bf16_plugin.py)

awaelchli (changes requested):

edward-io:
Removed TPU teardown, replaced test with a specific one for GPU teardown.

pytorch_lightning/strategies/ddp.py
```diff
@@ -429,5 +429,5 @@ def teardown(self) -> None:
             # GPU teardown
             log.detail(f"{self.__class__.__name__}: moving model to CPU")
             self.lightning_module.cpu()
-            # clean up memory
-            torch.cuda.empty_cache()
+
+        self.accelerator.teardown()
```
kaushikb11, line 433: Should we add `self.accelerator.teardown()` to base class Strategy's `teardown` method as well?

pytorch_lightning/strategies/dp.py
```diff
@@ -161,5 +161,5 @@ def teardown(self) -> None:
         if self.root_device.type == "cuda":
             # GPU teardown
             self.lightning_module.cpu()
```
carmocca, line 163: Do you plan to move this piece of code in a separate PR?

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -55,6 +55,9 @@ def get_device_stats(self, device: Union[str, torch.device]) -> Dict[str, Any]:
         """
         raise NotImplementedError
 
+    def teardown(self) -> None:
```
carmocca, line 58: Remember to update the CHANGELOG with this addition

carmocca (approved):

pytorch_lightning/strategies/ddp.py
four4fish: I think this PR is reproducing the logic. Add self.accelerator.teardown() could change teardown logic for some strategies. Maybe let's try that out in a separate PR?

four4fish (approved):

pytorch_lightning/strategies/ddp.py
awaelchli: > Should we add self.accelerator.teardown() to base class Strategy's teardown method as well?

Adding it to the base makes sense to me. However, we need to be aware of the `super().teardown()` call. 
We have the lines  `self.lightning_module.cpu()` so the call to empty cuda cache should run after that.

I'm good with the current version. If you change the PR to what @kaushikb11 suggested, please re-request review. Otherwise LGTM

tests/accelerators/test_gpu.py
```diff
@@ -72,3 +72,15 @@ def test_gpu_availability():
 def test_warning_if_gpus_not_used():
     with pytest.warns(UserWarning, match="GPU available but not used. Set `accelerator` and `devices`"):
         Trainer()
+
+
+@RunIf(min_gpus=1)
+def test_gpu_teardown():
+    acc = GPUAccelerator()
+    tensor = torch.zeros(100).cuda()
+    del tensor
```
awaelchli, line 81: are we counting on garbage collection being done at the appropriate time here?
It might be safer to add a `gc.collect()` call?

awaelchli (approved):

pytorch_lightning/strategies/ddp.py
tchaton: Let's keep the PR atomic. One PR for replacing the code, one PR to move this in the base class.

pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -287,6 +287,7 @@ def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_gra
     def teardown(self) -> None:
         super().teardown()
         os.environ.pop("PT_XLA_DEBUG", None)
```
tchaton, line 289: Should this be moved to the TPU teardown accelerator @kaushikb11

tchaton (approved):
LGTM ! Small comments

pytorch_lightning/strategies/strategy.py
```diff
@@ -442,7 +445,12 @@ def teardown(self) -> None:
         It is the right place to release memory and free other resources.
         """
         optimizers_to_device(self.optimizers, torch.device("cpu"))
+
+        if self.lightning_module is not None:
+            log.detail(f"{self.__class__.__name__}: moving model to CPU")
+            self.lightning_module.cpu()
```
carmocca, line 451: Why not have this be part of the GPUAccelerator teardown?

This might be limiting for other accelerators (but that's an issue in current master too)

carmocca (approved):

pytorch_lightning/strategies/strategy.py
justusschock: AFAIK our design is to move models back to cpu in general after we are done, no? When we are supposed to release all device memory claimed, there is no other way then to move the model to cpu imo. And this should be handled by the strategy and not the accelerator since the strategy is owning the model :)

PR #11886, opened 2022-02-11
Move model to device before creating optimizers in DDP

ananthsub:
## What does this PR do?

This came from investigating test failures in https://github.com/PyTorchLightning/pytorch-lightning/pull/11867/
The ZeroRedundancyOptimizer needs to have CUDA tensors. Upon further inspection, the optimizers were created before the models were moved to the device because of the `super()` call happening at the start of the method.

Now the flow matches the typical DistrbutedDataParallel initialization flow: https://pytorch.org/tutorials/intermediate/ddp_tutorial.html?highlight=ddp
- initialize the base module
- move the module to the device
- wrap the base module in DistributedDataParallel
- create the optimizers based on the module's parameters 

### Does your PR introduce any breaking changes? If yes, please list them.
None

## Before submitting

- [n/a] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [n/a] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @justusschock @kaushikb11 @awaelchli @akihironitta @rohitgr7

Issue #11863, opened 2022-02-10
Flatten the Strategy inheritance


four4fish:
## Proposed refactor

Flatten the Strategy inheritance:
Part of #10416

### Motivation

Reduce coupling between strategies, reduce unintentional overrides/inheritance and avoid silent failures
eg: 
1. DeepSpeed inherited configure_ddp() from DDP, which is unnecessary and error-prone.
2. DDP and FSDP have totally different distributed behavior, but with current inheritance, adding a DDP strategy method will automatically enable this method for FSDP

### Pitch
Tasks list:
- FSDP inherit ParallelStrategy instead of DDP
- DeepSpeed inherit ParallelStrategy instead of DDP
- [RFC] DDPShard inherit ParallelStrategy instead of DDP
- [RFC] DDPShardSpawn inherit ParallelStrategy instead of DDPSpawn 
- [RFC] TPUSpawn inherit ParallelStrategy instead of DDPSpawn

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

awaelchli:
Let's first get process launchers #11643 done so the ddp launch logic does not get duplicated into every strategy? 
cc @rohitgr7

PR #11952, opened 2022-02-17
Fix initialization of optimizers in DDP Strategy

ananthsub:
## What does this PR do?
Redo of #11886
also fixes: #12317

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/strategies/sharded.py
```diff
@@ -40,6 +40,23 @@ class DDPShardedStrategy(DDPStrategy):
...
+        # move the model to the correct device
+        self.model_to_device()
+
+        if self.sync_batchnorm:
+            self.model = self.configure_sync_batchnorm(self.model)
+
+        # skip wrapping the model if we are not fitting as no gradients need to be exchanged
+        trainer_fn = trainer.state.fn
+        if trainer_fn == TrainerFn.FITTING:
+            self.configure_ddp()
```
ananthsub, line 58: Sharded inheriting from DDP makes this unnecessarily complicated considering they have different behaviors. This is why issues like https://github.com/PyTorchLightning/pytorch-lightning/issues/11863 are important

carmocca (approved):

pytorch_lightning/strategies/sharded.py
awaelchli: We are missing a call to super().setup() here. Which means we won't setup optimizers there which is fine I guess because we do that in configure_ddp for this strategy. 

Just want to make sure you are aware that the missing super.setup()  means we won't call `accelerator.setup()` and `self.setup_precision_plugin() `

awaelchli (approved):

pytorch_lightning/strategies/sharded.py
ananthsub: @awaelchli - I am holding off on this until https://github.com/PyTorchLightning/pytorch-lightning/pull/11985 is reviewed & merged. then i'll redo the optimizer initialization, precision plugin setup, and accelerator setup

Borda (approved):

pytorch_lightning/strategies/sharded.py
ananthsub: I cannot call `super().setup()` because that will all DDPStrategy.setup() which does something completely different: it wraps the model in DistributedDataParallel, not ShardedDataParallel!

ananthsub: @awaelchli I don't even think we can rely on `super().setup()` inside of DDP since we should call the accelerator.setup(trainer) *before* we call `model_to_device` to handle any accelerator validation. 
https://github.com/PyTorchLightning/pytorch-lightning/blob/b5fe056765ce71b237c641a1b159c346e5e8bd33/pytorch_lightning/accelerators/gpu.py#L46-L50

at this point i really dislike how `setup` is packaged on the base strategy. i'd much prefer calling the individual methods at different points rather than try to dance around them.

pytorch_lightning/strategies/ddp.py
```diff
@@ -152,6 +154,17 @@ def setup(self, trainer: "pl.Trainer") -> None:
         if trainer_fn == TrainerFn.FITTING:
             self.configure_ddp()
 
+        # set up optimizers after the wrapped module has been moved to the device
+        self.setup_optimizers(trainer)
+        self.setup_precision_plugin()
+        optimizers_to_device(self.optimizers, self.root_device)
```
ananthsub: `setup_optimizers` returns early if the trainer isn't fitting or tuning.
but `setup_precision_plugin` looks up the model, optimizer, and lr scheduler configs.

are all precision plugins built to handle optimizers/LR schedulers not being available outside of fitting/tuning? 

https://github.com/PyTorchLightning/pytorch-lightning/blob/b5fe056765ce71b237c641a1b159c346e5e8bd33/pytorch_lightning/strategies/strategy.py#L146-L152

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -135,16 +135,15 @@ def setup_distributed(self) -> None:
 
     def setup(self, trainer: "pl.Trainer") -> None:
         self.accelerator.setup(trainer)
-        self.setup_optimizers(trainer)
-        self.setup_precision_plugin()
-        optimizers_to_device(self.optimizers, self.root_device)
```
ananthsub, line 138: we were calling `configure_optimizers` 3 times in this one method 😞 
i can split these out as separate PRs if desired

pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -119,10 +121,15 @@ def setup(self, trainer: "pl.Trainer") -> None:
...
 
         # skip wrapping the model if we are not fitting as no gradients need to be exchanged
-        trainer_fn = self.lightning_module.trainer.state.fn
+        trainer_fn = trainer.state.fn
         if trainer_fn == TrainerFn.FITTING:
             self.configure_ddp()
 
+        # set up optimizers after the wrapped module has been moved to the device
+        self.setup_optimizers(trainer)
+        optimizers_to_device(self.optimizers, self.root_device)
```
ananthsub, line 130: wondering if `self.setup_optimizers` should include `optimizers_to_device` by default, so users only ever need to call `setup_optimizers`

pytorch_lightning/strategies/ddp.py
```diff
@@ -135,12 +136,13 @@ def setup_environment(self) -> None:
         super().setup_environment()
 
     def setup(self, trainer: "pl.Trainer") -> None:
-        super().setup(trainer)
         # share ddp pids to all processes
         self._rank_0_will_call_children_scripts = self.broadcast(self._rank_0_will_call_children_scripts)
         if self._should_run_deadlock_detection():
             self._share_information_to_prevent_deadlock()
 
+        self.accelerator.setup(trainer)
```
tchaton, line 144: Why this change? This doesn't seem related to this PR.

pytorch_lightning/strategies/ddp.py
```diff
@@ -152,6 +154,17 @@ def setup(self, trainer: "pl.Trainer") -> None:
         if trainer_fn == TrainerFn.FITTING:
             self.configure_ddp()
 
+        # set up optimizers after the wrapped module has been moved to the device
+        self.setup_optimizers(trainer)
```
tchaton, line 158: Let's try to make the code more readable.

Let's call super().setup(trainer) and override `setup_optimizers` to add the post_localSGD within its own method.

```
super().setup(trainer)

def setup_optimizers(self):
    super().setup_optimizers()
    self._reinit_optimizers_with_post_localSGD() # contain the torch version checks
```

pytorch_lightning/strategies/ddp.py
```diff
@@ -257,6 +264,7 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
             optimizers[x] = post_localSGD_optimizer
             del optimizer
         self.optimizers = optimizers
+        optimizers_to_device(self.optimizers, self.root_device)
```
tchaton, line 267: This shouldn't be needed with the proposed refactor above as this would be called in the parent setup class?

pytorch_lightning/strategies/ddp_spawn.py
tchaton: Is `PostLocalSGDState` not supported for DDPSpawn? Why do we have a feature divergence there ?

tchaton: This is already in the parent setup method. I would prefer to rely as much as possible on the parent methods from the base `Strategy`.

Maybe the base class should be:

```py
        """Setup plugins for the trainer fit and creates optimizers.

        Args:
            trainer: the trainer instance
        """
        self.model_to_device()
        self.configure_model()
        self.setup_precision_plugin()
        optimizers_to_device(self.optimizers, self.root_device)
```

pytorch_lightning/strategies/fully_sharded.py
tchaton: Would be better.

pytorch_lightning/strategies/ddp_spawn.py
tchaton: @awaelchli @carmocca @justusschock

pytorch_lightning/strategies/ddp.py
awaelchli, line 158: by moving the `setup_optimizers()` after the configure_ddp(), it will now affect the BaguaStrategy. 
Bagua runs checks on the optimizer in configuer_ddp:

```py
        if self._bagua_algorithm == "qadam":
            self._check_qadam_optimizer()
```

We would need to move it also after the setup_optimizer() call was made.

ananthsub: great catch @awaelchli ! i forgot that bagua also extending DDP :/

pytorch_lightning/strategies/ddp_spawn.py
ananthsub: as @awaelchli  this doesn't work even now. Bagua expects the optimizers to be initialized before wrapping the model.

pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -140,7 +137,8 @@ def setup(self, trainer: "pl.Trainer") -> None:
             set_shared_parameters(self.model.module, shared_params)
 
         self.setup_optimizers(trainer)
-        self.precision_plugin.connect(self.model, None, None)
```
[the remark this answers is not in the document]
ananthsub: @kaushikb11 could you take a look?

rohitgr7:
hey @ananthsub!
is this blocked on something or ready to go?

rohitgr7:
assigned myself. Will complete this soon.

pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -186,6 +189,10 @@ def configure_ddp(self) -> None:
         self.model = self._setup_model(LightningDistributedModule(self.model))
         self._register_ddp_hooks()
 
+        # set up optimizers after the wrapped module has been moved to the device
+        self.setup_optimizers(self.lightning_module.trainer)
+        optimizers_to_device(self.optimizers, self.root_device)
```
awaelchli, line 194: why not place it in the setup method as with the DDPStrategy? These two will need to collapse anyway in the future.

pytorch_lightning/strategies/sharded.py
```diff
@@ -40,16 +41,41 @@ class DDPShardedStrategy(DDPStrategy):
     strategy_name = "ddp_sharded"
     _REDUCE_BUFFER_SIZE_DEFAULT: int = 2**23  # 8M
 
-    def configure_ddp(self) -> None:
-        trainer = self.lightning_module.trainer
-        if "reduce_buffer_size" not in self._ddp_kwargs:
-            # For multi-node training, enabling bucketing will improve performance.
-            self._ddp_kwargs["reduce_buffer_size"] = self._REDUCE_BUFFER_SIZE_DEFAULT if self.num_nodes > 1 else 0
+    def setup(self, trainer: "pl.Trainer") -> None:
```
awaelchli, line 44: isn't this the same as the DDP implementation? The only difference I see is in `configure_ddp`

rohitgr7: no.. there's a configuration for `PostLocalSGDState` as well.

austinmw:
Any updates on this PR?

carmocca (approved):

Issue #7550, opened 2021-05-14
DDP spawn no longer works in Jupyter environment


Borda:
## 🐛 Bug

Running multi-GPU training does not update training progress bar, used in Grid Jupyterlab

## Please reproduce using the BoringModel

https://github.com/Borda/kaggle_plant-pathology/blob/main/notebooks/Plant-Pathology-with-Lightning.ipynb

### To Reproduce

just change 
```
    gpus=2
    accelerator="ddp_spawn"
    precision=16,
    accumulate_grad_batches=8,
    val_check_interval=0.25,
    progress_bar_refresh_rate=1,
    weights_summary='top',
```
when I check GPU utilization it is running correctly

### Expected behaviour

a properly updated progress bar

### Environment

**Note**: `Bugs with code` are solved faster ! `Colab Notebook` should be made `public` !

cc @tchaton @rohitgr7 @akihironitta @awaelchli

tchaton:
Hey @Borda. I think this one is a `P0` as it is a very common usage.

Borda:
@awaelchli just testing with v1.3.1 and seems like the training does not start at all :/

awaelchli:
I did a quick test on Grid with this snippet:

```python
import torch.multiprocessing as mp


def foo(process_idx):
    print(process_idx)


if __name__ == "__main__":
    mp.spawn(foo, nprocs=2)  # should print 0 1
```

It crashes in Jupyter environment and works fine as a regular python program. 

```
---------------------------------------------------------------------------
ProcessExitedException                    Traceback (most recent call last)
<ipython-input-20-53d29309dc61> in <module>
      7 
      8 if __name__ == "__main__":
----> 9     mp.spawn(foo, nprocs=2)  # should print 0 1

~/conda/lib/python3.7/site-packages/torch/multiprocessing/spawn.py in spawn(fn, args, nprocs, join, daemon, start_method)
    228                ' torch.multiprocessing.start_process(...)' % start_method)
    229         warnings.warn(msg)
--> 230     return start_processes(fn, args, nprocs, join, daemon, start_method='spawn')

~/conda/lib/python3.7/site-packages/torch/multiprocessing/spawn.py in start_processes(fn, args, nprocs, join, daemon, start_method)
    186 
    187     # Loop on join until it returns True or raises an exception.
--> 188     while not context.join():
    189         pass
    190 

~/conda/lib/python3.7/site-packages/torch/multiprocessing/spawn.py in join(self, timeout)
[...]

PR #12008, opened 2022-02-19
Fix `is_interactive_compatible` logic after AcceleratorConnector rewrite

awaelchli:
## What does this PR do?

Follow up to #11448 
The PR introduced a bug with Jupyter notebooks because the strategy name property was changed. The check how it is implemented now doesn't make much sense. 

#11970 proposes to move it to the strategy base class itself but my proposal here is to move it to the newly added launcher. IMO this fixes the issue more efficiently. 

[Notebook for manual checking.](https://colab.research.google.com/drive/1DMclmZopMeThbJYcLRYAuzzZqYwq4O4V?usp=sharing)

No.

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7 @akihironitta @justusschock @kaushikb11 @awaelchli @ninginthecloud


pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -48,6 +48,10 @@ class _SpawnLauncher(_Launcher):
     def __init__(self, strategy: Strategy) -> None:
         self._strategy = strategy
 
+    @property
+    def is_interactive_compatible(self) -> bool:
+        return False  # TODO: the return value should depend on 1) start_method 2) CUDA vs. CPU
```
awaelchli, line 53: can do this once #11966 refactor is done. 

Here, it should be `return self.start_method == "spawn" or strategy.root_device == cuda`
See #7550 for more context.

tests/accelerators/test_accelerator_connector.py
```diff
@@ -393,19 +394,28 @@ def test_dist_backend_accelerator_mapping(*_):
...
-@mock.patch("pytorch_lightning.utilities._IS_INTERACTIVE", return_value=True)
-def test_ipython_compatible_backend(*_):
-    Trainer(strategy="ddp_spawn", num_processes=2)
+    with pytest.raises(MisconfigurationException, match=r"strategy='ddp'\)`.*is not compatible"):
+        # Edge case: AcceleratorConnector maps dp to ddp if accelerator != gpu
+        Trainer(strategy="dp")
+
+
+def test_ipython_compatible_backend(monkeypatch):
+    monkeypatch.setattr(pytorch_lightning.utilities, "_IS_INTERACTIVE", True)
```
awaelchli, line 415: using monkeypatch now because previously `_IS_INTERACTIVE` became a MagicMock, which was fine because it would still take the branch `if _IS_INTERACTIVE:` but not technically correct way of mocking.

four4fish: That's what happend for distermintic! Do we have similar issues in other places?
I remembered TPU_available mock below seems have similar issues.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -715,19 +715,16 @@ def _lazy_init_strategy(self) -> None:
 
         from pytorch_lightning.utilities import _IS_INTERACTIVE
 
-        # TODO move is_compatible logic to strategy API
-        interactive_compatible_strategy = (
+        interactive_recomended_strategy = (
```
four4fish, line 718: Do you think we should move it to _init_strategy? which fails faster.

four4fish: I was thinking add this to strategy API which could be called at init_strategy. I think in launcher make much more sense. +1

four4fish (approved):

pytorch_lightning/trainer/connectors/accelerator_connector.py
kaushikb11: Why change the variable from `compatible` to `recommended`?

kaushikb11 (approved):

pytorch_lightning/trainer/connectors/accelerator_connector.py
carmocca: +1, let's keep the previous name

tests/accelerators/test_accelerator_connector.py
```diff
@@ -393,19 +394,28 @@ def test_dist_backend_accelerator_mapping(*_):
...
+    Trainer()
+    Trainer(strategy="dp", accelerator="gpu")
+    Trainer(accelerator="tpu")
```
carmocca, line 418: What is this testing exactly? Is it that the instantiation works? Can we have some assertions?

carmocca (approved):
Maybe worth waiting until #11966 is merged so Adrian's comment can be addressed here

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: > Do you think we should move it to _init_strategy? which fails faster.

We can't because we need to run `configure_launcher` first and this needs to happen in `lazy_init`.


> Why change the variable from compatible to recommended?

I did this because we decoupled the launchers from the strategy. So the concept of being "compatible with interactive environments" has shifted away from the strategy and it made more sense to recomend (spelling error not intended) strategies that come with compatible launchers by default. Anyway, refactored now.

pytorch_lightning/strategies/launchers/spawn.py
kaushikb11: AFAIK, it should work for xla spawn as well

tests/accelerators/test_accelerator_connector.py
```diff
@@ -393,19 +394,31 @@ def test_dist_backend_accelerator_mapping(*_):
...
+        Trainer(strategy="ddp_spawn")
 
-@mock.patch("pytorch_lightning.utilities._IS_INTERACTIVE", return_value=True)
-def test_ipython_compatible_backend(*_):
-    Trainer(strategy="ddp_spawn", num_processes=2)
+    with pytest.raises(MisconfigurationException, match=r"strategy='ddp_sharded_spawn'\)`.*is not compatible"):
+        Trainer(strategy="ddp_sharded_spawn")
+
+    with pytest.raises(MisconfigurationException, match=r"strategy='ddp'\)`.*is not compatible"):
+        # Edge case: AcceleratorConnector maps dp to ddp if accelerator != gpu
```
kaushikb11, line 413: ># Edge case: AcceleratorConnector maps dp to ddp if accelerator != gpu

But we should let the users know about it through a warning?

pytorch_lightning/utilities/enums.py
```diff
@@ -254,8 +254,6 @@ def interactive_compatible_types() -> list[_StrategyType]:
         """Returns a list containing interactive compatible _StrategyTypes."""
         return [
             _StrategyType.DP,
-            _StrategyType.DDP_SPAWN,
-            _StrategyType.DDP_SHARDED_SPAWN,
```
kaushikb11, line 258: @awaelchli It's supported, right?

pytorch_lightning/strategies/launchers/spawn.py
awaelchli: yeah definitely, for xla spawn we return True.

awaelchli: for the xla spawn launcher, we can just return True

tests/accelerators/test_accelerator_connector.py
awaelchli: Well, this is and was the behavior captured by the accelerator connector and associated tests. it is out of scope for this PR for me to add a warning. Here, I'm concerned with raising an exception in an interactive environment.

pytorch_lightning/utilities/enums.py
awaelchli: No, pytorch does no longer support multiprocessing in jupyter notebooks and cuda. only fork+cpu seems to be working according to some tests I did weeks ago. #7550

awaelchli: added a comment in the code

PR #11944, opened 2022-02-16
Default `benchmark` based on `deterministic` flag

krshrimali:
## What does this PR do?

* This PR makes the `benchmark` flag of `Trainer` class `Optional`, and defaults to the value based on `deterministic` flag. Also allows manually overwriting the flag.
* Why not deprecate? We wish to allow overwriting the flag by the users, and prevent BC-breaking changes.
* Tests have been updated accordingly.

Fixes #9128.

### Does your PR introduce any breaking changes? If yes, please list them.

No!

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Thanks to @carmocca for helping me with this PR. 🎉

PR #12020, opened 2022-02-20
Avoid changing current cudnn benchmark

### LINKED ISSUE #12018: PL overrides globally set torch.backends.cudnn.benchmark value
## 🐛 Bug

Currently, globally turning on cudnn benchmarking in torch (`torch.backends.cudnn.benchmark = True)` does nothing as it is overridden when constructing a `Trainer` object. However, it's reasonable for users to expect modification of `torch.backends.cudnn.benchmark` to be respected by PL.

More intuitive behaviour would be for PL to only modify the current value of `torch.backends.cudnn.benchmark` when it is explicitly set when constructing the `Trainer` via the `benchmark` arg.

### To Reproduce

```python
import torch

torch.backends.cudnn.benchmark = True
trainer = Trainer(gpus=1)
print(torch.backends.cudnn.benchmark)
```
Output:
```
False
# When it should be True
```

### Expected behavior

For `torch.backends.cudnn.benchmark` to be changed by `Trainer` only when the `benchmark` arg is explicitly set.

| Instantiation | Behaviour |
|-|-|
| `Trainer()` | `torch.backends.cudnn.benchmark` is unchanged from current session value |
| `Trainer(benchmark=None)` | `torch.backends.cudnn.benchmark` is unchanged from current session value |
| `Trainer(benchmark=True)` | `torch.backends.cudnn.benchmark` set to `True` |
| `Trainer(benchmark=False)` | `torch.backends.cudnn.benchmark` set to `False` |

### Environment

```
* CUDA:
        - GPU:
                - NVIDIA GeForce RTX 3070 Laptop GPU
        - available:         True
        - version:           11.1
* Packages:
        - numpy:             1.21.4
        - pyTorch_debug:     False
        - pyTorch_version:   1.9.0+cu111
        - pytorch-lightning: 1.5.3
        - tqdm:              4.62.3
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.9.7
        - version:           #1 SMP Fri Apr 2 22:23:49 UTC 2021
```

timesler:
## What does this PR do?

This is a small change that prevents PL from silently overriding `torch.backends.cudnn.benchmark` when constructing a `Trainer` object.

Fixes #12018

With this change, the behaviour can be described by:

For `torch.backends.cudnn.benchmark` to be changed by `Trainer` only when the `benchmark` arg is explicitly set.

| Instantiation | Behaviour |
|-|-|
| `Trainer()` | `torch.backends.cudnn.benchmark` is unchanged from current session value |
| `Trainer(benchmark=None)` | `torch.backends.cudnn.benchmark` is unchanged from current session value |
| `Trainer(benchmark=True)` | `torch.backends.cudnn.benchmark` set to `True` |
| `Trainer(benchmark=False)` | `torch.backends.cudnn.benchmark` set to `False` |

## Before submitting

## PR review

## Did you have fun?


timesler:
This should likely wait until #11944 is merged, then be updated to ensure the same silent updating of `cudnn.benchmark` doesn't happen with the deterministic arg as well.

docs/source/common/trainer.rst
```diff
@@ -416,19 +416,19 @@ benchmark
...
 
 The speedup comes from allowing the cudnn auto-tuner to find the best
 algorithm for the hardware `[see discussion here]
 <https://discuss.pytorch.org/t/what-does-torch-backends-cudnn-benchmark-do/5936>`_.
 
 Example::
 
-    # default used by the Trainer
-    trainer = Trainer(benchmark=False)
+    # default used by the Trainer (current setting for torch.backends.cudnn.benchmark is unchanged)
```
krshrimali, line 430: minor nit for docs:

```suggestion
# default used by the Trainer (current setting for ``torch.backends.cudnn.benchmark`` is unchanged)
```

timesler: I think literal blocks have to be indented in RST https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

timesler: Actually, since this comment is also in the literal example block, it is code also. So no backticks are needed if we are to follow the convention in the rest of this file

krshrimali: Apologies for the indentation suggestion, it was by fault.

And sounds good, you are right!

docs/source/common/trainer.rst
```diff
@@ -416,19 +416,19 @@ benchmark
 
 |
 
-If true enables cudnn.benchmark.
-This flag is likely to increase the speed of your system if your
-input sizes don't change. However, if it does, then it will likely
-make your system slower.
+The value (``True`` or ``False``) to set ``torch.backends.cudnn.benchmark`` to. If not specified, the value
+set in the current session will be unchanged.
```
awaelchli, line 420:
```suggestion
set in the current session will be used.
```

docs/source/common/trainer.rst
```diff
@@ -416,19 +416,19 @@ benchmark
...
+Setting this flag to ``True`` is likely to increase the speed of your system if your input sizes don't
```
awaelchli, line 421:
```suggestion
Setting this flag to ``True`` is likely going to increase the speed of your system if your input sizes don't
```

docs/source/common/trainer.rst
awaelchli, line 430:
```suggestion
    # default used by the Trainer (will use whatever the current value for torch.backends.cudnn.benchmark is)
```

awaelchli (approved):

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -138,7 +138,8 @@ def __init__(
             C. TODO When multiple flag set to the same thing
         """
         # TODO: move to gpu accelerator
-        torch.backends.cudnn.benchmark = benchmark
+        if benchmark is not None:
+            torch.backends.cudnn.benchmark = benchmark
```
awaelchli, line 142: note, this code has changed on master
would you mind resolving the merge conflicts?

timesler: Resolved. This required a bit of thought though, due to the interaction between `deterministic` and `benchmark`. Here is the behaviour now implemented in this PR, which I think is the correct amount intuitiveness and best practice.

* Whenever the user explicitly sets `deterministic` or `benchmark`, those choices are respected (although a warning is raised if both are `True`, as was the behaviour already, and this is sensible)
* When `deterministic` is set to `True`, `benchmark` defaults to `False` (though it can be overridden)
* When `deterministic` is not set, it defaults to `False`, as it does currently in master. However, it is set as an optional argument defaulting to `None` so that we can determine if it was explicitly set by the user or not, then it is defaulted to `False` after some simple logic is applied to `benchmark`

docs/source/common/trainer.rst
```diff
@@ -416,22 +416,29 @@ benchmark
...
 
-    # defaults to True if not deterministic (which is False by default)
-    trainer = Trainer()
+    # default used by the Trainer (will use whatever the current value for torch.backends.cudnn.benchmark is)
+    trainer = Trainer(benchmark=None)
 
     # you can overwrite the value
     trainer = Trainer(benchmark=False)
 
+    # defaults to False when deterministic is True
```
carmocca, line 439:
```suggestion
    # `benchmark` defaults to False when deterministic is True
```

tests/trainer/test_trainer.py
```diff
@@ -641,12 +641,15 @@ def test_trainer_max_steps_accumulate_batches(tmpdir):
 @pytest.mark.parametrize(
     ["benchmark_", "deterministic", "expected"],
     [
-        (None, False, True),
+        (None, False, None),
         (None, True, False),
+        (None, None, None),
```
carmocca, line 646: So now, somebody that creates `Trainer()` with no arguments, won't get `cudnn.benchmark=True` as it does in master?

I think having `benchmark=True` by default is best.

The proposal in this PR only works if we were able to know if the user had set `benchmark` manually before.

ananthsub (approved):

tests/trainer/test_trainer.py
timesler: The justification for this change is that Pytorch Lightning shouldn't break standard Pytorch functionality, but extend it. In general, PL offers value by establishing sensible default values so users don't need to think about them usually. However, in this case, we are silently setting a global variable, resulting in this behaviour:

```python
import torch

torch.backends.cudnn.benchmark = True
trainer = Trainer(gpus=1)
print(torch.backends.cudnn.benchmark)
```

Output:

```
False
# When it should be True
```

carmocca: I understand that, but since there's no way for us to know whether the user previously set the value, we have 2 exclusive options:
1. Have the better default for most people which may override an existing value (current master)
2. Always respect the existing value but users have to remember to set this flag (this PR)

I personally prefer 1 as it establishes a "sensible default". We could request the torch folks to add an optional default so the options are not exclusive in the future.

ccing revierwers @awaelchli @ananthsub @krshrimali to see if they think differently.

awaelchli: > I understand that, but since there's no way for us to know whether the user previously set the value,

We can know, by making the default value for the argument None. A default `Trainer()` would result in taking the value from `torch.backends.cudnn.benchmark` which by default is True

carmocca: > torch.backends.cudnn.benchmark which by default is True

Is it? locally:

```bash
$ python -c 'import torch; print(torch.backends.cudnn.benchmark)'
False
```

Issue #11379, opened 2022-01-09
Resuming training with ModelCheckpoint can delete checkpoints in other runs


lucmos:
## 🐛 Bug
The `ModelCheckpoint` tracks the old checkpoints path and possibly deletes them while resuming even if the checkpoint is configured with a new dirpath.

### Expected behavior
The `ModelCheckpoint` should manage only the checkpoints in `default_root_dir/name/version`

### Environment
```bash
* CUDA:
        - GPU:
                - NVIDIA GeForce RTX 2080 Ti
        - available:         True
        - version:           10.2
* Packages:
        - numpy:             1.21.2
        - pyTorch_debug:     False
        - pyTorch_version:   1.8.1+cu102
        - pytorch-lightning: 1.5.8
        - tqdm:              4.62.3
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         
        - python:            3.9.7
        - version:           #1 SMP PREEMPT Wed Dec 29 18:09:17 UTC 2021
```

### Additional context

Issue opened after slack discussion.

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7 @carmocca @jjenniferdai

rohitgr7:
from what we discussed over slack, I think a more appropriate title would be that it tracks the old checkpoints path and possibly deletes them while resuming even if the checkpoint is configured with a new `dirpath`.

Possible workarounds I can think of right now:
1. Either we can stop tracking the best_k_models on reload if `dirpath` is changed or
2. A user has to backup/move their old checkpoints to be on the safe side.

awaelchli:
I think we could explore 1, by comparing `self.dir_path` in `ModelCheckpiont.on_load_checkpoint` with the one in the checkpoint.

krshrimali:
Thanks, @lucmos for creating the issue! @rohitgr7 @awaelchli - If it's fine, I am assigning myself as I plan to work on this. :) In case anyone has already started working on it, please let me know.

kaushikb11:
@krshrimali Go ahead with it! Assigning the issue to you.

rohitgr7:
re-opening since I think this is an issue still: https://github.com/PyTorchLightning/pytorch-lightning/pull/12045#issuecomment-1054334875

YishGene:
Hi,

I'm still encountering this with version 2.0.2.

PR #11794, opened 2022-02-07
Deprecate callback hooks `on_pretrain_routine_start` and `on_pretrain_routine_end`

krishnakalyan3:
## What does this PR do?

Fixes #10984

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #12045, opened 2022-02-22
Stop loading a few properties if checkpoint's `dirpath` has changed

krshrimali:
## What does this PR do?

* When resuming training, if a new checkpoint `dirpath` is provided, current state of `on_load_checkpoint` will track all the properties (like `best_model_score`, `kth_best_model_path`, etc.). We should only track `last_model_path` and `best_model_path` and stop tracking `best_model_score`, `kth_best_model_path`, `kth_value`, `best_k_models` if the checkpoint path has changed. Please see issue https://github.com/PyTorchLightning/pytorch-lightning/issues/11379 and comment by @rohitgr7 here: https://github.com/PyTorchLightning/pytorch-lightning/issues/11379#issuecomment-1008361651.
* This PR also adds a warning in case the path has changed, telling the user what will be tracked and what not.
* `__resolve_ckpt_dir` has been moved from `on_pretrain_routine_start` to `_setup` function. Please see the discussion [here](https://github.com/PyTorchLightning/pytorch-lightning/pull/12045#discussion_r812618473) for more context.

Yes, if the checkpoint has changed on resumed training, the following properties won't be tracked:

1. `best_model_score`
2. `kth_best_model_path`
3. `kth_value`
4. `best_k_models`

Only `last_model_path` and `best_model_path` will be tracked.

Additionally, a warning will be raised if the checkpoint path has changed.

## Before submitting

## PR review

## Did you have fun?


tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1231,14 +1231,21 @@ def test_model_checkpoint_saveload_ckpt(tmpdir):
...
+    # 1. "current_score", "dirpath" and "monitor" are currently not restored by on_load_checkpoint.
+    #    We therefore set "dirpath" and "monitor" to something different than for ckpt/cb_write so we can assert them.
+    # 2. "current_score" is left as initialized, i.e. None, and can therefore also be asserted
+    # 3. When a different `dirpath` is passed to `ModelCheckpoint` to resume training, only
+    #    `best_model_path` and `last_model_path` are tracked (tracking for others is stopped).
     cb_restore = ModelCheckpoint(dirpath=tmpdir + "restore", monitor=None, save_top_k=-1, save_last=True)
-    cb_restore.on_load_checkpoint("", "", written_ckpt)
+    with pytest.warns(UserWarning, match="The dirpath was changed from*"):
+        cb_restore.on_load_checkpoint("", "", written_ckpt)
     for key, val in written_ckpt.items():
```
carmocca, line 1243: Can you refactor the code below to assert like the following instead of with specific `if`s?

```python
expected_keys = {"current_score": True, ...}
# iterate over the written ckpt so it fails if new keys are added
for k, v in written_ckpt.items():
    should_match = expected_keys[k]
    if should_match:
        assert getattr(cb_restore, key) == val
    else:
        assert getattr(cb_restore, key) != val
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +354,30 @@ def on_save_checkpoint(
...
-        self.best_model_score = callback_state["best_model_score"]
-        self.best_model_path = callback_state["best_model_path"]
-        self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
-        self.kth_best_model_path = callback_state.get("kth_best_model_path", self.kth_best_model_path)
-        self.kth_value = callback_state.get("kth_value", self.kth_value)
-        self.last_model_path = callback_state.get("last_model_path", self.last_model_path)
+        # Need to get real path from the given `dirpath`, we follow the template from
+        # __init_ckpt_dir function to get the real path for comparison
+        dirpath_from_ckpt = callback_state.get("dirpath", self.dirpath)
+        _fs = get_filesystem(dirpath_from_ckpt if dirpath_from_ckpt else "")
```
carmocca, line 360: This does not need to be protected as it's a local variable

```suggestion
        fs = get_filesystem(dirpath_from_ckpt if dirpath_from_ckpt else "")
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +354,30 @@ def on_save_checkpoint(
...
+
+        if dirpath_from_ckpt and _fs.protocol == "file":
+            dirpath_from_ckpt = os.path.realpath(dirpath_from_ckpt)
```
carmocca, line 363: I don't think this is necessary

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +354,30 @@ def on_save_checkpoint(
...
+
+        # For more context, see: https://github.com/PyTorchLightning/pytorch-lightning/issues/11379
```
carmocca, line 365: You don't need to link to the issue or PR, git blame can be used to find the context

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +354,30 @@ def on_save_checkpoint(
...
+        if self.dirpath == dirpath_from_ckpt:
+            # TODO: Post discussion with @rohitgr7, maybe also consider monitor?
+            # If dirpath has changed, don't track the following properties
```
carmocca, line 368: This says "if dirpath has changed" but we are in the "dirpath has not changed" block

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali: Right! Should be fixed now.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +354,30 @@ def on_save_checkpoint(
...
+        # __init_ckpt_dir function to get the real path for comparison
+        dirpath_from_ckpt = callback_state.get("dirpath", self.dirpath)
+        _fs = get_filesystem(dirpath_from_ckpt if dirpath_from_ckpt else "")
+
+        if dirpath_from_ckpt and _fs.protocol == "file":
+            dirpath_from_ckpt = os.path.realpath(dirpath_from_ckpt)
+
+        # For more context, see: https://github.com/PyTorchLightning/pytorch-lightning/issues/11379
+        if self.dirpath == dirpath_from_ckpt:
+            # TODO: Post discussion with @rohitgr7, maybe also consider monitor?
```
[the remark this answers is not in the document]
krshrimali: I'll create an issue for this, and probably we can fix that in a follow-up PR.

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali: Sure, removed now.

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali: My bad, sorry. Fixed now.

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali: This wasn't needed, so removed this.

tests/checkpointing/test_model_checkpoint.py
krshrimali: Done! :)

pytorch_lightning/callbacks/model_checkpoint.py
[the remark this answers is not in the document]
rohitgr7: but if I change the metric being monitored, then old best_k_checkpoints become redundant no? since it will be best_k on the old one?

CHANGELOG.md
```diff
@@ -561,6 +561,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed an issue where `ModelCheckpoint` could delete older checkpoints when `dirpath` has changed during resumed training. [#12045](https://github.com/PyTorchLightning/pytorch-lightning/pull/12045)
```
rohitgr7, line 564:
```suggestion
- Fixed an issue where `ModelCheckpoint` could delete older checkpoints when `dirpath` has changed during resumed training ([#12045](https://github.com/PyTorchLightning/pytorch-lightning/pull/12045))
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -150,6 +151,9 @@ class ModelCheckpoint(Callback):
             if ``mode`` is none of ``"min"`` or ``"max"``.
         ValueError:
             If ``trainer.save_checkpoint`` is ``None``.
+        UserWarning:
```
rohitgr7, line 154: I think raises is used only for errors:
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -190,6 +194,9 @@ class ModelCheckpoint(Callback):
 
         *monitor, mode, every_n_train_steps, every_n_epochs, train_time_interval, save_on_train_epoch_end*
 
+    .. note:: If the checkpoint's `dirpath` changed from what it was before, only `last_model_path` and
+        `best_model_path` will be tracked and a warning will be issued.
```
rohitgr7, line 198:
```suggestion
    .. note:: If the checkpoint's ``dirpath`` changed from what it was before while resuming the training,
        only ``last_model_path`` and ``best_model_path`` will be tracked and a warning will be issued.
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +360,23 @@ def on_save_checkpoint(
...
+        dirpath_from_ckpt = callback_state.get("dirpath", self.dirpath)
+
+        if self.dirpath == dirpath_from_ckpt:
+            self.best_model_score = callback_state["best_model_score"]
+            self.kth_best_model_path = callback_state.get("kth_best_model_path", self.kth_best_model_path)
+            self.kth_value = callback_state.get("kth_value", self.kth_value)
+            self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
+        else:
+            # If dirpath has changed, raise a warning and only track last_model_path
+            # and best_model_path
```
rohitgr7, line 372:
```suggestion
```
comment isn't required

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +360,23 @@ def on_save_checkpoint(
...
+            warnings.warn(
+                f"The dirpath was changed from {self.dirpath} to {dirpath_from_ckpt},"
```
rohitgr7, line 374:
```suggestion
                f"The dirpath has changed from {dirpath_from_ckpt!r} to {self.dirpath!r},"
```
from old_path to new_path, right?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +360,23 @@ def on_save_checkpoint(
...
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", callback_state: Dict[str, Any]
     ) -> None:
-        self.best_model_score = callback_state["best_model_score"]
-        self.best_model_path = callback_state["best_model_path"]
-        self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
-        self.kth_best_model_path = callback_state.get("kth_best_model_path", self.kth_best_model_path)
-        self.kth_value = callback_state.get("kth_value", self.kth_value)
+        dirpath_from_ckpt = callback_state.get("dirpath", self.dirpath)
+
+        if self.dirpath == dirpath_from_ckpt:
```
rohitgr7, line 365: I think if dirpath is not set in the next run, a new version is assigned and checkpoints will be added there. This happens during `on_pretrain_routine_start`, thus the final checkpoint is decided after that hook is called. But the checkpoint states are loaded before that, so `self.dirpath` may not be initialized correctly by that time.

`on_pretrain_routine_start/end` is under deprecation. https://github.com/PyTorchLightning/pytorch-lightning/pull/11794/files
the alternative `on_fit_start` for it, will again be a problem for us since this is called after callback states are loaded in a general case.

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali:
```suggestion
                f"The dirpath has changed from {dirpath_from_ckpt} to {self.dirpath},"
```

krshrimali: > from old_path to new_path, right?

Right, good catch.

pytorch_lightning/callbacks/model_checkpoint.py
carmocca: Can't we call `__resolve_ckpt_dir` here in that case?

rohitgr7: or maybe I was rather thinking of moving the call from `on_pretrain_routine_start` to `setup` which will solve the problem. Need to think if there's any edge case here where it will be an issue.

just trying to avoid calling a private method inside a public hook.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -353,12 +357,21 @@ def on_save_checkpoint(
...
+            self.best_model_score = callback_state["best_model_score"]
+            self.kth_best_model_path = callback_state.get("kth_best_model_path", self.kth_best_model_path)
+            self.kth_value = callback_state.get("kth_value", self.kth_value)
+            self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
+        else:
+            warnings.warn(
+                f"The dirpath has changed from {dirpath_from_ckpt} to {self.dirpath},"
```
rohitgr7, line 369:
```suggestion
                f"The dirpath has changed from {dirpath_from_ckpt!r} to {self.dirpath!r},"
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -250,14 +254,14 @@ def state_key(self) -> str:
     def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: Optional[str] = None) -> None:
         # NOTE: setting these attributes needs to happen as early as possible BEFORE reloading callback states,
         # because the attributes are part of the state_key which needs to be fully defined before reloading.
+        self.__resolve_ckpt_dir(trainer)
```
rohitgr7, line 257: since it is moved now, can you add a test reflecting the case that covers this case, where `dirpath` is not set explicitly, but is created dynamically and will be different from what it was during the initial run?

carmocca, line 257: Can you put this before the comment directly above? It does not apply to the `dirpath`

carmocca (approved):

rohitgr7 (approved):

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -190,6 +191,9 @@ class ModelCheckpoint(Callback):
 
         *monitor, mode, every_n_train_steps, every_n_epochs, train_time_interval, save_on_train_epoch_end*
 
+    .. note:: If the checkpoint's ``dirpath`` changed from what it was before while resuming the training,
+        only ``last_model_path`` and ``best_model_path`` will be tracked and a warning will be issued.
```
awaelchli, line 195: can we add it to the existing Note section above?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -351,12 +355,21 @@ def on_save_checkpoint(
...
+        if self.dirpath == dirpath_from_ckpt:
+            self.best_model_score = callback_state["best_model_score"]
+            self.kth_best_model_path = callback_state.get("kth_best_model_path", self.kth_best_model_path)
+            self.kth_value = callback_state.get("kth_value", self.kth_value)
+            self.best_k_models = callback_state.get("best_k_models", self.best_k_models)
+        else:
+            warnings.warn(
+                f"The dirpath has changed from {dirpath_from_ckpt!r} to {self.dirpath!r},"
+                " therefore `best_model_score`, `kth_best_model_path`, `kth_value` and `best_k_models`"
+                " won't be tracked. Only `last_model_path` and `best_model_path` will be tracked."
```
awaelchli, line 369: what does it have to do with tracking? do you mean it won't get loaded?

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1210,37 +1210,64 @@ def test_check_val_every_n_epochs_top_k_integration(tmpdir):
...
+    with pytest.warns(UserWarning, match="The dirpath has changed from*"):
+        cb_restore.on_load_checkpoint("", "", written_ckpt)
+    make_assertions(cb_restore, written_ckpt)
+
+    # Case - 2
+    # Make sure that everything runs when dirpath is not initialized explicitly
+    cb_restore = CustomModelCheckpoint()
+    cb_restore.setup(Trainer(), BoringModel())
+    with pytest.warns(UserWarning, match="The dirpath has changed from*"):
+        cb_restore.on_load_checkpoint("", "", written_ckpt)
```
awaelchli, line 1269: Hint: you can use Mock() to pass in an arbitrary value while avoiding the linting errors

pytorch_lightning/callbacks/model_checkpoint.py
awaelchli: And I fail to see why "tracking" the remaining attributes would suddenly be prevented. 
It's not loading them, that's clear. But we can still encounter a new best model and start with the new top-k if we want.

awaelchli (approved):

pytorch_lightning/callbacks/model_checkpoint.py
awaelchli: Is it because we don't load best_model_score?

awaelchli:
Here is a full example of what I mean in my comments above.

1. The metric still gets tracked and we keep saving the best checkpoints. This is IMO expected behavior. All we want is to avoid that the checkpoints in the old folder get deleted and this is what this PR very well addresses <3 However, let's change the wording in the messages and comments. 

2. A bug remains where last.ckpt gets removed from the old directory. 

```py

import os
import shutil

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("latest_is_best", self.global_step)
        return {"loss": loss}

def run():
    shutil.rmtree("./first", ignore_errors=True)
    shutil.rmtree("./after-reload", ignore_errors=True)

    model = BoringModel()
    checkpoint = ModelCheckpoint(
        dirpath="./first", monitor="latest_is_best", mode="max", save_top_k=3, save_last=True,
        every_n_train_steps=1,
        filename='{epoch}-{step}-{latest_is_best}',
    )
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=5,
        max_epochs=1,
        enable_model_summary=False,
        callbacks=[checkpoint],
    )

    # NOTE: last exists here, but not later on.
    assert os.path.isfile("./first/last.ckpt")


    # ------------------------
    # RESUME WITH CHANGED PATH
    checkpoint = ModelCheckpoint(
        dirpath="./after-reload", monitor="latest_is_best", mode="max", save_top_k=3, save_last=True,
        every_n_train_steps=1,
        filename='{epoch}-{step}-{latest_is_best}',
    )
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=5,
        max_epochs=2,   # MORE EPOCHS
        enable_model_summary=False,
        callbacks=[checkpoint],
    )
    trainer.fit(model, train_dataloaders=train_data, ckpt_path="./first/epoch=0-step=3-latest_is_best=3.0.ckpt")

    # OBSERVE: last.ckpt has disappeared from the first folder
    # BUG?
    assert not os.path.isfile("./first/last.ckpt")


if __name__ == "__main__":
    run()
```

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali, line 369:
```suggestion
                " won't be reloaded. Only `last_model_path` and `best_model_path` will be reloaded."
```

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1210,37 +1210,64 @@ def test_check_val_every_n_epochs_top_k_integration(tmpdir):
...
-        if key not in ("current_score", "dirpath", "monitor"):
-            assert getattr(cb_restore, key) == val
-        else:
-            assert getattr(cb_restore, key) != val
+    # Notes:
+    # 1. "current_score", "dirpath" and "monitor" are currently not restored by on_load_checkpoint.
+    #    We therefore set "dirpath" and "monitor" to something different than for ckpt/cb_write so we can assert them.
+    # 2. "current_score" is left as initialized, i.e. None, and can therefore also be asserted
+    # 3. When a different `dirpath` is passed to `ModelCheckpoint` to resume training, only
+    #    `best_model_path` and `last_model_path` are tracked (tracking for others is stopped).
```
krshrimali, line 1258:
```suggestion
    #    `best_model_path` and `last_model_path` are reloaded (reloading for others is stopped).
```

tests/checkpointing/test_model_checkpoint.py
krshrimali: Yep, I and @carmocca tried using it here but unfortunately faced a few issues. Hence we decided to keep it this way.

krshrimali:
@awaelchli - Thank you so much for taking a look. You are right, the words were misleading, and they should be fixed now. Regarding the bug you mentioned, that's legit, and I'll create a follow-up PR to fix it. This also involves re-thinking if `last_model_path` should be loaded, and also modifying tests.

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali: Sorry, wrong words were used here. Yes, I meant `reloaded` instead of `tracked`.

pytorch_lightning/callbacks/model_checkpoint.py
krshrimali: > can we add it to the existing Note section above?

Done! :)

> And I fail to see why "tracking" the remaining attributes would suddenly be prevented.

Sorry, `tracking` was the wrong term to use. In the follow-up PR, to fix the bug you also mentioned, we'll take a look at what should be reloaded, and what not.

PR #12052, opened 2022-02-22
Suggested Change to Wiki Documentation

figgybit:
Shouldn't this be `y` not `x`?


rohitgr7 (commented):
nope!
it's an autoencoder so should be `x`.

figgybit:
@rohitgr7 thanks for pytorch lightning!

PR #12054, opened 2022-02-23
[not for review] parallel_device_ids demo

DuYicong515:
## What does this PR do?

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/callbacks/gpu_stats_monitor.py
+++ b/pytorch_lightning/callbacks/gpu_stats_monitor.py
@@ -28,6 +28,7 @@ from typing import Any, Dict, List, Optional, Tuple
 import torch
 
 import pytorch_lightning as pl
+from pytorch_lightning.accelerators import GPUAccelerator
 from pytorch_lightning.callbacks.base import Callback
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.parsing import AttributeDict
@@ -133,8 +134,16 @@ class GPUStatsMonitor(Callback):
             )
 
         # The logical device IDs for selected devices
-        # ignoring mypy check because `trainer.data_parallel_device_ids` is None when using CPU
-        self._device_ids = sorted(set(trainer.data_parallel_device_ids))  # type: ignore
+        # this is going to be wrong since parallel_device_ids=[0] if we use GPU + SingleDeviceStrategy
+        # However, parallel_devices now only exists in ParallelStrategy
+        # self._device_ids changes from [0] to [] (L121)
+        if hasattr(trainer.strategy, "parallel_devices"):
+            parallel_device_ids = (
+                [i for i in range(len(trainer.strategy.parallel_devices))]
+                if isinstance(trainer.accelerator, GPUAccelerator)
+                else None
+            )
+            self._device_ids = sorted(set(parallel_device_ids))
 
         # The unmasked real GPU IDs
         self._gpu_ids = self._get_gpu_ids(self._device_ids)
--- a/tests/strategies/test_single_device_strategy.py
+++ b/tests/strategies/test_single_device_strategy.py
@@ -54,6 +54,17 @@ def test_single_gpu():
     assert cuda_memory < model.start_cuda_memory
 
 
+@RunIf(skip_windows=True, min_gpus=1)
+def test_single_gpu_parallel_device():
+    """Tests if device is set correctly when training and after teardown for single GPU strategy."""
+    trainer = Trainer(gpus=1, fast_dev_run=True)
+    # assert training strategy attributes for device setting
+    assert isinstance(trainer.strategy, SingleDeviceStrategy)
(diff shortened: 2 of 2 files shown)


pytorch_lightning/callbacks/gpu_stats_monitor.py
```diff
@@ -133,8 +134,16 @@ def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: O
             )
 
         # The logical device IDs for selected devices
-        # ignoring mypy check because `trainer.data_parallel_device_ids` is None when using CPU
-        self._device_ids = sorted(set(trainer.data_parallel_device_ids))  # type: ignore
+        # this is going to be wrong since parallel_device_ids=[0] if we use GPU + SingleDeviceStrategy
+        # However, parallel_devices now only exists in ParallelStrategy
+        # self._device_ids changes from [0] to [] (L121)
+        if hasattr(trainer.strategy, "parallel_devices"):
```
DuYicong515, line 140: @four4fish So this migration is apparently wrong with  GPU + SingleDeviceStrategy, self._device_ids changes from [0] to []


Currently parallel_devices is only with ParallelStrategy https://github.com/PyTorchLightning/pytorch-lightning/blob/550d3a640d1bfeb731a16b2996d3160f0f7eb071/pytorch_lightning/strategies/parallel.py#L75-L77

Should I upstream it to Strategy and implement it?

four4fish: This file is deprecated, will be removed right after 1.6 release. We only need to keep logic as-is.
> Should I upstream it to Strategy and implement it?

no, we shouldn't

four4fish: Here you can still call trainer.data_parallel_device_ids
This file will be removed before trainer.data_parallel_device_ids removes

DuYicong515: okay, so I'll keep this. 

However, I feel `getattr(trainer.strategy, "parallel_devices", [])` behaviour isn't the same as `trainer._accelerator_connector.parallel_devices` on GPU + SingleDeviceStrategy. Thus `trainer._accelerator_connector.parallel_device_ids` behaviour will change as well, if we rely on `trainer.strategy.parallel_devices` 

  Do we still want to make the change?

DuYicong515: There's still one other place (not test) using trainer._accelerator_connector.parallel_device_ids if we keep this one unchanged.
https://github.com/PyTorchLightning/pytorch-lightning/blob/4c4b9d540f065c41ad288a78110cecbf45b96409/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py#L226

four4fish: The goal is to 
1. remove accelerator.parallel_device_ids
2. update trainer.data_parallel_device_ids to reproduce the same logic + add deprecation message
3. update tests with pytest.deprecated...

This file shouldn't change

four4fish: Oh I see! The logger connector is using parallel_device_ids. 
The [] and None make no different there I think, could you update it to trainer.strategy.parallel_devices to give it a try?

DuYicong515: If it's using `SingleDeviceStrategy` but with GPU=1 (like in my test case below) it won't have a `trainer.strategy.parallel_devices` since `parallel_devices` is only defined on `ParallelStrategy` so `SingleDeviceStrategy` does not have it.

trainer.strategy.parallel_devices will break on `SingleDeviceStrategy`, however the original `self.trainer._accelerator_connector.parallel_device_ids` returned [0], instead of []/None/Exceptions...

Maybe checkout my test case below?

DuYicong515: with  trainer = Trainer(gpus=1)
`getattr(trainer.strategy, "parallel_devices", [])` or `getattr(trainer.strategy, "parallel_devices", None)` or `trainer._accelerator_connector.parallel_device_ids`

four4fish: Just saw this method will be removed after 1.6 too

To handle single device you can do it in step 2 :update trainer.data_parallel_device_ids to reproduce the same logic + add deprecation message

do something similar to https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/connectors/accelerator_connector.py#L793-L799

The caller side can remain the same

DuYicong515: Oh I see, I'll handle the case in the new logic then. I thought `trainer.strategy.parallel_devices` should logically be an identical replacement for `trainer._accelerator_connector.parallel_devices`, and I was wrong. 

Thanks for the discussions and insights! @four4fish  I'm clear on this now.

Issue #11834, opened 2022-02-10
rework docs to encourage the best practice of having a separate nn.Module


williamFalcon:
A while back we shifted to recommending:
```python
encoder = ...
decoder = nn.Module(...)

class Autoencoder(LightningModule):
    def __init__(self, encoder, decoder):
        self.encoder = encoder
        self.decoder = decoder
```

instead of:
```python
class Autoencoder(LightningModule):
    def __init__(self):
        self.encoder = nn.Module(...)
        self.decoder = nn.Module(...)
```

we need to update all the docs including readme. This DOES NOT mean we don't support the first one, but the recommended thing is to keep your nn.Modules outside of LightningModule so that you don't HAVE to couple it with PL when you go do inference.

However, some people might still want to do:
```python
class Model(LightningModule):
    def __init__(self):
       self.l1 = nn.Module(...)
       self.l2 = nn.Module(...)
```
and that's STILL OK and likely to be the pattern for researchers

Related to https://github.com/PyTorchLightning/pytorch-lightning/issues/10270

cc @borda @rohitgr7

rohitgr7:
one issue, with the first one, is that while saving the hparams, the `nn.Modules` will be saved too in the hyperparameters while calling `self.save_hyperparameters()` unless the user explicitly calls it with `self.save_hyperparameters(ignore=['encoder', 'decoder'])`.

maybe we can improve the `save_hyperparameters()` to not save any `nn.Module`s by default?

awaelchli:
It's true that keeping the model definitions outside helps in production. And we've seen people adopt this pattern very often. However, we should also keep in mind that you can't simply load the nn Module weights from a lightning checkpoint. For example, in the auto encoder example above the parameter names will still include the "encoder." and "decoder." prefixes for all parameter names. 

So simply calling `load_state_dict` in your production code won't be enough, you have to iterate over the state dict and remove the prefix from all keys.

That's why I have recommended this pattern in the past as well

```py
# your lightning module for training
class MyLightningModule(LightningModule):
    def __init__(self):
       self.model = Model()
       
 
# a simple wrapper, instead of LightningModule it inherits from nn.Module
# use this in production
class ProductionModelWrapper(nn.Module):
    def __init__(self):
       self.model = Model()
```

You can now load the trained weights from the checkpoint and directly call
```py
production_model_wrapper = ProductionModelWrapper()
production_model_wrapper.load_state_dict(...)
```

[...]

PR #12068, opened 2022-02-23
Raise a warning when `nn.Module` instance is saved with `save_hyperparameters()`

rohitgr7:
## What does this PR do?

Addresses: https://github.com/PyTorchLightning/pytorch-lightning/issues/11834#issuecomment-1034752642

`nn.Module`s are not hparams and when someone uses LM as systems, they might forget to ignore the arguments and they will be saved inside hparams thus will increase the size of the checkpoint. 

## Before submitting

## PR review

## Did you have fun?

cc @borda


pytorch_lightning/utilities/parsing.py
```diff
@@ -247,9 +248,15 @@ def save_hyperparameters(
             hp = {arg: init_args[arg] for arg in args if isinstance(arg, str)}
             obj._hparams_name = "kwargs"
 
+    hp = obj._to_hparams_dict(hp)
+    # pop all nn.Module
+    for key in list(hp):
+        if isinstance(hp[key], nn.Module):
+            del hp[key]
```
krshrimali, line 255: https://github.com/PyTorchLightning/pytorch-lightning/blob/52013f066641f07c8abb4a27dc0583447101afb9/pytorch_lightning/utilities/parsing.py#L234

Can we just add a check if key is not a `nn.Module` instance in the line above and avoid deleting?

tests/models/test_hparams.py
```diff
@@ -320,8 +320,7 @@ def test_collect_init_arguments(tmpdir, cls):
     if isinstance(model, (SubClassBoringModel, NonSavingSubClassBoringModel)):
         assert model.hparams.subclass_arg == 1200
 
-    if isinstance(model, AggSubClassBoringModel):
-        assert isinstance(model.hparams.my_loss, torch.nn.CosineEmbeddingLoss)
+    assert "my_loss" not in model.hparams
```
krshrimali, line 323: Does this mean that the PR introduces a BC-breaking change?

krshrimali (changes requested):

pytorch_lightning/utilities/parsing.py
rohitgr7: yep!

tests/models/test_hparams.py
rohitgr7: sort of yes.. I actually wanted to discuss this. since my main motivation was to disable saving the modules only, but losses are also inherited from `nn.Module`. If we agree with this, we can mark this as breaking change.

awaelchli:
It's probably ok to do this, but we should stop here. There are many other parameter type that may ore may not be desired to serialize, however we can't do the guess work for everything. 

Note: If we go with this PR, there is no way for the user to force what is being ignored here. I hesitate to accept here, because normally we shouldn't make such assumptions for the user (the framework should not be limiting them).

rohitgr7:
> It's probably ok to do this, but we should stop here. There are many other parameter type that may or may not be desired to serialize, however we can't do the guess work for everything.
> 
> Note: If we go with this PR, there is no way for the user to force what is being ignored here. I hesitate to accept here, because normally we shouldn't make such assumptions for the user (the framework should not be limiting them).

I agree, my idea here was to avoid only `nn.Modules` because that's already been captured when we save `state_dict` inside the checkpoint. Other different types of arguments will not be captured automatically.

I was running t5-small, and it got stuck due to this, with no errors at all, so it was hard to figure out what went wrong here until I did some debugging with @krshrimali .

We can mention this inside docstrings if required.

tests/models/test_hparams.py
carmocca: Yeah, that's a big limitation as the loss is often made configurable.

carmocca:
I'm also hesitant about this.

> I was running t5-small, and it got stuck due to this, with no errors at all, so it was hard to figure out what went wrong here until I did some debugging with @krshrimali .

What if we focus instead on notifying the user about the associated problems? We can suggest excluding the keys with `save_hyperaprameters`

rohitgr7:
> I'm also hesitant about this.
> 
> > I was running t5-small, and it got stuck due to this, with no errors at all, so it was hard to figure out what went wrong here until I did some debugging with @krshrimali .
> 
> What if we focus instead on notifying the user about the associated problems? We can suggest excluding the keys with `save_hyperaprameters`

well, that's a good suggestion. works for me since it will happen during init. If everyone agrees on this, I'll update the PR :)

pytorch_lightning/utilities/parsing.py
```diff
@@ -225,12 +226,23 @@ def save_hyperparameters(
...
+    for k in list(init_args):
+        if k in ignore:
+            del init_args[k]
+            continue
+
+        if isinstance(init_args[k], nn.Module):
+            rank_zero_warn(
+                f"Attribute {k!r} is an instance of `nn.Module` and is already saved during checkpointing."
+                " It is recommended to ignore them using `self.save_hyperparameters(ignore=[k!r])`."
+            )
```
carmocca, line 245: I'd prefer

```python
modules = [arg for arg in init_args if arg not in ignore and isinstance(arg, nn.Module)]
for module in modules:
    warn(...)
```

carmocca (approved):

pytorch_lightning/utilities/parsing.py
rohitgr7: I am doing 2 things here. Filtering the ones that are required to be ignored and raising a warning. can't seem to find a way to do both in a single loop.

pytorch_lightning/utilities/parsing.py
```diff
@@ -247,9 +259,10 @@ def save_hyperparameters(
             hp = {arg: init_args[arg] for arg in args if isinstance(arg, str)}
             obj._hparams_name = "kwargs"
 
+    hp = obj._to_hparams_dict(hp)
```
rohitgr7, line 262:
```suggestion
```

pytorch_lightning/utilities/parsing.py
carmocca: But what's the issue with doing it in 2 loops as in my suggestion?

krshrimali: > But what's the issue with doing it in 2 loops as in my suggestion?

Probably, `del init_args[k]` is also required (and hence not a list comprehension)?

krshrimali (approved):
LGTM, thanks Rohit!

pytorch_lightning/utilities/parsing.py
```diff
@@ -225,12 +226,23 @@ def save_hyperparameters(
...
+        ignore = {ignore}
+    elif isinstance(ignore, (list, tuple)):
+        ignore = {arg for arg in ignore if isinstance(arg, str)}
+
+    for k in list(init_args):
+        if k in ignore:
+            del init_args[k]
+            continue
+
+        if isinstance(init_args[k], nn.Module):
```
Borda, line 241: shall we also add an argument that the user would need to explicitly allow saving `nn.Module` otherwise drop it?

Borda (approved):

pytorch_lightning/utilities/parsing.py
rohitgr7: this will save the `nn.Modules`, but will raise a warning in that case, because of checkpointing issue. I don't think we need an argument since that will only disable the warnings I guess.

PR #12072, opened 2022-02-23
Remove Accelerator.parallel_device_ids and deprecate Trainer.data_parallel_device_ids

### LINKED ISSUE #12051: Remove`AcceleratorConnector.parallel_device_ids` and deprecate `Trainer.data_parallel_device_ids`
## Proposed refactor
Remove the property `AcceleratorConnector.parallel_device_ids` and migrate its usage
https://github.com/PyTorchLightning/pytorch-lightning/blob/550d3a640d1bfeb731a16b2996d3160f0f7eb071/pytorch_lightning/trainer/connectors/accelerator_connector.py#L830-L832 

Deprecate the property `Trainer.parallel_device_ids` 
https://github.com/PyTorchLightning/pytorch-lightning/blob/550d3a640d1bfeb731a16b2996d3160f0f7eb071/pytorch_lightning/trainer/trainer.py#L2043-L2047

### Alternative
`parallel_device_ids` are meant to return GPU device indexes. You can use the logic below
```
if isinstance(Trainer.accelerator, GPUAccelerator):
  if isinstance(Trainer.strategy, ParallelStrategy):
    return [device.index for device in Trainer.strategy.parallel_devices]
  elif isinstance(Trainer.strategy, SingleDeviceStrategy):
    return [0]
return []
```

### Motivation
Part of the follow ups in #11449 are to deprecate unused properties in accelerator_connector.
The accelerator properties are not meant to be public, migrated its internal usage. The only internal usage in `logger_connector` is planned to remove in v1.6.  
https://github.com/PyTorchLightning/pytorch-lightning/blob/bc191af178474eccdd912059f728cb4db22dd0d8/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py#L226

The `Trainer.data_parallel_device_ids` were also deprecated in favor of getting the GPU device indexes from `Trainer.strategy.parallel_devices` directly. The only Internal usage is planned to remove in v1.6, other usages are in tests.

https://github.com/PyTorchLightning/pytorch-lightning/blob/bc191af178474eccdd912059f728cb4db22dd0d8/pytorch_lightning/callbacks/gpu_stats_monitor.py#L136

### Pitch

https://github.com/PyTorchLightning/pytorch-lightning/blob/550d3a640d1bfeb731a16b2996d3160f0f7eb071/pytorch_lightning/trainer/trainer.py#L2043-L2047

```
@property
def data_parallel_device_ids(self) -> Optional[List[int]]:
    rank_zero_deprecation(
            "The `Trainer.data_parallel_device_ids` is deprecated in v1.6."
    )
    if not isinstance(self.accelerator, GPUAccelerator):
        return None
    if isinstance(self.strategy, ParallelStrategy):
        return [i for i in range(len(self.strategy.parallel_devices))]
    elif isinstance(self.strategy, SingleDeviceStrategy):
        return [0] 
    return None
```
https://github.com/PyTorchLightning/pytorch-lightning/blob/4c4b9d540f065c41ad288a78110cecbf45b96409/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py#L226

```
parallel_device_ids = []
if isinstance(self.trainer.accelerator, GPUAccelerator):
    if isinstance(self.trainer.strategy, ParallelStrategy):
        parallel_device_ids = [i for i in range(len(self.trainer.strategy.parallel_devices))]
    elif isinstance(self.strategy, SingleDeviceStrategy):
       parallel_device_ids = [0]
 if gpu_id in parallel_device_ids:
```

cc @justusschock @awaelchli @rohitgr7 @tchaton @akihironitta

DuYicong515:
## What does this PR do?

Fixes #12051 (Motivation and alternatives listed in the issue)
Part of #11449: deprecate unused properties in accelerator_connector

N/A

## Before submitting

- [N/A] Did you make sure to **update the documentation** with your changes? (if necessary)

## PR review

- [N/A] Add labels and milestones (and optionally projects) to the PR so it can be classified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -725,14 +725,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed public attribute `sync_batchnorm` from strategies ([#11754](https://github.com/PyTorchLightning/pytorch-lightning/pull/11754))
 
 
-<<<<<<< HEAD
 - Removed `AcceleratorConnector.root_gpu` property ([#12262](https://github.com/PyTorchLightning/pytorch-lightning/pull/12262))
-=======
-- Removed deprecated `AcceleratorConnector.parallel_device_ids` property ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
->>>>>>> c3e41378b (rebase)
 
 
 - Removed `AcceleratorConnector.num_gpus` property ([#12384](https://github.com/PyTorchLightning/pytorch-lightning/pull/12384))
+
+
+- Removed `AcceleratorConnector.parallel_device_ids` property ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -368,6 +368,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The strategies that support `sync_batchnorm` now only apply it when fitting ([#11919](https://github.com/PyTorchLightning/pytorch-lightning/pull/11919))
 
 
+- Changed `Trainer.data_parallel_device_ids` to return a list of GPU device indexes ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
+
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -533,6 +533,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated `Trainer.num_gpus` in favor of `Trainer.num_devices` when GPU is used ([#12384](https://github.com/PyTorchLightning/pytorch-lightning/pull/12384))
 
 
+- Deprecated `Trainer.data_parallel_device_ids` in favor of `Trainer.device_ids` ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
+
+
 ### Removed
 
 - Removed deprecated parameter `method` in `pytorch_lightning.utilities.model_helpers.is_overridden` ([#10507](https://github.com/PyTorchLightning/pytorch-lightning/pull/10507))
@@ -720,7 +723,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed public attribute `sync_batchnorm` from strategies ([#11754](https://github.com/PyTorchLightning/pytorch-lightning/pull/11754))
 
 
+<<<<<<< HEAD
 - Removed `AcceleratorConnector.root_gpu` property ([#12262](https://github.com/PyTorchLightning/pytorch-lightning/pull/12262))
+=======
+- Removed deprecated `AcceleratorConnector.parallel_device_ids` property ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
+>>>>>>> c3e41378b (rebase)
 
 
 - Removed `AcceleratorConnector.num_gpus` property ([#12384](https://github.com/PyTorchLightning/pytorch-lightning/pull/12384))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -538,7 +538,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated `Trainer.num_processes` in favor of `Trainer.num_devices` ([#12388](https://github.com/PyTorchLightning/pytorch-lightning/pull/12388))
 
 
-- Deprecated `Trainer.data_parallel_device_ids` in favor of `Trainer.device_ids` ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))  
+- Deprecated `Trainer.data_parallel_device_ids` in favor of `Trainer.device_ids` ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
 
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -368,9 +368,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The strategies that support `sync_batchnorm` now only apply it when fitting ([#11919](https://github.com/PyTorchLightning/pytorch-lightning/pull/11919))
 
 
-- Changed `Trainer.data_parallel_device_ids` to return a list of GPU device indexes ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
-
-
 ### Deprecated
 
 - Deprecated `training_type_plugin` property in favor of `strategy` in `Trainer` and updated the references ([#11141](https://github.com/PyTorchLightning/pytorch-lightning/pull/11141))
--- a/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
+++ b/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
@@ -19,7 +19,6 @@ import pytorch_lightning as pl
 from pytorch_lightning.accelerators import GPUAccelerator
 from pytorch_lightning.loggers import LightningLoggerBase, TensorBoardLogger
 from pytorch_lightning.plugins.environments.slurm_environment import SLURMEnvironment
-from pytorch_lightning.strategies import ParallelStrategy, SingleDeviceStrategy
 from pytorch_lightning.trainer.connectors.logger_connector.result import _METRICS, _OUT_DICT, _PBAR_DICT
 from pytorch_lightning.trainer.states import RunningStage
 from pytorch_lightning.utilities import memory
@@ -222,13 +221,7 @@ class LoggerConnector:
                 self.trainer.lightning_module.log(key, mem, prog_bar=False, logger=True)
             else:
                 gpu_id = int(key.split("/")[0].split(":")[1])
-                parallel_device_ids = []
-                if isinstance(self.trainer.accelerator, GPUAccelerator):
-                    if isinstance(self.trainer.strategy, ParallelStrategy):
-                        parallel_device_ids = [i for i in range(len(self.trainer.strategy.parallel_devices))]
-                    elif isinstance(self.strategy, SingleDeviceStrategy):
-                        parallel_device_ids = [0]
-                if gpu_id in parallel_device_ids:
+                if gpu_id in self.trainer.data_parallel_device_ids:
                     self.trainer.lightning_module.log(
                         key, mem, prog_bar=False, logger=True, on_step=True, on_epoch=False
                     )
--- a/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
+++ b/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
@@ -221,7 +221,13 @@ class LoggerConnector:
                 self.trainer.lightning_module.log(key, mem, prog_bar=False, logger=True)
             else:
                 gpu_id = int(key.split("/")[0].split(":")[1])
-                if gpu_id in self.trainer._accelerator_connector.parallel_device_ids:
+                parallel_device_ids = []
+                if isinstance(self.trainer.accelerator, GPUAccelerator):
+                    if isinstance(self.trainer.strategy, ParallelStrategy):
+                        parallel_device_ids = [i for i in range(len(self.trainer.strategy.parallel_devices))]
+                    elif isinstance(self.strategy, SingleDeviceStrategy):
+                        parallel_device_ids = [0]
+                if gpu_id in parallel_device_ids:
                     self.trainer.lightning_module.log(
                         key, mem, prog_bar=False, logger=True, on_step=True, on_epoch=False
                     )
--- a/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
+++ b/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
@@ -221,7 +221,7 @@ class LoggerConnector:
                 self.trainer.lightning_module.log(key, mem, prog_bar=False, logger=True)
             else:
                 gpu_id = int(key.split("/")[0].split(":")[1])
-                if gpu_id in self.trainer.data_parallel_device_ids:
+                if self.trainer.data_parallel_device_ids and gpu_id in self.trainer.data_parallel_device_ids:
                     self.trainer.lightning_module.log(
                         key, mem, prog_bar=False, logger=True, on_step=True, on_epoch=False
                     )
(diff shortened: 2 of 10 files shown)


pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -223,7 +223,13 @@ def _log_gpus_metrics(self) -> None:
                 self.trainer.lightning_module.log(key, mem, prog_bar=False, logger=True)
             else:
                 gpu_id = int(key.split("/")[0].split(":")[1])
-                if gpu_id in self.trainer._accelerator_connector.parallel_device_ids:
+                parallel_device_ids = []
+                if isinstance(self.trainer.accelerator, GPUAccelerator):
+                    if isinstance(self.trainer.strategy, ParallelStrategy):
+                        parallel_device_ids = [i for i in range(len(self.trainer.strategy.parallel_devices))]
+                    elif isinstance(self.strategy, SingleDeviceStrategy):
+                        parallel_device_ids = [0]
```
four4fish, line 232: trainer.data_parallel_device_ids is the same logic as this. Let's use `trainer.data_parallel_device_ids`. 
This function is already deprecated, can be removed after 1.6 release.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2042,9 +2042,13 @@ def devices(self) -> Optional[Union[List[int], str, int]]:
 
     @property
     def data_parallel_device_ids(self) -> Optional[List[int]]:
-        return (
-            self._accelerator_connector.parallel_device_ids if self._accelerator_connector.parallel_device_ids else None
-        )
+        rank_zero_deprecation("`Trainer.data_parallel_device_ids` was deprecated in v1.6.")
+        if isinstance(self.accelerator, GPUAccelerator):
+            if isinstance(self.strategy, ParallelStrategy):
+                return [i for i in range(len(self.strategy.parallel_devices))]
```
four4fish, line 2048: Reviewers: here we can not use self.strategy.parallel_devices because self.strategy.parallel_devices could be a list of torch.devices

This is the existing logic, the correct one should be get ids from self.strategy.parallel_devices instead of `[i for i in range(len(self.strategy.parallel_devices))]`. Should we fix it? It will be a minor breaking change, and we have no active usage of this method (other than two gpu_log functions will be removed in 1.7)

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
DuYicong515: Sounds good

pytorch_lightning/trainer/trainer.py
carmocca: IMO yes

pytorch_lightning/trainer/trainer.py
```diff
@@ -2042,9 +2042,13 @@ def devices(self) -> Optional[Union[List[int], str, int]]:
...
+            elif isinstance(self.strategy, SingleDeviceStrategy):
+                return [0]
+        return None
```
carmocca, line 2051: I think this should return an empty list in the base case

pytorch_lightning/trainer/trainer.py
```diff
@@ -2042,9 +2042,13 @@ def devices(self) -> Optional[Union[List[int], str, int]]:
 
     @property
     def data_parallel_device_ids(self) -> Optional[List[int]]:
-        return (
-            self._accelerator_connector.parallel_device_ids if self._accelerator_connector.parallel_device_ids else None
-        )
+        rank_zero_deprecation("`Trainer.data_parallel_device_ids` was deprecated in v1.6.")
```
carmocca, line 2045: "and will be removed in v1.8"

Also suggest the updated alternative

pytorch_lightning/trainer/trainer.py
DuYicong515: The original function returns None  that's why I kept it the same. I'll change to empty list instead.

pytorch_lightning/trainer/trainer.py
DuYicong515: Changed the logic to use device.index for each device in parallel_devices

pytorch_lightning/trainer/trainer.py
DuYicong515: Do we still want to expose `parallel_device_ids` to users? There's no current alternative function yet. cc @four4fish

four4fish: Ideally user should just get parallel_devices instead of parallel_devices_id I think. Id can get from parallel_devices easily, it's duplicated with parallel_devices
I think after deprecate num_gpus, num_tpus and etc, we should update the device property in trainer which return the correct device for all strategies: single device return the root_device, parallel strategy return parallel_devices

Reviewers: should we provide both `devices` and `devices_id` in trainer api?

four4fish (approved):
LGTM! Thanks for working on this!

pytorch_lightning/trainer/trainer.py
```diff
@@ -2043,10 +2043,14 @@ def devices(self) -> Optional[Union[List[int], str, int]]:
...
-        return (
-            self._accelerator_connector.parallel_device_ids if self._accelerator_connector.parallel_device_ids else None
-        )
+    def data_parallel_device_ids(self) -> List[int]:
+        rank_zero_deprecation("`Trainer.data_parallel_device_ids` was deprecated in v1.6 and will be removed in v1.8.")
+        if isinstance(self.accelerator, GPUAccelerator):
+            if isinstance(self.strategy, ParallelStrategy):
+                return [device.index for device in self.strategy.parallel_devices]
+            elif isinstance(self.strategy, SingleDeviceStrategy):
+                return [0]
```
ananthsub, line 2052: why is this hardcoded to 0? someone might be running on a different device index on a host. 

use this instead on the strategy's `root_device`
https://github.com/PyTorchLightning/pytorch-lightning/blob/1b107c58928e75b42155aef4a3695b5ee243c276/pytorch_lightning/accelerators/gpu.py#L117

DuYicong515: Thanks for the suggestion, addressed it. I also added one test case `tests/utilities/test_cli.py::test_parse_args_parsing_gpus` to make sure it returns the right index on SingleDeviceStrategy

ananthsub (approved):

tests/callbacks/test_gpu_stats_monitor.py
```diff
@@ -84,7 +87,11 @@ def test_gpu_stats_monitor_no_queries(tmpdir):
         devices=1,
         callbacks=[gpu_stats],
     )
-    with mock.patch("pytorch_lightning.loggers.tensorboard.TensorBoardLogger.log_metrics") as log_metrics_mock:
+    with mock.patch(
+        "pytorch_lightning.loggers.tensorboard.TensorBoardLogger.log_metrics"
+    ) as log_metrics_mock, pytest.deprecated_call(
+        match="Trainer.data_parallel_device_ids` was deprecated in v1.6 and will be removed in v1.8."
```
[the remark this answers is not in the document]
DuYicong515: https://github.com/PyTorchLightning/pytorch-lightning/pull/12072#discussion_r813308009

This was still getting triggered, I previously migrated this callsite as well. But since it's going to be removed in v1.6, I kept using `Trainer.data_parallel_device_ids`. Any suggestions?

tests/models/test_gpu.py
```diff
@@ -236,14 +237,20 @@ def test_parse_gpu_returns_none_when_no_devices_are_available(mocked_device_coun
...
     sanitizing the gpus as only one of the GPUs is visible."""
     trainer = Trainer(gpus=gpus)
     assert isinstance(trainer._accelerator_connector.cluster_environment, TorchElasticEnvironment)
-    assert trainer.data_parallel_device_ids == device_parser.parse_gpu_ids(gpus)
     assert trainer.gpus == gpus
+    if rank_zero_only.rank == 0:
+        with pytest.deprecated_call(
+            match="Trainer.data_parallel_device_ids` was deprecated in v1.6 and will be removed in v1.8."
+        ):
+            assert (trainer.data_parallel_device_ids or None) == device_parser.parse_gpu_ids(gpus)
```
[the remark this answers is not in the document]
DuYicong515: It goes into this branch when gpus="0"

tests/callbacks/test_gpu_stats_monitor.py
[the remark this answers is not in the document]
awaelchli: No, it's not going to be removed in 1.6. It's going to be removed in 1.8 which means until then, if you don't change anything, the user would see this warning and they can't do anything about it, because the logger connector triggers it internally. We can't have that. 

What's clear is that we still need  a way to get the device ids. So my suggestion is to keep the method in the connector, deprecate the one on the trainer and then call the connector directly.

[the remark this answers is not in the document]
DuYicong515: That make sense! I feel the primary purpose of this diff is to remove `Accelerator.parallel_device_ids` as it's not intended to be a public API and internal usages can be moved to derive from `Trainer.strategy.parallel_devices`. I would prefer still remove this property.

Maybe I can hold-off this PR until I introduce `Trainer.device_ids` in #12126? Then `Trainer.data_parallel_device_ids` will be an easy migration. 

Thoughts?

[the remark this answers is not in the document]
awaelchli: > Maybe I can hold-off this PR until I introduce Trainer.device_ids in https://github.com/PyTorchLightning/pytorch-lightning/issues/12126? Then

Yes, this would be my preference.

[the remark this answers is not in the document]
four4fish: Yeah make sense, having trainer.device_ids is the desired state for 1.6, we can have that first then deprecate the rest
Thanks @awaelchli for suggesting. Thank you @DuYicong515 for working on this :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -2056,10 +2056,12 @@ def devices(self) -> int:
...
     @property
-    def data_parallel_device_ids(self) -> Optional[List[int]]:
-        return (
-            self._accelerator_connector.parallel_device_ids if self._accelerator_connector.parallel_device_ids else None
+    def data_parallel_device_ids(self) -> List[int]:
+        rank_zero_deprecation(
+            "`Trainer.data_parallel_device_ids` was deprecated in v1.6 and will be removed in v1.8."
+            " Please use `Trainer.device_ids` instead."
         )
+        return self.device_ids if isinstance(self.accelerator, GPUAccelerator) else []
```
ananthsub, line 2064: why doesn't this always return `self.device_ids` ? why is the GPU accelerator check needed?

DuYicong515: I think `data_parallel_device_ids` is originally intended to return the ids of GPU devices not device_ids? 

https://github.com/PyTorchLightning/pytorch-lightning/blob/f2d6a855af0c130f81e2c04506ba03a061eba924/pytorch_lightning/trainer/connectors/accelerator_connector.py#L836-L838.

Also its existing usage were all related to GPUs.

tests/callbacks/test_gpu_stats_monitor.py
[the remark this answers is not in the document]
DuYicong515: Hi @awaelchli, I just updated the PR, do you mind to recheck the changes here? Thanks!

pytorch_lightning/trainer/trainer.py
awaelchli, line 2064:
```suggestion
        return self.device_ids if isinstance(self.accelerator, GPUAccelerator) else None
```
a deprecation message only makes sense if we don't change the behavior.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2056,10 +2056,12 @@ def devices(self) -> int:
         return self.num_devices
 
     @property
-    def data_parallel_device_ids(self) -> Optional[List[int]]:
-        return (
-            self._accelerator_connector.parallel_device_ids if self._accelerator_connector.parallel_device_ids else None
+    def data_parallel_device_ids(self) -> List[int]:
```
awaelchli, line 2059:
```suggestion
    def data_parallel_device_ids(self) -> Optional[List[int]]:
```

pytorch_lightning/trainer/trainer.py
awaelchli: > why doesn't this always return self.device_ids ? why is the GPU accelerator check needed?

I think we simply want to keep the behavior during the deprecation phase so everybody who relies on it can keep using it.

tests/deprecated_api/test_remove_1-8.py
```diff
@@ -887,3 +887,30 @@ def test_trainer_config_device_ids():
...
     ):
         trainer.devices == 2
+
+
+@pytest.mark.parametrize(
+    ["trainer_kwargs", "expected_data_parallel_device_ids"],
+    [
+        ({}, []),
+        ({"devices": 1}, []),
+        ({"devices": "1"}, []),
```
awaelchli, line 897: the one on the trainer returns None, see my comment above.

awaelchli (changes requested):

pytorch_lightning/trainer/trainer.py
DuYicong515: Make sense. I've changed non-GPU case to keep returning None.

CHANGELOG.md
```diff
@@ -368,6 +368,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The strategies that support `sync_batchnorm` now only apply it when fitting ([#11919](https://github.com/PyTorchLightning/pytorch-lightning/pull/11919))
 
 
+- Changed `Trainer.data_parallel_device_ids` to return a list of GPU device indexes ([#12072](https://github.com/PyTorchLightning/pytorch-lightning/pull/12072))
+
```
awaelchli, line 372:
```suggestion
```
The trainer arg remains unchanged, only deprecated.

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
awaelchli: thanks

CHANGELOG.md
DuYicong515: Removed

ananthsub (approved):

kaushikb11 (approved):

tests/trainer/test_trainer_cli.py
```diff
@@ -177,8 +177,7 @@ def test_argparse_args_parsing_devices(cli_args, expected_parsed, expected_devic
         args = Trainer.parse_argparser(parser)
 
     assert args.devices == expected_parsed
-    trainer = Trainer.from_argparse_args(args)
-    assert trainer.data_parallel_device_ids == expected_device_ids
+    assert Trainer.from_argparse_args(args)
```
rohitgr7, line 180:
```suggestion
    trainer = Trainer.from_argparse_args(args)
    assert trainer.device_ids == expected_device_ids
```

rohitgr7 (approved):

PR #12083, opened 2022-02-24
Only allow one value for each plugin type in `plugins` flag

### LINKED ISSUE #11449: Rewrite Accelerator_connector and follow up tasks

four4fish:
## What does this PR do?

Fixes #11449 follow up item : check there is only 1 instance of each type at most in plugin flag

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/trainer/connectors/accelerator_connector.py
+++ b/pytorch_lightning/trainer/connectors/accelerator_connector.py
@@ -14,6 +14,7 @@
 
 import logging
 import os
+from collections import Counter
 from typing import List, Optional, Union
 
 import torch
@@ -304,13 +305,7 @@ class AcceleratorConnector:
             self._precision_flag = precision
 
         if plugins:
-            plugins_flags_count = {
-                "strategy": 0,
-                "precision_plugin": 0,
-                "precision": 0,
-                "checkpoint_io": 0,
-                "cluster_env": 0,
-            }
+            plugins_flags_types_list = []
             for plugin in plugins:
                 if isinstance(plugin, Strategy) or isinstance(plugin, str) and plugin in self._registered_strategies:
                     self._strategy_flag = plugin
@@ -318,19 +313,17 @@ class AcceleratorConnector:
                         f"Passing {plugin} `strategy` to the `plugins` flag in Trainer has been deprecated"
                         f" in v1.5 and will be removed in v1.7. Use `Trainer(strategy={plugin})` instead."
                     )
-                    plugins_flags_count["strategy"] += 1
+                    plugins_flags_types_list.append("Strategy")
 
                 elif isinstance(plugin, PrecisionPlugin):
                     self._precision_plugin_flag = plugin
-                    plugins_flags_count["precision_plugin"] += 1
-                elif isinstance(plugin, str) and plugin in self._precision_types:
-                    self._precision_flag = plugin
-                    plugins_flags_count["precision"] += 1
+                    plugins_flags_types_list.append("PrecisionPlugin")
                 elif isinstance(plugin, CheckpointIO):
                     self.checkpoint_io = plugin
-                    plugins_flags_count["checkpoint_io"] += 1
+                    plugins_flags_types_list.append("CheckpointIO")
                 elif isinstance(plugin, ClusterEnvironment):
                     self._cluster_environment_flag = plugin
+                    plugins_flags_types_list.append("ClusterEnvironment")
                 elif isinstance(plugin, LayerSync):
                     if sync_batchnorm and not isinstance(plugin, NativeSyncBatchNorm):
                         raise MisconfigurationException(
@@ -338,17 +331,17 @@ class AcceleratorConnector:
                             " plugin, but this is not allowed. Choose one or the other."
                         )
                     self._layer_sync = plugin
-                    plugins_flags_count["cluster_env"] += 1
                 else:
                     raise MisconfigurationException(
-                        f"Found invalid type for plugin {plugin}. Expected a precision plugin or training strategy."
+                        f"Found invalid type for plugin {plugin}. Expected PrecisionPlugin, "
+                        "CheckpointIO plugin, ClusterEnviroment plugin or a training strategy."
                     )
 
-            duplicated_plugin_key = [k for k, v in plugins_flags_count.items() if v > 1]
+            duplicated_plugin_key = [k for k, v in Counter(plugins_flags_types_list).items() if v > 1]
             if duplicated_plugin_key:
                 raise MisconfigurationException(
                     f"Received multiple values for {', '.join(duplicated_plugin_key)} flags in `plugins`."
-                    " Expected one value for each type at most"
+                    " Expected one value for each type at most."
                 )
 
         # handle the case when the user passes in a strategy instance which has an accelerator, precision,
--- a/pytorch_lightning/trainer/connectors/accelerator_connector.py
+++ b/pytorch_lightning/trainer/connectors/accelerator_connector.py
@@ -335,7 +335,7 @@ class AcceleratorConnector:
                 else:
                     raise MisconfigurationException(
                         f"Found invalid type for plugin {plugin}. Expected PrecisionPlugin, "
-                        "CheckpointIO plugin, ClusterEnviroment plugin or a training strategy."
+                        "CheckpointIO plugin, ClusterEnviroment plugin or a Strategy."
                     )
 
             duplicated_plugin_key = [k for k, v in Counter(plugins_flags_types_list).items() if v > 1]
--- a/pytorch_lightning/trainer/connectors/accelerator_connector.py
+++ b/pytorch_lightning/trainer/connectors/accelerator_connector.py
@@ -305,7 +305,7 @@ class AcceleratorConnector:
             self._precision_flag = precision
 
         if plugins:
-            plugins_flags_types_list = []
+            plugins_flags_types = Counter()
             for plugin in plugins:
                 if isinstance(plugin, Strategy) or isinstance(plugin, str) and plugin in self._registered_strategies:
                     self._strategy_flag = plugin
@@ -313,17 +313,17 @@ class AcceleratorConnector:
                         f"Passing {plugin} `strategy` to the `plugins` flag in Trainer has been deprecated"
                         f" in v1.5 and will be removed in v1.7. Use `Trainer(strategy={plugin})` instead."
                     )
-                    plugins_flags_types_list.append(Strategy.__name__)
+                    plugins_flags_types[Strategy.__name__] += 1
 
                 elif isinstance(plugin, PrecisionPlugin):
                     self._precision_plugin_flag = plugin
-                    plugins_flags_types_list.append(PrecisionPlugin.__name__)
+                    plugins_flags_types[PrecisionPlugin.__name__] += 1
                 elif isinstance(plugin, CheckpointIO):
                     self.checkpoint_io = plugin
-                    plugins_flags_types_list.append(CheckpointIO.__name__)
+                    plugins_flags_types[CheckpointIO.__name__] += 1
                 elif isinstance(plugin, ClusterEnvironment):
                     self._cluster_environment_flag = plugin
-                    plugins_flags_types_list.append("ClusterEnvironment")
+                    plugins_flags_types[ClusterEnvironment.__name__] += 1
                 elif isinstance(plugin, LayerSync):
                     if sync_batchnorm and not isinstance(plugin, NativeSyncBatchNorm):
                         raise MisconfigurationException(
@@ -331,14 +331,14 @@ class AcceleratorConnector:
                             " plugin, but this is not allowed. Choose one or the other."
                         )
                     self._layer_sync = plugin
-                    plugins_flags_types_list.append(ClusterEnvironment.__name__)
+                    plugins_flags_types[NativeSyncBatchNorm.__name__] += 1
                 else:
                     raise MisconfigurationException(
                         f"Found invalid type for plugin {plugin}. Expected PrecisionPlugin, "
                         "CheckpointIO plugin, ClusterEnviroment plugin or a Strategy."
                     )
 
-            duplicated_plugin_key = [k for k, v in Counter(plugins_flags_types_list).items() if v > 1]
+            duplicated_plugin_key = [k for k, v in plugins_flags_types.items() if v > 1]
             if duplicated_plugin_key:
                 raise MisconfigurationException(
                     f"Received multiple values for {', '.join(duplicated_plugin_key)} flags in `plugins`."
--- a/pytorch_lightning/trainer/connectors/accelerator_connector.py
+++ b/pytorch_lightning/trainer/connectors/accelerator_connector.py
@@ -348,7 +348,7 @@ class AcceleratorConnector:
             if duplicated_plugin_key:
                 raise MisconfigurationException(
                     f"Received multiple values for {', '.join(duplicated_plugin_key)} flags in `plugins`."
-                    " Expected one 1 value for each type at most"
+                    " Expected one value for each type at most"
                 )
 
         # handle the case when the user passes in a strategy instance which has an accelerator, precision,
--- a/pytorch_lightning/trainer/connectors/accelerator_connector.py
+++ b/pytorch_lightning/trainer/connectors/accelerator_connector.py
@@ -313,14 +313,14 @@ class AcceleratorConnector:
                         f"Passing {plugin} `strategy` to the `plugins` flag in Trainer has been deprecated"
                         f" in v1.5 and will be removed in v1.7. Use `Trainer(strategy={plugin})` instead."
                     )
-                    plugins_flags_types_list.append("Strategy")
+                    plugins_flags_types_list.append(Strategy.__name__)
 
                 elif isinstance(plugin, PrecisionPlugin):
                     self._precision_plugin_flag = plugin
-                    plugins_flags_types_list.append("PrecisionPlugin")
+                    plugins_flags_types_list.append(PrecisionPlugin.__name__)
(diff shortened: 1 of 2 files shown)


kaushikb11:
@four4fish How is this a breaking change? It was never supported.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -295,27 +295,46 @@ def _check_config_and_set_final_flags(
...
+                "cluster_env": 0,
+            }
             for plugin in plugins:
                 if isinstance(plugin, Strategy) or isinstance(plugin, str) and plugin in self._registered_strategies:
                     self._strategy_flag = plugin
                     rank_zero_deprecation(
                         f"Passing {plugin} `strategy` to the `plugins` flag in Trainer has been deprecated"
                         f" in v1.5 and will be removed in v1.7. Use `Trainer(strategy={plugin})` instead."
                     )
+                    plugins_flags_count["strategy"] += 1
```
kaushikb11, line 312: Not a huge fan of this approach.

You could also do

```python
if self._strategy_flag is not None:
	raise MisconfigurationException("Received multiple values for strategy in `plugins` ...")

File: tests/accelerators/test_accelerator_connector.py [hunk 2]
```diff
@@ -948,3 +948,16 @@ def test_deterministic_init(deterministic):
     if deterministic:
         assert os.environ.get("CUBLAS_WORKSPACE_CONFIG") == ":4096:8"
         assert os.environ.get("HOROVOD_FUSION_THRESHOLD") == "0"
+
+
+@pytest.mark.parametrize(
+    ["plugins", "expected"],
+    [
+        ([LightningEnvironment(), SLURMEnvironment()], "cluster_env"),
```
rohitgr7, line 956: can you add a check for IO as well?

pytorch_lightning/trainer/connectors/accelerator_connector.py
four4fish: Not really.. the self._strategy_flag could be set by strategy flag at L242-L276
The duplication of plugin[strategy] and strategy flag has been handled in L242-L276 as well

four4fish: Same for precision_plugin_flag, it could be set above already. We will need a seperate `count` variable to track whether plugin has multiple same type. 
Any suggestion @kaushikb11 :)?

four4fish:
> @four4fish How is this a breaking change? It was never supported.

@kaushikb11 If user set two checkpointIO in plugins before, we used the later one instead of throw exception. Does this considered breaking change?

pytorch_lightning/trainer/connectors/accelerator_connector.py
four4fish: I like this way because we can tell user all the duplications at one go

kaushikb11:
> @kaushikb11 If user set two checkpointIO in plugins before, we used the later one instead of throw exception. Does this considered breaking change?

IMO, we should throw exception for this case as well.

four4fish:
> > @kaushikb11 If user set two checkpointIO in plugins before, we used the later one instead of throw exception. Does this considered breaking change?
> 
> IMO, we should throw exception for this case as well.

Checkpoint io duplication check is included in this PR. Just a example to explain why this could be a breaking change

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -295,27 +295,46 @@ def _check_config_and_set_final_flags(
...
 
                 elif isinstance(plugin, PrecisionPlugin):
                     self._precision_plugin_flag = plugin
+                    plugins_flags_count["precision_plugin"] += 1
                 elif isinstance(plugin, str) and plugin in self._precision_types:
```
ananthsub, line 317: when was passing `precision` to the `plugins` argument allowed? this is completely redundant with the `precision` flag on the Trainer constructor. I don't think this should be allowed

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -295,27 +295,46 @@ def _check_config_and_set_final_flags(
...
                 else:
                     raise MisconfigurationException(
                         f"Found invalid type for plugin {plugin}. Expected a precision plugin or training strategy."
                     )
 
+            duplicated_plugin_key = [k for k, v in plugins_flags_count.items() if v > 1]
+            if duplicated_plugin_key:
+                raise MisconfigurationException(
+                    f"Received multiple values for {', '.join(duplicated_plugin_key)} flags in `plugins`."
+                    " Expected one value for each type at most"
```
ananthsub, line 335:
```suggestion
                    " Expected one value for each type at most."
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -295,27 +295,46 @@ def _check_config_and_set_final_flags(
...
+                    plugins_flags_count["cluster_env"] += 1
                 else:
                     raise MisconfigurationException(
                         f"Found invalid type for plugin {plugin}. Expected a precision plugin or training strategy."
                     )
 
+            duplicated_plugin_key = [k for k, v in plugins_flags_count.items() if v > 1]
+            if duplicated_plugin_key:
+                raise MisconfigurationException(
+                    f"Received multiple values for {', '.join(duplicated_plugin_key)} flags in `plugins`."
```
ananthsub, line 334: the flag keys here should be the base interface name. for instance, users know ClusterEnvironment, not "cluster_env"

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -295,27 +295,46 @@ def _check_config_and_set_final_flags(
             self._precision_flag = precision
 
         if plugins:
+            plugins_flags_count = {
+                "strategy": 0,
+                "precision_plugin": 0,
+                "precision": 0,
+                "checkpoint_io": 0,
+                "cluster_env": 0,
```
ananthsub, line 303: the keys here should be the base interface name. for instance, users know ClusterEnvironment, not "cluster_env" . this will improve the error message below

pytorch_lightning/trainer/connectors/accelerator_connector.py
ananthsub: But that would be a custom precision plug-in, right?

Was this supported?
```py
Trainer(plugins=["bf16"])
```

awaelchli: No it is not. On 1.5 it errors. 
Sorry, I only saw `elif isinstance(plugin, PrecisionPlugin):` and your question, but you marked the line below which checks for string. My bad.

awaelchli: In 1.5 we have this code:

```py
            if isinstance(plug, str) and plug in TrainingTypePluginsRegistry:
                if training_type is None:
                    training_type = TrainingTypePluginsRegistry.get(plug)
                else:
                        "You can only specify one precision and one training type plugin."
                        " Found more than 1 training type plugin:"
                        f' {TrainingTypePluginsRegistry[plug]["plugin"]} registered to {plug}'
                    )
            if isinstance(plug, str):
                # Reset the distributed type as the user has overridden training type
                # via the plugins argument
                self._distrib_type = None
                self.set_distributed_mode(plug)
```

which attempts to handle strings for training type plugins but there is no error handling for unsupported strings.

four4fish: Make sense. I will remove the precision str support and throw exceptions for non-supported strs

pytorch_lightning/trainer/connectors/accelerator_connector.py
tchaton: Yes. would be cleaner and remove the need to hardcode them.

pytorch_lightning/trainer/connectors/accelerator_connector.py
four4fish: Removed

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -295,27 +295,46 @@ def _check_config_and_set_final_flags(
             self._precision_flag = precision
 
         if plugins:
+            plugins_flags_count = {
```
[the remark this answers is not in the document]
four4fish: updated to use collections.Counter

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -301,27 +302,38 @@ def _check_config_and_set_final_flags(
             self._precision_flag = precision
 
         if plugins:
+            plugins_flags_types_list = []
```
carmocca, line 305: This can be the counter instance directly

carmocca (approved):

CHANGELOG.md
```diff
@@ -326,6 +326,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Disable loading dataloades if corresponding `limit_batches=0` ([#11576](https://github.com/PyTorchLightning/pytorch-lightning/pull/11576))
 
 
+- Only allow one value for each type in `plugins` trainer flag ([#12083](https://github.com/PyTorchLightning/pytorch-lightning/pull/12083))
```
ananthsub, line 329: does this need a changelog entry since this was already available before?

ananthsub (approved):

CHANGELOG.md
four4fish: I think this is new, it wasn't checked before

awaelchli: This was checked in 1.5.10 and raised an error. 
I verified with

```py
    trainer = Trainer(
        accelerator="cpu",
        devices=2,
        strategy="ddp",
        plugins=[DDPSpawnPlugin()]
    )
```

```py
    trainer = Trainer(
        accelerator="cpu",
        devices=2,
        plugins=[DDPSpawnPlugin(), DDPPlugin()]
    )
```

```py
    trainer = Trainer(
        accelerator="cpu",
        devices=2,
         plugins=[PrecisionPlugin(), NativeMixedPrecisionPlugin(16, device="cpu")]

```


So as far as I can see, this is fixing a AcceleratorConnector bug on master.

pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli:
```suggestion
                    f"Received multiple entries for {', '.join(duplicated_plugin_key)} in `plugins`."
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -301,27 +302,38 @@ def _check_config_and_set_final_flags(
...
                     self.checkpoint_io = plugin
+                    plugins_flags_types_list.append(CheckpointIO.__name__)
                 elif isinstance(plugin, ClusterEnvironment):
                     self._cluster_environment_flag = plugin
+                    plugins_flags_types_list.append(ClusterEnvironment.__name__)
                 else:
                     raise MisconfigurationException(
-                        f"Found invalid type for plugin {plugin}. Expected a precision plugin or training strategy."
+                        f"Found invalid type for plugin {plugin}. Expected PrecisionPlugin, "
+                        "CheckpointIO plugin, ClusterEnviroment plugin or a training strategy."
```
awaelchli, line 327:
```suggestion
                        "CheckpointIO plugin, ClusterEnviroment plugin or a Strategy."
```

awaelchli (approved):
While reviewing the PR, also found this malformatted error message:


```
pytorch_lightning.utilities.exceptions.MisconfigurationException: You have passed Trainer(strategy=ddp) and you can only specify one strategy, but you have passed <pytorch_lightning.plugins.training_type.ddp_spawn.DDPSpawnPlugin object at 0x7fc2bb458310> as a plugin.

```

it can be triggered by 
```py
Trainer(
        strategy="ddp",
        plugins=[DDPSpawnPlugin()]
)
```

for example. Would be nice follow up :)

CHANGELOG.md
awaelchli: So I recommend to remove the changelog entry.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -304,34 +305,46 @@ def _check_config_and_set_final_flags(
...
                             f"You set `Trainer(sync_batchnorm=True)` and provided a `{plugin.__class__.__name__}`"
                             " plugin, but this is not allowed. Choose one or the other."
                         )
                     self._layer_sync = plugin
+                    plugins_flags_types[NativeSyncBatchNorm.__name__] += 1
                 else:
                     raise MisconfigurationException(
-                        f"Found invalid type for plugin {plugin}. Expected a precision plugin or training strategy."
+                        f"Found invalid type for plugin {plugin}. Expected PrecisionPlugin, "
+                        "CheckpointIO plugin, ClusterEnviroment plugin or a Strategy."
```
ananthsub, line 338: error is missing LayerSync
```suggestion
                        f"Found invalid type for plugin {plugin}. Expected one of: PrecisionPlugin, "
                        "CheckpointIO, ClusterEnviroment, LayerSync, or Strategy."
```

ananthsub (approved):

awaelchli (approved):

Issue #3832, opened 2020-10-03
Mismatch between on_validation_epoch_end and on_train_epoch_end


pbmstrk:
## 🐛 Bug

The behaviour of the callback hooks `on_train_epoch_end` and `on_validation_epoch_end` do not match. While `on_train_epoch_end` can access metrics of the same epoch from the `validation_epoch_end` method the opposite is not true. More concretely, when accessing `trainer.callback_metrics` from `on_validation_epoch_end` the values logged in `training_epoch_end` correspond to the previous epoch rather than the current one.

### To Reproduce

On master:

```
class LitModel(pl.LightningModule):

    def __init__(self):
        super().__init__()
        self.layer_1 = torch.nn.Linear(28 * 28, 128)
        self.layer_2 = torch.nn.Linear(128, 10)

    def forward(self, x):
        x = x.view(x.size(0), -1)
        x = self.layer_1(x)
        x = F.relu(x)
        x = self.layer_2(x)
        return x

    def configure_optimizers(self):
        optimizer = torch.optim.Adam(self.parameters(), lr=1e-3)
        return optimizer

    def training_step(self, batch, batch_idx):
        x, y = batch
        y_hat = self(x)
        loss = F.cross_entropy(y_hat, y)
        self.log('loss', loss)
        return {'loss': loss}

    def validation_step(self, batch, batch_idx):
        x, y = batch
        y_hat = self(x)
        loss = F.cross_entropy(y_hat, y)
        self.log('val_loss', loss)
        return {"val_loss": loss}

    def training_epoch_end(self, outputs):
        loss_val = torch.stack([x['loss'] for x in outputs]).mean()
        self.log('train_loss_epoch', loss_val)

    def validation_epoch_end(self, outputs):
        loss_val = torch.stack([x['val_loss'] for x in outputs]).mean()
        self.log('val_loss_epoch', loss_val)

class CustomCallback(Callback):


    def on_epoch_end(self, trainer, pl_module):
        metrics = trainer.callback_metrics
        epoch = trainer.current_epoch
        print(f"Epoch {epoch}: {metrics}")

    def on_validation_epoch_end(self, trainer, pl_module):
        print(f"Val_epoch_end: {trainer.callback_metrics}")

    def on_train_epoch_end(self, trainer, pl_module):
        print(f"Train_epoch_end: {trainer.callback_metrics}")
        print("\n")


model = LitModel()

[...]

PR #12099, opened 2022-02-24
Update `validation_epoch_end` docs

AJSVB:
It took me time to understand that validation_epoch_end is called before training_epoch_end. This behavior is expected (see https://github.com/PyTorchLightning/pytorch-lightning/issues/3832), but not obvious, so I think that the documentation could be clearer

### DIFF
--- a/docs/source/common/lightning_module.rst
+++ b/docs/source/common/lightning_module.rst
@@ -398,7 +398,7 @@ Validation Epoch-level Metrics
 ==============================
 
 If you need to do something with all the outputs of each :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_step`,
-override the :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_epoch_end` method.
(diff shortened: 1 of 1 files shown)


awaelchli (approved):

tchaton (approved):
LGTM !

docs/source/common/lightning_module.rst
```diff
@@ -398,7 +398,7 @@ Validation Epoch-level Metrics
 ==============================
 
 If you need to do something with all the outputs of each :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_step`,
-override the :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_epoch_end` method.
+override the :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_epoch_end` method. Note that this method is called before  :meth:`~pytorch_lightning.core.lightning.LightningModule.training_epoch_end`.
```
carmocca, line 401:
```suggestion
override the :meth:`~pytorch_lightning.core.lightning.LightningModule.validation_epoch_end` method. Note that this method is called before :meth:`~pytorch_lightning.core.lightning.LightningModule.training_epoch_end`.
```

carmocca (approved):

PR #9052, opened 2021-08-23
3/n inter batch parallelism

tchaton:
## What does this PR do?

Blocked by https://github.com/PyTorchLightning/pytorch-lightning/pull/9047/

This PR adds remove entirely InterBatchProcessor and rely on DataFetcher. 

fixes #8867
fixes #8316

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #12100, opened 2022-02-24
Construct the hook kwargs inside each loop

carmocca:
## What does this PR do?

Same as #11466 but for training.

This encapsulates logic within each loop with the exception of hidden because they are passed from one optimizer index to the next one. I added a test for this behaviour which is unchanged from master.

Only left to do for prediction.

None

## Before submitting

## PR review

cc @borda @justusschock @awaelchli @rohitgr7 @carmocca @ananthsub @ninginthecloud


pytorch_lightning/loops/utilities.py
```diff
@@ -105,34 +105,25 @@ def _parse_loop_limits(
...
 
     Returns:
         the keyword arguments for the training step
     """
-    # enable not needing to add opt_idx to training_step
-    step_kwargs = OrderedDict([("batch", batch)])
-
     training_step_fx = getattr(lightning_module, "training_step")
-
-    if is_param_in_hook_signature(training_step_fx, "batch_idx", min_args=2):
```
carmocca, line 133: https://github.com/PyTorchLightning/pytorch-lightning/pull/9052 added support for not passing the `batch_idx` but there were no tests for it, I've removed the argument to a few.

tests/loops/batch/test_truncated_bptt.py
```diff
@@ -170,3 +170,35 @@ def training_step(self, *args, **kwargs):
     )
     trainer.fit(model)
     assert set(trainer.logged_metrics) == {"loss_step", "loss_epoch"}
+
+
+def test_hiddens_multiple_optimizers(tmpdir):
```
carmocca, line 175: We were missing a test for the hidden values with multiple optimizers and TBPTT enabled

pytorch_lightning/loops/optimization/manual_loop.py
```diff
@@ -149,3 +145,17 @@ def _on_before_step(self) -> None:
     def _on_after_step(self) -> None:
         self.trainer.profiler.stop("optimizer_step")
         self.optim_step_progress.increment_completed()
+
+    def _build_kwargs(self, kwargs: OrderedDict, hiddens: Optional[Any]) -> OrderedDict:
```
awaelchli, line 149: why the indirection of this method, if we could just directly call the function?
is this meant to be overridden for customization?

pytorch_lightning/loops/optimization/manual_loop.py
```diff
@@ -149,3 +145,17 @@ def _on_before_step(self) -> None:
...
+        """Helper function to build the arguments for the current step.
```
awaelchli, line 150:
```suggestion
        """Helper method to build the arguments for the current step.
```

awaelchli (approved):

pytorch_lightning/loops/optimization/manual_loop.py
carmocca: Yes, and for consistency. Although I realize that it's prefixed by `_`

I'll resolve this comment but tell me if you prefer to remove it. I don't mind.

rohitgr7 (approved):

pl_examples/loop_examples/yielding_training_step.py
```diff
@@ -56,28 +55,25 @@ def __init__(self):
     def connect(self, **kwargs):
         raise NotImplementedError(f"{self.__class__.__name__} does not connect any child loops.")
 
-    def on_run_start(self, batch, optimizers, batch_idx):
-        super().on_run_start(batch, optimizers, batch_idx)
+    def on_run_start(self, optimizers, kwargs):
```
Borda, line 58: shall we rather propagate it as `**kwargs` again?

pytorch_lightning/loops/batch/training_batch_loop.py
```diff
@@ -59,35 +58,35 @@ def reset(self) -> None:
         """Resets the loop state."""
         self._outputs = []
 
-    def on_run_start(self, batch: Any, batch_idx: int) -> None:  # type: ignore[override]
+    def on_run_start(self, kwargs: OrderedDict) -> None:  # type: ignore[override]
```
Borda, line 61:
```suggestion
    def on_run_start(self, kwargs: Dict[str, ANY]) -> None:  # type: ignore[override]
```

pytorch_lightning/loops/batch/training_batch_loop.py
```diff
@@ -59,35 +58,35 @@ def reset(self) -> None:
...
         """Splits the data into tbptt splits.
 
         Args:
-            batch: the current batch to run the trainstep on
-            batch_idx: the index of the current batch
+            kwargs: the kwargs passed down to the hooks.
         """
-        void(batch_idx)
+        batch = kwargs["batch"]
```
Borda, line 67: how about if the key is missing, shall we rather?
```suggestion
        batch = kwargs.get("batch")
```

pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -190,20 +189,16 @@ def reset(self) -> None:
         self._outputs = {}
 
     def on_run_start(  # type: ignore[override]
-        self, batch: Any, optimizers: List[Tuple[int, Optimizer]], batch_idx: int
+        self, optimizers: List[Tuple[int, Optimizer]], kwargs: OrderedDict
```
Borda, line 192: are you sure that it is order? I may suggest using more general Mapping

pl_examples/loop_examples/yielding_training_step.py
carmocca: No because we want to pass the container which is what's modified per-loop.

pytorch_lightning/loops/batch/training_batch_loop.py
carmocca: No, that would be a silent error. I want it to raise a KeyError in this case because it would clearly show a bug in the implementation

Borda (approved):

Issue #11409, opened 2022-01-11
[RFC] Deprecate `BaseProfiler` in favor of `AbstractProfiler` and move file-based code to utilities


daniellepintz:
## Proposed refactor

* Deprecate `BaseProfiler`:
https://github.com/PyTorchLightning/pytorch-lightning/blob/cec4f17df3cbdf2b585b723d518e03bd500a7c5b/pytorch_lightning/profiler/base.py#L51

    in favor of `AbstractProfiler`:
https://github.com/PyTorchLightning/pytorch-lightning/blob/cec4f17df3cbdf2b585b723d518e03bd500a7c5b/pytorch_lightning/profiler/base.py#L27

* Move file-based code to utility functions in `pytorch_lightning/utilities/profiler.py`

* Update all profilers to subclass `AbstractProfiler` directly and call the file-based utilities

* Update trainer to expect a profiler of class `AbstractProfiler`: 
https://github.com/PyTorchLightning/pytorch-lightning/blob/34c62da37dc1ed9a1de7e023c690c7528ee56c60/pytorch_lightning/trainer/trainer.py#L181

### Motivation

We currently have both `BaseProfiler` and `AbstractProfiler`, where all of our profilers subclass `BaseProfiler`, which in turn subclasses `AbstractProfiler`. Instead of two layers of inheritance, it would be simpler to only have one.

`BaseProfiler` assumes we want to save outputs to a file. This is not the case for all profilers, like the XLAProfiler, where we have to set filename to None as a workaround. Someone could also seek to publish data to other sinks, such as a remote service for monitoring. Thus, we should have all profilers subclass `AbstractProfiler` directly, and we can create utility functions to share the code for file-based handling.

cc @justusschock @awaelchli @akihironitta @rohitgr7 @carmocca @kaushikb11 @ninginthecloud

Part of #7740

carmocca:
Sounds good to me!

daniellepintz:
As part of this issue, we could also consider renaming the base profiler class to simply `Profiler`. This would be in line with the base classes for `Callback`, `Accelerator`, `Strategy`, and `Loop`. It seems that this is the standard across Lightning.

This would also allow us to make some changes to the API if we want, such as replacing `**kwargs` in `setup` and `teardown` here with real arguments:
[...]

PR #12084, opened 2022-02-24
Deprecate `weights_save_path` from the Trainer constructor 

daniellepintz:
## What does this PR do?
Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/11768

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #11832, opened 2022-02-10
Remove calls to agg_and_log_metrics

akashkw:
## What does this PR do?
Removes all calls to agg_and_log_metrics in the codebase. Danielle showed here https://github.com/PyTorchLightning/pytorch-lightning/issues/9145#issuecomment-1026186725 that agg_and_log_metrics is a no-op and can safely be removed.

Part of #9145 

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #12106, opened 2022-02-24
Deprecate `AbstractProfiler` in favor of `BaseProfiler`

akashkw:
## What does this PR do?
This PR combines `AbstractProfiler` and `BaseProfiler` into one `BaseProfiler` class
`AbstractProfiler` is deprecated in favor of `BaseProfiler`.

Part of #11409

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/profiler/base.py
```diff
@@ -25,7 +25,14 @@
...
 
 class AbstractProfiler(ABC):
-    """Specification of a profiler."""
+    """Specification of a profiler.
+
+    See deprecation message below
+
+    .. deprecated:: v1.6
+        `AbstractProfiler` was deprecated in v1.6 and will be removed in v1.8.
+        Please use `Profiler` instead
```
daniellepintz, line 34:
```suggestion
        Please use `BaseProfiler` instead
```

CHANGELOG.md
```diff
@@ -430,6 +430,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated `LightningLoggerBase.agg_and_log_metrics` in favor of `LightningLoggerBase.log_metrics` ([#11832](https://github.com/PyTorchLightning/pytorch-lightning/pull/11832))
 
 
+- Deprecated `pytorch_lightning.profiler.AbstractProfiler` in favor of `pytorch_lightning.profiler.BaseProfiler` ([#FILLTHISIN](https://github.com/PyTorchLightning/pytorch-lightning/pull/FILLTHISIN))
```
daniellepintz, line 433:
```suggestion
- Deprecated `pytorch_lightning.profiler.AbstractProfiler` in favor of `pytorch_lightning.profiler.BaseProfiler` ([#12106](https://github.com/PyTorchLightning/pytorch-lightning/pull/12106))
```

pytorch_lightning/profiler/base.py
```diff
@@ -64,6 +71,18 @@ def __init__(
         self._local_rank: Optional[int] = None
         self._stage: Optional[str] = None
 
+    @abstractmethod
```
daniellepintz, line 74: what is the benefit of changing these to abstractmethods?

akashkw: I think it more clearly indicates to users that they need to be implemented. Simply having a NotImplementedError is easier to miss in my opinion

daniellepintz (approved):
Looks good!

pytorch_lightning/profiler/base.py
daniellepintz: sg!

pytorch_lightning/profiler/base.py
```diff
@@ -25,7 +25,14 @@
 
 
 class AbstractProfiler(ABC):
-    """Specification of a profiler."""
+    """Specification of a profiler.
+
+    See deprecation message below
+
+    .. deprecated:: v1.6
```
rohitgr7, line 32:
```suggestion
    """Specification of a profiler.

    .. deprecated:: v1.6
```

rohitgr7 (approved):
I wonder if we could just remove it since we never recommended it to users anywhere.

pytorch_lightning/profiler/base.py
akashkw: That line is actually a workaround for a bug in docformatter https://github.com/PyCQA/docformatter/issues/78.
I've seen this line show up in some other places as well:
https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loggers/base.py#L187.

pytorch_lightning/profiler/base.py
```diff
@@ -65,6 +72,18 @@ def __init__(
...
+        """Defines how to start recording an action."""
+
+    @abstractmethod
+    def stop(self, action_name: str) -> None:
+        """Defines how to record the duration once an action is complete."""
+
+    @abstractmethod
+    def summary(self) -> str:
+        """Create profiler summary in text format."""
+
```
ananthsub, line 86: We don't need this. The trainer never explicitly calls `summary()` and it's redundant with `describe` already on the base class.

pytorch_lightning/profiler/base.py
daniellepintz: Would this also work as a workaround?
```
"""
Specification of a profiler.

    .. deprecated:: v1.6
"""
```

akashkw: Unfortunately that doesn't work. Docformatter moves `Specification of a profiler` up one line, and then the same bug returns

rohitgr7:
```
""""Specification of a profiler.

.. deprecated:: v1.6
    :class:`AbstractProfiler` was ...
```
can you try this? if this doesn't work, keep it just like it is now.

or if you end up removing the class altogether, then it won't be a problem anyways.

daniellepintz: hmm, I'm just wondering how I was able to do this in https://github.com/PyTorchLightning/pytorch-lightning/pull/12084/files#diff-667a2513c158c2b01d73b479e4dea75587d8531f9cd286891d34570a2fd145dcR2216

akashkw: > hmm, I'm just wondering how I was able to do this in https://github.com/PyTorchLightning/pytorch-lightning/pull/12084/files#diff-667a2513c158c2b01d73b479e4dea75587d8531f9cd286891d34570a2fd145dcR2216

Are you referring to line 2217? I think that works because there is a separate line with text above the deprecation message. This bug only appears if there is no text between the top line of text and the deprecation warning.

akashkw: You can see the same issue appears over here:
Commit: https://github.com/PyTorchLightning/pytorch-lightning/pull/11832/commits/27913e4b8002dfc91cb8c1a527a8e8822ef2b4c9
Commit after docformatter: https://github.com/PyTorchLightning/pytorch-lightning/pull/11832/commits/1b7a1bd05ac2fdecc7c1d765c54e3056b23a28b3
The docformatter bug appears in `_reduce_agg_metrics()` and `_finalize_agg_metrics()`, but notably it does not affect `_aggregate_metrics()`. It seems like this bug is not an issue if there is a new line with text either above or below the deprecation message (not include the top line).

akashkw: > ```
> """"Specification of a profiler.
> 
> .. deprecated:: v1.6
>     :class:`AbstractProfiler` was ...
> ```
> 
> can you try this? if this doesn't work, keep it just like it is now.
> 
> or if you end up removing the class altogether, then it won't be a problem anyways.

I gave this a try and it ended up formatting to
```
""""Specification of a profiler.

.. deprecated:: v1.6   :class:`AbstractProfiler` was ...
```
This is the exact same issue as here:
https://github.com/PyTorchLightning/pytorch-lightning/commit/1b7a1bd05ac2fdecc7c1d765c54e3056b23a28b3
If we can find a better workaround then there are 4 other places we should fix too.

pytorch_lightning/profiler/base.py
```diff
@@ -180,15 +195,6 @@ def teardown(self, stage: Optional[str] = None) -> None:
     def __del__(self) -> None:
         self.teardown(stage=self._stage)
 
-    def start(self, action_name: str) -> None:
-        raise NotImplementedError
-
-    def stop(self, action_name: str) -> None:
-        raise NotImplementedError
-
-    def summary(self) -> str:
```
ananthsub, line 189: summary can return an empty string instead of raising a NotImplementedError. this way the passthrough profiler doesn't need to override this, and users who don't implement this won't see exceptions

ananthsub (approved):

daniellepintz:
What do people think about outright removing AbstractProfiler as @rohitgr7 suggested, since we don't recommend it to users anywhere?

ananthsub:
> What do people think about outright removing AbstractProfiler as @rohitgr7 suggested, since we don't recommend it to users anywhere?

1. Even if it's not used anywhere, it's still part of the public API. If we can provide backwards compatibility according to the API evolution process, we should.
2. It's incredibly low-cost to mark deprecated and remove only for v1.8 since nothing depends on it. It's existence does not block further refactors or improvements

So I'd prefer not removing it outright

pytorch_lightning/profiler/base.py
```diff
@@ -205,6 +214,3 @@ def start(self, action_name: str) -> None:
 
     def stop(self, action_name: str) -> None:
         pass
-
-    def summary(self) -> str:
```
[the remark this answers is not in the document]
kaushikb11: Yup, as it subclassing `BaseProfiler`

kaushikb11 (approved):

ethanwharris (approved):

pytorch_lightning/profiler/base.py
```diff
@@ -26,7 +26,14 @@
...
 
 class AbstractProfiler(ABC):
-    """Specification of a profiler."""
+    """Specification of a profiler.
+
+    See deprecation warning below
+
+    .. deprecated:: v1.6
+        `AbstractProfiler` was deprecated in v1.6 and will be removed in v1.8.
+        Please use `BaseProfiler` instead
```
carmocca, line 35: It might be a bit silly because this doesn't raise a warning.

But it would be good to add a "fake" deprecation test so we remember this got deprecated in the future.

carmocca (approved):

PR #12124, opened 2022-02-26
Add profiling to dataloader

akashkw:
## What does this PR do?

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
@@ -107,11 +107,12 @@ class EvaluationEpochLoop(Loop):
         void(dl_max_batches)
 
         assert self._dataloader_iter is not None
-        if not isinstance(data_fetcher, DataLoaderIterDataFetcher):
-            batch_idx = self.batch_progress.current.ready
-            batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
-        else:
-            batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
+        with self.trainer.profiler.profile("{}_dataloader_next"):
+            if not isinstance(data_fetcher, DataLoaderIterDataFetcher):
+                batch_idx = self.batch_progress.current.ready
+                batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
+            else:
+                batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
 
         # configure step_kwargs
         kwargs = self._build_kwargs(kwargs, batch, batch_idx)
--- a/pytorch_lightning/loops/epoch/prediction_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/prediction_epoch_loop.py
@@ -89,7 +89,8 @@ class PredictionEpochLoop(Loop):
             num_dataloaders: the total number of dataloaders
             return_predictions: whether to return the obtained predictions
         """
-        batch_idx, batch = next(dataloader_iter)
+        with self.trainer.profiler.profile("{}_dataloader_next"):
+            batch_idx, batch = next(dataloader_iter)
         self._seen_batch_indices = self._get_batch_indices(dataloader_idx)
         # we need to truncate the list of batch indices due to prefetching in the dataloader and Lightning
         self._seen_batch_indices = self._seen_batch_indices[: (self.batch_progress.current.completed + 1)]
--- a/pytorch_lightning/loops/epoch/training_epoch_loop.py
+++ b/pytorch_lightning/loops/epoch/training_epoch_loop.py
@@ -158,11 +158,12 @@ class TrainingEpochLoop(loops.Loop[_OUTPUTS_TYPE]):
         self.val_loop.restarting = False
 
         assert self._dataloader_iter is not None
-        if not isinstance(data_fetcher, DataLoaderIterDataFetcher):
-            batch_idx = self.batch_idx + 1
-            batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
-        else:
-            batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
+        with self.trainer.profiler.profile("{}_dataloader_next"):
+            if not isinstance(data_fetcher, DataLoaderIterDataFetcher):
+                batch_idx = self.batch_idx + 1
+                batch, self.batch_progress.is_last_batch = next(self._dataloader_iter)
+            else:
+                batch_idx, (batch, self.batch_progress.is_last_batch) = next(self._dataloader_iter)
 
         self.batch_progress.increment_ready()
 


pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -107,7 +107,8 @@ def advance(  # type: ignore[override]
         void(dl_max_batches)
 
         assert self._dataloader_iter is not None
-        with self.trainer.profiler.profile("{}_dataloader_next"):
+        action_name = "test_dataloader_next" if self.trainer.testing else "val_dataloader_next"
```
ananthsub, line 110: let's follow the action name style here so it's uniform in the profiler later
https://github.com/PyTorchLightning/pytorch-lightning/blob/7e2f9fbad555242b0ceb2a24e5e4c004f0701bae/pytorch_lightning/trainer/trainer.py#L1604

        with self.profiler.profile(f"[LightningModule]{pl_module.__class__.__name__}.{hook_name}"):

```suggestion
        suffix = "test_dataloader_next" if self.trainer.testing else "val_dataloader_next"
        action_name = f"[{self.__class__.__name__}].{suffix}"
```

akashkw: There are a couple spots that don't follow this convention. Should these be updated too?

- https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loops/fit_loop.py#L273
- https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loops/epoch/training_epoch_loop.py#L198

carmocca (changes requested):
Hi @akashkw!

We cannot profile `next()` like this as this would not cover prefetched batches or inter-batch paralelism.

I recently had a call with @daniellepintz where I went over possible solutions. Here's some code pointers:

The concrete lines to profile:
https://github.com/PyTorchLightning/pytorch-lightning/blob/a0655611de460f1659b46150cda256d2e3fa974e/pytorch_lightning/utilities/fetching.py#L267


https://github.com/PyTorchLightning/pytorch-lightning/blob/a0655611de460f1659b46150cda256d2e3fa974e/pytorch_lightning/utilities/fetching.py#L336

Option 1:
Pass a profiler reference to the data fetchers, maybe also the stage and dataloader idx for more fine-grained profiling, then use these directly inside the fetching functions
Option 2:
Inject the profiler usage through callables that could be part of these hooks:
https://github.com/PyTorchLightning/pytorch-lightning/blob/a0655611de460f1659b46150cda256d2e3fa974e/pytorch_lightning/utilities/fetching.py#L61-L65
with a pattern similar to the one used in this PR to inject logic around the optimizer.step()
https://github.com/PyTorchLightning/pytorch-lightning/commit/4cc05b2cf98e49168a5f5dc265647d75d1d3aae9

akashkw:
> Option 1:
> Pass a profiler reference to the data fetchers, maybe also the stage and dataloader idx for more fine-grained profiling, then  use these directly inside the fetching functions
> Option 2:
> Inject the profiler usage through callables that could be part of these hooks:

@carmocca do you have any preference between these two approaches?

carmocca:
I prefer option (2) so the fetching does not need to keep a reference to the trainer. (the profiler is owned by the trainer)

pytorch_lightning/loops/epoch/prediction_epoch_loop.py
```diff
@@ -89,7 +89,8 @@ def advance(  # type: ignore[override]
             num_dataloaders: the total number of dataloaders
             return_predictions: whether to return the obtained predictions
         """
-        batch_idx, batch = next(dataloader_iter)
+        with self.trainer.profiler.profile(f"[{self.__class__.__name__}].pred_dataloader_next"):
+            batch_idx, batch = next(dataloader_iter)
```
akashkw, line 93: `prediction_epoch_loop` doesn't seem to use data fetchers, should I just leave this as is?

carmocca: Yes

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -88,6 +88,11 @@ def on_run_start(  # type: ignore[override]
         # add the previous `fetched` value to properly track `is_last_batch` with no prefetching
         data_fetcher.fetched += self.batch_progress.current.ready
 
+        data_fetcher._start_profiler = self.trainer.profiler.start
+        data_fetcher._stop_profiler = self.trainer.profiler.stop
+        suffix = "test_dataloader_next" if self.trainer.testing else "val_dataloader_next"
```
carmocca, line 93: I think we should differentiate per dataloader idx too

pytorch_lightning/loops/epoch/prediction_epoch_loop.py
```diff
@@ -89,7 +89,8 @@ def advance(  # type: ignore[override]
             num_dataloaders: the total number of dataloaders
             return_predictions: whether to return the obtained predictions
         """
-        batch_idx, batch = next(dataloader_iter)
+        with self.trainer.profiler.profile(f"[{self.__class__.__name__}].pred_dataloader_next"):
```
carmocca, line 92:
```suggestion
        with self.trainer.profiler.profile(f"[{self.__class__.__name__}].predict_dataloader_next"):
```

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -88,6 +88,11 @@ def on_run_start(  # type: ignore[override]
...
+        data_fetcher._profiler_action_name = f"[{self.__class__.__name__}].{suffix}"
```
carmocca, line 94: Let's avoid the `_profiler_action_name` attribute in the fetcher

```python
data_fetcher._start_profiler = self._on_before_fetch

def on_before_fetch(self):
    suffix = "test_dataloader_next" if self.trainer.testing else "val_dataloader_next"
    self.trainer.profiler.start(f"[{self.__class__.__name__}].{suffix}")
```


You could also use `functools.partial` but I would say this is nicer

pytorch_lightning/utilities/fetching.py
```diff
@@ -385,3 +398,9 @@ def fetching_function(self) -> Tuple[int, Iterator]:
         if not self.done:
             return self.fetched, self.iterator
         raise StopIteration
+
+    def on_fetch_start(self) -> None:
```
carmocca, line 402: These are only called in `DataFetcher._fetch_next_batch` but that method is not used by this class (this class subclasses `AbstractDataFetcher`)

Particularly, this start and stop needs to wrap the `next` in `StepFuncDataLoaderIter`

carmocca (commented):
Good progress!

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
akashkw: Didn't know about `functools.partial`, thanks for sharing!
I agree, this is approach is cleaner.

pytorch_lightning/utilities/fetching.py
```diff
@@ -320,9 +330,11 @@ def move_to_device(self, batch: Any) -> Any:
 
     def on_fetch_start(self) -> "torch.cuda.Event":
         # create a cuda event used to record the async stream of data to device.
+        self._start_profiler()
         return torch.cuda.Event()
```
carmocca, line 334: Probably overkill but if creating the `cuda.Event()` was a very expensive call, then we would want to do

```suggestion
        event = torch.cuda.Event()
        self._start_profiler()
        return event
```

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -146,6 +146,15 @@ def on_run_start(self, data_fetcher: AbstractDataFetcher) -> None:  # type: igno
         # add the previous `fetched` value to properly track `is_last_batch` with no prefetching
         data_fetcher.fetched += self.batch_progress.current.ready
 
+        data_fetcher._start_profiler = self._on_before_fetch
+        data_fetcher._stop_profiler = self._on_after_fetch
+
+    def _on_before_fetch(self):
```
carmocca, line 152: Remember typing (see mypy failures in CI)

carmocca (commented):
You should check for these new entries in the profiler tests

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -88,6 +88,21 @@ def on_run_start(  # type: ignore[override]
         # add the previous `fetched` value to properly track `is_last_batch` with no prefetching
         data_fetcher.fetched += self.batch_progress.current.ready
 
+        dataloader_idx = f"_{kwargs['dataloader_idx']}" if "dataloader_idx" in kwargs else ""
+        if self.trainer.testing:
+            suffix = f"test_dataloader{dataloader_idx}_next"
```
carmocca, line 93: In result logging, we use the convention `dataloader_idx_{}`. We should do the same here too

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -88,6 +88,21 @@ def on_run_start(  # type: ignore[override]
...
+        else:
+            suffix = f"val_dataloader{dataloader_idx}_next"
+
+        def _on_before_fetch() -> None:
+            self.trainer.profiler.start(f"[{self.__class__.__name__}].{suffix}")
+
+        def _on_after_fetch() -> None:
+            self.trainer.profiler.stop(f"[{self.__class__.__name__}].{suffix}")
```
carmocca, line 101: I'm uncomfortable using local functions as they will not work if this class gets pickled.

carmocca (commented):
CHANGELOG and tests!

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
tchaton: Same here. Let's avoid them.

pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -146,6 +146,15 @@ def on_run_start(self, data_fetcher: AbstractDataFetcher) -> None:  # type: igno
         # add the previous `fetched` value to properly track `is_last_batch` with no prefetching
         data_fetcher.fetched += self.batch_progress.current.ready
 
+        def _on_before_fetch() -> None:
+            self.trainer.profiler.start(f"[{self.__class__.__name__}].train_dataloader_next")
```
tchaton, line 150: Same here.

carmocca (approved):

tests/utilities/test_fetching.py
```diff
@@ -485,3 +486,88 @@ def validation_step(self, batch, batch_idx):
     assert dm.count_called_on_before_batch_transfer == 4
     assert dm.count_called_transfer_batch_to_device == 4
     assert dm.count_called_on_after_batch_transfer == 4
+
+
+@RunIf(skip_windows=True)  # TODO: all durations are 0 on Windows
```
carmocca, line 491: I'm unsure if this is expected or a bug...

rohitgr7 (approved):

otaj (approved):

PR #12142, opened 2022-02-28
Remove version argument from WandbLogger to fix directory structure

### LINKED ISSUE #12036: Several issues with `WandbLogger` output directory structure in offline mode
## Proposed refactor

Change the default `name` to "lightning_logs" and the default `save_dir` to "wandb".
https://github.com/PyTorchLightning/pytorch-lightning/blob/60fe1520fd8ac8fa9006237f31045465200c9803/pytorch_lightning/loggers/wandb.py#L252-L253

### Motivation

Currently the `name` and `save_dir` properties on the `WandbLogger` default to `None`, and the version can be `None` as well.  This causes several issues in the directory structure as shown in the image below:

wandb currentl

1. The main logging directory is named `None` which is an odd name for a directory to store logs
2. Wandb already automatically creates a directory called `wandb` which contains metadata. It is weird to have two separate logging directories for Wandb - let's instead default to storing everything in one. Plus, the doc literally says the save_dir is `wandb` by default, which is not true at the moment:
https://github.com/PyTorchLightning/pytorch-lightning/blob/60fe1520fd8ac8fa9006237f31045465200c9803/pytorch_lightning/loggers/wandb.py#L221
3. The profiler output is unable to save to a path since the logger's save_dir is `None`, and we get the error `The PyTorchProfiler failed to export trace as dirpath is None`
4. A fourth issue, but one that is not addressed by this proposed change, is that the subdirectory is named "version_None" which is also an odd name for a directory.

How the directory structure will look after this change:
wandb after

cc @awaelchli @morganmcg1 @AyushExel @borisdayma @scottire @manangoel99

manangoel99:
## What does this PR do?
Fixes #12036 

The PR fixes the directory structure of the logs when using `WandbLogger` without any arguments while creating the instance. The default arguments initially created directories named "None". Now they will be named according to the wandb project and run id.

Massive shoutout to @daniellepintz for raising this issue!

155562617-d1195605-73d5-4e87-bf36-ca13bab5bd9b

None
## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loggers/wandb.py
```diff
@@ -293,7 +292,7 @@ def __init__(
         self._wandb_init = dict(
             name=name,
             project=project,
-            id=version or id,
+            id=id,
```
daniellepintz, line 295: I am not following how this would fix the issue. if both version and id are None, we still have this issue right?

daniellepintz:
@manangoel99 thanks so much for the PR and sorry for the delay! I think there might be some confusion here - in https://github.com/PyTorchLightning/pytorch-lightning/issues/12036 I am talking about the directory structure in offline mode. Are we on the same page?

manangoel99:
@daniellepintz Hi! Sorry for the late response. Yes I completely get that you were talking about the directory structure and this is a very weird bug which is getting fixed with this change. I've been trying to track down exactly how that is happening but somehow it works. I think it could be something to do how pytorch lightning creates it's directories internally using the version property.
Since for wandb, `version` and `id` are the same thing, I think the bug came from that so removing version as an argument worked.

pytorch_lightning/loggers/wandb.py
manangoel99: So whenever the run object is initialized, if id is passed as None, wandb creates it's own id which is used everywhere else.

manangoel99:
@awaelchli @carmocca I think this PR is good to go. Please have a look whenever possible. Thanks!

pytorch_lightning/loggers/wandb.py
```diff
@@ -254,7 +254,6 @@ def __init__(
         offline: Optional[bool] = False,
         id: Optional[str] = None,
         anonymous: Optional[bool] = None,
-        version: Optional[str] = None,
```
carmocca, line 257: If this is removed without deprecation, it should be mentioned in the "Removed" section of the CHANGELOG

carmocca (approved):
There's one test failing @manangoel99 

`tests/loggers/test_wandb.py::test_wandb_logger_init`

pytorch_lightning/loggers/wandb.py
rohitgr7: shouldn't this be deprecated first?

manangoel99: @rohitgr7 it's a very weird bug. Keeping `version=None` and providing a deprecation warning doesn't work. This works correctly only if the `version` argument is removed. So I think I can add it to the "Removed" section of the CHANGELOG as suggested by @carmocca .

PR #12147, opened 2022-02-28
Deprecate `LoggerCollection` in favor of `trainer.loggers`

### LINKED ISSUE #11232: [RFC] Better support using multiple loggers simultaneously by deprecating `LoggerCollection`

akashkw:
## What does this PR do?

Fixes #11232

## Before submitting

## PR review

## Did you have fun?


tests/trainer/properties/test_loggers.py
```diff
@@ -52,6 +54,10 @@ def test_trainer_loggers_setters():
     logger2 = CustomLogger()
     logger_collection = LoggerCollection([logger1, logger2])
     logger_collection_2 = LoggerCollection([logger2])
+    logger_collection_dep = "Using `trainer.logger` when Trainer is configured to use multiple loggers."
+    " `LoggerCollection` is deprecated in v1.6 in favor of `trainer.loggers` and will be removed in v1.8."
+    " This behavior will change in v1.8 such that `trainer.logger` will return the first"
+    " logger in `trainer.loggers`."
```
carmocca, line 60: Don't need to match the full string, `match=...` takes a regex

```suggestion
    logger_collection_dep = "LoggerCollection` is deprecated in v1.6"
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2599,10 +2599,11 @@ def logger(self) -> Optional[LightningLoggerBase]:
...
         else:
-            rank_zero_warn(
-                "Using trainer.logger when Trainer is configured to use multiple loggers."
-                " This behavior will change in v1.8 when LoggerCollection is removed, and"
-                " trainer.logger will return the first logger in trainer.loggers"
+            rank_zero_deprecation(
+                "Using `trainer.logger` when Trainer is configured to use multiple loggers."
+                " `LoggerCollection` is deprecated in v1.6 in favor of `trainer.loggers` and will be removed in v1.8."
+                " This behavior will change in v1.8 such that `trainer.logger` will return the first"
+                " logger in `trainer.loggers`."
```
carmocca, line 2606: Shouldn't this be inside the `LoggerCollection.__init__`?

akashkw: It could be, this is still up for discussion. @daniellepintz and I talked about it earlier and we concluded that since there was already a warning here, it made sense to replace it with a deprecation warning. 

Most people who are using `LoggerCollection` are doing so through `trainer.logger` so they should be made aware of both the deprecation of `LoggerCollection` and the changing behavior of `trainer.logger`.  

What we could do is have one warning inside the `trainer.logger` getter and another warning in `LoggerCollection.__init__`. The `trainer.logger` getter constructs a `LoggerCollection` each time it's called so both warnings would be displayed.

akashkw: For now, having two deprecation warnings makes the most sense to me, so I went ahead and added a deprecation message to `LoggerCollection.__init__`.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2599,10 +2599,10 @@ def logger(self) -> Optional[LightningLoggerBase]:
         if len(self.loggers) == 1:
             return self.loggers[0]
         else:
-            rank_zero_warn(
-                "Using trainer.logger when Trainer is configured to use multiple loggers."
-                " This behavior will change in v1.8 when LoggerCollection is removed, and"
-                " trainer.logger will return the first logger in trainer.loggers"
+            rank_zero_deprecation(
```
daniellepintz, line 2602: I think this should be `rank_zero_warn` because otherwise the framework will emit dep warnings every time it accesses `trainer.logger` when we have multiple loggers.

pytorch_lightning/trainer/trainer.py
daniellepintz: I agree it makes sense to have a warning in both places

pytorch_lightning/trainer/trainer.py
```diff
@@ -2600,9 +2600,9 @@ def logger(self) -> Optional[LightningLoggerBase]:
             return self.loggers[0]
         else:
             rank_zero_warn(
-                "Using trainer.logger when Trainer is configured to use multiple loggers."
-                " This behavior will change in v1.8 when LoggerCollection is removed, and"
-                " trainer.logger will return the first logger in trainer.loggers"
+                "Using `trainer.logger` when Trainer is configured to use multiple loggers."
```
ananthsub, line 2603:
```suggestion
                "You are using `trainer.logger` when the Trainer is configured to use multiple loggers."
```

ananthsub (approved):

rohitgr7 (approved):

carmocca (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2600,9 +2600,9 @@ def logger(self) -> Optional[LightningLoggerBase]:
             return self.loggers[0]
         else:
             rank_zero_warn(
-                "Using trainer.logger when Trainer is configured to use multiple loggers."
-                " This behavior will change in v1.8 when LoggerCollection is removed, and"
-                " trainer.logger will return the first logger in trainer.loggers"
+                "You are using `trainer.logger` when the Trainer is configured to use multiple loggers."
+                " This behavior will change in v1.8 such that `trainer.logger` will return the first"
+                " logger in `trainer.loggers`."
```
ananthsub, line 2605:
```suggestion
                "Using trainer.logger when Trainer is configured to use multiple loggers."
                " This behavior will change in v1.8 when LoggerCollection is removed, and"
                " trainer.logger will return the first logger in trainer.loggers"
```

pytorch_lightning/loggers/base.py
```diff
@@ -221,13 +221,19 @@ def version(self) -> Union[int, str]:
 class LoggerCollection(LightningLoggerBase):
     """The :class:`LoggerCollection` class is used to iterate all logging actions over the given `logger_iterable`.
 
+    .. deprecated:: v1.6
+        `LoggerCollection` is deprecated in v1.6 in favor of `trainer.loggers` and will be removed in v1.8.
```
awaelchli, line 225: I would prefer to make the error message more explicit. The short changelog entry is fine but here we need to say what the user should do instead. 

```suggestion
        `LoggerCollection` is deprecated in v1.6 and will be removed in v1.8. Directly pass a list of loggers to the Trainer and access the list via the `trainer.loggers` attribute. 
```

pytorch_lightning/loggers/base.py
```diff
@@ -221,13 +221,19 @@ def version(self) -> Union[int, str]:
...
+
     Args:
         logger_iterable: An iterable collection of loggers
     """
 
     def __init__(self, logger_iterable: Iterable[LightningLoggerBase]):
         super().__init__()
         self._logger_iterable = logger_iterable
+        rank_zero_deprecation(
+            "`LoggerCollection` is deprecated in v1.6 in favor of `trainer.loggers` and will be removed in v1.8."
```
awaelchli, line 235: same here :))

tests/trainer/properties/test_loggers.py
```diff
@@ -50,8 +52,10 @@ def test_trainer_loggers_setters():
     """Test the behavior of setters for trainer.logger and trainer.loggers."""
     logger1 = CustomLogger()
     logger2 = CustomLogger()
-    logger_collection = LoggerCollection([logger1, logger2])
-    logger_collection_2 = LoggerCollection([logger2])
+    with pytest.deprecated_call(match="`LoggerCollection` is deprecated in v1.6"):
+        logger_collection = LoggerCollection([logger1, logger2])
+    with pytest.deprecated_call(match="`LoggerCollection` is deprecated in v1.6"):
+        logger_collection_2 = LoggerCollection([logger2])
```
awaelchli, line 58: this seems to be unused, weird.

awaelchli (approved):

tests/trainer/properties/test_loggers.py
akashkw: Not sure I understand what you mean. Can you clarify?
`logger_collection` and `logger_collection_2` are both used later on in this unit test.

pytorch_lightning/loggers/base.py
daniellepintz: I think this is fine, but the user actually doesn't have to change any behavior, they still pass a list of loggers to the Trainer

awaelchli: But hold on, this warning is when LoggerCollection gets instantiated by the user. This is what we want to warn about, and passing it to the trainer should trigger a warning as well. However, we need to avoid triggering the warning internally or multiple times. This is currently  a problem, because for the getter in 

`trainer.logger`

we return 

`return LoggerCollection(self.loggers)`

which will once again trigger a warning that the user will find confusing and can't change. Do you agree?
Can we make an internal flag that we flip when we want to suppress the warning internally?

ananthsub: We should still warn about that scenario because the behavior will change following v1.8 as users will receive the first logger in the list instead of a logger collection. To update their code, they would need to use trainer.loggers

daniellepintz: @awaelchli after https://github.com/PyTorchLightning/pytorch-lightning/pull/11920 there are virtually no places where we call `trainer.logger` internally - we updated them all to `trainer.loggers`

awaelchli: @daniellepintz To be very clear:

```py
from pytorch_lightning.loggers import TensorBoardLogger


if __name__ == "__main__":
    trainer = Trainer(
        logger=[TensorBoardLogger(save_dir="."), TensorBoardLogger(save_dir=".")]
    )
    _ = trainer.logger
```

Nothing in the above code is deprecated usage. Yet we see this message:

```
/Users/adrian/repositories/pytorch-lightning/pytorch_lightning/trainer/trainer.py:2621: UserWarning: Using trainer.logger when Trainer is configured to use multiple loggers. This behavior will change in v1.8 when LoggerCollection is removed, and trainer.logger will return the first logger in trainer.loggers
  rank_zero_warn(
/Users/adrian/repositories/pytorch-lightning/pytorch_lightning/loggers/base.py:235: LightningDeprecationWarning: `LoggerCollection` is deprecated in v1.6 and will be removed in v1.8. Directly pass a list of loggers to the Trainer and access the list via the `trainer.loggers` attribute.
  rank_zero_deprecation(
```

There is nothing I as a user can do to avoid the second warning, except change to `trainer.loggers` but that's not deprecated API. Plus, the warning mentions `LoggerCollection` which I have not used anywhere in my code. 

It's triggered internally because we instantiate `LoggerCollection` and we need to suppress that warning, for example with https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings or the toggle as I showed above. No matter how unlikely it is that a user runs into this issue, we can't do it because it's not correct.

akashkw: We were actually having a similar discussion earlier. https://github.com/PyTorchLightning/pytorch-lightning/pull/12147#r816308356. Having the user see both the warning and the deprecation was an intentional choice (originally both messages were combined into one). I agree though that it would be confusing for the user to see the deprecation warning from instantiating `LoggerCollection` so this is a good solution

PR #12151, opened 2022-03-01
Deprecate `Trainer.devices` in favor of `Trainer.num_devices` and  `Trainer.device_ids` 

DuYicong515:
## What does this PR do?
Part of #12171 

N/A

## Before submitting

## PR review

cc @four4fish  @awaelchli

### DIFF
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -2012,7 +2012,7 @@ class Trainer(
 
     @property
     def device_ids(self) -> List[int]:
-        """List of device indexes."""
+        """List of device indexes per node."""
         devices = getattr(self.strategy, "parallel_devices", [self.strategy.root_device])
         device_ids = []
         for idx, device in enumerate(devices):
@@ -2024,6 +2024,7 @@ class Trainer(
 
     @property
     def num_devices(self) -> int:
+        """Number of devices per node."""
         return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
 
     @property
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -58,7 +58,7 @@ from pytorch_lightning.profiler import (
     SimpleProfiler,
     XLAProfiler,
 )
-from pytorch_lightning.strategies import ParallelStrategy, SingleDeviceStrategy, Strategy
+from pytorch_lightning.strategies import ParallelStrategy, Strategy
 from pytorch_lightning.strategies.ddp_spawn import DDPSpawnStrategy
 from pytorch_lightning.trainer.callback_hook import TrainerCallbackHookMixin
 from pytorch_lightning.trainer.configuration_validator import verify_loop_configurations
@@ -2012,7 +2012,7 @@ class Trainer(
 
     @property
     def device_ids(self) -> List[int]:
-        """List of device indexes per node."""
+        """List of device indexes."""
         devices = getattr(self.strategy, "parallel_devices", [self.strategy.root_device])
         device_ids = []
         for idx, device in enumerate(devices):
@@ -2024,12 +2024,7 @@ class Trainer(
 
     @property
     def num_devices(self) -> int:
-        """Number of devices per node."""
-        if isinstance(self.strategy, SingleDeviceStrategy):
-            return 1
-        elif isinstance(self.strategy, ParallelStrategy):
-            return len(self.strategy.parallel_devices)
-        return 0
+        return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
 
     @property
     def num_processes(self) -> int:
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -2024,7 +2024,7 @@ class Trainer(
 
     @property
     def num_devices(self) -> int:
-        """Number of devices per node."""
+        """Number of devices the trainer uses per node."""
         return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
 
     @property
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -2025,7 +2025,7 @@ class Trainer(
     @property
     def num_devices(self) -> int:
         """Number of devices the trainer uses per node."""
-        return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
+        return len(self.device_ids)
 
     @property
     def num_processes(self) -> int:
@@ -2048,7 +2048,7 @@ class Trainer(
         return self._accelerator_connector.num_gpus
 
     @property
-    def devices(self) -> Optional[Union[List[int], str, int]]:
+    def devices(self) -> int:
         rank_zero_deprecation(
             "`Trainer.devices` was deprecated in v1.6 and will be removed in v1.8."
             " Please use `Trainer.num_devices` or `Trainer.device_ids` to get device information instead."
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -58,7 +58,7 @@ from pytorch_lightning.profiler import (
     SimpleProfiler,
     XLAProfiler,
 )
-from pytorch_lightning.strategies import ParallelStrategy, Strategy
+from pytorch_lightning.strategies import ParallelStrategy, SingleDeviceStrategy, Strategy
 from pytorch_lightning.strategies.ddp_spawn import DDPSpawnStrategy
 from pytorch_lightning.trainer.callback_hook import TrainerCallbackHookMixin
 from pytorch_lightning.trainer.configuration_validator import verify_loop_configurations
@@ -2011,6 +2011,18 @@ class Trainer(
         return getattr(self.strategy, "num_nodes", 1)
 
     @property
+    def device_ids(self) -> List[int]:
+        if isinstance(self.strategy, ParallelStrategy):
+            return [torch._utils._get_device_index(device, allow_cpu=True) for device in self.strategy.parallel_devices]
+        elif isinstance(self.strategy, SingleDeviceStrategy):
+            return [torch._utils._get_device_index(self.strategy.root_device, allow_cpu=True)]
+        return []
+
+    @property
+    def num_devices(self) -> int:
+        return len(self.device_ids)
+
+    @property
     def num_processes(self) -> int:
         return self._accelerator_connector.num_processes
 
@@ -2032,7 +2044,11 @@ class Trainer(
 
     @property
     def devices(self) -> Optional[Union[List[int], str, int]]:
-        return self._accelerator_connector.devices
+        rank_zero_deprecation(
+            "`Trainer.devices` was deprecated in v1.6 and will be removed in v1.8."
+            " Please use `Trainer.num_devices` or `Trainer.device_ids` to get device information instead."
+        )
+        return self.num_devices
 
     @property
     def data_parallel_device_ids(self) -> Optional[List[int]]:
--- a/pytorch_lightning/trainer/trainer.py
+++ b/pytorch_lightning/trainer/trainer.py
@@ -58,7 +58,7 @@ from pytorch_lightning.profiler import (
     SimpleProfiler,
     XLAProfiler,
 )
-from pytorch_lightning.strategies import ParallelStrategy, Strategy
+from pytorch_lightning.strategies import ParallelStrategy, SingleDeviceStrategy, Strategy
 from pytorch_lightning.strategies.ddp_spawn import DDPSpawnStrategy
 from pytorch_lightning.trainer.callback_hook import TrainerCallbackHookMixin
 from pytorch_lightning.trainer.configuration_validator import verify_loop_configurations
@@ -2012,6 +2012,7 @@ class Trainer(
 
     @property
     def device_ids(self) -> List[int]:
+        """List of device indexes per node."""
         devices = getattr(self.strategy, "parallel_devices", [self.strategy.root_device])
         device_ids = []
         for idx, device in enumerate(devices):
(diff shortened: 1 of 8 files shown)


pytorch_lightning/trainer/trainer.py
```diff
@@ -2026,6 +2026,21 @@ def should_rank_save_checkpoint(self) -> bool:
...
+        for idx, device in enumerate(devices):
+            if isinstance(device, torch.device):
+                device_ids.append(device.index or idx)
+            elif isinstance(device, int):
+                device_ids.append(device)
+        return device_ids
+
+    @property
+    def num_devices(self) -> int:
+        return len(self.device_ids)
```
rohitgr7, line 2042: can it rely on `trainer._accelerator_connector.devices` here? since we are only concerned with the count, and not the actual devices. `device_ids` does some additional minor computation.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2048,7 +2063,11 @@ def num_gpus(self) -> int:
 
     @property
     def devices(self) -> Optional[Union[List[int], str, int]]:
-        return self._accelerator_connector.devices
+        rank_zero_deprecation(
```
rohitgr7, line 2066: can you add a deprecation test?

rohitgr7 (approved):

pytorch_lightning/trainer/trainer.py
DuYicong515: I am planning to remove devices from `AcceleratorConnector` since that's not internally used by `AcceleratorConnector` and it's not meant to be a public property. 

But I could definitely implement the same logic here so the computations are more efficient.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2026,6 +2026,21 @@ def should_rank_save_checkpoint(self) -> bool:
...
+        device_ids = []
+        for idx, device in enumerate(devices):
+            if isinstance(device, torch.device):
+                device_ids.append(device.index or idx)
+            elif isinstance(device, int):
+                device_ids.append(device)
+        return device_ids
+
+    @property
+    def num_devices(self) -> int:
```
ananthsub, line 2041: Please add docstrings for these. The docstring should clarify that that these are the device_ids *per node*

ananthsub (approved):

pytorch_lightning/trainer/trainer.py
DuYicong515: Added in `tests/trainer/test_trainer.py`

pytorch_lightning/trainer/trainer.py
DuYicong515: Added!

pytorch_lightning/trainer/trainer.py
DuYicong515: Moved the logic here. Also in tests checked they are the same with len(device_ids)

tests/trainer/test_trainer.py
```diff
@@ -2146,3 +2146,26 @@ def test_dataloaders_are_not_loaded_if_disabled_through_limit_batches(running_st
...
+    [
+        ({}, [0]),
+        ({"devices": 1}, [0]),
+        ({"accelerator": "gpu", "devices": 1}, [0]),
+        ({"strategy": "ddp", "devices": 1}, [0]),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": 1}, [0]),
+        ({"strategy": "ddp", "devices": 2}, [0, 1]),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": 2}, [0, 1]),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": [2]}, [2]),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": [0, 2]}, [0, 2]),
```
kaushikb11, line 2162: It's not required to have `strategy` as a parameter for the test

tests/deprecated_api/test_remove_1-8.py
```diff
@@ -688,3 +688,29 @@ def on_load_checkpoint(self, checkpoint):
...
+    [
+        ({}, 1),
+        ({"devices": 1}, 1),
+        ({"accelerator": "gpu", "devices": 1}, 1),
+        ({"strategy": "ddp", "devices": 1}, 1),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": 1}, 1),
+        ({"strategy": "ddp", "devices": 2}, 2),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": 2}, 2),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": [2]}, 1),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": [0, 2]}, 2),
```
kaushikb11, line 704: Same here

kaushikb11 (approved):

tests/trainer/test_trainer.py
```diff
@@ -2146,3 +2146,26 @@ def test_dataloaders_are_not_loaded_if_disabled_through_limit_batches(running_st
...
+@pytest.mark.parametrize(
+    ["trainer_kwargs", "expected_device_ids"],
+    [
+        ({}, [0]),
+        ({"devices": 1}, [0]),
+        ({"accelerator": "gpu", "devices": 1}, [0]),
+        ({"strategy": "ddp", "devices": 1}, [0]),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": 1}, [0]),
+        ({"strategy": "ddp", "devices": 2}, [0, 1]),
+        ({"strategy": "ddp", "accelerator": "gpu", "devices": 2}, [0, 1]),
```
kaushikb11, line 2160: Could add set of kwargs with strings being passed to `devices`

tests/trainer/test_trainer.py
```diff
@@ -2146,3 +2146,26 @@ def test_dataloaders_are_not_loaded_if_disabled_through_limit_batches(running_st
...
     )
     assert dl is None
+
+
+@pytest.mark.parametrize(
+    ["trainer_kwargs", "expected_device_ids"],
+    [
+        ({}, [0]),
+        ({"devices": 1}, [0]),
+        ({"accelerator": "gpu", "devices": 1}, [0]),
```
kaushikb11, line 2156: Please add tests for TPUs and IPUs, you could mock the availability

tests/trainer/test_trainer.py
```diff
@@ -2146,3 +2146,26 @@ def test_dataloaders_are_not_loaded_if_disabled_through_limit_batches(running_st
         else getattr(trainer, f"{dl_prefix}_dataloaders")
     )
     assert dl is None
+
+
+@pytest.mark.parametrize(
+    ["trainer_kwargs", "expected_device_ids"],
```
four4fish, line 2152: Can we test num_devices as well please?

pytorch_lightning/trainer/trainer.py
```diff
@@ -2026,6 +2026,27 @@ def should_rank_save_checkpoint(self) -> bool:
     def num_nodes(self) -> int:
         return getattr(self.strategy, "num_nodes", 1)
 
+    @property
+    def device_ids(self) -> List[int]:
+        """List of device indexes per node."""
```
kaushikb11, line 2031:
```suggestion
        """List of device indexes"""
```

tests/trainer/test_trainer.py
DuYicong515: assert len(trainer.device_ids) == trainer.num_devices should already test it. See L2171

four4fish: Oh  I see! nice!

pytorch_lightning/trainer/trainer.py
```diff
@@ -2026,6 +2026,27 @@ def should_rank_save_checkpoint(self) -> bool:
...
+        return device_ids
+
+    @property
+    def num_devices(self) -> int:
+        """Number of devices per node."""
+        if isinstance(self.strategy, SingleDeviceStrategy):
+            return 1
+        elif isinstance(self.strategy, ParallelStrategy):
+            return len(self.strategy.parallel_devices)
+        return 0
```
ethanwharris, line 2048: Is this line reachable? Doesn't seem to come up in tests. Might be cleaner to just return `len(parallel_devices)` for parallel strategy or 1 otherwise

ethanwharris (approved):
LGTM, small comment

pytorch_lightning/trainer/trainer.py
```diff
@@ -2026,6 +2026,22 @@ def should_rank_save_checkpoint(self) -> bool:
...
+        for idx, device in enumerate(devices):
+            if isinstance(device, torch.device):
+                device_ids.append(device.index or idx)
+            elif isinstance(device, int):
+                device_ids.append(device)
+        return device_ids
+
+    @property
+    def num_devices(self) -> int:
+        return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
```
ananthsub, line 2043: did the docstrings get lost? please clarify that these properties are per node
```suggestion
    def device_ids(self) -> List[int]:
        """List of device indexes used per node."""
        devices = getattr(self.strategy, "parallel_devices", [self.strategy.root_device])
        device_ids = []
        for idx, device in enumerate(devices):
            if isinstance(device, torch.device):
                device_ids.append(device.index or idx)
            elif isinstance(device, int):
                device_ids.append(device)
        return device_ids

    @property
    def num_devices(self) -> int:
        """ Number of devices the trainer uses per node. """
        return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
```

tests/trainer/test_trainer.py
DuYicong515: The TPU have a lot of things to mock. It uses functions like `xm.xla_device` frequently not only we have to mock `_TPU_AVAILABLE`. I put more test cases in test_tpu.py and only run when we have tpus available. Should be similar for testing purposes. 

let me know what you think!

 https://github.com/PyTorchLightning/pytorch-lightning/blob/15364c18c81909e213f1910a57982f100f8a356d/pytorch_lightning/strategies/single_tpu.py#L44

pytorch_lightning/trainer/trainer.py
ananthsub: also shouldn't these properties be kept in sync? why isn't `num_devices` implemented like this?
```py
    return max(len(self.device_ids), 1)
```
?

pytorch_lightning/trainer/trainer.py
DuYicong515: Changed! Thanks for the advice.

pytorch_lightning/trainer/trainer.py
DuYicong515: I received a comment previously about `self.device_ids` requiring additional computations. 
https://github.com/PyTorchLightning/pytorch-lightning/pull/12151#discussion_r817486457

I also add a test case to asserting `self.num_devices == len(expected_device_ids) == len(self.device_ids)`

I'm okay with both weighing more towards logic in sync or efficient computations.

pytorch_lightning/trainer/trainer.py
DuYicong515: I feel per node is more clear? Thoughts? cc @ananthsub

tests/deprecated_api/test_remove_1-8.py
```diff
@@ -688,3 +689,38 @@ def on_load_checkpoint(self, checkpoint):
         " v1.6 and will be removed in v1.8. Use `load_state_dict` instead."
     ):
         trainer.fit(model)
+
+
+@pytest.mark.parametrize(
```
rohitgr7, line 694: I don't think all of these are required here since they are already tested in other files. Just take one of these cases and check the deprecated_call as you have already done at the end.

tests/trainer/test_trainer.py
rohitgr7: I agree! for TPU it might not be that trivial. You can test it on TPU directly.

four4fish (approved):

pytorch_lightning/trainer/trainer.py
ananthsub: I agree. This property is ambiguously named, so the comments should clarify that it's per node, and not the global list of device indices being used.

pytorch_lightning/trainer/trainer.py
```diff
@@ -2026,6 +2026,23 @@ def should_rank_save_checkpoint(self) -> bool:
...
+            if isinstance(device, torch.device):
+                device_ids.append(device.index or idx)
+            elif isinstance(device, int):
+                device_ids.append(device)
+        return device_ids
+
+    @property
+    def num_devices(self) -> int:
+        """Number of devices the trainer uses per node."""
+        return len(self.strategy.parallel_devices) if isinstance(self.strategy, ParallelStrategy) else 1
```
ananthsub, line 2044: my assumption is that `num_devices` will be infrequently accessed, so I would prefer we rely on `device_ids` for the length to avoid these falling out of sync. If we need to optimize the performance, we can always do so without it being a breaking change.

tchaton (approved):
LGTM !

Borda (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2010,6 +2010,23 @@ def should_rank_save_checkpoint(self) -> bool:
     def num_nodes(self) -> int:
         return getattr(self.strategy, "num_nodes", 1)
 
+    @property
+    def device_ids(self) -> List[int]:
```
awaelchli, line 2014: I would have preferred if this was implemented on the accelerator connector. It is IMO not the responsibility of the Trainer to compute the device_ids off the strategy. Note the the strategy is owned by the connector, not the trainer. As the trainer has already many responsibilities, we should delegate as much as possible to the concrete components.

awaelchli (commented):
Apologies for the delayed review.

pytorch_lightning/trainer/trainer.py
DuYicong515: Thanks for reviewing @awaelchli!

My personal thoughts: the accelerator connector should stay internal, and accelerator connector should not expose those device related public properties, since they are easily computed off the strategy. 

> Note the the strategy is owned by the connector, not the trainer. 

Personally I feel the accelerator connector does a lot of checks/set up/initialization and configuration of the accelerator and strategy. After things are set up already, strategy/accelerator are already accessible on the Trainer. 

I think delegating computations using accelerator/strategy to the accelerator_connector will make us add many public properties to the internal-facing accelerator_connector. Deriving directly from strategy/accelerator directly are more straight forward, and they(strategy, accelerator) are very external facing as well. 

I'm very new to this and my feelings might be wrong. Feel free to add more thoughts to issue #12171!

PR #12030, opened 2022-02-21
Add support for pluggable Accelerators

kaushikb11:
## What does this PR do?

Supports #10687

```python
trainer = Trainer(accelerator=NewSOTAAccelerator(), devices=4)
```

Also, fixes #12044 and #10592

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @akihironitta @rohitgr7 @justusschock

PR #12180, opened 2022-03-02
Add `AcceleratorRegistry`

kaushikb11:
## What does this PR do?

Follow up to #12030 

```python
accelerator_name = "custom_accelerator"

class TestAccelerator(Accelerator):
    @staticmethod
    def parse_devices(devices):
        return devices

    @staticmethod
    def get_parallel_devices(devices):
        return ["foo"] * devices

    @staticmethod
    def auto_device_count():
        return 3

    @staticmethod
    def is_available():
        return True

AcceleratorRegistry.register(
    accelerator_name, CustomAccelerator, description=accelerator_description, param1="abc", param2=123
)

assert accelerator_name in AcceleratorRegistry
assert isinstance(AcceleratorRegistry.get(accelerator_name), TestAccelerator)

trainer = Trainer(accelerator="custom_accelerator", devices="auto")
assert isinstance(trainer.accelerator, TestAccelerator)
assert trainer._accelerator_connector.parallel_devices == ["foo"] * 3
 ```

## Before submitting

## PR review

## Did you have fun?

cc @borda @akihironitta @rohitgr7 @justusschock

### DIFF
--- a/pytorch_lightning/accelerators/__init__.py
+++ b/pytorch_lightning/accelerators/__init__.py
@@ -10,8 +10,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from pathlib import Path
+
 from pytorch_lightning.accelerators.accelerator import Accelerator  # noqa: F401
 from pytorch_lightning.accelerators.cpu import CPUAccelerator  # noqa: F401
 from pytorch_lightning.accelerators.gpu import GPUAccelerator  # noqa: F401
 from pytorch_lightning.accelerators.ipu import IPUAccelerator  # noqa: F401
+from pytorch_lightning.accelerators.registry import AcceleratorRegistry, register_accelerators  # noqa: F401
 from pytorch_lightning.accelerators.tpu import TPUAccelerator  # noqa: F401
+
+FILE_ROOT = Path(__file__).parent
+ACCELERATORS_BASE_MODULE = "pytorch_lightning.accelerators"
+
+register_accelerators(FILE_ROOT, ACCELERATORS_BASE_MODULE)
--- a/pytorch_lightning/accelerators/__init__.py
+++ b/pytorch_lightning/accelerators/__init__.py
@@ -10,16 +10,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from pathlib import Path
-
 from pytorch_lightning.accelerators.accelerator import Accelerator  # noqa: F401
 from pytorch_lightning.accelerators.cpu import CPUAccelerator  # noqa: F401
 from pytorch_lightning.accelerators.gpu import GPUAccelerator  # noqa: F401
 from pytorch_lightning.accelerators.ipu import IPUAccelerator  # noqa: F401
-from pytorch_lightning.accelerators.registry import AcceleratorRegistry, register_accelerators  # noqa: F401
+from pytorch_lightning.accelerators.registry import AcceleratorRegistry, call_register_accelerators  # noqa: F401
 from pytorch_lightning.accelerators.tpu import TPUAccelerator  # noqa: F401
 
-FILE_ROOT = Path(__file__).parent
 ACCELERATORS_BASE_MODULE = "pytorch_lightning.accelerators"
 
-register_accelerators(FILE_ROOT, ACCELERATORS_BASE_MODULE)
+call_register_accelerators(ACCELERATORS_BASE_MODULE)
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -77,4 +77,4 @@ class Accelerator(ABC):
 
     @classmethod
     def register_accelerators(cls, accelerator_registry: Dict) -> None:
-        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__class__.__name__}")
+        pass
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -75,7 +75,6 @@ class Accelerator(ABC):
     def is_available() -> bool:
         """Detect if the hardware is available."""
 
-    @staticmethod
-    @abstractmethod
-    def name() -> str:
-        """Name of the Accelerator."""
+    @classmethod
+    def register_accelerators(cls, accelerator_registry: Dict) -> None:
+        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__name__}")
--- a/pytorch_lightning/accelerators/accelerator.py
+++ b/pytorch_lightning/accelerators/accelerator.py
@@ -77,4 +77,4 @@ class Accelerator(ABC):
 
     @classmethod
     def register_accelerators(cls, accelerator_registry: Dict) -> None:
-        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__name__}")
+        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__class__.__name__}")
--- a/pytorch_lightning/accelerators/cpu.py
+++ b/pytorch_lightning/accelerators/cpu.py
@@ -63,7 +63,10 @@ class CPUAccelerator(Accelerator):
         """CPU is always available for execution."""
         return True
 
-    @staticmethod
-    def name() -> str:
-        """Name of the Accelerator."""
-        return "cpu"
+    @classmethod
+    def register_accelerators(cls, accelerator_registry: Dict) -> None:
+        accelerator_registry.register(
+            "cpu",
+            cls,
+            description=f"{cls.__class__.__name__}",
+        )
--- a/pytorch_lightning/accelerators/registry.py
+++ b/pytorch_lightning/accelerators/registry.py
@@ -1,0 +1,78 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import importlib
+from inspect import getmembers, isclass
+from pathlib import Path
+from typing import Any, List, Optional
+
+from pytorch_lightning.accelerators.accelerator import Accelerator
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+
+
+class _AcceleratorRegistry(dict):
+    def register(self, accelerator: Accelerator, name: Optional[str] = None, override: bool = False) -> Accelerator:
+        """Registers an accelerator mapped to a name.
+
+        Args:
+            accelerator: the accelerator to be mapped.
+            name: the name that identifies the provided accelerator.
+            override: Whether to override an existing key.
+        """
+        if name is None:
+            name = accelerator.name()
+        elif not isinstance(name, str):
+            raise TypeError(f"`name` must be a str, found {name}")
+
+        if name in self and not override:
+            raise MisconfigurationException(f"'{name}' is already present in the registry. HINT: Use `override=True`.")
+        self[name] = accelerator
+        return accelerator
+
+    def get(self, name: str, default: Optional[Any] = None) -> Any:
+        """Calls the registered Accelerator and returns the Accelerator object.
+
+        Args:
+            name (str): the name that identifies a Accelerator, e.g. "tpu"
+        """
+        if name in self:
+            accelerator = self[name]
+            return accelerator()
+
+        if default is not None:
+            return default
+
+        err_msg = "'{}' not found in registry. Available names: {}"
+        available_names = ", ".join(sorted(self.keys())) or "none"
+        raise KeyError(err_msg.format(name, available_names))
+
+    def remove(self, name: str) -> None:
+        """Removes the registered accelerator by name."""
+        self.pop(name)
+
+    def available_accelerators(self) -> List:
+        """Returns a list of registered accelerators."""
+        return list(self.keys())
+
+    def __str__(self) -> str:
+        return "Registered Accelerators: {}".format(", ".join(self.keys()))
+
+
+AcceleratorRegistry = _AcceleratorRegistry()
+
+
+def register_accelerators(root: Path, base_module: str) -> None:
+    module = importlib.import_module(base_module)
+    for _, mod in getmembers(module, isclass):
+        if issubclass(mod, Accelerator) and mod is not Accelerator:
+            AcceleratorRegistry.register(mod)
--- a/pytorch_lightning/accelerators/registry.py
+++ b/pytorch_lightning/accelerators/registry.py
@@ -48,7 +48,7 @@ class _AcceleratorRegistry(dict):
         self,
         name: str,
         accelerator: Optional[Callable] = None,
-        description: Optional[str] = None,
+        description: str = "",
         override: bool = False,
         **init_params: Any,
     ) -> Callable:
--- a/pytorch_lightning/accelerators/registry.py
+++ b/pytorch_lightning/accelerators/registry.py
@@ -69,6 +69,7 @@ class _AcceleratorRegistry(dict):
 
         data: Dict[str, Any] = {}
 
+        data["description"] = description
         data["init_params"] = init_params
 
         def do_register(name: str, accelerator: Callable) -> Callable:
--- a/pytorch_lightning/accelerators/registry.py
+++ b/pytorch_lightning/accelerators/registry.py
@@ -41,7 +41,7 @@ class _AcceleratorRegistry(dict):
 
         or
 
-        AcceleratorRegistry.register("sota", SOTAAAccelerator, description="Custom sota accelerator", a=1, b=True)
+        AcceleratorRegistry.register("sota", SOTAAccelerator, description="Custom sota accelerator", a=1, b=True)
     """
 
     def register(
--- a/pytorch_lightning/accelerators/registry.py
+++ b/pytorch_lightning/accelerators/registry.py
@@ -96,7 +96,7 @@ class _AcceleratorRegistry(dict):
             return default
 
         err_msg = "'{}' not found in registry. Available names: {}"
-        available_names = ", ".join(sorted(self.keys())) or "none"
+        available_names = self.available_accelerators()
         raise KeyError(err_msg.format(name, available_names))
 
     def remove(self, name: str) -> None:
--- a/pytorch_lightning/accelerators/registry.py
+++ b/pytorch_lightning/accelerators/registry.py
@@ -13,41 +13,85 @@
 # limitations under the License.
 import importlib
 from inspect import getmembers, isclass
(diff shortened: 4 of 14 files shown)

### CODE REVIEW

File: pytorch_lightning/accelerators/registry.py [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+    def register(self, accelerator: Accelerator, name: Optional[str] = None, override: bool = False) -> Accelerator:
+        """Registers an accelerator mapped to a name.
+
+        Args:
+            accelerator: the accelerator to be mapped.
+            name: the name that identifies the provided accelerator.
+            override: Whether to override an existing key.
+        """
+        if name is None:
+            name = accelerator.name()
```
rohitgr7, line 33: strategy is using class variables instead of a property. Should we choose one of them for consistency?

tests/accelerators/test_accelerator_registry.py
```diff
@@ -0,0 +1,57 @@
...
+        @staticmethod
+        def name():
+            return accelerator_name
+
+    AcceleratorRegistry.register(TestAccelerator)
+
+    assert accelerator_name in AcceleratorRegistry
+    assert isinstance(AcceleratorRegistry.get(accelerator_name), TestAccelerator)
+
+    trainer = Trainer(accelerator="custom_accelerator", devices="auto")
```
rohitgr7, line 48:
```suggestion
    trainer = Trainer(accelerator=accelerator_name, devices="auto")
```

tests/accelerators/test_accelerator_registry.py
```diff
@@ -0,0 +1,57 @@
...
+        def auto_device_count():
+            return 3
+
+        @staticmethod
+        def is_available():
+            return True
+
+        @staticmethod
+        def name():
+            return accelerator_name
```
ananthsub, line 41: Now the name isn't used anywhere outside of the accelerator registry. We don't need to require it from the API/interface, which would reduce the overhead. Instead, I think we shouldoffer the register_accelerator classmethod  on the accelerator to allow for customization. In the future, we could have multiple configurations per accelerator. One example is for GPUs, and enabling inter-batch parallelism 

https://github.com/PyTorchLightning/pytorch-lightning/blob/d923dff62763b2096bb2df1b1bbfc7f8540b4bbc/pytorch_lightning/strategies/strategy.py#L445-L447

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,78 @@
...
+        elif not isinstance(name, str):
+            raise TypeError(f"`name` must be a str, found {name}")
+
+        if name in self and not override:
```
four4fish, line 37: Why do we need this override? aka why do we want to allow overrides? Will it be less error prone to require register accelerator with different names?

four4fish (approved):
I like this! Provides more flexibilities for customized accelerator and it's consistent with Strategy behavior. 

Do you think we can remove AcceleratorType after this PR?

tests/accelerators/test_accelerator_registry.py
kaushikb11: @ananthsub Couple of questions.

1. How do you think we could have multiple of configurations? In Strategies/`register_strategies`, we have different configurations with different values being passed to class's constructor.

2. Do you think we should go ahead with having `name` as a class attribute or have it being an attribute?

ananthsub: > How do you think we could have multiple of configurations? In Strategies/register_strategies, we have different configurations with different values being passed to class's constructor.

It'd be the same here. For instance, the GPU accelerator could accept a flag like `enable_inter_batch_parallelism: bool = False` to trigger this feature: https://github.com/PyTorchLightning/pytorch-lightning/blob/bc1c8b926c5072f58f42ad4b7413a8ef5c904c85/pytorch_lightning/loops/fit_loop.py#L350-L353

to me, this is no different than how someone could specify DDP find unused parameters on the DDP strategy.

> Do you think we should go ahead with having name as a class attribute or have it being an attribute?

Is name referenced anywhere outside of the registry? If it isn't, then we don't need to require this attribute at all.

pytorch_lightning/accelerators/registry.py
kaushikb11: We need `name` to be an abstract method. We wish to make it mandatory for the users to define it, so they could pass the name to the `accelerator` flag to the Trainer.

```python
trainer = Trainer(accelerator="sota_accelerator", devices="auto")
```

kaushikb11: Also, the reason to make it a static method is to get the name without instantiating the Accelerator class

justusschock: @kaushikb11, I would still prefer class variables. They are also static, so you can access them without instantiation. And more important: They are the same for every instance of the class. Whereas the function approach could still contain some logic to change this dynamically. And I think, we could check that this has been set during the accelerator's init?

tests/accelerators/test_accelerator_registry.py
justusschock: It wouldn't be a difference whether you require the name or the class-method to register it, except for the class-method to be more flexible since it allows multiple configurations :)

pytorch_lightning/accelerators/registry.py
kaushikb11: @justusschock But the end goal is 

```python
accelerator_name = "custom_accelerator"

class TestAccelerator(Accelerator):
    @staticmethod
    def parse_devices(devices):
        return devices

    @staticmethod
    def get_parallel_devices(devices):
        return ["foo"] * devices

    @staticmethod
    def auto_device_count():
        return 3

    @staticmethod
    def is_available():
        return True

    @staticmethod
    def name():
        return accelerator_name

AcceleratorRegistry.register(TestAccelerator)

trainer = Trainer(accelerator="custom_accelerator", devices="auto")
assert isinstance(trainer.accelerator, TestAccelerator)
```

How would we enforce users to define the class variable if they wish to use their custom accelerator?

tests/accelerators/test_accelerator_registry.py
justusschock: After discussing offline with @kaushikb11 , we both agreed that a `register_accelerators` approach would support everything we need and be more flexible compared to the name attribute or function :)

pytorch_lightning/accelerators/registry.py
justusschock: Resolved offline, see https://github.com/PyTorchLightning/pytorch-lightning/pull/12180#discussion_r832509497

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -75,7 +75,6 @@ def auto_device_count() -> int:
     def is_available() -> bool:
         """Detect if the hardware is available."""
 
-    @staticmethod
-    @abstractmethod
-    def name() -> str:
-        """Name of the Accelerator."""
+    @classmethod
+    def register_accelerators(cls, accelerator_registry: Dict) -> None:
+        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__name__}")
```
kaushikb11, line 80:
```suggestion
        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__class__.__name__}")
```

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+    Example::
+
+        @AcceleratorRegistry.register("sota", description="Custom sota accelerator", a=1, b=True)
+        class SOTAAccelerator(Accelerator):
+            def __init__(self, a, b):
+                ...
+
+        or
+
+        AcceleratorRegistry.register("sota", SOTAAAccelerator, description="Custom sota accelerator", a=1, b=True)
```
kaushikb11, line 44:
```suggestion
        AcceleratorRegistry.register("sota", SOTAAccelerator, description="Custom sota accelerator", a=1, b=True)
```

tests/accelerators/test_accelerator_registry.py
kaushikb11: @justusschock @ananthsub Updated!

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -75,7 +75,6 @@ def auto_device_count() -> int:
     def is_available() -> bool:
         """Detect if the hardware is available."""
 
-    @staticmethod
-    @abstractmethod
-    def name() -> str:
-        """Name of the Accelerator."""
+    @classmethod
+    def register_accelerators(cls, accelerator_registry: Dict) -> None:
+        raise NotImplementedError(f"`register_accelerators` is not implemented for {cls.__class__.__name__}")
```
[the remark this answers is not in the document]
kaushikb11: Updated!

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+        or
+
+        AcceleratorRegistry.register("sota", SOTAAccelerator, description="Custom sota accelerator", a=1, b=True)
+    """
+
+    def register(
+        self,
+        name: str,
+        accelerator: Optional[Callable] = None,
+        description: Optional[str] = None,
```
ananthsub, line 51:
```suggestion
        description: str = "",
```

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+            init_params: parameters to initialize the accelerator
+        """
+        if not (name is None or isinstance(name, str)):
+            raise TypeError(f"`name` must be a str, found {name}")
+
+        if name in self and not override:
+            raise MisconfigurationException(f"'{name}' is already present in the registry. HINT: Use `override=True`.")
+
+        data: Dict[str, Any] = {}
+        data["description"] = description if description is not None else ""
```
ananthsub, line 71:
```suggestion
```

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        if name in self:
+            data = self[name]
+            return data["accelerator"](**data["init_params"])
+
+        if default is not None:
+            return default
+
+        err_msg = "'{}' not found in registry. Available names: {}"
+        available_names = ", ".join(sorted(self.keys())) or "none"
```
ananthsub, line 100: None shouldn't be listed. Otherwise people may think accelerator=None is supported.

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+        raise KeyError(err_msg.format(name, available_names))
+
+    def remove(self, name: str) -> None:
+        """Removes the registered accelerator by name."""
+        self.pop(name)
+
+    def available_accelerators(self) -> List:
```
ananthsub, line 107:
```suggestion
    def available_accelerators(self) -> List[str]:
```

pytorch_lightning/accelerators/registry.py
ananthsub, line 100:
```suggestion
        available_names = self.available_accelerators()
```

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+        """Returns a list of registered accelerators."""
+        return list(self.keys())
+
+    def __str__(self) -> str:
+        return "Registered Accelerators: {}".format(", ".join(self.keys()))
```
ananthsub, line 112:
```suggestion
        return "Registered Accelerators: {}".format(", ".join(self.available_accelerators()))
```

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,121 @@
...
+        return list(self.keys())
+
+    def __str__(self) -> str:
+        return "Registered Accelerators: {}".format(", ".join(self.available_accelerators()))
+
+
+AcceleratorRegistry = _AcceleratorRegistry()
+
+
+def call_register_accelerators(base_module: str) -> None:
```
justusschock, line 117: maybe merge this with https://github.com/PyTorchLightning/pytorch-lightning/blob/7a6efb38b25d6d4de75a09199809f70e0b930488/pytorch_lightning/strategies/strategy_registry.py#L133 so that we have a single function taking the base_module, the registry and the type to register?

No need to fully duplicate this. We could also think of an abstract registry class, having this as a function so that we spare the registry argument (and get `self` instead).

justusschock (approved):

ananthsub (approved):

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -157,7 +158,7 @@ def __init__(
         # 1. Parsing flags
         # Get registered strategies, built-in accelerators and precision plugins
         self._registered_strategies = StrategyRegistry.available_strategies()
-        self._accelerator_types = ("tpu", "ipu", "gpu", "cpu")
+        self._accelerator_types = AcceleratorRegistry.available_accelerators()
```
Borda, line 161: I see the value of having it dynamically generated, gut it is not a friendly way users looking to code without really running it... :/

Borda (approved):

pytorch_lightning/accelerators/registry.py
kaushikb11: Could take a stab at it with a follow up PR, need the registry changes for Habana PR now

pytorch_lightning/accelerators/registry.py
kaushikb11: There could be a case where the user would wish to override the internal accelerator as well as keeping the api consistent with `register_strategies`

pytorch_lightning/trainer/connectors/accelerator_connector.py
ananthsub, line 161: is this used when "auto" is selected ? if so, we need to be very careful about the ordering inside of the registry

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+        **init_params: Any,
+    ) -> Callable:
+        """Registers a accelerator mapped to a name and with required metadata.
+
+        Args:
+            name : the name that identifies a accelerator, e.g. "gpu"
+            accelerator : accelerator class
+            description : accelerator description
+            override : overrides the registered accelerator, if True
+            init_params: parameters to initialize the accelerator
```
kaushikb11, line 62:
```suggestion
            init_params: parameters to initialize the Accelerator
```

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+    The motivation for having a AcceleratorRegistry is to make it convenient
+    for the Users to try different accelerators by passing mapped aliases
+    to the accelerator flag to the Trainer.
+
+    Example::
+
+        @AcceleratorRegistry.register("sota", description="Custom sota accelerator", a=1, b=True)
+        class SOTAAccelerator(Accelerator):
+            def __init__(self, a, b):
+                ...
```
carmocca, line 40: This doesn't work:

```python
  File "/Users/carmocca/git/pytorch-lightning/thing.py", line 5, in <module>
    class SOTAAccelerator:
TypeError: do_register() missing 1 required positional argument: 'accelerator'
```

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -75,7 +75,6 @@ def auto_device_count() -> int:
     def is_available() -> bool:
         """Detect if the hardware is available."""
 
-    @staticmethod
-    @abstractmethod
-    def name() -> str:
```
carmocca, line 80: I don't get the reasoning for removing the `name`.

If we kept it, the registry could use it to automatically define the name for the class

pytorch_lightning/utilities/registry.py
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import inspect
+from typing import Any
+
+
+def _is_register_method_overridden(mod: type, base_cls: Any, method: str) -> bool:
```
carmocca, line 18: Why not use `is_overridden`?

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+
+        or
+
+        AcceleratorRegistry.register("sota", SOTAAccelerator, description="Custom sota accelerator", a=1, b=True)
+    """
+
+    def register(
+        self,
+        name: str,
+        accelerator: Optional[Callable] = None,
```
carmocca, line 50: The type should be `Optional[Type]`, not Callable

pytorch_lightning/accelerators/__init__.py
```diff
@@ -14,4 +14,9 @@
 from pytorch_lightning.accelerators.cpu import CPUAccelerator  # noqa: F401
 from pytorch_lightning.accelerators.gpu import GPUAccelerator  # noqa: F401
 from pytorch_lightning.accelerators.ipu import IPUAccelerator  # noqa: F401
+from pytorch_lightning.accelerators.registry import AcceleratorRegistry, call_register_accelerators  # noqa: F401
 from pytorch_lightning.accelerators.tpu import TPUAccelerator  # noqa: F401
+
+ACCELERATORS_BASE_MODULE = "pytorch_lightning.accelerators"
```
carmocca, line 20: Why expose this string?

If it was to avoid registering these by default, it would not work because to change the path you'd need to import the variable

And at import time, they would get registered anyways.

pytorch_lightning/accelerators/cpu.py
```diff
@@ -63,7 +63,10 @@ def is_available() -> bool:
         """CPU is always available for execution."""
         return True
 
-    @staticmethod
-    def name() -> str:
-        """Name of the Accelerator."""
-        return "cpu"
+    @classmethod
+    def register_accelerators(cls, accelerator_registry: Dict) -> None:
```
carmocca, line 67: The argument should be typed as `_AcceleratorRegistry`, not `Dict`

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -75,7 +75,6 @@ def auto_device_count() -> int:
...
-        """Name of the Accelerator."""
+    @classmethod
+    def register_accelerators(cls, accelerator_registry: Dict) -> None:
```
awaelchli, line 79: Typing is wrong. should be the actual accelerator registry

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+from pytorch_lightning.utilities.registry import _is_register_method_overridden
+
+
+class _AcceleratorRegistry(dict):
+    """This class is a Registry that stores information about the Accelerators.
+
+    The Accelerators are mapped to strings. These strings are names that identify
+    an accelerator, e.g., "gpu". It also returns Optional description and
+    parameters to initialize the Accelerator, which were defined during the
+    registration.
```
awaelchli, line 29: ... an optional description ...

pytorch_lightning/accelerators/registry.py
```diff
@@ -0,0 +1,122 @@
...
+
+    The motivation for having a AcceleratorRegistry is to make it convenient
+    for the Users to try different accelerators by passing mapped aliases
```
awaelchli, line 32: users capitalized

pytorch_lightning/accelerators/accelerator.py
kaushikb11: Even if it subclasses Dict?

pytorch_lightning/accelerators/accelerator.py
kaushikb11: Related discussion https://github.com/PyTorchLightning/pytorch-lightning/pull/12180#discussion_r818023970

PR #12182, opened 2022-03-02
Have the outputs match the loops format

### LINKED ISSUE #9737: RFC: Change input format of `training_epoch_end` hook in case of multiple optimizers or TBPTT

Change the format in the outputs that the `training_epoch_end` and `training_step_end` hooks receive. This is a breaking change and requires careful deprecation. It will affect users with either multiple optimizers or truncated backprop IF they also implement the aforementioned hooks.

### Motivation

When using multiple optimizers or truncated backprop (or both) the inputs passed to the `training_epoch_end` hook are the outputs from the training step arranged in a 2D or 3D nested list of lists. The shape of this multi-dim array is
`(num_optimizers, num_batches, num_tbptt_splits)` in the general case and when num_optimizers=1 or truncated backprop is deactivated, the dimensions get squeezed. The problem with this is that the order of these dimensions does not correspond to the loop structure: 

```python
for batch in dataloader:
    for split in batch:
        for opt in optimizers:
            ...
```
It means this output format will never generalize for loop customization as the ordering is arbitrary. 
Currently, this [permutation of dimensions is hard-coded](https://github.com/PyTorchLightning/pytorch-lightning/blob/198aa852ef0d45a26edcf2ec5434cc9df614063b/pytorch_lightning/loops/epoch/training_epoch_loop.py#L329-L340) and will break for custom loops. 

### Pitch

Deprecate the current format and make it consistent with the loop structure, meaning, we adopt the format
`(num_batches, num_tbptt_splits, num_optimizers)`. This corresponds 1:1 with the loop structure. The standardization here will unblock custom loops with arbitrary nesting and output aggregation with less effort.

### Proposed deprecation plan:

1. In 1.5, log a message that the format will change in the future (if using multiple optimizers and hook is overridden)
2. In 1.5, the user will change their code given our recommendation and will signal this by adding a new argument to the hook: 

    ```python
    def training_step_end(self, outputs, new_format=True):
        ...

    def training_epoch_end(self, outputs, new_format=True):
        ...
    ```
    This will trigger the loop to call the hook with the new format for `outputs` instead of the old one.

3. In 1.7, the new format will be used unconditionally and be a breaking change if users did not adapt their code until now. The argument `new_format=True/False` will become ineffective and can be removed again.

Note: The only purpose the `new_format` argument serves is for inspection by our loop to infer what the user expects to get. We will not pass a value so the user must make it a keyword argument.

### Alternative deprecation plan:

Instead of a `new_format` argument in the signature, one can also add properties to the LightningModule:

```python
class MyLightningModule(LightningModule):
    def __init__(self):
         self.v1_7_training_epoch_end_format = True
```

<sub>

</sub>

carmocca:
## What does this PR do?

Fixes #9737

When multiple optimizers are used, the previous format allowed knowing which optimizers stepped by looking at the structure. This is no longer possible with the new format. A test relying on this behaviour was updated.

## Before submitting

## PR review

cc @borda @tchaton @rohitgr7 @carmocca @justusschock @ananthsub @ninginthecloud


tests/loops/optimization/test_optimizer_loop.py
```diff
@@ -109,16 +109,13 @@ class CurrentModel(BoringModel):
         def training_step(self, batch, batch_idx, optimizer_idx):
             return super().training_step(batch, batch_idx)
 
-        def training_epoch_end(self, outputs):
-            assert len(outputs[0]) == sum(idx == 0 for idx, _ in expected)
-            assert len(outputs[1]) == sum(idx == 1 for idx, _ in expected)
```
carmocca, line 114: We were using the `training_epoch_end` outputs format as a side effect to check the optimizer indices called, this is not possible anymore and can be considered a breaking change.

The expected format is still checked after `fit` below

tests/loops/epoch/test_training_epoch_loop.py
```diff
@@ -29,122 +32,213 @@
...
-    ],
-)
-def test_prepare_outputs_training_epoch_end_automatic(num_optimizers, batch_outputs, expected):
-    """Test that the loop converts the nested lists of outputs to the format that the `training_epoch_end` hook
-    currently expects in the case of automatic optimization."""
-    prepared = TrainingEpochLoop._prepare_outputs_training_epoch_end(
-        batch_outputs,
-        automatic=True,
-        num_optimizers=num_optimizers,
+class TestPrepareOutputs:
```
carmocca, line 35: I rewrote this into a class to reuse code, but the parametrizations aren't changed.

I kept the tests using the old format, and added new tests for the new one.

tests/loops/optimization/test_optimizer_loop.py
awaelchli: note, if one needs the index of the optimizer, then that can be put as an output in the training step and it will show up here.

pytorch_lightning/core/lightning.py
```diff
@@ -705,10 +705,9 @@ def training_epoch_end(self, outputs: EPOCH_OUTPUT) -> None:
             training_epoch_end(train_outs)
 
         Args:
-            outputs: List of outputs you defined in :meth:`training_step`.
-                If there are multiple optimizers, it is a list containing a list of outputs for each optimizer.
-                If using ``truncated_bptt_steps > 1``, each element is a list of outputs corresponding to the outputs
-                of each processed split batch.
+            outputs: List of outputs you defined in :meth:`training_step`. If there are multiple optimizers or when
+                using ``truncated_bptt_steps > 0``, the lists have the dimensions
+                (n_batches, tbptt_steps, n_optimizers. Dimensions of length 1 are squeezed.
```
awaelchli, line 710:
```suggestion
                (n_batches, tbptt_steps, n_optimizers). Dimensions of length 1 are squeezed.
```

awaelchli (approved):
thanks for taking this on!

ananthsub (approved):

SeanNaren (approved):

PR #12184, opened 2022-03-02
Fix logging's step values when multiple dataloaders are used during evaluation

### LINKED ISSUE #11616: Inconsistent step values when using multiple val/test dataloaders when logged on step-level
## 🐛 Bug

Currently, the `LoggerConnector` handles the val_step index for step-level logging for val/test steps, due to which it shows incorrect step (x-axis) in case of multiple dataloaders.

### To Reproduce

<p>

```py
import torch
import time

from pytorch_lightning import LightningDataModule, LightningModule, Trainer


class RandomDataset(Dataset):
    def __init__(self, size: int, length: int):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        return {"loss": loss}

    def validation_step(self, batch, batch_idx, dataloader_idx=None):
        loss = self(batch).sum()
        self.log('valid_loss', loss, on_step=True)

        optimizer = torch.optim.SGD(self.layer.parameters(), lr=0.1)
        return optimizer

    def train_dataloader(self):
        return DataLoader(RandomDataset(32, 64))

    def val_dataloader(self):
        return [DataLoader(RandomDataset(32, 64)), DataLoader(RandomDataset(32, 64))]
```
</p>
</details>

```py
model = BoringModel()
trainer = Trainer(max_epochs=1)
trainer.fit(model)
```

Output:
The log for dataloader_idx=1 starts from 64, which is the step right after the first dataloader is completed.

Screenshot 2022-01-25 at 5 10 43 PM

Expected:
for each datalaoder, it should start from 0.

If we go a little more crazy and do:

```py
model = BoringModel()
trainer = Trainer(max_epochs=1)
trainer.fit(model)
trainer.validate(model)
```

Output:
Screenshot 2022-01-25 at 7 17 09 PM

Expected:
I think it should replace the logs created during fit in such a case.

### Proposal
If we consider the above issues:
I think a better solution would be to let `EvaluationLoop` track the logging step for each dataloader, instead of logger connector, in a list where each index represents the current dataloader and will be the only one that should be incremented.

But there can be another issue, where the user resets the dataloader after every n epochs, then in such a case, should we reset the list if we go with this? The reason is, we can't for sure say whether the new dataloaders will be the same/more/less, and if we don't reset, it might be a problem.

### Environment

- Python version (e.g., 3.9): 3.0
- OS (e.g., Linux): mac
- CUDA/cuDNN version:

### Additional context

Open to better alternative solutions.

cc @carmocca @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy @justusschock @ninginthecloud

rohitgr7:
## What does this PR do?

Fixes #11616

## Before submitting

## PR review

## Did you have fun?

cc @otaj @borda @carmocca @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy @justusschock @ninginthecloud


pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -137,35 +134,15 @@ def log_metrics(self, metrics: _OUT_DICT, step: Optional[int] = None) -> None:
...
     def _evaluation_epoch_end(self) -> None:
         results = self.trainer._results
         assert results is not None
         results.dataloader_idx = None
 
-    def update_eval_step_metrics(self) -> None:
+    def update_eval_step_metrics(self, step: int) -> None:
         assert not self._epoch_end_reached
-        if self.trainer.sanity_checking:
-            return
```
carmocca, line 162: Why remove this?

pytorch_lightning/loops/fit_loop.py
```diff
@@ -218,7 +218,8 @@ def reset(self) -> None:
     def on_run_start(self) -> None:  # type: ignore[override]
         """Calls the ``on_train_start`` hook."""
         # reset train dataloader and val dataloader
-        self.trainer.reset_train_val_dataloaders(self.trainer.lightning_module)
+        self.trainer.reset_train_dataloader(self.trainer.lightning_module)
+        self.epoch_loop.val_loop._reload_evaluation_dataloaders()
```
carmocca, line 222: Can we defer this call to the moment validation loop actually runs?

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
rohitgr7: moved to loop instead where we are managing the step values. Since we don't need to increment the step during sanity check

pytorch_lightning/loops/fit_loop.py
rohitgr7: sorry, what do you mean? I made this change for the case when `num_sanity_steps=0`, the validation dataloaders are loaded here, but we need to use the loop method to load them to ensure that the tracker is updated correctly.

tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -924,3 +925,57 @@ def test_rich_print_results(inputs, expected):
...
+    trainer.test(model)
+
+    for dl_idx in range(num_dataloaders):
+        suffix = f"/dataloader_idx_{dl_idx}" if num_dataloaders == 2 else ""
+        assert logger.logged_metrics[f"val_log_fit{suffix}"]["logged_values"] == list(range(limit_batches)) * max_epochs
+        assert logger.logged_metrics[f"val_log_fit{suffix}"]["logged_steps"] == list(range(limit_batches * max_epochs))
+        assert logger.logged_metrics[f"val_log_validate{suffix}"]["logged_values"] == list(range(limit_batches))
+        assert logger.logged_metrics[f"val_log_validate{suffix}"]["logged_values"] == list(range(limit_batches))
+        assert logger.logged_metrics[f"test_log{suffix}"]["logged_steps"] == list(range(limit_batches))
+        assert logger.logged_metrics[f"test_log{suffix}"]["logged_steps"] == list(range(limit_batches))
```
ethanwharris, line 981: Some duplication here:
```suggestion
        assert logger.logged_metrics[f"val_log_validate{suffix}"]["logged_values"] == list(range(limit_batches))
        assert logger.logged_metrics[f"test_log{suffix}"]["logged_steps"] == list(range(limit_batches))
```

ethanwharris (approved):

tests/trainer/logging_/test_eval_loop_logging.py
rohitgr7: good catch!!

pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -219,10 +220,17 @@ def _get_max_batches(self) -> List[int]:
...
+        dataloaders = None
+
         if self.trainer.testing:
             self.trainer.reset_test_dataloader()
+            dataloaders = self.trainer.test_dataloaders
         elif self.trainer.val_dataloaders is None or self.trainer._data_connector._should_reload_val_dl:
             self.trainer.reset_val_dataloader()
+            dataloaders = self.trainer.val_dataloaders
+
+        if dataloaders:
```
justusschock, line 232:
```suggestion
        if dataloaders is not None:
```

just a nit: but let's be precise :)

justusschock (approved):

pytorch_lightning/loops/dataloader/evaluation_loop.py
rohitgr7: did this way to make sure `[]` are handled too

pytorch_lightning/loops/fit_loop.py
carmocca: My question is, why call this here in `FitLoop.on_run_start` instead of on `FitLoop.EpochLoop.EvaluationLoop.on_run_start`?

pytorch_lightning/loops/fit_loop.py
```diff
@@ -205,7 +205,8 @@ def reset(self) -> None:
     def on_run_start(self) -> None:  # type: ignore[override]
         """Calls the ``on_train_start`` hook."""
         # reset train dataloader and val dataloader
-        self.trainer.reset_train_val_dataloaders(self.trainer.lightning_module)
```
carmocca, line 208: This method is now unused. Deprecate it?

tests/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -929,3 +930,57 @@ def test_rich_print_results(inputs, expected):
...
+def test_eval_step_logging(tmpdir, num_dataloaders):
+    class CustomLogger(CSVLogger):
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.logged_metrics = collections.defaultdict(lambda: collections.defaultdict(list))
+
+        def log_metrics(self, metrics, step):
+            for key in metrics:
+                self.logged_metrics[key]["logged_values"].append(metrics[key])
+                self.logged_metrics[key]["logged_steps"].append(step)
```
carmocca, line 945: Using the CSV logger to test this is a bit ugly.

Why not mock `logger.log_metrics` instead and test the call inputs?

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -287,3 +291,10 @@ def _should_track_batch_outputs_for_epoch_end(self) -> bool:
         if self.trainer.testing:
             return is_overridden("test_epoch_end", model)
         return is_overridden("validation_epoch_end", model)
+
+    def _initialize_batch_idx_tracker(self, num_dataloaders: int) -> None:
+        self._seen_batches = [0] * num_dataloaders
```
carmocca, line 296: shouldn't this initialize with the existing ptracking value in case of reload?

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -287,3 +291,10 @@ def _should_track_batch_outputs_for_epoch_end(self) -> bool:
...
+
+    def _increment_batch_idx_tracker(self, dataloader_idx: int) -> None:
+        if not self.trainer.sanity_checking:
+            self._seen_batches[dataloader_idx] += self.batch_progress.current.completed
```
carmocca, line 300: Why not do this one by one inside `advance` where you get the `dl_eval_step`?

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
```diff
@@ -135,7 +136,10 @@ def advance(  # type: ignore[override]
         self.batch_progress.increment_completed()
 
         # log batch metrics
-        self.trainer._logger_connector.update_eval_step_metrics()
+        if not self.trainer.sanity_checking:
+            dl_idx = kwargs.get("dataloader_idx", 0)
+            dl_eval_step = self._seen_batches[dl_idx] + self.batch_progress.current.completed - 1
```
carmocca, line 141: Could this just be a += 1?

pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -37,8 +36,6 @@ def __init__(self, trainer: "pl.Trainer", log_gpu_memory: Optional[str] = None)
                 "Please monitor GPU stats with the `DeviceStatsMonitor` callback directly instead."
             )
         self.log_gpu_memory = log_gpu_memory
-        self._val_log_step: int = 0
-        self._test_log_step: int = 0
```
carmocca, line 41: Nice

pytorch_lightning/loops/fit_loop.py
rohitgr7: for the progress bar. The main progress bar requires the total batches for both which is created during training.

pytorch_lightning/loops/fit_loop.py
rohitgr7: follow up PR! Will be quick if this PR get's merged quickly today.

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
rohitgr7: what if the number of dataloaders changes? if they are new dataloaders, I'd suggest reinitializing them

tests/trainer/logging_/test_eval_loop_logging.py
rohitgr7: can be done that way too. will update

pytorch_lightning/loops/epoch/evaluation_epoch_loop.py
rohitgr7: I thought since we already have a tracker, why not use that? batch_progress resets after each dataloader call.

carmocca (approved):
I included the deprecation here. It can be easily removed when cherry-picked into a bugfix release.

CHANGELOG.md
```diff
@@ -235,6 +235,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed issue where the CLI could not pass a `Profiler` to the `Trainer` ([#13084](https://github.com/PyTorchLightning/pytorch-lightning/pull/13084))
 
 
+- Fixed logging on step level for eval mode ([#12184](https://github.com/PyTorchLightning/pytorch-lightning/pull/12184))
```
awaelchli, line 238: This is too general. You need to mention that this is fixed for the case of multiple dataloaders. As I understand, there is no known bug for single dataloader use cases.

awaelchli (approved):
The fact that lightning offers step level logging for validation/testing still baffles me today. Absolute insanity.
Anyway, thanks @rohitgr7 for working on this!

CHANGELOG.md
carmocca: What about "Fixed logging's step values when multiple dataloaders are used during evaluation"?

I'll resolve this comment and change it in a separate PR as it's ready to go otherwise. Save the trees :evergreen_tree:

PR #12194, opened 2022-03-02
Fix deepspeed keeping old sub-folders in same ckpt path

### LINKED ISSUE #11687: DeepSpeed model checkpoint saves all the checkpoints even when ModelCheckpoint.save_top_k=1
## 🐛 Bug

When using DeepSpeed with ModelCheckpoint, even when I set save_top_k to 1, the checkpoint directory named `last.ckpt` contains all the checkpoints, instead of just the top 1. This causes the checkpoints to take way more disk space.

This behavior is due to the [DeepSpeed strategy](https://github.com/PyTorchLightning/pytorch-lightning/blob/86b177ebe5427725b35fde1a8808a7b59b8a277a/pytorch_lightning/strategies/deepspeed.py#L749) saving each checkpoint in `last.ckpt` with a distinct the tag containing `global_step` (default for DeepSpeed). These are never deleted.
For example, if I save a checkpoint every 1000 steps and train for 400k steps, the directory `last.ckpt` will contain: `global_step_999`, `global_step_1999`, ..., `global_step_399999`, which takes way too much disk space.

### Expected behavior

When saving checkpoints with DeepSpeed, `last.ckpt` should only contain the last global step instead of all the global steps if save_last_k is set to 1.

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.5.9
- PyTorch Version (e.g., 1.10): 1.10.1
- Python version (e.g., 3.9): 3.8
- OS (e.g., Linux): Ubuntu 20.04
- Any other relevant information: deepspeed 0.5.10

cc @awaelchli @ananthsub @ninginthecloud @rohitgr7 @SeanNaren @carmocca @jjenniferdai

### [Edit] To reproduce:
```
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

def run():

    model = BoringModel()
    trainer = Trainer(
        strategy='deepspeed_stage_1',
        accelerator='gpu',
        devices=1,
        callbacks=[ModelCheckpoint(save_top_k=1, save_last=True)],
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=10,
        enable_model_summary=False,
    )


if __name__ == "__main__":
    run()


# last.ckpt contains ALL the checkpoints even though we limit the number of checkpoints to 1.
# This can take up way too much disk space.

# ls lightning_logs/version_0/checkpoints/last.ckpt
# global_step1  global_step2  global_step3  global_step4  global_step5  global_step6  global_step7  global_step8  global_step9  global_step10  latest  zero_to_fp32.py
```

rohitgr7:
## What does this PR do?

Fixes #11687

## Before submitting

## PR review

## Did you have fun?

cc @borda @SeanNaren @awaelchli @rohitgr7


pytorch_lightning/strategies/deepspeed.py
```diff
@@ -736,6 +736,11 @@ def save_checkpoint(self, checkpoint: Dict, filepath: _PATH) -> None:
             checkpoint: The checkpoint state dictionary
             filepath: write-target file's path
         """
+        # since deepspeed creates subfolders instead of checkpoint files at the filepath
+        # it's better to remove the filepath completely before saving a new one to avoid
+        # creating multiple sub-folders (more info: #11687)
+        self.remove_checkpoint(filepath)
```
rohitgr7, line 742: another patch here can be at L-756
```
self.deepspeed_engine.save_checkpoint(filepath, client_state=checkpoint, tag='some_static_string')
```
so that deepspeed doesn't use dynamic tags to generate multiple sub-folders rather replace it.

carmocca (approved):

tests/strategies/test_deepspeed_strategy.py
```diff
@@ -1077,3 +1077,23 @@ def test_deepspeed_with_meta_device(tmpdir):
     )
     trainer.fit(model)
     assert model.layer.weight.device.type == "cpu"
+
+
+@RunIf(min_gpus=2, deepspeed=True, standalone=True)
+def test_deepspeed_multi_save_same_filepath(tmpdir):
```
awaelchli, line 1083: can we add a description to the test? especially since this test is to prevent regression

awaelchli (approved):

Borda (approved):

pytorch_lightning/strategies/deepspeed.py
rohitgr7: update tag='checkpoint' instead since `tag=''` saves the sharded weights in the folder itself where 2 more files are present.
so now it will be `["latest", "zero_to_fp32.py", "checkpoint"]`.

tests/strategies/test_deepspeed_strategy.py
rohitgr7: done!

tests/strategies/test_deepspeed_strategy.py
```diff
@@ -1085,3 +1085,25 @@ def test_deepspeed_with_meta_device(tmpdir):
     )
     trainer.fit(model)
     assert model.layer.weight.device.type == "cpu"
+
+
+@RunIf(min_gpus=2, deepspeed=True)
```
rohitgr7, line 1090:
```suggestion
@RunIf(min_gpus=2, deepspeed=True, standalone=True)
```

PR #12223, opened 2022-03-04
Update type hint for log_model

roschly:
According to the documentation, `log_model` should either be boolean or the string "all".
Updated type hint to reflect that: `Union[str, bool]`


akihironitta (approved):

carmocca (approved):

krshrimali (approved):
Thanks! 🚀

PR #12267, opened 2022-03-08
Reduce number of times optimizers are instantiated with FSDP

ananthsub:
## What does this PR do?

Follow up to comment here: https://github.com/PyTorchLightning/pytorch-lightning/pull/11952#discussion_r820060750

Currently we are calling `LightningModule.configure_optimizers` 3 times in the FSDP strategy 😞
We only need to do this one, after the model is wrapped with FSDP inside of `model_sharded_context` 

## Before submitting
No

## PR review

## Did you have fun?


pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -135,16 +135,11 @@ def setup_distributed(self) -> None:
...
         self.accelerator.setup(trainer)
-        self.setup_optimizers(trainer)
-        self.setup_precision_plugin()
-        optimizers_to_device(self.optimizers, self.root_device)
 
         if self._layer_sync:
             self.model = self._layer_sync.apply(self.model)
 
         self.configure_ddp()
-        self.barrier()
```
awaelchli, line 146: Are you sure the barrier can be dropped? It is there for a reason, we should probably keep it.

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -171,19 +166,19 @@ def wrap_policy(*args, **kwargs):
             yield
 
         log.detail(f"{self.__class__.__name__}: exiting model_sharded_context.")
+        self.setup_optimizers(self.lightning_module.trainer)
+        optimizers_to_device(self.optimizers, self.root_device)
+        self.setup_precision_plugin()
```
awaelchli, line 171: why can it not be done in setup? setup() is mean to do exactly that, and setup gets called after the model is wrapped (see trainer.py). I highly recommend to keep it in setup()

ananthsub: Yes, I can change that. I needed to double check the ordering around wrapping FSDP vs move to device. 

But looking again at the trainer code, why does the Trainer need to explicitly call configure_sharded_model. Could we move that to the Strategy's setup instead?

pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -135,16 +135,15 @@ def setup_distributed(self) -> None:
...
-        optimizers_to_device(self.optimizers, self.root_device)
 
         if self._layer_sync:
             self.model = self._layer_sync.apply(self.model)
 
         self.configure_ddp()
         self.barrier()
         self.setup_optimizers(trainer)
+        optimizers_to_device(self.optimizers, self.root_device)
+        self.setup_precision_plugin()
```
four4fish, line 146: Does the order for these two matter? I see inconsistency across different strategies

four4fish (approved):

pytorch_lightning/strategies/fully_sharded.py
ananthsub: i prefer updating the optimizers first. setup_precision_plugin simply passes the model, optimizers, and lr_schedulers to the precision plugin to view

awaelchli (approved):

rohitgr7 (approved):

Borda (approved):

PR #12292, opened 2022-03-10
Set `MLFlowLogger` status to FAILED when training raises an error

### LINKED ISSUE #12291: `MLFlowLogger`'s status is "RUNNING" even after training failed
## 🐛 Bug

If a trainer with `MLFlowLogger` raises an error, the user should be able to see the MLflow's screen to check the training has been failed.
MLflow's status remains "RUNNING" even after `trainer.fit` raises an error in the current implementation, so the user cannot know whether the training is still in progress or failed.

#### Current behavior when training finished with an error: 

スクリーンショット 2022-03-10 21 41 18

#### Expected behavior: 

スクリーンショット 2022-03-10 21 31 23

### To Reproduce

```py
class CustomModel(BoringModel):
        super().training_step(batch, batch_idx)
        raise BaseException
trainer = Trainer(logger=MLFlowLogger("test"))
try:
    trainer.fit(CustomModel())
finally:
    print(trainer.logger.experiment.get_run(trainer.logger.run_id).info.status) # This should be 'FAILED'
```

cc @borda

ritsuki1227:
## What does this PR do?

If a trainer with MLFlowLogger raises an error, the user should be able to see the MLflow's screen to check the training has been failed.
MLflow's status remains "RUNNING" even after trainer.fit raises an error in the current implementation, so the user cannot know whether the training is still in progress or failed.

Fixes #12291

## Before submitting

## PR review

## Did you have fun?


ritsuki1227:
The commit has succeeded to fix the bug, but failed to pass an existing unit test, which I couldn't find a solution to.
Can anyone help me with the draft PR?

ritsuki1227:
I would appreciate it if you approve running the workflows 🙂

pytorch_lightning/loggers/mlflow.py
```diff
@@ -230,7 +230,10 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
     @rank_zero_only
     def finalize(self, status: str = "FINISHED") -> None:
         super().finalize(status)
-        status = "FINISHED" if status == "success" else status
+        if status == "success":
+            status = "FINISHED"
+        elif status == "failed":
```
awaelchli, line 235: an else is missing, which can lead to an undefined value for status. 
let's choose a default value, or not call set_terminated if status is neither finished nor failed.

pytorch_lightning/loggers/tensorboard.py
```diff
@@ -265,9 +265,10 @@ def save(self) -> None:
 
     @rank_zero_only
     def finalize(self, status: str) -> None:
-        self.experiment.flush()
-        self.experiment.close()
-        self.save()
+        if status != "failed":
```
awaelchli, line 268: I think we can still flush and close even if status is failed. Or is this not correct?

pytorch_lightning/trainer/trainer.py
```diff
@@ -730,12 +730,16 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
                 self.state.status = TrainerStatus.INTERRUPTED
                 self._call_callback_hooks("on_keyboard_interrupt")
                 self._call_callback_hooks("on_exception", exception)
+                for logger in self.loggers:
+                    logger.finalize("failed")
```
awaelchli, line 734: Great. I think we should follow up on this PR with a small refactor that defines an enum for statuses.

tests/loggers/test_mlflow.py
```diff
@@ -238,3 +238,23 @@ def test_mlflow_logger_experiment_calls(client, mlflow, time, tmpdir):
     logger._mlflow_client.create_experiment.assert_called_once_with(
         name="test", artifact_location="my_artifact_location"
     )
+
+
+@mock.patch("pytorch_lightning.loggers.mlflow.mlflow")
+@mock.patch("pytorch_lightning.loggers.mlflow.MlflowClient")
+def test_mlflow_logger_run_status_failed(client, mlflow):
```
awaelchli, line 245: great test!

awaelchli:
> I would appreciate it if you approve running the workflows 🙂

Most of team was OOO. Sorry about the delay. 

> The commit has succeeded to fix the bug, but failed to pass an existing unit test, which I couldn't find a solution to.
> Can anyone help me with the draft PR?

Which test was it?

ritsuki1227:
@awaelchli 
Thank you for your review and modification!

> Which test was it?

It was `tests/utilities/test_cli.py::test_cli_distributed_save_config_callback`.
Although the latest implementation has passed the test, the failed test can be reproduced on commit dbbfef817a3fff52a4d0b98ef68ae39f03863f56.

pytorch_lightning/loggers/tensorboard.py
ritsuki1227: `tests/utilities/test_cli.py::test_cli_distributed_save_config_callback` fails without the condition on commit dbbfef817a3fff52a4d0b98ef68ae39f03863f56.
I don't know why but somehow the problem was fixed on the current implementation.

ritsuki1227:
@awaelchli @daniellepintz 
Do you have any update on this?
I understand some PRs may be pending because of the 1.6 release.

pytorch_lightning/loggers/tensorboard.py
carmocca: Is this test still a problem? I am its author

ritsuki1227:
@carmocca Is there anything I can help to merge the PR?

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -267,9 +267,10 @@ def save(self) -> None:
 
     @rank_zero_only
     def finalize(self, status: str) -> None:
-        self.experiment.flush()
-        self.experiment.close()
-        self.save()
+        if status != "failed":
```
carmocca, line 270: Why don't we want to close it if it failed?

tests/tests_pytorch/loggers/test_mlflow.py
```diff
@@ -259,3 +259,23 @@ def test_mlflow_logger_experiment_calls(client, mlflow, time, tmpdir):
     logger._mlflow_client.create_experiment.assert_called_once_with(
         name="test", artifact_location="my_artifact_location"
     )
+
+
+@mock.patch("pytorch_lightning.loggers.mlflow.mlflow")
+@mock.patch("pytorch_lightning.loggers.mlflow.MlflowClient")
+def test_mlflow_logger_run_status_failed(client, mlflow):
+    class CustomModel(BoringModel):
+        def training_step(self, batch, batch_idx):
```
carmocca, line 268: You can raise the exception from an earlier hook like `on_fit_start`

carmocca (approved):
The changes look good to me. I'll merge master to see what CI thinks

src/pytorch_lightning/loggers/tensorboard.py
ritsuki1227: @carmocca Hi, sorry for the late reply.
I've added the modification, but it makes `test_cli.py::test_cli_distributed_save_config_callback` fail.
I'd appreciate it if you help me with that.
FYI, I've found 10e133af6407463fe43951e9aa3153ff9b95c8c7 solves the problem as a workaround.

tests/tests_pytorch/loggers/test_mlflow.py
ritsuki1227: fixed the test to raise the exception on `on_fit_start`.

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -101,12 +101,12 @@ def __init__(
         super().__init__(agg_key_funcs=agg_key_funcs, agg_default_func=agg_default_func)
         self._save_dir = save_dir
         self._name = name or ""
-        self._version = version
         self._sub_dir = sub_dir
         self._log_graph = log_graph
         self._default_hp_metric = default_hp_metric
         self._prefix = prefix
         self._fs = get_filesystem(save_dir)
+        self._version = version if version is not None else self._get_next_version()
```
awaelchli, line 109: We shouldn't have any changes in tensorboard logger. Could you remove them?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -654,12 +654,16 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
             if not self.interrupted:
                 self.state.status = TrainerStatus.INTERRUPTED
                 self._call_callback_hooks("on_exception", exception)
+                for logger in self.loggers:
+                    logger.finalize("failed")
```
awaelchli, line 658: If this gets moved here, then it shoud be removed from `Trainer._call_teardown_hook()`, right?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -654,12 +654,16 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
...
         except BaseException as exception:
             self.state.status = TrainerStatus.INTERRUPTED
             if distributed_available() and self.world_size > 1:
                 # try syncing remaining processes, kill otherwise
                 self.strategy.reconciliate_processes(traceback.format_exc())
             self._call_callback_hooks("on_exception", exception)
+            for logger in self.loggers:
+                logger.finalize("failed")
```
awaelchli, line 666: Note that with strategies like ddp_spawn, this won't do anything. On the main process, the logger has never done anything, so sending a "failed" signal will be wrong. 

There is currently no way we can handle this situation, and this is mostly the reason why only the "success" calls exist today. Sorry for the bad news :(

src/pytorch_lightning/loggers/tensorboard.py
carmocca: See https://github.com/Lightning-AI/lightning/pull/12292#discussion_r959484188

src/pytorch_lightning/trainer/trainer.py
ritsuki1227: Thank you for sharing. Is there anything I can do to help with the situation? I don't know If I have got the skills to do that.

awaelchli: A quick temporary fix is to do this in mlflowlogger.finalize:

```py
if self._mlflow_client is None:
    return

# else do your thing here to set the status to failed
```

This logic relies on the fact that _mlflow_client is only ever not None in the worker process. In the main process it is None unless the user called the logger manually. 

This "hack" is not ideal, because it shifts responsibility to the logger that should ideally be handled by probably the launcher (i.e., only calling the finalize on the right process). 
This mechanism to process exceptions before they bubble up to the main process would be necessary to have for handling this properly. It needs some brainstorming, so probably the next step is to open a GH issue. 

For now, let's try to go with my suggestion and try to land the PR, what do you think? 

Note the above snippet would have to be adapted for each logger that implements finalize, i.e., checking whether `self._experiment is None`.

ritsuki1227: @awaelchli 
> checking whether self._experiment is None.

In addition to the fact that objects to be checked were different for each logger, it is unclear how other loggers behave when `finalize("failed")` is called.
So I fixed it so that `finalize("failed")` is called on an exception only when the logger is mlfow. Can you review the new commits?
A quick overview of the changes is as follows:

trainer.py:
```py
def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs: Any) -> Any:
        ...
        except BaseException as exception:
            ...
            for logger in self.loggers:
                logger.finalize_when_exception()
```

logger.py:
```py
class Logger(ABC):
    ...
    def finalize_when_exception(self) -> None:
        pass
```

mlflow.py: 
```py
class MLFlowLogger(Logger):
    ...
    def finalize_when_exception(self) -> None:
        super().finalize_when_exception()
        if self._mlflow_client is None:
            return
        self.finalize("failed")
```

awaelchli: > In addition to the fact that objects to be checked were different for each logger, it is unclear how other loggers behave when finalize("failed") is called.

If they don't do anything now, then we leave it like that. 

I wouldn't introduce any new api here, especially not in the base class when it's just affecting one logger. Did my [suggestion](https://github.com/Lightning-AI/lightning/pull/12292#discussion_r961550279) not work?
Happy to help and get hands-on here. Should I push a commit? I can take care of the other loggers.

ritsuki1227: @awaelchli It would be helpful if you push a commit or give me some advice!

> Did my https://github.com/Lightning-AI/lightning/pull/12292#discussion_r961550279 not work?

If you fix `tensorboard.py` like this: 

```py
    @rank_zero_only
    def finalize(self, status: str) -> None:
        if self._experiment is None:
            return
        self.experiment.flush()
        self.experiment.close()
        self.save()
```

Then `tests/tests_pytorch/loggers/test_tensorboard.py::test_tensorboard_save_hparams_to_yaml_once` fails: 

```py
    def test_tensorboard_save_hparams_to_yaml_once(tmpdir):
        model = BoringModel()
        logger = TensorBoardLogger(save_dir=tmpdir, default_hp_metric=False)
        trainer = Trainer(max_steps=1, default_root_dir=tmpdir, logger=logger)
        assert trainer.log_dir == trainer.logger.log_dir
        trainer.fit(model)
    
        hparams_file = "hparams.yaml"
>       assert os.path.isfile(os.path.join(trainer.log_dir, hparams_file))
E       AssertionError: assert False
E        +  where False = <function isfile at 0x1059b8670>('/private/var/folders/01/y61dt9vx52lds6ylr3975mbc0000gq/T/pytest-of-ritsuki.yamada/pytest-93/test_tensorboard_save_hparams_0/lightning_logs/version_0/hparams.yaml')
E        +    where <function isfile at 0x1059b8670> = <module 'posixpath' from '/Users/ritsuki.yamada/.pyenv/versions/3.9.13/lib/python3.9/posixpath.py'>.isfile
E        +      where <module 'posixpath' from '/Users/ritsuki.yamada/.pyenv/versions/3.9.13/lib/python3.9/posixpath.py'> = os.path
E        +    and   '/private/var/folders/01/y61dt9vx52lds6ylr3975mbc0000gq/T/pytest-of-ritsuki.yamada/pytest-93/test_tensorboard_save_hparams_0/lightning_logs/version_0/hparams.yaml' = <function join at 0x1059b8e50>('/private/var/folders/01/y61dt9vx52lds6ylr3975mbc0000gq/T/pytest-of-ritsuki.yamada/pytest-93/test_tensorboard_save_hparams_0/lightning_logs/version_0', 'hparams.yaml')
E        +      where <function join at 0x1059b8e50> = <module 'posixpath' from '/Users/ritsuki.yamada/.pyenv/versions/3.9.13/lib/python3.9/posixpath.py'>.join
E        +        where <module 'posixpath' from '/Users/ritsuki.yamada/.pyenv/versions/3.9.13/lib/python3.9/posixpath.py'> = os.path
E        +      and   '/private/var/folders/01/y61dt9vx52lds6ylr3975mbc0000gq/T/pytest-of-ritsuki.yamada/pytest-93/test_tensorboard_save_hparams_0/lightning_logs/version_0' = <pytorch_lightning.trainer.trainer.Trainer object at 0x130584dc0>.log_dir

tests/tests_pytorch/loggers/test_tensorboard.py:291: AssertionError
```

awaelchli: @ritsuki1227 I noticed this too. This was because the save() in finalize of TensorBoard was expected to be called, even when no self.experiment is initialized. This is because the saving of hparams happens independently of the tensorboard summary writer. Now, I implemented it so that it behaves the same as on master. 

Otherwise, I took care of:
- Other loggers
- Improving the tests

I let the CI run now and check back later. If other tests fail, I will take another look.

awaelchli:
@ritsuki1227 I created a fix in #14762 to unblock this PR for the remaining issues.

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -644,12 +644,16 @@ def _call_and_handle_interrupt(self, trainer_fn: Callable, *args: Any, **kwargs:
             if not self.interrupted:
                 self.state.status = TrainerStatus.INTERRUPTED
                 self._call_callback_hooks("on_exception", exception)
+                for logger in self.loggers:
+                    logger.finalize("failed")
```
awaelchli, line 648: Right now, KeyboardInterrupt is the same as "failed". Perhaps we want to introduce something like "stopped" in the future. How the loggers react here depends on how the third-party implementations expect things to be reported when an experiment gets "stopped" or "interrupted" in a controlled way.

awaelchli (approved):

src/pytorch_lightning/loggers/wandb.py
```diff
@@ -552,6 +552,10 @@ def use_artifact(self, artifact: str, artifact_type: Optional[str] = None) -> "w
 
     @rank_zero_only
     def finalize(self, status: str) -> None:
+        if self._experiment is None:
```
rohitgr7, line 555: experiment is initialized during init, so will it be None on the main process?

rohitgr7 (approved):
mind add a changelog?

src/pytorch_lightning/loggers/wandb.py
awaelchli: You're right! On wandb >= 0.12.10 that's the case. 
I changed the logic now to keep the behavior on master, which is: only upload checkpoints on success

awaelchli:
@ritsuki1227 before we merge, would you mind checking again your MLFlow use case, that the status is reported as you expect? Thank you!

ritsuki1227:
@awaelchli Thank you for your commits! I've checked mlflow logger worked correctly on the use case.

awaelchli:
Awesome, let's get it in then! Thanks a lot for your help in making the integration with PL better!

PR #11526, opened 2022-01-18
fix available modules

Borda:
## What does this PR do?

Fixes #11524
Original fix was taken from https://github.com/PyTorchLightning/metrics/pull/772

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda

PR #12377, opened 2022-03-19
Fix _module_available to detect horovod.torch properly

### LINKED ISSUE #12370: _module_available does not correctly check child modules
## 🐛 Bug

I tried to use horovod accelerator which is correctly installed in my environment, as I can `import horovod.torch` without any problem but I get an error stating that horovod is not installed. I checked the source code to see how module availability was checked and was surprised to see that lightning uses `hasattr` to check child modules. I guess it comes handy when you are not trying to check for a module but for a function or a class, but I don't understand why the function does not simply try to call `import horovod.torch` (which works) instead of `hasattr(horovod, "torch")` (which returns `False`). Indeed, a submodule is not necessarily an attribute of its parent module unless stated explicitly, which is apparently not the case for horovod. If you still want to handle the case where you are trying to import an object and not a module, you can check for a `ModuleNotFoundError` and then use `hasattr` instead.

### To Reproduce

With horovod=0.24.1 installed with PyTorch compatibility, use `_module_available("horovod.torch")` (or simply `import horovod; hasattr(horovod, "torch")`). 

### Expected behavior

The function should return `True`. It has to be noted that if a `import horovod.torch` has been performed prior to the function call, it returns `True`, which can serve as a workaround in the meantime. 

### Environment

PyTorch version: 1.10.1
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: CentOS Stream release 8 (x86_64)
GCC version: (GCC) 8.5.0 20210514 (Red Hat 8.5.0-10)
Clang version: Could not collect
CMake version: version 3.20.2
Libc version: glibc-2.28

Python version: 3.9.7 (default, Sep 16 2021, 13:09:58)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-4.18.0-365.el8.x86_64-x86_64-with-glibc2.28
Is CUDA available: True
CUDA runtime version: 11.6.55
GPU models and configuration:
GPU 0: Tesla P100-PCIE-16GB
GPU 1: Tesla P100-PCIE-16GB

Nvidia driver version: 510.47.03
cuDNN version: Probably one of the following:
/usr/lib64/libcudnn.so.8.3.2
/usr/lib64/libcudnn_adv_infer.so.8.3.2
/usr/lib64/libcudnn_adv_train.so.8.3.2
/usr/lib64/libcudnn_cnn_infer.so.8.3.2
/usr/lib64/libcudnn_cnn_train.so.8.3.2
/usr/lib64/libcudnn_ops_infer.so.8.3.2
/usr/lib64/libcudnn_ops_train.so.8.3.2
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.22.1
[pip3] pytorch-lightning==1.5.10
[pip3] torch==1.10.1
[pip3] torchaudio==0.10.1
[pip3] torchmetrics==0.7.2
[pip3] torchvision==0.11.2
[pip3] horovod==0.24.1
[conda] blas                      1.0                         mkl
[conda] cudatoolkit               11.3.1               h2bc3f7f_2
[conda] mkl                       2021.4.0           h06a4308_640
[conda] mkl-service               2.4.0            py39h7f8727e_0
[conda] mkl_fft                   1.3.1            py39hd3c417c_0
[conda] mkl_random                1.2.2            py39h51133e4_0
[conda] mypy-extensions           0.4.3                    pypi_0    pypi
[conda] numpy                     1.22.1                   pypi_0    pypi
[conda] numpy-base                1.21.2           py39h79a1101_0
[conda] pytorch                   1.10.1          py3.9_cuda11.3_cudnn8.2.0_0    pytorch
[conda] pytorch-lightning         1.5.10                   pypi_0    pypi
[conda] pytorch-mutex             1.0                        cuda    pytorch
[conda] torchaudio                0.10.1               py39_cu113    pytorch
[conda] torchmetrics              0.7.2                    pypi_0    pypi
[conda] torchvision               0.11.2               py39_cu113    pytorch

### Additional context

If you confirm that this is unexpected behavior, I can write a PR that should fix this by directly trying to import the module.

cc @awaelchli

JoostvDoorn:
## What does this PR do?

Fixes #12370, such that horovod.torch is again properly detected.

This PR simplifies the logic of the _module_available function. I think in general it's not a good idea to try to deviate too much from importlib when testing for module availability, if the specific logic of `hasattr` is necessary for any check it would be better to handle that case separately. I tried reproducing the error that first triggered this change in the torchmetrics repo https://github.com/PyTorchLightning/metrics/issues/770 by testing `_module_available("transformers.models.auto")` and this didn't give me any problems, but a maintainer might be a better judge of this. I can replicate this code to that repo later.

Note I run horovod in develop mode not sure if that's related, but latest pypi version of horovod has a very restrictive pin on the pytorch lightning version.

## Before submitting

## PR review

## Did you have fun?


awaelchli:
Thanks for the PR @JoostvDoorn 
I triggered CI, let's see what happens :)

JoostvDoorn:
Some of the horovod tests are failing and seem to be skipped in master.

Borda (commented):
does it also work for any other modules?
we had the for loop there on purpose...

JoostvDoorn:
> does it also work for any other modules? we had the for loop there on purpose...

I tested some manually, but I think this is a good point, so I added some tests in 0d798211455b59be55b15e60d953b07bbeaaee0f to confirm imports match with the flags allowing the CI to test for this. Let me know if there are any specific cases that need to be added to this, as the exact flags for which the loop is relevant in this repository isn't fully clear to me.

pytorch_lightning/utilities/imports.py
```diff
@@ -53,13 +53,9 @@ def _module_available(module_path: str) -> bool:
     if not _package_available(module_names[0]):
         return False
     try:
-        module = importlib.import_module(module_names[0])
-    except ImportError:
+        importlib.import_module(module_path)
+    except ModuleNotFoundError:
         return False
-    for name in module_names[1:]:
```
awaelchli, line 59: looking at git blame, this came from #11526 which on the other hand was carried over from a report in torchmetrics: https://github.com/PyTorchLightning/metrics/issues/770

So, to verify that @JoostvDoorn's fix is valid, we can check against that original issue to make sure its' covered :)

awaelchli:
@JoostvDoorn It looks like #12379 will have to be completed first, right? If you want, you can merge #12379 into this branch to see if both work together.

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -623,7 +623,7 @@ def _handle_horovod(self) -> None:
         hvd.init()
         if isinstance(self.accelerator, GPUAccelerator):
             # Horovod assigns one local GPU per process
-            self._parallel_devices = list(range(hvd.local_size()))
+            self._parallel_devices = [torch.device(f"cuda:{i}") for i in range(hvd.local_size())]
```
four4fish, line 626: Does hvd expect a list of id or a list of cuda devices? Both works right?
+1 to this change if both work, cuda devices list is more consistent with other strategies' parallel devices

JoostvDoorn: This is required to be device since the value will propegate to root_device used here: https://github.com/PyTorchLightning/pytorch-lightning/blob/3079416ae5aaecad80511d901454096a00ce2c88/pytorch_lightning/strategies/horovod.py#L140

pytorch_lightning/utilities/imports.py
carmocca: cc @Borda as the author of this piece of code

carmocca (approved):

awaelchli (approved):
Great!

tests/utilities/test_imports.py
```diff
@@ -45,3 +54,48 @@ def test_compare_version(monkeypatch):
     assert not _compare_version("torch", operator.ge, "1.10.0.rc0")
     assert _compare_version("torch", operator.ge, "1.10.0", use_base_version=True)
     assert not _compare_version("torch", operator.ge, "1.10.0")
+
+
+def test_imports():
+    try:
+        import apex  # noqa
+
+        assert _APEX_AVAILABLE
```
Borda, line 63: lest move it to finalize

tests/utilities/test_imports.py
```diff
@@ -45,3 +54,48 @@ def test_compare_version(monkeypatch):
...
+    except ModuleNotFoundError:
+        assert not _APEX_AVAILABLE
```
Borda, line 65:
```suggestion
    try:
        import apex  # noqa
    except ModuleNotFoundError:
        assert not _APEX_AVAILABLE
    else:
        assert _APEX_AVAILABLE

```
lest move it to `else` https://note.nkmk.me/en/python-try-except-else-finally/

Borda (approved):

pytorch_lightning/utilities/imports.py
rohitgr7: cc @Borda

Borda: yep, did test adjustments and seems to be passing

Borda:
just found another type of error:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
File /usr/lib/python3.9/site-packages/horovod/torch/mpi_ops.py:33, in <module>
     32 try:
---> 33     from horovod.torch import mpi_lib_v2 as mpi_lib
     34 except Exception as e:

ImportError: /usr/lib/python3.9/site-packages/horovod/torch/mpi_lib_v2.cpython-39-x86_64-linux-gnu.so: undefined symbol: _ZN2at4_ops15to_dtype_layout4callERKNS_6TensorEN3c108optionalINS5_10ScalarTypeEEENS6_INS5_6LayoutEEENS6_INS5_6DeviceEEENS6_IbEEbbNS6_INS5_12MemoryFormatEEE

The above exception was the direct cause of the following exception:

HorovodVersionMismatchError               Traceback (most recent call last)
Input In [2], in <cell line: 8>()
      6 import matplotlib
      7 import matplotlib.pyplot as plt
----> 8 import pytorch_lightning as pl
      9 import seaborn as sns
     10 import torch

File ~/.local/lib/python3.9/site-packages/pytorch_lightning/__init__.py:30, in <module>
     27     _logger.addHandler(logging.StreamHandler())
     28     _logger.propagate = False
---> 30 from pytorch_lightning.callbacks import Callback  # noqa: E402
     31 from pytorch_lightning.core import LightningDataModule, LightningModule  # noqa: E402
     32 from pytorch_lightning.trainer import Trainer  # noqa: E402
```

PR #12245, opened 2022-03-06
Update logging docs

daniellepintz:
## What does this PR do?
Fixes various things in the logging docs.

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #12378, opened 2022-03-19
Replace PostLocalSGDOptimizer with a dedicated model averaging component

daniellepintz:
## What does this PR do?

_In this PR I am finishing https://github.com/PyTorchLightning/pytorch-lightning/pull/9446 which was created by @wayi1 before the Strategy refactor. The description below is copied from that PR._

This is an improvement of https://github.com/PyTorchLightning/pytorch-lightning/pull/8967.

Replace `PostLocalSGDOptimizer` by a dedicated model averaging component that can run after `optimizer_step`.

The previous implementation can cause data races and fail the training for some cases. For example, if the data loading phase also involves allreduce (e.g., for checking if the input data stream has reached the end), and such allreduce can be kicked off during `optimizer.step()`. Therefore, the implementation in this PR can fix such issue.

Proposal: https://github.com/pytorch/pytorch/issues/59699

There are two options to enable post-local SGD by using vanilla PyTorch:

1. Post-LocalSGD Comm Hook + Periodic Model Averager
https://github.com/pytorch/pytorch/blob/master/torch/distributed/algorithms/model_averaging/averagers.py#L48-L78

    This option requires 1-line addition (such as `averager.average_parameters(model.parameters())`) in the training loop.

2. Post-LocalSGD Comm Hook + Post-LocalSGD Optimizer
https://github.com/pytorch/pytorch/blob/master/torch/distributed/optim/post_localSGD_optimizer.py#L15-L48

    This option does not need any code change in the training loop. I tried this option earlier. However, one limitation is that if there is another phase such as data loading that also runs some communication like allreduce and overlaps with `optimizer_step()`, then this can cause potential data race and fail the training. Since such concurrent phase like data loading is out of the control of model averaging, I didn't choose this implementation.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/strategies/ddp.py
+++ b/pytorch_lightning/strategies/ddp.py
@@ -98,9 +98,8 @@ class DDPStrategy(ParallelStrategy):
         self._ddp_comm_state = ddp_comm_state
         self._ddp_comm_hook = ddp_comm_hook
         self._ddp_comm_wrapper = ddp_comm_wrapper
-        if _TORCH_GREATER_EQUAL_1_10:
-            self._model_averaging_period = model_averaging_period
-            self._model_averager: Optional[ModelAverager] = None
+        self._model_averaging_period = model_averaging_period
+        self._model_averager: Optional[ModelAverager] = None
         self._pids: Optional[List[int]] = None
         self._sync_dir: Optional[str] = None
         self._rank_0_will_call_children_scripts: bool = False
@@ -238,7 +237,8 @@ class DDPStrategy(ParallelStrategy):
                 if isinstance(self._ddp_comm_state, post_localSGD.PostLocalSGDState):
                     self._enable_model_averaging()
 
-    def _enable_model_averaging(self):
+    def _enable_model_averaging(self) -> None:
+        # Only called when PyTorch version >= 1.10
         log.detail(f"{self.__class__.__name__}: reinitializing optimizers with post localSGD")
         if self._model_averaging_period is None:
             raise ValueError(
@@ -282,15 +282,19 @@ class DDPStrategy(ParallelStrategy):
             model: reference to the model, optionally defining optimizer step related hooks
             **kwargs: Any extra arguments to ``optimizer.step``
         """
-        super().optimizer_step(optimizer, opt_idx, closure, model, **kwargs)
+        optimizer_output = super().optimizer_step(optimizer, opt_idx, closure, model, **kwargs)
 
-        if _TORCH_GREATER_EQUAL_1_10 and self._model_averager is not None:
-            for opt in self.optimizers():
-                for group in opt.param_groups:
-                    for param in group["params"]:
-                        if param.grad is None:
-                            continue
-                        self._model_averager.average_parameters(iter(param))
+        if not _TORCH_GREATER_EQUAL_1_10 or self._model_averager is None:
+            return optimizer_output
+
+        for opt in self.optimizers():
+            for group in opt.param_groups:
+                for param in group["params"]:
+                    if param.grad is None:
+                        continue
+                    self._model_averager.average_parameters(iter(param))
+
+        return optimizer_output
 
     def configure_ddp(self) -> None:
         log.detail(f"{self.__class__.__name__}: configuring DistributedDataParallel")
--- a/pytorch_lightning/strategies/ddp.py
+++ b/pytorch_lightning/strategies/ddp.py
@@ -18,12 +18,13 @@ import signal
 import tempfile
 import time
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
 import torch.distributed
 from torch.nn import Module
 from torch.nn.parallel.distributed import DistributedDataParallel
+from torch.optim.optimizer import Optimizer
 
 import pytorch_lightning as pl
 from pytorch_lightning.core.optimizer import LightningOptimizer
@@ -58,6 +59,8 @@ if _FAIRSCALE_AVAILABLE:
     from fairscale.optim import OSS
 if _TORCH_GREATER_EQUAL_1_8:
     from pytorch_lightning.utilities.distributed import register_ddp_comm_hook
+if _TORCH_GREATER_EQUAL_1_10:
+    from torch.distributed.algorithms.model_averaging.averagers import ModelAverager
 
 
 log = logging.getLogger(__name__)
@@ -95,7 +98,9 @@ class DDPStrategy(ParallelStrategy):
         self._ddp_comm_state = ddp_comm_state
         self._ddp_comm_hook = ddp_comm_hook
         self._ddp_comm_wrapper = ddp_comm_wrapper
-        self._model_averaging_period = model_averaging_period
+        if _TORCH_GREATER_EQUAL_1_10:
+            self._model_averaging_period = model_averaging_period
+            self._model_averager: Optional[ModelAverager] = None
         self._pids: Optional[List[int]] = None
         self._sync_dir: Optional[str] = None
         self._rank_0_will_call_children_scripts: bool = False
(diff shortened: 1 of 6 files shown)


awaelchli (approved):
I wonder if this is compatible with ddp spawn. I don't immediately see a reason why not.
Since launchers have been extracted, we want to align the two strategies as much as possible and eventually merge them into one.

pytorch_lightning/strategies/ddp.py
```diff
@@ -256,24 +255,42 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
...
+        """performs the actual optimizer step.
+
+        Args:
+            optimizer: the optimizer performing the step
+            opt_idx: index of the current optimizer
+            closure: closure calculating the loss value
+            model: reference to the model, optionally defining optimizer step related hooks
+            **kwargs: Any extra arguments to ``optimizer.step``
+        """
+        super().optimizer_step(optimizer, opt_idx, closure, model, **kwargs)
```
ananthsub, line 285: this improperly overrides the base class. you need to capture the return value of this and return it after the model averaging has been applied.

pytorch_lightning/strategies/ddp.py
```diff
@@ -256,24 +255,42 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
...
+
+        if _TORCH_GREATER_EQUAL_1_10 and self._model_averager is not None:
```
ananthsub, line 287: please return early instead to improve readability

```py
if not _TORCH_GREATER_EQUAL_1_10 or self._model_averager is None:
    return ...
...
```

pytorch_lightning/strategies/ddp.py
```diff
@@ -231,23 +236,17 @@ def _register_ddp_hooks(self) -> None:
                 import torch.distributed.algorithms.ddp_comm_hooks.post_localSGD_hook as post_localSGD
 
                 if isinstance(self._ddp_comm_state, post_localSGD.PostLocalSGDState):
-                    self._reinit_optimizers_with_post_localSGD(self._ddp_comm_state.start_localSGD_iter)
+                    self._enable_model_averaging()
 
-    def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
+    def _enable_model_averaging(self):
```
ananthsub, line 241: add a typehint for the return value
```suggestion
    def _enable_model_averaging(self) -> None:
```

pytorch_lightning/strategies/ddp.py
```diff
@@ -95,7 +98,9 @@ def __init__(
         self._ddp_comm_state = ddp_comm_state
         self._ddp_comm_hook = ddp_comm_hook
         self._ddp_comm_wrapper = ddp_comm_wrapper
-        self._model_averaging_period = model_averaging_period
+        if _TORCH_GREATER_EQUAL_1_10:
+            self._model_averaging_period = model_averaging_period
+            self._model_averager: Optional[ModelAverager] = None
```
ananthsub, line 103: this will face issues with typechecking. these attributes should be set if _TORCH_GREATER_EQUAL_1_10 is less than 1.10

ananthsub (commented):
please add a test and changelog entry as well

pytorch_lightning/strategies/ddp.py
```diff
@@ -256,24 +255,42 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
                 is_distributed_optimizer
                 or isinstance(optimizer, ZeroRedundancyOptimizer)
                 or (_FAIRSCALE_AVAILABLE and isinstance(optimizer, OSS))
+                or isinstance(optimizer, PostLocalSGDOptimizer)
             ):
                 raise ValueError(
-                    f"Cannot wrap a distributed optimizer of type {optimizer.__name__} by PostLocalSGDOptimizer."
+                    f"Currently model averaging cannot work with a distributed optimizer of type {optimizer.__name__}."
                 )
```
krshrimali, line 262: Should this be tested? (also other changes in this PR?)

krshrimali (approved):
Thanks, @daniellepintz for the PR. Looks good to me, left a couple of comments for your review. :)

pytorch_lightning/strategies/ddp.py
```diff
@@ -95,7 +98,9 @@ def __init__(
         self._ddp_comm_state = ddp_comm_state
         self._ddp_comm_hook = ddp_comm_hook
         self._ddp_comm_wrapper = ddp_comm_wrapper
-        self._model_averaging_period = model_averaging_period
+        if _TORCH_GREATER_EQUAL_1_10:
+            self._model_averaging_period = model_averaging_period
```
Borda, line 102: also, what is the attribute when PL<1.10

daniellepintz: do you mean PT? I will remove the `if _TORCH_GREATER_EQUAL_1_10:`

ananthsub (approved):

daniellepintz:
For some reason the tests for PT <1.10 are complaining `name 'post_localSGD' is not defined` even though the test is marked as `min_torch="1.10.0"`. Seems to be an issue with pytest.

pytorch_lightning/strategies/ddp.py
```diff
@@ -256,24 +255,47 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
...
-        self.optimizers = optimizers
+    def optimizer_step(
+        self,
+        optimizer: Optimizer,
+        opt_idx: int,
+        closure: Callable[[], Any],
+        model: Optional[Union["pl.LightningModule", Module]] = None,
+        **kwargs: Any,
+    ) -> Any:
+        """performs the actual optimizer step.
```
tchaton, line 277:
```suggestion
        """Performs the actual optimizer step.
```

pytorch_lightning/strategies/ddp.py
```diff
@@ -256,24 +255,47 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
...
+            closure: closure calculating the loss value
+            model: reference to the model, optionally defining optimizer step related hooks
+            **kwargs: Any extra arguments to ``optimizer.step``
+        """
+        optimizer_output = super().optimizer_step(optimizer, opt_idx, closure, model, **kwargs)
+
+        if not _TORCH_GREATER_EQUAL_1_10 or self._model_averager is None:
+            return optimizer_output
+
+        for opt in self.optimizers:
```
tchaton, line 291: Is it correct? From the documentation, it seems we need to provide averager.average_parameters(model.parameters()) entirely. Won't this trigger several average step?

tchaton (approved):
LGTM !

pytorch_lightning/strategies/ddp.py
daniellepintz: @rohan-varma

wayi1: Sorry for the confusion here.

Actually the documentation is not accurate -- `averager.average_parameters(model.parameters())` is not always correct, because if a model has untrainable parameters, then we shouldn't average all the parameters, so the implementation here is more accurate.

Thus, `averager.average_parameters(model.parameters())` is not added to PyTorch Lightning, and the code here is a better replacement.

daniellepintz: Sounds good, thanks @wayi1 for the info!!

ananthsub (approved):
@daniellepintz please also update the docs for DDP optimizations with this support

pytorch_lightning/strategies/ddp.py
tchaton: Thanks :)

pytorch_lightning/strategies/ddp.py
```diff
@@ -232,23 +236,18 @@ def _register_ddp_hooks(self) -> None:
                 import torch.distributed.algorithms.ddp_comm_hooks.post_localSGD_hook as post_localSGD
 
                 if isinstance(self._ddp_comm_state, post_localSGD.PostLocalSGDState):
-                    self._reinit_optimizers_with_post_localSGD(self._ddp_comm_state.start_localSGD_iter)
+                    self._enable_model_averaging()
```
rohan-varma, line 239: I see, I guess this tightly couples localSGD state + DDP comm hook to model averaging component, this should be fine at the moment, although @wayi1 is also working on hierarchical SGD which might require additional design thought to integrate into lightning.

pytorch_lightning/strategies/ddp.py
```diff
@@ -257,24 +256,47 @@ def _reinit_optimizers_with_post_localSGD(self, warmup_steps: int):
...
+            closure: closure calculating the loss value
+            model: reference to the model, optionally defining optimizer step related hooks
+            **kwargs: Any extra arguments to ``optimizer.step``
+        """
+        optimizer_output = super().optimizer_step(optimizer, opt_idx, closure, model, **kwargs)
+
+        if not _TORCH_GREATER_EQUAL_1_10 or self._model_averager is None:
+            return optimizer_output
+
+        for opt in self.optimizers:
```
rohan-varma, line 292: Noob q because I am not familiar with lightning, but it seems `optimizer_step` will be called for each `optimizer` in `self.optimizer`s? 

If this is true, shouldn't we only average parameters that belong to the current optimizer that's running `optimizer_step`? Otherwise we'd indeed have multiple averages on the same param?

pytorch_lightning/strategies/ddp.py
daniellepintz: sounds good!

pytorch_lightning/strategies/ddp.py
daniellepintz: yes I believe you are right! I was just copying the logic directly from #9446 but it seems this should be changed! Will fix

daniellepintz:
I updated the docs but something seems wrong with the way it is displaying in the CI. The "note" and code sections get combined: https://153988-178626720-gh.circle-artifacts.com/0/html/advanced/model_parallel.html#ddp-optimizations compared to the current docs: https://pytorch-lightning.readthedocs.io/en/latest/advanced/model_parallel.html?highlight=ddp#ddp-optimizations

however it seems this is an issue in all CI jobs, for example this is the same page from https://github.com/PyTorchLightning/pytorch-lightning/pull/12245: https://153982-178626720-gh.circle-artifacts.com/0/html/advanced/model_parallel.html#ddp-optimizations

pytorch_lightning/strategies/ddp.py
wayi1: Thanks for bringing this up @rohan-varma. This should be compatible with hierarchical SGD, which will still use localSGD state + DDP comm hook.

daniellepintz:
Confirmed the docs on master look good: https://pytorch-lightning.readthedocs.io/en/latest/advanced/model_parallel.html?highlight=ddp#ddp-optimizations

So it was just a weird thing with the CI

PR #12379, opened 2022-03-20
Fix broken horovod tests

JoostvDoorn:
## What does this PR do?

Fixes horovod tests that were broken in some previous PRs due to CI not running them, as mentioned here https://github.com/PyTorchLightning/pytorch-lightning/pull/12377#issuecomment-1073066919. This PR depends on https://github.com/PyTorchLightning/pytorch-lightning/pull/12377 for the tests to actually run. For me locally the test test_nccl_is_available_on_gpu_environment is still failing, I haven't looked into this. I separated this code from https://github.com/PyTorchLightning/pytorch-lightning/pull/12377 since I think it may be slightly easier to discuss this separately, but I can also merge the code into that PR if desired.

I have also been testing horovod with the example that's located here https://github.com/horovod/horovod/blob/master/examples/pytorch/pytorch_lightning_mnist.py, and these changes are important to get this to work on a gpu.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/tests/models/test_horovod.py
+++ b/tests/models/test_horovod.py
@@ -71,6 +71,8 @@ def _run_horovod(trainer_options):
     ]
     if trainer_options.get("accelerator", "cpu") == "gpu":
         cmdline += ["--on-gpu"]
+    if devices == 2:
+        cmdline += ["--check-size"]
     exit_code = subprocess.call(" ".join(cmdline), shell=True, env=os.environ.copy())
     assert exit_code == 0
 
@@ -93,7 +95,7 @@ def test_horovod_cpu(tmpdir):
 @RunIf(skip_windows=True, horovod=True, skip_49370=True)
 def test_horovod_cpu_accumulate_grad_batches(tmpdir):
     trainer_options = dict(
-        default_root_dir=tmpdir,
+        default_root_dir=str(tmpdir),
         enable_progress_bar=False,
         max_epochs=1,
         limit_train_batches=4,
--- a/pytorch_lightning/trainer/connectors/accelerator_connector.py
+++ b/pytorch_lightning/trainer/connectors/accelerator_connector.py
@@ -623,7 +623,7 @@ class AcceleratorConnector:
         hvd.init()
         if isinstance(self.accelerator, GPUAccelerator):
             # Horovod assigns one local GPU per process
-            self._parallel_devices = list(range(hvd.local_size()))
+            self._parallel_devices = [torch.device(f"cuda:{i}") for i in range(hvd.local_size())]
         else:
             self._parallel_devices = [torch.device("cpu")] * hvd.local_size()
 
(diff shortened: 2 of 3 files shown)


tests/models/test_horovod.py
```diff
@@ -93,7 +95,7 @@ def test_horovod_cpu(tmpdir):
 @RunIf(skip_windows=True, horovod=True, skip_49370=True)
 def test_horovod_cpu_accumulate_grad_batches(tmpdir):
     trainer_options = dict(
-        default_root_dir=tmpdir,
+        default_root_dir=str(tmpdir),
```
awaelchli, line 98: interesting. was this needed because of the json dump on the other side?

tests/models/test_horovod.py
```diff
@@ -71,6 +71,8 @@ def _run_horovod(trainer_options):
     ]
     if trainer_options.get("accelerator", "cpu") == "gpu":
         cmdline += ["--on-gpu"]
+    if devices == 2:
```
awaelchli, line 74: was it meant to be this?
```suggestion
    if devices > 1:
```

tests/models/test_horovod.py
```diff
@@ -71,6 +71,8 @@ def _run_horovod(trainer_options):
...
+        cmdline += ["--check-size"]
```
awaelchli, line 75: I couldn't find this option in the horovodrun help page. Is this a new option?

awaelchli:
Thank you for investigating this!

awaelchli (approved):

tests/models/test_horovod.py
JoostvDoorn: This is an argument for the script tests/models/data/horovod/train_default_model.py that I add in this pr. This script has an assert checking if the replication is two in here: https://github.com/PyTorchLightning/pytorch-lightning/blob/c5ff4f903f8be20173a112fc89674d7bde3c5d98/tests/models/data/horovod/train_default_model.py#L75-L76

Since some tests will have hvd.size()==1 it won't pass the test in those cases.

tests/models/test_horovod.py
JoostvDoorn: Not really, but I suppose we could pass the number of devices instead as well. See also my other comments regarding this option.

tests/models/test_horovod.py
JoostvDoorn: Indeed, all other tests actually have this as well.

tests/models/test_horovod.py
awaelchli: ah right 🤦

Borda (approved):

rohitgr7 (approved):

akihironitta:
Hi @JoostvDoorn! Thank you for reporting and investigating the issue!

As suggested in https://github.com/PyTorchLightning/pytorch-lightning/pull/12377#issuecomment-1073231264, how about merging this PR to #12377 since we can't make sure that the change here fixes the failing tests in CI? I believe this is not too complicated to have both PRs together :]

PR #12418, opened 2022-03-23
`ModelCheckpoint`'s `save_last` now ignores `every_n_epochs`

### LINKED ISSUE #12391: Fix interaction with `save_last` and `every_n_epochs`
@carmocca for the following snippet, a `last.ckpt` is generated before this PR, but not anymore after:

```
import uuid

import torch
from pytorch_lightning import Trainer, LightningModule


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

if __name__ == "__main__":
    tmpdir = f"/tmp/{uuid.uuid4()}"
    print(tmpdir)

    trainer = Trainer(
        default_root_dir=tmpdir,
        max_epochs=1,
        callbacks=[ModelCheckpoint(dirpath=tmpdir, every_n_epochs=10, save_last=True)],
        enable_checkpointing=True,
    )
    model = BoringModel()
    trainer.fit(
        model, train_dataloaders=DataLoader(RandomDataset(32, 64), batch_size=2)
    )
```

The following contract was respected prior to this PR, but not anymore after:

> save_last: When True, always saves the model at the end of the epoch to a file last.ckpt

It is a BC-breaking change that changes a behavior some users rely on, regardless of whether it is believed to be a "bug".

_Originally posted by @yifuwang in https://github.com/PyTorchLightning/pytorch-lightning/issues/11805#issuecomment-1068230683_

cc @tchaton @rohitgr7 @akihironitta @carmocca @awaelchli @ninginthecloud @jjenniferdai

carmocca:
## What does this PR do?

Fixes #12391
And other minor bugs I noticed while working on this.

None

## Before submitting

## PR review

cc @borda @tchaton @rohitgr7 @akihironitta @carmocca @awaelchli @ninginthecloud @jjenniferdai

### DIFF
--- a/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/pytorch_lightning/callbacks/model_checkpoint.py
@@ -108,12 +108,10 @@ class ModelCheckpoint(Callback):
         every_n_train_steps: Number of training steps between checkpoints.
             If ``every_n_train_steps == None or every_n_train_steps == 0``, we skip saving during training.
             To disable, set ``every_n_train_steps = 0``. This value must be ``None`` or non-negative.
-            This must be mutually exclusive with ``train_time_interval`` and ``every_n_epochs``.
         train_time_interval: Checkpoints are monitored at the specified time interval.
             For all practical purposes, this cannot be smaller than the amount
             of time it takes to process a single training batch. This is not
             guaranteed to execute at the exact time specified, but should be close.
-            This must be mutually exclusive with ``every_n_train_steps`` and ``every_n_epochs``.
         every_n_epochs: Number of epochs between checkpoints.
             This value must be ``None`` or non-negative.
             To disable saving after each epoch, set ``every_n_epochs = 0``.
@@ -123,11 +121,11 @@ class ModelCheckpoint(Callback):
             If ``every_n_epochs == None`` and either ``every_n_train_steps != None`` or ``train_time_interval != None``,
             saving at the end of each epoch is disabled
             (equivalent to ``every_n_epochs = 0``).
-            This must be mutually exclusive with ``every_n_train_steps`` and ``train_time_interval``.
             Setting both ``ModelCheckpoint(..., every_n_epochs=V, save_on_train_epoch_end=False)`` and
             ``Trainer(max_epochs=N, check_val_every_n_epoch=M)``
             will only save checkpoints at epochs 0 < E <= N
             where both values for ``every_n_epochs`` and ``check_val_every_n_epoch`` evenly divide E.
+            This argument does not impact the saving of ``save_last=True`` checkpoints.
         save_on_train_epoch_end: Whether to run checkpointing at the end of the training epoch.
             If this is ``False``, then the check runs at the end of the validation.
 
@@ -238,7 +236,9 @@ class ModelCheckpoint(Callback):
 
         self.__init_monitor_mode(mode)
         self.__init_ckpt_dir(dirpath, filename)
-        self.__init_triggers(every_n_train_steps, every_n_epochs, train_time_interval)
+        self._train_time_interval = train_time_interval
+        self._every_n_epochs = 1 if every_n_epochs is None else every_n_epochs
+        self._every_n_train_steps = 0 if every_n_train_steps is None else every_n_train_steps
         self.__validate_init_configuration()
 
     @property
@@ -299,24 +299,13 @@ class ModelCheckpoint(Callback):
 
     def on_train_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         """Save a checkpoint at the end of the training epoch."""
-        if (
-            not self._should_skip_saving_checkpoint(trainer)
-            and self._save_on_train_epoch_end
-            and self._every_n_epochs > 0
-            and (trainer.current_epoch + 1) % self._every_n_epochs == 0
-        ):
+        if not self._should_skip_saving_checkpoint(trainer) and self._save_on_train_epoch_end:
             self.save_checkpoint(trainer)
 
     def on_validation_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         """Save a checkpoint at the end of the validation stage."""
-        if (
-            self._should_skip_saving_checkpoint(trainer)
-            or self._save_on_train_epoch_end
-            or self._every_n_epochs < 1
-            or (trainer.current_epoch + 1) % self._every_n_epochs != 0
-        ):
-            return
-        self.save_checkpoint(trainer)
+        if not self._should_skip_saving_checkpoint(trainer) and not self._save_on_train_epoch_end:
+            self.save_checkpoint(trainer)
 
     def on_save_checkpoint(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", checkpoint: Dict[str, Any]
@@ -358,19 +347,34 @@ class ModelCheckpoint(Callback):
         This method runs on all ranks. It is the responsibility of `trainer.save_checkpoint` to correctly handle the
         behaviour in distributed training, i.e., saving only on rank 0 for data parallel use cases.
         """
-        self._validate_monitor_key(trainer)
-
-        # what can be monitored
-        monitor_candidates = self._monitor_candidates(trainer, epoch=trainer.current_epoch, step=trainer.global_step)
-
-        # callback supports multiple simultaneous modes
-        # here we call each mode sequentially
-        # Mode 1: save the top k checkpoints
-        self._save_top_k_checkpoint(trainer, monitor_candidates)
-        # Mode 2: save monitor=None checkpoints
-        self._save_none_monitor_checkpoint(trainer, monitor_candidates)
-        # Mode 3: save last checkpoints
+        monitor_candidates = self._monitor_candidates(trainer)
+        self._save_topk_checkpoint(trainer, monitor_candidates)
         self._save_last_checkpoint(trainer, monitor_candidates)
+
+    def _save_topk_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
+        if self.save_top_k == 0 or self._every_n_epochs < 1 or (trainer.current_epoch + 1) % self._every_n_epochs != 0:
+            # `every_n_epochs` only applies to monitored checkpoints
+            return
+
+        # validate metric
+        if self.monitor is not None:
+            if self.monitor not in monitor_candidates:
+                m = (
+                    f"`ModelCheckpoint(monitor='{self.monitor}')` could not find the monitored key in the returned"
+                    f" metrics: {list(monitor_candidates)}."
+                    f" HINT: Did you call `log({self.monitor!r}, value)` in the `LightningModule`?"
+                )
+                if trainer.fit_loop.epoch_loop.val_loop._has_run:
+                    raise MisconfigurationException(m)
+                warning_cache.warn(m)
+            self._save_monitor_checkpoint(trainer, monitor_candidates)
+        else:
+            self._save_none_monitor_checkpoint(trainer, monitor_candidates)
+
+    def _actual_save_checkpoint(self, trainer: "pl.Trainer", filepath: str) -> None:
+        trainer.save_checkpoint(filepath, self.save_weights_only)
+
+        self._last_global_step_saved = trainer.global_step
 
         # notify loggers
         if trainer.is_global_zero:
@@ -396,16 +400,6 @@ class ModelCheckpoint(Callback):
             )
         if self._every_n_epochs < 0:
             raise MisconfigurationException(f"Invalid value for every_n_epochs={self._every_n_epochs}. Must be >= 0")
-
-        every_n_train_steps_triggered = self._every_n_train_steps >= 1
-        every_n_epochs_triggered = self._every_n_epochs >= 1
-        train_time_interval_triggered = self._train_time_interval is not None
-        if every_n_train_steps_triggered + every_n_epochs_triggered + train_time_interval_triggered > 1:
-            raise MisconfigurationException(
-                f"Combination of parameters every_n_train_steps={self._every_n_train_steps}, "
-                f"every_n_epochs={self._every_n_epochs} and train_time_interval={self._train_time_interval} "
-                "should be mutually exclusive."
-            )
 
         if self.monitor is None:
             # -1: save all epochs, 0: nothing is saved, 1: save last epoch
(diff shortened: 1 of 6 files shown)


tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1253,7 +1234,7 @@ def test_save_last_saves_correct_last_model_path(tmpdir):
     trainer = Trainer(callbacks=mc)
     trainer.strategy.connect(BoringModel())
 
-    mc._save_last_checkpoint(trainer, {"foo": 1})
+    mc._save_last_checkpoint(trainer, {"foo": 1, "step": 0})
```
carmocca, line 1237: This isn't a breaking change for the users. Only necessary here because it's an unit test

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1086,7 +1067,7 @@ def __init__(self, hparams):
             super().__init__()
             self.save_hyperparameters(hparams)
 
-    model_checkpoint = ModelCheckpoint(dirpath=tmpdir, save_top_k=1, monitor="foo")
+    model_checkpoint = ModelCheckpoint(dirpath=tmpdir, save_top_k=1)
```
carmocca, line 1070: The model doesn't log `foo` so this was a bug. It didn't fail before because `_is_valid_monitor_key` has `or len(metrics) == 0` and the length was 0 because we were using accessing `trainer.callback_metrics` instead of the existing `monitor_candidates`

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -675,10 +657,9 @@ def test_model_checkpoint_train_time_interval(mock_datetime, tmpdir) -> None:
     )
 
     trainer.fit(model)
-    # Each batch takes 7 sec and we checkpoint every minute. There are 64
-    # batches per epoch, so total time to run is 7*64*2 = 896 sec < 14.96 minutes,
-    # so we should have 14 checkpoints.
-    assert len(os.listdir(tmpdir)) == 14
+    # Each batch takes 7 sec and we checkpoint every minute. There are 64 batches per epoch, so total time to run is
+    # 7*64*2 = 896 sec < 14.96 minutes so we should have 14 checkpoints. +2 for those saved at the end of the each epoch
+    assert len(os.listdir(tmpdir)) == 14 + 2
```
carmocca, line 662: The triggers are not exclusive now so we also save at the end of each epoch (2 epochs)

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -625,7 +608,6 @@ def test_ckpt_every_n_train_steps(tmpdir):
     epoch_length = 64
     checkpoint_callback = ModelCheckpoint(
         filename="{step}",
-        every_n_epochs=0,
```
carmocca, line 628: The triggers aren't exclusive now so setting it to 0 would mean we save nothing.

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -109,7 +109,6 @@ def training_step(self, batch, batch_idx):
         def validation_step(self, batch, batch_idx):
             log_value = self.val_logs[self.current_epoch, batch_idx]
             self.log("val_log", log_value)
-            self.log("epoch", self.current_epoch, on_epoch=True)
```
carmocca, line 112: Since we use `setdefault` now inside `ModelCheckpoint`, this `log` call would set the `epoch` value but it would be a float so the filename would be `epoch=1.0.ckpt`.

Anyways, we don't need to log it anymore

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -108,26 +108,21 @@ class ModelCheckpoint(Callback):
...
-            This must be mutually exclusive with ``train_time_interval`` and ``every_n_epochs``.
         train_time_interval: Checkpoints are monitored at the specified time interval.
             For all practical purposes, this cannot be smaller than the amount
             of time it takes to process a single training batch. This is not
             guaranteed to execute at the exact time specified, but should be close.
-            This must be mutually exclusive with ``every_n_train_steps`` and ``every_n_epochs``.
         every_n_epochs: Number of epochs between checkpoints.
             This value must be ``None`` or non-negative.
-            To disable saving after each epoch, set ``every_n_epochs = 0``.
+            To disable saving top-k checkpoints, set ``every_n_epochs = 0``.
```
carmocca, line 117: This flag now only applies to top-k checkpoints (with monitor or monitor=None) as described in the issue

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -358,20 +354,34 @@ def save_checkpoint(self, trainer: "pl.Trainer") -> None:
...
+            self._save_monitor_checkpoint(trainer, monitor_candidates)
+        else:
+            self._save_none_monitor_checkpoint(trainer, monitor_candidates)
+
+    def _actual_save_checkpoint(self, trainer: "pl.Trainer", filepath: str) -> None:
+        trainer.save_checkpoint(filepath, self.save_weights_only)
+
+        self._last_global_step_saved = trainer.global_step
+
         # notify loggers
```
carmocca, line 385: Now we only do this when we actually save

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -352,26 +350,41 @@ def on_load_checkpoint(
...
-        monitor_candidates = self._monitor_candidates(trainer, epoch=trainer.current_epoch, step=trainer.global_step)
-
-        # callback supports multiple simultaneous modes
-        # here we call each mode sequentially
-        # Mode 1: save the top k checkpoints
-        self._save_top_k_checkpoint(trainer, monitor_candidates)
-        # Mode 2: save monitor=None checkpoints
-        self._save_none_monitor_checkpoint(trainer, monitor_candidates)
-        # Mode 3: save last checkpoints
+        # TODO: unused method. deprecate it
```
carmocca, line 359: Left for a follow-up PR

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -621,51 +619,40 @@ def _get_metric_interpolated_filepath_name(
...
-    def _save_top_k_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
-        if self.monitor is None or self.save_top_k == 0:
-            return
-        self._last_global_step_saved = monitor_candidates.get("step", trainer.global_step)
-
+    def _save_monitor_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
         current = monitor_candidates.get(self.monitor)
         if self.check_monitor_top_k(trainer, current):
             self._update_best_and_save(current, trainer, monitor_candidates)
         elif self.verbose:
```
krshrimali, line 643: Asking for my knowledge, what happens if none of the 2 conditions are True? (since there is no `else`)

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -352,26 +350,41 @@ def on_load_checkpoint(
...
+                    f" HINT: Did you call `log({self.monitor!r}, value)` in the `LightningModule`?"
+                )
+                if trainer.fit_loop.epoch_loop.val_loop._has_run:
+                    raise MisconfigurationException(m)
+                warning_cache.warn(m)
+            self._save_monitor_checkpoint(trainer, monitor_candidates)
+        else:
+            self._save_none_monitor_checkpoint(trainer, monitor_candidates)
+
+    def _actual_save_checkpoint(self, trainer: "pl.Trainer", filepath: str) -> None:
```
krshrimali, line 383: I might be missing some context here, but why is "actual" used here? Are there places where we intend to save checkpoints, but not actually save them?

tests/checkpointing/test_model_checkpoint.py
krshrimali: > but it would be a float so the filename would be

How do you feel about filenames having epoch numbers as floating points? Should we just cast them to integers instead?

tests/checkpointing/test_model_checkpoint.py
```diff
@@ -1281,3 +1280,24 @@ def test_last_global_step_saved():
     trainer.callback_metrics = {"foo": 123}
     model_checkpoint.save_checkpoint(trainer)
     assert model_checkpoint._last_global_step_saved == 0
+
+
+def test_save_last_every_n_epochs_interaction(tmpdir):
+    """Test that `save_last` ignores `every_n_epochs`."""
+    mc = ModelCheckpoint(every_n_epochs=5, save_last=True, save_top_k=0, save_on_train_epoch_end=True)
```
krshrimali, line 1287: Should we also test for `every_n_epochs=0` and `save_last=True`? To make sure that the last checkpoint is saved irrespective of `every_n_epochs`.

krshrimali (approved):
Thanks, @carmocca! Looks really good to me, I just have a few questions for my knowledge and a minor test extension suggestion. Shouldn't block merging this PR once everyone concerned is happy with it :)

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -621,51 +619,40 @@ def _get_metric_interpolated_filepath_name(
...
             trainer.strategy.remove_checkpoint(previous)
 
-    def _save_top_k_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
-        if self.monitor is None or self.save_top_k == 0:
-            return
-        self._last_global_step_saved = monitor_candidates.get("step", trainer.global_step)
-
+    def _save_monitor_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
         current = monitor_candidates.get(self.monitor)
         if self.check_monitor_top_k(trainer, current):
```
krshrimali, line 641: Looks like the tests are failing because we didn't validate the monitor key here. `monitor_candidates` will just have `step, epoch` as the keys here and `monitor_candidates.get(self.monitor)` will return `None` if not present. `if self.check_monitor_top_k` just returns False if current is None.

pytorch_lightning/callbacks/model_checkpoint.py
carmocca: Reworded

pytorch_lightning/callbacks/model_checkpoint.py
carmocca: Nothing happens, as in, we don't save a checkpoint

tests/checkpointing/test_model_checkpoint.py
carmocca: Probably it's best to do it so accidents are avoided.

tests/trainer/test_trainer.py
```diff
@@ -403,6 +403,7 @@ def test_model_freeze_unfreeze():
         assert param.requires_grad
 
 
+@pytest.mark.xfail(reason="FIXME(@carmocca): this test wasn't running and is now broken")
```
carmocca, line 406: Since the monitor does not exist and it did not fail, no checkpoints were written so the assertions did not run

Meaning, the test is broken in master.

Will work on fixing it in a separate PR

ananthsub (approved):

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -100,8 +100,8 @@ class ModelCheckpoint(Callback):
             based on either the maximization or the minimization of the monitored quantity.
             For ``'val_acc'``, this should be ``'max'``, for ``'val_loss'`` this should be ``'min'``, etc.
         auto_insert_metric_name: When ``True``, the checkpoints filenames will contain the metric name.
-            For example, ``filename='checkpoint_{epoch:02d}-{acc:02d}`` with epoch 1 and acc 80 will resolve to
-            ``checkpoint_epoch=01-acc=80.ckp``. Is useful to set it to ``False`` when metric names contain ``/``
+            For example, ``filename='checkpoint_{epoch:02d}-{acc:02.0f}`` with epoch ``1`` and acc ``1.12`` will resolve
+            to ``checkpoint_epoch=01-acc=01.ckp``. Is useful to set it to ``False`` when metric names contain ``/``
```
rohitgr7, line 104:
```suggestion
            to ``checkpoint_epoch=01-acc=01.ckpt``. Is useful to set it to ``False`` when metric names contain ``/``
```

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -352,26 +350,41 @@ def on_load_checkpoint(
...
 
+    def _save_topk_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
+        if self.save_top_k == 0:
+            return
+
+        # validate metric
+        if self.monitor is not None:
+            if self.monitor not in monitor_candidates:
+                m = (
+                    f"`ModelCheckpoint(monitor='{self.monitor}')` could not find the monitored key in the returned"
```
rohitgr7, line 372:
```suggestion
                    f"`ModelCheckpoint(monitor={self.monitor!r})` could not find the monitored key in the returned"
```

rohitgr7 (approved):
great! very small comments.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -352,26 +350,41 @@ def on_load_checkpoint(
...
+        monitor_candidates = self._monitor_candidates(trainer)
+        self._save_topk_checkpoint(trainer, monitor_candidates)
         self._save_last_checkpoint(trainer, monitor_candidates)
 
+    def _save_topk_checkpoint(self, trainer: "pl.Trainer", monitor_candidates: Dict[str, _METRIC]) -> None:
+        if self.save_top_k == 0:
+            return
+
+        # validate metric
+        if self.monitor is not None:
```
Borda, line 369:
```suggestion
        if self.monitor:
```
or can it be False?

Borda (approved):

pytorch_lightning/callbacks/model_checkpoint.py
carmocca: It could be an empty string (maybe? we don't have any tests for that actually)

I prefer being explicit

PR #12431, opened 2022-03-23
Do not configure launcher if processes are launched externally

ananthsub:
## What does this PR do?

If processes are created externally, Lightning should not use the subscript process launcher at all. Currently, this is very confusing when jobs are launched w/ torchelastic or SLURM, yet we see the subprocess script launcher configured here:
https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bfcc7407d3ca24716497806298c6a1d050b2be/pytorch_lightning/strategies/ddp.py#L138-L143

which in turn causes the trainer to invoke this path here: https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bfcc7407d3ca24716497806298c6a1d050b2be/pytorch_lightning/trainer/trainer.py#L722-L723

but the launcher itself checks this property again and calls the trainer function directly if processes are created externally: https://github.com/PyTorchLightning/pytorch-lightning/blob/a9bfcc7407d3ca24716497806298c6a1d050b2be/pytorch_lightning/strategies/launchers/subprocess_script.py#L81-L93

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/strategies/ddp.py
+++ b/pytorch_lightning/strategies/ddp.py
@@ -137,8 +137,8 @@ class DDPStrategy(ParallelStrategy):
         return self._process_group_backend
 
     def _configure_launcher(self) -> None:
-        self._launcher = _SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
         if not self.cluster_environment.creates_processes_externally:
+            self._launcher = _SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
             self._rank_0_will_call_children_scripts = True
 
     def setup_environment(self) -> None:
--- a/tests/strategies/test_ddp_strategy.py
+++ b/tests/strategies/test_ddp_strategy.py
@@ -19,7 +19,7 @@ import torch
 from torch.nn.parallel import DistributedDataParallel
 
 from pytorch_lightning import LightningModule, Trainer
-from pytorch_lightning.plugins.environments import LightningEnvironment
+from pytorch_lightning.plugins.environments import ClusterEnvironment, LightningEnvironment
 from pytorch_lightning.strategies import DDPStrategy
 from pytorch_lightning.trainer.states import TrainerFn
 from tests.helpers.boring_model import BoringModel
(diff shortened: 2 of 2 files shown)


awaelchli:
@ananthsub even if the condition `if self.strategy.launcher is not None: ` is True in a second call to e.g. Trainer.fit (inside a subprocess), we would expect the SubprocessLauncher to skip the call to `call_children_scripts` and only call the passed through function directly. In other words, the launcher should know itself whether it is in a subprocess already or not.

awaelchli (approved):
I think this is great as it makes the strategy compatible with the cluster environments that unconditionally return True for the `creates_processes_externally` property, but probably we have to revisit this carefully when DDPSpawnStrategy gets merged into DDPStrategy.

pytorch_lightning/strategies/ddp.py
```diff
@@ -137,8 +137,8 @@ def process_group_backend(self) -> Optional[str]:
         return self._process_group_backend
 
     def _configure_launcher(self) -> None:
-        self._launcher = _SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
         if not self.cluster_environment.creates_processes_externally:
+            self._launcher = _SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
             self._rank_0_will_call_children_scripts = True
```
carmocca, line 142: noob question unrelated to the PR:
why is this attribute not part of the launcher?

carmocca (approved):

pytorch_lightning/strategies/ddp.py
awaelchli: Because the value of this variable has to be broadcast by the strategy and the deadlock detection is currently part of the strategy. At the time of extracting the launcher out of the strategy, there was too much complexity to separate this cleanly. This has to be revisited if we want to combine ddp and ddp-spawn into one strategy.

tchaton (approved):
Nice catch !

rohitgr7 (approved):

PR #12461, opened 2022-03-25
[WIP] Accelerator registry follow-up

carmocca:
## What does this PR do?

My review of #12180

None as this is unreleased.

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -829,3 +831,9 @@ def is_distributed(self) -> bool:
         if isinstance(self.accelerator, TPUAccelerator):
             is_distributed |= self.strategy.is_distributed
         return is_distributed
+
+
+def _populate_registries() -> None:
```
kaushikb11, line 836: Why have it in `accelerator_connector`, not during import time `__init__`?

pytorch_lightning/accelerators/accelerator.py
```diff
@@ -76,6 +76,8 @@ def auto_device_count() -> int:
     def is_available() -> bool:
         """Detect if the hardware is available."""
 
-    @classmethod
-    def register_accelerators(cls, accelerator_registry: Dict) -> None:
-        pass
+    @staticmethod
+    @abstractmethod
+    def name() -> str:
```
kaushikb11, line 81: The discussion for removing the `name` method. https://github.com/PyTorchLightning/pytorch-lightning/pull/12180#discussion_r818023970

pytorch_lightning/trainer/connectors/accelerator_connector.py
kaushikb11: Why not? What's the benefit of delaying it? 

If I do this 

```python
    from pytorch_lightning.accelerators.registry import ACCELERATOR_REGISTRY
    print(ACCELERATOR_REGISTRY.names)
```

I would be expecting a list of accelerator names, rather than calling one more function `_populate_registeries` before it. Also `registeries` is plural, and we are populating only one Registry

carmocca: Doing things at import time is usually problematic. In addition to slowing time to start running, it means users are very limited when they want to override or customize behaviour.

> Also registeries is plural, and we are populating only one Registry

Yes, but we would eventually do the same changes for strategies.

kaushikb11: For strategies, we really can't get rid of `register_strategies`, as we have multiple configurations for certain strategies.
That's the major con of this approach, we can't have multiple subclasses with different defaults in our code.

We can't have an inconsistent api for registering strategies and accelerators.

Issue #12474, opened 2022-03-27
Standalone tests can silently fail


awaelchli:
## 🐛 Bug

The CI passes with a green check mark, but the logs clearly show that the standalone tests fail.

### To Reproduce

Can be seen on the[ latest commits on master](https://dev.azure.com/PytorchLightning/pytorch-lightning/_build/results?buildId=64144&view=logs&j=3afc50db-e620-5b81-6016-870a6976ad29&t=d9f671c5-a304-5675-5394-961fd7f98b9b). 

At least two parametrization of this test fail: test_progress_bar_max_val_check_interval

### Expected behavior

The CI fails when a test fails.

### Environment

PL master

### Additional context

As I always say, don't trust the CI. 

cc @tchaton @rohitgr7 @akihironitta @carmocca @borda

awaelchli:
@rohitgr7 I'm not 100% sure but it looks like this started to happen since #11657

Borda:
ok, just checking and we already have `set -e` which shall enforce upstream the error
https://github.com/PyTorchLightning/pytorch-lightning/blob/2e5728a4841cb1d3b75123ac941cd36f681f9a11/.azure-pipelines/gpu-tests.yml#L112

rohitgr7:
this is weird. Just checked, test is passing on lightning cluster

carmocca:
Did you try using the script?

`./tests/standalone_tests.sh -k test_progress_bar_max_val_check_interval`

and running all standalone tests?

rohitgr7:
Thanks, @carmocca!! Now it's failing. Let me check the issue.

btw how is this different from running the tests manually?

carmocca:
It manages calling multiple (or all) standalone tests with a printed report at the end.

rohitgr7:
okay.. but still trying to understand why it failed when running with `./tests/standalone_tests.sh` and not when running manually, although the test had a bug.

awaelchli:
@rohitgr7 As to why the CI doesn't pick it up, I guess it's because the test passed on rank 0 but not on rank 1. The error on rank 1 does not seem to get picked up even though it appears in the logs. If one added a barrier at the end of the test, we most likely would have seen a hang instead of a silent pass. 

This will happen again in the future, so the issue is not yet resolved. I vote for reopening this.

rohitgr7:
> This will happen again in the future, so the issue is not yet resolved. I vote for reopening this.

[...]

PR #7863, opened 2021-06-07
[Test] Add extra test for val_check_interval in distributed scenario

tchaton:
## What does this PR do?

From some investigation, the current behaviour seems correct. 

When using val_check_interval. If you have 8 train batches, and val_check_interval is 0.2, val_check_batch=1 as `val_check_batch=max(1, int(train_batches * val_check_interval))`.

Alternatively, we could randomly sample int(1 / val_check_interval) batches from train to run the validation on, but the validation won't be run at the frequency, but it might be confusing for the users.

This PR adds a test to make sure the correct number total number of batches is selected.

Additionally, it adds a section in the doc to compute the total number of batches.

Fixes #7751

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [ ] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
Make sure you had fun coding 🙃

PR #12563, opened 2022-04-01
Run main progress bar independent of val progress bar in `TQDMProgressBar`

rohitgr7:
## What does this PR do?

Fixes some issues with tqdm progress bar. Explained in comments

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @rohitgr7


pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -268,8 +268,9 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", *_: Any) -> None:
         self.main_progress_bar.set_description(f"Epoch {trainer.current_epoch}")
 
     def on_train_batch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", *_: Any) -> None:
-        if self._should_update(self.train_batch_idx, self.total_train_batches):
-            _update_n(self.main_progress_bar, self.train_batch_idx + self._val_processed)
+        current = self.train_batch_idx + self._val_processed
```
rohitgr7, line 271: here we need to track the current index for the main progress bar. With `val_check_interval`, this was incorrect. Added a test that fails on master.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -294,10 +295,12 @@ def on_validation_batch_start(
         self.val_progress_bar.set_description(f"{desc} DataLoader {dataloader_idx}")
 
     def on_validation_batch_end(self, trainer: "pl.Trainer", *_: Any) -> None:
-        if self._should_update(self.val_batch_idx, self.total_val_batches_current_dataloader):
+        if self._should_update(self.val_batch_idx, self.val_progress_bar.total):
             _update_n(self.val_progress_bar, self.val_batch_idx)
-            if trainer.state.fn == "fit":
-                _update_n(self.main_progress_bar, self.train_batch_idx + self._val_processed)
+
+        current = self.train_batch_idx + self._val_processed
```
rohitgr7, line 301: here, the `main progress bar` now updates independently of `val progress bar`. So in case `val progress bar` doesn't need an update but the `main progress bar` does, it will update.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -338,7 +341,7 @@ def on_predict_batch_start(
         self.predict_progress_bar.set_description(f"{self.predict_description} DataLoader {dataloader_idx}")
 
     def on_predict_batch_end(self, *_: Any) -> None:
-        if self._should_update(self.predict_batch_idx, self.total_predict_batches_current_dataloader):
+        if self._should_update(self.predict_batch_idx, self.predict_progress_bar.total):
```
rohitgr7, line 344: we assign `total` to the progress bar already, so we can extract it from there itself. Just makes it less error-prone.

pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -318,7 +321,7 @@ def on_test_batch_start(
         self.test_progress_bar.set_description(f"{self.test_description} DataLoader {dataloader_idx}")
 
     def on_test_batch_end(self, *_: Any) -> None:
-        if self._should_update(self.test_batch_idx, self.total_test_batches_current_dataloader):
```
carmocca, line 321: All these `self.total_*` properties are still used by Rich, correct?

Does that mean Rich also has this problem?

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -347,10 +348,10 @@ def test_tqdm_progress_bar_value_on_colab(tmpdir):
         [2, 3, 1, [1, 2, 3, 4, 5], [1, 2, 3]],
         [0, 0, 3, None, None],
         [1, 0, 3, [1], None],
-        [1, 1, 3, [1, 2], [1]],
```
carmocca, line 350: Can you explain why this changed?

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -347,10 +348,10 @@ def test_tqdm_progress_bar_value_on_colab(tmpdir):
...
+        [1, 1, 3, [2], [1]],
         [5, 0, 3, [3, 5], None],
-        [5, 2, 3, [3, 5, 7], [2]],
-        [5, 2, 6, [5, 7], [2]],
```
carmocca, line 353: Same

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -549,16 +550,56 @@ def test_tqdm_progress_bar_can_be_pickled():
...
+    total_val_batches = limit_batches * val_checks_per_epoch
+
+    assert pbar_callback.val_progress_bar.n == limit_batches
+    assert pbar_callback.val_progress_bar.total == limit_batches
+    assert pbar_callback.main_progress_bar.n == limit_batches + total_val_batches
+    assert pbar_callback.main_progress_bar.total == limit_batches + total_val_batches
+    assert pbar_callback.is_enabled
+
+
+@RunIf(min_gpus=2, standalone=True)
```
carmocca, line 594: I know you didn't add it, but why is this a gpu standalone test?

Pbar doesn't have any distributed specific logic...

pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: ideally no.. since
```py
self.test_progress_bar.total == self.total_test_batches_current_dataloader
```
since I had to change it for the main progress bar, changed then here too.

But yes rich has some issues currently in some edge cases. Actually, I started with Rich, and ended up fixing tqdm.. there will be a follow-up for rich

tests/callbacks/test_tqdm_progress_bar.py
rohitgr7: yes, so since `train batches = 1` and `val batches = 1` and refresh rate is 2, so ideally main progress bar should only update when total progress is 2 since it contains train+val.

tests/callbacks/test_tqdm_progress_bar.py
rohitgr7: same here. `train batches = 5` and `val batches = 2` thus `total = 7` and refresh_rate=3, so for main progress bar updates should happen with multiples of 3 i.e 3, 6, 7(this is the end)

tests/callbacks/test_tqdm_progress_bar.py
rohitgr7: will remove it.. might have been added by mistake.
but the interesting thing is that this test was failing with `standalone=True` but not normally. #12474

will investigate more on this.

awaelchli: If we follow git blame back, it was added here #7863. I guess the decorator and global_zero check was added because the source issue was reported for a multi-gpu/node case.

tests/callbacks/test_tqdm_progress_bar.py
carmocca: I disagree here. The refresh rate is meant to reduce step-level updates but I don't think it should apply to epoch-level updates.

With this change, metrics logged on_epoch do not appear in the training progress and I think that's counterintuitive for the user, even if it's a stricter view of the refresh rate concept.

tests/callbacks/test_tqdm_progress_bar.py
carmocca: (Can be done in a follow-up) I still suggest we remove it from this test. Standalone tests are the current bottleneck in CI

carmocca (approved):
LGTM

awaelchli (approved):

CHANGELOG.md
```diff
@@ -82,6 +82,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Run main progress bar independent of val progress bar in `TQDMProgressBar` ([#12563](https://github.com/PyTorchLightning/pytorch-lightning/pull/12563))
```
awaelchli, line 85: this could be misinterpreted. the two bars are not independent. the main bar always contains the count of all validation steps. 

```suggestion
- Run main progress bar updates independent of val progress bar updates in `TQDMProgressBar` ([#12563](https://github.com/PyTorchLightning/pytorch-lightning/pull/12563))
```

PR #12588, opened 2022-04-03
Support deterministic="warn" in Trainer for Pytorch 1.11+

weiji14:
## What does this PR do?

In Pytorch 1.11, operations that do not have a deterministic implementation can be set to throw a warning instead of an error when ran in deterministic mode. See https://pytorch.org/docs/1.11/generated/torch.use_deterministic_algorithms.html

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?


carmocca (approved):
Cool! Pushed a commit with some minor changes

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -79,13 +80,26 @@
...
 
 log = logging.getLogger(__name__)
 
 if _HOROVOD_AVAILABLE:
     import horovod.torch as hvd
 
+if _PYTHON_GREATER_EQUAL_3_8_0:
+    _LITERAL_WARN = Literal["warn"]
+else:
+    # there's an issue with the LightningCLI's help message using Literal on 3.7
```
carmocca, line 100: @mauvilsa looks like `Literal` does not work with the CLI and Python 3.7

https://github.com/PyTorchLightning/pytorch-lightning/runs/5843851902?check_suite_focus=true#step:17:3177

mauvilsa: `Literal` was introduced in python 3.8, see https://docs.python.org/3/library/typing.html#typing.Literal

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -18,6 +18,7 @@
 from typing import Dict, List, Optional, Union
 
 import torch
+from typing_extensions import Literal
```
mauvilsa, line 21: jsonargparse currently does not consider backports in `typing_extensions`. Could be added.

mauvilsa: It is now added, [jsonargparse/commit/16cec7e](https://github.com/omni-us/jsonargparse/commit/16cec7e8a6aff99be5995eafa232b6e3b5c46dc2). Will be part of the next release.

akihironitta (approved):
Nice addition!

---

The type check is failing in CI: https://github.com/PyTorchLightning/pytorch-lightning/runs/5853483314?check_suite_focus=true

Once the failure is resolved, LGTM!

rohitgr7 (approved):

weiji14:
Cool, thanks everyone for sorting out the type hints! Great to learn about "Literal" types :laughing:

weiji14:
> I find this is a confusing feature. Either I want determinism or not. Something in-between like "warn" does not exist for me. Perhaps I'm missing the use case here.

This thread might be a useful read: https://github.com/pytorch/pytorch/issues/64883. Ideally yes, full determinism would be desirable, but in practice, not every algorithm has a deterministic implementation and this gets us 90% of the way there (depending on the model being used). I'm coming at it more from a reproducibility standpoint, it's better to be able to reproduce a model's result by 90% than to not be able to reproduce it at all because of all the randomness going on.

.github/workflows/code-checks.yml
```diff
@@ -37,15 +30,8 @@ jobs:
           ${{ runner.os }}-pip-
 
     - name: Install dependencies
-      env:
-        # TORCH_VERSION: ${{ matrix.pytorch-version }}
-        TORCH_VERSION: "1.10"
```
Borda, line 42: let's keep the pin version, and if needed, change it to 1.11

Borda (changes requested):

Borda (approved):

carmocca (approved):

mareikethies:
Hi, nice update which is useful for me. Could it be that the add_argparse_args() method in Trainer needs to be adapted as well? Currently, it accepts only boolean values for the deterministic keyword. At least I don't know how to pass the warn flag using argparse. Thanks!

weiji14:
> Yes, I did install from github today. I have not had a look at the source code, but I observe that when I run something like `python train.py --deterministic warn` using `Trainer.add_argparse_args()` in train.py, I get this error: `error: argument --deterministic: invalid str_to_bool value: 'warn'`. But I might miss something here or pass it the wrong way. Just plain `python train.py --deterministic` works.

Thanks @mareikethies, that error message is very helpful. I can confirm the `invalid str_to_bool value: 'warn'` error. Minimal working example:

1. Do `pip install https://github.com/PyTorchLightning/pytorch-lightning/archive/6490996b3974019a254f2750f5c01e6d38ff6a6e.zip`
2. Create a file called `train_determinstic.py` like so:

```python
import argparse
import torch


class BasicNet(pl.LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(2, 2)

    def forward(self, x):
        return self.layer(x)

        output = self(batch)
        loss = torch.nn.functional.mse_loss(batch, output)
        return {"loss": loss}

        return torch.optim.Adam(params=self.parameters(), lr=0.1, weight_decay=0.0005)


class RandomDataset(torch.utils.data.Dataset):
    def __init__(self):
        pass

    def __len__(self):
        return 1

    def __getitem__(self, idx: int):
        return torch.randn(2)


def cli_main():
    parser = pl.Trainer.add_argparse_args(parent_parser=parser)
    args = parser.parse_args()
    trainer = pl.Trainer.from_argparse_args(args=args, max_epochs=1, accelerator="auto")

    model = BasicNet()
    dataloader = torch.utils.data.DataLoader(dataset=RandomDataset())

    trainer.fit(model=model, train_dataloaders=dataloader)


if __name__ == "__main__":
    cli_main()
```

3. Run `python deterministic.py --deterministic=warn`, which will produce `deterministic.py: error: argument --deterministic: invalid str_to_bool value: 'warn'`

So my guess is that the code here might need to change:
https://github.com/PyTorchLightning/pytorch-lightning/blob/0fe2ae74e919c8ef220a029cab2203572ca536a5/pytorch_lightning/utilities/argparse.py#L232-L251

PR #12597, opened 2022-04-03
Provide access to unwrapped model in Lite

### LINKED ISSUE #12596: Provide access to unwrapped module in Lite
## 🚀 Feature

LightningLite has a method `setup` which wraps the model:

```py
model, optimizer = self.setup(model, optimizer)
```

Here model is an instance of `_LiteModule` and to access the attributes of the original model, you would do
`model.module.method_a()` instead of `model.method_a()`. 

However, depending on the Strategy, the model may be wrapped by other modules like `DistributedDataParallel` for example.

### Motivation

Normally one only calls the model through forward. However, in some cases it is useful to expose extra method on the model that don't do distributed communication. 

### Pitch

Make `_LiteModule.module` return the original unwrapped model, as it was passed into the setup() method.

### Additional context

awaelchli:
## What does this PR do?

Fixes #12596

This PR enables attribute pass-through on the _LiteModule wrapper. When an attribute is not found on the wrapper, it tries to look it up on the original model that was passed into the setup method.

The _LiteModule wrapper now takes an optional argument `original_module`, which is the one passed in to the `Lite.setup` method. If we didn't do this, we wouldn't be able to access it, because the `forward_module` is potentially already wrapped by something like `DistributedDataParallel`.

## Breaking change

This is a breaking change if you previously relied on `LiteModule.state_dict()` to save or load checkpoints. The old checkpoint won't load after this PR because of missing keys (_original_module, _forward_module).

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7 @carmocca @justusschock @awaelchli


pytorch_lightning/lite/wrappers.py
```diff
@@ -65,23 +65,27 @@ def step(self, closure: Optional[Callable] = None) -> Any:
...
         super().__init__()
-        self._module = module
+        self._forward_module = forward_module
+        self._original_module = original_module
         self._precision_plugin = precision_plugin
 
     @property
     def module(self) -> nn.Module:
-        return self._module
+        return self._original_module or self._forward_module
```
justusschock, line 88: Should we really have a separate reference here? Can't we just call the unwrap function, we used to have?

awaelchli: The "unwrap" function is implemented for LightningModule and has a stopping condition with an isinstance check against LightningModule. This can't be so easily converted to an equivalent unwrapping method for pure nn.Modules.
Afaik, this is the safest way to guarantee we actually obtain the original untouched model that was passed in to the Lite.setup method.

Borda (approved):

pytorch_lightning/lite/wrappers.py
carmocca: Off topic to the PR: I find the unwrapping logic we have in Lightning hard to follow. I would like that we clean it up someday

This solution by Adrian is clean and explicit IMO

pytorch_lightning/lite/wrappers.py
```diff
@@ -102,12 +108,21 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
...
         to_type = torch.get_default_dtype()
         output = apply_to_collection(output, function=_convert_float_tensor, dtype=Tensor)
         return output
 
+    def __getattr__(self, item: Any) -> Any:
+        try:
+            # __getattr__ gets called as a last resort if the attribute does not exist
+            # call nn.Module's implementation first
+            return super().__getattr__(item)
+        except AttributeError:
```
carmocca, line 122: Why the try-except instead of `hasattr`?

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
--
+- Added attribute lookup pass-through for the wrapped model returned by `LightningLite.setup()` ([#12597](https://github.com/PyTorchLightning/pytorch-lightning/pull/12597))
```
carmocca, line 12: If this is set to the 1.6.x milestone, shouldn't this be in "Fixed"?

carmocca (approved):

pytorch_lightning/lite/wrappers.py
justusschock: Afaik `hasattr` is implemented by calling `getattr` and catching `AttributeError`. So if you would call `hasattr` here, `getattr` would call this very method again and you'd end up with recursion errors since you can only call this on an instance (`self`) and not the superclass

pytorch_lightning/lite/wrappers.py
justusschock: I agree, The only motivation for asking was consistency so that when we clean it in one place, it would still all behave the same :)

pytorch_lightning/lite/wrappers.py
awaelchli: Yes exactly, that is the reason. 
However, there is still something weird going on with one of the failing deepspeed test and haven't figured it out yet. 

```
tests/lite/test_lite.py:459: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pytorch_lightning/lite/lite.py:405: in _run_impl
    return self._strategy.launcher.launch(run_method, *args, **kwargs)
pytorch_lightning/strategies/launchers/subprocess_script.py:93: in launch
    return function(*args, **kwargs)
pytorch_lightning/lite/lite.py:412: in _run_with_strategy_setup
    return run_method(*args, **kwargs)
tests/lite/test_lite.py:402: in run
    model, optimizer = self.setup(model, optimizer)
pytorch_lightning/lite/lite.py:174: in setup
    model = _LiteModule(model, self._precision_plugin, original_module=original_model)
/usr/local/lib/python3.7/dist-packages/deepspeed/runtime/zero/partition_parameters.py:249: in wrapper
    if not hasattr(module, "_ds_child_entered"):
pytorch_lightning/lite/wrappers.py:124: in __getattr__
    return getattr(self._original_module, item)
pytorch_lightning/lite/wrappers.py:124: in __getattr__
    return getattr(self._original_module, item)
pytorch_lightning/lite/wrappers.py:124: in __getattr__
    return getattr(self._original_module, item)
E   RecursionError: maximum recursion depth exceeded while calling a Python object
!!! Recursion detected (same locals & position)
```

My suspicion is that deepspeed does some patching which I haven't found yet.

awaelchli: Ok, finally figured this out. Deepspeed patches the init globally for all nn Modules. That wrapper called hasattr which then called our getattr and at this point `_original_module ` wasn't defined yet so it would recurse infinitely.

akihironitta (commented):
lgtm 🚀

otaj (approved):

justusschock (approved):

rohitgr7 (approved):

awaelchli:
Gatekeeper merge

PR #9443, opened 2021-09-10
[bugfix] Revert inference mode support

ananthsub:
## What does this PR do?

Fixes #9431
Reverts  #8813 until torchmetrics integration is sorted out

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

Issue #8499, opened 2021-07-21
Support for PyTorch's optimize_for_inference mode


ananthsub:
## 🚀 Feature

Leverage PyTorch's optimize_for_inference mode for performance benefits during model evaluation and inference

PyTorch has recently introduced an experimental API `optimize_for_inference`:
- https://pytorch.org/docs/master/generated/torch.inference_mode.html#torch.inference_mode
- https://pytorch.org/docs/master/notes/autograd.html#locally-disabling-gradient-computation 
- https://github.com/pytorch/pytorch/issues/56608 / https://github.com/pytorch/pytorch/pull/58045

### Motivation

Reap performance improvements

### Pitch

This can be used during `Trainer.predict` in place of the `no_grad` if optimize_for_inference is available: https://github.com/PyTorchLightning/pytorch-lightning/blob/4c79b3a5b343866217784c66d122819c59a92c1d/pytorch_lightning/trainer/trainer.py#L1078-L1083

### Alternatives

Keep as is

#### If you enjoy PL, check out our other projects:

- [Metrics](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.
- [Flash](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning
- [Bolts](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch
- [Lightning Transformers](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

carmocca:
We could also enable it for test, right?

ananthsub:
> We could also enable it for test, right?

Yup! I wonder if we could with validation too, as long as we're careful to reset the mode before the next train loop call

carmocca:
> I wonder if we could with validation too, as long as we're careful to reset the mode before the next train loop call

Not sure about this one, I can image somebody saving a validation `inference-only` tensor and trying use it on a computation that has a training tensor. This would fail, right?

PR #12715, opened 2022-04-11
Enable inference mode for evaluation

### LINKED ISSUE #11018: use `torch.inference_mode()` in `Trainer.predict`
## 🚀 Feature

pytorch 1.9 introduced a new decorator/context manager specifically for model inference `torch.inference_mode()`. Like `torch.no_grad()` it disables gradient tracking for faster forward passes through a model and is, as the name implies, useful for model inference. Unlike `torch.no_grad()`, it disables view tracking, so computations made using `torch.inference_mode()` can not later be used in computations that require gradients (as opposed to `torch.no_grad()`). Disabling view tracking further speeds up model forward passes.

### Motivation

The motivation here is to implement `torch.inference_mode()` inside the `Trainer.predict()` logic, replacing any occurrences of `torch.no_grad()`. Because this code right now is contextually used for inference, I don't believe the lack of view tracking represents a significant drawback/limitation for downstream code. Conceptually, clients of `Trainer.predict()` should not be relying on these tensors for downstream code that requires gradient calculation. However, clients of this code are likely relying on pytorch lightning to transparently scale and speed up their own code, and using `torch.inference_mode()` will further this goal.

### Pitch

change
```python
def _run_predict(self) -> Optional[_PREDICT_OUTPUT]:
        self.reset_predict_dataloader(self.lightning_module)
        # reset trainer on this loop and all child loops in case user connected a custom loop
        self.predict_loop.trainer = self
        with torch.no_grad():
            return self.predict_loop.run()
```

to
```python
def _run_predict(self) -> Optional[_PREDICT_OUTPUT]:
        self.reset_predict_dataloader(self.lightning_module)
        # reset trainer on this loop and all child loops in case user connected a custom loop
        self.predict_loop.trainer = self
        with torch.inference_mode():
            return self.predict_loop.run()
```

I think this would be the only change to make

cc @borda

### LINKED ISSUE #11530: inference_mode
PyTorch has an `inference_mode` (https://pytorch.org/docs/stable/generated/torch.inference_mode.html) that it recommends as a replacement for `no_grad`. But I don't think lightning is currently using it. Should it?

rohitgr7:
## What does this PR do?

Fixes #11530
Fixes #11018

## Before submitting

## PR review

## Did you have fun?

cc @borda


pytorch_lightning/trainer/trainer.py
```diff
@@ -2786,6 +2788,17 @@ def configure_optimizers(self):
         return max_estimated_steps
 
 
+@contextmanager
+def _evaluation_context() -> Generator:
+    cm = (
```
carmocca, line 2793: nit: we can spare some letters :)

```suggestion
    context_manager_class = (
```

carmocca (approved):
Awesome!

ananthsub:
We previously had support for this but needed to revert it: https://github.com/PyTorchLightning/pytorch-lightning/issues/8499
https://github.com/PyTorchLightning/pytorch-lightning/pull/9443

have you confirmed this is safe to roll out now?

rohitgr7:
> 

we have extensive tests that seem to be passing with this change, so a good indicator that this is safe. I have also confirmed that it's working with the example shared in the issue that caused the revert last time. But even if it still has some issues, we should explore the issue more in-depth rather than reverting it just like last time. I couldn't find anything related in the PyTorch docs that some backend like `gloo` is not supported or any other limitation. So I think it's safe to roll out now and explore the issue more next time if any comes up.

rohitgr7:
> We previously had support for this but needed to revert it: #8499 #9443
> 
> have you confirmed this is safe to roll out now?

we have extensive tests that seem to be passing with this change, so a good indicator that this is safe. I have also confirmed that it's working with the example shared in the issue that caused the revert last time. But even if it still has some issues, we should explore the issue more in-depth rather than reverting it just like last time. I couldn't find anything related in the PyTorch docs that some backend like gloo is not supported or any other limitation. So I think it's safe to roll out now and explore the issue more next time if any comes up.

pytorch_lightning/trainer/trainer.py
rohitgr7: done!

krshrimali (approved):
LGTM! 🚀

akihironitta (approved):
LGTM! 🚀

Borda (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2744,6 +2746,17 @@ def configure_optimizers(self):
         return max_estimated_steps
 
 
+@contextmanager
+def _evaluation_context() -> Generator:
+    context_manager_class = (
+        torch.inference_mode
+        if _TORCH_GREATER_EQUAL_1_9 and not (dist.is_initialized() and dist.get_backend() == "gloo")
```
awaelchli, line 2753: I don't find any information in docs or anywhere about inference_mode not being compatible with the gloo backend. A comment here in the code would probably be appropriate.

awaelchli (approved):

pytorch_lightning/trainer/trainer.py
```diff
@@ -2744,6 +2746,17 @@ def configure_optimizers(self):
         return max_estimated_steps
 
 
+@contextmanager
+def _evaluation_context() -> Generator:
```
rohitgr7, line 2750:
```suggestion
def _evaluation_context() -> Generator:
    # inference mode is not supported on gloo backend
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -2744,6 +2746,18 @@ def configure_optimizers(self):
...
+    # inference mode is not supported on gloo backend
```
rohitgr7, line 2751:
```suggestion
    # inference mode is not supported with gloo backend (#9431)
```

krshrimali (approved):

pytorch_lightning/trainer/trainer.py
awaelchli: thanks for adding the comment. i'm definitely not satisfied still, and will investigate. this is very sus, especially since I cannot find any open or closed issue ticket on the pytorch github.

rohitgr7: I couldn't find it either. Let me open an issue on PT GitHub.

PR #12744, opened 2022-04-13
Convolution on habana accelerators

jerome-habana:
## What does this PR do?

 Gaudi HW performs convolution operations with filter (weights) in filters last format - RSCK format where:
 R = height of the filter ,S = width of the filter, C = number of channels per filter,K = number of filters
 The default PyTorch convolution weight ordering is ‘filters first’ (KCRS).
 Therefore a re-ordering/permutation of all the convolution weights from KCRS to RSCK format is required
 before convolution operations

Added a new helper to aid in permutes before moving the model to Gaudi

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/docs/source/accelerators/hpu.rst
+++ b/docs/source/accelerators/hpu.rst
@@ -147,6 +147,7 @@ The subsequent snippet showcases the methodology to do the param permutes when u
 
     # Gaudi HW performs convolution operations with filter (weights) in filters last format
     from pytorch_lightning.utilities.hpu_device import HPUDeviceUtils
+
     HPUDeviceUtils.permute_params(model, True)
 
     # Train the model ⚡
--- a/docs/source/accelerators/hpu.rst
+++ b/docs/source/accelerators/hpu.rst
@@ -112,6 +112,49 @@ This enables advanced users to provide their own BF16 and FP32 operator list ins
 
 For more details, please refer to `PyTorch Mixed Precision Training on Gaudi <https://docs.habana.ai/en/latest/PyTorch_User_Guide/PyTorch_User_Guide.html#pytorch-mixed-precision-training-on-gaudi>`__.
 
+
+Enabling Models with Convolution layers
+---------------------------------------
+Convolution operations are central to vision topologies like ResNet. Gaudi HW performs convolution operations with filter (weights) in filters last format - RSCK format where:
+R = height of the filter; S = width of the filter; C = number of channels per filter; K = number of filters
+
+The default PyTorch convolution weight ordering is ‘filters first’ (KCRS). Therefore a re-ordering/permutation of all the convolution weights from KCRS to RSCK format is required before convolution operations.
+Such a permutation of weights is done once at the beginning of training in the PyTorch Habana vision topologies.
+Post permutation the weights are in RSCK format during training, a conversion back to KCRS format is necessary when saving intermediate checkpoints or saving the final trained weights.
+This helps bring the weights back to the default PyTorch format (KCRS), say, for use across DL training platforms/accelerators.
+
+Due to the permutation of weights to RSCK format, the gradients of these weights will also be in the same format on the HPU automatically.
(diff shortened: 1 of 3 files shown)


docs/source/accelerators/hpu.rst
```diff
@@ -112,6 +112,50 @@ This enables advanced users to provide their own BF16 and FP32 operator list ins
 
 For more details, please refer to `PyTorch Mixed Precision Training on Gaudi <https://docs.habana.ai/en/latest/PyTorch_User_Guide/PyTorch_User_Guide.html#pytorch-mixed-precision-training-on-gaudi>`__.
 
+
+Enabling Models with Convolution layers
+---------------------------------------
+Convolution operations are central to vision topologies like ResNet. Gaudi HW performs convolution operations with filter (weights) in filters last format - RSCK format where:
+R = height of the filter; S = width of the filter; C = number of channels per filter; K = number of filters
```
justusschock, line 119: Can we use the terms PyTorch uses in its docs for this: KCHW?

pl_examples/hpu_examples/convolution_example.py
```diff
@@ -0,0 +1,93 @@
...
+val_loader = DataLoader(val_ds, batch_size=16)
+
+# Initialize a trainer
+trainer = pl.Trainer(accelerator="hpu", max_epochs=1)
+
+if _HPU_AVAILABLE:
+    # Gaudi HW performs convolution operations with filter (weights) in filters last format
+    from pytorch_lightning.utilities.hpu_device import HPUDeviceUtils
+
+    HPUDeviceUtils.permute_params(model, True)
```
justusschock, line 87: can't we automate this with something like:

```python
if any(isinstance(x, _ConvNd) for x in model.modules())
```

and put this in the accelerator? Usually with PL we try to have code agnostic across hardware and this is likely something to be forgotten

if course, this approach won't work with users invoking functional convolution calls, but the `permute_params` likely won't catch this at all.

pytorch_lightning/utilities/hpu_device.py
```diff
@@ -0,0 +1,58 @@
...
+    # Therefore a re-ordering/permutation of all the convolution weights from KCRS to RSCK format is required
+    # before convolution operations
+    def permute_params(model: "pl.LightningModule", to_filters_last: bool) -> None:
+        """permute the params from filters first (KCRS) to filters last(RSCK) or vice versa.
+
+        and permute from RSCK to KCRS is used for checkpoint saving
+        """
+        with torch.no_grad():
+            for name, param in model.named_parameters():
+                if param.ndim == 4:
```
justusschock, line 37: what about other types of convolutions like 1D or 3D? This won't be catched at all and IMO this is relevant as well.
We cannot assume 2D operations only.

Also we should do some filtering to only apply this to weights of `_ConvNd`. If I have a custom model that for any reason has a 4D parameter that is never used for convolution, I would be pretty confused to have that one permuted like that. This is really error-prone and has to be regarded.

rohitgr7:
can you add some description in the PR or link the issue if there's any? :)

pytorch_lightning/utilities/hpu_device.py
jerome-habana: Agreed @justusschock . We will be removing this need in the subsequent releases. The idea here it to showcase one example to users who might want to work on existing versions of synapse AI

docs/source/accelerators/hpu.rst
jerome-habana: This is specifically for the filters. I'll see if I can add some more document for data formats too

kaushikb11:
@jerome-habana let's resolve the conflicts?

jerome-habana:
> @jerome-habana let's resolve the conflicts?

@kaushikb11  I'll close this PR. We are handling this internally now and expect no user script change

PR #12750, opened 2022-04-13
Improve LigtningEnum, etc.

### LINKED ISSUE #12749: Improve `LightningEnum` and its deprecation
## Proposed refactor

- Use the built-in `Enum` capabilities to make `LightningEnum` simpler.
- Modify and rename the `_OnAccessEnumMeta` type to only call a `deprecate()` method (it's not used for anything else.
- Create a `DeprecatedEnum` base class and use that instead of constantly setting `metaclass=...` on deprecated enums.
- Correct attribution for `_OnAccessEnumMeta` (current attribution is to a website and article that falsely claims authorship of `Enum`).

### Motivation

Fixing the attribution (which came to my attention due to [an `aenum` issue](https://github.com/ethanfurman/aenum/issues/18), and efficient use of `Enum`.

cc @justusschock @awaelchli @rohitgr7 @tchaton

ethanfurman:
Fixes #12749 

- change `from_str` to use `__members__` for enum names
- change `from_str` to use built-in `__getitem__` syntax for names
- make `_OnAccessEnumMeta` more robust and efficient
- create `DeprecatedEnum` as a base class
- change `DistributedType` and `DeviceType` to use `DeprecatedEnum`


Borda (approved):

ethanwharris (approved):

pytorch_lightning/utilities/enums.py
```diff
@@ -68,6 +68,12 @@ def __call__(cls, *args: Any, **kwargs: Any) -> Any:
         return obj
 
 
+class DeprecatedEnum(LightningEnum, metaclass=DeprecatedEnumMeta):
```
rohitgr7, line 71:
```suggestion
class DeprecatedEnum(LightningEnum, metaclass=DeprecatedEnumMeta):
```
let's keep this protected?

rohitgr7 (approved):

pytorch_lightning/utilities/enums.py
ethanfurman: I don't see a change -- what is your concern and suggestion?

rohitgr7: updated!

ethanfurman: I also added an `_` to `DeprecatedEnumMeta`.

awaelchli (approved):
Someone is playing with metaclasses! <3

carmocca (approved):

carmocca (approved):

Issue #12725, opened 2022-04-12
AttributeError: module 'torch.distributed' has no attribute 'ProcessGroup'


carmocca:
@carmocca This is the stack trace I get when I import PyTorch lightning with the following `environment.yml`

```yaml
name: pl_error
channels:
  - defaults
  - pytorch
  - conda-forge

dependencies:
  - python=3.9
  - numpy=1.21.2
  - pytorch=1.11
  - pytorch-lightning=1.6
```

Import error

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/__init__.py", line 30, in <module>
    from pytorch_lightning.callbacks import Callback  # noqa: E402
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/callbacks/__init__.py", line 26, in <module>
    from pytorch_lightning.callbacks.pruning import ModelPruning
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/callbacks/pruning.py", line 31, in <module>
    from pytorch_lightning.core.lightning import LightningModule
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/core/__init__.py", line 16, in <module>
    from pytorch_lightning.core.lightning import LightningModule
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/core/lightning.py", line 41, in <module>
    from pytorch_lightning.trainer.connectors.data_connector import _DataHookSelector
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/trainer/__init__.py", line 16, in <module>
    from pytorch_lightning.trainer.trainer import Trainer
  File "/opt/homebrew/Caskroom/miniforge/base/envs/pl_error/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 34, in <module>
    from pytorch_lightning.accelerators import Accelerator, GPUAccelerator, HPUAccelerator, IPUAccelerator, TPUAccelerator
[...]

PR #12794, opened 2022-04-18
Fix import error when `distributed` module not available

krshrimali:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/12725.

Confirmed that this fixes it (built PyTorch from source with `USE_DISTRIBUTED=0`)

No

## Before submitting

## PR review

## Did you have fun?

cc @borda


pytorch_lightning/strategies/ddp.py
```diff
@@ -63,7 +63,8 @@
 if _FAIRSCALE_AVAILABLE:
     from fairscale.optim import OSS
 if _TORCH_GREATER_EQUAL_1_10:
```
akihironitta, line 65: This is simpler maybe?
```suggestion
if _TORCH_GREATER_EQUAL_1_10 and torch.distributed.is_available():
```

akihironitta (approved):

pytorch_lightning/strategies/ddp.py
krshrimali: Ah, smart!!! @akihironitta - Thanks!

pytorch_lightning/strategies/ddp.py
```diff
@@ -62,8 +62,9 @@
 
 if _FAIRSCALE_AVAILABLE:
     from fairscale.optim import OSS
-if _TORCH_GREATER_EQUAL_1_10:
-    from torch.distributed.algorithms.model_averaging.averagers import ModelAverager
+if _TORCH_GREATER_EQUAL_1_10 and torch.distributed.is_available():
+    if torch.distributed.is_available():
+        from torch.distributed.algorithms.model_averaging.averagers import ModelAverager
```
rohitgr7, line 67:
```suggestion
if _TORCH_GREATER_EQUAL_1_10 and torch.distributed.is_available():
    from torch.distributed.algorithms.model_averaging.averagers import ModelAverager
```
??

rohitgr7 (approved):

Borda (approved):

pytorch_lightning/strategies/ddp.py
krshrimali: Good catch. Something went wrong with the previous suggestion. I didn't notice. Thanks!

pytorch_lightning/overrides/torch_distributed.py
```diff
@@ -15,7 +14,10 @@
 logger = logging.getLogger(__name__)
 
 if torch.distributed.is_available():
+    from torch._C._distributed_c10d import ProcessGroup
     from torch.distributed import Backend, broadcast, get_backend, get_rank, GroupMember
+else:
+    ProcessGroup = None
```
awaelchli, line 20: ProcessGroup only gets imported for typing. Wouldn't the type check fail if the else branch is considered, as None is not a valid type annotation.

awaelchli (approved):

pytorch_lightning/overrides/torch_distributed.py
krshrimali: I think `None` is a valid type annotation. Also, it doesn't error out on my system when `torch.distrubted.is_available()` is `False`. (source: https://stackoverflow.com/a/38516083)

```python
>>> def nothing(nun: None) -> None:
...     return nun
... 
>>> nothing(None)
>>> 
```

tchaton (approved):
LGTM !

pytorch_lightning/overrides/torch_distributed.py
akihironitta: I think `None` may be too strict in this case. Should it be `object` instead? @awaelchli @krshrimali 
```suggestion
else:
    ProcessGroup = object
```

krshrimali: Thanks, @akihironitta for taking another look. I was just reading about it on the internet and found this interesting answer: https://stackoverflow.com/a/36286947. Apparently, this fixes for my environment (which doesn't have `distributed` module in PyTorch). What do you think? (just pushed a commit with my proposed solution)

carmocca (approved):

Issue #11912, opened 2022-02-14
[RFC] Introduce `.fit(ckpt_path="last")`


carmocca:
## 🚀 Feature

Add support for passing just `"last"` to `trainer.{fit,validate,test,predict}(ckpt_path=...)`

This would pick the latest model checkpoint saved from the set of:
- "top-k" checkpoints
- "last" checkpoints
- "fault-tolerant" checkpoints.

The tracking logic would most likely be inside the checkpoint connector.

Also, just for `trainer.fit`, `ckpt_path` should default to `"last"` when fault-tolerance is enabled.

### Motivation

Until now, we've recommended passing `trainer.model_checkpoint.last_model_path`, but with fault-tolerance enabled, the last model path might be one generated by fault-tolerance.

Fault-tolerance logic is meant to be hidden from the users so we don't want them to track this.

### Alternatives

The best alternative (after #11862) would be: 
```python
ft_checkpoints = [cb for cb in self.callbacks if isinstance(cb, _FaultToleranceCheckpoint)]
if ft_checkpoints:
    last_ckpt_path = ft_checkpoints[0].ckpt_path
else:
    last_ckpt_path = trainer.model_checkpoint.last_model_path
trainer.fit(model, ckpt_path=last_ckpt_path)
```
but it does not consider that the `model_checkpoint.last_model_path` could have been saved after.

We should track this for the user.

### Additional context

Proposed by @tchaton 

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

[...]

Issue #5030, opened 2020-12-08
Add version suffix to last.ckpt


carmocca:
## 🚀 Feature

If you use `ModelCheckpoint(save_last=True)` and you run an experiment twice in the same directory, then this set of checkpoints is generated:

```
file-v0.ckpt
file-v1.ckpt
...
last.ckpt (the last of the second run)
```

the idea is to add a version also to `last.ckpt` if it would get overwritten:

```
file-v0.ckpt
file-v1.ckpt
...
last-v0.ckpt (last of the first run)
last-v1.ckpt (last of the second run)
```

### Motivation

Avoid overwriting existing checkpoints

### Alternatives

Modifying the default `CHECKPOINT_NAME_LAST` to avoid the conflict

### Additional context

Discussed in https://github.com/PyTorchLightning/pytorch-lightning/issues/5000#issuecomment-740701415

To be done after #5008 is merged.

cc @borda @carmocca @awaelchli @ninginthecloud @jjenniferdai @rohitgr7

carmocca:
@awaelchli answering https://github.com/PyTorchLightning/pytorch-lightning/issues/5000#issuecomment-741503135 here

How do you propose we fix this then?

awaelchli:
@carmocca Before I make a suggestion could you help me understand the context a bit more. Is this issue a) about damage control for users that accidentally write to the same directory, or b) is this about a particular use case where a user is doing this on purpose? 

I am asking this because of two reasons. 1) I notice in your description you didn't append epoch numbers to the filenames and the primary reason why the save_last option is exists is for sequential checkpoints numbered by some metric like epoch or step. 2) last.ckpt should always be a copy of the last epoch, i.e. epoch_N.ckpt is identical to last.ckpt. Therefore it would be save to overwrite last.ckpt in a subsequent run.

Note that situation a) cannot be solved by appending version affixes to the files. Consider the case where user changes epoch size between runs so they would get a mix of files and you wouldn't be able to tell if epoch=x_step=y.ckpt or epoch=x_step=z.ckpt belonged to the first or second run.

awaelchli:
One may consider this option:
```
file-v0.ckpt
file-v1.ckpt
...
last-v0.ckpt (last of the first run)
last-v1.ckpt (last of the second run)
last.ckpt (copy of last-v1)
```

carmocca:
It is about a)

>  I notice in your description you didn't append epoch numbers

Sorry, I didn't for brevity

[...]

PR #12816, opened 2022-04-20
Add support for reloading the last checkpoint saved by passing `ckpt_path="last"`

otaj:
## What does this PR do?

Implements #11912 

Allows you to pass keyword `"last"` to `ckpt_path` argument. `"best"` still remains the default value if `None` is passed, except one particular case - in method `fit` when fault tolerant training is enabled.

Value `last` selects the checkpoint with the latest modified timestamp from set of all last checkpoints (currently only one, because they overwrite each other, but once #5030 is solved, it could be a different one) and all fault tolerant checkpoints. If no checkpoint in these two sets is found (due to fault tolerant training not being enabled and passing `save_last=False` to `ModelCheckpoint`), warning is issued and no checkpoint is loaded.

#### What needs to be done before making it non-draft

- [x] Write tests
- [x] Add changelog entry
- [x] Update documentation
    - [ ] is the change of docs big enough?
- [x] Discuss whether this design makes sense

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7 @carmocca @justusschock @jjenniferdai


pytorch_lightning/trainer/trainer.py
```diff
@@ -1390,25 +1392,45 @@ def __set_ckpt_path(self, ckpt_path: Optional[str], model_provided: bool, model_
         if ft_checkpoints:
             ft_ckpt_path = ft_checkpoints[0].ckpt_path
             fs = get_filesystem(ft_ckpt_path)
-            if fs.exists(ft_ckpt_path):
-                return ft_ckpt_path
+            if not fs.exists(ft_ckpt_path):
+                ft_ckpt_path = None
+        else:
+            ft_ckpt_path = None
```
awaelchli, line 1398: this can all be simplified by setting `ft_ckpt_path = None` from the beginning

pytorch_lightning/trainer/trainer.py
```diff
@@ -1433,6 +1455,21 @@ def __set_ckpt_path(self, ckpt_path: Optional[str], model_provided: bool, model_
             # load best weights
             ckpt_path = self.checkpoint_callback.best_model_path
 
+        if ckpt_path == "last":
+            candidates = [ft.ckpt_path for ft in ft_checkpoints] + [
+                cb.last_model_path for cb in self.checkpoint_callbacks
+            ]
+            candidates_fs = {path: get_filesystem(path) for path in candidates if path}
+            candidates_ts = {path: fs.modified(path) for path, fs in candidates_fs.items() if fs.exists(path)}
```
awaelchli, line 1463: I believe we should only take the last_model_path from the main checkpoint callback (the first one).

pytorch_lightning/trainer/trainer.py
```diff
@@ -1390,25 +1392,45 @@ def __set_ckpt_path(self, ckpt_path: Optional[str], model_provided: bool, model_
...
 
         if model_provided and ckpt_path is None:
             # use passed model to function without loading weights
             return
 
-        fn = self.state.fn.value
-
         if model_connected and ckpt_path is None:
-            rank_zero_warn(
+            full_msg = (
```
awaelchli, line 1417: can we clean this up a little? either we use f-string formatting only, or we do the string.format but mixing both is a bit harder to read and I believe it's not necessary.

tests/trainer/test_trainer.py
```diff
@@ -664,6 +665,118 @@ def test_benchmark_option(benchmark_, deterministic, expected):
     torch.backends.cudnn.benchmark = original_val
 
 
+@pytest.mark.parametrize("ckpt_path", (None, "last"))
+@pytest.mark.parametrize("save_last", (True, False))
+@pytest.mark.parametrize("fn", ("fit", "validate"))
+@pytest.mark.parametrize("ft_enabled", (True, False))
+def test_checkpoint_path_input_last(tmpdir, ckpt_path, save_last, fn, ft_enabled):
```
awaelchli, line 672: there is a bit much going on in this test. The highly nested if-else blocks make the test hard to follow and make the parametrizations almost pointless. 
I would at least disentangle the the fault tolerance from this test.

awaelchli (changes requested):

pytorch_lightning/trainer/trainer.py
otaj: Done

pytorch_lightning/trainer/trainer.py
otaj: Done

tests/trainer/test_trainer.py
otaj: Cleaned up - now we have separate test for fault tolerancy enabled

pytorch_lightning/trainer/trainer.py
otaj: We were talking with @carmocca about this and the conclusion was that at this moment, it does not really matter as they all point to the same file and in the future, once issue #5030 is done, then it will actually make sense the load the latest available from multiple runs. Does it make sense to you as well?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Allow to load last checkpoint using `ckpt_path="last"` ([#12816](https://github.com/PyTorchLightning/pytorch-lightning/pull/12816))
```
carmocca, line 12:
```suggestion
- Added support for reloading the last checkpoint saved by passing `ckpt_path="last"` ([#12816](https://github.com/PyTorchLightning/pytorch-lightning/pull/12816))
```

pytorch_lightning/trainer/trainer.py
carmocca: Yes. Also note that when multiple checkpoint instances are used, not all might have `save_last=True` set

carmocca (approved):
LGTM! I've pushed some minor simplification to the code and greater simplifications to the test to narrow their scope to just these warnings.

tests/trainer/test_trainer.py
```diff
@@ -666,71 +667,29 @@ def test_benchmark_option(benchmark_, deterministic, expected):
...
-    )
-    assert trainer.ckpt_path is None
-    trainer_fn = getattr(trainer, fn)
-
-    from pytorch_lightning.callbacks.fault_tolerance import _FaultToleranceCheckpoint
-
-    ft_checkpoints = [cb for cb in trainer.callbacks if isinstance(cb, _FaultToleranceCheckpoint)]
-    ft_ckpt_path = ft_checkpoints[0].ckpt_path
+@pytest.mark.parametrize("fn", (TrainerFn.FITTING, TrainerFn.VALIDATING))
+def test_checkpoint_path_input_last_fault_tolerant(ckpt_path, fn):
```
otaj, line 671: This rewritten test now fails. The reason for that is that loading of last checkpoint checks whether the file exists and the fault tolerant checkpoint does not exist

otaj: I fixed it with a simple addition

rohitgr7 (approved):
great!

kaushikb11 (approved):

PR #12821, opened 2022-04-20
Fix fit loop restart logic to enable resume using the checkpoint

### LINKED ISSUE #12724: PT 1.6.0 could not resume a training with plugins monitoring on metrics
## 🐛 Bug

As mentioned in the title

### To Reproduce

The code (boring.py) is provided below:

<details>

```
from typing import List

import torch
from pytorch_lightning import LightningDataModule, LightningModule
from pytorch_lightning.utilities.cli import (LightningArgumentParser, LightningCLI)
from jsonargparse import lazy_instance


class Arch(torch.nn.Linear):

    def __init__(self, input_size: int = 10, output_size: int = 2) -> None:
        super().__init__(input_size, output_size)


class RandomDataset(Dataset):

        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class MyDataModule(LightningDataModule):

    def __init__(self, input_size: int = 10, train_transforms=None, val_transforms=None, test_transforms=None, dims=None):
        super().__init__(train_transforms=train_transforms, val_transforms=val_transforms, test_transforms=test_transforms, dims=dims)
        self.input_size = input_size

    def train_dataloader(self) -> DataLoader:
        return DataLoader(RandomDataset(self.input_size, 64), batch_size=2)

    def val_dataloader(self) -> DataLoader:
        return DataLoader(RandomDataset(self.input_size, 64), batch_size=2)

    def test_dataloader(self) -> DataLoader:
        return DataLoader(RandomDataset(self.input_size, 64), batch_size=2)


    def __init__(self, arch: Arch = lazy_instance(Arch), channels: List[int] = [0, 1]):
        super().__init__()
        self.arch = arch

    def forward(self, x):
        print(x.shape)
        return self.arch(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

        return torch.optim.SGD(self.arch.parameters(), lr=0.1)


class MyCLI(LightningCLI):

    def add_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
        from pytorch_lightning.callbacks import EarlyStopping, ModelCheckpoint
        parser.add_lightning_class_args(EarlyStopping, "early_stopping")
        parser.set_defaults({
            "early_stopping.monitor": "valid_loss",
            "early_stopping.min_delta": 0.01,
            "early_stopping.patience": 10,
            "early_stopping.mode": "min",
        })
        
        # ModelCheckpoint
        parser.add_lightning_class_args(ModelCheckpoint, "model_checkpoint")
        model_checkpoint_defaults = {
            "model_checkpoint.filename": "epoch{epoch}_valid_loss{valid_loss:.4f}",
            "model_checkpoint.monitor": "valid_loss",
            "model_checkpoint.mode": "min",
            "model_checkpoint.every_n_epochs": 1,
            "model_checkpoint.save_top_k": 5,
            "model_checkpoint.auto_insert_metric_name": False,
            "model_checkpoint.save_last": True
        }
        parser.set_defaults(model_checkpoint_defaults)
        return super().add_arguments_to_parser(parser)


if __name__ == '__main__':
    cli = MyCLI(BoringModel, MyDataModule, seed_everything_default=None, save_config_overwrite=True, parser_kwargs={"parser_mode": "omegaconf"})
```

</details>

1. fit: `python boring.py fit`, and kill the training program when we have ckeckpoints  
2. resume (something like the command): `python boring.py fit --config lightning_logs/version_7/config.yaml --ckpt_path=lightning_logs/version_7/checkpoints/last.ckpt`

### Expected behavior

resume sucessfully.

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.6.0
- OS (e.g., Linux):
- CUDA/cuDNN version:

### Additional context

The error: 

<details>

```
python boring.py fit --config lightning_logs/version_7/config.yaml --ckpt_path=lightning_logs/version_7/checkpoints/last.ckpt
/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/loops/utilities.py:91: PossibleUserWarning: `max_epochs` was not set. Setting it to 1000 epochs. To train without an epoch limit, set `max_epochs=-1`.
  rank_zero_warn(
GPU available: True, used: False
HPU available: False, using: 0 HPUs
/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py:1823: PossibleUserWarning: GPU available but not used. Set `accelerator` and `devices` using `Trainer(accelerator='gpu', devices=8)`.
  rank_zero_warn(
Restoring states from the checkpoint path at lightning_logs/version_7/checkpoints/last.ckpt
/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/callbacks/model_checkpoint.py:342: UserWarning: The dirpath has changed from '/data/home/quancs/projects/NBSS_pmt/lightning_logs/version_7/checkpoints' to '/data/home/quancs/projects/NBSS_pmt/lightning_logs/version_8/checkpoints', therefore `best_model_score`, `kth_best_model_path`, `kth_value`, `last_model_path` and `best_k_models` won't be reloaded. Only `best_model_path` will be reloaded.
  warnings.warn(

  | Name | Type | Params
------------------------------
0 | arch | Arch | 22    
------------------------------
22        Trainable params
0         Non-trainable params
22        Total params
0.000     Total estimated model params size (MB)
Restored all states from the checkpoint file at lightning_logs/version_7/checkpoints/last.ckpt
/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/connectors/data_connector.py:240: PossibleUserWarning: The dataloader, train_dataloader, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 128 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  rank_zero_warn(
/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py:1938: PossibleUserWarning: The number of training samples (32) is smaller than the logging interval Trainer(log_every_n_steps=50). Set a lower value for log_every_n_steps if you want to see logs for the training epoch.
  rank_zero_warn(
/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/connectors/data_connector.py:240: PossibleUserWarning: The dataloader, val_dataloader 0, does not have many workers which may be a bottleneck. Consider increasing the value of the `num_workers` argument` (try 128 which is the number of cpus on this machine) in the `DataLoader` init to improve performance.
  rank_zero_warn(
Epoch 19:  50%|██████████████████████████████████████████████████████████████████████████▌                                                                          | 32/64 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/data/home/x/projects/NBSS_pmt/boring.py", line 100, in <module>
    cli = MyCLI(BoringModel, MyDataModule, seed_everything_default=None, save_config_overwrite=True, parser_kwargs={"parser_mode": "omegaconf"})
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/utilities/cli.py", line 564, in __init__
    self._run_subcommand(self.subcommand)
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/utilities/cli.py", line 835, in _run_subcommand
    fn(**fn_kwargs)
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 771, in fit
    self._call_and_handle_interrupt(
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 724, in _call_and_handle_interrupt
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 812, in _fit_impl
    results = self._run(model, ckpt_path=self.ckpt_path)
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1237, in _run
    results = self._run_stage()
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1324, in _run_stage
    return self._run_train()
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1354, in _run_train
    self.fit_loop.run()
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/loops/base.py", line 205, in run
    self.on_advance_end()
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/loops/fit_loop.py", line 297, in on_advance_end
    self.trainer._call_callback_hooks("on_train_epoch_end")
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1637, in _call_callback_hooks
    fn(self, self.lightning_module, *args, **kwargs)
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/callbacks/early_stopping.py", line 179, in on_train_epoch_end
    self._run_early_stopping_check(trainer)
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/callbacks/early_stopping.py", line 190, in _run_early_stopping_check
    if trainer.fast_dev_run or not self._validate_condition_metric(  # disable early_stopping with fast_dev_run
  File "/data/home/x/miniconda3/lib/python3.9/site-packages/pytorch_lightning/callbacks/early_stopping.py", line 145, in _validate_condition_metric
    raise RuntimeError(error_msg)
RuntimeError: Early stopping conditioned on metric `valid_loss` which is not available. Pass in or modify your `EarlyStopping` callback to use any of the following: ``
Epoch 19:  50%|█████     | 32/64 [00:00<?, ?it/s]
```

</details>

cc @tchaton @rohitgr7 @carmocca @awaelchli

rohitgr7:
## What does this PR do?

Fixes #12724

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7 @carmocca @justusschock @ananthsub @ninginthecloud


pytorch_lightning/loops/fit_loop.py
```diff
@@ -128,10 +128,10 @@ def restarting(self, restarting: bool) -> None:
         values = (
             self.epoch_progress.current.ready,
             self.epoch_progress.current.started,
-            self.epoch_progress.current.processed,
         )
-        finished_before_on_train_end = any(v != self.epoch_progress.current.completed for v in values)
+        finished_before_on_train_end = any(v != self.epoch_progress.current.processed for v in values)
```
rohitgr7, line 132: in an ideal case when a checkpoint is saved using `ModelCheckpoint`, they are saved inside `on_train_epoch_end`, and `process.current.completed` is not incremented by that time, so during restart, `restarting` will be True and it ends up calling https://github.com/PyTorchLightning/pytorch-lightning/blob/bb81802bff29fa06399d704688df256813e05ad2/pytorch_lightning/loops/fit_loop.py#L271

for the last completed epoch again and then start the next epoch. Due to these certain hooks get called which are not supposed to be called.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -205,6 +205,10 @@ def reset(self) -> None:
 
     def on_run_start(self) -> None:  # type: ignore[override]
         """Calls the ``on_train_start`` hook."""
+        # update the current_epoch in-case of checkpoint reload
+        if not self._iteration_based_training():
```
rohitgr7, line 209: during restart if it's not iteration-based training, we need to update the current epoch so that it starts from a fresh epoch rather than the old one.

rohitgr7:
failing ci is on master and unrelated to this PR.

carmocca (approved):

pytorch_lightning/loops/fit_loop.py
```diff
@@ -128,10 +128,10 @@ def restarting(self, restarting: bool) -> None:
         values = (
             self.epoch_progress.current.ready,
             self.epoch_progress.current.started,
-            self.epoch_progress.current.processed,
         )
-        finished_before_on_train_end = any(v != self.epoch_progress.current.completed for v in values)
+        finished_before_on_train_end = any(v != self.epoch_progress.current.processed for v in values)
         restarting &= finished_before_on_train_end
+        restarting |= self._iteration_based_training()
```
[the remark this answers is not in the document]
rohitgr7: not 100% sure but in the case of iteration-based training, set `restarting=True` when total batches processed != current, ready. Need to check how it works with step-based training.

pytorch_lightning/loops/fit_loop.py
rohitgr7: yes! good catch! will update

rohitgr7: > Does this make the following 2 lines unnecessary?

maybe... will check. I remember a test was failing: https://github.com/PyTorchLightning/pytorch-lightning/runs/6101153350?check_suite_focus=true

but it should not I guess since we increment `current.completed` after every epoch.

rohitgr7: okay looks like `done` is called within `skip`, so we need to keep it.

tests/models/test_hooks.py
```diff
@@ -651,23 +650,12 @@ def test_trainer_model_hook_system_fit_no_val_and_resume_max_epochs(tmpdir):
...
-        dict(name="Callback.state_dict"),
-        dict(name="Callback.on_save_checkpoint", args=(trainer, model, saved_ckpt1)),
-        dict(name="on_save_checkpoint", args=(saved_ckpt1,)),
-        dict(name="on_train_epoch_end"),
-        dict(name="Callback.on_epoch_end", args=(trainer, model)),
-        dict(name="on_epoch_end"),
-        dict(name="Callback.on_epoch_start", args=(trainer, model)),
-        dict(name="on_epoch_start"),
-        dict(name="Callback.on_train_epoch_start", args=(trainer, model)),
-        dict(name="on_train_epoch_start"),
```
Borda, line 664: a bit confused why this is not called... :/

Borda (approved):

tests/models/test_hooks.py
rohitgr7:
```
        dict(name="on_train_epoch_start"),
        dict(name="Callback.on_train_epoch_end", args=(trainer, model)),
```
check the call order. What was happening is that right after epoch_start, it didn't process any batches and went directly to `epoch_end` hooks, which was incorrect.

otaj (approved):

carmocca (approved):

pytorch_lightning/loops/fit_loop.py
awaelchli: Hey @rohitgr7 
I don't understand this logic. There is no true iteration based training in Lightning and we always have epochs. We may restart from a completed epoch or from an incomplete epoch, regardless of how the max_ flags on the trainer is set.

pytorch_lightning/loops/fit_loop.py
```diff
@@ -205,6 +200,10 @@ def reset(self) -> None:
 
     def on_run_start(self) -> None:  # type: ignore[override]
         """Calls the ``on_train_start`` hook."""
+        # update the current_epoch in-case of checkpoint reload
```
awaelchli, line 203: in case

pytorch_lightning/loops/fit_loop.py
rohitgr7: if a user is using `max_steps=7` and uses a checkpoint from `step=4` to restart, we need to start from step 5 in that case. Although I guess the dataloaders are re-iterated from the beginning in that case.

awaelchli: This explanation is not satisfying. It doesn't answer how max_steps has anything to do with the way we restart. 

You could also have `max_epochs=1` where the epoch size is 7 and you would still restore the checkpoint on step 4 the exact same way. 

max_steps / max_epochs are the stopping conditions. Them affecting the way we restart is beyond my understanding.

rohitgr7: > You could also have max_epochs=1 where the epoch size is 7 (equivalent to max_steps=7) and you would still restore the checkpoint on step 4 the exact same way.

yes, it does but in that case we do start from step=4 but on an entirely new epoch. The reason we need to separate this a little is that we don't need to update the current epoch. Although just noticed that we do update the current_epoch at the end even if the user is performing training based on `max_steps`. Do you think we should remove this condition and keep incrementing the current epoch on each restart even though the training is based on `max_steps`?

awaelchli: Whatever the fix is, it will not at all be conditioned on max_steps / max_epochs trainer flags. I don't know how to solve this problem and we need to brainstorm this what should be fixed. 

All I know right now is that this PR did something weird 😅 that can't be the fix.

Issue #12804, opened 2022-04-19
Automatic seeding when using the LightningCLI


Schinkikami:
## 🚀 Feature

Currently the LightningCLI offers the argument `seed_everything_default: Optional[int] = None`.

As it is good practice in scientific research to provide reproducible experiments, I propose to have an option to automatically set a randomly chosen seed at runtime. This then has the advantage that every experiment with the CLI is automatically reproducible as the seed is logged in the config.yaml via the `SafeConfigCallback`.

### Pitch

I propose to extend the signature to
`seed_everything_default: Optional[Union[bool,int]] = False`, with the Optional being kept for backwards compatibility. Here the argument None is then equivalent to False.

In this case, setting seed_everything_default to `True`, automatically generate's a seed that is logged to the argument `seed_everything`, which like before is then saved to  the config.yaml by the `SafeConfigCallback`.

The code change could be as following:

```

    def add_default_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
        """Adds default arguments to the parser."""
        parser.add_argument(
            "--seed_everything",
            type=Optional[int],
            default=self.seed_everything_default,
            help="Set to an int to run seed_everything with this value before classes instantiation",
        )
```
to 

```
    def add_default_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
        """Adds default arguments to the parser."""

        default_seed = self.seed_everything_default

        # If self.seed_everything_default is a bool, choose a random seed or none at all
        if isinstance(self.seed_everything_default, bool) and self.seed_everything_default:
            default_seed = random.randint(int(torch.iinfo(torch.int32).min), int(torch.iinfo(torch.int32).max))
        else:
            default_seed = None

        parser.add_argument(
            "--seed_everything",
            type=Optional[int],
            default=default_seed,
            help="Set to an int to run seed_everything with this value before classes instantiation. Set to True to automatically choose a seed at runtime",
        )


```
This change also has the advantage that it prevents passing an int greater than pytorchs `int32.max`, which results, without raising a warning, in using a different random seed (same behaviour for numpy).

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

[...]

PR #12822, opened 2022-04-20
Feature/12804 Automatic Seeding of the CLI

Schinkikami:
## What does this PR do?

The argument `seed_default_value` of the `LightningCLI` is changed to the type `Optional[Union[bool,int]] = True` (from `Optional[int] = None`). The `Optional` flag is set for backwards compatibility. Setting it to `None` has the same effect as setting it to `False`, but raises a `rankzero - LightningDeprecationWarning`.

If the value is set to `True` a valid (range `np.iiinfo(np.int32).min` to `np.iiinfo(np.int32).max` ) seed is drawn at random and added as default to the `seed_everything` argument. 

The issue was discussed and approved in [12804](https://github.com/PyTorchLightning/pytorch-lightning/issues/12804). The type of `--seed_everything` is set to `Optional[int32]`, as I do not think, that setting it to `Optional[Union[bool,int]]` is required. By default the CLI now chooses a new seed with every instantiation, which can of course be overwritten with `--seed_everything` from the CLI or a config file. The only time, one would like to set it to `False`, would be if somewhere in the code the seed is already set, but in this case the automatic seeding should be disabled in the code (by setting `seed_default_value` to False) and not in a config file with `--seed_everything`.

Choosing a seed outside of the valid seed range, raises and error and avoids unnoticed accidental mistakes.

The version in the deprecation warning still needs to be set.

Fixes #\<[12804](https://github.com/PyTorchLightning/pytorch-lightning/issues/12804)>
For a motivation and discussion, have a look at the issue. 

No.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Marked `swa_lrs` argument in `StochasticWeightAveraging` callback as required ([#12556](https://github.com/PyTorchLightning/pytorch-lightning/pull/12556))
 
 
--
+- Changed `seed_default_value` argument in the `LightningCLI` to type `Union[bool, int]`. If set to `True` a seed is automatically generated for the parser argument `--seed_everything`. Setting it to `None` raises a `LightningDeprecationWarning` ([#12804](https://github.com/PyTorchLightning/pytorch-lightning/issues/12804)). 
 
 
 -
--- a/pytorch_lightning/utilities/cli.py
+++ b/pytorch_lightning/utilities/cli.py
@@ -15,6 +15,7 @@
 
 import inspect
 import os
+import random
 import sys
 from functools import partial, update_wrapper
 from types import MethodType, ModuleType
@@ -25,6 +26,8 @@ import torch
 import yaml
 from torch.optim import Optimizer
 
+import numpy as np
+
 import pytorch_lightning as pl
 from pytorch_lightning import Callback, LightningDataModule, LightningModule, seed_everything, Trainer
 from pytorch_lightning.utilities.cloud_io import get_filesystem
@@ -32,13 +35,13 @@ from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.imports import _JSONARGPARSE_AVAILABLE
 from pytorch_lightning.utilities.meta import get_all_subclasses
 from pytorch_lightning.utilities.model_helpers import is_overridden
-from pytorch_lightning.utilities.rank_zero import _warn, rank_zero_warn
+from pytorch_lightning.utilities.rank_zero import _warn, rank_zero_warn, rank_zero_deprecation
 from pytorch_lightning.utilities.types import LRSchedulerType, LRSchedulerTypeTuple, LRSchedulerTypeUnion
 
 if _JSONARGPARSE_AVAILABLE:
     from jsonargparse import ActionConfigFile, ArgumentParser, class_from_function, Namespace, set_config_read_mode
     from jsonargparse.optionals import import_docstring_parse
-
+    from jsonargparse.typing import restricted_number_type
     set_config_read_mode(fsspec_enabled=True)
 else:
     locals()["ArgumentParser"] = object
@@ -470,7 +473,7 @@ class LightningCLI:
         save_config_multifile: bool = False,
         trainer_class: Union[Type[Trainer], Callable[..., Trainer]] = Trainer,
         trainer_defaults: Optional[Dict[str, Any]] = None,
-        seed_everything_default: Optional[int] = None,
+        seed_everything_default: Optional[Union[bool, int]] = True,
         description: str = "pytorch-lightning trainer command line tool",
         env_prefix: str = "PL",
         env_parse: bool = False,
@@ -508,8 +511,8 @@ class LightningCLI:
                 this argument will not be configurable from a configuration file and will always be present for
                 this particular CLI. Alternatively, configurable callbacks can be added as explained in
                 :ref:`the CLI docs <lightning-cli>`.
-            seed_everything_default: Default value for the :func:`~pytorch_lightning.utilities.seed.seed_everything`
-                seed argument.
+            seed_everything_default: Value for the :func:`~pytorch_lightning.utilities.seed.seed_everything`
+                seed argument. Set to True to automatically choose a valid seed.
             description: Description of the tool shown when running ``--help``.
             env_prefix: Prefix for environment variables.
             env_parse: Whether environment variable parsing is enabled.
@@ -532,6 +535,14 @@ class LightningCLI:
         self.trainer_defaults = trainer_defaults or {}
         self.seed_everything_default = seed_everything_default
 
+        if self.seed_everything_default is None:
+            rank_zero_deprecation(
+                "Setting `LightningCLI.seed_everything_default` to `None` is deprecated in v1.X and will be"
+                " removed in v1.(X+1). Set it to `False` instead."
+            )
+            self.seed_everything_default = False
+
+
         self.model_class = model_class
         # used to differentiate between the original value and the processed value
         self._model_class = model_class or LightningModule
@@ -553,9 +564,10 @@ class LightningCLI:
 
         self.subcommand = self.config["subcommand"] if run else None
 
-        seed = self._get(self.config, "seed_everything")
-        if seed is not None:
-            seed_everything(seed, workers=True)
+
+        self._set_seed()
+
+
 
         self.before_instantiate_classes()
         self.instantiate_classes()
@@ -591,11 +603,24 @@ class LightningCLI:
 
     def add_default_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
         """Adds default arguments to the parser."""
+
+        min_value = np.iinfo(np.int32).min
+        max_value = np.iinfo(np.int32).max
+        int32 = restricted_number_type('int32', int, [('>=', min_value), ('<=', max_value)])
+
+        if isinstance(self.seed_everything_default, bool):
+            if self.seed_everything_default:
+                seed = random.randint(min_value, max_value)
+            else:
+                seed = None
+        else:
+            seed = self.seed_everything_default
+
         parser.add_argument(
             "--seed_everything",
-            type=Optional[int],
-            default=self.seed_everything_default,
-            help="Set to an int to run seed_everything with this value before classes instantiation",
+            type=Optional[int32],
+            default=seed,
+            help="Set to an int to run seed_everything with this value before classes instantiation."
         )
 
     def add_core_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
@@ -848,6 +873,14 @@ class LightningCLI:
         if self.datamodule is not None:
             fn_kwargs["datamodule"] = self.datamodule
         return fn_kwargs
+
+    def _set_seed(self):
+        """Sets the seed."""
+
+        config_seed = self._get(self.config, "seed_everything")
+
+        if config_seed is not None:
+            seed_everything(config_seed, workers=True)
 
 
 def _class_path_from_class(class_type: Type) -> str:
--- a/pytorch_lightning/utilities/cli.py
+++ b/pytorch_lightning/utilities/cli.py
@@ -512,7 +512,7 @@ class LightningCLI:
                 this particular CLI. Alternatively, configurable callbacks can be added as explained in
                 :ref:`the CLI docs <lightning-cli>`.
             seed_everything_default: Value for the :func:`~pytorch_lightning.utilities.seed.seed_everything`
-                seed argument. Set to True to automatically choose a valid seed.
+                seed argument. Set to True to automatically choose a valid seed. Set to False if you want to set your seeds manually.
             description: Description of the tool shown when running ``--help``.
             env_prefix: Prefix for environment variables.
             env_parse: Whether environment variable parsing is enabled.
--- a/tests/utilities/test_cli.py
+++ b/tests/utilities/test_cli.py
@@ -24,6 +24,7 @@ from typing import List, Optional, Union
 from unittest import mock
 from unittest.mock import ANY
 
+import numpy as np
 import pytest
 import torch
 import yaml
@@ -52,6 +53,7 @@ from pytorch_lightning.utilities.cli import (
 )
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.imports import _TORCHVISION_AVAILABLE
+from pytorch_lightning.utilities.rank_zero import LightningDeprecationWarning
 from tests.helpers import BoringDataModule, BoringModel
 from tests.helpers.runif import RunIf
 from tests.helpers.utils import no_warning_call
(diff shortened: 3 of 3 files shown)


pytorch_lightning/utilities/cli.py
```diff
@@ -532,6 +536,13 @@ def __init__(
         self.trainer_defaults = trainer_defaults or {}
         self.seed_everything_default = seed_everything_default
 
+        if self.seed_everything_default is None:
+            rank_zero_deprecation(
+                "Setting `LightningCLI.seed_everything_default` to `None` is deprecated in v1.X and will be"
+                " removed in v1.(X+1). Set it to `False` instead."
```
carmocca, line 542:
```suggestion
                "Setting `LightningCLI.seed_everything_default` to `None` is deprecated in v1.7 and will be"
                " removed in v1.9. Set it to `False` instead."
```

CHANGELOG.md
```diff
@@ -40,7 +40,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Marked `swa_lrs` argument in `StochasticWeightAveraging` callback as required ([#12556](https://github.com/PyTorchLightning/pytorch-lightning/pull/12556))
 
 
--
+- Changed `seed_default_value` argument in the `LightningCLI` to type `Union[bool, int]`. If set to `True` a seed is automatically generated for the parser argument `--seed_everything`. Setting it to `None` raises a `LightningDeprecationWarning` ([#12804](https://github.com/PyTorchLightning/pytorch-lightning/issues/12804)).
```
carmocca, line 43: The comment about the deprecation should go into the "Deprecated" section below. One PR can add multiple CHANGELOG entries

pytorch_lightning/utilities/cli.py
```diff
@@ -470,7 +473,7 @@ def __init__(
         save_config_multifile: bool = False,
         trainer_class: Union[Type[Trainer], Callable[..., Trainer]] = Trainer,
         trainer_defaults: Optional[Dict[str, Any]] = None,
-        seed_everything_default: Optional[int] = None,
+        seed_everything_default: Optional[Union[bool, int]] = True,
```
carmocca, line 476: @mauvilsa would it be useful to also annotate this with `int32`?

tests/utilities/test_cli.py
```diff
@@ -1547,3 +1549,54 @@ def test_cli_logger_shorthand():
     with mock.patch("sys.argv", ["any.py", "--trainer.logger=False"]):
         cli = LightningCLI(TestModel, run=False)
     assert cli.trainer.logger is None
+
+
+def test_cli_auto_seeding():
+
+    with mock.patch("sys.argv", ["any.py"]):
+        from jsonargparse import ParserError
```
carmocca, line 1557: Import this only once at the top of the test

tests/utilities/test_cli.py
```diff
@@ -1547,3 +1549,54 @@ def test_cli_logger_shorthand():
...
+                TestModel,
+                run=False,
+                seed_everything_default=int(np.iinfo(np.int32).min - 1),
+                trainer_defaults={"logger": False},
+                parser_kwargs={"error_handler": None},
+            )
+            assert cli.seed_everything_default is False
+
+    with mock.patch("sys.argv", ["any.py"]):
+        with pytest.warns(LightningDeprecationWarning):
```
carmocca, line 1583: The deprecation warning tests should go here https://github.com/PyTorchLightning/pytorch-lightning/blob/master/tests/deprecated_api/test_remove_1-9.py

pytorch_lightning/utilities/cli.py
```diff
@@ -508,8 +511,9 @@ def __init__(
                 this argument will not be configurable from a configuration file and will always be present for
                 this particular CLI. Alternatively, configurable callbacks can be added as explained in
                 :ref:`the CLI docs <lightning-cli>`.
-            seed_everything_default: Default value for the :func:`~pytorch_lightning.utilities.seed.seed_everything`
-                seed argument.
+            seed_everything_default: Value for the :func:`~pytorch_lightning.utilities.seed.seed_everything`
+                seed argument. Set to True to automatically choose a valid seed.
+                Set to False if you want to set your seeds manually.
```
mauvilsa, line 516: `int32` is to set the seed manually. `False` is to not call `seed_everything`.

pytorch_lightning/utilities/cli.py
```diff
@@ -591,11 +600,24 @@ def setup_parser(
 
     def add_default_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
         """Adds default arguments to the parser."""
+
+        min_value = np.iinfo(np.int32).min
+        max_value = np.iinfo(np.int32).max
+        int32 = restricted_number_type("int32", int, [(">=", min_value), ("<=", max_value)])
```
mauvilsa, line 606: Don't define a type inside a method. This should be at the top of this file or in [pytorch_lightning/utilities/types.py](https://github.com/PyTorchLightning/pytorch-lightning/tree/master/pytorch_lightning/utilities/types.py). It needs to be inside an `if _JSONARGPARSE_AVAILABLE:` block since jsonargparse is an optional dependency.

After a second thought, we could do something even better. Instead of creating a new type, we use `np.int32`. But this type needs to be registered so that the parsing fails on invalid values.

```python
from jsonargparse.typing import register_type

int32_min = np.iinfo(np.int32).min
int32_max = np.iinfo(np.int32).max

def int32_deserializer(value):
    value = int(value)
    if value < int32_min or value > int32_max:
        raise ValueError(f"value outside int32 range: {value}")
    return value

register_type(np.int32, deserializer=int32_deserializer)
```

pytorch_lightning/utilities/cli.py
```diff
@@ -591,11 +600,24 @@ def setup_parser(
...
+
+        if isinstance(self.seed_everything_default, bool):
+            if self.seed_everything_default:
+                seed = random.randint(min_value, max_value)
+            else:
+                seed = None
+        else:
+            seed = self.seed_everything_default
```
mauvilsa, line 614: To me this creates a weird behavior. When I do `cli.py --help` what I expect to see as default value is `True` and not a number that changes each time it is run. Similarly when I do `--print_config` the output should have the default value `True`.

Only in the case that the code is run, the saved config would include the used seed. As commented in the issue, the value should be overwritten in the config after parsing. Here the default as it was before.

pytorch_lightning/utilities/cli.py
```diff
@@ -591,11 +600,24 @@ def setup_parser(
...
+
         parser.add_argument(
             "--seed_everything",
-            type=Optional[int],
-            default=self.seed_everything_default,
-            help="Set to an int to run seed_everything with this value before classes instantiation",
+            type=Optional[int32],
```
mauvilsa, line 618: The type should be `Optional[Union[int32, bool]]`.

pytorch_lightning/utilities/cli.py
mauvilsa: Yes, here it should also be `int32`.

mauvilsa (changes requested):

tests/utilities/test_cli.py
Schinkikami: Debugging fragment, sry.

pytorch_lightning/utilities/cli.py
Schinkikami: The first paragraph is clear. I'm not sure if I understand the second paragraph correctly, as I'm not so familiar with the jsonargparse internals. 

Instead of defining the new type int32 we use the existing type np.int32. The deserializer converts the CLI/config/default value to the internal representation and throws an error if the value is out of range. What irritates me is that  there isn't a cast to np.int32... So we are just using the class_name np.int32 but actually we are just using ints?

So if I understand this correctly, we will then set the type of `seed_everything_default` to `Optional[Union[np.int32, bool]]` instead of `Optional[Union[int/int32, bool]]` and the type of `--seed_everything` to `Union[bool, np.int32] `instead of` Union[bool, int/int32]` ? But in this case the linter will throw warnings if the user passes an int as argument to seed_everything_default.... 

For now I set it to uint32.

pytorch_lightning/utilities/cli.py
```diff
@@ -849,6 +862,19 @@ def _prepare_subcommand_kwargs(self, subcommand: str) -> Dict[str, Any]:
...
         return fn_kwargs
 
+    def _set_seed(self) -> None:
+        """Sets the seed."""
+
+        config_seed = self._get(self.config, "seed_everything")
+
+        if isinstance(config_seed, bool):
+            if config_seed:
+                config_seed = _select_seed_randomly()
```
carmocca, line 872:
```suggestion
        if isinstance(config_seed, bool) and config_seed:
            config_seed = _select_seed_randomly()
```

pytorch_lightning/utilities/cli.py
```diff
@@ -849,6 +862,19 @@ def _prepare_subcommand_kwargs(self, subcommand: str) -> Dict[str, Any]:
...
+
+        if config_seed is not False:
+            seed_everything(config_seed, workers=True)
+            self.config["seed_everything"] = config_seed
```
carmocca, line 876: This key might not exist if the user overrides `add_default_arguments_to_parser` and does not add the argument

pytorch_lightning/utilities/cli.py
```diff
@@ -470,7 +476,7 @@ def __init__(
         save_config_multifile: bool = False,
         trainer_class: Union[Type[Trainer], Callable[..., Trainer]] = Trainer,
         trainer_defaults: Optional[Dict[str, Any]] = None,
-        seed_everything_default: Optional[int] = None,
+        seed_everything_default: Optional[Union[bool, uint32]] = True,
```
carmocca, line 479: Even though we have the deprecation path for Optional, I would keep the non-deprecated type here

```suggestion
        seed_everything_default: Union[bool, uint32] = True,
```

CHANGELOG.md
```diff
@@ -57,7 +57,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Deprecated `num_processes`, `gpus`, `tpu_cores,` and `ipus` from the `Trainer` constructor in favor of using the `accelerator` and `devices` arguments ([#11040](https://github.com/PyTorchLightning/pytorch-lightning/pull/11040))
 
 
--
+- Deprecated setting `seed_everything_default` in `LightningCLI` to `None`. Set it to `False` instead.  ([#12804](https://github.com/PyTorchLightning/pytorch-lightning/issues/12804)).
```
carmocca, line 60:
```suggestion
- Deprecated setting `LightningCLI(seed_everything_default=None)` in favor of `False` ([#12804](https://github.com/PyTorchLightning/pytorch-lightning/issues/12804)).
```

pytorch_lightning/utilities/cli.py
Schinkikami: That could be. How do we deal with this then? 

if config_seed is not None and config_seed is not False should catch these cases.

However, do we also deal with the edge cases, where the user removes the attribute but sets seed_everything_default to True (the default value now, which could easily be forgotten) or some integer. Because otherwise, the user could accidentally think, that he seeds his program.

carmocca: > However, do we also deal with the edge cases, where the user removes the attribute but sets seed_everything_default to True (the default value now, which could easily be forgotten) or some integer.

I think this is fine, but we shouldn't add the key to the config if it didn't exist already IMO

pytorch_lightning/utilities/cli.py
mauvilsa: If `seed_everything_default` has as type `Optional[Union[np.int32, bool]]` and someone wants to give a value, they will need to cast it to not get a warning, e.g. `seed_everything_default=np.int32(9876)`. Either this or add a comment to disable the warning.

tests/utilities/test_cli.py
```diff
@@ -1547,3 +1549,62 @@ def test_cli_logger_shorthand():
...
+        LightningCLI(
+            TestModel,
+            run=False,
+            seed_everything_default=int(np.iinfo(np.uint32).min - 1),
+            trainer_defaults={"logger": False},
+            parser_kwargs={"error_handler": None},
+        )
+
+    with mock.patch("sys.argv", ["any.py"]):
+        cli = LightningCLI(TestModel, run=False, seed_everything_default=False, trainer_defaults={"logger": False})
```
carmocca, line 1573: question: why do you set `logger: False` in all these cases?

carmocca (approved):

tests/utilities/test_cli.py
Schinkikami: Mostly, because I copied the test from above (test_cli_logger_shorthand) and as the tests do not require a logger I just left it there. I also wanted to make sure no log_dirs or anything are created (didn't know out of my head, when exactly the loggers create their log_dir. If upon initialization or somewhere later).
I can remove it if you want.

carmocca: Yes. You can remove them

The test above set it because it was explicitly testing the logger argument.

This is not going to generate any artifacts since `run=False` is set

pytorch_lightning/utilities/cli.py
Schinkikami: But is forcing the user to cast their `int `seed to `numpy.uint32` (in a pytorch based library) the right way? 
I think the current implementation, with the custom type, is the more user friendly implementation.

If you really think the implementation with the custom type shouldn't be used, an alternative would be to set the argument type of `seed_everything_default` back to `int `(like it was before.. the `seed_everything` function also takes an `int`, so I dont think these two arguments should have different types) and enforce the correct range just in the ArgumentParser with your `register_type(np.int32, deserializer=int32_deserializer)` line.

mauvilsa: This is just how type hints work in python. The value you give must be one of the accepted types. If `int` is not an accepted type, then the value must be casted. It is the same if it is a type created with `restricted_number_type` or `np.int32`, both require to cast. The only benefit of `np.int32` is that a well known type is used instead of creating an equivalent one for this single use case.

The only reason that I mentioned all of this is because you said in #12804

> This change also has the advantage that it prevents passing an int greater than pytorchs int32.max, which results, without raising a warning, in using a different random seed (same behaviour for numpy).

Can you please clarify the problem exactly. Did I understand correctly that giving to `seed_everything` a value `>int32.max` will give different results each time it is run? i.e. it becomes non-deterministic. If this is the case, it does seem like a problem to fix. And not just to fix in the cli, but also in the `seed_everything` function. Also I don't understand why you are now using `uint32`. `np.iinfo(np.uint32).max` is not the same as `np.iinfo(np.int32).max` and with `uint32` no negative is allowed. Which range of values are the ones that create problems?

If for any int value the runs are deterministic, even `>int32.max`, then the type should be kept as `int`.

CHANGELOG.md
```diff
@@ -40,7 +40,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Marked `swa_lrs` argument in `StochasticWeightAveraging` callback as required ([#12556](https://github.com/PyTorchLightning/pytorch-lightning/pull/12556))
 
 
--
+- Changed `seed_everything_default` argument in the `LightningCLI` to type `Union[bool, int]`. If set to `True` a seed is automatically generated for the parser argument `--seed_everything`.
```
awaelchli, line 43: missing PR reference

tests/utilities/test_cli.py
```diff
@@ -1547,3 +1549,67 @@ def test_cli_logger_shorthand():
...
+        cli = LightningCLI(TestModel, run=False, seed_everything_default=False)
+        assert cli.seed_everything_default is False
+        assert cli.config["seed_everything"] is False
+
+    with mock.patch("sys.argv", ["any.py"]):
+        cli = LightningCLI(TestModel, run=False, seed_everything_default=True)
+        assert cli.seed_everything_default is True
+        assert isinstance(cli.config["seed_everything"], int)
+
+    with mock.patch("sys.argv", ["any.py", "--seed_everything", "3"]):
```
awaelchli, line 1586: looks great. you could parametrize all these into a single test.

pytorch_lightning/utilities/cli.py
```diff
@@ -849,6 +860,17 @@ def _prepare_subcommand_kwargs(self, subcommand: str) -> Dict[str, Any]:
             fn_kwargs["datamodule"] = self.datamodule
         return fn_kwargs
 
+    def _set_seed(self) -> None:
+        """Sets the seed."""
+        config_seed = self._get(self.config, "seed_everything")
+
+        if isinstance(config_seed, bool) and config_seed:
+            config_seed = _select_seed_randomly()
```
awaelchli, line 868: why are we setting it randomly though? shouldn't we choose a fixed one, like 0? Otherwise subsequent runs will always be different, unless you pass in a config file.

awaelchli (approved):

pytorch_lightning/utilities/cli.py
Schinkikami: But we want the seed of every run to be different. That is the normal behavior, that you also have when setting no seed (numpy and pytorch then draw random seeds). What we want here is to automatically log them in the config.yaml, so that all experiments that were done with the LightningCLI are automatically reproducible by using the config.yaml.

pytorch_lightning/utilities/cli.py
```diff
@@ -32,12 +33,16 @@
...
+from pytorch_lightning.utilities.rank_zero import _warn, rank_zero_deprecation, rank_zero_warn
+from pytorch_lightning.utilities.seed import _select_seed_randomly
 from pytorch_lightning.utilities.types import LRSchedulerType, LRSchedulerTypeTuple, LRSchedulerTypeUnion
 
 if _JSONARGPARSE_AVAILABLE:
     from jsonargparse import ActionConfigFile, ArgumentParser, class_from_function, Namespace, set_config_read_mode
     from jsonargparse.optionals import import_docstring_parse
+    from jsonargparse.typing import restricted_number_type
+
+    uint32 = restricted_number_type("uint32", int, [(">=", np.iinfo(np.uint32).min), ("<=", np.iinfo(np.uint32).max)])
```
carmocca, line 45: You should define a fallback in the else block

`unit32 = int`

And probably needs a `TypeAlias` given the current mypy error: https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

tests/deprecated_api/test_remove_1-9.py
```diff
@@ -12,9 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from unittest import mock
+
 import pytest
 
 import pytorch_lightning.loggers.base as logger_base
+from pytorch_lightning.core.datamodule import LightningDataModule
```
carmocca, line 20: Unused

```suggestion
```

pytorch_lightning/utilities/cli.py
Schinkikami: The switch from `in32` to `uint32` is because I made a mistake in the original issue. It should have been uint32.

Regarding passing in a wrong seed, if a a value `>np.uint32.max` is passed in the behaviour will be non-deterministic and the pytorch and numpy seeding functions in this case do not throw an warning.
However, I just looked in `seed_everything()` and it seems it already throws a warning (there could be a discussion if it should raise an error instead, but thats not the point of this PR), which I seemed to have missed in the output, when I had an issue with out of range seeds. In this case I would now change back the type of `seed_everything_default` and `--seed_everything` to `int`, as warning would already be raised by `seed_everything()`. Sorry for the confusion.

mauvilsa (approved):

otaj (approved):

carmocca:
You are correct @Schinkikami. I opened #13110 with a fix.

adosar:
@carmocca Does `--seed-everything` includes `workers==True`?

Issue #12647, opened 2022-04-06
Introduce Collaborative Training Strategy!


SeanNaren:
## 🚀 Feature

Over the past few months, I've been working with a library called [hivemind](https://github.com/learning-at-home/hivemind/tree/master/hivemind), where the hivemind team have done amazing things such as the [training transformers together](https://training-transformers-together.github.io/) project. The goal of hivemind is to be able to train across the internet collaboratively, over different machines (like a swarm of machines) rather than having to rely on specialized machine setups (as we traditionally see for distributed training).

I have a working hacky PoC [here](https://github.com/SeanNaren/hivemind-lightning), however, I've made a few iterations privately namely making it a Strategy! The `CollaborativeStrategy` allows machines to connect to each other by passing a list of peers. The Strategy helps makes the experience with hivemind far easier (small things like handling peers/DHT or making changes to the scaler/module where configurations require) and reducing boilerplate in code. I've also successfully trained with the `CollaborativeStrategy` across spot instances, showing that unreliable GPU training is possible!

**The main goal of this ticket is to come to an agreement on whether the `CollaborativeStrategy` should live in PyTorch Lightning, or as a separate integration (in its own repo).**

### Motivation

I believe that a new strategy within PyTorch Lightning will bring users who would like to use spot instances or unreliable GPU machines distributed training to Pytorch Lightning, and make them aware that it is possible!

### Suggested API

The Strategy option makes more sense, as we have to control some of the behaviour of the precision plugin as well as control certain pieces of training. More importantly, currently, the hivemind integration will not work with any other strategy. Being a strategy ensures that the strategy is exclusive.

```python
import pytorch_lightning as pl
from pytorch_lightning.strategies import CollaborativeStrategy

trainer = pl.Trainer(
    strategy=CollaborativeStrategy(target_batch_size=8192)
),
```

When users run the code, they are given a message on how to get clients to join:
```python
python train.py
# Other peers can connect via:
# "INITIAL_PEERS=<PEERS> python ...
# or pass the peers to the strategy: 
# CollaborativeStrategy(initial_peers='<PEERS>')
```

### Pros/Cons

**Why we should add this to PyTorch Lightning**
- Easier access for users who want to use the `CollaborativeStrategy` (just have to install hivemind), hopefully drawing more users who are interested!
- No need to maintain a separate entire repo, with its own CI/docs/maintainers
- Relatively few lines (~400 lines), relying on HiveMind for the heavy lifting

[...]

PR #12723, opened 2022-04-12
Separate strategies' requirements

Borda:
## What does this PR do?

moving Horovod and the additional strategy requirement to own file
updating and cleaning all dockers
update CI workflows
**Fixes:** failing HPU testing

### Does your PR introduce any breaking changes? If yes, please list them.

Horovod is not installed with `.[extra]` but with `.[strategies]`

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @carmocca @akihironitta @borda @tchaton @rohitgr7

PR #12842, opened 2022-04-21
[1/2] Collaborative Strategy

SeanNaren:
## What does this PR do?

Related #12647 

This PR represents the code portion of the strategy. The next PR will be the documentation.

None

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        self.dht.shutdown()
+
+
+class HiveMindScheduler:
+    """Wrapper for schedulers to prevent Lightning from stepping the scheduler too soon.
+
+    This code ensures that we only step when the HiveMind optimizer reaches the global step.
+    """
+
+    def __init__(self, optimizer: "hivemind.Optimizer", scheduler: LRSchedulerTypeUnion):
```
SeanNaren, line 323: Someone pls help me figure out this MyPy issue 😢 

```
pytorch_lightning/strategies/collaborative.py:229: error: Argument "scheduler" to "HiveMindScheduler" has incompatible type "Union[_LRScheduler, ReduceLROnPlateau]"; expected "Union[_LRScheduler, ReduceLROnPlateau]"  [arg-type]
```

I've tried various configurations from the Types utilities but have had no success.

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        return tensor
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        return tensor
+
+    @property
+    def root_device(self) -> torch.device:
+        # todo: cyclic import, someone show me the way
+        from pytorch_lightning.accelerators.cpu import CPUAccelerator
+        from pytorch_lightning.accelerators.gpu import GPUAccelerator
```
SeanNaren, line 276: Anyone know how to fix this?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+            return torch.device(f"cuda:{torch.cuda.current_device()}")
+        elif isinstance(self.accelerator, CPUAccelerator):
+            return torch.device("cpu")
+        raise ValueError
+
+    def model_to_device(self) -> None:
+        assert self.model is not None
+        self.model.to(self.root_device)
+
+    def setup(self, trainer: "pl.Trainer") -> None:
```
awaelchli, line 288: could the setup be moved up, after the properties?

pytorch_lightning/strategies/collaborative.py
awaelchli, line 323:
```suggestion
    def __init__(self, optimizer: "hivemind.Optimizer", scheduler: LRSchedulerTypeUnion) -> None:
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        host_maddrs: Optional[List],
+        initial_peers: Optional[Union[str, List]],
+        persistent: bool,
+        endpoint: Optional[bool],
+        peer_endpoint: Optional[str],
+        host: Optional[str],
+        port: Optional[int],
+        retry_initial_peers: int = 5,
+        retry_peer_sleep_duration: int = 5,
+    ):
```
awaelchli, line 364:
```suggestion
    ) -> None:
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        """Manages the `hivemind.DHT` connection and provides a side-car endpoint server for initial peer access.
+
+        Arguments:
+
+            host_maddrs: List of multi-addrs to create visible peers for other processes.
+            https://learning-at-home.readthedocs.io/en/latest/user/dht.html#running-across-the-internet
+
+            initial_peers: If connecting to a running process, a list of initial peers needs to be passed in.
```
awaelchli, line 372: the formatting is not correct. the new line under it needs to be indented a tab.
for the others too

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+
+            port: When creating the endpoint, the host port to use.
+
+            retry_initial_peers: When connecting to the `peer_endpoint`, how many time to retry before raising
+            an exception.
+
+            retry_peer_sleep_duration: When connecting to the `peer_endpoint`, how long to wait between retries.
+        """
+        logging.basicConfig()
+        log.setLevel(logging.INFO)
```
awaelchli, line 393: this looks wrong. what are you trying to achieve here?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        port = os.environ.get(self.PORT_ENV, self.port)
+        self.port = int(port) if isinstance(port, str) else port
+        self.host = os.environ.get(self.HOST_ENV, self.host)
+
+    @property
+    def disable_logging_checkpointing(self) -> bool:
+        if self.persistent and (self.initial_peers or self.peer_endpoint):
+            # if this node is a peer, we do not log/checkpoint in persistent mode.
+            return True
+        return False
```
awaelchli, line 511:
```
if x:
    return True
return False
```

is the same as 

`return x`

which is preferred.

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        thread.daemon = True
+        thread.start()
+
+    def _get_initial_peers_from_endpoint(self, retry_initial_peers: int, retry_peer_sleep_duration: int) -> List:
+        peers = None
+        for x in range(retry_initial_peers):
+            try:
+                peers = self._get_peers()
+                break
+            except Exception:
```
awaelchli, line 477: do we know a specific type of exception that is being raised?
ConnectionError? ConnectionTimeout?
if it is known, we should catch the specific one(s), please

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+                self.wfile.write("\n".join(visible_peers).encode())
+
+        server = http.server.ThreadingHTTPServer((host, int(port)), DHTHandler)
+        thread = threading.Thread(target=server.serve_forever)
+        thread.daemon = True
+        thread.start()
+
+    def _get_initial_peers_from_endpoint(self, retry_initial_peers: int, retry_peer_sleep_duration: int) -> List:
+        peers = None
+        for x in range(retry_initial_peers):
```
awaelchli, line 473:
```suggestion
        for _ in range(retry_initial_peers):
```

awaelchli (commented):
this is an awesome addition. can't wait for the full docs!

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+        return self.root_device.type == "cuda" and torch.cuda.is_available()
+
+    def reduce(self, tensor: Union[Any, torch.Tensor], *args: Any, **kwargs: Any) -> Union[Any, torch.Tensor]:
+        return tensor
+
+    def all_gather(self, tensor: torch.Tensor, group: Optional[Any] = None, sync_grads: bool = False) -> torch.Tensor:
+        return tensor
+
+    @property
+    def root_device(self) -> torch.device:
```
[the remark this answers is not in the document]
justusschock: > Wouldn't the CollaborativeStrategy better fit as SingleDeviceStrategy?
@awaelchli I disagree here. SingleDeviceStrategy subclasses should not rely/interact with other processes like this one does. It allows training on multiple devices. 

In DDPStrategy you also only have a single device per process. Following your recommendation, ddp would also be a SingleDeviceStrategy?

> Can you clarify this for me: This strategy always runs on a single device. So Trainer(devices=1) is the only valid option?
@SeanNaren what happens if I set devices>1? Is there a possibility to have usual ddp-training per node and only use the elastic/collaborative part across nodes?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+
+    @property
+    def is_global_zero(self) -> bool:
+        return True
+
+    def barrier(self, *args: Any, **kwargs: Any) -> None:
+        pass
+
+    def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
+        return obj
```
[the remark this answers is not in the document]
justusschock: I think that the contract should be to sync/broadcast *if necessary*, so this should really be the strategies responsibility. However, we rely on things like this for syncrhonizing checkpointing and early stopping decisions

@SeanNaren do these work with the CollaborativeStrategy? Will it checkpoint on every process independently?

pytorch_lightning/strategies/collaborative.py
justusschock: I agree, this is nothing a framework (we) should set, this should only be done by the user

SeanNaren: Woops, from debugging days, ignore...

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+from pytorch_lightning.utilities.types import LRSchedulerTypeUnion
+
+if _HIVEMIND_AVAILABLE:
+    import hivemind
+
+log = logging.getLogger(__name__)
+
+
+class CollaborativeStrategy(Strategy):
+    def __init__(
```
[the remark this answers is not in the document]
SeanNaren: Yep, the arguments at the top are nearly always used, the one below are only used (and are lightning specific) for the case of orchestration.

pytorch_lightning/strategies/collaborative.py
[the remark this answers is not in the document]
SeanNaren: Ah this actually fell through the cracks, embarrassing...

Setting devices > 1 right now isn't supported. I need to add a catch for this and test it! I was so close to making it work but it was far too ugly :(

In the future we'll be able to support multiple processes locally via DDP on the local machine, but that will require work to support. The hivemind team will assist here.

In the meantime if the consensus is that it should be setup as a SingleDeviceStrategy I'm alright with this, curious what your preferences are @justusschock @awaelchli

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,511 @@
...
+
+            retry_initial_peers: When connecting to the `peer_endpoint`, how many time to retry before raising
+                an exception.
+
+            retry_peer_sleep_duration: When connecting to the `peer_endpoint`, how long to wait between retries.
+
+            optimizer_kwargs: kwargs are passed to the `hivemind.Optimizer` class.
+        """
+        super().__init__()
+        self.dht_manager = DHTManager(
```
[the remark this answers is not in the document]
SeanNaren: Definitely not, will protect the variables!

I'm unsure what the typical stable/experimental is for strategies.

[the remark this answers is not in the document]
SeanNaren: Protected vars!

Not entirely sure what the process is for stable/experimental, but will make a note in a comment so it doesn't get lost here

SeanNaren:
Huge thanks @awaelchli @justusschock for going through and reviewing, really appreciate it ❤️ 

Two things I need to address:

- Maybe we should consider inheriting from the `SingleDeviceStrategy` instead as currently only single device is supported?
- Should we mark this strategy as experimental/stable?

justusschock:
> Maybe we should consider inheriting from the SingleDeviceStrategy instead as currently only single device is supported?

IMO we shouldn't do that. It has one single device per process, but in total allows training on many devices. That way it behaves like an elastic multi-node DDP if the processes are spawned externally.

> Should we mark this strategy as experimental/stable?
Given most of this code is a wrapper over hivemind, I think calling it experimental is a bit overkill but I can be swayed either way

I think, I'd still mark it as experimental. This gives us more freedom to change stuff (like the per-node DDP thingy you mentioned).
Also we can mark it as stable whenever we want later on.
I think marking this directly as stable may yield false expectations from users.

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+        if not self._hivemind_initialized:
+            self._hivemind_initialized = True
+            if self._batch_size is None:
+                try:
+                    self._batch_size = extract_batch_size(batch)
+                    log.info(f"Found per machine batch size automatically from the batch: {self._batch_size}")
+                except Exception as e:
+                    raise MisconfigurationException(
+                        "We tried to infer the batch size from the first batch of data are were unable to. "
+                        "Please provide the batch size to the Strategy (CollaborativeStrategy"
```
ethanwharris, line 280: Something missing at end of here, maybe:
```suggestion
                        "Please provide the batch size to the Strategy (CollaborativeStrategy) directly."
```

pytorch_lightning/strategies/collaborative.py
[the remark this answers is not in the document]
SeanNaren: These will not work with the collaborative strategy, and currently are orthogonal to the integration. All processes would checkpoint/early stop (since it's fully decentralized).

More importantly synchronization across processes of varying throughput isn't practical especially at the extremes!

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+        retry_peer_sleep_duration: int = 5,
+        **optimizer_kwargs: Any,
+    ):
+        """Provides capabilities to train using the Hivemind Library, training collaboratively across the internet
+        with unreliable machines. `For more information: https://pytorch-
+        lightning.readthedocs.io/en/latest/strategies/collaborative_training.html`.
+
+        .. warning:: ``CollaborativeStrategy`` is experimental and subject to change.
+
+        Arguments:
```
ethanwharris, line 64: I think this needs to be `Args`:
```suggestion
        Args:
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+        endpoint: Optional[bool],
+        peer_endpoint: Optional[str],
+        host: Optional[str],
+        port: Optional[int],
+        retry_initial_peers: int = 5,
+        retry_peer_sleep_duration: int = 5,
+    ) -> None:
+        """Manages the `hivemind.DHT` connection and provides a side-car endpoint server for initial peer access.
+
+        Arguments:
```
ethanwharris, line 364:
```suggestion
        Args:
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+                peers = self._get_peers()
+                break
+            except Exception:
+                log.info(f"Failed to get peers, retrying in {retry_peer_sleep_duration} seconds...")
+                time.sleep(retry_peer_sleep_duration)
+        if peers is None:
+            raise MisconfigurationException(
+                f"Was unable to get peers. Tried {retry_initial_peers} times waiting {retry_peer_sleep_duration}s."
+                f"These parameters can be extended by passing "
+                "to the strategy (CollaborativeStrategy(retry_connection=x, retry_sleep_duration=y)"
```
ethanwharris, line 479:
```suggestion
                f"Unable to get peers. Tried {retry_initial_peers} times waiting {retry_peer_sleep_duration}s."
                f"These parameters can be extended by passing "
                "to the strategy (CollaborativeStrategy(retry_connection=x, retry_sleep_duration=y))."
```

ethanwharris (approved):
LGTM, just some small comments :smiley:

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+        persistent: bool = True,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
+        retry_initial_peers: int = 5,
+        retry_peer_sleep_duration: int = 5,
+        **optimizer_kwargs: Any,
+    ):
+        """Provides capabilities to train using the Hivemind Library, training collaboratively across the internet
+        with unreliable machines. `For more information: https://pytorch-
+        lightning.readthedocs.io/en/latest/strategies/collaborative_training.html`.
```
kaushikb11, line 60:
```suggestion
        with unreliable machines. For more information, `refer to the docs <https://pytorch-
        lightning.readthedocs.io/en/latest/strategies/collaborative_training.html>`__.
```
The previous link reference wouldn't work.

requirements/extra.txt
```diff
@@ -8,3 +8,4 @@ hydra-core>=1.0.5
 jsonargparse[signatures]>=4.6.0
 gcsfs>=2021.5.0
 rich>=10.2.2,!=10.15.*
+hivemind>=1.0.1
```
kaushikb11, line 11: @Borda Have we decided on adding a separate requirements file for strategies?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+        self.world_size = 1
+
+    @property
+    def num_peers(self) -> int:
+        if self._opt:
+            return self._opt.tracker.global_progress.num_peers
+        return 1
+
+    @property
+    def dht(self) -> "hivemind.DHT":
```
kaushikb11, line 189: Could add a docstring what it means. Had to search for it.

```
Distributed Hash Table: distributed “dictionary” where each peer can store and get values.
```

pytorch_lightning/strategies/collaborative.py
kaushikb11: I prefer to be explicit with the code in the error for the users.

```bash
Please provide the batch size to the Strategy by ``Trainer(strategy=CollaborativeStrategy(batch_size=x))`` .....
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+    def state_dict(self) -> Dict:
+        return self.scheduler.state_dict()
+
+
+class DHTManager:
+    ENDPOINT_ENV: str = "ENDPOINT"
+    PEER_ENDPOINT_ENV: str = "PEER_ENDPOINT"
+    INITIAL_PEERS_ENV: str = "INITIAL_PEERS"
+    HOST_ENV: str = "HOST"
+    PORT_ENV: str = "PORT"
```
kaushikb11, line 346: We usually add `PL_` prefix to the env variables used by PyTorch Lightning internally.
Should we do that here as well to avoid potential conflicts?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+
+    def __init__(
+        self,
+        host_maddrs: Optional[List],
+        initial_peers: Optional[Union[str, List]],
+        persistent: bool,
+        endpoint: Optional[bool],
+        peer_endpoint: Optional[str],
+        host: Optional[str],
+        port: Optional[int],
```
kaushikb11, line 358: Why not add defaults here?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+        if self._endpoint:
+            self._host = self._host if self._host is not None else self.DEFAULT_HOST
+            self._port = self._port if self._port is not None else self.DEFAULT_PORT
+            self._start_server_process(self._host, self._port)
+            self._log_endpoint_helper_message(visible_addresses)
+        elif self._peer_endpoint:
+            log.info("Machine received initial peers from endpoint.")
+        elif self._initial_peers is None:
+            log.info(
+                f"\nOther machines can connect running the same command:\n"
```
kaushikb11, line 422:
```suggestion
                "\nOther machines can connect running the same command:\n"
```

SeanNaren:
Lots of untested pieces of code, making a TODO list:

- Test the batch size finder throws the error when unsuccessful
- Test the endpoint is setup correctly when a URL is passed in
- Test that `get_peers` is called the correct number of times based on `retry_initial_peers` (probably want to change this to `retry_endpoint`

tests/strategies/test_collaborative.py
```diff
@@ -0,0 +1,344 @@
...
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.types import STEP_OUTPUT
+from tests.helpers import BoringModel
+from tests.helpers.runif import RunIf
+
+if _HIVEMIND_AVAILABLE:
+    import hivemind
+
+
+@mock.patch("pytorch_lightning.strategies.collaborative._HIVEMIND_AVAILABLE", new_callable=lambda: False)
```
kaushikb11, line 26:
```suggestion
@mock.patch("pytorch_lightning.strategies.collaborative._HIVEMIND_AVAILABLE", return_value=False)
```

pytorch_lightning/strategies/collaborative.py
SeanNaren: Looking at the other strategies they all follow the `Arguments` name, am I missing something?

tests/strategies/test_collaborative.py
```diff
@@ -0,0 +1,344 @@
...
+@mock.patch.dict(os.environ, {"HIVEMIND_MEMORY_SHARING_STRATEGY": "file_descriptor"}, clear=True)
+def test_scaler_updated_precision_16(mocked_cuda_available, mocked_device_count):
+    class TestModel(BoringModel):
+        def on_fit_start(self) -> None:
+            assert isinstance(self.trainer.precision_plugin.scaler, hivemind.GradScaler)
+            raise SystemExit
+
+    model = TestModel()
+    trainer = pl.Trainer(
+        strategy=CollaborativeStrategy(target_batch_size=1),
```
kaushikb11, line 337: I think it's the first strategy in PyTorch Lightning without an alias 🦜

pytorch_lightning/utilities/imports.py
```diff
@@ -103,6 +103,7 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
 _FAIRSCALE_OSS_FP16_BROADCAST_AVAILABLE = _FAIRSCALE_AVAILABLE and _compare_version("fairscale", operator.ge, "0.3.3")
 _FAIRSCALE_FULLY_SHARDED_AVAILABLE = _FAIRSCALE_AVAILABLE and _compare_version("fairscale", operator.ge, "0.3.4")
 _GROUP_AVAILABLE = not _IS_WINDOWS and _module_available("torch.distributed.group")
+_HIVEMIND_AVAILABLE = _package_available("hivemind")
```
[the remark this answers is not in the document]
SeanNaren: Will wait till the sample PR is done just to make sure I do it right (I also saw a potential way to just make it a one line change, idk if it's a lazy implementation though)

tests/strategies/test_collaborative.py
SeanNaren: I was thinking about how I could make an alias, as it would be neat to do it from the command line, but I felt like for now we shouldn't try to shoehorn this in

From my experiments, you're always having to extend the class with a bunch of parameters! I think once we start making more QoL improvements we can revisit

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,504 @@
...
+            self._hivemind_initialized = True
+            if self._batch_size is None:
+                try:
+                    self._batch_size = extract_batch_size(batch)
+                    log.info(f"Found per machine batch size automatically from the batch: {self._batch_size}")
+                except Exception as e:
+                    raise MisconfigurationException(
+                        "We tried to infer the batch size from the first batch of data are were unable to. "
+                        "Please provide the batch size to the Strategy (CollaborativeStrategy"
+                        f"exception raised: {e}"
```
kaushikb11, line 286: This is the error message inside `extract_batch_size`

```
    error_msg = (
        "We could not infer the batch_size from the batch. Either simplify its structure"
        " or provide the batch_size as `self.log(..., batch_size=batch_size)`."
    )
```

So, there would be a duplication and confusion with both the errors being mentioned.

pytorch_lightning/strategies/collaborative.py
ethanwharris: Ah, looks like both `Args` and `Arguments` are ok, although the google style uses `Args`. Within PL usage is just inconsistent haha

pytorch_lightning/strategies/collaborative.py
SeanNaren: Since we catch the underlying exception, this will never be printed to the stdout

I've modified this exception to be more specific (I was just catching Exception beforehand)

pytorch_lightning/strategies/collaborative.py
ethanwharris: So I think this can be ignored :smiley:

requirements/extra.txt
Borda: #12723

pytorch_lightning/strategies/collaborative.py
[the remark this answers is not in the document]
justusschock: Understood. I think for checkpointing that's fine, however for early stopping we probably need a way to sync this as it does not make sense to only continue training for some processes (while the others quit).

Does that mean that in general there is no way to sync anything in hivemind?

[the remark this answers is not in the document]
SeanNaren: For early stopping, all processes are technically in sync so they should all stop at the same time (there is some stuff we could probably do with syncing state via the distributed hash table, but I think it's unnecessary).

the DHT is how you would sync state, however this should definitely be saved for a future PR (and is exposed for users who want to mess with it via `self.strategy.dht`!

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,509 @@
...
+                try:
+                    self._batch_size = extract_batch_size(batch)
+                    log.info(f"Found per machine batch size automatically from the batch: {self._batch_size}")
+                except (MisconfigurationException, RecursionError) as e:
+                    raise MisconfigurationException(
+                        "We tried to infer the batch size from the first batch of data. "
+                        "Please provide the batch size to the Strategy by "
+                        "``Trainer(strategy=CollaborativeStrategy(batch_size=x))``. "
+                        f"exception raised: {e}"
+                    )
```
akihironitta, line 287: nit: even though the original error message is included, maybe we should explicitly chain the exceptions.
```suggestion
                    ) from e
```

tests/strategies/test_collaborative.py
```diff
@@ -0,0 +1,370 @@
...
+    """Test when we are a second peer (passing initial peers or peer endpoint) we warn the user that
+    logging/checkpointing will be disabled."""
+    with pytest.warns(UserWarning, match="This machine is not a persistent machine"):
+        CollaborativeStrategy(target_batch_size=1, initial_peers=initial_peers, peer_endpoint=peer_endpoint)
+
+
+@RunIf(hivemind=True)
+@mock.patch.dict(
+    os.environ,
+    {"HIVEMIND_MEMORY_SHARING_STRATEGY": "file_descriptor", "PORT": str(find_free_network_port())},
```
akihironitta, line 59: noob question: Should this env var be PL-specific one which gets read by `DHTManager` class?
```suggestion
    {"HIVEMIND_MEMORY_SHARING_STRATEGY": "file_descriptor", "PL_PORT": str(find_free_network_port())},
```

akihironitta (commented):
🚀

tests/strategies/test_collaborative.py
SeanNaren: Yes it should! Great catch :)

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,509 @@
...
+        batch_size: Optional[int] = None,
+        delay_state_averaging: bool = False,
+        delay_optimizer_step: Optional[bool] = None,
+        delay_grad_averaging: bool = False,
+        offload_optimizer: Optional[bool] = None,
+        reuse_grad_buffers: bool = False,
+        scheduler_fn: Optional[Callable] = None,
+        matchmaking_time: float = 5.0,
+        averaging_timeout: float = 30.0,
+        verbose: bool = True,
```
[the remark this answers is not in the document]
SeanNaren: Fully agree here

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+        lightning.readthedocs.io/en/latest/strategies/collaborative_training.html>`__.
+
+        .. warning:: ``CollaborativeStrategy`` is experimental and subject to change.
+
+        Arguments:
+
+            target_batch_size: When training, the batch size to accumulate to before running a step. The larger this
+                batch size, the more work can be done asynchronously without communication.
+
+            run_id: A unique identifier of this training run, used as a common prefix for all DHT keys.
```
carmocca, line 69: For new users, it's unclear what DHT is when reading from top to bottom.

Maybe link to the DHT class

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+            batch_size: The local batch size per process. If not provided,
+                we infer this from the first batch of data passed in at training.
+
+            delay_state_averaging: If enabled (default), average parameters and extra tensors in a background thread;
+                if set to False, average parameters synchronously within the corresponding hivemind.Optimizer.step call.
+
+            delay_optimizer_step: Run optimizer in background, apply results in future .step.
+                requires `offload_optimizer`.
+
+            delay_grad_averaging: Average gradients in background; requires offload_optimizer and delay_optimizer_step.
```
carmocca, line 80: Same comment about the paramrefs (please build the docs and check I didn't make a mistake)

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+
+            offload_optimizer: Offload the optimizer to host memory, saving GPU memory for parameters and gradients.
+
+            reuse_grad_buffers: Use model's .grad buffers for gradient accumulation.
```
carmocca, line 84:
```suggestion
            reuse_grad_buffers: Use the model's gradient buffers for gradient accumulation.
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+                This is more memory efficient, but it requires that we do not call `zero_grad` in the `LightningModule`.
```
carmocca, line 85:
```suggestion
                This is more memory efficient, but it requires that we do not call ``zero_grad`` in the ``LightningModule``.
```

carmocca, line 85: This comment is a bit confusing. Does that mean that when you set this flag, `zero_grad` will automatically be skipped in the LightningModule?

Or does the user need to do it manually?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+
+            scheduler_fn: callable(optimizer) -> PyTorch LRScheduler or a pre-initialized PyTorch scheduler.
+                When using `offload_optimizer`/`delay_optimizer_step`/`delay_state_averaging` a scheduler_fn is required
+                instead of passing a scheduler to the `pl.LightningModule`. This is because the optimizer is re-created
```
carmocca, line 89: What do you mean with "passing a scheduler to the LM"? Are you referring to returning a scheduler from `configure_optimizers`?

Can't this strategy re-create it for the user when necessary?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+                and the scheduler needs to be re-created as well.
+
+            matchmaking_time: When looking for group, wait for peers to join for up to this many seconds.
+                Increase if you see "averaged gradients with N peers" where N is below 0.9x on >=25% of epochs.
```
carmocca, line 93: Is this a heuristic? Is that why you dont' do it automatically?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+            host: When creating the endpoint, the host IP to use.
+
+            port: When creating the endpoint, the host port to use.
+
+            retry_endpoint_attempts: When connecting to the `peer_endpoint`, how many time to retry before raising
+                an exception.
+
+            retry_endpoint_sleep_duration: When connecting to the `peer_endpoint`, how long to wait between retries.
+
+            optimizer_kwargs: kwargs are passed to the `hivemind.Optimizer` class.
```
carmocca, line 130:
```suggestion
            \**optimizer_kwargs: kwargs are passed to the :class:`hivemind.Optimizer` class.
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+                server log/checkpoint. If `persistent` is True, we do not log/checkpoint from other processes.
+
+            host: When creating the endpoint, the host IP to use.
+
+            port: When creating the endpoint, the host port to use.
+
+            retry_endpoint_attempts: When connecting to the `peer_endpoint`, how many time to retry before raising
+                an exception.
+
+            retry_endpoint_sleep_duration: When connecting to the `peer_endpoint`, how long to wait between retries.
```
carmocca, line 128: Same comment about paramref

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+            initial_peers: If connecting to a running process, a list of initial peers needs to be passed in.
+                This can also be set via the env variable `INITIAL_PEERS`.
+
+            endpoint: Enable if a side-car endpoint server is required on the process to server initial peers.
+                This is useful when using some form of orchestration such as torchelastic.
+
+            peer_endpoint: The endpoint to request initial peers from.
+
+            persistent: When using an endpoint, this controls whether other processes that are not the endpoint
+                server log/checkpoint. If `persistent` is True, we do not log/checkpoint from other processes.
```
carmocca, line 119:
```suggestion
                server log/checkpoint. If ``persistent`` is True, we do not log/checkpoint from other processes.
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+
+            verbose: Report internal Hivemind events such as accumulating gradients and running background tasks.
+
+            averager_opts: Additional keyword arguments forwarded to both GradientAverager and TrainingStateAverager.
+
+            host_maddrs: List of multi-addrs to create visible peers for other processes.
+                `https://learning-at-home.readthedocs.io/en/latest/user/dht.html#running-across-the-internet`
+
+            initial_peers: If connecting to a running process, a list of initial peers needs to be passed in.
+                This can also be set via the env variable `INITIAL_PEERS`.
```
carmocca, line 111:
```suggestion
                This can also be set via the env variable ``INITIAL_PEERS``.
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+            averaging_timeout: If an averaging step hangs for this long, it will be cancelled automatically.
+                Increase averaging_timeout if you see "Proceeding with local gradients" at least 25% of the time.
+                Do not set this timeout too high, as it may cause your optimizer to hang
+                after some types of network errors.
+
+            allreduce_timeout: Timeout for a single attempt to run all-reduce.
+
+            verbose: Report internal Hivemind events such as accumulating gradients and running background tasks.
+
+            averager_opts: Additional keyword arguments forwarded to both GradientAverager and TrainingStateAverager.
```
carmocca, line 105:
```suggestion
            averager_opts: Additional keyword arguments forwarded to both ``GradientAverager`` and ``TrainingStateAverager``.
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+            offload_optimizer=offload_optimizer,
+            averager_opts=averager_opts if averaging_timeout is not None else dict(request_timeout=1.0),
+            verbose=verbose,
+            reuse_grad_buffers=reuse_grad_buffers,
+            **optimizer_kwargs,
+        )
+
+        # a bit of a hack to only log from the stable server
+        if self.dht_manager.disable_logging_checkpointing:
+            warnings.warn(
```
carmocca, line 172: should this use `rank_zero_warn`?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+                "This machine is not a persistent machine. Checkpointing/Logging has been disabled.", UserWarning
+            )
+        rank_zero_only.rank = 1 if self.dht_manager.disable_logging_checkpointing else 0
+        self._hivemind_initialized = False
+
+        self.global_rank = 0
+        self.local_rank = 0
+        self.world_size = 1
```
carmocca, line 180: Why does these need to be set here? Isn't this a responsibility of the environment?

And should they be read-only?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+    def dht(self) -> "hivemind.DHT":
+        """Hivemind Distributed Hash Table which stores values across all peers.
+
+        See documentation for more details: `https://learning-at-home.readthedocs.io/en/latest/modules/dht.html`
+        """
+        return self.dht_manager.dht
+
+    @property
+    def root_device(self) -> torch.device:
+        # todo: cyclic import, someone show me the way
```
carmocca, line 198: IMO this is fine to do:

```suggestion
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+
+        if isinstance(self.accelerator, GPUAccelerator):
+            return torch.device(f"cuda:{torch.cuda.current_device()}")
+        elif isinstance(self.accelerator, CPUAccelerator):
+            return torch.device("cpu")
+        raise ValueError
+
+    @property
+    def is_global_zero(self) -> bool:
+        return True
```
carmocca, line 210: If you set `self.global_rank` in the __init__, shouldn't this check that?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+
+    def _initialize_hivemind(self) -> None:
+        if len(self.optimizers) > 1:
+            raise MisconfigurationException("Hivemind only supports training with one optimizer.")
+        optimizer = self.optimizers[0]
+
+        if self._require_scheduler_fn and self._scheduler_fn is None:
+            rank_zero_warn(
+                "Enabling delay_optimizer_step, delay_state_averaging or offload_optimizer "
+                "requires a scheduler_fn to be passed to the strategy if a scheduler is being used "
```
carmocca, line 226:
```suggestion
                "Enabling `delay_optimizer_step`, `delay_state_averaging` or `offload_optimizer` "
                "requires a `scheduler_fn` to be passed to the strategy if a scheduler is being used "
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+        self.optimizers = [opt]
+        self._opt = opt
+
+        if self._reuse_grad_buffers:
+            assert self.lightning_module is not None
+            # turn off zero_grad by Lightning
+            if is_overridden("optimizer_zero_grad", self.lightning_module):
+                rank_zero_warn(
+                    "You have overridden `optimizer_zero_grad` which will be disabled. "
+                    "When CollaborativeStrategy(reuse_grad_buffers=True), "
```
carmocca, line 256:
```suggestion
                    "When `CollaborativeStrategy(reuse_grad_buffers=True)`, "
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,513 @@
...
+                    "the optimizer cannot call zero grad, as this would "
+                    "delete the gradients before they are averaged."
+                )
+
+            def override_fn(*args: Any, **kwargs: Any) -> None:
+                pass
+
+            self.lightning_module.optimizer_zero_grad = override_fn  # type: ignore[assignment]
```
carmocca, line 264: This *should* work too, and has the benefit that when this is called again later, `is_overridden` should return False

```suggestion
            self.lightning_module.optimizer_zero_grad = None
```

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,509 @@
...
+
+        Arguments:
+
+            target_batch_size: When training, the batch size to accumulate to before running a step. The larger this
+                batch size, the more work can be done asynchronously without communication.
+
+            run_id: A unique identifier of this training run, used as a common prefix for all DHT keys.
+
+            batch_size: The local batch size per process. If not provided,
+                we infer this from the first batch of data passed in at training.
```
carmocca, line 72: This should mention that the batch size cannot (or should not) change during training

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,509 @@
...
+            peer_endpoint: The endpoint to request initial peers from.
+
+            persistent: When using an endpoint, this controls whether other processes that are not the endpoint
+                server log/checkpoint. If `persistent` is True, we do not log/checkpoint from other processes.
+
+            host: When creating the endpoint, the host IP to use.
+
+            port: When creating the endpoint, the host port to use.
+
+            retry_endpoint_attempts: When connecting to the `peer_endpoint`, how many time to retry before raising
```
carmocca, line 389: I think the `CollaborativeStrategy` docstring could paramref all these arguments or the other way around to avoid duplication

pytorch_lightning/utilities/imports.py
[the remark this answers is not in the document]
SeanNaren: I'll close this till the PR upstream gets worked on.

pytorch_lightning/strategies/collaborative.py
SeanNaren: It's extremely difficult to recreate the scheduler, this is why it's unfortunately cleaner to ask the user to pass an FN. This is something that can be looked at in future work!

pytorch_lightning/strategies/collaborative.py
SeanNaren: It's tricky as you do not want to hang for too long (if processes are on a very unstable connection). automating some of these variables is a next step from this PR, this just exposes vars that are important for hivemind.

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,528 @@
...
+
+        Arguments:
+
+            target_batch_size: When training, the batch size to accumulate to before running a step. The larger this
+                batch size, the more work can be done asynchronously without communication.
+
+            run_id: A unique identifier of this training run, used as a common prefix for all DHT keys.
+                See ``https://learning-at-home.readthedocs.io/en/latest/user/dht.html``.
+
+            batch_size: The local batch size per process. If not provided, we infer this from the first batch of data
```
Borda, line 72: can we comment on how this works together with `batch_size` defined in datamodule?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,528 @@
...
+    @property
+    def root_device(self) -> torch.device:
+        from pytorch_lightning.accelerators.cpu import CPUAccelerator
+        from pytorch_lightning.accelerators.gpu import GPUAccelerator
+
+        if isinstance(self.accelerator, GPUAccelerator):
+            return torch.device(f"cuda:{torch.cuda.current_device()}")
+        elif isinstance(self.accelerator, CPUAccelerator):
+            return torch.device("cpu")
+        raise ValueError
```
Borda, line 211: any message?

Borda (approved):

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,528 @@
...
+                :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.peer_endpoint`,
+                how many time to retry before raising an exception.
+
+            retry_endpoint_sleep_duration: When connecting to the
+                :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.peer_endpoint`,
+                how long to wait between retries.
+
+            **optimizer_kwargs: kwargs are passed to the :class:`hivemind.Optimizer` class.
+        """
+        if not _HIVEMIND_AVAILABLE:
```
kaushikb11, line 141: Should we also check here if the system is linux?

kaushikb11 (approved):
🔥

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,528 @@
...
+
+            host_maddrs: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.host_maddrs`
+
+            initial_peers: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.initial_peers`
+
+            endpoint: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.endpoint`
+
+            peer_endpoint: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.peer_endpoint`
+
+            persistent: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.persistent`
```
rohitgr7, line 401:
```suggestion
            persistent: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.persistent`

            endpoint: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.endpoint`

            peer_endpoint: :paramref:`~pytorch_lightning.strategies.collaborative.CollaborativeStrategy.peer_endpoint`
```

requirements/strategies.txt
```diff
@@ -1,3 +1,4 @@
 fairscale>=0.4.5
 deepspeed<0.6.0
 horovod>=0.21.2,!=0.24.0  # no need to install with [pytorch] as pytorch is already installed
+hivemind>=1.0.1; sys_platform == 'linux'
```
rohitgr7, line 4: it works only on linux? if yes, should we add some checks and throw exceptions otherwise?

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,528 @@
...
+            matchmaking_time: When looking for group, wait for peers to join for up to this many seconds.
+                Increase if you see "averaged gradients with N peers" where N is below 0.9x on >=25% of epochs.
+                Training with low-latency network, decreasing matchmaking_time allows training with smaller batch sizes.
+
+            averaging_timeout: If an averaging step hangs for this long, it will be cancelled automatically.
+                Increase averaging_timeout if you see "Proceeding with local gradients" at least 25% of the time.
+                Do not set this timeout too high, as it may cause your optimizer to hang
+                after some types of network errors.
+
+            allreduce_timeout: Timeout for a single attempt to run all-reduce.
```
rohitgr7, line 107: this is not an argument. is it misseD?

rohitgr7 (commented):
still reviewing

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,529 @@
...
+                optimizer=opt,
+                scheduler=scheduler,
+            )
+
+    def on_train_batch_start(self, batch: Any, batch_idx: int, dataloader_idx: int = 0) -> None:
+        if not self._hivemind_initialized:
+            self._hivemind_initialized = True
+            # todo (sean): we could technically support a dynamic batch size by inferring each step
+            # and passing it to the ``hivemind.Optimizer``.
+            if self._batch_size is None:
```
rohitgr7, line 295: what if batch_size differs for validation/test.. will it consider it the same for all?

rohitgr7 (approved):
awesome 🚀

pytorch_lightning/strategies/collaborative.py
SeanNaren: This only affects training, no sync during test/validation

requirements/strategies.txt
SeanNaren: Has been done as suggested by @kaushikb11!

pytorch_lightning/strategies/collaborative.py
```diff
@@ -0,0 +1,527 @@
...
+    @property
+    def root_device(self) -> torch.device:
+        from pytorch_lightning.accelerators.cpu import CPUAccelerator
+        from pytorch_lightning.accelerators.gpu import GPUAccelerator
+
+        if isinstance(self.accelerator, GPUAccelerator):
+            return torch.device(f"cuda:{torch.cuda.current_device()}")
+        elif isinstance(self.accelerator, CPUAccelerator):
+            return torch.device("cpu")
+        raise MisconfigurationException(f"Was unable to infer device type from the accelerator: {self.accelerator}")
```
rohitgr7, line 210:
```suggestion
        raise MisconfigurationException(
            f"Was unable to infer device type from the accelerator: {self.accelerator.__class__.__name__}."
        )
```

Borda (commented):
gatekeeper/pass: @SeanNaren

Issue #12883, opened 2022-04-26
Address current CI blocker


carmocca:
Updating fairscale triggered the following error:

```python
____________________ test_registries_register_automatically ____________________

    def test_registries_register_automatically():
        assert "SaveConfigCallback" not in CALLBACK_REGISTRY
        with mock.patch("sys.argv", ["any.py"]):
>           LightningCLI(BoringModel, run=False, auto_registry=True)

/__w/pytorch-lightning/pytorch-lightning/tests/utilities/test_cli.py:956: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/__w/pytorch-lightning/pytorch-lightning/pytorch_lightning/utilities/cli.py:551: in __init__
    self.setup_parser(run, main_kwargs, subparser_kwargs)
/__w/pytorch-lightning/pytorch-lightning/pytorch_lightning/utilities/cli.py:590: in setup_parser
    self._add_arguments(self.parser)
/__w/pytorch-lightning/pytorch-lightning/pytorch_lightning/utilities/cli.py:630: in _add_arguments
    parser.add_optimizer_args(OPTIMIZER_REGISTRY.classes)
/__w/pytorch-lightning/pytorch-lightning/pytorch_lightning/utilities/cli.py:223: in add_optimizer_args
    self.add_subclass_arguments(optimizer_class, nested_key, **kwargs)
/root/miniconda3/envs/lightning/lib/python3.8/site-packages/jsonargparse/signatures.py:481: in add_subclass_arguments
    doc_group = self._gather_docstrings(baseclass, get_class_init_and_base_docstrings)[0]
/root/miniconda3/envs/lightning/lib/python3.8/site-packages/jsonargparse/signatures.py:527: in _gather_docstrings
    docstring = docstring_parse(doc)
/root/miniconda3/envs/lightning/lib/python3.8/site-packages/docstring_parser/parser.py:42: in parse
    raise exc
/root/miniconda3/envs/lightning/lib/python3.8/site-packages/docstring_parser/parser.py:35: in parse
    ret = module.parse(text)
/root/miniconda3/envs/lightning/lib/python3.8/site-packages/docstring_parser/google.py:295: in parse
    return GoogleParser().parse(text)
[...]

PR #12884, opened 2022-04-26
Update `jsonargparse` to unblock master

carmocca:
Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/12883

cc @tchaton @rohitgr7 @carmocca @akihironitta @borda


requirements/extra.txt
```diff
@@ -6,5 +6,6 @@ torchtext>=0.9.*
 omegaconf>=2.0.5
 hydra-core>=1.0.5
 jsonargparse[signatures]>=4.6.0
+docstring-parser==0.13
```
kaushikb11, line 9: @carmocca Could you add a comment next to it on why we pinned it?

carmocca: There's git blame for that

Borda: can we have some limited range? `docstring-parser<=0.13`

rohitgr7 (approved):

requirements/extra.txt
carmocca:
```suggestion
docstring-parser<=0.13
```

kaushikb11 (approved):

requirements/extra.txt
```diff
@@ -6,5 +6,6 @@ torchtext>=0.9.*
 omegaconf>=2.0.5
 hydra-core>=1.0.5
 jsonargparse[signatures]>=4.6.0
+docstring-parser<0.14
```
mauvilsa, line 9: I can fix it in jsonargparse and create a patch release so that we can bump jsonargparse instead of adding docstring-parse.

carmocca: Whatever lands faster is good with me :)

mauvilsa, line 9:
```suggestion
jsonargparse[signatures]>=4.7.1
```

otaj (approved):

akihironitta:
It now fails with `TypeError` in CPU tests. Is it due to the new version of jsonargparse? @mauvilsa 
```
>           docstring = parse(component.__doc__)
E           TypeError: 'tuple' object is not callable

/home/runner/work/pytorch-lightning/pytorch-lightning/pytorch_lightning/utilities/cli.py:894: TypeError
```
https://github.com/PyTorchLightning/pytorch-lightning/runs/6177267111?check_suite_focus=true

pytorch_lightning/utilities/cli.py
```diff
@@ -889,7 +889,7 @@ def instantiate_class(args: Union[Any, Tuple[Any, ...]], init: Dict[str, Any]) -
 
 
 def _get_short_description(component: object) -> Optional[str]:
-    parse = import_docstring_parse("LightningCLI(run=True)")
+    parse, _ = import_docstring_parse("LightningCLI(run=True)")
```
carmocca, line 892: This was a breaking change. @mauvilsa you might want to have jsonargparse expose the `get_short_description` function

SeanNaren (approved):

PR #12897, opened 2022-04-27
Fix number of references to LightningModule

### LINKED ISSUE #12622: sys.getrefcount() of a LightningModule
## 🐛 Bug

When creating a `LightningModule`, `sys.getrefcount()` returns 3 (vs. returning 2 for a conventional torch `Module`). In my particular case, for instance, this results in memory issues: even if I explicitly do `del <LightningModule>`, the memory is not released, and I believe this is the root of the issue.

I had originally discussed this problem on Slack. As discussed, I'm creating an issue for it and pinging @carmocca. 

### To Reproduce

Input:
```
import sys
import torch

a = torch.nn.Module()
b = pl.LightningModule()
print(sys.getrefcount(b))
print(sys.getrefcount(a))
```
Output:
```
3
2
```

### Expected behavior

Output:
```
2
2
```

### Environment

- PyTorch Lightning Version: 1.5.9
- PyTorch Version: 1.10.2+cu113
- Python version: 3.8.12
- OS: Linux
- CUDA/cuDNN version: 11.6
- GPU models and configuration: Quadro RTX 8000
- How you installed PyTorch: pip

### Additional context

Per @carmocca's advice on Slack, a workaround is to do `b._load_state_dict_pre_hooks.clear()` before deleting the `LightningModule`. This indeed seems to work. However, I'm not sure about an explanation as to why.

Thanks!

cc @carmocca @justusschock @awaelchli @borda @ananthsub @ninginthecloud @jjenniferdai @rohitgr7

otaj:
## What does this PR do?

I know, this looks way too hacky, the proposal in bug report proposes to deal with it in `__del__` method. Problem is, the `__del__` method will never be called, exactly because of this issue, because there will always be an internal reference to self. There will be accompanying PR to pytorch to fix it on their side

Fixes #12622

## Before submitting

## PR review

## Did you have fun?

cc @otaj @borda @carmocca @justusschock @awaelchli @ananthsub @ninginthecloud @jjenniferdai @rohitgr7


tests/core/test_lightning_module.py
```diff
@@ -399,3 +400,11 @@ def configure_gradient_clipping(self, optimizer, optimizer_idx, gradient_clip_va
         match=r"gradient_clip_algorithm='norm'\)` and have passed `clip_gradients\(gradient_clip_algorithm='foo'",
     ):
         trainer.fit(model)
+
+
+@RunIf(skip_windows=True, min_torch="1.10")
```
rohitgr7, line 405: qq:
why skip on windows and min torch?

CHANGELOG.md
```diff
@@ -157,6 +157,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Use only a single instance of `rich.console.Console` throughout codebase ([#12886](https://github.com/PyTorchLightning/pytorch-lightning/pull/12886))
 
 
+- Fixed number of references of LightningModule so it can be deleted ([#12897](https://github.com/PyTorchLightning/pytorch-lightning/pull/12897))
```
rohitgr7, line 160:
```suggestion
- Fixed the number of references of `LightningModule` so it can be deleted ([#12897](https://github.com/PyTorchLightning/pytorch-lightning/pull/12897))
```

rohitgr7 (approved):

tests/core/test_lightning_module.py
carmocca: Since it's a small test, I suggest removing the skip.

Even if it's not a bug under those circumstances, it's a nice check to have

```suggestion
```

pytorch_lightning/core/lightning.py
```diff
@@ -1976,4 +1977,6 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
         from torch.distributed._sharded_tensor import pre_load_state_dict_hook, state_dict_hook
 
         self._register_state_dict_hook(state_dict_hook)
-        self._register_load_state_dict_pre_hook(pre_load_state_dict_hook, True)
+
+        # We need to make sure the self inside the method is a weakref proxy
+        Module._register_load_state_dict_pre_hook(weakref.proxy(self), pre_load_state_dict_hook, True)
```
carmocca, line 1982: Nice hack!

Might be overkill, but I would suggest getting the parent class dynamically

```suggestion
        parent = inspect.getmro(self.__class__)[1]
        parent._register_load_state_dict_pre_hook(weakref.proxy(self), pre_load_state_dict_hook, True)
```

otaj: Is it possible that the parent would be anything other than Module? And do we even want to invoke it if it is something else than Module? I can imagine in user defined classes which inherits from LightningModule, this would actually resolve to LightningModule, right?

Borda (approved):

pytorch_lightning/core/lightning.py
```diff
@@ -1976,4 +1977,6 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
         from torch.distributed._sharded_tensor import pre_load_state_dict_hook, state_dict_hook
 
         self._register_state_dict_hook(state_dict_hook)
-        self._register_load_state_dict_pre_hook(pre_load_state_dict_hook, True)
+
+        # We need to make sure the self inside the method is a weakref proxy
```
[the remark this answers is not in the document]
otaj: This has been merged in upstream, although a little bit differently.

carmocca (approved):

awaelchli (approved):

PR #12918, opened 2022-04-28
Remove use of jsonargparse internals

mauvilsa:
## What does this PR do?

The `import_docstring_parse` function is not part of the jsonargparse public API, see https://jsonargparse.readthedocs.io/en/stable/#api-reference. This pull request removes its use to avoid possible failures due to changes in jsonargparse internals. An example of this happening can be seen in https://github.com/PyTorchLightning/pytorch-lightning/pull/12884#discussion_r858812880.

I think this change should be included in 1.6.X. Even though `import_docstring_parse` is internal, I will change it again to be backward compatible so that released versions of `LightningCLI` do not fail due to this, see [jsonargparse/commit/6dc8539](https://github.com/omni-us/jsonargparse/commit/6dc85394e10ea4ee30e5f5dd2617bbee6e5c2893).

None

## Before submitting

## PR review

## Did you have fun?


carmocca (approved):

Borda (approved):

pytorch_lightning/utilities/cli.py
```diff
@@ -889,9 +893,11 @@ def instantiate_class(args: Union[Any, Tuple[Any, ...]], init: Dict[str, Any]) -
...
-    try:
-        docstring = parse(component.__doc__)
-        return docstring.short_description
-    except ValueError:
-        rank_zero_warn(f"Failed parsing docstring for {component}")
+    if not docstring_parser:
+        rank_zero_warn(f"Failed parsing docstring for {component}: docstring-parser package is required")
+    else:
+        try:
+            docstring = docstring_parser.parse(component.__doc__)
```
otaj, line 900: We should resolve failing mypy check, which is saying, that `docstring_parser.parse` expects `str`, but we are passing `Optional[str]`

otaj (approved):

pytorch_lightning/utilities/cli.py
```diff
@@ -37,13 +37,17 @@
...
     set_config_read_mode(fsspec_enabled=True)
 else:
     locals()["ArgumentParser"] = object
     locals()["Namespace"] = object
 
+try:
+    import docstring_parser
+except ImportError:
+    docstring_parser = False
+
```
otaj, line 50: Oh, mypy is (rightfully) complaining also about this. Can we have some guarding variable so that we are not assigning bool to module?

pytorch_lightning/utilities/cli.py
mauvilsa: mypy still fails, but now due to the optional import. Will fix it.

pytorch_lightning/utilities/cli.py
```diff
@@ -889,9 +891,13 @@ def instantiate_class(args: Union[Any, Tuple[Any, ...]], init: Dict[str, Any]) -
...
 def _get_short_description(component: object) -> Optional[str]:
-    parse, _ = import_docstring_parse("LightningCLI(run=True)")
-    try:
-        docstring = parse(component.__doc__)
-        return docstring.short_description
-    except ValueError:
-        rank_zero_warn(f"Failed parsing docstring for {component}")
+    if component.__doc__ is None:
+        return None
+    if not _DOCSTRING_PARSER_AVAILABLE:
```
rohitgr7, line 896: qq:

`docstring_parse` is part of base `jsonargparse[signatures]`
https://github.com/omni-us/jsonargparse/blob/6dc85394e10ea4ee30e5f5dd2617bbee6e5c2893/setup.cfg#L23-L24

and isn't `jsonargparse[signatures]` required for cli?

mauvilsa: Yes. Though it is possible that someone installs pytorch-lightning without extras and installs jsonargparse without signatures.

akihironitta (approved):

pytorch_lightning/utilities/cli.py
rohitgr7: does cli work without signatures? I haven't used it much but when I did, it always recommended to install `jsonargparse[signatures]`. so just confirming this.

if it works with jsonargparse, what extensions does `signatures` provide?

mauvilsa: Yes, it works without `signatures`. Without it the docstrings are not used, which is quite a shortcoming because the help is not that useful. But the running of code is not affected. Some time ago `signatures` included more dependencies but were removed on some refactoring.

rohitgr7: then should we do `JSONARGPARSE_SIGNATURES_AVAILABLE` instead of `_DOCSTRING_PARSER_AVAILABLE`? since this is the only dependency within `signatures`.

mauvilsa: > then should we do `JSONARGPARSE_SIGNATURES_AVAILABLE` instead of `_DOCSTRING_PARSER_AVAILABLE`? since this is the only dependency within `signatures`.
> 
> just trying to keep only one external dependency here i.e `jsonargparse` and not users confuse it with another package that is not even present anywhere in PL requirements.

That does sound better to me. Later I can create a pull request with this change.

mauvilsa: Created #13035

PR #12932, opened 2022-04-29
Add hook test for reloading with max epochs

carmocca:
## What does this PR do?

Add test to ease the review of #12821

Setting the 1.6.x milestone because #12821 will need to land there

None

### DIFF
--- a/tests/models/test_hooks.py
+++ b/tests/models/test_hooks.py
@@ -574,7 +574,114 @@ def test_trainer_model_hook_system_fit(tmpdir, kwargs, automatic_optimization):
     assert called == expected
 
 
-def test_trainer_model_hook_system_fit_no_val_and_resume(tmpdir):
+def test_trainer_model_hook_system_fit_no_val_and_resume_max_epochs(tmpdir):
+    # initial training to get a checkpoint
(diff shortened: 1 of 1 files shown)


rohitgr7 (approved):

otaj (approved):

Borda (approved):

tests/models/test_hooks.py
```diff
@@ -574,7 +574,114 @@ def training_step(self, batch, batch_idx):
...
+        enable_progress_bar=False,
+        enable_model_summary=False,
+        callbacks=[HookedCallback([])],
+    )
+    trainer.fit(model)
+    best_model_path = trainer.checkpoint_callback.best_model_path
+
+    called = []
+    callback = HookedCallback(called)
+    # already performed 1 step, resume and do 2 more
```
awaelchli, line 594: Can you explain the comment? I think it's wrong.

We first train 2 steps (1 epoch). 
Then, we resume to do 2 more steps (1 additional epoch)

rohitgr7: yep... it's a small typo... must be copied from the test right below

PR #12935, opened 2022-04-29
fix recursive call to deletion in sys.modules from pl_legacy_patch

### LINKED ISSUE #12922: Checkpoint loading patch fails when using a pre-trained featurizer within a Lightning module.
## 🐛 Bug

I am getting the error: `ModuleNotFoundError: No module named 'pytorch_lightning.utilities.argparse_utils'` coming from 
migration.py when loading checkpoints for models which leverage pre-trained models also trained via Pytorch Lightning.

Essentially, the context manager is called recursively, since the model relies on features generated from a pre-trained model. This model also requires a call to load_from_checkpoint(). As a result, the __exit__ function of the context manager gets called twice in a row, with the error occurring on the second call.

### Expected behavior

I would expect checkpoint loading to not raise en error because of a pre-trained featurizer model.

### Environment

pytorch-lightning == 1.6.1

cc @awaelchli @ananthsub @ninginthecloud @rohitgr7

jdavid04:
This PR fixes errors occurring from recursive calls to the checkpoint loading patch described in #12922. As mentioned in the issue, I had trouble checking the tests locally, but this should be a safe modification.

Fixes  #12922.

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/utilities/migration.py
+++ b/pytorch_lightning/utilities/migration.py
@@ -52,4 +52,3 @@ class pl_legacy_patch:
 
         if "pytorch_lightning.utilities.argparse_utils" in sys.modules:
             del sys.modules["pytorch_lightning.utilities.argparse_utils"]
-
--- a/pytorch_lightning/utilities/migration.py
+++ b/pytorch_lightning/utilities/migration.py
@@ -14,13 +14,9 @@
 from __future__ import annotations
 
 import sys
-import threading
 from types import ModuleType, TracebackType
 
 import pytorch_lightning.utilities.argparse
-
-# Create a global lock to ensure no race condition with deleting sys modules
-_lock = threading.Lock()
 
 
 class pl_legacy_patch:
@@ -39,7 +35,7 @@ class pl_legacy_patch:
     """
 
     def __enter__(self) -> None:
-        _lock.acquire()
+
         # `pl.utilities.argparse_utils` was renamed to `pl.utilities.argparse`
         legacy_argparse_module = ModuleType("pytorch_lightning.utilities.argparse_utils")
         sys.modules["pytorch_lightning.utilities.argparse_utils"] = legacy_argparse_module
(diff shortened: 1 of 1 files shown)


jdavid04:
@awaelchli let me know if you'd like me to change anything here. I'm not sure what is going wrong with the CI but it does not seem to be related to the changes made.

pytorch_lightning/utilities/migration.py
```diff
@@ -14,14 +14,10 @@
...
 
 import sys
-import threading
 from types import ModuleType, TracebackType
 
 import pytorch_lightning.utilities.argparse
 
-# Create a global lock to ensure no race condition with deleting sys modules
-_lock = threading.Lock()
-
```
awaelchli, line 24: @jdavid04 I don't see how removing the lock is solving the issue. Is this a bad merge?

pytorch_lightning/utilities/migration.py
```diff
@@ -53,5 +49,6 @@ def __exit__(
     ) -> None:
         if hasattr(pytorch_lightning.utilities.argparse, "_gpus_arg_default"):
             delattr(pytorch_lightning.utilities.argparse, "_gpus_arg_default")
-        del sys.modules["pytorch_lightning.utilities.argparse_utils"]
-        _lock.release()
+
+        if "pytorch_lightning.utilities.argparse_utils" in sys.modules:
+            del sys.modules["pytorch_lightning.utilities.argparse_utils"]
```
awaelchli, line 54: Is this the fix you are proposing?

awaelchli, line 54: Is this the fix you are proposing?

pytorch_lightning/utilities/migration.py
jdavid04: Yes sorry, I mentioned this in the original issue. I had additional issues with the thread locking in place when using Lightning with Ray Tune. Model selection never terminates and I believe it's related to the addition of this lock because it doesn't occur without it. I propose to remove it because the fix below should fix the multithreading issues with checkpointing also.

awaelchli: @jdavid04 Ok. Unfortunately we added the thread locking here just a few days ago to solve another issue reported by users. cc @krshrimali 

We should try to find a fix for both.

awaelchli: Could you explain to us (ideally with a reproducible example) what issues you are experiencing with the lock and why it is happening?

jdavid04: Sure, I'll have a chance to get back to this in a few days and will try to reproduce both issues with a simpler example.

pytorch_lightning/utilities/migration.py
jdavid04: Yes, as far as I can tell this simple check should fix both the previous issue with multithreading and this issue.

Issue #12438, opened 2022-03-24
Whether clarification/documentation/redesign is needed for customizing LightningCLI subcommands


mauvilsa:
The [LightningCLI documentation](https://pytorch-lightning.readthedocs.io/en/latest/common/lightning_cli.html) does not mention any case in which trainer methods (CLI subcommands) are customized. It would be good to include it so that people are more aware of a recommended way to do this.

On the other hand, the `LightningCLI` class currently has an undocumented feature that supports the implementation of methods with the same name as the `Trainer` class, see https://github.com/PyTorchLightning/pytorch-lightning/pull/11714#discussion_r798721175. Despite that discussion, still to me it seems weird that `return_predictions` option is there by default. It does nothing unless you implement the `LightningCLI.predict`. Documenting this would not change this weirdness. Furthermore, if documented it would seem that this is the recommended way to customize a `Trainer` method. But this has the limitation that parameters added e.g. to `LightningCLI.predict`, would not be exposed as command line options. I think there is a need for some redesign and only document once it is clear what are the recommended ways to implement the most common use cases. This github issue could be used for the discussion about this posible redesign.

I am not convinced that it is a good idea that to customize `Trainer` methods, people implement methods with the same name in the `LightningCLI` class. I think a better way to do this is to subclass `Trainer` and pass this to the `trainer_class` init parameter of `LightningCLI`. One reason I meantioned before is that new parameters of the method would become CLI options. Another reason is that these kind of customizations can be useful without using `LightningCLI`, so recommending to implement this logic in a subclass of `LightningCLI` would go against it. An one more reason, it is more natural to customize the method where it was originally defined than in a different class.

Another thought I have is that by default, as far as possible, all `LightningCLI` subcommands and options should do something obvious/useful, and not have options that don't do anything. What could be the default behavior for `predict`? I haven't used predict, so I could be misinterpreting its purpose. But I would expect as default one of two possibilities. The first is the predictions are always printed to stdout. This would imply that `return_predictions` is not a CLI option. The second possibility is that `predict` has an `--output` option which could support stdout, a local file or even some remote location. Again `return_predictions` wouldn't be a CLI option, though internally its value could depend on the `--output` value. If someone wants a behavior different from the default, then some implementation is needed.

To anayze whether a redesign is needed and come up with a good solution, would be good to list a few use cases and discuss how this could/should be implemented. I give a few that come to mind.

[...]

PR #12940, opened 2022-04-30
Add ability to load checkpoints from cli

randombenj:
## What does this PR do?

**What does it do?**

This PR adds the ability to load checkpoints from the Lightning CLI. At the moment, checkpoint loading
is limited to the model (trainer, data loader could be added easily if requested).

**Why is this change important?**

I think the lightningcli is a fantastic tool to manage different trainer/model/dataloader configurations.
Especially that the model params are loaded from `yaml` config files.

In my case I would like to export the trained models to `onnx` or `torchscript` with the same hyperparameter config 
(loaded by lightningcli).

```python
# run with --load_from_checkpoint path/to/checkpoint.ckpt
cli = LightningCLI(run=False)
cli.model.to_onnx("model.onnx")
```

In the future it would also be nice to load trainer and dataloaders from checkpoints to continue training ...

No

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Include a version suffix for new "last" checkpoints of later runs in the same directory ([#12902](https://github.com/PyTorchLightning/pytorch-lightning/pull/12902))
 
 
+- Add ability to load checkpoints when using the `LightningCLI` by providing `--load_from_checkpoint path/to/checkpoint.ckpt` ([#12940](https://github.com/PyTorchLightning/pytorch-lightning/pull/12940))
+
+
 -
 
 ### Changed
--- a/pytorch_lightning/utilities/cli.py
+++ b/pytorch_lightning/utilities/cli.py
@@ -609,6 +609,12 @@ class LightningCLI:
                 "Set to True to use a random seed."
             ),
         )
+        parser.add_argument(
+            "--load_from_checkpoint",
+            type=Optional[str],
+            default=None,
+            help="Load the lightningmodule from a given checkpoint",
+        )
 
     def add_core_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
         """Adds arguments from the core classes to the parser."""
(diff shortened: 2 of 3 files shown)


randombenj:
After looking through some PR's I think the failing tests are not due to my change (I might be mistaken however).
Even the main branch seems to not have passing checks for quite some time now :)

carmocca:
Any thoughts on this @mauvilsa?

> In the future it would also be nice to load trainer and dataloaders from checkpoints to continue training ...

This cannot be easily done as they are supposed to be reloaded when the `Trainer` starts running

randombenj:
@carmocca Well maybe not for the dataloaders but `ckpt_path` could also be added to the `trainer.fit` function (according to the docs) https://pytorch-lightning.readthedocs.io/en/stable/common/checkpointing.html#checkpoint-loading.

This could enable interrupting and continuing training in via LightningCLI.

Another idea would be to  add an `export` subcommand to https://github.com/PyTorchLightning/pytorch-lightning/blob/fd14bbb7e44b77eb8500326e22e1de18e2ea1d6d/pytorch_lightning/utilities/cli.py#L661 to easily export a trained model with weights to `onnx/torchscript/...` (forwarding sample data for tracing could then be another feature, maybe taken from the dataloader, when `Module.example_input_array` is not defined in the Module to export

Or also just run inference with pretrained weights using LightningCLI ...

mauvilsa:
Adding globally an `--load_from_checkpoint` option means that it will be available for all subcommands and instantiation only, i.e. `run=False`. But the subcommands already have the `ckpt_path` parameter, which would make it redundant and confusing.

The idea of having a subcommand for exporting makes more sense to me. This would be for a use case in which someone trains a bunch of models and only wants to export one or a few of them. If the use case is that always when a model is fit it should also be exported, then this is more for the user to add to `after_fit` or in a callback that implements `on_fit_end`.

> Or also just run inference with pretrained weights using LightningCLI ...

There is already the `predict` subcommand for this, no? This might have some relation to the discussion in #12438.

randombenj:
> Adding globally an --load_from_checkpoint option means that it will be available for all subcommands and instantiation only, i.e. run=False. But the subcommands already have the ckpt_path parameter, which would make it redundant and confusing.

I was not aware that one could use `ckpt_path` for the subcommands, but I do agree that two checkpoint options do not make sense here. RTFM I guess :) 

> The idea of having a subcommand for exporting makes more sense to me. This would be for a use case in which someone trains a bunch of models and only wants to export one or a few of them. If the use case is that always when a model is fit it should also be exported, then this is more for the user to add to `after_fit` or in a callback that implements `on_fit_end`.

Should close this PR then and maybe add an `export` subcommand as this would cover my use case. But wouldn't you anyway something similar like this PR to only load the model weights for exporting? I guess one shouldn't instantiate a trainer just to export the model?

PR #12981, opened 2022-05-05
Improve support for custom `DataLoader`s when instantiated in `*_dataloader` hook

### LINKED ISSUE #12917: MisconfigurationException: Trying to inject `DistributedSampler` into the `AnnLoader` instance
## 🐛 Bug

This a copy of the [issue 757](https://github.com/scverse/anndata/issues/757) posted at the anndata github repository. 

I have been trying to implement an MLP to predict cell type labels using pyTorch Lightning and the AnnLoader function from the anndata Python package. For the implementation, I followed the [AnnLoader tutorial to interface with pyTorch models](https://anndata-tutorials.readthedocs.io/en/latest/annloader.html) and the PyTorch Lightning tutorial.
I aim to implement the training, test and prediction methods, and run it on a GPU. I tested my code on a Google Colabs instance. The error message is the same for GPU and CPU runtime.  
When I try to predict a cell type label using the predict function, pyTorch lightning wants to use the DistributedSampler as sampler, which is not implemented in the AnnLoader and I could not figure out how to disable the sampler.
My intuition is that this is an issue with the setup of the prediction step as the test step works without error. 

### To Reproduce

```
import gdown
import torch
import torch.nn as nn

import numpy as np
import scanpy as sc
from sklearn.preprocessing import OneHotEncoder, LabelEncoder
from torchmetrics.functional import accuracy
from anndata.experimental.pytorch import AnnLoader

#define model class 
class MLP(pl.LightningModule):
  
  def __init__(self, input_dim, hidden_dims, out_dim):
    super().__init__()
    modules = []
    for in_size, out_size in zip([input_dim]+hidden_dims, hidden_dims):
        modules.append(nn.Linear(in_size, out_size))
        modules.append(nn.LayerNorm(out_size))
        modules.append(nn.ReLU())
        modules.append(nn.Dropout(p=0.05))
    modules.append(nn.Linear(hidden_dims[-1], out_dim))
    self.layers = nn.Sequential(*modules)
    
    self.ce = nn.CrossEntropyLoss()
    
  def forward(self, x):
    return self.layers(x)
  
    # here, a batch has data (x) and labels (y). What is returned by
    # batch depends on the __get_item__() implementation in your Dataset
    x = batch.X
    y = batch.obs['cell_type'] #hard coded, please adapt
    x = x.view(x.size(0), -1)
    y_hat = self.layers(x)
    loss = self.ce(y_hat, y)
    self.log('train_loss', loss)
    return loss
  
    # here, a batch has data (x) and labels (y). What is returned by
    # batch depends on the __get_item__() implementation in your Dataset
    x = batch.X
    y = batch.obs['cell_type'] #hard coded, please adapt
    x = x.view(x.size(0), -1)
    y_hat = self.layers(x)
    loss = self.ce(y_hat, y)
    y_hat = torch.argmax(y_hat, dim=1)
    acc = accuracy(y_hat, y)
    metrics = dict({
        'test_loss': loss.clone().detach(),
        'test_acc': acc.clone().detach(),
    })
    self.log_dict(metrics, batch_size=len(y))
    return metrics

  def predict_step(self, batch, batch_idx, dataloader_idx=0):
    # here, a batch has data (x) and labels (y). What is returned by
    # batch depends on the __get_item__() implementation in your Dataset
    x = batch.X
    x = x.view(x.size(0), -1)
    y_hat = self.model(x)
    return y_hat

    optimizer = torch.optim.Adam(self.parameters(), lr=1e-4)
    return optimizer

#load normalized pancreas data from AnnLoader tutorial 
from google.colab import drive
drive.mount('/content/drive')
file_path = '/content/drive/My Drive/pancreas_normalized.h5ad'

adata = sc.read(file_path)
adata.X = adata.raw.X # put raw counts to .X

#prepare AnnLoader 
encoder_study = OneHotEncoder(sparse=False, dtype=np.float32)
encoder_study.fit(adata.obs['study'].to_numpy()[:, None])

encoder_celltype = LabelEncoder()
encoder_celltype.fit(adata.obs['cell_type'])

use_cuda = torch.cuda.is_available()

encoders = {
    'obs': {
        'study': lambda s: encoder_study.transform(s.to_numpy()[:, None]),
        'cell_type': encoder_celltype.transform
    }
}

# Load data as dataLoader, split in train and test data  
dataloader = AnnLoader(adata[adata.obs['study']!='Pancreas Fluidigm C1'], batch_size=128, shuffle=True, convert=encoders, use_cuda=use_cuda)
dataloader_test = AnnLoader(adata[adata.obs['study']=='Pancreas Fluidigm C1'], batch_size=128, #sampler = sampler,  
                            shuffle=False, convert=encoders, use_cuda=use_cuda)

#create MLP model, configure pytorch lightning trainer 
mlp = MLP(input_dim = adata.n_vars, hidden_dims = [128,128], out_dim=8)
trainer = pl.Trainer(auto_scale_batch_size='power', gpus=1, deterministic=True, 
                     max_epochs=5, replace_sampler_ddp=False) 
# Train the model
trainer.fit(mlp, dataloader)

# Perform evaluation
trainer.test(mlp, dataloader_test)
## output [{'test_acc': 0.9620253443717957, 'test_loss': 0.12309074401855469}]

# Return predictions
trainer.predict(mlp, dataloader_test)
```

Here is the error message from the prediction step:
```
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
/usr/local/lib/python3.7/dist-packages/pytorch_lightning/trainer/connectors/data_connector.py:489: PossibleUserWarning: Your `predict_dataloader`'s sampler has shuffling enabled, it is strongly recommended that you turn shuffling off for val/test/predict dataloaders.
  category=PossibleUserWarning,
---------------------------------------------------------------------------
MisconfigurationException                 Traceback (most recent call last)
[<ipython-input-18-0ce04d8b9a10>](https://localhost:8080/#) in <module>()
      1 # Return predictions
----> 2 trainer.predict(mlp, dataloader_test)

11 frames
[/usr/local/lib/python3.7/dist-packages/pytorch_lightning/utilities/data.py](https://localhost:8080/#) in _get_dataloader_init_kwargs(dataloader, sampler, mode)
    240         dataloader_cls_name = dataloader.__class__.__name__
    241         raise MisconfigurationException(
--> 242             f"Trying to inject `DistributedSampler` into the `{dataloader_cls_name}` instance. "
    243             "This would fail as some of the `__init__` arguments are not available as instance attributes. "
    244             f"The missing attributes are {required_args}. "

MisconfigurationException: Trying to inject `DistributedSampler` into the `AnnLoader` instance. This would fail as some of the `__init__` arguments are not available as instance attributes. The missing attributes are ['adatas']. HINT: If you wrote the `AnnLoader` class, define `self.missing_arg_name` or manually add the `DistributedSampler` as: `AnnLoader(dataset, sampler=DistributedSampler(dataset))`.
```

### Expected behavior

I expected a return of the predicted label per cell (i.e. per input). 

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.6.0
- PyTorch Version (e.g., 1.10): 1.10.0+cu111
- Python version (e.g., 3.9): 3.7.13
- OS (e.g., Linux): Linux-5.4.144+-x86_64-with-Ubuntu-18.04-bionic
- CUDA/cuDNN version: cu111 (from a Google Colab instance)

Output of all versions:
```
pyasn1                                      0.4.8
pyasn1_modules                              0.2.8
pydev_ipython                               NA
pydevconsole                                NA
pydevd                                      2.0.0
pydevd_concurrency_analyser                 NA
pydevd_file_utils                           NA
pydevd_plugins                              NA
pydevd_tracing                              NA
pydot_ng                                    2.0.0
pygments                                    2.6.1
pyparsing                                   3.0.7
pytorch_lightning                           1.6.0
pytz                                        2018.9
regex                                       2.5.72
requests                                    2.23.0
rsa                                         4.8
scipy                                       1.4.1
session_info                                1.0.0
setuptools                                  57.4.0
simplegeneric                               NA
sitecustomize                               NA
six                                         1.15.0
sklearn                                     1.0.2
socks                                       1.7.1
sphinxcontrib                               NA
storemagic                                  NA
tblib                                       1.7.0
tensorboard                                 2.8.0
tensorflow                                  2.8.0
termcolor                                   1.1.0
threadpoolctl                               3.1.0
toolz                                       0.11.2
torch                                       1.10.0+cu111
torchmetrics                                0.7.3
torchtext                                   0.11.0
torchvision                                 0.11.1+cu111
tornado                                     5.1.1
tqdm                                        4.63.0
traitlets                                   5.1.1
typing_extensions                           NA
uritemplate                                 3.0.1
urllib3                                     1.24.3
wcwidth                                     0.2.5
webencodings                                0.5.1
wrapt                                       1.14.0
yaml                                        6.0
zipp                                        NA
zmq                                         22.3.0
-----
IPython             5.5.0
jupyter_client      5.3.5
jupyter_core        4.9.2
notebook            5.3.1
-----
Python 3.7.13 (default, Mar 16 2022, 17:37:17) [GCC 7.5.0]
Linux-5.4.144+-x86_64-with-Ubuntu-18.04-bionic
-----
Session information updated at 2022-04-13 15:04
```

### Additional context

cc @justusschock @awaelchli @ninginthecloud @rohitgr7

otaj:
## What does this PR do?

Fixes #12917
Improve support for custom `DataLoader`s when instantiated in `*_dataloader` hook. If the DataLoader is a valid DataLoader, there is now a lower chance that things we will break the DataLoader when injecting custom arguments.

Also, if things break due to our injecting, we throw better error messages.

`DataLoader` instantiated inside a hook will now not instantiate attributes for you automatically. Before this PR, this code would work:

```python
class MyDataLoader(DataLoader):
    def __init__(dataset, my_arg, **kwargs):
        # we explicitly do not set my_arg, as PL sets it for us automatically
        super().__init__(dataset, **kwargs)

class MyModule(pl.LightningModule):
    ...
    def train_dataloader(self):
        loader = MyDataLoader(range(10), 10, batch_size=2)
        assert loader.my_arg == 10
        return loader
```

After this PR, this code would no longer work. This was an internal behavior.

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @ninginthecloud @rohitgr7 @otaj


pytorch_lightning/utilities/data.py
```diff
@@ -197,6 +197,8 @@ def _update_dataloader(
             f" can be passed to its parent's `{argument}=...` `__init__` argument. This is likely caused by allowing"
             f" passing both a custom argument that will map to the `{argument}` argument as well as `**kwargs`."
             f" `kwargs` should be filtered to make sure they don't contain the `{argument}` key."
+            " This argument was automatically passed to your DataLoader by PyTorch Lightning,"
+            " however you are also passing it in your implementation."
```
carmocca, line 201: Don't these 2 new lines say the same as the message above?

pytorch_lightning/utilities/data.py
```diff
@@ -336,23 +352,28 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
     return wrapper
 
 
 # https://stackoverflow.com/a/63851681/9201239
 def _get_all_subclasses(cls: Type[Any]) -> Set[Type[Any]]:
-    """Returns a list of all classes that inherit directly or indirectly from the given class."""
-    subclasses = set()
+    """Returns a list of all classes that inherit directly or indirectly from the given class, including the base
+    class."""
+    subclasses = {cls}
```
carmocca, line 376: I would not make this modification. For example, it's used in the CLI where we don't want to include the class itself.

I suggest creating a new function that includes itself or manually adding it where you need it.

pytorch_lightning/utilities/data.py
```diff
@@ -336,23 +352,28 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
-                continue
-            setattr(obj, arg_name, arg_value)
+            if not hasattr(obj, "__" + arg_name):
+                # Set the value privately if it has not been set yet. This achieves two things:
+                # 1. The argument is the one that was passed in the outermost call
+                #    and not overriden by a call to `super().__init__()`
+                # 2. The argument is the passed value and does not get overwritten in `__init__()`
+                setattr(obj, "__" + arg_name, arg_value)
+                set_arg_names.append(arg_name)
+        if not hasattr(obj, "_set_arg_names"):
```
carmocca, line 364: Why this if?

And if we want it, shouldn't it also avoid the `for` loop entirely?

pytorch_lightning/utilities/data.py
```diff
@@ -336,23 +352,28 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
             if param.name != "self" and param.kind not in (param.VAR_POSITIONAL, param.VAR_KEYWORD)
         ]
 
-        cls = type(obj)
+        set_arg_names = []
```
carmocca, line 355: Why not initialize it as a set?

pytorch_lightning/utilities/data.py
```diff
@@ -336,23 +352,28 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
-                # will set `self._arg_name = arg_value` in `__init__` which is the attribute returned by the `arg_name`
-                # property so we are fine skipping in that case
-                continue
-            setattr(obj, arg_name, arg_value)
+            if not hasattr(obj, "__" + arg_name):
+                # Set the value privately if it has not been set yet. This achieves two things:
+                # 1. The argument is the one that was passed in the outermost call
+                #    and not overriden by a call to `super().__init__()`
+                # 2. The argument is the passed value and does not get overwritten in `__init__()`
+                setattr(obj, "__" + arg_name, arg_value)
```
carmocca, line 362: noob q: is this going to do the double underscore name mangling? Or not because we are using setattr?

pytorch_lightning/utilities/data.py
```diff
@@ -336,23 +352,28 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
+        set_arg_names = []
         for arg_name, arg_value in chain(zip(param_names, args), kwargs.items()):
-            if hasattr(cls, arg_name) and getattr(cls, arg_name).fset is None:
-                # the class defines a read-only (no setter) property of this name. it's likely that the implementation
-                # will set `self._arg_name = arg_value` in `__init__` which is the attribute returned by the `arg_name`
-                # property so we are fine skipping in that case
-                continue
-            setattr(obj, arg_name, arg_value)
+            if not hasattr(obj, "__" + arg_name):
+                # Set the value privately if it has not been set yet. This achieves two things:
```
carmocca, line 358: I understand the benefits achieved, but are they a requirement for the intended changes in this PR?

pytorch_lightning/utilities/data.py
carmocca, line 362: (this comment applies to master too): do you think we should set a proxy here? These would be owning references otherwise, right?

pytorch_lightning/utilities/data.py
```diff
@@ -259,12 +277,10 @@ def _get_dataloader_init_kwargs(
             missing_kwargs = sorted(missing_kwargs)
             dataloader_cls_name = dataloader.__class__.__name__
             raise MisconfigurationException(
-                f"Trying to inject `DistributedSampler` into the `{dataloader_cls_name}` instance. "
+                f"Trying to inject custom `Sampler` into the `{dataloader_cls_name}` instance. "
```
carmocca, line 280: I would be broad:

```suggestion
                f"Trying to inject parameters into the `{dataloader_cls_name}` instance. "
```

pytorch_lightning/utilities/data.py
```diff
@@ -259,12 +277,10 @@ def _get_dataloader_init_kwargs(
...
                 "This would fail as it doesn't expose all its attributes in the `__init__` signature. "
-                f"The missing arguments are {missing_kwargs}. "
-                f"HINT: If you wrote the `{dataloader_cls_name}` class, add the `__init__` arguments or "
-                "manually add the `DistributedSampler` as: "
-                f"`{dataloader_cls_name}(dataset, sampler=DistributedSampler(dataset))`."
+                f"The missing arguments are {missing_kwargs}. HINT: If you wrote the `{dataloader_cls_name}` class, "
+                "add the `__init__` arguments or allow passing **kwargs"
```
carmocca, line 283:
```suggestion
                "add the `__init__` arguments or allow passing `**kwargs`"
```

pytorch_lightning/utilities/data.py
```diff
@@ -244,12 +264,10 @@ def _get_dataloader_init_kwargs(
...
         raise MisconfigurationException(
-            f"Trying to inject `DistributedSampler` into the `{dataloader_cls_name}` instance. "
+            f"Trying to inject custom `Sampler` into the `{dataloader_cls_name}` instance. "
             "This would fail as some of the `__init__` arguments are not available as instance attributes. "
-            f"The missing attributes are {required_args}. "
-            f"HINT: If you wrote the `{dataloader_cls_name}` class, define `self.missing_arg_name` or "
-            "manually add the `DistributedSampler` as: "
-            f"`{dataloader_cls_name}(dataset, sampler=DistributedSampler(dataset))`."
+            f"The missing attributes are {required_args}. If you instantiate your `{dataloader_cls_name}` inside a "
+            "`*_dataloader` hook of your module, this will likely go away."
```
carmocca, line 270: (the suggestion needs a line break)

```suggestion
            "`*_dataloader` hook of your module, we will do this for you. Otherwise, define `self.missing_arg_name` inside your `__init__`."
```

pytorch_lightning/utilities/data.py
```diff
@@ -208,8 +210,14 @@ def _get_dataloader_init_kwargs(
...
-    # get the dataloader instance attributes
-    attrs = {k: v for k, v in vars(dataloader).items() if not k.startswith("_")}
+    was_wrapped = hasattr(dataloader, "_set_arg_names")
+    if was_wrapped:
+        attrs = {k: getattr(dataloader, "__" + k) for k in dataloader._set_arg_names}
+        original_dataset = dataloader.__dataset  # we have this saved from _wrap_init
+    else:
+        # get the dataloader instance attributes
+        attrs = {k: v for k, v in vars(dataloader).items() if not k.startswith("_")}
+        original_dataset = attrs["dataset"]
```
carmocca, line 220: If it wasn't wrapped, can we do this? The DataLoader might not define it.

pytorch_lightning/utilities/data.py
```diff
@@ -218,17 +226,29 @@ def _get_dataloader_init_kwargs(
...
-        params.update(inspect.signature(DataLoader.__init__).parameters)
-        del params["self"]
+
+        # if the dataloader was wrapped in a hook, only take arguments with default values
+        #  and assume user passes their kwargs correctly
+        params.update(
+            {
+                k: v
+                for k, v in inspect.signature(DataLoader.__init__).parameters.items()
+                if not was_wrapped or (v.default is not v.empty)
```
carmocca, line 236: was_wrapped is invariant to this loop so it should go outside.

tests/utilities/test_data.py
```diff
@@ -133,23 +133,28 @@ def test_has_len_all_rank():
...
-    dataloader = BadImpl([1, 2, 3])
+    dataloader = BadStandaloneGoodHookImpl([1, 2, 3])
     with pytest.raises(MisconfigurationException, match="`DataLoader` implementation has an error.*`dataset`"):
         _update_dataloader(dataloader, dataloader.sampler)
 
-    class BadImpl2(DataLoader):
+    with _replace_dataloader_init_method():
+        dataloader = BadStandaloneGoodHookImpl([1, 2, 3])
+        new_dataloader = _update_dataloader(dataloader, dataloader.sampler)
+        assert isinstance(new_dataloader, BadStandaloneGoodHookImpl)
```
carmocca, line 149:
```suggestion
    new_dataloader = _update_dataloader(dataloader, dataloader.sampler)
    assert isinstance(new_dataloader, BadStandaloneGoodHookImpl)
```

pytorch_lightning/utilities/data.py
```diff
@@ -336,23 +352,28 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
 
-        cls = type(obj)
+        set_arg_names = []
         for arg_name, arg_value in chain(zip(param_names, args), kwargs.items()):
-            if hasattr(cls, arg_name) and getattr(cls, arg_name).fset is None:
-                # the class defines a read-only (no setter) property of this name. it's likely that the implementation
-                # will set `self._arg_name = arg_value` in `__init__` which is the attribute returned by the `arg_name`
-                # property so we are fine skipping in that case
-                continue
-            setattr(obj, arg_name, arg_value)
```
carmocca, line 346: Not setting this would now be a breaking change for people using 1.6 who rely on this behaviour

pytorch_lightning/utilities/data.py
otaj: Suprisingly, no. The Python doc is not very clear on that, but name mangling happens only if the invoking class is the one where it should happen. The last paragraph of [this](https://docs.python.org/3/tutorial/classes.html#private-variables) is probably closest to reasonable official explanation.

otaj: Actually, no, and it would be a bad idea to set it as a proxy. We want to be able to pull these values in the moment we are reinstantiating the instance with our parameters. And, what is even better is the fact that when we are reinstantiating, we throw away the old instance and return don't use this wrapped init, but the original one, therefore in the end, we don't have these references.

pytorch_lightning/utilities/data.py
otaj: Do people rely on this particular behavior? Or do they rely on the fact that their class will be reinstantiated correctly? I believe how we manage the reinstantiation is an implementation detail.

I believe a good "test" for whether people rely on this particular behavior and we shouldn't remove it is whether we advertised it in the docs (such as "if you create a DataLoader in a hook, you don't have to set your attributes, we'll do it for you")

pytorch_lightning/utilities/data.py
otaj: Good catch!

pytorch_lightning/utilities/data.py
otaj: So, ideally we want to only capture variables in the original, first call + `dataset`. The reason why we want `dataset` variable is just this line https://github.com/PyTorchLightning/pytorch-lightning/pull/12981/files#diff-f21f92490e6c1417367a48324385cf66c908d35fb4321a4b1b20e5ba410830f4R251

The same reason actually applies to changing the function `_get_all_subclasses`

pytorch_lightning/utilities/data.py
otaj: I don't think so - the message above does not say anything about the fact it is PL code that is injecting some kwargs there and might confuse people

pytorch_lightning/utilities/data.py
otaj: No, they are not a requirement, but I think it's only a matter of time before someone will come with an implementation along the lines of

```python
class MyIncompleteLoader(DataLoader):
    def __init__(self, dataset, batch_size, **kwargs):
        batch_size = min(batch_size - 5, 0) # five last batches are broken
        super().__init__(dataset, batch_size=batch_size, **kwargs)
```

And then will be very surprised, why their loader loads all the batches when they explicitly wanted to not load 5 last batches.

This basically ensures correctness in such a case. Sure, it might be slightly degenerate case, but I don't think it is too far-fetched

pytorch_lightning/utilities/data.py
otaj: So, code inspection found only one usage of this function in the code base and that is in the `_replace_dataloader_init_method`. That's why I thought it to be safe to change. However, I can absolutely change it back.

pytorch_lightning/utilities/data.py
otaj: You are right. During development, I had the condition outside of this dict comprehension and then I thought it looked neater this way.

pytorch_lightning/utilities/data.py
otaj: Well, sort of. `DataLoader` (the PyTorch one) does it and the class has to inherit from it. I can change it to `None`. Again, the only place where this is needed is to figure out the code path on this line https://github.com/PyTorchLightning/pytorch-lightning/pull/12981/files#diff-f21f92490e6c1417367a48324385cf66c908d35fb4321a4b1b20e5ba410830f4R251

otaj:
I think I replied to all of your comments @carmocca. I applied most of your suggestions, would you mind taking another look? Thanks!

pytorch_lightning/utilities/data.py
carmocca: Somebody might (maybe PyTorch geometric?) but I agree their authors should create the attribute themselves.

Still I would mention this explicitly as a breaking change

pytorch_lightning/utilities/data.py
```diff
@@ -336,14 +353,29 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
+        if not hasattr(obj, "_set_arg_names"):
+            set_arg_names = set()
+            for arg_name, arg_value in chain(zip(param_names, args), kwargs.items()):
+                if not hasattr(obj, "__" + arg_name):
```
carmocca, line 359: Could it happen that `hasattr(obj, "_set_arg_names") is True` but `hasattr(obj, "__" + arg_name) is False`?

As in, is this second if redundant?

pytorch_lightning/utilities/data.py
```diff
@@ -336,14 +353,29 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
+        if not hasattr(obj, "__dataset"):
+            # We have not found value for dataset yet, but we will find it eventually
+            # because it has to be passed to the original DataLoader
+            dataset_value = None
+            if "dataset" in param_names[: len(args)]:
+                dataset_value = args[param_names.index("dataset")]
+            elif "dataset" in kwargs:
+                dataset_value = kwargs["dataset"]
+            if dataset_value is not None:
+                setattr(obj, "__dataset", dataset_value)
```
carmocca, line 377: Why does this need to happen in this function?

It makes it coupled to dataloaders

otaj: I separated it dataloader specifics into its own function

pytorch_lightning/utilities/data.py
otaj: You're right, it was redundant, I removed it now

pytorch_lightning/utilities/data.py
otaj: Added a test case for it

pytorch_lightning/utilities/data.py
otaj: Mentioned it in breaking changes. Should I also update Changelog?

carmocca: Yes

pytorch_lightning/utilities/data.py
```diff
@@ -320,30 +338,44 @@ def _auto_add_worker_init_fn(dataloader: DataLoader, rank: int) -> None:
...
+
+        if not hasattr(obj, "__dataset"):
+            # We have not found value for dataset yet, but we will find it eventually
+            # because it has to be passed to the original DataLoader
+            dataset_value = None
+            if "dataset" in param_names:
+                dataset_value = args[param_names.index("dataset")]
+            elif "dataset" in kwargs:
+                dataset_value = kwargs["dataset"]
+            if dataset_value is not None:
```
carmocca, line 376: Edge case: if the actual value passed is None then this would break: `dataset=None`

You can fix it by duplicating the `setattr` inside the `if`s above

And add a test case for that!

carmocca (approved):
Super!

pytorch_lightning/utilities/data.py
otaj: Done

CHANGELOG.md
```diff
@@ -91,6 +91,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Raise an error if there are insufficient training batches when using a float value of `limit_train_batches` ([#12885](https://github.com/PyTorchLightning/pytorch-lightning/pull/12885))
 
 
+- `DataLoader` instantiated inside a `*_dataloader` hook will now not instantiate attributes for you automatically ([#12981](https://github.com/PyTorchLightning/pytorch-lightning/pull/12981))
```
carmocca, line 94:
```suggestion
- `DataLoader` instantiated inside a `*_dataloader` hook will not set the passed arguments as attributes anymore ([#12981](https://github.com/PyTorchLightning/pytorch-lightning/pull/12981))
```

pytorch_lightning/utilities/data.py
```diff
@@ -197,38 +196,71 @@ def _update_dataloader(
...
-    attrs = {k: v for k, v in vars(dataloader).items() if not k.startswith("_")}
-    # not part of `vars`
-    attrs["multiprocessing_context"] = dataloader.multiprocessing_context
+    was_wrapped = hasattr(dataloader, "__pl_dl_args")
+    if was_wrapped:
+        args = dataloader.__pl_dl_args
+        attrs = dataloader.__pl_dl_kwargs
+        arg_names = dataloader.__pl_dl_arg_names
+        used_args = min(len(args), len(arg_names))
+        for arg, arg_name in zip(args, arg_names):
```
carmocca, line 217: super nit: keys are usually in the left

```suggestion
        for arg_name, arg in zip(arg_names, args):
```

pytorch_lightning/utilities/data.py
```diff
@@ -197,38 +196,71 @@ def _update_dataloader(
...
+            attrs[arg_name] = arg
+
+        args = args[used_args:]
+        arg_names = arg_names[used_args:]
```
carmocca, line 221: Why don't we do all this logic in the wrapper?

pytorch_lightning/utilities/data.py
```diff
@@ -197,38 +196,71 @@ def _update_dataloader(
...
-        params.update(inspect.signature(DataLoader.__init__).parameters)
-        del params["self"]
 
-    # keep only the params whose default is different to the current attr value
-    non_defaults = {name for name, p in params.items() if name in attrs and p.default != attrs[name]}
-    # add `dataset` as it might have been replaced with `*args`
-    non_defaults.add("dataset")
+        if was_wrapped:
+            # if the dataloader was wrapped in a hook, only take arguments with default values
+            #  and assume user passes their kwargs correctly
```
carmocca, line 240:
```suggestion
            # and assume user passes their kwargs correctly
```

pytorch_lightning/utilities/data.py
```diff
@@ -197,38 +196,71 @@ def _update_dataloader(
...
+        # and hope we can get it from the instance attributes
+        original_dataset = None
+        # not part of `vars`
+        attrs["multiprocessing_context"] = dataloader.multiprocessing_context
 
     # get the dataloader instance `__init__` parameters
     params = dict(inspect.signature(dataloader.__init__).parameters)
     has_variadic_kwargs = any(p.kind is p.VAR_KEYWORD for p in params.values())
     if has_variadic_kwargs:
         # if the signature takes **kwargs, assume they will be passed down with `super().__init__(**kwargs)`
```
carmocca, line 236: This comment can go now?

pytorch_lightning/utilities/data.py
```diff
@@ -243,13 +275,13 @@ def _get_dataloader_init_kwargs(
     if required_args:
         required_args = sorted(required_args)
         dataloader_cls_name = dataloader.__class__.__name__
+        missing_args_message = ", ".join([f"`self.{arg_name}`" for arg_name in required_args])
```
carmocca, line 278: Can be a generator

```suggestion
        missing_args_message = ", ".join(f"`self.{arg_name}`" for arg_name in required_args)
```

pytorch_lightning/utilities/data.py
otaj: This comment was there originally, so I am not really sure about that one?

pytorch_lightning/utilities/data.py
otaj: Not there anymore

pytorch_lightning/utilities/data.py
```diff
@@ -320,30 +350,35 @@ def _auto_add_worker_init_fn(dataloader: DataLoader, rank: int) -> None:
...
         # We need to inspect `init`, as inspecting `obj.__init__`
         # can lead to inspecting the wrong function with multiple inheritance
         params = inspect.signature(init).parameters
-
         param_names = [
             param.name
             for param in params.values()
             if param.name != "self" and param.kind not in (param.VAR_POSITIONAL, param.VAR_KEYWORD)
         ]
+        param_names = tuple(param_names[: len(args)])
```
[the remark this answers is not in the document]
otaj: You're absolutely right, for some reason, I thought that slicing a tuple would not yield a tuple but list, but ti turns out it still is a tuple :)

pytorch_lightning/utilities/data.py
```diff
@@ -237,40 +260,41 @@ def _get_dataloader_init_kwargs(
...
-            f"`{dataloader_cls_name}(dataset, sampler=DistributedSampler(dataset))`."
+            f"The missing attributes are {required_args}. If you instantiate your `{dataloader_cls_name}` inside a "
+            "`*_dataloader` hook of your module, we will do this for you."
+            f" Otherwise, define {missing_args_message} inside your `__init__`."
         )
 
     if not has_variadic_kwargs:
         # the dataloader signature does not allow keyword arguments that need to be passed
-        missing_kwargs = dl_kwargs.keys() - params.keys()
+        missing_kwargs = (set(dl_kwargs.keys()) | set(arg_names)) - set(params.keys())
```
carmocca, line 283: Once the left operand is a set, it can be used with a dict keyview.

```suggestion
        missing_kwargs = (set(dl_kwargs) | set(arg_names)) - params.keys()
```

pytorch_lightning/utilities/data.py
```diff
@@ -406,16 +416,40 @@ def _replace_dataloader_init_method() -> Generator[None, None, None]:
...
-    dataset = dl_kwargs["dataset"]
+def _apply_fault_tolerant_automatic_capture_dataset_wrapper(dataset: Dataset) -> Dataset:
     if isinstance(dataset, IterableDataset):
         # wrap the `IterableDataset` into a `CaptureIterableDataset` to record sampler states.
-        dl_kwargs["dataset"] = CaptureIterableDataset(dataset=dataset)
+        return CaptureIterableDataset(dataset=dataset)
     elif get_len(dataset) != float("inf"):
-        dl_kwargs["dataset"] = CaptureMapDataset(dataset=dataset)
+        return CaptureMapDataset(dataset=dataset)
     else:
```
carmocca, line 425: This else can now be removed

tests/utilities/test_data.py
```diff
@@ -194,16 +194,16 @@ def __init__(self, attribute2, *args, **kwargs):
     assert dataloader.batch_size == 2
     assert dataloader.at1 == "attribute1"  # But the value still gets passed when it should
 
-    dataloader = _update_dataloader(dataloader, dataloader.sampler)
+    updated_dataloader = _update_dataloader(dataloader, dataloader.sampler)
```
carmocca, line 197: nit: is the name change necessary? you should try to minimize changes in long PRs like this one

pytorch_lightning/utilities/data.py
```diff
@@ -406,16 +416,40 @@ def _replace_dataloader_init_method() -> Generator[None, None, None]:
         del subclass._old_init
 
 
-def _apply_fault_tolerant_automatic_capture_dataset_wrapper(dl_kwargs: Dict) -> Dict:
-    dataset = dl_kwargs["dataset"]
+def _apply_fault_tolerant_automatic_capture_dataset_wrapper(dataset: Dataset) -> Dataset:
```
carmocca, line 419: Let's rename it to something simpler: `_wrap_with_capture_dataset`

pytorch_lightning/utilities/data.py
```diff
@@ -406,16 +416,40 @@ def _replace_dataloader_init_method() -> Generator[None, None, None]:
...
+        return CaptureIterableDataset(dataset=dataset)
     elif get_len(dataset) != float("inf"):
-        dl_kwargs["dataset"] = CaptureMapDataset(dataset=dataset)
+        return CaptureMapDataset(dataset=dataset)
     else:
         raise MisconfigurationException("This shouldn't happen, please open an issue on Lightning Github repository.")
-    return dl_kwargs
+
+
+def _apply_fault_tolerant_automatic_capture_dataset_wrapper_args_kwargs(
```
carmocca, line 429: Now this can keep the old name

pytorch_lightning/utilities/data.py
```diff
@@ -406,16 +416,40 @@ def _replace_dataloader_init_method() -> Generator[None, None, None]:
...
+        dl_kwargs["dataset"] = _apply_fault_tolerant_automatic_capture_dataset_wrapper(dl_kwargs["dataset"])
+    elif "dataset" in arg_names:
+        dataset_idx = arg_names.index("dataset")
+        dataset = _apply_fault_tolerant_automatic_capture_dataset_wrapper(dl_args[dataset_idx])
+        dl_args = dl_args[:dataset_idx] + (dataset,) + dl_args[dataset_idx + 1 :]
+    else:
+        if was_wrapped:
+            avoid_message = (
+                " To avoid this, either pass your dataset in kwargs or use argument name `dataset`"
+                " in signature of `__init__` method of your DataLoader."
```
carmocca, line 442: Let's be more explicit and mention passing `DataLoader(your_dataset)` or `DataLoader(dataset=your_dataset)`

pytorch_lightning/utilities/data.py
```diff
@@ -320,30 +354,35 @@ def _auto_add_worker_init_fn(dataloader: DataLoader, rank: int) -> None:
...
-        ]
-
-        cls = type(obj)
-        for arg_name, arg_value in chain(zip(param_names, args), kwargs.items()):
-            if hasattr(cls, arg_name) and getattr(cls, arg_name).fset is None:
-                # the class defines a read-only (no setter) property of this name. it's likely that the implementation
-                # will set `self._arg_name = arg_value` in `__init__` which is the attribute returned by the `arg_name`
-                # property so we are fine skipping in that case
-                continue
-            setattr(obj, arg_name, arg_value)
```
carmocca, line 346: Just noticed that since you removed this setattr, this PR also needed to remove these liens:

https://github.com/PyTorchLightning/pytorch-lightning/blob/3f78c4ca7a89fed8f79d9015e9510f0a0dcc46b5/pytorch_lightning/utilities/data.py#L393-L396

pytorch_lightning/utilities/data.py
otaj: Done

pytorch_lightning/utilities/data.py
otaj: Done

pytorch_lightning/utilities/data.py
otaj: Done

pytorch_lightning/utilities/data.py
otaj: Didn't really remove, just adapted to current codebase

tests/utilities/test_data.py
otaj: It isn't, but it is change only on the commit level, not on the PR level

pytorch_lightning/utilities/data.py
otaj: Are you sure about it? It was there even before and we are still passing the same dataset to it

carmocca: I was referring to https://pheanex.github.io/pylint/#:~:text=Unnecessary%20else%20after%20return%20.,them%20containing%20a%20return%20statement.

otaj: Oh, absolutely! I thought that it cannot happen anymore now! Thanks for the website, looks very useful. Wonder why precommit didn't catch it though

justusschock (approved):

PR #12383, opened 2022-03-21
logic for lock requirements

Borda:
## What does this PR do?

- Replace all latest with locked versions
- In the CI we augment all requirements files by adding max version

To prepare the latest lock, run:
```bash
python .actions/assistant.py create_locked_versions
```
This can be also automated, to run this generation nightly, and if there is a diff, send slack msg or create PR

### Does your PR introduce any breaking changes? If yes, please list them.

@awaelchli 

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @carmocca @akihironitta @borda

PR #12994, opened 2022-05-06
add freeze for development and full range for install

Borda:
## What does this PR do?

Closes #12383
set all max versions of today's latest released version and define dependabot to open PR when a new release appears...
in setup add logic for removing this upper bound unless it is commented in as strict

## Before submitting

## PR review

## Did you have fun?


.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+    labels:
+      - "dependencies"
+    pull-request-branch-name:
+      # Separate sections of the branch name with a hyphen
+      # for example, `dependabot-npm_and_yarn-next_js-acorn-6.4.1`
+      separator: "-"
+    # Allow up to 10 open pull requests for pip dependencies
+    open-pull-requests-limit: 10
+    reviewers:
+      - "octocat"
```
carmocca, line 23:
```suggestion
```

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+  # Enable version updates for python
+  - package-ecosystem: "pip"
+    # Look for a `requirements` in the `root` directory
+    directory: "/"
+    # Check for updates once a week
+    schedule:
+      interval: "weekly"
+    # Labels on pull requests for version updates only
+    labels:
+      - "dependencies"
```
carmocca, line 15:
```suggestion
      - "ci"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
+# Basic dependabot.yml file with
+# minimum configuration for two package managers
+
+version: 2
+updates:
+  # Enable version updates for python
+  - package-ecosystem: "pip"
+    # Look for a `requirements` in the `root` directory
+    directory: "/"
```
carmocca, line 9: Won't this only modify the root requirements.txt? We need to tell it to modify all

requirements/strategies.txt
```diff
@@ -1,4 +1,4 @@
-fairscale>=0.4.5
+fairscale>=0.4.5, <=0.4.6
 deepspeed<0.6.0
 horovod>=0.21.2,!=0.24.0  # no need to install with [pytorch] as pytorch is already installed
 hivemind>=1.0.1; sys_platform == 'linux'
```
carmocca, line 4: These have no max pin. On purpose?

Borda: probably would need to come with better reg expression, so let's first evaluate the core of this PR and then we can finish this last one :)

awaelchli (approved):
awesome!

pytorch_lightning/setup_tools.py
```diff
@@ -19,7 +19,9 @@
 _PROJECT_ROOT = os.path.dirname(os.path.dirname(__file__))
 
 
-def _load_requirements(path_dir: str, file_name: str = "requirements.txt", comment_char: str = "#") -> List[str]:
+def _load_requirements(
```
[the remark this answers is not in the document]
Borda: when you install PL as a python package our requirements listed in files need to be loaded and passed to the setup function... till now we just dropped all comments and HTTP links, but now we also remove all upper bound version limitations unless it is intended (the commend after package includes word `strict`)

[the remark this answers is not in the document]
awaelchli: Okay I see. So basically if the requirement says `x >= 1.5, <1.7`, we will install as if `x >= 1.5`. Sounds good! Do we actually have examples where the comment includes "strict"?

awaelchli (approved):

requirements/strategies.txt
Borda: fixed :rabbit:

pytorch_lightning/setup_tools.py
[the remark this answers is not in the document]
Borda: > Okay I see. So basically if the requirement says `x >= 1.5, <1.7`, we will install as if x >= 1.5

the pattern for replacement is `<=` not `<` as the dependabot would list the last feasible version

> Do we actually have examples where the comment includes "strict"?

not yet as we do not have such case yet :/

.github/dependabot.yml
Borda: well, seems there is only one folder, so moving `./requirements.txt` to `./requirements/base.txt` and keeping the original `./requirements.txt` ad link

requirements/base.txt
```diff
@@ -0,0 +1,10 @@
+numpy>=1.17.2, <=1.22.3
+torch>=1.8.*, <=1.11.0
+tqdm>=4.57.0, <=4.63.0
+PyYAML>=5.4, <=6.0
+fsspec[http]>=2021.05.0, !=2021.06.0, <=2022.2.0
+tensorboard>=2.2.0, <=2.8.0
+torchmetrics>=0.4.1, <=0.7.2
+pyDeprecate>=0.3.1, <0.4.0, <=0.3.2
```
akihironitta, line 8: This is now redundant? https://github.com/Borda/pyDeprecate/releases

```suggestion
pyDeprecate>=0.3.1, <=0.3.2
```

requirements/extra.txt
```diff
@@ -1,9 +1,9 @@
...
-matplotlib>3.1
-torchtext>=0.9.*
-omegaconf>=2.0.5
-hydra-core>=1.0.5
-jsonargparse[signatures]>=4.7.1
-gcsfs>=2021.5.0
-rich>=10.2.2,!=10.15.*
+matplotlib>3.1, <=3.5.1
+torchtext>=0.9.*, <=0.12.0
+omegaconf>=2.0.5, <=2.1.*
```
akihironitta, line 5: noob question: How is it different from specifying the exact version? e.g. `<=2.1.*` vs. `<=2.1.2`

Not sure why some use `.*` and others use exact versions.

akihironitta (commented):
The change in `setup_tools.py` is very interesting 🚀

requirements/extra.txt
Borda: `.*` includes all bugfixes...
`<=2.1.*` is the same as `<2.2`

akihironitta: Should all upper bounds be written that way then?

For example,
```diff
- matplotlib>3.1, <=3.5.1
- torchtext>=0.9.*, <=0.12.0
+ matplotlib>3.1, <=3.5.*
+ torchtext>=0.9.*, <=0.12.*
```

akihironitta: I think it's nothing critical, but I'm just curious why some are written so while others aren't.

Borda: we can convert them to a specific version :)

carmocca (approved):

.github/dependabot.yml
akihironitta: SGTM!

pytorch_lightning/setup_tools.py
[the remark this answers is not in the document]
akihironitta: This is nice! Please correct me if I'm wrong, but with this logic, we don't need to change anything in the requirements files (i.e., we don't need to unpin the upper-bounded versions) when releasing, right?

requirements/extra.txt
akihironitta: IMHO, we should have specific versions for all packages to further reduce the possibility of something getting broken.

pytorch_lightning/setup_tools.py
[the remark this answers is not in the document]
Borda: @kaushikb11 ^^

pytorch_lightning/setup_tools.py
```diff
@@ -31,12 +33,17 @@ def _load_requirements(path_dir: str, file_name: str = "requirements.txt", comme
...
         # skip directly installed dependencies
-        if ln.startswith("http") or "@http" in ln:
+        if not req or req.startswith("http") or "@http" in req:
             continue
-        if ln:  # if requirement is not empty
-            reqs.append(ln)
+        # remove version restrictions unless they are strict
+        if unfreeze and "<" in req and "strict" not in comment:
+            req = re.sub(r",? *<= *[\d\.\*]+", "", req).strip()
+            req = req[:-1] if req[-1] == "," else req
```
Borda, line 45:
```suggestion
```

pytorch_lightning/setup_tools.py
```diff
@@ -31,12 +33,16 @@ def _load_requirements(path_dir: str, file_name: str = "requirements.txt", comme
...
+        comment = ln[ln.index(comment_char) :] if comment_char in ln else ""
+        req = ln.strip()
         # skip directly installed dependencies
-        if ln.startswith("http") or "@http" in ln:
+        if not req or req.startswith("http") or "@http" in req:
             continue
-        if ln:  # if requirement is not empty
-            reqs.append(ln)
+        # remove version restrictions unless they are strict
+        if unfreeze and "<" in req and "strict" not in comment:
```
Borda, line 43:
```suggestion
        if unfreeze and "<=" in req and "strict" not in comment:
```

Borda (commented):
->

requirements/strategies.txt
```diff
@@ -1,4 +1,5 @@
-fairscale>=0.4.5
+fairscale>=0.4.5, <=0.4.6
 deepspeed<0.6.0
-horovod>=0.21.2,!=0.24.0  # no need to install with [pytorch] as pytorch is already installed
-hivemind>=1.0.1; sys_platform == 'linux'
+# no need to install with [pytorch] as pytorch is already installed
+horovod>=0.21.2,!=0.24.0, <=0.24.3
```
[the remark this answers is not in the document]
Borda: it was too long line, but yes, we can move it back...

PR #13014, opened 2022-05-09
HPU & TPU doesn't support torch.inference_mode

jerome-habana:
Signed-off-by: Jerome <janand@habana.ai>

## What does this PR do?

https://github.com/PyTorchLightning/pytorch-lightning/pull/12715/ introduced torch.inference_mode without checking all backends. This resulted in the following error on habana backends
"RuntimeError: Cannot set version_counter for inference tensor"

This PR disables inference_mode and uses no_grad mode for eval/predict stages with hpu backend. Once hpu impl adds support for inference_mode, this check will be removed.

Fixes #\<issue_number>

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #13019, opened 2022-05-09
Patch release v1.6.4

carmocca:
## What does this PR do?

As usual.


kaushikb11:
Need #13014 as well.

carmocca:
@kaushikb11 Inference mode is in 1.7 only: #12715

carmocca:
@Borda Quantization performance tests are failing in the LTS job. Do you think this is just flakiness?

Borda:
> @Borda Quantization performance tests are failing in the LTS job. Do you think this is just flakiness?

I do not see any reason for such failure so I believe it is just flakiness

rohitgr7:
is it still wIP?

carmocca:
The PR is ready but blocked by #13152 (which happens on master too)

jerome-habana:
@carmocca can you include this as well  #13067

requirements.txt
```diff
@@ -1,12 +1,2 @@
...
-torch>=1.8.*
-tqdm>=4.57.0
-PyYAML>=5.4
-fsspec[http]>=2021.05.0, !=2021.06.0
-tensorboard>=2.2.0
-torchmetrics>=0.4.1
-pyDeprecate>=0.3.1, <0.4.0
-packaging>=17.0
-typing-extensions>=4.0.0
+-r ./requirements/base.txt
```
awaelchli, line 2: This contains the pinned dependencies, which is probably undesired for the release

Borda: The pin shall be removed setup, but rather double check that also this is included

carmocca: How do we double-check it?

awaelchli: Oh right, I forgot that setup.py does this. 

> How do we double-check it?
I guess we would reduce one of the pinned version to a lower number, install lightning, then try to upgrade that package to the latest and observe whether pip complains about dependency resolution. Might be too much effort though.

Borda: Checked and it is there...

awaelchli (approved):

CHANGELOG.md
```diff
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 
+## [1.6.4] - 2022-05-10
```
carmocca, line 8: Self note: update the date

PR #13024, opened 2022-05-10
Add BaseModelCheckpoint class to inherit from

### LINKED ISSUE #6504: [RFC] Create a ModelCheckpointBase callback
## 🚀 Feature
Create a ModelCheckpointBase callback, and have the existing checkpoint callback extend it

### Motivation
The model checkpoint callback is growing in complexity. Features that have been recently added or will soon be proposed:
- Checkpoint every n train steps: #6146 
- Checkpointing using a time-based interval during training: #6286 
- Checkpointing on train end to fix this hack 
https://github.com/PyTorchLightning/pytorch-lightning/blob/680e83adab38c2d680b138bdc39d48fc35c0cb58/pytorch_lightning/trainer/training_loop.py#L152-L163

The decision was made in #6146 to keep these triggers mutually exclusive, at least based on the phase they run in. Why? It's very hard to get the state management right. For instance, the `monitor` might be added for something that's available only during validation, but the checkpoint callback is configured to run during training too, and crashes when it tries to look up the `monitor` key in the available metrics for tracking. Tracking top-K models and scores is another huge pain. Supporting multiple monitor metrics on top of this is another beast.

cc @borda @carmocca @awaelchli @ninginthecloud @jjenniferdai @rohitgr7 

### Pitch

Move the existing logic for the following into a base class:
- Core saving functionality
- Top-K model management
- formatting checkpoint names
- validation (though sub-classes should override this)

And have thin wrappers on top which extend this class and implement callback hook(s) for when to save the checkpoint.

### Alternatives

The checkpoint callback gets bigger and bigger as we add more features to it.

otaj:
## What does this PR do?

Introduces `BaseModelCheckpoint`, empty class for custom Checkpoint callbacks, that can be subclassed and still recognized in `Trainer`.

This is a prerequisite for future split of `ModelCheckpoint` into multiple classes each doing their own thing.

Fixes #6504

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7 @otaj @carmocca @jjenniferdai


rohitgr7 (approved):

Borda (approved):

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -45,7 +45,15 @@
 warning_cache = WarningCache()
 
 
-class ModelCheckpoint(Callback):
+class BaseModelCheckpoint(Callback):
+    r"""
+    This is the base class for Model checkpointing. Expert users may want to subclass it in case of writing
+    custom :class:`~pytorch_lightning.callbacks.model_checkpoint.ModelCheckpoint` callback, so that
+    the trainer recognizes the custom class as a checkpointing callback
```
awaelchli, line 52:
```suggestion
    the trainer recognizes the custom class as a checkpointing callback.
```

docs/source/common/checkpointing_expert.rst
```diff
@@ -87,3 +87,10 @@ Custom Checkpoint IO Plugin
 .. note::
 
     Some ``TrainingTypePlugins`` like ``DeepSpeedStrategy`` do not support custom ``CheckpointIO`` as checkpointing logic is not modifiable.
+
+
+*********************************
+Writing your own Checkpoint class
+*********************************
```
awaelchli, line 94: I would move this above the checkpoint io plugin. I predict that it is more likely or more intuitive that a user would want to override the callback first before looking at the checkpoin io api.

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -45,7 +45,15 @@
 warning_cache = WarningCache()
 
 
-class ModelCheckpoint(Callback):
+class BaseModelCheckpoint(Callback):
```
awaelchli, line 48: Just so that someone commented about it: 
We have gradually moved away from naming base classes with the "Base" as pre- or postfix. Here of course, we are in the situation where we want to create a new base class and can't rename the existing one. Adding Base is probably the best here. Have other names be proposed? I can't think of a better one, all alternatives sound silly or are inconsistent: CheckpointCallback, Checkpoint, AbstractModelCheckpoint.

awaelchli (approved):

pytorch_lightning/callbacks/model_checkpoint.py
otaj: I don't think other names have been proposed, but I agree with you that most of the alternatives sound a bit silly. The only one that sounds at least a bit reasonable is `AbstractModelCheckpoint` as it shows, that if you use just this without subclassing, you're not actually doing anything at all.

docs/source/common/checkpointing_expert.rst
otaj: Makes sense, moved

pytorch_lightning/callbacks/model_checkpoint.py
justusschock: I would only call it Abstract if it really is an abstract class (i.e. if an instance of this class cannot be created without subclassing. Otherwise i'd stick to base

docs/source/common/checkpointing_expert.rst
```diff
@@ -8,6 +8,13 @@ Checkpointing (expert)
 
 TODO: I don't understand this...
 
+*********************************
+Writing your own Checkpoint class
+*********************************
+
+We provide ``BaseModelCheckpoint`` class, for easier subclassing. Users may want to subclass it in case of writing custom ``ModelCheckpoint`` callback, so that the ``Trainer`` recognizes the custom class as a checkpointing callback.
```
carmocca, line 15: @RobertLaurella Can you double-check this sentence? It reads weird to me

RobertLaurella: I suggested a small change below. When the line reads "You may also want to subclass it", it may be worth specifying what "it" refers to (the BaseModelCheckpoint class?)

pytorch_lightning/callbacks/model_checkpoint.py
awaelchli: Actually, I like Checkpoint as the best option out of the selection we have. 
Checkpointer would be an outlier among all the other names (e.g. ModelSummary vs ModelSummarizer).

otaj: Yeah, I like Checkpoint too. Applied

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -45,7 +45,15 @@
 warning_cache = WarningCache()
 
 
-class ModelCheckpoint(Callback):
+class Checkpoint(Callback):
```
ananthsub, line 48: why not split this out into its own file?

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -259,18 +260,18 @@ def _attach_model_callbacks(self) -> None:
 
     @staticmethod
     def _reorder_callbacks(callbacks: List[Callback]) -> List[Callback]:
-        """Moves all ModelCheckpoint callbacks to the end of the list. The sequential order within the group of
+        """Moves all BaseModelCheckpoint callbacks to the end of the list. The sequential order within the group of
```
ananthsub, line 263:
```suggestion
        """Moves all Checkpoint callbacks to the end of the list. The sequential order within the group of
```

pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -259,18 +260,18 @@ def _attach_model_callbacks(self) -> None:
...
         checkpoint callbacks is preserved, as well as the order of all other callbacks.
 
         Args:
             callbacks: A list of callbacks.
 
         Return:
-            A new list in which the last elements are ModelCheckpoints if there were any present in the
+            A new list in which the last elements are BaseModelCheckpoints if there were any present in the
```
ananthsub, line 270:
```suggestion
            A new list in which the last elements are Checkpoint if there were any present in the
```

ananthsub (commented):
what about areas where the trainer makes an assumption of properties that are set on the ModelCheckpoint class but which are not available on the Checkpoint class.

For instance, the checks here: https://github.com/PyTorchLightning/pytorch-lightning/blob/29fe1dad155ee80e8934cf91004b4ea9cdbc455d/pytorch_lightning/trainer/trainer.py#L1416-L1441

Now if I subclass the `Checkpoint` class that does not add these attributes, passing `trainer.fit(model, ckpt_path="best")` will fail. 

Do we need to change those to be `hasattr` checks instead?

pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -45,7 +45,15 @@
...
+    r"""
+    This is the base class for Model checkpointing. Expert users may want to subclass it in case of writing
+    custom :class:`~pytorch_lightning.callbacks.model_checkpoint.ModelCheckpoint` callback, so that
+    the trainer recognizes the custom class as a checkpointing callback.
+    """
+
+
+class ModelCheckpoint(Checkpoint):
```
carmocca, line 56: `_FaultToleranceCheckpoint` should also be updated to subclass it

otaj:
> what about areas where the trainer makes an assumption of properties that are set on the ModelCheckpoint class but which are not available on the Checkpoint class.
> 
> For instance, the checks here:
> 
> https://github.com/PyTorchLightning/pytorch-lightning/blob/29fe1dad155ee80e8934cf91004b4ea9cdbc455d/pytorch_lightning/trainer/trainer.py#L1416-L1441
> 
> Now if I subclass the `Checkpoint` class that does not add these attributes, passing `trainer.fit(model, ckpt_path="best")` will fail.
> 
> Do we need to change those to be `hasattr` checks instead?

We absolutely have to, you're right! I'm gonna go through the codebase to see where are all the places we should be doing it.

pytorch_lightning/callbacks/model_checkpoint.py
otaj: It is now

pytorch_lightning/callbacks/model_checkpoint.py
otaj: It is now

otaj:
To be fair, I see this PR as a prelude for much larger PR which will:

* define a basic BaseCheckpoint class which will include all the basic functionality of checkpointing (saving, etc) (btw, don't focus on name, that's just a first idea)
* split out current ModelCheckpoint class into multiple classes, which will inherit from this BaseCheckpoint and rely on it for basic functionality, but they will each take care of functionally different things such as saving last checkpoint, saving best checkpoint, saving top-k checkpoints - basically untangle all the weird interactions arising from setting weird combinations of arguments to `ModelCheckpoint`

The plan isn't really laid out in stone yet and I'm absolutely open to any and all suggestions :)

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -8,6 +8,13 @@ Checkpointing (expert)
 
 TODO: I don't understand this...
 
+*********************************
+Writing your own Checkpoint class
+*********************************
+
+We provide ``Checkpoint`` class, for easier subclassing. Users may want to subclass this class in case of writing custom ``ModelCheckpoint`` callback, so that the ``Trainer`` recognizes the custom class as a checkpointing callback.
```
awaelchli, line 15: I don't think this fits here, because this page is about the CheckpointIO plugin and we don't want to confuse the two concepts I think. 

> so that the ``Trainer`` recognizes the custom class as a checkpointing callback.

This is not something the user cares about, this is just an implementation detail.

I propose to write the section with the main point being the recommendation of using the Checkpoint as the base class for highly custom checkpointing callbacks, and also provide a very simple example.

src/lightning_app/cli/pl-app-template/core/callbacks.py
```diff
@@ -291,7 +291,11 @@ def setup(
         self._collect_logger_metadata(trainer)
 
     def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
-        if trainer.checkpoint_callback and trainer.checkpoint_callback.dirpath is not None:
```
awaelchli, line 294: Let's drop this change so it's not blocked for merge.

src/pytorch_lightning/callbacks/checkpoint.py
```diff
@@ -0,0 +1,9 @@
+from pytorch_lightning.callbacks.callback import Callback
+
+
+class Checkpoint(Callback):
+    r"""
+    This is the base class for Model checkpointing. Expert users may want to subclass it in case of writing
+    custom :class:`~pytorch_lightning.callbacks.model_checkpoint.ModelCheckpoint` callback, so that
```
awaelchli, line 7: it wouldn't be a custom ModelCheckpoint though, because it doesn't subclass it.

src/pytorch_lightning/callbacks/checkpoint.py
```diff
@@ -0,0 +1,9 @@
+from pytorch_lightning.callbacks.callback import Callback
+
+
+class Checkpoint(Callback):
+    r"""
+    This is the base class for Model checkpointing. Expert users may want to subclass it in case of writing
```
awaelchli, line 6:
```suggestion
    This is the base class for model checkpointing. Expert users may want to subclass it in case of writing
```

src/pytorch_lightning/loggers/wandb.py
```diff
@@ -474,25 +479,33 @@ def finalize(self, status: str) -> None:
...
         # get checkpoints to be saved with associated score
-        checkpoints = {
-            checkpoint_callback.last_model_path: checkpoint_callback.current_score,
-            checkpoint_callback.best_model_path: checkpoint_callback.best_model_score,
-            **checkpoint_callback.best_k_models,
-        }
-        checkpoints = sorted((Path(p).stat().st_mtime, p, s) for p, s in checkpoints.items() if Path(p).is_file())
+        checkpoints = dict()
+        if hasattr(checkpoint_callback, "last_model_path") and hasattr(checkpoint_callback, "current_score"):
+            checkpoints[checkpoint_callback.last_model_path] = (checkpoint_callback.current_score, "latest")
```
awaelchli, line 486: Where do these strings "latest", "best" come from? This wasn't there before?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1406,11 +1407,15 @@ def __set_ckpt_path(self, ckpt_path: Optional[str], model_provided: bool, model_
                     f'`.{fn}(ckpt_path="best")` is set but `ModelCheckpoint` is not configured to save the best model.'
                 )
             # load best weights
-            ckpt_path = self.checkpoint_callback.best_model_path
+            ckpt_path = (
+                self.checkpoint_callback.best_model_path
+                if hasattr(self.checkpoint_callback, "best_model_path")
+                else None
+            )
```
awaelchli, line 1414:
```suggestion
            ckpt_path = getattr(self.checkpoint_callback, "best_model_path", None)
```

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1406,11 +1407,15 @@ def __set_ckpt_path(self, ckpt_path: Optional[str], model_provided: bool, model_
...
 
         if ckpt_path == "last":
-            candidates = [ft.ckpt_path for ft in ft_checkpoints] + [
-                cb.last_model_path for cb in self.checkpoint_callbacks
+            candidates = [ft.ckpt_path if hasattr(ft, "ckpt_path") else None for ft in ft_checkpoints] + [
+                cb.last_model_path if hasattr(cb, "last_model_path") else None for cb in self.checkpoint_callbacks
```
awaelchli, line 1418:
```suggestion
            candidates = [getattr(ft, "ckpt_path", None) for ft in ft_checkpoints] + [
                getattr(cb, "last_model_path", None) for cb in self.checkpoint_callbacks
```

otaj: Oh my, I feel dumb for using this antipattern. Thanks @awaelchli for pointing it out :)

src/pytorch_lightning/trainer/trainer.py
otaj: Oh my, I feel dumb for using this antipattern. Thanks @awaelchli for pointing it out :)

src/pytorch_lightning/loggers/wandb.py
otaj: It was, down the line, the variable `aliases`. I just thought it makes sense to generate them here

src/pytorch_lightning/callbacks/checkpoint.py
otaj: You're right, done

docs/source-pytorch/common/checkpointing_expert.rst
otaj: But, while I'm defending this particular sentence, I fully agree that the section could get a rewrite and maybe moved somewhere else :)

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -8,6 +8,13 @@ Checkpointing (expert)
 
 TODO: I don't understand this...
```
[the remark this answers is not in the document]
otaj: We can, though it isn't mine and it doesn't show up in rendered docs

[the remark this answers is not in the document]
carmocca: If you want to keep it, then it should stay above the section it previously referenced (the checkpointIO section)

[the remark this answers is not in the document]
otaj: You're right. Moved

Issue #11826, opened 2022-02-09
Move `_*_AVAILABLE` checks to their respective files.


carmocca:
## Proposed refactor

Move all the variables defined here that are only used in one file
https://github.com/PyTorchLightning/pytorch-lightning/blob/1e36cffbca3f92855d82907370544fade1656da6/pytorch_lightning/utilities/imports.py#L98-L134
to their respective files.

### Motivation

This encapsulates code and avoids clutter from the utilities.

### Pitch

Move:

https://github.com/Lightning-AI/lightning/blob/50fd12f841bc10ec093e80d0acf3e3b51cd927fe/src/pytorch_lightning/utilities/imports.py#L28-L51

To the files that use them.

The rest I don't think should be moved.
______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @borda

Borda:
@carmocca could you elaborate more on why this is needed?
I would say having import change in one place makes it easier to find and this `imports` is a primitive without any dependency so there is no risk of cyclic imports compared to your proposal...

carmocca:
My point is that if one of these primitives is used in only one file, then it should be defined there. For example `_JSONARGPARSE_AVAILABLE` is only used the CLI:

```python
$ grep -iIRn "_JSONARGPARSE_AVAILABLE" pytorch_lightning/
pytorch_lightning//utilities/__init__.py:44:    _JSONARGPARSE_AVAILABLE,
[...]

Issue #12786, opened 2022-04-17
Reduce the import time of pytorch_lightning


awaelchli:
## Proposed refactor

The current import time for the pytorch_lightning package on my machine is several seconds. There are some opportunities to improve this.

### Motivation

High import times have an impact on the development and debugging speed. 

### Benchmark

I benchmarked the import time in two environments:
1. Fresh environment with pytorch lightning installed, no extras.
2. My currrent environment, with many extras installed such as loggers, horovod, etc.

To measure the import time, I created a simple file which only imports pytorch_lightning:

```py
import pytorch_lightning as pl
```

Then I use the `importtime` package to measure the time and create a profile:

```
python -X importtime simple.py 2> import.log
```

Finally, I used tuna to visualize the profile:

```
tuna import.log
```

For the fresh environment, the total import time is <2 secs with the following profile:

<img width="1466" alt="image" src="https://user-images.githubusercontent.com/5495193/163724305-4f927dbd-cd52-4273-a37b-89782fd28104.png">

```
>  pip freeze  | grep torch                                            (clean-pl-env) 

pytorch-lightning==1.6.1
torch==1.11.0
torchmetrics==0.7.3
```

For a full development environment, the total import time is >4 seconds:

<img width="1466" alt="image" src="https://user-images.githubusercontent.com/5495193/163723996-8f600c91-0d0a-467c-a2d5-e54a73bbac36.png">

The times vary a bit between multiple runs. However, I have observed that the time is consistently higher when running in an environment with extras installed. Looking at the profiles, it looks like the origin of a large waste of time is coming from our 

`pytorch_lightning.utilities.imports` module where we evaluate some constants at import time:

https://github.com/PyTorchLightning/pytorch-lightning/blob/ae3226ced96e2bc7e62f298d532aaf2290e6ef34/pytorch_lightning/utilities/imports.py#L98-L124

It looks like if a 3rd party package is installed and takes a long time to import, this time gets added to our loading time as well, even if the package never ends up being used. This is because our `_module_available` and `_package_available` implementations attempt to [import the modules](https://github.com/PyTorchLightning/pytorch-lightning/blob/ae3226ced96e2bc7e62f298d532aaf2290e6ef34/pytorch_lightning/utilities/imports.py#L56) to check their availability. This can be very costly. 

### Pitch

Evaluate the import checks lazily. 
Convert 

```py
_X_AVAILALBE = _module_available("x")
```

to

```py
@lru_cache()
[...]

PR #13035, opened 2022-05-11
Improved the jsonargparse[signatures] availability variable

mauvilsa:
## What does this PR do?

This PR addresses https://github.com/PyTorchLightning/pytorch-lightning/pull/12918#discussion_r862050327. Also saw #11826 so I moved the code to cli.py.

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/pytorch_lightning/utilities/imports.py
+++ b/pytorch_lightning/utilities/imports.py
@@ -85,6 +85,25 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
     return op(pkg_version, Version(version))
 
 
+class _RequirementAvaliable:
+    """Boolean-like class for check of requirement with extras and version specifiers."""
+
+    def __init__(self, requirement: str):
+        self.available = True
+        self.message = ""
+        try:
+            pkg_resources.require([requirement])
+        except Exception as ex:
+            self.available = False
+            self.message = f"Requirement '{requirement}' not met, {ex.__class__.__name__}: {ex}"
+
+    def __bool__(self):
+        return self.available
+
+    def __str__(self):
+        return self.message
+
+
 _IS_WINDOWS = platform.system() == "Windows"
 _IS_INTERACTIVE = hasattr(sys, "ps1")  # https://stackoverflow.com/a/64523765
 _PYTHON_GREATER_EQUAL_3_8_0 = Version(platform.python_version()) >= Version("3.8.0")
@@ -100,7 +119,6 @@ _BAGUA_AVAILABLE = _package_available("bagua")
 _DEEPSPEED_AVAILABLE = _package_available("deepspeed")
 _DEEPSPEED_GREATER_EQUAL_0_5_9 = _DEEPSPEED_AVAILABLE and _compare_version("deepspeed", operator.ge, "0.5.9")
 _DEEPSPEED_GREATER_EQUAL_0_6 = _DEEPSPEED_AVAILABLE and _compare_version("deepspeed", operator.ge, "0.6.0")
-_DOCSTRING_PARSER_AVAILABLE = _package_available("docstring_parser")
 _FAIRSCALE_AVAILABLE = not _IS_WINDOWS and _module_available("fairscale.nn")
 _FAIRSCALE_OSS_FP16_BROADCAST_AVAILABLE = _FAIRSCALE_AVAILABLE and _compare_version("fairscale", operator.ge, "0.3.3")
 _FAIRSCALE_FULLY_SHARDED_AVAILABLE = _FAIRSCALE_AVAILABLE and _compare_version("fairscale", operator.ge, "0.3.4")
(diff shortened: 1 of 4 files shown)


pytorch_lightning/utilities/imports.py
```diff
@@ -85,6 +85,25 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
     return op(pkg_version, Version(version))
 
 
+class _RequirementAvaliable:
```
carmocca, line 88: Typo:

```suggestion
class _RequirementAvailable:
```

pytorch_lightning/utilities/imports.py
```diff
@@ -85,6 +85,25 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
...
+    """Boolean-like class for check of requirement with extras and version specifiers."""
+
+    def __init__(self, requirement: str):
```
carmocca, line 91:
```suggestion
    def __init__(self, requirement: str) -> None:
```

pytorch_lightning/utilities/imports.py
```diff
@@ -85,6 +85,25 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
     return op(pkg_version, Version(version))
 
 
+class _RequirementAvaliable:
+    """Boolean-like class for check of requirement with extras and version specifiers."""
```
carmocca, line 89: Cool!

carmocca (approved):

pytorch_lightning/utilities/imports.py
```diff
@@ -85,6 +85,25 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
...
+
+    def __init__(self, requirement: str):
+        self.available = True
+        self.message = ""
+        try:
+            pkg_resources.require([requirement])
+        except Exception as ex:
+            self.available = False
+            self.message = f"Requirement '{requirement}' not met, {ex.__class__.__name__}: {ex}"
```
rohitgr7, line 98:
```suggestion
            self.message = f"Requirement {requirement!r} not met, {ex.__class__.__name__}: {ex}"
```

rohitgr7, line 98: just wondering if we should check this inside `__bool__` only once. then it will be checked only when it's required thus might reduce the import time #12786

rohitgr7 (approved):

pytorch_lightning/utilities/imports.py
carmocca: Good idea!

rohitgr7: >  where this class is instantiated the object is used right after

not all the time. It will still by imported even when user is not using LightningCli.
but anyways is fine, do it now or later for that issue.

pytorch_lightning/utilities/imports.py
Borda: it is a replacement of `_compare_version` and what is the advantage of class over function?

mauvilsa: This is much more than `_compare_version`. It checks an entire dependency, including the package, install+extras requires and version specifiers. Could be used to replace pairs of `_package_available` and `_compare_version` or even `_module_available` if in some case it is implied by an extra require.

The benefit of having a class is that it is not just a boolean. It also gives a clear message saying why the dependency is not met. This is being used in `LightningCLI` to show to the user.

pytorch_lightning/utilities/imports.py
```diff
@@ -85,24 +85,44 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
...
+    def _check_requirement(self) -> None:
+        if self.available is None:
+            try:
+                pkg_resources.require([self.requirement])
+            except Exception as ex:
+                self.available = False
+                self.message = f"Requirement {self.requirement!r} not met, {ex.__class__.__name__}: {ex}"
+            else:
+                self.available = True
+                self.message = f"Requirement {self.requirement!r} met"
```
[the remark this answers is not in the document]
mauvilsa: You are right, changed it.

pytorch_lightning/utilities/imports.py
```diff
@@ -85,6 +85,42 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
...
+    False
+    """
+
+    def __init__(self, requirement: str) -> None:
+        self.requirement = requirement
+
+    def _check_requirement(self) -> None:
+        if not hasattr(self, "available"):
+            try:
+                pkg_resources.require([self.requirement])
```
carmocca, line 105: Docs for this say:

> require(*requirements)
> [...]
> requirements must be a string or a (possibly-nested) sequence thereof, specifying the distributions and versions required.

https://setuptools.pypa.io/en/latest/pkg_resources.html

So shouldn't it be:

```suggestion
                pkg_resources.require(self.requirement)
```

akihironitta (commented):
🚀

pytorch_lightning/utilities/imports.py
```diff
@@ -85,6 +85,42 @@ def _compare_version(package: str, op: Callable, version: str, use_base_version:
     return op(pkg_version, Version(version))
 
 
+class _RequirementAvailable:
```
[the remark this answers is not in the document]
awaelchli: See also #12786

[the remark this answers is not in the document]
mauvilsa: Could be, but it would be `_RequirementAvailable("deepspeed>=0.5.9")`.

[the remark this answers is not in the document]
mauvilsa: Anyway I think that is something for a different pull request.

PR #13113, opened 2022-05-20
Avoid annoying firewall message from `find_free_network_port`

awaelchli:
## What does this PR do?

For years, we have learned to ignored the annoying firewall message that pops up when we run for example ddp, tests locally on CPU. **No more.** The `sock.listen(1)` call in the `find_free_network_port` utility function is causing this, but the call itself seems to be redundant after all. 

image

No.

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub

### DIFF
--- a/pytorch_lightning/plugins/environments/lightning_environment.py
+++ b/pytorch_lightning/plugins/environments/lightning_environment.py
@@ -96,7 +96,6 @@ def find_free_network_port() -> int:
     """
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     s.bind(("", 0))
-    s.listen(1)
     port = s.getsockname()[1]
     s.close()
     return port


pytorch_lightning/plugins/environments/lightning_environment.py
```diff
@@ -96,7 +96,6 @@ def find_free_network_port() -> int:
     """
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
     s.bind(("", 0))
-    s.listen(1)
```
justusschock, line 99: So this is redundant because we don't want to actually listen here but just create the socket to find a free port?

justusschock (approved):

pytorch_lightning/plugins/environments/lightning_environment.py
awaelchli: Yes, I'm basing this around the answer here https://stackoverflow.com/questions/1365265/on-localhost-how-do-i-pick-a-free-port-number
I saw the listen(1) call in one example but couldn't figure out why it is there.

justusschock: as far as I understood this is only necessary if this socket is actually going to receive some info which is nothing we handle there :)

kaushikb11 (approved):

carmocca (approved):

Issue #13102, opened 2022-05-18
MPS (Mac M1) device support


carmocca:
## 🚀 Feature

https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/
Docs: https://pytorch.org/docs/master/notes/mps.html

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @borda @akihironitta @rohitgr7 @justusschock

carmocca:
I head @awaelchli just got an M1 :smiling_imp:

awaelchli:
@justusschock also has one if I remember correctly 😃

justusschock:
I'll investigate on the weekend

carmocca:
After a bit of offline discussion, we thought about this API:

- The `MPSAccelerator` is introduced.
- The `GPUAccelerator` is deprecated and renamed in favor of `CUDAAccelerator`
- 2 new accelerator options: `Trainer(accelerator="cuda"|"mps")` to explicitly choose one of the above.
- `Trainer(accelerator="gpu")` now chooses from the 2 accelerators above based on the available hardware. (side note: this is how we would also support ROCm).
- `Trainer(accelerator="auto")` is different in that it also considers IPUs, TPUs, HPUs...

This shouldn't be a problem because a machine cannot have both accelerators available.

eubinecto:
Just when I thought I wanted to raise an issue! I'm so rooting for this.

ananthsub:
[...]

PR #13123, opened 2022-05-21
MPS Accelerator

justusschock:
## What does this PR do?

Adds a prototype of M1 GPU support via PyTorch's MPS backend (part of #13102 , renaming existing accelerators will be done separately)

To give it a try, install PL from this branch and change the accelerator to `'mps'`

## Remaining ToDos:
- [x] Extend Tests
- [x] Docs
- [ ] Getting tests to pass on all platforms
- [x] Move `_MPS_AVAILABLE` to accelerator file
- [x] Integrate MPS into GPU debug prints

Nope

## Before submitting

## PR review

## Did you have fun?

cc @borda @akihironitta @rohitgr7 @justusschock


tests/accelerators/test_accelerator_connector.py
```diff
@@ -291,7 +291,7 @@ def test_accelererator_invalid_type_devices(mock_is_available, mock_device_count
         _ = Trainer(accelerator="gpu", devices=device_count)
 
 
-@RunIf(min_gpus=1)
+@RunIf(min_cuda_gpus=1)
```
Borda, line 294: maybe make this rename as separate PR so it minimizes this change?

justusschock: Can do, but to me it is strongly coupled to this PR. Let me first finish here and then I can look into splitting :)

docs/source/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,41 @@
+:orphan:
+
+.. _mps_basic:
+
+MPS training (basic)
```
kaushikb11, line 5: I would suggest Apple or M1 in the title to make it easier for users to discover it.

justusschock: done :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -1746,6 +1754,7 @@ def __setup_profiler(self) -> None:
         self.profiler.setup(stage=self.state.fn._setup_fn, local_rank=local_rank, log_dir=self.log_dir)
 
     def _log_device_info(self) -> None:
+        # TODO(justusschock): should this include the nvidia gpus only or mps as well?
```
carmocca, line 1757: IMO MPS as well

Just need to change `torch.cuda.is_available` for `GPUAccelerator.is_available()`

Where `GPUAccelerator` is something like:

```python
if MPSAccelerator.is_available():
    GPUAccelerator = MPSAccelerator
else:
    GPUAccelerator = CUDAccelerator
```

And perhaps include a "type: mps | cuda" in the message

pytorch_lightning/utilities/device_parser.py
```diff
@@ -199,15 +220,37 @@ def _normalize_parse_gpu_input_to_list(gpus: Union[int, List[int], Tuple[int, ..
...
+
+def _get_all_available_mps_gpus() -> List[int]:
+    """
+    Returns:
+        a list of all available MPS gpus
+    """
+    if _MPS_AVAILABLE:
+        return [0]
+    else:
+        return []
```
carmocca, line 247: nit: 
```suggestion
    return [0] if _MPS_AVAILABLE else []
```

pytorch_lightning/accelerators/mps.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @staticmethod
+    def auto_device_count() -> int:
+        """Get the devices when set to auto."""
+        return 1
+
+    @staticmethod
+    def is_available() -> bool:
+        """MPS is only available for certain torch builds starting at torch>=1.12."""
+        return _MPS_AVAILABLE
```
carmocca, line 64: I would inline the _MPS_AVAILABLE flag here

```suggestion
    if _TORCH_GREATER_EQUAL_1_12:
        return torch.backends.mps.is_available()
    return False
```

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -566,11 +570,13 @@ def _choose_strategy(self) -> Union[Strategy, str]:
         if self._num_nodes_flag > 1:
             return DDPStrategy.strategy_name
         if len(self._parallel_devices) <= 1:
-            device = (
-                device_parser.determine_root_gpu_device(self._parallel_devices)  # type: ignore
-                if self._accelerator_flag == "gpu"
-                else "cpu"
-            )
+            # TODO: Change this once gpu accelerator was renamed to cuda accelerator
```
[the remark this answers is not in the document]
justusschock: I don’t think so. I think it’s easier to first introduce this PR and then just reroute the accelerator selection for gpu when both accelerators are present.

pytorch_lightning/trainer/trainer.py
justusschock: currently this is separate, but can easily change this :)

pytorch_lightning/accelerators/mps.py
carmocca: You would still have a single source of truth with this change, but inside the accelerator itself, not in a global variable

justusschock: that way we do the same check multiple times

pytorch_lightning/utilities/imports.py
```diff
@@ -181,6 +181,11 @@ def __repr__(self) -> str:
 else:
     _HPU_AVAILABLE = False
 
+if _TORCH_GREATER_EQUAL_1_12:
+    _MPS_AVAILABLE = torch.backends.mps.is_available()
+else:
+    _MPS_AVAILABLE = False
```
carmocca, line 187: Can we put this in the MPS accelerator file directly? https://github.com/PyTorchLightning/pytorch-lightning/issues/11826

pytorch_lightning/accelerators/mps.py
```diff
@@ -0,0 +1,93 @@
...
+        parsed_devices = device_parser.parse_gpu_ids(devices, include_mps=True)
+        return parsed_devices
+
+    @staticmethod
+    def get_parallel_devices(devices: Union[int, str, List[int]]) -> List[torch.device]:
+        """Gets parallel devices for the Accelerator."""
+        parsed_devices = device_parser.parse_gpu_ids(devices, include_mps=True)
+        assert parsed_devices is not None
+
+        return [torch.device("mps")] * len(parsed_devices)
```
kaushikb11, line 54: But you mentioned in the docs, that it supports one device at a time?

justusschock: yes, and `parse_gpu_ids for mps always only returns a single device. This is implemented that way, that we only need to change the device parsing logic if there will ever be a machine with multiple gpus (which I assume is not that unlikely given the still missing pro setup fro the mac line).

justusschock:
This now depends on #11935 to move the device back to cpu :)

docs/source/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,41 @@
+:orphan:
+
+.. _mps_basic:
```
SeanNaren, line 3: Shall we include a warning that this is still experimental and further work within PyTorch is happening?

SeanNaren (approved):
🚀 LGTM great work @justusschock! I assume you've run the CI manually locally to verify it works :)

docs/source/accelerators/mps_basic.rst
justusschock: yes we should! Good point

justusschock:
yes, for me everything is passing :)

docs/source/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,49 @@
...
+MPS training (basic)
+====================
+**Audience:** Users looking to train on their Apple silicon GPUs.
+
+.. warning::
+
+   The MPS accelerator as well as the PyTorch backend are still very experimental.
+   So far not all operations are supported, but more ops are coming every day due to development from the PyTorch Team.
+   You can use ``PYTORCH_ENABLE_MPS_FALLBACK=1 python your_script.py`` to fall back to cpu for unsupported operations.
+   We will do our very best to keep this accelerator up and running, but so far were unable to find a suitable CI provider making this a lot of manual work.
```
carmocca, line 14: I would not include this line in the docs

docs/source/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,49 @@
+:orphan:
```
carmocca, line 1: cc @RobertLaurella to check out this file :smile:

pytorch_lightning/accelerators/mps.py
```diff
@@ -0,0 +1,98 @@
...
+        return get_device_stats()
+
+    @staticmethod
+    def parse_devices(devices: Union[int, str, List[int]]) -> Optional[List[int]]:
+        """Accelerator device parsing logic."""
+        parsed_devices = device_parser.parse_gpu_ids(devices, include_mps=True)
+        return parsed_devices
+
+    @staticmethod
+    def get_parallel_devices(devices: Union[int, str, List[int]]) -> List[torch.device]:
```
carmocca, line 54: This method does not return an optional but `patse_devices` does. I think they should both do it or not do it for consistency

pytorch_lightning/accelerators/mps.py
```diff
@@ -0,0 +1,98 @@
...
+        """Gets parallel devices for the Accelerator."""
+        parsed_devices = device_parser.parse_gpu_ids(devices, include_mps=True)
```
carmocca, line 56: Could call `MPSAccelerator.parse_devices(devices)`

pytorch_lightning/trainer/trainer.py
```diff
@@ -1745,9 +1752,20 @@ def __setup_profiler(self) -> None:
...
-            f"GPU available: {torch.cuda.is_available()}, used: {isinstance(self.accelerator, GPUAccelerator)}"
-        )
+
+        if GPUAccelerator.is_available():
+            gpu_available = True
+            gpu_type = " (cuda)"
+        elif MPSAccelerator.is_available():
+            gpu_available = True
+            gpu_type = " (mps)"
+
```
carmocca, line 1762: nit: empty line

```suggestion
```

pytorch_lightning/utilities/device_parser.py
```diff
@@ -199,15 +219,37 @@ def _normalize_parse_gpu_input_to_list(gpus: Union[int, List[int], Tuple[int, ..
...
 
-def _get_all_available_gpus() -> List[int]:
+def _get_all_available_gpus(include_cuda: bool = False, include_mps: bool = False) -> List[int]:
+    """
+    Returns:
+        a list of all available gpus
+    """
+    cuda_gpus = _get_all_available_cuda_gpus() if include_cuda else []
+    mps_gpus = _get_all_available_mps_gpus() if include_mps else []
+    # This is where in the future ROCM parsing for GPUs should also be called
```
carmocca, line 234: I don't think we should add this type of comment. So easy for it to become stale

tests/callbacks/test_tqdm_progress_bar.py
```diff
@@ -29,9 +29,15 @@
...
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.imports import _TORCH_GREATER_EQUAL_1_12
 from tests.helpers.boring_model import BoringModel, RandomDataset
 from tests.helpers.runif import RunIf
 
+if _TORCH_GREATER_EQUAL_1_12:
+    torch_test_assert_close = torch.testing.assert_close
+else:
+    torch_test_assert_close = torch.testing.assert_allclose
+
```
carmocca, line 40: Interesting, why was this necessary?

tests/helpers/runif.py
```diff
@@ -103,6 +105,7 @@ def __new__(
             tpu: Require that TPU is available.
             ipu: Require that IPU is available.
             hpu: Require that HPU is available.
+            mps: Require that MPS (M1 Metal GPU) is available or not available.
```
carmocca, line 108: For consistency. Also let's not mention M1 as there will likely be an M2 and beyond

```suggestion
            mps: Require that MPS (Apple Sillicon) is available.
```

carmocca (approved):

pytorch_lightning/utilities/device_parser.py
justusschock: will remove it :)

pytorch_lightning/accelerators/mps.py
justusschock: I just sticked to the interface the other accelerators had. If I modify the typing here, wouldn't that violate liskov substitution principle?

pytorch_lightning/utilities/device_parser.py
```diff
@@ -181,7 +197,9 @@ def _sanitize_gpu_ids(gpus: List[int]) -> List[int]:
         MisconfigurationException:
             If machine has fewer available GPUs than requested.
     """
-    all_available_gpus = _get_all_available_gpus()
+    if not sum([int(include_cuda), int(include_mps)]) > 0:
```
akihironitta, line 200: nit: Could be simpler?
```suggestion
    if sum((include_cuda, include_mps)) == 0:
```

tests/plugins/test_amp_plugins.py
```diff
@@ -21,9 +21,15 @@
...
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.imports import _TORCH_GREATER_EQUAL_1_12
 from tests.helpers import BoringModel
 from tests.helpers.runif import RunIf
 
+if _TORCH_GREATER_EQUAL_1_12:
+    torch_test_assert_close = torch.testing.assert_close
+else:
+    torch_test_assert_close = torch.testing.assert_allclose
+
```
akihironitta, line 32: This seems unnecessary anymore?
```suggestion
```

pytorch_lightning/accelerators/mps.py
carmocca: I don't see why. The base indicates `Any` anyways.

Feel free to resolve it as my comment is more about why does https://github.com/PyTorchLightning/pytorch-lightning/blob/dd475183227644a8d22dca3deb18c99fb0a9b2c4/pytorch_lightning/utilities/device_parser.py#L88-L90 return `None` instead of `[]`, which would remove the assert you had to add in the MPSAccelerator

akihironitta (approved):
This is amazing! 🚀 Ran a simple CIFAR10 training script locally and confirmed that it speeds up training! LGTM!

---

Haven't looked into these failures further yet, but for now, let me share that some tests failing on my M1 machine.

@justusschock Have you seen these failures on yours?

<details><summary> pytest failures </summary>

```
========================================================================================================================== FAILURES ===========================================================================================================================
___________________________________________________________________________________________________________________ test_callbacks_restore ____________________________________________________________________________________________________________________

tmpdir = local('/private/var/folders/pj/kh01cnqs2bgchr2wl2q_7s3c0000gn/T/pytest-of-nitta/pytest-13/test_callbacks_restore0')

    def test_callbacks_restore(tmpdir):
        """Test callbacks restore is called after optimizers have been re-created but before optimizer states
        reload."""
        chk = ModelCheckpoint(dirpath=tmpdir, save_last=True)

        model = FinetuningBoringModel()
        callback = TestCallbacksRestoreCallback()

        trainer_kwargs = dict(
            default_root_dir=tmpdir, limit_train_batches=1, limit_val_batches=1, callbacks=[callback, chk], max_epochs=2
        )

        trainer = Trainer(**trainer_kwargs)
        trainer.fit(model)

        # only 1 optimizer
        assert len(callback._internal_optimizer_metadata) == 1

        # only 2 param groups
        assert len(callback._internal_optimizer_metadata[0]) == 2

        # original parameters
        expected = {
            "lr": 0.1,
            "momentum": 0,
            "dampening": 0,
            "weight_decay": 0,
            "nesterov": False,
            "params": ["layer.3.weight", "layer.3.bias"],
        }
        if _TORCH_GREATER_EQUAL_1_11:
            expected["maximize"] = False

>       assert callback._internal_optimizer_metadata[0][0] == expected
E       AssertionError: assert {'dampening': 0,\n 'foreach': None,\n 'lr': 0.1,\n 'maximize': False,\n 'momentum': 0,\n 'nesterov': False,\n 'params': ['layer.3.weight', 'layer.3.bias'],\n 'weight_decay': 0} == {'dampening': 0,\n 'lr': 0.1,\n 'maximize': False,\n 'momentum': 0,\n 'nesterov': False,\n 'params': ['layer.3.weight', 'layer.3.bias'],\n 'weight_decay': 0}
E         Common items:
E         {'dampening': 0,
E          'lr': 0.1,
E          'maximize': False,
E          'momentum': 0,
E          'nesterov': False,
E          'params': ['layer.3.weight', 'layer.3.bias'],
E          'weight_decay': 0}
E         Left contains 1 more item:
E         {'foreach': None}
E         Full diff:
E           {
E            'dampening': 0,
E         +  'foreach': None,
E            'lr': 0.1,
E            'maximize': False,
E            'momentum': 0,
E            'nesterov': False,
E            'params': ['layer.3.weight',
E                       'layer.3.bias'],
E            'weight_decay': 0,
E           }

tests/callbacks/test_finetuning_callback.py:372: AssertionError
-------------------------------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------------------------------
Epoch 1: 100%|██████████| 2/2 [00:00<00:00, 525.21it/s, loss=1.01, v_num=0]
-------------------------------------------------------------------------------------------------------------------- Captured stderr call ---------------------------------------------------------------------------------------------------------------------
GPU available: True (mps), used: False
HPU available: False, using: 0 HPUs
`Trainer(limit_train_batches=1)` was configured so 1 batch per epoch will be used.
`Trainer(limit_val_batches=1)` was configured so 1 batch will be used.
Missing logger folder: /private/var/folders/pj/kh01cnqs2bgchr2wl2q_7s3c0000gn/T/pytest-of-nitta/pytest-13/test_callbacks_restore0/lightning_logs

  | Name  | Type       | Params
-------------------------------------
0 | layer | Sequential | 3.2 K
-------------------------------------
66        Trainable params
3.2 K     Non-trainable params
3.2 K     Total params
0.013     Total estimated model params size (MB)
____________________________________________________________________________________________________________________ test_proper_refcount _____________________________________________________________________________________________________________________

    def test_proper_refcount():
        torch_module = nn.Module()
        lightning_module = LightningModule()

>       assert sys.getrefcount(torch_module) == sys.getrefcount(lightning_module)
E       assert 2 == 3
E        +  where 2 = <built-in function getrefcount>(Module())
E        +    where <built-in function getrefcount> = sys.getrefcount
E        +  and   3 = <built-in function getrefcount>(LightningModule())
E        +    where <built-in function getrefcount> = sys.getrefcount

tests/core/test_lightning_module.py:415: AssertionError
------------------------------------------------------------------------------------------------------------------ Captured stdout teardown -------------------------------------------------------------------------------------------------------------------
Epoch 0:   0%|          | 0/2 [00:00<?, ?it/s]
Epoch 0:   0%|          | 0/2 [00:00<?, ?it/s]
___________________________________________________________________________________________________________ test_trainer_num_prefetch_batches[mps] ____________________________________________________________________________________________________________

tmpdir = local('/private/var/folders/pj/kh01cnqs2bgchr2wl2q_7s3c0000gn/T/pytest-of-nitta/pytest-13/test_trainer_num_prefetch_batc3'), accelerator = 'mps'

    @pytest.mark.flaky(reruns=3)
    @pytest.mark.parametrize(
        "accelerator", [pytest.param("gpu", marks=RunIf(min_cuda_gpus=1)), pytest.param("mps", marks=RunIf(mps=True))]
    )
    def test_trainer_num_prefetch_batches(tmpdir, accelerator):

>       model = RecommenderModel()

/Users/nitta/work/github.com/PyTorchLightning/pytorch-lightning/tests/utilities/test_fetching.py:202:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/Users/nitta/work/github.com/PyTorchLightning/pytorch-lightning/tests/utilities/test_fetching.py:167: in __init__
    self.CYCLES_PER_MS = int(get_cycles_per_ms())
/Users/nitta/work/github.com/PyTorchLightning/pytorch-lightning/tests/utilities/test_fetching.py:142: in get_cycles_per_ms
    vals.append(measure())
/Users/nitta/work/github.com/PyTorchLightning/pytorch-lightning/tests/utilities/test_fetching.py:130: in measure
    start = torch.cuda.Event(enable_timing=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'torch.cuda.streams.Event'>, enable_timing = True, blocking = False, interprocess = False

    def __new__(cls, enable_timing=False, blocking=False, interprocess=False):
>       return super(Event, cls).__new__(
            cls,
            enable_timing=enable_timing, blocking=blocking, interprocess=interprocess)
E       TypeError: object.__new__() takes exactly one argument (the type to instantiate)

/Users/nitta/.miniconda3/envs/dev39_nightly/lib/python3.9/site-packages/torch/cuda/streams.py:164: TypeError
=================================================================================================================== short test summary info ===================================================================================================================
FAILED tests/callbacks/test_finetuning_callback.py::test_callbacks_restore - AssertionError: assert {'dampening': 0,\n 'foreach': None,\n 'lr': 0.1,\n 'maximize': False,\n 'momentum': 0,\n 'nesterov': False,\n 'params': ['layer.3.weight', 'layer.3.bias...
FAILED tests/core/test_lightning_module.py::test_proper_refcount - assert 2 == 3
FAILED tests/utilities/test_fetching.py::test_trainer_num_prefetch_batches[mps] - TypeError: object.__new__() takes exactly one argument (the type to instantiate)
```

</details>

pytorch_lightning/accelerators/mps.py
justusschock: I agree that this case could be handled in the accelerator connector as well. But I think it is rather device specific, so I also understand why it is not there.
Maybe this should be resolved and me moved outside to be checked beforehand separately :)

pytorch_lightning/trainer/trainer.py
```diff
@@ -1745,9 +1752,20 @@ def __setup_profiler(self) -> None:
...
+        else:
+            gpu_available = False
+            gpu_type = ""
+
+        gpu_used = isinstance(self.accelerator, (GPUAccelerator, MPSAccelerator))
+        rank_zero_info(f"GPU available: {gpu_available}{gpu_type}, used: {gpu_used}")
```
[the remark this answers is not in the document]
justusschock: It really get's quite verbose and there is no way to disable this at all...

Personally I also like this better

docs/source/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,49 @@
...
+   trainer = Trainer(accelerator="mps", devices=1)
+
+.. note::
+   The ``MPSAccelerator`` only supports 1 device at a time. Currently there are no machines with multiple MPS-capable GPUs.
+
+----
+
+What does MPS stand for?
+------------------------
+MPS is short for Metal Performance Shaders which is the technology used in the back for gpu communication and computing.
```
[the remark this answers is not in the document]
justusschock: added (this isn't md so links have to be provided differently ;) )

justusschock:
@akihironitta I once saw the finetuning callback, but then could reproduce it to be an issue with pt 1.12 not mps . However, when I reinstalled pt1.12 with mps it was gone. The others I haven't seen.

carmocca:
The `test_proper_refcount` failure probably depends on your nightly version because we check whether 1.12 is installed since @otaj sent a fix upstream for it. But it got recently reverted: https://github.com/pytorch/pytorch/pull/76435#issuecomment-1129151004

Borda (approved):

pytorch_lightning/utilities/device_parser.py
```diff
@@ -199,15 +219,36 @@ def _normalize_parse_gpu_input_to_list(gpus: Union[int, List[int], Tuple[int, ..
...
 
-def _get_all_available_gpus() -> List[int]:
+def _get_all_available_gpus(include_cuda: bool = False, include_mps: bool = False) -> List[int]:
+    """
+    Returns:
+        a list of all available gpus
+    """
+    cuda_gpus = _get_all_available_cuda_gpus() if include_cuda else []
+    mps_gpus = _get_all_available_mps_gpus() if include_mps else []
+    return cuda_gpus + mps_gpus
```
awaelchli, line 234: On an M1 with an additional cuda gpu, the sum of these lists could have duplicate entries like [0, 0] for example. Shouldn't the two options be mutually exclusive?

pytorch_lightning/utilities/device_parser.py
```diff
@@ -74,13 +80,19 @@ def parse_gpu_ids(gpus: Optional[Union[int, str, List[int]]]) -> Optional[List[i
             indicates specific GPUs to use.
             An int 0 means that no GPUs should be used.
             Any int N > 0 indicates that GPUs [0..N) should be used.
+        include_cuda: A boolean indicating whether to include cuda devices for gpu parsing.
+        include_mps: A boolean indicating whether to include mps devices for gpu parsing.
```
awaelchli, line 84: I didn't understand where a user would set these. Is this just for testing?

pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -479,6 +480,8 @@ def _choose_accelerator(self) -> str:
                 return "ipu"
             if _HPU_AVAILABLE:
                 return "hpu"
+            if MPSAccelerator.is_available():
```
awaelchli, line 483: Shouldn't this come after cuda?
On a M1 with an additional CUDA GPU, which one would we prefer? Probably the CUDA, right?

tests/helpers/runif.py
```diff
@@ -103,6 +105,7 @@ def __new__(
             tpu: Require that TPU is available.
             ipu: Require that IPU is available.
             hpu: Require that HPU is available.
+            mps: Require that MPS (Apple Sillicon) is available.
```
awaelchli, line 108: The implementation is different from the others. The way it is right now, mps=False forces the test to only run if mps is absent, if I understand correctly. So the docstring should make that clear

tests/lite/test_lite.py
```diff
@@ -448,11 +451,11 @@ def run(self):
                 optimizer_2.step()
 
             for mw_1, mw_2 in zip(model_1.state_dict().values(), model_2.state_dict().values()):
-                assert torch.equal(mw_1, mw_2)
+                assert torch.allclose(mw_1, mw_2)
 
             # Verify collectives works as expected
             ranks = self.all_gather(torch.tensor([self.local_rank]).to(self.device))
-            assert torch.equal(ranks.cpu(), torch.tensor([[0], [1]]))
+            assert torch.allclose(ranks.cpu(), torch.tensor([[0], [1]]))
```
awaelchli, line 458: how does this change relate to the mps accelerator?

tests/accelerators/test_mps.py
```diff
@@ -0,0 +1,165 @@
...
+
+@RunIf(mps=True)
+def test_mps_availability():
+    assert MPSAccelerator.is_available()
+
+
+@RunIf(mps=True)
+def test_warning_if_mps_not_used():
+    with pytest.warns(UserWarning, match="MPS available but not used. Set `accelerator` and `devices`"):
+
```
awaelchli, line 47:
```suggestion
```

tests/accelerators/test_accelerator_connector.py
```diff
@@ -714,12 +715,21 @@ def test_devices_auto_choice_cpu(
 
 @mock.patch("torch.cuda.is_available", return_value=True)
 @mock.patch("torch.cuda.device_count", return_value=2)
+@RunIf(mps=False)
```
awaelchli, line 718: see my comment regarding priorities cuda vs mps

pytorch_lightning/accelerators/mps.py
```diff
@@ -0,0 +1,98 @@
...
+from pytorch_lightning.accelerators.accelerator import Accelerator
+from pytorch_lightning.utilities import device_parser
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.imports import _PSUTIL_AVAILABLE, _TORCH_GREATER_EQUAL_1_12
+from pytorch_lightning.utilities.types import _DEVICE
+
+if _TORCH_GREATER_EQUAL_1_12:
+    _MPS_AVAILABLE = torch.backends.mps.is_available()
+else:
+    _MPS_AVAILABLE = False
```
awaelchli, line 27:
```suggestion
 _MPS_AVAILABLE = _TORCH_GREATER_EQUAL_1_12 and torch.backends.mps.is_available()
```

awaelchli (commented):
Great addition!

pytorch_lightning/utilities/device_parser.py
justusschock: apple does not support cuda gpus and there simply are no drivers :)

pytorch_lightning/utilities/device_parser.py
justusschock: the user never uses these. They are just to unify logic between cuda/gpu and mps

pytorch_lightning/trainer/connectors/accelerator_connector.py
justusschock: that isn’t even a possible combination. Apple does not have support for cuda at all

tests/helpers/runif.py
justusschock: yes that’s true and unfortunately necessary. Will adapt the docstrings

awaelchli (approved):

johnnynunez:
I'm testing this! And works fine! It is cool to use. The problem now is the pytorch backend that most of features hadn't been implemented.
https://github.com/pytorch/pytorch/issues/77764

justusschock:
@johnnynunez Thanks for testing it. You can still use the cpu fallback with `PYTORCH_ENABLE_MPS_FALLBACK=1 python your_script.py` . That will move the computation for unsupported ops to cpu, but luckily due to the shared memory concept it should be very cheap moving the actual tensors

johnnynunez:
> @johnnynunez Thanks for testing it. You can still use the cpu fallback with `PYTORCH_ENABLE_MPS_FALLBACK=1 python your_script.py` . That will move the computation for unsupported ops to cpu, but luckily due to the shared memory concept it should be very cheap moving the actual tensors

thank you! I will test it. I have M1 Max with 32gb

Borda:
@justusschock mind checking the collision? (in git there are much smaller, don't be scared by the GH message) I think we are ready to merge it... :)

docs/source-pytorch/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,48 @@
...
+
+MPS training (basic)
+====================
+**Audience:** Users looking to train on their Apple silicon GPUs.
+
+.. warning::
+
+   The MPS accelerator as well as the PyTorch backend are still very experimental.
+   So far not all operations are supported, but more ops are coming every day due to development from the PyTorch Team.
+   You can use ``PYTORCH_ENABLE_MPS_FALLBACK=1 python your_script.py`` to fall back to cpu for unsupported operations.
```
RobertLaurella, line 13:
```suggestion
   Both the MPS accelerator and the PyTorch backend are still experimental.
   As such, not all operations are currently supported. However, with ongoing development from the PyTorch team, an increasingly large number of operations are becoming available.
   You can use ``PYTORCH_ENABLE_MPS_FALLBACK=1 python your_script.py`` to fall back to cpu for unsupported operations.
```

docs/source-pytorch/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,48 @@
...
+
+
+----
+
+What is Apple silicon?
+----------------------
+Apple silicon chips are a unified system on a chip (SoC) developed by Apple based on the ARM design.
+Among other things, they feature a CPU-cores, GPU-cores, a neural engine and shared memory between all of those.
+
```
RobertLaurella, line 22:
```suggestion
Among other things, they feature CPU-cores, GPU-cores, a neural engine and shared memory between all of these features.

```

docs/source-pytorch/accelerators/mps_basic.rst
```diff
@@ -0,0 +1,48 @@
...
+----
+
+So it's a CPU?
+--------------
+Among other things yes, it includes CPU-cores. However, when running on the ``CPUAccelerator``, not the full potential of hardware acceleration the M-Socs are capable of, is used because they also feature a GPU and a neural engine.
```
RobertLaurella, line 27:
```suggestion
Apple silicon includes CPU-cores among several other features. However, the full potential for the hardware acceleration of which the M-Socs are capable is unavailable when running on the ``CPUAccelerator``. This is because they also feature a GPU and a neural engine.
```

saikatkumardey:
@johnnynunez were you able to fix the issue? I have a similar error while trying to use `AveragePrecision()` on Mac M1 Pro.

By debugging it further I see that [this](https://github.com/Lightning-AI/metrics/blob/58e7ac124ca8800c507558c5b0b477e5fa38a2af/src/torchmetrics/functional/classification/precision_recall_curve.py#L59) line generates the following error.

```
loc("mps_subtract"("(mpsFileLoc): /AppleInternal/Library/BuildRoots/b6051351-c030-11ec-96e9-3e7866fcf3a1/Library/Caches/com.apple.xbs/Sources/MetalPerformanceShadersGraph/mpsgraph/MetalPerformanceShadersGraph/Core/Files/MPSGraphUtilities.mm":219:0)): error: input types 'tensor<128xi64>' and 'tensor<128xf32>' are not broadcast compatible
```

awaelchli:
@saikatkumardey Thanks, since you are mentioning AveragePrecision, I suggest to open an issue directly on https://github.com/Lightning-AI/metrics. Please note, not all function in PyTorch are implemented for MPS.

PR #13142, opened 2022-05-24
Fix torchelastic detection with non-distributed installations

### LINKED ISSUE #13118: Impossible to check whether a current process was launched using the torchelastic command on MacOS
## 🐛 Bug

When checking whether a current process was launched using the torchelastic command on a system running MacOS, it results in an error stating that `module 'torch.distributed' has no attribute 'is_torchelastic_launched'`. This is due to the `torch.distributed` not being available by default on MacOS, which leads to other APIs in the `distributed` package being hidden, as per the PyTorch [documentation](https://pytorch.org/docs/stable/distributed.html).

### To Reproduce

Any call of the `detect()` method from the `TorchElasticEnvironment` class.

### Expected behavior

The `detect()`method should return `False` when `torch.distributed` is not available.

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.6.3
- PyTorch Version (e.g., 1.10): 1.11.0
- Python version (e.g., 3.9): 3.8.13
- OS (e.g., Linux): MacOS
- CUDA/cuDNN version: -
- GPU models and configuration: M1 Chip
- If compiling from source, the output of `torch.__config__.show()`: - 

### Additional context

I discovered this bug when using the [Lightning-Hydra-Template](https://github.com/ashleve/lightning-hydra-template), which calls that method by default to check for availability of distributed job scheduling.

carmocca:
## What does this PR do?
Fixes #13118.

Fixes bug reported in Slack

ss_2022-05-24_20-52-47

None

## Before submitting

## PR review

cc @borda @awaelchli

### DIFF
--- a/pytorch_lightning/plugins/environments/torchelastic_environment.py
+++ b/pytorch_lightning/plugins/environments/torchelastic_environment.py
@@ -62,7 +62,8 @@ class TorchElasticEnvironment(ClusterEnvironment):
     def detect() -> bool:
         """Returns ``True`` if the current process was launched using the torchelastic command."""
         if _TORCH_GREATER_EQUAL_1_9_1:
-            return torch.distributed.is_torchelastic_launched()
+            # if not available (for example on MacOS), `is_torchelastic_launched` is not defined
+            return torch.distributed.is_available() and torch.distributed.is_torchelastic_launched()
         required_env_vars = {"RANK", "GROUP_RANK", "LOCAL_RANK", "LOCAL_WORLD_SIZE"}
         return required_env_vars.issubset(os.environ.keys())
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -223,6 +223,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Avoid redundant callback restore warning while tuning ([#13026](https://github.com/PyTorchLightning/pytorch-lightning/pull/13026))
 
 
+- Fixed torchelastic detection with non-distributed installations ([#13142](https://github.com/PyTorchLightning/pytorch-lightning/pull/13142))
+
+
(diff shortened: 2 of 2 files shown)


ananthsub (approved):

akihironitta (approved):

pytorch_lightning/plugins/environments/torchelastic_environment.py
```diff
@@ -62,7 +62,8 @@ def main_port(self) -> int:
     def detect() -> bool:
         """Returns ``True`` if the current process was launched using the torchelastic command."""
         if _TORCH_GREATER_EQUAL_1_9_1:
-            return torch.distributed.is_torchelastic_launched()
+            # if not available (for example on MacOS), `is_torchelastic_launched` is not defined
+            return torch.distributed.is_available() and torch.distributed.is_torchelastic_launched()
```
justusschock, line 66:
```suggestion
            return torch.distributed.is_available() and getattr(torch.distributed, 'is_torchelastic_launched', lambda : False)()
```
Note that for me on the mac `torch.distributed.is_available()` returns `True` which makes sense for the cpu gloo backend. So this just may not work for all the cases.

justusschock (approved):

pytorch_lightning/plugins/environments/torchelastic_environment.py
carmocca: It might be because yours has the M1. This comment says MacOS builds without it by default

https://github.com/pytorch/pytorch/blob/87148f2b5921557774a535495d7c77b20aa795b6/torch/distributed/__init__.py#L15

Then `is_torchelastic_launched` comes from this wildcard import under `is_available`:

https://github.com/pytorch/pytorch/blob/87148f2b5921557774a535495d7c77b20aa795b6/torch/distributed/__init__.py#L55

PR #13198, opened 2022-06-01
Eager pip install

carmocca:
## What does this PR do?

Fixes the problem described in https://github.com/PyTorchLightning/pytorch-lightning/pull/13048#issuecomment-1142399958

None

## Before submitting

## PR review

### DIFF
--- a/.github/workflows/ci-pytorch_test-full.yml
+++ b/.github/workflows/ci-pytorch_test-full.yml
@@ -78,13 +78,19 @@ jobs:
     - name: Pull legacy checkpoints
       run: bash .actions/pull_legacy_checkpoints.sh
 
+    - run: |
+        flag=$(python -c "print('--pre' if '${{matrix.release}}' == 'pre' else '')" 2>&1)
+        url=$(python -c "print('test/cpu/torch_test.html' if '${{matrix.release}}' == 'pre' else 'cpu/torch_stable.html')" 2>&1)
+        echo "PIP_FLAG=$flag" >> $GITHUB_ENV
+        echo "TORCH_URL="https://download.pytorch.org/whl/${url}" >> $GITHUB_ENV
+
     - name: Install dependencies
       env:
         PACKAGE_NAME: pytorch
       run: |
         flag=$(python -c "print('--pre' if '${{matrix.release}}' == 'pre' else '')" 2>&1)
         url=$(python -c "print('test/cpu/torch_test.html' if '${{matrix.release}}' == 'pre' else 'cpu/torch_stable.html')" 2>&1)
-        pip install -e .[test] --upgrade $flag --find-links "https://download.pytorch.org/whl/${url}"
+        pip install -e .[test] $PIP_FLAG --find-links "$TORCH_URL" -U --upgrade-strategy eager
         pip list
       shell: bash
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-checks.yml
```diff
@@ -42,9 +42,9 @@ jobs:
           pip --version
           pip install -q fire
           # python -m pip install --upgrade --user pip
-          pip install -r requirements.txt --upgrade-strategy only-if-needed --find-links https://download.pytorch.org/whl/cpu/torch_stable.html --quiet
-          pip install -r requirements/devel.txt
-          pip install -r requirements/docs.txt
+          pip install -r requirements.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html --quiet --upgrade-strategy eager
+          pip install -r requirements/devel.txt --upgrade-strategy eager
```
carmocca, line 46: These also need `-U`, right? `--upgrade-strategy` does not upgrade by itself.

Borda (approved):
seem complete for me

.github/workflows/ci_test-full.yml
```diff
@@ -89,19 +89,19 @@ jobs:
...
         flag=$(python -c "print('--pre' if '${{matrix.release}}' == 'pre' else '')" 2>&1)
         url=$(python -c "print('test/cpu/torch_test.html' if '${{matrix.release}}' == 'pre' else 'cpu/torch_stable.html')" 2>&1)
-        pip install -r requirements.txt --upgrade $flag --find-links "https://download.pytorch.org/whl/${url}"
+        pip install -r requirements.txt $flag -U --upgrade-strategy eager --find-links "https://download.pytorch.org/whl/${url}"
         # adjust versions according installed Torch version
         python ./requirements/adjust-versions.py requirements/examples.txt
-        pip install -r requirements/examples.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html --upgrade
-        pip install -r requirements/test.txt --upgrade
+        pip install -r requirements/examples.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html -U --upgrade-strategy eager
+        pip install -r requirements/test.txt -U --upgrade-strategy eager
```
carmocca, line 96: CI is failing because a `test.txt` dependency requires `protobuf` and since we are marking eager here, it gets upgraded to the latest version since it's not considering the pin in `requirements.txt`

Not sure about how we can resolve this problem. All requirements would need to be installed together instead of in steps so the resolution algorithm can find a working combination.

carmocca:
Closing, as I don't think we can fix the issue this way reliably.

I think the best option now would be to have some sort of lock file by using a tool like poetry.

Borda:
fyi, shall be passing after #13441

Borda:
> I don't see how it solves the issue described in [#13198 (comment)](https://github.com/Lightning-AI/lightning/pull/13198#discussion_r887842276)

yes, this won't address installing requirements together but set the proper upgrade for testing the bumped version are correct/feasible

carmocca:
> merge this and check how to join and install the remaining req. or do you prefer to do it single PR?

I would prefer to have a general solution that works for all requirements before starting to make changes. This will make any possible issues pop up

Borda:
> I would prefer to have a general solution that works for all requirements before starting to make changes.

but be aware that no other CI/workflow requires running the latest, other rather uses cache, for example, Conda has specified PT versions and other like dockers are fine to build with a version available not necessary the latest... so I believe we shall just update this CI job

.github/workflows/ci-pytorch_test-full.yml
```diff
@@ -143,7 +147,7 @@ jobs:
       run: |
         # adjust versions according installed Torch version
         python ./requirements/pytorch/adjust-versions.py requirements/pytorch/examples.txt
-        pip install -r requirements/pytorch/examples.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html --upgrade
+        pip install $PIP_FLAG -r requirements/pytorch/examples.txt --find-links "$TORCH_URL --upgrade
```
Borda, line 150: so what we can do here is reinstall the all above to reduce installing different versions to be not sliding with example requirements...

Issue #13181, opened 2022-05-30
Unexpected behavior of .save_hyperparameters() for same parameter name


serena-ruan:
## 🐛 Bug

When a pytorch_lightning model is created as a parameter for a class, and they contain parameters with same name, the original class's value overrides the model's value when calling `.save_hyperparameters()`. Besides, it also saves all parameters for the wrapper class.

### To Reproduce

I created a BoringParent class (not actually a parent) that contains a param named `same_arg` with a default value `parent`, and inside the init function I created a Boring_saving_hprams object that also contains a param named `same_arg` and I call `save_hyperparameters` to save the params for this pytorch_lightning module. However, the value of `same_arg` is overriden by the default BoringParent class's value.
Please see the colab notebook for details.
https://colab.research.google.com/drive/1De82M7GEzgtc18UTSRHJOV_nNnIJrji6?usp=sharing

### Expected behavior

I'm expecting the model should save its own value even when there're param name conflicts. And in this case the class BoringParent is not actually the parent, it's just the object holding this model, we shouldn't iterate over its parameters. I also just want it to save the parameters for this pytorch_lightning modle only, not parameters from the class that holds this model.

### Environment

* CUDA:
	- GPU:
		- Tesla T4
	- available:         True
	- version:           11.3
* Packages:
	- numpy:             1.21.6
	- pyTorch_debug:     False
	- pyTorch_version:   1.11.0+cu113
	- pytorch-lightning: 1.6.3
	- tqdm:              4.64.0
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- 
	- processor:         x86_64
	- python:            3.7.13
	- version:           #1 SMP Sun Apr 24 10:03:06 PDT 2022

### Additional context

Inside the save_hyperparameters function, it's updating the init_args here so the later args just override the previous one:
<img width="287" alt="image" src="https://user-images.githubusercontent.com/82044803/170971090-ebd76b6c-8eaf-4e90-9e53-0701b59d7e32.png">
And inside `collect_init_args` it's recursively checking the params passed in, which is not the correct order for my case. I think that we could check whether there's a real inheritance relationship between classes and if not we don't add the param inside the path_args.

Also let me know if you want me to draft a PR. Thanks!

cc @carmocca @justusschock @awaelchli @borda @ananthsub @ninginthecloud @jjenniferdai @rohitgr7

tanmoyio:
@serena-ruan taking a look

PR #13202, opened 2022-06-01
Make `save_hyperparamters` not save another class's args when `LightningModule` instantiated within another class's constructor

tanmoyio:
## Duplicate args conflict with `.save_hyperparameters()`

Simple hack to fix the issue https://github.com/PyTorchLightning/pytorch-lightning/issues/13181 by @[serena-ruan](https://github.com/serena-ruan)

quick test
```python3
class Mod1(LightningModule):
    def __init__(self, same_arg):
        super().__init__()
        self.same_arg = same_arg
        self.save_hyperparameters()

class Mod2(LightningModule):
    def __init__(self, same_arg):
        super().__init__()
        self.same_arg = same_arg
        self.save_hyperparameters()

class Parent():
    def __init__(self, same_arg="parent", diff_arg="test"):
        super().__init__()
        self.m1 = Mod1(same_arg='child1')
        self.m2 = Mod2(same_arg='child2')

parent = Parent()
print(parent.m1.hparams, parent.m2.hparams)
``` 

## Before submitting

## PR review

## Did you have fun?


awaelchli:
@tanmoyio could this introduce any breaking changes? I wasn't sure

tanmoyio:
@awaelchli It will not. One unit test broke in `test_parsing.py` which was expected, I modified that.

serena-ruan:
@tanmoyio Just one more question regarding the issue I raised, this PR will take children's param as priority, but it still saves whatever params existing in the parent. And in my issue the two classes are not inheritance relationship, so will you solve this problem in a separate PR?

tanmoyio:
@serena-ruan 
```python3
class Boring_saving_hparams(LightningModule):
    def __init__(self, same_arg):
        super().__init__()
        self.same_arg = same_arg
        self.save_hyperparameters()

class BoringParent():
    def __init__(self, same_arg="parent", diff_arg="test"):
        super().__init__()
        self.model = Boring_saving_hparams(same_arg="child")
```
in your code, even if the `BoringParent` is not inheriting it is still initializing a object of `Boring_saving_hparams` which uses `save_hyperparameters()`. So it will keep track of the hparams, and it should. 

@awaelchli what do you think ?

tanmoyio:
@rohitgr7 reference https://github.com/Lightning-AI/lightning/issues/13181

akihironitta:
> can you share some cases where such kind of inheritance is used? `LightningModule`is a system. If your child components just contain some layers, they should be defined as `nn.Modules`.

@rohitgr7 I have the same thought and am still curious about the actual use cases (cc @tanmoyio @serena-ruan), but at the same time, I think we can't assume that a lightning module is always instantiated outside any class's `__init__`.

akihironitta:
> in your code, even if the `BoringParent` is not inheriting it is still initializing a object of `Boring_saving_hparams` which uses `save_hyperparameters()`. So it will keep track of the hparams, and it should. 

@tanmoyio I don't think it should since it's completely out of scope of PL. If one wants to log other args of the wrapping class, they should be passed to the lightning module's `__init__`.

In short, I believe the following assertions need to pass in this PR:

```python

    def __init__(self, same_arg):
        super().__init__()
        self.save_hyperparameters()

class SomeClass:
    def __init__(self, same_arg="parent", diff_arg="test"):
        super().__init__()
        self.model = BoringModel(same_arg="child")
        assert self.model.hparams.same_arg == "child", f"'{self.model.hparams.same_arg}' should be 'child'."
        assert not hasattr(self.model.hparams, "diff_arg"), "`diff_arg` shouldn't be saved in `self.model.hparams`."

some_class = SomeClass()
```

awaelchli:
I propose to close this PR. Not having the `save_hyperparameters()` at the "root" module called is a misuse and should be prohibited. If it was done like this, we can't reliably instantiate the model using the `load_from_checkpoint` function.

carmocca:
@awaelchli It's not clear to me if your argument also applies to the linked issue and whether it should be closed too as non-supported usage. Closing is your call, as you are the most familiar on this topic :)

awaelchli:
> It's not clear to me if your argument also applies to the linked issue and whether it should be closed too

Yes it would apply to both. The only reason to keep the issue open I think would be to check whether it is possible to communicate to the user better that this is unintended use case / can have unintended side effects.

awaelchli:
@carmocca Actually, I am simply referring to the use case given in the description here. I just realized it is not at all the same as in the linked issue. 

The linked issue should be addressed, but it is clearly different. This PR tries to resolve a limitation with inheritance that I argue can't be logically supported.
On the other hand, the reported issue is about composition, and I think there is a bug.

PR #13224, opened 2022-06-03
DummyLogger can be called with unknown methods

### LINKED ISSUE #13203: DummyLogger AttributeError when calling WandbLogger methods with fast_dev_run
## 🐛 Bug
When using the `fast_dev_run` flag `pytorch-lightning` replaces the logger with a `DummyLogger` object. But with a WandbLogger I also log media with `self.logger.log_text` in my pl module.  I therefore get an AttributeError that the DummyLogger does not have the attribute `log_text`. 

### To Reproduce

```python
import os

import torch

from pytorch_lightning.loggers import WandbLogger


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        self.logger.log_text(
            "samples",
            columns=["epoch", "temperatue", "text"],
            data=[[1, 0.7, "generated_text"]],
        )
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

def run():
    wandb_logger = WandbLogger(project="myproject", offline=False)

    model = BoringModel()
    trainer = Trainer(
        logger=wandb_logger,
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        enable_model_summary=False,
        fast_dev_run=True
    )


if __name__ == "__main__":
    run()

```

### Expected behavior

The `DummyLogger` ignores attribute access of external loggers. 

### Environment

```bash

* CUDA:
	- GPU:
	- available:         False
	- version:           None
* Packages:
	- numpy:             1.22.3
	- pyTorch_debug:     False
	- pyTorch_version:   1.11.0
	- pytorch-lightning: 1.6.3
	- tqdm:              4.64.0
* System:
	- OS:                Darwin
	- architecture:
		- 64bit
		-
	- processor:         i386
	- python:            3.9.10
	- version:           Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64
```

### Additional context

Requires probably a `wandb` setup.

cc @borda @awaelchli @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy

HallerPatrick:
## What does this PR do?

This PR overrides the `__getattr__` method of the `DummyLogger` class, such that an unkown method calls due to external 
loggers do not cause an `AttributeError` but return a dummy method and print information about the call to a non-existing method.

Fixes #13203 

-

## Before submitting

## PR review

## Did you have fun?


pytorch_lightning/loggers/logger.py
```diff
@@ -346,6 +346,16 @@ def __iter__(self):
         # if DummyLogger is substituting a logger collection, pretend it is empty
         yield from ()
 
+    def __getattr__(self, name):
+        """Allows the DummyLogger to be called with arbitrary methods, to avoid AttributeErrors."""
+
+        def method(*args, **kwargs):
+            rank_zero_info(
+                f"The DummyLogger is being used and the method `{name}` was called on it but it does not exist"
```
carmocca, line 354: Let's mention fast_dev_run
```suggestion
                f"The `DummyLogger` is being used and the method `{name}` was called on it but it does not exist."
                " This is the case when `fast_dev_run` is enabled"
```

pytorch_lightning/loggers/logger.py
```diff
@@ -346,6 +346,16 @@ def __iter__(self):
         # if DummyLogger is substituting a logger collection, pretend it is empty
         yield from ()
```
carmocca, line 347: Can you update the CHANGELOG? The "Added" section would be the best fit

pytorch_lightning/loggers/logger.py
```diff
@@ -346,6 +346,16 @@ def __iter__(self):
...
 
+    def __getattr__(self, name):
+        """Allows the DummyLogger to be called with arbitrary methods, to avoid AttributeErrors."""
+
+        def method(*args, **kwargs):
+            rank_zero_info(
+                f"The DummyLogger is being used and the method `{name}` was called on it but it does not exist"
+            )
+
+        return method
```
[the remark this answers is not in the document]
carmocca: TIL :)

pytorch_lightning/loggers/logger.py
```diff
@@ -346,6 +346,17 @@ def __iter__(self):
...
         yield from ()
 
+    def __getattr__(self, name: str) -> Callable:
+        """Allows the DummyLogger to be called with arbitrary methods, to avoid AttributeErrors."""
+
+        def method(*args, **kwargs):
+            rank_zero_info(
+                f"The `DummyLogger` is being used and the method `{name}` was called on it but it does not exist."
+                " This is the case when `fast_dev_run` is enabled"
+            )
```
awaelchli, line 356: What is the motivation behind emitting an info here? This would lead to a spam of messages for every logging step.
Also it is not only used for fast dev run. Can we drop it? A dummy method with return value None should suffice I think.

pytorch_lightning/loggers/logger.py
```diff
@@ -346,6 +346,17 @@ def __iter__(self):
         # if DummyLogger is substituting a logger collection, pretend it is empty
         yield from ()
 
+    def __getattr__(self, name: str) -> Callable:
```
awaelchli, line 349: Note, now that is is implemented, getattr would be called on any attribute lookup that cant be relsolved by getattribute(). We assume the attribute is always a method, which is the best we can do I guess.

pytorch_lightning/loggers/logger.py
carmocca: I suggested this, because otherwise this could be silently being executed without users noticing

awaelchli: I think we shouldn't do this. The whole point is that using the dummy logger silently makes all logging calls dumb and do nothing to simulate logging being turned off. If we need a message, it should be done through the trainer when fast dev run is active, IMO, we can say something along the lines of logging and checkpointing is suppressed for fast dev run.

HallerPatrick: I rewrote it

awaelchli: This looks great! Just the test needs to be adjusted

tests/tests_pytorch/loggers/test_logger.py
```diff
@@ -257,6 +257,18 @@ def test_dummylogger_noop_method_calls():
     logger.log_metrics("1", 2, three="three")
 
 
+def test_dummlogger_arbitrary_method_calls(caplog):
+    """Test that the DummyLogger can be called with non existing methods."""
+    logger = DummyLogger()
+    # Example method from WandbLogger
+    logger.log_text("1", columns=["1", "2", "3"], data=[[1, 2, 3]])
```
awaelchli, line 264: Since getattr was implemented, you could now assert

```py
assert hasattr(logger, "cocofruit")
assert callable(logger.cocofruit)
```

awaelchli (approved):
Just the test needs an update

CHANGELOG.md
```diff
@@ -74,6 +74,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 -
 
 
+
+- Override `__getattr__` in `DummyLogger` to avoid `AttributeError` from third party logger calls in `fast_dev_run` ([#13224](https://github.com/PyTorchLightning/pytorch-lightning/pull/13224))
```
rohitgr7, line 78:
```suggestion
- Added support for calling unknown methods with `DummyLogger` ([#13224](https://github.com/PyTorchLightning/pytorch-lightning/pull/13224))
```

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -589,7 +589,10 @@ def _init_debugging_flags(
             self.check_val_every_n_epoch = 1
             self.loggers = [DummyLogger()] if self.loggers else []
 
-            rank_zero_info(f"Running in fast_dev_run mode: will run the requested loop using {num_batches} batch(es).")
+            rank_zero_info(
+                f"Running in fast_dev_run mode: will run the requested loop using {num_batches} batch(es). "
+                "Logging and checkpointing is suppressed."
```
rohitgr7, line 594:
```suggestion
                f"Running in `fast_dev_run` mode: will run the requested loop using {num_batches} batch(es). "
                "Logging and checkpointing is suppressed."
```

tests/tests_pytorch/loggers/test_logger.py
```diff
@@ -257,6 +257,18 @@ def test_dummylogger_noop_method_calls():
...
+
+    assert (
+        "The `DummyLogger` is being used and the method `log_text` was called on it "
+        "but it does not exist. This is the case when `fast_dev_run` is enabled\n" in caplog.text
+    )
```
rohitgr7, line 269:
```suggestion
    with caplog.at_level(level=logging.INFO):
        logger.log_text("1")

    assert "method `log_text` was called on it but it does not exist" in caplog.text
```
do import `logging`

src/pytorch_lightning/loggers/logger.py
```diff
@@ -346,6 +346,14 @@ def __iter__(self):
         # if DummyLogger is substituting a logger collection, pretend it is empty
         yield from ()
 
+    def __getattr__(self, name: str) -> Callable:
+        """Allows the DummyLogger to be called with arbitrary methods, to avoid AttributeErrors."""
+
+        def method(*args, **kwargs):
```
rohitgr7, line 352: where is the info generated?

tests/tests_pytorch/loggers/test_logger.py
HallerPatrick: Capturing logging is not required as the dummy method is not calling `rank_zero_info` anymore

src/pytorch_lightning/loggers/logger.py
HallerPatrick: Now only general info is generated when running in `fast_dev_run`, see [here](https://github.com/HallerPatrick/pytorch-lightning/blob/2833cc19e2f59f793be9cb91c42a79b6b8f51caf/src/pytorch_lightning/trainer/trainer.py#L594)

tests/tests_pytorch/loggers/test_logger.py
HallerPatrick: Yes, that looks more clean!

otaj (approved):

PR #13262, opened 2022-06-09
Fix erroneous UserWarning for unset max_epochs

### LINKED ISSUE #13257: Erroneous UserWarning for unset max_epochs
## 🐛 Bug

When I use:
```python
Trainer(fast_dev_run=True)
```
I see the following warning:
```
pytorch_lightning.utilities.warnings.PossibleUserWarning: `max_epochs` was not set. Setting it to 1000 epochs. To train without an epoch limit, set `max_epochs=-1`.
```
However, `fast_dev_run=True` implies `max_epochs=1`, so shouldn't this warning be skipped?

### To Reproduce

Use code like:
```python
Trainer(fast_dev_run=True)
```
in your code with warnings enabled or captured by pytest.

### Expected behavior

No warning message.

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.6.4
- PyTorch Version (e.g., 1.10): 1.11.0
- Python version (e.g., 3.9): 3.10.4, etc.
- OS (e.g., Linux): Linux, etc.
- CUDA/cuDNN version: N/A
- If compiling from source, the output of `torch.__config__.show()`: N/A
- Any other relevant information: N/A

cc @justusschock @kaushikb11 @awaelchli @borda @rohitgr7

adamjstewart:
## What does this PR do?

Fixes #13257. @akihironitta

Now the warning is raised only during `trainer.fit` call.

None

## Before submitting

## PR review

## Did you have fun?


tests/trainer/flags/test_min_max_epochs.py
```diff
@@ -41,3 +43,11 @@ def test_max_epochs_not_set_warning():
     with pytest.warns(PossibleUserWarning, match="`max_epochs` was not set. Setting it to 1000 epochs."):
         trainer = Trainer(max_epochs=None)
         assert trainer.max_epochs == 1000
+
+
+def test_fast_dev_run_no_warning():
+    """Test that no warning is emitted when `fast_dev_run` is set."""
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
```
adamjstewart, line 51: See https://docs.pytest.org/en/7.0.x/how-to/capture-warnings.html#additional-use-cases-of-warnings-in-tests

adamjstewart:
Tests and fix were separated into 2 commits to show that the tests do indeed fail without the fix.

tests/trainer/flags/test_min_max_epochs.py
carmocca: You can use this utility we have:

```
from tests.helpers.utils import no_warning_call

with no_warning_call(match="..."):
    ...
```

pytorch_lightning/trainer/trainer.py
```diff
@@ -449,7 +449,7 @@ def __init__(
         self.tuner = Tuner(self)
 
         min_steps, max_steps, min_epochs, max_epochs, max_time = _parse_loop_limits(
-            min_steps, max_steps, min_epochs, max_epochs, max_time
+            min_steps, max_steps, min_epochs, max_epochs, max_time, fast_dev_run
```
carmocca, line 452: IMO instead of passing `fast_dev_run` here, a better solution would be to call this

https://github.com/PyTorchLightning/pytorch-lightning/blob/bfa8b7be2d99b980afa62f5cb0433326bcfd2ef0/pytorch_lightning/trainer/trainer.py#L544

before `parse_loop_limits` since `fast_dev_run` might modify the limits.

adamjstewart: It looks like moving `_init_debugging_flags` above `_parse_loop_limits` breaks because it requires `fit_loop` to be set, but that is set using values from the output of `_parse_loop_limits`. Should I revert to my previous solution or is there a way to break this loop?

carmocca: I pushed a commit that should fix the problem. It required inlining the `init_debugging_flags` method.

pytorch_lightning/trainer/trainer.py
```diff
@@ -448,6 +448,59 @@ def __init__(
         self._signal_connector = SignalConnector(self)
```
carmocca, line 448: Can you update the CHANGELOG? "Fixed" section

carmocca (approved):

akihironitta (dismissed):

akihironitta (commented):
@adamjstewart Thank you for working on this! Would you mind having a look at the failing tests in CI? https://github.com/PyTorchLightning/pytorch-lightning/runs/6895684273?check_suite_focus=true

I think applying the following patch should fix the error :)

adamjstewart:
@carmocca it seems like your solution isn't passing the tests

tests/tests_pytorch/trainer/flags/test_min_max_epochs.py
```diff
@@ -37,7 +38,13 @@ def test_min_max_steps_epochs(tmpdir, min_epochs, max_epochs, min_steps, max_ste
...
+    """Test that a warning is only emitted when `max_epochs` was not set by the user."""
+    match = "`max_epochs` was not set. Setting it to 1000 epochs."
+
+    with pytest.warns(PossibleUserWarning, match=match):
         trainer = Trainer(max_epochs=None)
         assert trainer.max_epochs == 1000
+
+    with no_warning_call(expected_warning=PossibleUserWarning, match=match):
+        Trainer(fast_dev_run=True)
+        Trainer(fast_dev_run=1)
```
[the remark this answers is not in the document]
adamjstewart: We could try 2. I think it's important to try all input types and make sure the parameter behaves as expected. For example, if this PR checked to see `if fast_dev_run == True:` (not a good solution, I know), then this test would catch that mistake.

carmocca:
> it seems like your solution isn't passing the tests

Should be passing now

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -536,75 +578,9 @@ def __init__(
...
-            limit_predict_batches,
-            val_check_interval,
-            overfit_batches,
-            fast_dev_run,
-        )
-
         # Callback system
         self._call_callback_hooks("on_init_end")
 
-    def _init_debugging_flags(
```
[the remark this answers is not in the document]
rohitgr7: qq: do you think we could dep & remove `overfit_batches`?

all it does is this:
```py
if overfit_batches_enabled:
    self.limit_train_batches = overfit_batches
    self.limit_val_batches = overfit_batches
```

akihironitta (approved):
It looks good to me other than the possible readability issue https://github.com/Lightning-AI/lightning/pull/13262/files#r903701888

adamjstewart:
Is this PR ready to merge or is there anything else it needs?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -456,6 +456,65 @@ def __init__(
...
+            self.limit_predict_batches = _determine_batch_limits(limit_predict_batches, "limit_predict_batches")
+            self.num_sanity_val_steps = float("inf") if num_sanity_val_steps == -1 else num_sanity_val_steps
+            self.val_check_interval = _determine_batch_limits(val_check_interval, "val_check_interval")
+        if overfit_batches_enabled:
+            self.limit_train_batches = overfit_batches
+            self.limit_val_batches = overfit_batches
+
         min_steps, max_steps, min_epochs, max_epochs, max_time = _parse_loop_limits(
             min_steps, max_steps, min_epochs, max_epochs, max_time
         )
```
awaelchli, line 520: While `_parse_loop_limits` doesn't do much with it, max_time by construction of the function call could be modified. Previously it was passed to `self._callback_connector.on_trainer_init` after these lines but now it has moved to before.

In other words, before we did the parsing first before using the values, now it is the other way around.

An alternative solution that avoids the re-ordering would have been to pass the `fast_dev_run` flag down to the `_parse_loop_limits` to be able to avoid emitting the warning. Was this considered? If yes, what was the blocker for not being able to do it this way?

adamjstewart: > An alternative solution that avoids the re-ordering would have been to pass the `fast_dev_run` flag down to the `_parse_loop_limits` to be able to avoid emitting the warning. Was this considered?

This was actually my initial solution. Aside from testing, the entire PR was a single commit: https://github.com/Lightning-AI/lightning/pull/13262/commits/97e73aab50f6192bae29fc5d419c989ded21a85b.

> If yes, what was the blocker for not being able to do it this way?

https://github.com/Lightning-AI/lightning/pull/13262#discussion_r897041894

rohitgr7 (changes requested):
I'd consider implementing https://github.com/Lightning-AI/lightning/pull/13262#issuecomment-1164448685

this way we can keep those methods that handles debugging flags.

carmocca:
> start raising this warning during .fit call cause during that call it actually makes sense

This PR addresses a bug in our existing logic, whereas this proposal would be a refactor and should be left for a follow-up

> do you think we could dep & remove overfit_batches?

Same comment about this suggestion. This should be left for a separate issue where we discuss the advantages and disadvantages.

I don't think it's relevant to keep the separate method because it's not called from anywhere else and it simplifies the attribute setting of the Trainer instance. The code had to be moved around to make sure things are set where expected, but not looking at the diff I feel like the code is somewhat easier to follow now for a new user.

adamjstewart:
I assume these test failures are unrelated to the PR?

rohitgr7 (approved):

src/pytorch_lightning/loops/fit_loop.py
```diff
@@ -50,10 +50,10 @@ class FitLoop(Loop[None]):
     def __init__(
         self,
         min_epochs: int = 0,
-        max_epochs: int = 1000,
+        max_epochs: Optional[int] = None,
```
carmocca, line 53: @rohitgr7 Why this change? It's cleaner if the loop takes the argument already validated (converted to int). We should always try to avoid Optionals

rohitgr7: the issue is max_epochs doesn't do anything for `trainer.test/validate/predict` but we initialize the fit_loop in the trainer constructor. We need to ensure it gets updated and that warning should be raised during `.fit` only. Due to this, I don't see a possibility of setting it to an int before the call.

adamjstewart:
Is there still interest in this PR? My original commits solved the issue in only a few lines of code and passed all tests. Since then, this PR has blown up into a huge refactor with 4+ contributors, 23 commits, and 9 files changed, and is still failing tests. I no longer have any idea how this code works or if this is the right thing to do. Would it be possible to merge the first couple of commits and leave the refactor for another day?

awaelchli:
@adamjstewart We haven't forgotten about this. I would very much prefer to go with the simple solution (see these two [comments](https://github.com/Lightning-AI/lightning/pull/13262#discussion_r909087735)). Any larger issues about how we should validate some input arguments that only apply when certain trainer functions get called should be left for a different issue in my opinion, as this is a broader question than just fast dev run.

carmocca (approved):

carmocca:
LGTM. Just needs passing CI and your approval @awaelchli

rohitgr7:
> @adamjstewart We haven't forgotten about this. I would very much prefer to go with the simple solution (see these two [comments](https://github.com/Lightning-AI/lightning/pull/13262#discussion_r909087735)). Any larger issues about how we should validate some input arguments that only apply when certain trainer functions get called should be left for a different issue in my opinion, as this is a broader question than just fast dev run.
> 
> @rohitgr7 @carmocca thoughts?

I understand, if you want, I'll move out the commits from this PR to another one. But it's more like just moving some of the code pieces somewhere else and then moving them back in follow-up.

Borda (approved):

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -455,9 +455,22 @@ def __init__(
         self._signal_connector = SignalConnector(self)
         self.tuner = Tuner(self)
 
-        min_steps, max_steps, min_epochs, max_epochs, max_time = _parse_loop_limits(
-            min_steps, max_steps, min_epochs, max_epochs, max_time
+        # Declare attributes to be set in _callback_connector on_trainer_init
+        self._callback_connector.on_trainer_init(
```
awaelchli, line 459: I'm not happy that we are changing the order of these. We still have the issue that some of these connectors write back attributes using their internal trainer reference, and the order in which things get initialized here can matter. Have we carefully checked this? 

It looks like all you wanted to do is move `_parse_loop_limit`. Would we be able to move the callback connector and init_debugging_flags and logger connector, back to where they were? For sanity? Or have we done this to resolve another bug?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -567,66 +563,64 @@ def _init_debugging_flags(
...
 
         self.should_stop = False
         self.state = TrainerState()
         self.num_training_batches = float("inf")
         self.train_dataloader = None
 
-        if num_sanity_val_steps == -1:
-            self.num_sanity_val_steps = float("inf")
-        else:
-            self.num_sanity_val_steps = num_sanity_val_steps
```
awaelchli, line 628: I suppose this was an unrelated refactor because you want to classify sanity_val_steps as a debugging flag?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1138,6 +1132,12 @@ def _run(
         # ----------------------------
         # TRAIN
         # ----------------------------
+        if self.state.fn in (TrainerFn.FITTING, TrainerFn.TUNING):
+            min_epochs, max_epochs = _parse_loop_limits(
+                self.min_steps, self.max_steps, self.min_epochs, self.max_epochs, self
+            )
+            self.fit_loop.min_epochs = min_epochs
+            self.fit_loop.max_epochs = max_epochs
```
awaelchli, line 1140: why aren't we doing it earlier? At this point we've already run model hooks that could have accessed the trainer.
Note, a few lines above we  `verify_loop_configurations(self)`, so I think this should at least move up there

awaelchli (approved):
Thanks @adamjstewart for working on this and @rohitgr7 / @carmocca for finishing it. 
I think it looks good as a whole solution but the way to get here could have been different IMO. We could have merged the perfectly fine fix that was initially implemented and then iterated on it later. These wouldn't have been conflicting changes and I hope @adamjstewart isn't too upset that we've overridden the direction here. Nonetheless, great work everyone! <3

adamjstewart:
> I hope @adamjstewart isn't too upset that we've overridden the direction here.

Not at all, just glad to see things fixed. My fix was a quick hack, but I agree that a refactor makes sense here.

src/pytorch_lightning/trainer/trainer.py
carmocca: Tests should be failing if they aren't moved (I was who initially moved it!). It's very tricky to see these dependencies between connector, the easiest way to double-check would be to move it back and run CI

src/pytorch_lightning/trainer/trainer.py
carmocca: I believe I made this change. I moved it because in one of the `_init_debugging_flags`'s branches we were already setting `self.num_sanity_val_steps = 0`, so it made sense that we managed the attribute entirely inside that method

src/pytorch_lightning/trainer/trainer.py
rohitgr7: done!

src/pytorch_lightning/trainer/trainer.py
rohitgr7: tried moving back and it failed

awaelchli: I reverted the re-ordering in [64e4506](https://github.com/Lightning-AI/lightning/pull/13262/commits/64e450663f4e891f2b7f6b3971ad105eef51efe4) and all tests pass. I am still convinced not touching the ordering here is the best way for sanity.

If you still think we should be doing the re-ordering, please feel free revert my last commit, then merge the PR.

PR #13271, opened 2022-06-11
collaborative strategy: client mode

t-vi:
## What does this PR do?

Hivemind features a client mode for cases where a computer that is not reachable by the peers in a collaborative run is not reachable from them (e.g. behind a NAT). This exposes the client mode flag to the strategy.

No.

I'm unsure how to test this properly.

### DIFF
--- a/pytorch_lightning/strategies/collaborative.py
+++ b/pytorch_lightning/strategies/collaborative.py
@@ -53,6 +53,7 @@ class CollaborativeStrategy(Strategy):
         port: Optional[int] = None,
         retry_endpoint_attempts: int = 5,
         retry_endpoint_sleep_duration: int = 5,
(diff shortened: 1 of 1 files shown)


justusschock (approved):
I think this does not really require tests as this only passes through arguments to hivemind.

A changelog entry would be nice though :)

Borda:
@SeanNaren ^^ :rabbit:

akihironitta:
@SeanNaren Could you review this PR?

t-vi:
I think it is still relevant, I'll fix it.

t-vi:
I update things and filed https://github.com/Lightning-AI/lightning/pull/13761 , so this is obsolete. Thanks again.

PR #13306, opened 2022-06-15
Future 5/n: Move requirements

carmocca:
## What does this PR do?

Package specific requirements


requirements/pytorch/collect_env_details.py
```diff
@@ -24,7 +24,7 @@
 import torch
```
Borda, line 24: let's keep this at the root of the requirement and, if needed, extend by collecting info needed for LAI?

akihironitta (approved):

Borda (approved):

awaelchli (approved):

PR #13311, opened 2022-06-16
Update README and approval config

carmocca:
## What does this PR do?

See title

cc @borda @rohitgr7


.github/approve_config.yml
```diff
@@ -2,16 +2,25 @@ approvals:
   # check will fail if there is no approval
   minimum: 1
   groups:
-    - name: GridAI
-      minimum: 1
+    - name: 'PyTorch Lightning'
```
Borda, line 5:
```suggestion
    - name: 'Lightning AI'
```

.github/approve_config.yml
```diff
@@ -2,16 +2,25 @@ approvals:
...
         - carmocca
-        - edenlightning
         - ethanwharris
         - kaushikb11
         - krshrimali
+        - otaj
         - rohitgr7
         - tchaton
         - williamFalcon
+    - name: 'Lightning Apps'
```
Borda, line 18: How does this link to the source folder?

carmocca: It doesn't. But I don't think it's necessary because this is just the employee approval.

.github/approve_config.yml
carmocca: This is the group name

Borda (approved):

PR #13350, opened 2022-06-21
[BUG] `estimated_stepping_batches` requires distributed comms in `configure_optimizers` for `DeepSpeedStrategy`

### LINKED ISSUE #12317: Can't use `estimated_stepping_batches` in `configure_optimizers` with DDP
## 🐛 Bug

When using DDP and calling `estimated_stepping_batches` in `configure_optimizers`, an error is thrown.
It happens because there's an attempt to sync between the processes using the model's device, but the model hasn't been moved to a non-cpu device yet.
https://github.com/PyTorchLightning/pytorch-lightning/blob/7ee690758ccad7f702460d056f6369c1d4371a46/pytorch_lightning/utilities/data.py#L124
The error:
```
  File "~/bug_report_model.py", line 36, in configure_optimizers
    self.trainer.estimated_stepping_batches
  File "~/pytorch-lightning/pytorch_lightning/trainer/trainer.py", line 2702, in estimated_stepping_batches
    self.reset_train_dataloader()
  File "~/pytorch-lightning/pytorch_lightning/trainer/trainer.py", line 1848, in reset_train_dataloader
    if has_len_all_ranks(self.train_dataloader, self.strategy, module)
  File "~/pytorch-lightning/pytorch_lightning/utilities/data.py", line 124, in has_len_all_ranks
    total_length = training_type.reduce(torch.tensor(local_length).to(model.device), reduce_op="sum")
  File "~/pytorch-lightning/pytorch_lightning/strategies/ddp_spawn.py", line 224, in reduce
    tensor = sync_ddp_if_available(tensor, group, reduce_op=reduce_op)
  File "~/pytorch-lightning/pytorch_lightning/utilities/distributed.py", line 95, in sync_ddp_if_available
    return sync_ddp(result, group=group, reduce_op=reduce_op)
  File "~/pytorch-lightning/pytorch_lightning/utilities/distributed.py", line 129, in sync_ddp
    torch.distributed.all_reduce(result, op=op, group=group, async_op=False)
  File "~/lib/python3.9/site-packages/torch/distributed/distributed_c10d.py", line 1287, in all_reduce
    work = group.allreduce([tensor], opts)
RuntimeError: Tensors must be CUDA and dense
```

### To Reproduce

```python
import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        self.trainer.estimated_stepping_batches  # Can be used here to define the LR scheduler 


def run(ckpt_path=None):

    model = BoringModel()
    trainer = Trainer(
        num_sanity_val_steps=0,
        max_epochs=1,
        gpus=2,
        logger=False,
        enable_checkpointing=False
    )


if __name__ == "__main__":
    run()

```

### Environment
- PyTorch Lightning Version (e.g., 1.5.0): [master ](https://github.com/PyTorchLightning/pytorch-lightning/commit/7ee690758ccad7f702460d056f6369c1d4371a46)

cc @justusschock @kaushikb11 @awaelchli @akihironitta @rohitgr7

SeanNaren:
## What does this PR do?

Closes #12317.

This is a solution to fixing `estimated_stepping_batches` for the `DeepSpeedStrategy`.

This solution is optimal, however, touches internals which may prove error-prone in the future. The reason we do not use `model_to_device` is that we do not want to move the entire model to the device when using DeepSpeed Stage 3. We rely on DeepSpeed to manage device sharding/assignment.

## Before submitting

- [x] Did you **update the [CHANGELOG](https://github.com/Lightning-AI/lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

## Did you have fun?

cc @borda @SeanNaren @awaelchli @rohitgr7 @akihironitta

### DIFF
--- a/src/pytorch_lightning/strategies/deepspeed.py
+++ b/src/pytorch_lightning/strategies/deepspeed.py
@@ -357,7 +357,8 @@ class DeepSpeedStrategy(DDPStrategy):
 
     def setup(self, trainer: "pl.Trainer") -> None:
         self.accelerator.setup(trainer)
-        self.model_to_device()
+        # we set the device so that optimizers can be created with distributed comms.
+        self.lightning_module._device = self.root_device
         self.setup_optimizers(trainer)
         self.setup_precision_plugin()
         optimizers_to_device(self.optimizers, self.root_device)
--- a/src/pytorch_lightning/strategies/deepspeed.py
+++ b/src/pytorch_lightning/strategies/deepspeed.py
@@ -357,6 +357,7 @@ class DeepSpeedStrategy(DDPStrategy):
 
     def setup(self, trainer: "pl.Trainer") -> None:
         self.accelerator.setup(trainer)
+        self.model_to_device()
         self.setup_optimizers(trainer)
         self.setup_precision_plugin()
         optimizers_to_device(self.optimizers, self.root_device)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -232,6 +232,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed issue where the CLI fails with certain torch objects ([#13153](https://github.com/PyTorchLightning/pytorch-lightning/pull/13153))
 
 
+- Fixed `estimated_stepping_batches` requiring distributed comms in `configure_optimizers` for the `DeepSpeedStrategy` ([#13350](https://github.com/PyTorchLightning/pytorch-lightning/pull/13350))
+
+
 -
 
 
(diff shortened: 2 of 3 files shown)


src/pytorch_lightning/strategies/deepspeed.py
```diff
@@ -357,6 +357,8 @@ def setup_distributed(self):
 
     def setup(self, trainer: "pl.Trainer") -> None:
         self.accelerator.setup(trainer)
+        # we set the device so that optimizers can be created with distributed comms.
+        self.lightning_module._device = self.root_device
```
justusschock, line 361: does this only apply to deepspeed or could this be relevant to other strategies as well (e.g. sharded)? Just wondering if we should do this for all strategies

justusschock (approved):

carmocca (approved):

src/pytorch_lightning/strategies/deepspeed.py
SeanNaren: Looking at other plugins, there isn't as much of a worry as they can move the model to device fine! I think its just the deepspeed plugin that has issues (maybe bagua? but seems the original PR took care)

Issue #10535, opened 2021-11-15
[RFC] Deprecate `auto_select_gpus` from Trainer constructor


daniellepintz:
## Proposed refactoring or deprecation

Deprecate `auto_select_gpus` from Trainer constructor:
https://github.com/PyTorchLightning/pytorch-lightning/blob/7a9a08c5d3ca4699fb439f691c40e1320b37507a/pytorch_lightning/trainer/trainer.py#L209-L212

### Motivation

Users can achieve the same functionality by setting the `accelerator` and `devices` flags - we can simplify by only having one way of doing this.

cc @justusschock @awaelchli @rohitgr7 @tchaton @kaushikb11 @borda

kaushikb11:
@daniellepintz We do have users using it. For instance, https://github.com/PyTorchLightning/pytorch-lightning/issues/10456.
It could be pretty convenient when users are running multiple experiments on bunch of GPUs.

daniellepintz:
@kaushikb11 can this same functionality not be achieved with `accelerator` and `devices` flags?

rharish101:
@daniellepintz Currently, I think that if we want 3 available GPUs, then we need to specify `devices=3` and `accelerator='gpu'`. However, this means that GPUs [0, 1, 2] are selected, even if one of them is fully used by another process:
https://github.com/PyTorchLightning/pytorch-lightning/blob/ffb1a754af5301608c71955c12468216e920af51/pytorch_lightning/utilities/device_parser.py#L76

However, with `auto_select_gpus`, it chooses 3 GPUs that are available. This will be [0, 1, 2] if all three are available, else it will also check GPUs 3, 4, etc. until it finds 3 available GPUs.

PR #13356, opened 2022-06-21
Deprecate `auto_select_gpus` and enable it by default

### LINKED ISSUE #13079: RFC: Deprecate  `auto_select_gpus` Trainer argument
## Proposed refactor

Deprecate the `Trainer(auto_select_gpus=True|False)` option and just enable it always when `devices=int` gets selected. 

### Motivation

- The feature is misleading, we don't actually have an algorithm to select the gpus automatically in a smart way. A confusion by user was raised recently: #13012
- The flag only applies to the GPU accelerator. There are no equivalent flags for tpus, ipus, etc.
- What auto_select_gpus does is such a niche use case that it's almost not worth talking about. The implementation just runs through all available GPUs and tests if it can place a tensor in memory. This essentially tests wether the GPU is in exclusive mode or not. [See the implementation here](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/tuner/auto_gpu_select.py).
- Internally, the flag is framed as "tuning" as the functions are under the tuner module, but this does not fall under the term "tuning" in my opinion.

### Pitch

Keep the feature, but simply enable it by default and remove the flag from the Trainer. It is still useful to have the check against exclusivity, for example on managed clusters, and I can't think of a reason why it would be undesired. This would only apply when passing `devices=int` and not when indices get passed. 

### Alternatives

Keep it and come up with sophisticated methods to determine whether a GPU should be selected or not based on memory profile, utilization etc. This IMO is not feasible because not enough information is available before training.

### Additional context

#13012

awaelchli:
## What does this PR do?

Fixes #13079, #10535

Before
```py
# Picks the first two available GPUs
Trainer(devices=2, auto_select_gpus=True)
```

Now:
```py
# Picks the first two available GPUs
Trainer(devices=2)
```

Yes. The `auto_select_gpus` argument in Trainer will be removed in v1.9. Users of `auto_select_gpus` can just remove the flag and still make use of it as it is now turned on by default.

## Before submitting

- [ ] Did you **update the [CHANGELOG](https://github.com/Lightning-AI/lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

## Did you have fun?


src/pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -517,7 +515,7 @@ def _set_parallel_devices_and_init_accelerator(self) -> None:
         self._gpus = self._devices_flag if not self._gpus else self._gpus
         self._tpu_cores = self._devices_flag if not self._tpu_cores else self._tpu_cores
 
-        self._set_devices_flag_if_auto_select_gpus_passed()
+        self._set_devices_flag_if_count_passed()
```
awaelchli, line 518: We could merge this with `Accelerator.parse_devices` (see line below). I.e., we would implement `GPUAccelerator.parse_devices` by calling the `pick_multiple_gpus` function there.

src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -49,20 +49,6 @@ def determine_root_gpu_device(gpus: List[_DEVICE]) -> Optional[_DEVICE]:
     return root_gpu
 
 
-def _parse_devices(
```
awaelchli, line 52: removed dead code

carmocca:
Is this blocked or something?

awaelchli:
This just needs more work. Requires to look at tests etc. Need to discuss priority and then add to sprint.

PR #13405, opened 2022-06-24
Add support for DDP fork

### LINKED ISSUE #7550: DDP spawn no longer works in Jupyter environment

### LINKED ISSUE #8230: [Feature Request] Support forking for DDP
## 🚀 Feature
PyTorch Lightning DDP uses `spawn` for creating trainer processes for each GPU.  This issue is a request to allow creating with forking instead.

### Motivation
This feature request is primarily motivated by training graph neural networks for node classification/link prediction on a large graph.  A common strategy is to load the entire graph into CPU memory (if the CPU memory can hold but GPU cannot) for fast access of the graph structure as well as node/edge feature sfor neighbor sampling or subgraph sampling.

However, for multi-GPU training one needs to create multiple processes, each for one GPU, and they need to perform read-only access to the same graph and features.  If the processes are created with `spawn`, each process will have its own (redundant) replica of the graph and features, which linearly scales up the memory consumption with the number of processes spawned.

`fork`, on the other hand, allows the subprocesses to inherit the same in-memory data structure from the parent process via copy-on-write, thereby saving memory.

An example of using `fork` for multi-GPU training is [DGL's Multi-GPU GraphSAGE implementation](https://github.com/dmlc/dgl/blob/master/examples/pytorch/graphsage/train_sampling_multi_gpu.py).  Note the use of `mp.Process` to create the subprocesses.  I believe this functionality can benefit other frameworks that require other GNN frameworks too, e.g. PyTorch Geometric.

### Pitch

Add a new option `ddp_fork` as a new trainer accelerator.  Something like:

```python
# train on 8 GPUs, fork-based (same machine (ie: node))
trainer = Trainer(gpus=8, accelerator='ddp_fork')
```

### Alternatives

Putting the data into shared memory is also an option.  However under certain circumstances (e.g. Docker containers in some production environment) shared memory is quite limited and cannot hold a large object.

cc @borda @akihironitta

awaelchli:
## What does this PR do?

Fixes #7550
Fixes #8230

Adds support for DDP Fork. This version of ddp can be used in Jupyter notebooks with GPU unlike ddp_spawn!!!
Usage:

```py
trainer = Trainer(
    accelerator="gpu",
    devices=2,
    strategy="ddp_fork",
)
```

See the added docs for a comparison of ddp_spawn vs. ddp_fork

Remarks:
- Thanks for the advice and support by @justusschock 
- DDP fork should be faster to launch the processes. We could consider replacing ddp_spawn with ddp_fork in some tests to speed up CI, but one should be careful as process forking is considered unsafe in general. 

**Q: Why did we need to replace almost all instances of `torch.cuda.device_count()` and `torch.cuda.is_available()`?**
**A:** These function calls unfortunately create a CUDA context, i.e., they init the CUDA memory and tie it to the current process. Once this happens, we won't be able to re-initialize the CUDA anymore in the forked processes. This is a limitation of torch + forking. 

**Q: Now that we support two different start methods in DDPSpawnStrategy, shouldn't we rename the strategy?**
**A:** Yes technically we should. Especially since TPUSpawnStrategy is also using the fork start method and does not even support spawn. However, renaming everything here is too premature, considering that in the longterm the strategies DDPSpawn and DDP will get merged eventually anyway. What we can do in a follow up is renaming the internal launcher classes `_SpawnLauncher` and `_XLASpawnLauncher`, and  associated terminology in docs, comments, etc. 

## Open questions

- Should DDP fork become the default when `strategy=None` and `devices>1` in a Jupyter notebook?
  - Result: Can be done in follow up by modifying acc connector
- Investigate forkserver.
   - Result: Can't be supported in Jupyter. But user can select it if they want in the launcher. 
  
## Follow-up work

- Renaming internal classes: #13743
- Adding alias ddp_notebook #13744
- Set ddp_notebook as the default in Jupyter #13746
- More and cleaner error handling in connector (TODO)

No known ones.

## Before submitting

- [x] Did you **update the [CHANGELOG](https://github.com/Lightning-AI/lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7 @justusschock @kaushikb11 @awaelchli @akihironitta


src/pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -46,16 +46,16 @@ class _SpawnLauncher(_Launcher):
...
-        self._start_method = "spawn"
+        self._start_method = start_method
 
     @property
     def is_interactive_compatible(self) -> bool:
-        # The start method 'spawn' is currently the only one that works with DDP and CUDA support
-        # The start method 'fork' is the only one supported in Jupyter environments but not compatible with CUDA
-        # For more context, see https://github.com/Lightning-AI/lightning/issues/7550
-        return self._start_method == "fork" and self._strategy.root_device.type != "cuda"
+        # The start method 'spawn' is not supporrted in interactive environments
```
awaelchli, line 55:
```suggestion
        # The start method 'spawn' is not supported in interactive environments
```

src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -330,3 +332,13 @@ def parse_hpus(devices: Optional[Union[int, str, List[int]]]) -> Optional[int]:
...
+
+
+def num_cuda_devices() -> int:
+    with multiprocessing.Pool(1) as pool:
+        return pool.apply(torch.cuda.device_count)
+
+
+def is_cuda_available() -> bool:
+    with multiprocessing.Pool(1) as pool:
+        return pool.apply(torch.cuda.is_available)
```
justusschock, line 344:
```suggestion
# the following functions are run in a separate process to avoid creating 
# a cuda context in the main process where we later might want to fork. 
def num_cuda_devices() -> int:
    with multiprocessing.Pool(1) as pool:
        return pool.apply(torch.cuda.device_count)


def is_cuda_available() -> bool:
    with multiprocessing.Pool(1) as pool:
        return pool.apply(torch.cuda.is_available)
```

src/pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         ddp_comm_hook: Optional[callable] = None,
         ddp_comm_wrapper: Optional[callable] = None,
         process_group_backend: Optional[str] = None,
+        start_method: str = "spawn",
```
justusschock, line 71: with another start method, this isn't really ddp_*spawn* anymore, is it? :D

justusschock (approved):

src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -330,3 +332,13 @@ def parse_hpus(devices: Optional[Union[int, str, List[int]]]) -> Optional[int]:
         raise MisconfigurationException("`devices` for `HPUAccelerator` must be int, string or None.")
 
     return int(devices) if isinstance(devices, str) else devices
+
+
+def num_cuda_devices() -> int:
```
justusschock, line 337:
```suggestion
# the following methods are run in a separate process to avoid creating a
# cuda-context in the main process where we later might want to fork processes
def num_cuda_devices() -> int:
```

src/pytorch_lightning/strategies/ddp_spawn.py
awaelchli: Yes, but I would not rename the DDPSpawn strategy class yet. Long-term, I believe it will eventually be merged together with the DDP strategy. The names in the registry remain consistent though, so we will have "ddp_spawn" and "ddp_fork" as string names.

What we should do is rename the SpawnLauncher. It's a protected API so should be no problem there.

src/pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -280,17 +282,20 @@ def post_training_step(self):
...
-        strategy_registry.register(
-            cls.strategy_name,
-            cls,
-            description=f"{cls.__class__.__name__}",
-        )
+        for start_method in ("spawn", "fork"):
+            strategy_registry.register(
+                f"ddp_{start_method}_find_unused_parameters_false",
+                cls,
+                description="DDPSpawn Strategy with `find_unused_parameters` as False",
```
carmocca, line 289: This is definitely confusing haha

I would name it DDPFork already even though it doesn't exist in the implementation

src/pytorch_lightning/strategies/launchers/xla_spawn.py
```diff
@@ -51,8 +51,7 @@ class _XLASpawnLauncher(_SpawnLauncher):
     """
 
     def __init__(self, strategy: "Strategy") -> None:
-        super().__init__(strategy)
-        self._start_method = "fork"
```
carmocca, line 55: TIL XLASpawn is actually XLAFork. How confusing!

src/pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -613,7 +613,11 @@ def _check_strategy_and_fallback(self) -> None:
                 f"You selected strategy to be `{DDPFullyShardedNativeStrategy.strategy_name}`, "
                 "but GPU accelerator is not used."
             )
-
+        if strategy_flag in ("ddp_fork", "ddp_fork_find_unused_parameters_false") and not hasattr(os, "fork"):
```
carmocca, line 616: Do we need to hardcode these options again? Can't we extract them from the registry?

src/pytorch_lightning/utilities/device_parser.py
carmocca: Such comments should be part of the docstring IMO

src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -330,3 +333,27 @@ def parse_hpus(devices: Optional[Union[int, str, List[int]]]) -> Optional[int]:
...
+    if not hasattr(os, "fork"):
+        return torch.cuda.device_count()
+    with multiprocessing.get_context("fork").Pool(1) as pool:
+        return pool.apply(torch.cuda.device_count)
+
+
+def is_cuda_available() -> bool:
+    """Returns a bool indicating if CUDA is currently available.
+
+    Unlike :func:`torch.cuda.is_available`, this function will do its best not to create a CUDA context to for fork
```
carmocca, line 353: If we didn't do this. Would there be failures or bugs or is it just a matter of leaving leftover CUDA context using memory?

carmocca (approved):
Has there been a recent change that suddenly enabeld this?

And can you open separate issues for the leftover items mentioned in the description?

Great job!

src/pytorch_lightning/strategies/launchers/xla_spawn.py
awaelchli: Yes exactly :)) The terminology was a bit fuzzy until recently, at least in the torch world. If we look at the `mp.spawn` function here, it also used to support two options spawn and fork despite being grouped under the name "spawn": https://pytorch.org/docs/stable/multiprocessing.html#torch.multiprocessing.spawn

src/pytorch_lightning/utilities/device_parser.py
awaelchli: If we didn't do this, we would see an immediate error upon launching processes:

> RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method

src/pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: Hmm, it would be nice if we could avoid it, but there is no elegant way to do this right now. We could add a method to the StrategyRegistry that returns us a view of the registered items for a particular class only. 

Or we could naively filter the keys for the substring "fork", if that's what you meant?

docs/source-pytorch/accelerators/gpu_intermediate.rst
```diff
@@ -199,6 +201,61 @@ You can then call your scripts anywhere
...
+   :header-rows: 1
+
+   * -
+     - DDP
+     - DDP Spawn
+     - DDP Fork
+   * - Works in Jupyter notebooks / IPython environments
+     - No
+     - Yes
+     - No
```
otaj, line 232: isn't this mixed up? as in, DDP and DDP Spawn don't work in Jupyter whereas DDP Fork does.

rohitgr7 (commented):
still reviewing

src/pytorch_lightning/trainer/connectors/accelerator_connector.py
carmocca: > We could add a method to the StrategyRegistry that returns us a view of the registered items for a particular class only.

This is what I meant. But fine to leave it for latter if you prefer.

kaushikb11 (approved):
Awesome! We should set it as a strategy after detecting the interactive environment.

docs/source-pytorch/accelerators/gpu_intermediate.rst
awaelchli: Yes, thank you! Should be swapped

docs/source-pytorch/accelerators/gpu_intermediate.rst
```diff
@@ -199,6 +201,61 @@ You can then call your scripts anywhere
...
+     - Linux, Mac, Win
+     - Linux, Mac, Win
+     - Linux, Mac
+   * - Requires all objects to be picklable
+     - No
+     - Yes
+     - No
+   * - Is the guard ``if "__name__"=="__main__"`` required?
+     - Yes
+     - Yes
```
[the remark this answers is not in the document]
awaelchli: Yes it is required. 
And no, ddp_spawn does not work in notebooks, which is the motivation for introducing fork.

src/pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -41,21 +42,30 @@ class _SpawnLauncher(_Launcher):
...
+              the Windows platform.
     """
 
-    def __init__(self, strategy: Strategy) -> None:
+    def __init__(self, strategy: Strategy, start_method: Literal["spawn", "fork"] = "spawn") -> None:
         self._strategy = strategy
-        self._start_method = "spawn"
+        self._start_method = start_method
+        if start_method == "fork" and not hasattr(os, "fork"):
+            raise ValueError("The start method 'fork' is not available on this platform. Use 'spawn' instead.")
```
kaushikb11, line 61: @awaelchli Is there a case where this error would be thrown?
As we are validating it in Accelerator Connector and the error would be thrown there

kaushikb11: @justusschock It does check for platform compatibly in Accelerator Connector
https://github.com/Lightning-AI/lightning/blob/feature/ddp-fork2/src/pytorch_lightning/trainer/connectors/accelerator_connector.py#L616

justusschock: Didn't see that. But theoretically the launcher could be reused for other third-party plugins as well, so I think it is good to error out here again

SeanNaren (approved):
Amazing @awaelchli, have I made the right assumption that it was the combination of `start_processes` from `torch.multiprocessing` and ensuring that the CUDA context is only created within the processes first (not in the main process) that allowed this to be possible?

src/pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -613,7 +613,11 @@ def _check_strategy_and_fallback(self) -> None:
...
+            raise ValueError(
+                f"You selected `Trainer(strategy='{strategy_flag}')` but process forking is not supported on this"
+                f" platform. We recommed to choose `Trainer(strategy='ddp_spawn')` instead."
```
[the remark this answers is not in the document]
awaelchli: We could yes. I chose spawn here because it's the closest to fork.

awaelchli:
@SeanNaren Thanks and yes, that's exactly the main reason why and how.

src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -330,3 +332,27 @@ def parse_hpus(devices: Optional[Union[int, str, List[int]]]) -> Optional[int]:
         raise MisconfigurationException("`devices` for `HPUAccelerator` must be int, string or None.")
 
     return int(devices) if isinstance(devices, str) else devices
+
+
+def num_cuda_devices() -> int:
```
tchaton, line 337: Neat ! This was the secret ;)

tchaton (approved):
Looks awesome! Great work and intuition ;)

src/pytorch_lightning/trainer/connectors/accelerator_connector.py
awaelchli: @kaushikb11 A different error would show up when attempting to launch with .fit(), pointing that the start method is not supported. We can generalize and improve the error handling further, but I think this is good enough and catches the most common use cases for 99% of the users. I will leave this as fine tuning for a follow up.

src/pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -69,13 +78,19 @@ def launch(self, function: Callable, *args: Any, trainer: Optional["pl.Trainer"]
                 a selected set of attributes get restored in the main process after processes join.
             **kwargs: Optional keyword arguments to be passed to the given function.
         """
+        if self._start_method not in mp.get_all_start_methods():
```
rohitgr7, line 81: also raise this during init?

rohitgr7 (approved):

src/pytorch_lightning/strategies/launchers/spawn.py
```diff
@@ -34,27 +35,35 @@ class _SpawnLauncher(_Launcher):
...
+          the launcher is invoked. E.g., one should avoid creating cuda tensors or calling ``torch.cuda.*`` functions
+          before calling ``Trainer.fit``.
 
     Args:
         strategy: A reference to the strategy that is used together with this launcher.
+        start_method: The method how to start the processes.
+            - 'spawn': The default start method. Requires all objects to be pickleable.
+            - 'fork': Preferrable for IPython/Jupyter environments where 'spawn' is not available. Not available on
+              the Windows platform for example.
+            - 'forkserver': Alternative implementation to 'fork'.
```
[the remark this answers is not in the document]
awaelchli: Nope, does not work well with interactive. I only list forkserver for completeness. Custom strategies/launchers can use it if they want, but I don't want to expose it in any other way right now.

src/pytorch_lightning/strategies/launchers/spawn.py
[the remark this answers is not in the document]
awaelchli: Yes, in general we can't just fork a process on the cpu if we already have data in cuda, like tensors etc. Which is why they say it's not supported. Hence also the note in the docs that you shouldn't call any cuda related functions before calling trainer.fit(). However, in Lightning we can control all of this and ensure we create processes before we setup the model on GPU or move any data there. 

This is the reason why https://github.com/Lightning-AI/lightning/issues/7550 was open for so long, because until recently Lightning didn't spawn processes early enough for this to become possible.

src/pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -280,17 +283,20 @@ def post_training_step(self):
...
-            cls,
-            description="DDPSpawn Strategy with `find_unused_parameters` as False",
-            find_unused_parameters=False,
-        )
-        strategy_registry.register(
-            cls.strategy_name,
-            cls,
-            description=f"{cls.__class__.__name__}",
-        )
+        for start_method in ("spawn", "fork"):
```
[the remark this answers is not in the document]
awaelchli: Nope, not going to expose this right now. forkserver is not well known and not going to help in Jupyter. So unless someone asks for it, let's not expose it.

src/pytorch_lightning/strategies/launchers/spawn.py
awaelchli: I prefer not to. It makes the validation in accelerator connector easier, as it allows the instantiation to happen independently.

src/pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -71,6 +72,7 @@ def __init__(
         ddp_comm_wrapper: Optional[callable] = None,
         process_group_backend: Optional[str] = None,
         timeout: Optional[timedelta] = default_pg_timeout,
+        start_method: Literal["spawn", "fork", "forkserver"] = "spawn",
```
akihironitta, line 75: Should forkserver be removed from the type:
```suggestion
        start_method: Literal["spawn", "fork"] = "spawn",
```
following https://github.com/Lightning-AI/lightning/pull/13405/files#r925506674?

awaelchli: I don't understand. If someone wants to use it in a custom strategy, they can but would otherwise get a type error.

Borda (approved):

PR #13501, opened 2022-07-01
Fix FullyShardedDataParallel wrapped models not being unwrapped, leading to incorrect checkpoints

### LINKED ISSUE #13500: FullyShardedDataParallel wrapped models not being unwrapped, leading to incorrect checkpoints.
## 🐛 Bug

When training in parallel with the `fsdp` strategy, the saved checkpoints are somehow messed up. When I try to resume training from those, the epoch number is properly resumed, but the loss spikes dramatically, like as if it went back to an initial/random state. When I do the same train/checkpoint/resume loop with `ddp_sharded` I do not have this issue and the checkpoint resumes with a similar loss to where it left off. I further saw that when I point a model with strategy `fsdp` at a checkpoint saved with `ddp_sharded` it also resumes with a reasonable loss that is roughly at the previous level. This suggests that `fsdp` loads a checkpoint ok, but there is something wrong with how it saves checkpoints in parallel. Conversely when I resume using `ddp_sharded` from an `fsdp` saved checkpoint, the loss is dramatically worse as if weights were randomly initialized, further suggesting that the issue is with how weights are saved in `fsdp`. Knowing all of this, I am able to just switch to using `ddp_sharded` but this seems like a really nasty bug that could cause other people headaches so I wanted to make sure it was known.

The fix seems to be to make sure to unwrap the `FullyShardedDataParallel` wrapper. One key difference between the `fsdp` strategy implementation and the `ddp_sharded` strategy implementation is that `ddp_sharded` overrides `self.lightning_module` and does calls a custom `unwrap_...` function which unwraps the `ShardedDataParallel` layer prior to calling the shared `unwrap_lightning_module(...)` function. `fsdp` doesn't do any of this, and it defaults to the method implemented in `ParallelStrategy.lightning_module` which only calls the `unwrap_lightning_module(...)` function. 

I am going to open a PR and link it here which makes `unwrap_lightning_module(...)` aware of  `FullyShardedDataParallel` (both flavors) as well as `ShardedDataParallel` so that all of the strategies that use one of those wrappers would benefit. Also in the future hopefully that will be a piece of code that is noticed which needs to be modified as new wrappers are added.

### To Reproduce

1. Train a model in parallel that saves checkpoints for a few epochs, use `--strategy fsdp`. Note the loss at the beginning and make sure it drops.
2. Resume a model using any strategy from one of those saved checkpoints and note that the loss is similar to the beginning of training. Based on code I would guess that the `fsdp native` strategy, whatever that is called, is also broken. Maybe others.
3. Repeat 1,2 but this time use `--strategy ddp_sharded` and note that the loss resumes from where it left off.

### Expected behavior

Model training continues when resuming.

### Environment

* CUDA:
	- GPU:
		- NVIDIA A100 80GB PCIe
		- NVIDIA A10
		- NVIDIA A10
	- available:         True
	- version:           11.5
* Packages:
	- numpy:             1.22.3
	- pyTorch_debug:     False
	- pyTorch_version:   1.11.0
	- pytorch-lightning: 1.6.3
	- tqdm:              4.64.0
* System:
	- OS:                Linux
	- architecture:
		- 64bit
		- ELF
	- processor:         x86_64
	- python:            3.9.12
	- version:           #100-Ubuntu SMP Fri Sep 24 14:50:10 UTC 2021

### Additional context

cc @SeanNaren @awaelchli @rohitgr7 @akihironitta

jstjohn:
## What does this PR do?

Possibly fixes #13500, needs more testing.

Potential breaking changes (unsure because behavior looks ok on my single node multi-gpu setup):

- In the code block below in FSDP, it's possible that this change will change the interaction between the optimizer and the model and sharding. The model does train and weights are optimized, but I do not have any large language model type tests to make sure that sharding is working properly. Prior to this change, the outer Shard wrapper would not have been unwrapped, while other things would be unwrapped? **It is worth checking that these changes do not break sharding for large language models**, however the current behavior is arguably worse, the checkpoints themselves are likely corrupted in some way.
```python
    def configure_ddp(self) -> None:
        log.detail(f"{self.__class__.__name__}: configuring FSDP... (cpu_offload: [{self.cpu_offload}])")
        if not self.cpu_offload:
            # When using CPU Offload, FSDP will manage the CUDA movement for us.
            # Note: this would be problematic for large model (which could not fit in one GPU)
            # as FSDP module.to(device) would first summon all parameters
            # (TODO: need to figure out solution)
            self.model_to_device()
        # setup optimizers after fully sharded has wrapped the lightning module
        self.setup_optimizers(self.lightning_module.trainer)
```

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -271,6 +271,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Fixed a bug where checkpoints were not properly saved with `FSDP` strategy models ([#13500](https://github.com/Lightning-AI/lightning/issues/13500))
+
 
 - Improved support for custom `DataLoader`s when instantiated in `*_dataloader` hook ([#12981](https://github.com/PyTorchLightning/pytorch-lightning/pull/12981))
 
--- a/src/pytorch_lightning/overrides/base.py
+++ b/src/pytorch_lightning/overrides/base.py
@@ -17,9 +17,11 @@ import torch
 import torch.nn as nn
 from torch.nn import DataParallel
 from torch.nn.parallel import DistributedDataParallel
-
 import pytorch_lightning as pl
 from pytorch_lightning.core.mixins import DeviceDtypeModuleMixin
+from pytorch_lightning.overrides.fairscale import _FAIRSCALE_AVAILABLE
+from pytorch_lightning.utilities import _FAIRSCALE_FULLY_SHARDED_AVAILABLE
+from pytorch_lightning.utilities.imports import _TORCH_GREATER_EQUAL_1_11
 
 
 class _LightningPrecisionModuleWrapperBase(DeviceDtypeModuleMixin, torch.nn.Module):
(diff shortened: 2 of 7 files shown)


awaelchli:
@jstjohn Thanks for looking into this. 
The unwrapping logic is a bit overcomplicated and as you found a source for bugs. I propose to completely remove it and simplify it greatly. Here is the proposal #13502. Not sure whether that would solve your bug as well but just wanted to point out the related issue here.

jstjohn:
> @jstjohn Thanks for looking into this. The unwrapping logic is a bit overcomplicated and as you found a source for bugs. I propose to completely remove it and simplify it greatly. Here is the proposal #13502. Not sure whether that would solve your bug as well but just wanted to point out the related issue here.

@awaelchli your solution seems great. Should I close this PR and let you take this on with that approach? I am also wondering whether or not any of the tests should have caught this? I see tests in place to make sure that weights match after checkpointing but I don't know if they cover the parallel case and also it looks like they only run on pytorch 1.12?

jstjohn:
Closing this PR since I like the solution outlined in https://github.com/Lightning-AI/lightning/issues/13502 better, and will let @awaelchli execute on that.

PR #13534, opened 2022-07-04
Fix mypy typing errors in pytorch_lightning/strategies/single_tpu.py

Cyprien-Ricque:
## What does this PR do?

Fixes mypy typing errors in pytorch_lightning/strategies/single_tpu.py in https://github.com/Lightning-AI/lightning/issues/13445.

This could be a breaking change in case I missed something/relative to older lightning versions.

in ```SingleTPUStrategy``` class, in function ```setup(self, trainer: "pl.Trainer")```, the call to ```self.model.on_post_move_to_device()``` has been removed because the condition to trigger it cannot be ```True```. 
The condition is ```is_overridden("on_post_move_to_device", self.lightning_module)```. To be ```True```, among other things, this condition requires the parent class to contain the ```on_post_move_to_device``` method. This ```on_post_move_to_device``` method is not present in any of the parent classes listed in ```is_overridden```, therefore the condition cannot be ```True```. (parents classes being ```LightningModule```, ```LightningDataModule``` and ```Callback```).

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/strategies/single_tpu.py
+++ b/src/pytorch_lightning/strategies/single_tpu.py
@@ -19,7 +19,6 @@ from pytorch_lightning.plugins.io.xla_plugin import XLACheckpointIO
 from pytorch_lightning.plugins.precision import PrecisionPlugin
 from pytorch_lightning.strategies.single_device import SingleDeviceStrategy
 from pytorch_lightning.utilities import _TPU_AVAILABLE, find_shared_parameters, set_shared_parameters
-from pytorch_lightning.utilities.model_helpers import is_overridden
 
 if _TPU_AVAILABLE:
     import torch_xla.core.xla_model as xm
--- a/src/pytorch_lightning/strategies/single_tpu.py
+++ b/src/pytorch_lightning/strategies/single_tpu.py
@@ -55,13 +55,10 @@ class SingleTPUStrategy(SingleDeviceStrategy):
         return False
 
     def setup(self, trainer: "pl.Trainer") -> None:
+        assert self.model, "self.model must be set before find_shared_parameters(self.model)"
         shared_params = find_shared_parameters(self.model)
         self.model_to_device()
-        if is_overridden("on_post_move_to_device", self.lightning_module):
-            self.model.on_post_move_to_device()
-        else:
-            set_shared_parameters(self.model, shared_params)
-
+        set_shared_parameters(self.model, shared_params)
         super().setup(trainer)
 
         if self.debug:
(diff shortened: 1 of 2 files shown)


src/pytorch_lightning/strategies/single_tpu.py
```diff
@@ -55,13 +55,10 @@ def is_distributed(self) -> bool:
         return False
 
     def setup(self, trainer: "pl.Trainer") -> None:
+        assert self.model, "self.model must be set before find_shared_parameters(self.model)"
         shared_params = find_shared_parameters(self.model)
         self.model_to_device()
-        if is_overridden("on_post_move_to_device", self.lightning_module):
-            self.model.on_post_move_to_device()
-        else:
-            set_shared_parameters(self.model, shared_params)
```
justusschock, line 63: why did you remove this logic? Never change existing logic for the sake of typing!

src/pytorch_lightning/strategies/single_tpu.py
```diff
@@ -70,9 +67,6 @@ def setup(self, trainer: "pl.Trainer") -> None:
         self.tpu_local_core_rank = xm.get_local_ordinal()
         self.tpu_global_core_rank = xm.get_ordinal()
 
-    def model_to_device(self) -> None:
-        self.model.to(self.root_device)
```
justusschock, line 74: why remove this?

justusschock (changes requested):

src/pytorch_lightning/strategies/single_tpu.py
Cyprien-Ricque: Sure I correct.

Put aside the fact that I don't have to do it in this PR, I changed it because as explained in the PR, It appeared to me that this ```if``` cannot be ```True```

src/pytorch_lightning/strategies/single_tpu.py
Cyprien-Ricque: Because it looked like code duplication since the function is the exact same as the one in the parent class ```SingleDeviceStrategy```. 
I noticed it because the mypy error is the same as the one in the parent class which is fixed in https://github.com/Lightning-AI/lightning/pull/13532.
Do I have to put it back?

awaelchli: Either way is fine with me but keep in mind we would prefer focus on fixing the typing errors and not mix different types of refactors :) Otherwise LGTM

awaelchli (approved):

justusschock (approved):

Borda (approved):

Issue #12521, opened 2022-03-30
Remove deprecated code after the 1.6 release


akihironitta:
## Proposed refactor

Remove deprecated code after the 1.6 release.

**NOTE**: Please pick up a _single_ item from the list (by commenting here in the issue) - and if there are no conflicts - we will happily assign you and put your name in front of the item in the list.

Please note that unless mentioned, the classes are importable from `pytorch_lightning`, example: `from pytorch_lightning import Trainer`.

* [x] `LightningModule.summarize` -> #12559
* [x] `pytorch_lightning.core.memory.LayerSummary` -> #12593
* [x] `pytorch_lightning.core.memory.ModelSummary` -> #12593
* [x] `pytorch_lightning.core.memory.get_gpu_memory_map` -> #12644
* [x] `pytorch_lightning.core.memory.get_memory_profile` -> #12659
* [x] `LightningModule.model_size` -> #12641
* [x] `LightningDataModule.train_transforms` -> #12662
* [x] `LightningDataModule.val_transforms` -> #12763
* [x] `LightningDataModule.test_transforms` -> #12773
* [x] `LightningDataModule.size` -> #12780
* [x] `LightningDataModule.dims` and `LightningDataModule(dims=...)` -> #12780
* [x] `LightningModule.get_progress_bar_dict` -> #12839 
* [x] `Trainer.progress_bar_dict` -> #12839 
* [x] `Trainer(prepare_data_per_node=...)` -> #12536
* [x] `Trainer(stochastic_weight_avg=...)` -> #12535
* [x] `Trainer(terminate_on_nan=...)` and `Trainer.terminate_on_nan` -> #12553 
* [x] `LightningModule.on_{train,val,test,predict}_dataloader` -> #13033
* [x] `pytorch_lightning.loggers.TestTubeLogger` -> #12859
* [x] `pytorch_lightning.Callback.on_keyboard_interrupt` -> #13438
* [x] `Trainer(process_position=...)` -> #13071
* [x] `Trainer(flush_logs_every_n_steps=...)` -> #13074
* [x] `LightningModule.add_to_queue` -> @shenoynikhil
* [x] `LightningModule.get_from_queue` -> @shenoynikhil
* [x] `Trainer(progress_bar_refresh_rate=...)` -> #12514
* [x] `LightningLoggerBase.close` and `pytorch_lightning.loggers.LoggerCollection.close` -> #13149
[...]

PR #13544, opened 2022-07-05
Revert "Fix mypy typing errors in pytorch_lightning/strategies/single_tpu.py"

Cyprien-Ricque:
Reverts Lightning-AI/lightning#13534

In the previous PR I explained that the condition ```is_overridden("on_post_move_to_device", self.lightning_module)``` can not be True. This is incorrect. I missed the heritage of ```LightningModule``` with ```ModelHooks```.
Indeed, ```ModelHooks``` implements ```on_post_move_to_device```.
I am very sorry for this.

I think I did not see it in the unit tests because this function has been deprecated in v1.5 and will be removed in v1.7, therefore it is not tested. According to this test: https://github.com/Lightning-AI/lightning/blob/61c28cb428a13c2aea6d7f3f55e0f00431a4ea4e/tests/tests_pytorch/deprecated_api/test_remove_1-7.py#L64

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,6 +72,7 @@ module = [
     "pytorch_lightning.strategies.parallel",
     "pytorch_lightning.strategies.sharded",
     "pytorch_lightning.strategies.sharded_spawn",
+    "pytorch_lightning.strategies.single_tpu",
     "pytorch_lightning.strategies.tpu_spawn",
     "pytorch_lightning.strategies.strategy",
     "pytorch_lightning.profilers.advanced",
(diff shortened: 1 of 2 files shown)


justusschock:
Probably no need to revert it, as #13534 will be merged to 1.7 only and won't be included in 1.6.X

akihironitta:
@Cyprien-Ricque Thank you for your action! The change shouldn't have been part of your previous PR #13534, but as mentioned above, there's no need to revert the change because my PR #13548 will remove it anyway as part of #12521 and the change won't be included in 1.6.x. Could you close this PR?

PR #13570, opened 2022-07-07
fix mypy typing errors in pytorch_lightning/strategies/horovod.py

Cyprien-Ricque:
## What does this PR do?

Fixes mypy typing errors in pytorch_lightning/strategies/horovod.py in https://github.com/Lightning-AI/lightning/issues/13445.

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/strategies/hivemind.py
+++ b/src/pytorch_lightning/strategies/hivemind.py
@@ -315,6 +315,7 @@ class HiveMindScheduler:
         # implemented custom logic which we would not want to call on destruction of the `HiveMindScheduler`
         self.__dict__ = {k: v for k, v in scheduler.__dict__.items() if k not in ("step", "__del__")}
 
+        self.base_lrs = scheduler.base_lrs
         self.optimizer = optimizer
         self.scheduler = scheduler
         self.current_step = -1
--- a/src/pytorch_lightning/strategies/hivemind.py
+++ b/src/pytorch_lightning/strategies/hivemind.py
@@ -309,13 +309,13 @@ class HiveMindScheduler:
 
     This code ensures that we only step when the HiveMind optimizer reaches the global step.
     """
+    base_lrs: List[float]
 
     def __init__(self, optimizer: "hivemind.Optimizer", scheduler: _LRScheduler) -> None:
         # copy most of the `Scheduler` methods into this instance. `__del__` is skipped in case the scheduler has
         # implemented custom logic which we would not want to call on destruction of the `HiveMindScheduler`
         self.__dict__ = {k: v for k, v in scheduler.__dict__.items() if k not in ("step", "__del__")}
 
-        self.base_lrs = scheduler.base_lrs
         self.optimizer = optimizer
         self.scheduler = scheduler
         self.current_step = -1
--- a/src/pytorch_lightning/strategies/horovod.py
+++ b/src/pytorch_lightning/strategies/horovod.py
@@ -154,8 +154,12 @@ class HorovodStrategy(ParallelStrategy):
         else:
             hvd.join()
 
-    def reduce(self, tensor: Union[Any, Tensor], group: Optional[Any] = None,
-               reduce_op: Optional[Union[ReduceOp, str]] = "mean") -> Union[Any, Tensor]:
+    def reduce(
+        self,
+        tensor: Union[Any, Tensor],
+        group: Optional[Any] = None,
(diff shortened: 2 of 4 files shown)


src/pytorch_lightning/strategies/horovod.py
```diff
@@ -129,27 +133,33 @@ def _unpack_lightning_optimizer(opt):
...
+    def join(self) -> None:
         if self.root_device.type == "cuda":
             hvd.join(self.local_rank)
         else:
             hvd.join()
 
-    def reduce(self, tensor, group: Optional[Any] = None, reduce_op: Optional[Union[ReduceOp, str]] = "mean"):
+    def reduce(
+        self,
+        tensor: Union[Any, Tensor],
```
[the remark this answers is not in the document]
Borda: yes `Union[Any, Tensor]` -> `Any` but it is confusing as `tensor` argument can be any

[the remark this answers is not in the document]
akihironitta: I left a similar comment in a PR quite a while ago, but I remember we concluded that we'd continue to have types like this instead of merging them to `Any`. (cc @justusschock if I remember it correctly)

otaj (approved):

src/pytorch_lightning/strategies/hivemind.py
```diff
@@ -315,6 +315,7 @@ def __init__(self, optimizer: "hivemind.Optimizer", scheduler: _LRScheduler) ->
         # implemented custom logic which we would not want to call on destruction of the `HiveMindScheduler`
         self.__dict__ = {k: v for k, v in scheduler.__dict__.items() if k not in ("step", "__del__")}
 
+        self.base_lrs = scheduler.base_lrs
```
carmocca, line 318: This should be already set from the `__dict__` override 2 lines above

carmocca (approved):

src/pytorch_lightning/strategies/hivemind.py
otaj: That is true, but I think mypy does not pick it up since it's set by `__dict__` override

justusschock: But then we probably should just add stubs before the init instead of doing that again

src/pytorch_lightning/strategies/horovod.py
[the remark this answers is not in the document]
justusschock: The advantage that way is that actually most IDEs nowadays parse the typehints and that way you do get autocompletion for tensors which in many cases is the correct type instead of just doing Any, where the IDE doesn't know what to suggest. So yeah, technically this is redundant, but it really helps with delevoper experience

src/pytorch_lightning/strategies/horovod.py
```diff
@@ -111,8 +113,10 @@ def _unpack_lightning_optimizer(opt):
         lr_scheduler_configs = self.lr_scheduler_configs
         for config in lr_scheduler_configs:
             scheduler = config.scheduler
+            assert isinstance(scheduler, _LRScheduler), "ReduceLROnPlateau does not have attribute 'base_lr'"
```
awaelchli, line 116: The assertion does not relate to the attribute base_lr. 
The class of the scheduler is also not ReduceLROnPlateau in general. 
I suggest you remove the comment

src/pytorch_lightning/strategies/horovod.py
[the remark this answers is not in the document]
awaelchli: I am in favor of having Any, Tensor

awaelchli (approved):

src/pytorch_lightning/strategies/horovod.py
Cyprien-Ricque: Sorry, indeed I made a typo, it is ```base_lrs```.
According to the type of ```scheduler``` being ```Union[_LRScheduler, ReduceLROnPlateau]``` If the type is not ```_LRScheduler``` then it is ```ReduceLROnPlateau```, right? 
Do you mean that it is also used with other schedulers not listed is that ```Union```?
In that case shouldn't we update the ```scheduler``` type in the data class ```LRSchedulerConfig```?

src/pytorch_lightning/strategies/hivemind.py
Cyprien-Ricque: I did it. Thank you for the advice!

Issue #13445, opened 2022-06-29
Improve typing coverage (4/n)


otaj:
## 🚀 Typing coverage

Let's improve typing coverage of PyTorch Lightning together!

I'm creating a new issue in order to increase visibility. There are three older issues (#7037, #5023, #4698) which became stale over time.

### Plan

Currently, there are 55 files which are excluded from mypy checks so that our CI does not fail. These files vastly differ in difficulty in order to make the typing complete. For this reason, we are introducing difficulty estimate for each file so that community members can choose to work on the files appropriate to their skill level.

Please, comment on this issue in order to reserve a particular file to work on. Once you do so, I will edit this top comment to avoid collisions. Once you think your work is finished, please open a PR referencing this issue which:

 * removes the corresponding line from `pyproject.toml`
 * and passes `mypy` checks with the corresponding line removed. You can test it locally by running `mypy` from root directory

If you are struggling with pushing it over the finish line, open the PR anyway and someone from our team will help you to get it there. 🚀

Please note, that it can happen that you may need to edit more than just one file. This is fine, but please keep in mind, that the goal of your PR will be to make the check passing for the chosen file. Also, please note that the difficulty is just an educated guess.

For those of you who are not familiar with the process of contributing a PR, we have prepared a [simple guide](https://devblog.pytorchlightning.ai/quick-contribution-guide-86d977171b3a) that will walk you through the necessary steps. You can do it! :rocket: :muscle: 

#### Difficulty 3 of 3

- [x] `pytorch_lightning/callbacks/progress/rich_progress.py` @donlapark #14963 

### Completed

<details>

#### Difficulty 1 of 3

- [x] `pytorch_lightning/core/decorators.py` #14044
- [x] `pytorch_lightning/profilers/advanced.py` @nninept #13792 
~- [ ] `pytorch_lightning/profilers/base.py` @LeeChanHyuk #13879~
- [x] `pytorch_lightning/loggers/base.py` @JustinGoheen #13494 
- [x] `pytorch_lightning/__setup__.py` @CyprienRicque #13472 
~- [ ] `pytorch_lightning/distributed/dist.py` @puhuk #13492~
- [x] `pytorch_lightning/strategies/single_device.py` @CyprienRicque #13532 
- [x] `pytorch_lightning/trainer/optimizers.py` @gautierdag #13470
- [x] `pytorch_lightning/utilities/distributed.py` @krishnakalyan3 #13678
[...]

PR #13603, opened 2022-07-11
Fix mypy errors attributed to `pytorch_lightning.core.module.py`

JustinGoheen:
## What does this PR do?

Fix mypy errors attributed to `pytorch_lightning.core.module.py` for issue #13445

## Before submitting

## PR review


src/pytorch_lightning/core/module.py
```diff
@@ -1563,10 +1563,7 @@ def lr_scheduler_step(self, scheduler, optimizer_idx, metric):
                 scheduler.step(epoch=self.current_epoch)
 
         """
-        if metric is None:
-            scheduler.step()
-        else:
-            scheduler.step(metric)
+        scheduler.step(metric)
```
JustinGoheen, line 1566: @rohitgr7 will this change break anything? mypy is complaining about ln: 1567 of the previous implementation

rohitgr7: not always.. currently it should be passed only when user is using `ReduceLROnPlateau`.

For others the signature can vary: https://pytorch.org/docs/stable/_modules/torch/optim/lr_scheduler.html

src/pytorch_lightning/utilities/types.py
```diff
@@ -29,6 +29,8 @@
 from torchmetrics import Metric
 from typing_extensions import Protocol, runtime_checkable
 
+import pytorch_lightning
```
JustinGoheen, line 32: add to support LRSchedulerPLType

src/pytorch_lightning/utilities/types.py
```diff
@@ -131,6 +133,9 @@ def no_sync(self) -> Generator:
 LRSchedulerTypeTuple = (torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau)
 LRSchedulerTypeUnion = Union[torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau]
 LRSchedulerType = Union[Type[torch.optim.lr_scheduler._LRScheduler], Type[torch.optim.lr_scheduler.ReduceLROnPlateau]]
+LRSchedulerPLType = Union[
```
JustinGoheen, line 136: added to support return type when `lr_schedulers` encounters:

```python
        if len(lr_schedulers) == 1:
            return lr_schedulers[0]
```

src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -76,7 +76,7 @@ def apply_to_collection(
     dtype: Union[type, Any, Tuple[Union[type, Any]]],
     function: Callable,
     *args: Any,
-    wrong_dtype: Optional[Union[type, Tuple[type]]] = None,
+    wrong_dtype: Optional[Union[type, Tuple[type], Tuple]] = None,
```
JustinGoheen, line 79: resolves

```sh
src/pytorch_lightning/core/module.py:371: error: Argument "wrong_dtype" to "apply_to_collection" has incompatible type "Tuple[Type[Number], Type[Metric], Type[Tensor], Type[Dict[Any, Any]]]"; expected "Union[type, Tuple[type], None]"  [arg-type]
```

src/pytorch_lightning/core/module.py
```diff
@@ -440,7 +446,7 @@ def log(
             logger=logger,
             on_step=on_step,
             on_epoch=on_epoch,
-            reduce_fx=reduce_fx,
+            reduce_fx=reduce_fx,  # type: ignore[arg-type]
```
JustinGoheen, line 449: the implementation of `module.log` (ln319) allows the user to pass in a str or Callable for arg `reduce_fx`; overriding the _Metadata dataclass in `pl.trainer.connectors.logger_connector`, which uses `reduce_fx: Callable = torch.mean`.

updating _Metadata type hints did not resolve the issue - it caused more. 

 an alternative to ignoring the error would be to use `pl.trainer.connectors.logger_connector._Metadata._parse_reduce_fx` in `pl.core.module` before calling `results.log` at ln 441.

src/pytorch_lightning/core/module.py
```diff
@@ -93,7 +98,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         torch._C._log_api_usage_once(f"lightning.module.{self.__class__.__name__}")
 
         # pointer to the trainer object
-        self.trainer: Optional["pl.Trainer"] = None
+        self.trainer = None
```
JustinGoheen, line 101: I removed the annotation here because the annotation was causing errors.

This breaks from convention found in other modules, where there is an `Optional["pl.Trainer"]`; however, self.trainer is used in `core.module` in a way that attribute errors occur when self.Trainer is `None`

src/pytorch_lightning/core/module.py
JustinGoheen: reverted to original and ignored

src/pytorch_lightning/core/module.py
```diff
@@ -49,12 +49,20 @@
...
+    LRSchedulerPLType,
+    LRSchedulerTypeUnion,
+    STEP_OUTPUT,
+)
 from pytorch_lightning.utilities.warnings import WarningCache
 
 warning_cache = WarningCache()
 log = logging.getLogger(__name__)
 
+MODULE_OPTIMIZERS = Union[Optimizer, LightningOptimizer, List[Optimizer], List[LightningOptimizer]]
```
JustinGoheen, line 64: added to correct ln 134, 137, 149

src/pytorch_lightning/core/module.py
```diff
@@ -175,12 +179,13 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
 
     @property
     def trainer(self) -> "pl.Trainer":
-        if not self._running_torchscript and self._trainer is None:
+        if self._running_torchscript and self._trainer is not None:
```
JustinGoheen, line 182: inverted logic to correct:

```sh
src/pytorch_lightning/core/module.py:188: error: Incompatible return value type (got "Optional[Trainer]", expected "Trainer")  [return-value]
```

awaelchli: But it's not correct. Most of the time we are not running with torchscript but we have the trainer reference not None. 

I guess it should be

```py
        if self._running_torchscript or self._trainer is None:
            raise RuntimeError(...)
        return self._trainer
```

src/pytorch_lightning/core/module.py
```diff
@@ -175,12 +179,13 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
...
+            return self._trainer
+        else:
             raise RuntimeError(f"{self.__class__.__qualname__} is not attached to a `Trainer`.")
-        return self._trainer
 
     @trainer.setter
-    def trainer(self, trainer: Optional["pl.Trainer"]) -> None:
+    def trainer(self, trainer: "pl.Trainer") -> None:
```
awaelchli, line 188: see below the check whether trainer is None. 
cc @carmocca

src/pytorch_lightning/core/module.py
```diff
@@ -230,12 +235,12 @@ def local_rank(self) -> int:
...
 
     @property
-    def on_gpu(self):
+    def on_gpu(self) -> bool:
         """Returns ``True`` if this model is currently located on a GPU.
 
         Useful to set flags around the LightningModule for different CPU vs GPU behavior.
         """
-        return self.device.type == "cuda"
+        return self.device.type == "cuda"  # type: ignore[union-attr]
```
awaelchli, line 243: I think we should investigate instead why we annotated `DeviceDtypeModuleMixin.device` with `Union[str, torch.device]`. Let's promote it to `torch.device` without str?

src/pytorch_lightning/core/module.py
```diff
@@ -1577,7 +1586,7 @@ def lr_scheduler_step(self, scheduler, optimizer_idx, metric):
 
         """
         if metric is None:
-            scheduler.step()
+            scheduler.step()  # type: ignore[call-arg]
```
awaelchli, line 1589: I think we should raise an error here when `metric=None` and the scheduler is `ReduceOnPlateau` instead of ignoring the type error. 
cc @carmocca

src/pytorch_lightning/core/module.py
```diff
@@ -1737,7 +1746,7 @@ def tbptt_split_batch(self, batch, split_size):
                 if isinstance(x, Tensor):
                     split_x = x[:, t : t + split_size]
                 elif isinstance(x, collections.Sequence):
-                    split_x = [None] * len(x)
+                    split_x = [None] * len(x)  # type: ignore[assignment]
```
awaelchli, line 1749: would this be solved if we just wrote it as a list comprehension (the loop below)?

src/pytorch_lightning/core/module.py
```diff
@@ -1775,15 +1784,15 @@ def unfreeze(self) -> None:
...
+    def _verify_is_manual_optimization(self, fn_name: str) -> None:
         if self.automatic_optimization:
             raise MisconfigurationException(
                 f"to use {fn_name}, please disable automatic optimization:"
                 " set model property `automatic_optimization` as False"
             )
 
     @classmethod
-    def _auto_collect_arguments(cls, frame=None) -> Tuple[Dict, Dict]:
+    def _auto_collect_arguments(cls, frame=None) -> Tuple[Dict, Dict]:  # type: ignore[no-untyped-def]
```
awaelchli, line 1795:
```suggestion
    def _auto_collect_arguments(cls, frame: Optional[types.FrameType] = None) -> Tuple[Dict, Dict]:

File: src/pytorch_lightning/utilities/types.py (hunk 3 above)
Comment (awaelchli) [reply src/pytorch_lightning/utilities/types.py:136]: I don't understand this

File: src/pytorch_lightning/core/module.py (hunk 8 above)
Comment (JustinGoheen) [reply src/pytorch_lightning/core/module.py:182]: If left as the original, the mypy error will raise. Should I just ignore the error?

File: src/pytorch_lightning/utilities/types.py (hunk 3 above)
Comment (JustinGoheen) [reply src/pytorch_lightning/utilities/types.py]: cleaned in https://github.com/Lightning-AI/lightning/pull/13603/commits/205e546805833061d225c8237d3a39d40cbbd0dc

File: src/pytorch_lightning/core/module.py (hunk 8 above)
Comment (awaelchli) [reply src/pytorch_lightning/core/module.py:182]: If you apply my suggestion, there won't be an error. I couldn't see one.

File: src/pytorch_lightning/core/module.py (hunk 10 above)
Comment (JustinGoheen) [reply src/pytorch_lightning/core/module.py:243]: I think anything that accepts a str variant of a device or metric was done so to allow the user to pass in a str

File: src/pytorch_lightning/core/module.py (hunk 12 above)
Comment (JustinGoheen) [reply src/pytorch_lightning/core/module.py:1750]: I used

```python
split_x = [x[batch_idx][t : t + split_size] for batch_idx in range(len(x))]
```

and got

```sh
src/pytorch_lightning/core/module.py:1750: error: Incompatible types in assignment (expression has type "List[Any]", variable has type "Tensor")  [assignment]
```

src/pytorch_lightning/core/module.py
```diff
@@ -1792,7 +1792,7 @@ def _verify_is_manual_optimization(self, fn_name: str) -> None:
             )
 
     @classmethod
-    def _auto_collect_arguments(cls, frame=None) -> Tuple[Dict, Dict]:  # type: ignore[no-untyped-def]
+    def _auto_collect_arguments(cls, frame: Optional[types.FrameType] = None) -> Tuple[Dict, Dict]:
```
JustinGoheen: @awaelchli this caused different errors

```sh
rc/pytorch_lightning/core/module.py:1811: error: Item "None" of "Optional[FrameType]" has no attribute "f_back"  [union-attr]
src/pytorch_lightning/core/module.py:1811: error: Argument 1 to "collect_init_args" has incompatible type "Union[FrameType, None, Any]"; expected "FrameType"  [arg-type]
```

src/pytorch_lightning/core/module.py
JustinGoheen: previous convo with @rohitgr7 https://github.com/Lightning-AI/lightning/pull/13603#discussion_r918291207

rohitgr7: this is not possible since this is an internal call that is made based on the config returned from `configure_optimizers`.
and the user is expected to override this hook rather than calling it explictly.

src/pytorch_lightning/core/module.py
```diff
@@ -265,8 +271,10 @@ def logger(self) -> Optional[Logger]:
         """Reference to the logger object in the Trainer."""
         # this should match the implementation of `trainer.logger`
         # we don't reuse it so we can properly set the deprecation stacklevel
+        if self.trainer is None:
+            return None
```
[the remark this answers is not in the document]
JustinGoheen: should I just ignore the return-value error?

src/pytorch_lightning/utilities/types.py
```diff
@@ -131,6 +131,7 @@ def no_sync(self) -> Generator:
 LRSchedulerTypeTuple = (torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau)
 LRSchedulerTypeUnion = Union[torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau]
 LRSchedulerType = Union[Type[torch.optim.lr_scheduler._LRScheduler], Type[torch.optim.lr_scheduler.ReduceLROnPlateau]]
+LRSchedulerPLType = Union[_LRScheduler, ReduceLROnPlateau, Any]
```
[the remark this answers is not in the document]
JustinGoheen:
```sh
src/pytorch_lightning/core/module.py:179: error: Incompatible return value type (
got "List[Union[pytorch_lightning.utilities.types._LRScheduler, pytorch_lightning.utilities.types.ReduceLROnPlateau]]",
expected "Union[torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau, List[Union[torch.optim.lr_scheduler._LRScheduler, torch.optim.lr_scheduler.ReduceLROnPlateau]], pytorch_lightning.utilities.types._LRScheduler, pytorch_lightning.utilities.types.ReduceLROnPlateau, None]")  [return-value]
```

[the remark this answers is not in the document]
JustinGoheen: removing Any and replacing with the following resolves the error:

```python
LRSchedulerPLType = Union[
    _LRScheduler, ReduceLROnPlateau, List[Union[_LRScheduler, ReduceLROnPlateau], _LRScheduler, ReduceLROnPlateau, None]
]
```

src/pytorch_lightning/core/module.py
```diff
@@ -106,7 +115,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._current_fx_name: Optional[str] = None
         self._automatic_optimization: bool = True
         self._truncated_bptt_steps: int = 0
-        self._param_requires_grad_state = {}
+        self._param_requires_grad_state: Dict = {}
```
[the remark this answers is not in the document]
JustinGoheen: Optimizer parameters (str) and if the param requires grad (bool).

set with:

```python
  param_requires_grad_state = {}
  for opt in self.trainer.optimizers:
      for group in opt.param_groups:
          for param in group["params"]:
              # If a param already appear in param_requires_grad_state, continue
              if param in param_requires_grad_state:
                  continue
              param_requires_grad_state[param] = param.requires_grad
              param.requires_grad = False

  # Then iterate over the current optimizer's parameters and set its `requires_grad`
  # properties accordingly
  for group in optimizer.param_groups:  # type: ignore[union-attr]
      for param in group["params"]:
          param.requires_grad = param_requires_grad_state[param]
  self._param_requires_grad_state = param_requires_grad_state
```

src/pytorch_lightning/core/module.py
```diff
@@ -175,12 +180,13 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
 
     @property
     def trainer(self) -> "pl.Trainer":
-        if not self._running_torchscript and self._trainer is None:
+        if self._running_torchscript or self._trainer is None:
```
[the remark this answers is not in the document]
JustinGoheen: original logic causes mypy to raise

```sh
src/pytorch_lightning/core/module.py:186: error: Incompatible return value type (got "Optional[Trainer]", expected "Trainer")  [return-value]
```

src/pytorch_lightning/utilities/types.py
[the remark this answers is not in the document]
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/38aeef8fe492040b3a50a2d4647340f66c1bde45

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/38aeef8fe492040b3a50a2d4647340f66c1bde45

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
JustinGoheen: updated to match default branch and ignored the return-value error in https://github.com/Lightning-AI/lightning/pull/13603/commits/cf4aeb1d1eed408075c932e4eccb67d62779206b

src/pytorch_lightning/core/module.py
JustinGoheen: @awaelchli I updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/049ace94f1fb7e3137c95d4e82c10e15e249b8e2

I had to add (at ln1810):

```python
frame_args = collect_init_args(frame.f_back, [])  # type: ignore[union-attr, arg-type]
```

because updating `pytorch_lightning.utilities.parsing.collect_init_args` to accept an Optional[types.FrameType] caused more mypy errors than it resolved.

src/pytorch_lightning/core/module.py
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/4ddf8415d562021c0bfd191bccda89d29d66d36f

src/pytorch_lightning/core/module.py
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/d59f1ccd467dc743c4d438744396c9c8c2983ff0

src/pytorch_lightning/core/mixins/device_dtype_mixin.py
```diff
@@ -39,7 +39,7 @@ def dtype(self, new_dtype: Union[str, torch.dtype]) -> None:
         raise RuntimeError("Cannot set the dtype explicitly. Please use module.to(new_dtype).")
 
     @property
-    def device(self) -> Union[str, torch.device]:
+    def device(self) -> torch.device:
```
JustinGoheen, line 42: updated to resolve:

```sh
src/pytorch_lightning/core/module.py:245: error: Item "str" of "Union[str, device]" has no attribute "type"  [union-attr]
```

src/pytorch_lightning/overrides/data_parallel.py
```diff
@@ -65,7 +65,7 @@ def forward(self, *inputs: Any, **kwargs: Any) -> Any:
         output = super().forward(*inputs, **kwargs)
 
         def output_transform(data: Any) -> Any:
-            device = cast(torch.device, self.module.device)
+            device = self.module.device
```
JustinGoheen, line 68: resolved redundant cast after setting DeviceDtypeModuleMixin.device to torch.device

src/pytorch_lightning/core/mixins/device_dtype_mixin.py
awaelchli: This makes sense, thx

src/pytorch_lightning/core/module.py
awaelchli: Perhaps this should go into the types utility module? Not sure. Let's ask @carmocca for opinion.

src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -76,7 +76,7 @@ def apply_to_collection(
     dtype: Union[type, Any, Tuple[Union[type, Any]]],
     function: Callable,
     *args: Any,
-    wrong_dtype: Optional[Union[type, Tuple[type, ...]]] = None,
+    wrong_dtype: Optional[Union[type, Tuple[type], Tuple]] = None,
```
awaelchli, line 79: Let's revert this. I don't see any errors locally. This must have been due to a conflict with changes from master.

awaelchli (commented):
Great progress!

src/pytorch_lightning/core/module.py
```diff
@@ -177,13 +182,15 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
     def trainer(self) -> "pl.Trainer":
         if not self._running_torchscript and self._trainer is None:
             raise RuntimeError(f"{self.__class__.__qualname__} is not attached to a `Trainer`.")
-        return self._trainer
+        if self._trainer is not None:
+            return self._trainer
```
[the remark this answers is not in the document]
JustinGoheen: I have to check locally however I'm pretty certain I did this to avoid a return-value error

src/pytorch_lightning/core/module.py
```diff
@@ -1795,14 +1804,14 @@ def _auto_collect_arguments(cls, frame=None) -> Tuple[Dict, Dict]:
             self_arguments: arguments dictionary of the first instance
             parents_arguments: arguments dictionary of the parent's instances
         """
+
         if not frame:
             frame = inspect.currentframe()
 
-        frame_args = collect_init_args(frame.f_back, [])
-        self_arguments = frame_args[-1]
+        frame_args = collect_init_args(frame.f_back, [])  # type: ignore[union-attr, arg-type]
```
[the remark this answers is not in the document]
awaelchli: Turns out this code is never used. Removing it in #14015.

src/pytorch_lightning/core/module.py
```diff
@@ -177,13 +182,15 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
...
+        if self._trainer is not None:
+            return self._trainer
 
     @trainer.setter
     def trainer(self, trainer: Optional["pl.Trainer"]) -> None:
         for v in self.children():
             if isinstance(v, LightningModule):
-                v.trainer = trainer
+                if trainer is not None:
+                    v.trainer = trainer
```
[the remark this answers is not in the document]
JustinGoheen: same as previous comment regarding return-value error.

src/pytorch_lightning/core/module.py
```diff
@@ -266,7 +273,7 @@ def logger(self) -> Optional[Logger]:
         # this should match the implementation of `trainer.logger`
         # we don't reuse it so we can properly set the deprecation stacklevel
         if self._trainer is None:
-            return
+            return  # type: ignore[return-value]
```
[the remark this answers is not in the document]
JustinGoheen: see https://github.com/Lightning-AI/lightning/pull/13603#discussion_r929958803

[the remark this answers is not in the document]
awaelchli: I don't understand what the problem is here. There is nothing to ignore here. Mypy just wants you to write explicitly None, instead of an implicit blank return.

src/pytorch_lightning/core/module.py
```diff
@@ -604,7 +613,7 @@ def forward(self, *args, **kwargs) -> Any:
         """
         return super().forward(*args, **kwargs)
 
-    def training_step(self, *args, **kwargs) -> STEP_OUTPUT:
+    def training_step(self, *args: Any, **kwargs: Any) -> STEP_OUTPUT:
```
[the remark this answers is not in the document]
JustinGoheen: I agree with above and would add that a follow-on type hints project be added to create additional bespoke Lightning types to help uniformity - as is reasonably possible - across files for explicitly named args and kwargs.

Also noting that these types are pushed to docs, under the parameters block; which raises the question of - should Dict, List, Mappings be left simple - as Dict and not Dict[str, List[...]] or some other variation - for the sake of brevity in docs.

src/pytorch_lightning/core/module.py
carmocca: If we see an use anywhere other than this file, I'd say yes.

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
carmocca: I would go with `Any` for now. It is weird to leave this unannotated when every other method in `ModelHooks` has annotations that might also be undesired. To me, it's an all-or-nothing decision.

I wonder if things could be improved by using `typing.overloads` so that we don't have to just allow `*args, **kwargs`. But the simplest solution right now is to add `Any` now

src/pytorch_lightning/core/module.py
rohitgr7: let's keep it in `types.py` anyway. It might be hard to find if we need it again for other modules in the future.

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
rohitgr7: @carmocca added this.

what's up with `_running_torchscript`? if it's True and trainer is None, should it return None or raise an error?

src/pytorch_lightning/core/module.py
```diff
@@ -569,7 +575,9 @@ def log_grad_norm(self, grad_norm_dict):
         """
         self.log_dict(grad_norm_dict, on_step=True, on_epoch=True, prog_bar=False, logger=True)
 
-    def all_gather(self, data: Union[Tensor, Dict, List, Tuple], group: Optional[Any] = None, sync_grads: bool = False):
+    def all_gather(
+        self, data: Union[Tensor, Dict, List, Tuple], group: Optional[Any] = None, sync_grads: bool = False
+    ) -> Tensor:
```
rohitgr7, line 580: I don't think it returns `Tensor`. Should return the type of data.

rohitgr7 (commented):
It looks good, but there are some ignores which need some investigation. Will do.

src/pytorch_lightning/core/module.py
JustinGoheen: by `type of data`, do you mean `Union[Tensor, Dict, List, Tuple]`

rohitgr7: yes

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
carmocca: The running torchscript hack is required by a bug in PyTorch's jit (https://github.com/pytorch/pytorch/issues/67146), which makes `mypy` get confused. But this would never return `None` normally.

My suggestion is to undo this change and add an ignore for this error. Hopefully they fix this bug eventually.

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
carmocca: No, this is caused by a different issue. Mypy doesn't seem to recognize that `v.trianer = trauber` is going to call `v`'s trainer setter, which has `Optional[Trainer]` as the input annotation.

Instead, it uses the output annotation of the trainer getter. In my eyes, this is a mypy bug and I suggest ignoring it. Haven't checked their issue tracker though

src/pytorch_lightning/core/module.py
```diff
@@ -291,15 +296,14 @@ def _apply_batch_transfer_handler(
...
+        def call_hook(hook_name: str, *args: Any) -> Callable:
             if self._trainer:
                 datahook_selector = self._trainer._data_connector._datahook_selector
                 obj = datahook_selector.get_instance(hook_name)
-                trainer_method = (
-                    self._trainer._call_lightning_module_hook
-                    if isinstance(obj, self.__class__)
-                    else self._trainer._call_lightning_datamodule_hook
-                )
+                if isinstance(obj, self.__class__):
```
JustinGoheen, line 303: this is the same logic, it was changed because mypy was complaining 

```sh
src/pytorch_lightning/core/module.py:308: error: Cannot call function of unknown type  [operator]
```

src/pytorch_lightning/core/module.py
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/b9ed56e9b7898b61b6491292412a4ec1dd0b8df7

JustinGoheen:
apologies, I'm no longer able to keep this PR up to date.

src/pytorch_lightning/core/module.py
awaelchli: I couldn't do it because of circular imports.

src/pytorch_lightning/core/module.py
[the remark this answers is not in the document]
awaelchli: I vote for keeping it unannotated.

awaelchli (approved):

awaelchli:
> apologies, I'm no longer able to keep this PR up to date.

No worries, I will address reviews and finish it. There shouldn't be too much work left here. 
@rohitgr7 @carmocca

rohitgr7 (approved):

Borda (approved):

PR #13614, opened 2022-07-12
Introduce ServableModuleValidator Callback

### LINKED ISSUE #13480: Add Inference support to PyTorch Lightning
## 🚀 Feature

From my current investigation, it seems TorchServe has matured and it is now the default Server used within Sagemaker AWS PyTorch Image.

Until now, PyTorch Lightning has offered very limited support for serving and the transition from training to production has been quite challenging for many users.

This issue is meant to start a conversation on how PyTorch Lightning can smooth the transition between those worlds.  

My proposition is the following:

Add a new hook to the LightningModule called `configure_serve_handler` where users can return a [BaseHandler](https://pytorch.org/serve/_modules/ts/torch_handler/base_handler.html#BaseHandler) subclass to accommodate their models' serving requirements.

Here is how it could look like:

```py
class MNISTClassifier(LightningModule):

    def configure_serve_handler(self):
        return MNISTDigitClassifierHandler(...) 
```

Here are some opportunities available with such an introduction:

* Sanity serve checking when starting training. 
* Sanity TorchScripting when starting training.
* Auto-generate an handler.py file for the users without the pytorch lightning dependency.
* Auto-generate mar file with torchscripted model, handlers, ....
* Provide DataType to help users to write down serialization and deserialization layers for their models or upstream to Serve.

Furthermore, using the newly introduced Lightning App, it would be possible to provide even more to PyTorch Lightning users by automating all those steps and then serving their model to any services of their choice.  

### Additional context

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @borda @tchaton @justusschock @awaelchli @carmocca @ananthsub @ninginthecloud @jjenniferdai @rohitgr7

tchaton:
## What does this PR do?

Fixes #13480

**Context**: Improve model mobility from training to serving. MLOps Engineers would argue that a production model shouldn't be trained and resources wasted if the model can't be served to provide value to customers. They would argue that the model should be unit tested as soon as possible to validate its conformity for its production usage. This PR investigates the addition of serving functionalities in PyTorch Lightning.

As a user, you would need to add the `SanityServing` callback to your Trainer and make your model subclass `ServableModule`.

The `ServableModule` requires 3 hooks to be implemented in order to fully describe how the model behaves when being served.

* configure_payload: Returns an example of a payload object: Lightning can provide example payloads for images, text, videos, etc..
* configure_serialization: Provide serialization / deserialization methods. Lightning can provide common data types.
* serve_step: The logic to be performed when you have received a request.

```py
from typing import Dict
import torch
from pytorch_lightning import seed_everything, Trainer
from pytorch_lightning.serve import ServableModuleValidator, ServableModule
from pytorch_lightning.demos.boring_classes import BoringModel

class ServableBoringModel(BoringModel, ServableModule):
    def configure_payload(self) -> ...:
        return {"body": {"x": list(range(32))}}

    def configure_serialization(self):
        class Tensor:
            @staticmethod
            def deserialize(x):
                return torch.tensor(x).float()

            @staticmethod
            def serialize(x):
                return x.numpy().tolist()

        return {"x": Tensor.deserialize}, {"output": Tensor.serialize}

    def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
        return {"output": self.forward(x)}

trainer = Trainer(max_epochs=1, limit_train_batches=2, limit_val_batches=0, callbacks=[ServableModuleValidator()])
trainer.fit(ServableBoringModel())
```

The Ideal API would rely only on the type to apply the deserialization and serialization and tensor shape could be added:

```py

from pydantic import BaseModel
from typing import TypedDict

class Tensor(BaseModel):
    @staticmethod
    def deserialize(x):
        return torch.tensor(x).float()

    @staticmethod
    def serialize(x):
        return x.numpy().tolist()

class OutputTensor(TypedDict):
    output: Tensor[2]

class ServableBoringModel(BoringModel, ServableModule):
    def configure_payload(self) -> ...:
        return {"body": {"x": list(range(32))}}

    def serve_step(self, x: Tensor[32]) -> OutputTensor:
        return {"output": self.forward(x)}

callback = ServableModuleValidator(
       optimization="trace|script|onxx|tensor_rt|...",
       server='fastapi|torch_serve|ml_server|sagemaker|triton"
)
trainer = Trainer(max_epochs=1, limit_train_batches=2, limit_val_batches=0, callbacks=[callback])
trainer.fit(ServableBoringModel())
```

**Extra functionalities for sanity serving**

* Performance Testing. Latency and RPS
* Apply serving check at start and end of quantization and pruning.
* etc...

This would be even more impactful once https://github.com/pytorch/torchdynamo is available in PyTorch and models and their transforms served in pure python are further optimized.

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @justusschock @awaelchli @carmocca @ananthsub @ninginthecloud @jjenniferdai @rohitgr7 @akihironitta


src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class ServableModule(Module):
+    def configure_payload(self) -> Dict[str, Any]:
+        ...
+
+    def configure_inputs_outputs(self) -> Tuple[Dict[str, Callable], Dict[str, Callable]]:
+        ...
+
+    def serve_step(self, *args, **kwargs):
```
carmocca, line 22:
```suggestion
    def serve_step(self, *args: Any, **kwargs: Any) -> Any:
```

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+        class Tensor:
+            @staticmethod
+            def deserialize(x):
+                return torch.tensor(x).float()
+
+            @staticmethod
+            def serialize(x):
+                return x.numpy().tolist()
+
+        return ({"x": Tensor.deserialize}, {"output": Tensor.serialize})
```
carmocca, line 24: nit:

```suggestion
        return {"x": Tensor.deserialize}, {"output": Tensor.serialize}
```

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+
+class ServableBoringModel(BoringModel, ServableModule):
+    def configure_payload(self) -> ...:
+        return {"body": {"x": list(range(32))}}
+
+    def configure_inputs_outputs(self):
+        class Tensor:
+            @staticmethod
+            def deserialize(x):
+                return torch.tensor(x).float()
```
carmocca, line 18:
```suggestion
                return torch.tensor(x, dtype=torch.float)
```

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+
+            @staticmethod
+            def serialize(x):
+                return x.numpy().tolist()
```
carmocca, line 22: This should work

```suggestion
                return x.tolist()
```

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+
+        return ({"x": Tensor.deserialize}, {"output": Tensor.serialize})
+
+    def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
```
carmocca, line 26: Do you still plan to add support for serialization through annotations?

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+        return {"output": self.forward(x)}
```
carmocca, line 27: would it be better to call `self(x)` instead?

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+from pytorch_lightning import seed_everything, Trainer
+from pytorch_lightning.callbacks.sanity_serving import SanityServing, ServableModule
+from pytorch_lightning.demos.boring_classes import BoringModel
+
+
+class ServableBoringModel(BoringModel, ServableModule):
+    def configure_payload(self) -> ...:
+        return {"body": {"x": list(range(32))}}
+
+    def configure_inputs_outputs(self):
```
carmocca, line 14: I would suggest a name like `configure_serialization` instead. Inputs/outputs is a very broad term

tests/tests_pytorch/callbacks/test_sanity_serving.py
```diff
@@ -0,0 +1,43 @@
...
+    def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
+        return {"output": self.forward(x)}
+
+
+def test_sanity_serving():
+    seed_everything(42)
+    model = ServableBoringModel()
+    callback = SanityServing()
+    callback.on_train_start(None, model)
+    assert callback.resp.json() == {"output": [-7.451034069061279, 1.635885238647461]}
```
carmocca, line 35: Can you mock the numbers so the tests dont depend on the RNG?

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+
+    run(app, host=host, port=port, log_level="error")
+
+
+class SanityServing(Callback):
+
+    """This callback enables to validate a model is servable following the ServableModule API."""
+
+    @requires("fastapi")
+    @requires("uvicorn")
```
carmocca, line 65: Can we avoid this new decorator and instead use the `_RequirementAvailable` class we've recently started using for these? We are trying to standardize all these utilities into one: #12786 

In the past, William has been very against decorators.

example:
```python
FASTAPI_AVAILABLE = _RequirementAvailable("fastapi")

if not FASTAPI_AVAILABLE:
    ...
```

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self, method: Optional[str] = None):
+        super().__init__()
+        assert method in (None, "trace", "script")
+        if method in ("trace", "script"):
+            raise NotImplementedError(f"The method {method} is currently not supported.")
```
carmocca, line 70: Isn't this validation part of the `Process` implementation already?

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+        self.method = method
+        self.resp = None
+
+    @rank_zero_only
+    def on_train_start(self, trainer: "pl.Trainer", servable_model: "ServableModule"):
+        isinstance(servable_model, ServableModule)
+
+        process = Process(target=_start_server, args=(servable_model, "127.0.0.1", 8080, self.method))
```
carmocca, line 78: Should the host and port be part of the __init__?

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+    app = FastAPI()
+
+    inputs, outputs = servable_model.configure_inputs_outputs()
+
+    @app.get("/ping")
+    def ping() -> bool:
+        return True
+
+    @app.post("/predict")
+    async def predict(payload: dict = Body(...)) -> Dict[str, Any]:
```
carmocca, line 40: I don't think this should be called `predict`. It would be very confusing with the predict functionality that exists already

Why not `/serve`?

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+        body = payload["body"]
+
+        for key, deserializer in inputs.items():
+            body[key] = deserializer(body[key])
+
+        with torch.no_grad():
```
carmocca, line 46: Should this also do `servable_model.eval()`?

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+        ...
+
+    def configure_inputs_outputs(self) -> Tuple[Dict[str, Callable], Dict[str, Callable]]:
+        ...
+
+    def serve_step(self, *args, **kwargs):
+        ...
+
+
+def _start_server(servable_model: ServableModule, host: str, port: int, method: bool) -> None:
```
carmocca, line 26: Shouldn't this function be part of the Callback so it can be overridden?

src/pytorch_lightning/callbacks/sanity_serving.py
```diff
@@ -0,0 +1,92 @@
...
+
+        for key, serializer in outputs.items():
+            output[key] = serializer(output[key])
+
+        return output
+
+    run(app, host=host, port=port, log_level="error")
+
+
+class SanityServing(Callback):
```
carmocca, line 60: I'm not a fan of the name. What about `SimpleServer`?

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: Done !

tests/tests_pytorch/callbacks/test_sanity_serving.py
tchaton: Done !

tests/tests_pytorch/callbacks/test_sanity_serving.py
tchaton: Ideally yes, but I think there might be some difficulties to do so.

tests/tests_pytorch/callbacks/test_sanity_serving.py
tchaton: Done !

tests/tests_pytorch/callbacks/test_sanity_serving.py
tchaton: Done !

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: @williamFalcon was against decorator for the users, not internal usage. I don't like `_RequirementAvailable`, I would prefer requires if we need to uniformize something.

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: No, this is for model optimization.

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: Fine, but predict is the common name for it. Remember, this isn't a server, simply sanity checking. Users won't see it.

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: Done !

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: Done !

src/pytorch_lightning/callbacks/sanity_serving.py
tchaton: The name is associated with its function. It is doing sanity checking, not serving. The process is killed right after the requests has been made:

```py
        payload = servable_model.configure_payload()
        self.resp = requests.post(f"http://{self.host}:{self.port}/serve", json=payload)
        process.kill()
```

tests/tests_pytorch/callbacks/test_sanity_serving.py
tchaton: Done !

zippeurfou:
I am not sure I am a fan of the sanityServe to be a callback.
Here are a few reasons:
1. Serving and model training are two different things (separation of concern). You have multiple scenario where they are not tied as much as this (eg. serving as model composition, serving when you depend on different party)
2. Serving look at different performance metrics (eg. RPS, uptime...), you do care about theses benchmark. I am not sure tying it up to training step helps (this is not a now but how do I load test here? Or how would we help the users in the future do load testing?) which brings me to the next point.
3. Hardware / auto-scaling.. is important there. I am not sure we allow good separation there by putting it in a callback.

I am bringing this up because I think it's important SanityServe provides value otherwise no one will use it. However, to provide value we should think about what can it provide that you would not have in the Serve module.

zippeurfou:
Spoke with @tchaton async and we synced on it.

tests/tests_pytorch/serve/test_servable_module_validator.py
```diff
@@ -0,0 +1,44 @@
...
+        class Tensor:
+            @staticmethod
+            def deserialize(x):
+                return torch.tensor(x, dtype=torch.float)
+
+            @staticmethod
+            def serialize(x):
+                return x.tolist()
+
+        return {"x": Tensor.deserialize}, {"output": Tensor.serialize}
```
carmocca, line 24: Actually, since we are not using the class, this could be two local functions

```suggestion
        def deserialize(x):
            return torch.tensor(x, dtype=torch.float)

        def serialize(x):
            return x.tolist()

        return {"x": deserialize}, {"output": serialize}
```

tests/tests_pytorch/serve/test_servable_module_validator.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
+        assert torch.equal(x, torch.arange(32).float())
```
carmocca, line 27:
```suggestion
        assert torch.equal(x, torch.arange(32, dtype=torch.float))
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,145 @@
...
+            raise Exception("The provided model should be subclass of pytorch_lightning.server.ServableModule.")
+
+        if not is_overridden("configure_payload", servable_model, ServableModule):
+            raise Exception("The ``configure_payload`` method needs to be overridden.")
+
+        if not is_overridden("configure_serialization", servable_model, ServableModule):
+            raise Exception("The ``configure_serialization`` method needs to be overridden.")
+
+        if not is_overridden("serve_step", servable_model, ServableModule):
+            raise Exception("The ``serve_step`` method needs to be overridden.")
```
carmocca, line 82:
```suggestion
            raise NotImplementedError("The `configure_payload` method needs to be overridden.")

        if not is_overridden("configure_serialization", servable_model, ServableModule):
            raise NotImplementedError("The `configure_serialization` method needs to be overridden.")

        if not is_overridden("serve_step", servable_model, ServableModule):
            raise NotImplementedError("The `serve_step` method needs to be overridden.")
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,145 @@
...
+    @rank_zero_only
+    def on_train_start(self, trainer: "pl.Trainer", servable_model: "ServableModule"):
+        if isinstance(trainer.strategy, _NOT_SUPPORTED_STRATEGIES):
+            raise Exception(
+                f"The current strategy {trainer.strategy} used "
+                "by the trainer isn't supported for sanity serving yet."
+            )
+
+        if not isinstance(servable_model, ServableModule):
+            raise Exception("The provided model should be subclass of pytorch_lightning.server.ServableModule.")
```
carmocca, line 73:
```suggestion
            raise TypeError(f"The provided model should be subclass of {ServableModule.__qualname__}.")
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.resp = requests.post(f"http://{self.host}:{self.port}/serve", json=payload)
+        process.kill()
+
+    @property
+    def successful(self) -> bool:
+        """Returns whether the model was successful served."""
+        return self.resp.status_code == 200 if self.resp else False
+
+    def state_dict(self) -> Dict[str, Any]:
```
carmocca, line 109: Is this useful now?

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,145 @@
...
+        if not _RequirementAvailable("uvicorn"):
+            raise ModuleNotFoundError("The package uvicorn is required by the ServableModuleValidator callback.")
+
+        # TODO: Add support for those optimizations
+        assert optimization in (None, "trace", "script", "onnx", "tensor_rt")
+        if optimization is not None:
+            raise NotImplementedError(f"The optimization {optimization} is currently not supported.")
+
+        # TODO: Add support for testing with those server services
+        assert server in ("fastapi", "ml_server", "torch_serve", "sagemaker")
```
carmocca, line 54: You could remove this assertion and type server as

```python
from typing_extensions import Literal

server: Literal["fastapi"]
```

then IDEs and type checkers will do the job of complaining and you can remove this assert

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,145 @@
...
+        optimization: Optional[str] = None,
+        server: str = "fastapi",
+        host: str = "127.0.0.1",
+        port: int = 8080,
+    ):
+        super().__init__()
+        if not _RequirementAvailable("fastapi"):
+            raise ModuleNotFoundError("The package fastapi is required by the ServableModuleValidator callback.")
+        if not _RequirementAvailable("uvicorn"):
+            raise ModuleNotFoundError("The package uvicorn is required by the ServableModuleValidator callback.")
```
carmocca, line 46:
```suggestion
        fastapi_installed = _RequirementAvailable("fastapi")
        if not fastapi_installed:
            raise ModuleNotFoundError(fastapi_installed.message)
        uvicorn_installed = _RequirementAvailable("uvicorn")
        if not uvicorn_installed:
            raise ModuleNotFoundError(uvicorn_installed.message)
```

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: I would say yes. It would like to use this information to know if a checkpoint has been successfully validated for serving.

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: This is meant primarily for the readers and to show the ambition to support more.

justusschock:
With
> Lightning can provide common data types.

You mean types like the `Tensor` class in your example, right? Because we kind of agreed that we don't want domain-specific types within PL.

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,15 @@
+from modulefinder import Module
```
justusschock, line 1: What does this class do and why can't we simply use the built-in protocols for that?

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+from pytorch_lightning.utilities.rank_zero import rank_zero_only
+
+_NOT_SUPPORTED_STRATEGIES = (
+    DeepSpeedStrategy,
+    DDPFullyShardedNativeStrategy,
+    DDPFullyShardedStrategy,
+)
+
+
+class ServableModuleValidator(Callback):
```
justusschock, line 24: Why should this be a Callback? There is no connection to training. The sanity check could also be a simple function executing all these steps.

Is there a real benefit from doing that inside the trainer? 

it's not more boilerplate to call the checking function compared to instantiate the callback and add it to the trainer.

lantiga (commented):
Awesome job, left a few comments.

tests/tests_pytorch/serve/test_servable_module_validator.py
```diff
@@ -0,0 +1,41 @@
+from typing import Dict
+
+import torch
+
+from pytorch_lightning import seed_everything, Trainer
+from pytorch_lightning.demos.boring_classes import BoringModel
+from pytorch_lightning.serve.servable_module_validator import ServableModule, ServableModuleValidator
+
+
+class ServableBoringModel(BoringModel, ServableModule):
```
justusschock, line 10: I like this, but with this separate mixin, I strongly think that the validation/sanity checking shouldn't happen as part of a callback (and thereby inside the trainer). I think training and serving/deployment are completely different concerns and should also be treated that way.

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+    @property
+    def successful(self) -> bool:
+        """Returns whether the model was successfully served."""
+        return self.resp.status_code == 200 if self.resp else False
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {"successful": self.successful, "optimization": self.optimization, "server": self.server}
+
+    @staticmethod
+    def _start_server(servable_model: ServableModule, host: str, port: int, _: bool) -> None:
```
justusschock, line 110: did you test this with ddp-spawn?

I'm asking because you do have some locally defined functions here which may be an issue when pickling the callback

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+                resp = requests.get(f"http://{self.host}:{self.port}/ping")
+                ready = resp.status_code == 200
+            except requests.exceptions.ConnectionError:
+                pass
+            time.sleep(0.1)
+
+        payload = servable_module.configure_payload()
+
+        if "body" not in payload:
+            raise Exception(f"Your provided payload {payload} should have a field body.")
```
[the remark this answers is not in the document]
justusschock: @lantiga probably need to have `f"Your provided payload {payload} should have a field named 'body'."` to satisfy black :)

src/pytorch_lightning/serve/servable_module.py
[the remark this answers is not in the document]
tchaton: ... VSCode auto-imports ...

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+    Arguments:
+        optimization: The format in which the model should be tested while being served.
+        server: The library used to evaluate the model serving. Default is FastAPI.
+        host: The host address associated the server.
+        port: The port associated the server.
+    """
+
+    def __init__(
+        self,
+        optimization: Optional[Literal["trace", "script", "onnx", "tensorRt"]] = None,
```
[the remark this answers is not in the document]
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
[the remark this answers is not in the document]
tchaton: Thanks for the catch.

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+        process.start()
+
+        ready = False
+        while not ready:
+            try:
+                resp = requests.get(f"http://{self.host}:{self.port}/ping")
+                ready = resp.status_code == 200
+            except requests.exceptions.ConnectionError:
+                pass
+            time.sleep(0.1)
```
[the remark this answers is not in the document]
tchaton: Added a timeout in order to exit out.

src/pytorch_lightning/serve/servable_module_validator.py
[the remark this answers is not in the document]
tchaton: Done !

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+    def successful(self) -> bool:
+        """Returns whether the model was successfully served."""
+        return self.resp.status_code == 200 if self.resp else False
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {"successful": self.successful, "optimization": self.optimization, "server": self.server}
+
+    @staticmethod
+    def _start_server(servable_model: ServableModule, host: str, port: int, _: bool) -> None:
+        """This optimization starts a simple FastAPI server with a predict and ping endpoint."""
```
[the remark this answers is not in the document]
tchaton: Done !

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+        from fastapi import Body, FastAPI
+        from uvicorn import run
+
+        app = FastAPI()
+
+        deserializers, serializers = servable_model.configure_serialization()
+        servable_model.eval()
```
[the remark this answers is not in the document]
tchaton: Done !

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+        self.resp = requests.post(f"http://{self.host}:{self.port}/serve", json=payload)
+        process.kill()
+
+    @property
+    def successful(self) -> bool:
+        """Returns whether the model was successfully served."""
+        return self.resp.status_code == 200 if self.resp else False
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {"successful": self.successful, "optimization": self.optimization, "server": self.server}
```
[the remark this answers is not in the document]
tchaton: Good catch !

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: I believe this is a common sentiment shared by researchers, but I don't share it. Validating Serving for me is part of the training and should be done on_train_start and possibly on on_train_end.

It might evolve to do more, so it fits better in a callbacks, especially with shared models.

tests/tests_pytorch/serve/test_servable_module_validator.py
tchaton: Again, I strongly disagree and I think this common sentiment among researchers is the reason why model mobility is so hard to do.

In my view, the model is the vessel that should implement all contracts from data processing to serving.

In PyTorch Lightning, the Trainer is the entry point. If I care about serving, then the training should validate my model is even worth training.

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: The function shouldn't be a problem as the callback can be pickled without issues. The local functions aren't deepcopied.

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+
+_NOT_SUPPORTED_STRATEGIES = (
+    DeepSpeedStrategy,
+    DDPFullyShardedNativeStrategy,
+    DDPFullyShardedStrategy,
+)
+
+
+class ServableModuleValidator(Callback):
+    """The ServableModuleValidator callback enables to validate a model is servable following the ServableModule
```
carmocca, line 25: Can you add the warning about this being experimental? Also, should we include it in the docs reference already?

carmocca (approved):
LGTM!

krshrimali (approved):
LGTM, thanks @tchaton - minor suggestion, should we add a `serve.txt` requirements file [here](https://github.com/Lightning-AI/lightning/tree/master/requirements/pytorch)?

src/pytorch_lightning/serve/servable_module_validator.py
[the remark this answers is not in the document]
hhsecond: forgot to push?

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Any, Callable, Dict, Tuple
+
+import torch
+
+
+class ServableModule(Module):
```
hhsecond, line 7: Could you add some docs here about what each of these functions stand for? Doesn't have to be elaborate but perhaps with some examples as you have in the PR description?

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+            try:
+                resp = requests.get(f"http://{self.host}:{self.port}/ping")
+                ready = resp.status_code == 200
+            except requests.exceptions.ConnectionError:
+                pass
+            time.sleep(0.1)
+
+        payload = servable_module.configure_payload()
+
+        if "body" not in payload:
```
hhsecond, line 95: Not a blocker really but we could pick a better name than `body` :)

hhsecond (approved):
Nice work!

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,142 @@
...
+        server: The library used to evaluate the model serving. Default is FastAPI.
+        host: The host address associated the server.
+        port: The port associated the server.
+    """
+
+    def __init__(
+        self,
+        optimization: Optional[Literal["trace", "script", "onnx", "tensorRt"]] = None,
+        server: Literal["fastapi", "ml_server", "torchserve", "sagemaker"] = "fastapi",
+        host: str = "127.0.0.1",
```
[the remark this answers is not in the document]
tchaton: It is usually cleaner to set 127.0.0.1. This is what I found in FastAPI and uvicorn.

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: Yes, I am planning to add it to the docs with some examples.

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: Resolved this bug.

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: Let's keep this for now.

examples/pl_servable_module/production.py
```diff
@@ -0,0 +1,137 @@
...
+class Image:
+
+    height: Optional[int] = None
+    width: Optional[int] = None
+    extension: str = "JPEG"
+    mode: str = "RGB"
+    channel_first: bool = False
+
+    def deserialize(self, data: str) -> torch.Tensor:
+        encoded_with_padding = (data + "===").encode("ascii")
```
justusschock, line 76: what is this padding used for?

examples/pl_servable_module/production.py
```diff
@@ -0,0 +1,137 @@
...
+        # 1: Access the train dataloader and load a single sample.
+        image, _ = self.trainer.train_dataloader.loaders.dataset[0]
+
+        # 2: Convert the image into a PIL Image and save it to disk
+        pil_image = T.ToPILImage()(image)
+        pil_image.save("payload_image.png")
+
+        # 3: Load the image as bytes and encode it with base64
+        with open("payload_image.png", "rb") as f:
+            imgstr = base64.b64encode(f.read()).decode("UTF-8")
```
justusschock: could we maybe always stick to ascii or always to utf8 (preferred)?

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -107,6 +112,13 @@ def on_train_start(self, trainer: "pl.Trainer", servable_module: "pl.LightningMo
         self.resp = requests.post(f"http://{self.host}:{self.port}/serve", json=payload)
         process.kill()
 
+        if is_overridden("configure_response", servable_module, ServableModule):
+            response = servable_module.configure_response()
+            assert self.resp.json() == response
+
+        if self.exit_on_failure and not self.successful:
+            raise SystemExit("The ServableModel API hasn't been properly implemented.")
```
justusschock, line 121: that is one of the few cases where IMO a MisconfigurationException would fit :)

tests/tests_pytorch/serve/test_servable_module_validator.py
```diff
@@ -41,6 +43,7 @@ def test_servable_module_validator_with_trainer():
         limit_train_batches=2,
         limit_val_batches=0,
         callbacks=[callback],
+        strategy="ddp_spawn",
```
justusschock, line 46: should we test thus with all the supported strategies just to be sure?

docs/source-pytorch/deploy/production_advanced.rst
```diff
@@ -58,3 +58,19 @@ Once you have the exported model, you can run it on your ONNX runtime in the fol
...
     ort_inputs = {input_name: np.random.randn(1, 64)}
     ort_outs = ort_session.run(None, ort_inputs)
+
+----
+
+****************************
+Validate a Model Is Servable
+****************************
+
+Production ML Engineers would argue that a model shouldn't be trained if it can't be deployed reliability and in a fully automated manner.
```
carmocca, line 68:
```suggestion
Production ML Engineers would argue that a model shouldn't be trained if it can't be deployed reliably and in a fully automated manner.
```

docs/source-pytorch/deploy/production_advanced.rst
```diff
@@ -58,3 +58,19 @@ Once you have the exported model, you can run it on your ONNX runtime in the fol
...
+
+In order to ease transition from training to production, PyTorch Lightning provides a way for you to validate a model can be served even before starting training.
+
+In order to do so, your LightningModule needs to subclass the :class:`~pytorch_lightning.serve.servable_module.ServableModule`, implements its hooks and pass a :class:`~pytorch_lightning.serve.servable_module_validator.ServableModuleValidator` callback to the Trainer.
+
+Find below an example for validating a resnet18 can be served.
```
carmocca, line 74:
```suggestion
Below you can find an example of how the serving of a resent18 can be validated.
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+                def deserialize(x):
+                    return torch.tensor(x, dtype=torch.float)
+
+                def serialize(x):
+                    return x.tolist()
+
+                return {"x": deserialize}, {"output": serialize}
+
+            def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
+                assert torch.equal(x, torch.arange(32, dtype=torch.float))
```
carmocca, line 41: The assertion shouldn't be part of the example

```suggestion
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+
+        import torch
+
+        from pytorch_lightning import seed_everything, Trainer
+        from pytorch_lightning.demos.boring_classes import BoringModel
+        from pytorch_lightning.serve.servable_module_validator import ServableModule, ServableModuleValidator
+
+
+        class ServableBoringModel(BoringModel, ServableModule):
+            def configure_payload(self) -> ...:
```
carmocca, line 28:
```suggestion
            def configure_payload(self):
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+        r"""
+        Returns the predictions of your model as a dictionary.
+
+        .. code-block:: python
+
+            def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
+                return {"predictions": self(x)}
+
+        Args:
+            args|kwargs (:class:`~torch.Tensor` | (:class:`~torch.Tensor`, ...) | [:class:`~torch.Tensor`, ...]):
```
carmocca, line 81: The types are included automatically

```suggestion
            args|kwargs:
```

Also does this `|` notation render as expected? I haven't seen it before

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+                The output of the serializer functions provided by the ``configure_serialization`` hook.
+
+        Return:
+            - ``dict`` - A dictionary. Can include any keys, but must include the key ``'loss'``
```
carmocca, line 85: Why does it need to include the key loss?

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+            limit_val_batches=0,
+            callbacks=[callback],
+        )
+        trainer.fit(ServableBoringModel())
+        assert callback.resp.json() == {"output": [0, 1]}
+    """
+
+    def configure_payload(self) -> Dict[str, Any]:
+        """Returns a request payload as a dictionary."""
+        ...
```
carmocca, line 58:
```suggestion
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+    def configure_serialization(self) -> Tuple[Dict[str, Callable], Dict[str, Callable]]:
+        """Returns a tuple of dictionaries.
+
+        The first dictionary contains the name of the ``serve_step`` input variables name as its keys
+        and the associated de-serialization function (e.g function to convert a payload to tensors).
+
+        The second dictionary contains the name of the ``serve_step`` output variables name as its keys
+        and the associated serialization function (e.g function to convert a tensors into payload).
+        """
+        ...
```
carmocca, line 69:
```suggestion
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+                return {"predictions": self(x)}
+
+        Args:
+            args|kwargs (:class:`~torch.Tensor` | (:class:`~torch.Tensor`, ...) | [:class:`~torch.Tensor`, ...]):
+                The output of the serializer functions provided by the ``configure_serialization`` hook.
+
+        Return:
+            - ``dict`` - A dictionary. Can include any keys, but must include the key ``'loss'``
+        """
+        ...
```
carmocca, line 87:
```suggestion
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def configure_response(self) -> Dict[str, Any]:
+        """Returns a response to validate the server response."""
+        ...
```
carmocca, line 91:
```suggestion
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,169 @@
...
+            raise TypeError(f"The provided model should be subclass of {ServableModule.__qualname__}.")
+
+        if not is_overridden("configure_payload", servable_module, ServableModule):
+            raise NotImplementedError("The `configure_payload` method needs to be overridden.")
+        if not is_overridden("configure_serialization", servable_module, ServableModule):
+            raise NotImplementedError("The `configure_serialization` method needs to be overridden.")
+        if not is_overridden("serve_step", servable_module, ServableModule):
+            raise NotImplementedError("The `serve_step` method needs to be overridden.")
+
+        servable_module.trainer = None
```
carmocca, line 88: Why set it to None and to Trainer right after?

src/pytorch_lightning/serve/servable_module_validator.py
carmocca: "hasn't been properly implemented" is unclear for users. Maybe direct them to a specific resource?

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,169 @@
...
+        if is_overridden("configure_response", servable_module, ServableModule):
+            response = servable_module.configure_response()
+            if self.resp.json() != response:
+                raise Exception(f"The expected response {response} doesn't match the generated one {self.resp.json()}.")
+
+        if self.exit_on_failure and not self.successful:
+            raise SystemExit("The ServableModel API hasn't been properly implemented.")
+
+        if self.successful:
+            print(f"Your model is servable and the received payload was {self.resp.json()}.")
```
carmocca, line 124: This should be a debug statement if anything. Also, use a logger instead of basic print

src/pytorch_lightning/serve/servable_module_validator.py
justusschock: The issue is that the trainer (or at least some of it's components - likely loggers, maybe also some other stuff) cannot be pickled. This is required for passing it to a new process.

carmocca: A comment would be helpful then

examples/pl_servable_module/production.py
tchaton: This was used in Flash. I kept it as it was.

examples/pl_servable_module/production.py
tchaton: Changed to `UTF-8`.

tests/tests_pytorch/serve/test_servable_module_validator.py
tchaton: I think it is an overkill right now.

docs/source-pytorch/deploy/production_advanced.rst
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Typo.

src/pytorch_lightning/serve/servable_module.py
tchaton: This is needed for the IDE to not complain somehow.

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,169 @@
...
+        ready = False
+        t0 = time.time()
+        while not ready:
+            try:
+                resp = requests.get(f"http://{self.host}:{self.port}/ping")
+                ready = resp.status_code == 200
+            except requests.exceptions.ConnectionError:
+                pass
+            if time.time() - t0 > self.timeout:
+                raise Exception(f"The Server didn't start in {self.timeout}")
```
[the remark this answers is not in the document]
tchaton: This is meant to exit the infinite loop.

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: Should be info. If you are using the callback, you want to know it was successful.

examples/pl_servable_module/production.py
```diff
@@ -0,0 +1,135 @@
...
+from pytorch_lightning import cli_lightning_logo, LightningDataModule, LightningModule
+from pytorch_lightning.serve import ServableModule, ServableModuleValidator
+from pytorch_lightning.utilities.cli import LightningCLI
+
+DEFAULT_CMD_LINE = (
+    "fit",
+    "--trainer.max_epochs=1",
+    "--trainer.limit_train_batches=15",
+    "--trainer.limit_val_batches=15",
+)
```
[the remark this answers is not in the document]
tchaton: I personally prefer how it is right now.

examples/pl_servable_module/production.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def cli_main():
+
+    seed_everything(42)
+
+    cli = LightningCLI(
+        ProductionReadyModel,
+        CIFAR10DataModule,
+        save_config_overwrite=True,
```
carmocca, line 115:
```suggestion
    cli = LightningCLI(
        ProductionReadyModel,
        CIFAR10DataModule,
        seed_everything_default=42,
        save_config_overwrite=True,
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,175 @@
...
+        process = Process(target=self._start_server, args=(servable_module, self.host, self.port, self.optimization))
+        process.start()
+
+        servable_module.trainer = trainer
+
+        ready = False
+        t0 = time.time()
+        while not ready:
+            try:
+                resp = requests.get(f"http://{self.host}:{self.port}/ping")
```
adriangonz, line 105: Would it make sense to test that the model can be served with specific inference servers, like Triton or MLServer? This would ensure that the environment is as close to prod as similar. Conscious that this would mean having separate validators for each serving engine though (i.e. one for Sagemaker, one for the inference servers listed above, one for Seldon Core, etc.)....

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,175 @@
...
+        self,
+        optimization: Optional[Literal["trace", "script", "onnx", "tensorrt"]] = None,
+        server: Literal["fastapi", "ml_server", "torchserve", "sagemaker"] = "fastapi",
+        host: str = "127.0.0.1",
+        port: int = 8080,
+        timeout: int = 10,
+        exit_on_failure: bool = True,
+    ):
+        super().__init__()
+        fastapi_installed = _RequirementAvailable("fastapi")
```
adriangonz, line 52: Probably outside the scope of this PR, but would it make sense to save the full set of dependencies for a model after training so that we can check whether these are present on the serving environment? I think that this is something that the user could declare and then Lightning could save after training.

adriangonz (commented):
This looks great @tchaton !

Following up from our brief discussion in Slack, just added a couple comments below. It would be great to hear your thoughts!

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: This is 100 % a great idea. I think more metadata should be stored, like requirements, payload, etc...

justusschock: Probably integrates quite nice with https://pytorch.org/docs/stable/package.html

williamFalcon:
the premise here is that every model needs to be served which is not true. 

for example a model to fold proteins does not need “serving”… this is a model that runs once in a while to generate a new protein sequence for a lab to synthesize. 

so, no, we can’t “force” every model to be “production ready.”. 

however, if a team ops for forcing their particular research code to always be production ready, they should have a mechanism to enforce that behavior and can opt in.

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,90 @@
...
+        Returns the predictions of your model as a dictionary.
+
+        .. code-block:: python
+
+            def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
+                return {"predictions": self(x)}
+
+        Args:
+            args: The output from serializer functions provided by the ``configure_serialization`` hook.
+            kwargs: The keyword output of the serializer functions provided by the ``configure_serialization`` hook.
```
rohitgr7, line 81:
```suggestion
            args: The output from de-serializer functions provided by the ``configure_serialization`` hook.
            kwargs: The keyword output of the de-serializer functions provided by the ``configure_serialization`` hook.
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,175 @@
...
+        self.optimization = optimization
+        self.host = host
+        self.port = port
+        self.server = server
+        self.timeout = timeout
+        self.exit_on_failure = exit_on_failure
+        self.resp: Optional[requests.Response] = None
+
+    @rank_zero_only
+    def on_train_start(self, trainer: "pl.Trainer", servable_module: "pl.LightningModule") -> None:
```
rohitgr7, line 75: should it be enabled with `fast_dev_run`?

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,175 @@
...
+        if isinstance(trainer.strategy, _NOT_SUPPORTED_STRATEGIES):
+            raise Exception(
+                f"The current strategy {trainer.strategy} used "
```
rohitgr7, line 78:
```suggestion
                f"The current strategy {trainer.strategy.__class__.__qualname__} used "
```

src/pytorch_lightning/serve/servable_module_validator.py
```diff
@@ -0,0 +1,175 @@
...
+
+    .. warning::
+
+        This is currently an experimental feature and API changes are to be expected.
+
+    Arguments:
+        optimization: The format in which the model should be tested while being served.
+        server: The library used to evaluate the model serving. Default is FastAPI.
+        host: The host address associated the server.
+        port: The port associated the server.
```
rohitgr7, line 39:
```suggestion
        optimization: The format in which the model should be tested while being served.
        server: The library used to evaluate the model serving. The default is FastAPI.
        host: The host address associated with the server.
        port: The port associated with the server.
        timeout: Timeout period in seconds, that the process should wait for the server to start.
        exit_on_failure: Whether to exit the process on failure.
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,90 @@
...
+
+        The first dictionary contains the name of the ``serve_step`` input variables name as its keys
+        and the associated de-serialization function (e.g function to convert a payload to tensors).
+
+        The second dictionary contains the name of the ``serve_step`` output variables name as its keys
+        and the associated serialization function (e.g function to convert a tensors into payload).
+        """
+        ...
+
+    def serve_step(self, *args: torch.Tensor, **kwargs: torch.Tensor) -> Dict[str, torch.Tensor]:
```
rohitgr7, line 70: will there be args in any case?
```py
output = servable_model.serve_step(**body)
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,90 @@
...
+
+class ServableModule(torch.nn.Module):
+
+    """The ServableModule provides a simple API to make your model servable.
+
+    .. warning::
+
+        This is currently an experimental feature and API changes are to be expected.
+
+    Here is an example on how to use the `ServableModule` module.
```
rohitgr7, line 14:
```suggestion
    Here is an example of how to use the ``ServableModule`` module.
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,90 @@
...
+
+    .. code-block:: python
+
+        from typing import Dict, Any, Callable
+
+        import torch
+
+        from pytorch_lightning import seed_everything, Trainer
```
rohitgr7, line 22:
```suggestion
        from pytorch_lightning import Trainer
```

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,90 @@
...
+
+        callback = ServableModuleValidator()
+        trainer = Trainer(
+            max_epochs=1,
+            limit_train_batches=2,
+            limit_val_batches=0,
+            callbacks=[callback],
+        )
+        trainer.fit(ServableBoringModel())
+        assert callback.resp.json() == {"output": [0, 1]}
```
rohitgr7, line 52:
```suggestion
        serve_cb = ServableModuleValidator()
        trainer = Trainer(
            max_epochs=1,
            limit_train_batches=2,
            limit_val_batches=0,
            callbacks=[serve_cb],
        )
        trainer.fit(ServableBoringModel())
        assert serve_cb.resp.json() == {"output": [0, 1]}
```

tests/tests_pytorch/serve/test_servable_module_validator.py
```diff
@@ -0,0 +1,49 @@
...
+    model = ServableBoringModel()
+    callback = ServableModuleValidator()
+    callback.on_train_start(Trainer(), model)
+
+
+def test_servable_module_validator_with_trainer():
+    seed_everything(42)
+    callback = ServableModuleValidator()
+    trainer = Trainer(
+        max_epochs=1,
```
rohitgr7, line 42:
```suggestion
def test_servable_module_validator_with_trainer(tmpdir):
    seed_everything(42)
    callback = ServableModuleValidator()
    trainer = Trainer(
        default_root_dir=tmpdir,
        max_epochs=1,
```

rohitgr7 (approved):

src/pytorch_lightning/serve/servable_module.py
```diff
@@ -0,0 +1,90 @@
...
+            def configure_serialization(self) -> Tuple[Dict[str, Callable], Dict[str, Callable]]:
+                def deserialize(x):
+                    return torch.tensor(x, dtype=torch.float)
+
+                def serialize(x):
+                    return x.tolist()
+
+                return {"x": deserialize}, {"output": serialize}
+
+            def serve_step(self, x: torch.Tensor) -> Dict[str, torch.Tensor]:
```
rohitgr7, line 40: I think the example should also use `configure_response` to show that a user doesn't have to do the output validation.

justusschock:
@williamFalcon that is why this was designed to be a combination of an optional mixin + a callback instead of directly added to the lightning module and the trainer. So this is opt-in as it is currently implemented :)

justusschock (approved):

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: I think so too. This is quite fast.

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module_validator.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

src/pytorch_lightning/serve/servable_module.py
tchaton: Done !

awaelchli (approved):
unblock

PR #13628, opened 2022-07-12
CI: adjust GPU timeout

Borda:
## What does this PR do?

seems that for some cases the GPU testing hangs on the standard test 66% for 1h40min
so let's kill it sooner and have better control of how long each step shall take...

## Before submitting

## PR review

## Did you have fun?


carmocca:
A per-test timeout would be much more reliable in this case. The cpu job uses `pytest-timeout` successfully

Borda:
> A per-test timeout would be much more reliable in this case. The cpu job uses `pytest-timeout` successfully

good point, we can check but in past, we had some problems running the timeout on GPU (but worse to check it again :)...
also, an advantage of this one is that this is driven by the DevOps so in case the pytest hangs it will be terminated by Pipelines...

carmocca (approved):
I'm not saying the azure timeout shouldn't be tuned (it should!) But stopping once a test hangs for, say, 3 minutes will improve the situation more than reducing the total timeout from 100 -> 80 min

justusschock (approved):

awaelchli (approved):

akihironitta (approved):

PR #13634, opened 2022-07-13
[App] Application logs in CLI

adam-lightning:
## What does this PR do?

This PR adds CLI command `lightning show logs <app_name> [<components>]` that prints logs

We've extended the `run_app_in_cloud` context manager to return `app_name` as well. This is to enable easier testing, but might also affect other existing tests from different repositories.

## Before submitting

## PR review

## Did you have fun?

cc @borda


williamFalcon:
is “show” the best command word here? @adam-lightning. 
how will this generalize as the CLI evolves?

adam-lightning:
@williamFalcon  The idea is that `show` would be for showing / displaying / printing all information about given resource(s).
Something like just saying "Show me logs of X", "show me something ...", "show me my application" etc.

Tho, for the future, I'd distinguish the "show" from "list" as one first one answers the question about the object's contents / details, and the other lists the objects we might want to interact with.

Or You mean something else by the CLI evolution ?

williamFalcon:
that makes sense. i'm just thinking about what else "show" will apply to other than logs

src/lightning_app/cli/lightning_cli.py
```diff
@@ -29,10 +31,68 @@ def get_app_url(runtime_type: RuntimeType, *args) -> str:
...
 
 
+@main.group()
+def show():
+    """Show info about resource."""
+    pass
+
+
+@show.command()
+@click.argument("app_name", required=False)
```
manskx, line 44: this should be required? 

should we also allow using `app_id`?

src/lightning_app/cli/lightning_cli.py
```diff
@@ -29,10 +31,68 @@ def get_app_url(runtime_type: RuntimeType, *args) -> str:
...
+@click.argument("components", nargs=-1, required=False)
+@click.option("-f", "--follow", required=False, is_flag=True)
+def logs(app_name: str, components: List[str], follow: bool) -> None:
```
manskx, line 47: Docstrings and help strings will be nice. Specially for components to indicate the `flow` can be used

src/lightning_app/cli/lightning_cli.py
```diff
@@ -29,10 +31,68 @@ def get_app_url(runtime_type: RuntimeType, *args) -> str:
...
+    if not apps:
+        raise click.ClickException("Your app list is empty. Please run an application first.")
+
+    if not app_name:
+        raise click.ClickException(
+            f"You have not specified any LightningApp. Please select one of available: [{', '.join(apps.keys())}]"
+        )
+
+    if app_name not in apps:
+        raise click.ClickException(f"LightningApp '{app_name}' does not exist.")
```
manskx, line 66: nit: show list of apps like above `Please select one of available: [{', '.join(apps.keys())}]`

src/lightning_app/utilities/app_logs.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _error_callback(ws_app: WebSocketApp, error: Exception):
+    errors = {
+        KeyError: "Malformed log message, missing key",
+        JSONDecodeError: "Malformed log message",
+        TypeError: "Malformed log format",
+        ValueError: "Malformed date format",
+    }
+    print(f"Error while reading logs ({errors.get(type(error), 'Unknown')})", file=sys.stderr)
```
manskx, line 64: nit: show the exception msg also

src/lightning_app/utilities/app_logs.py
```diff
@@ -0,0 +1,118 @@
...
+        # We strongly trust that the contract on API will hold atm :D
+        event_dict = json.loads(msg)
+        labels = _LogEventLabels(**event_dict["labels"])
+        if "message" in event_dict:
+            event = _LogEvent(
+                message=event_dict["message"],
+                timestamp=dateutil.parser.isoparse(event_dict["timestamp"]),
+                labels=labels,
+            )
+            read_queue.put((event.timestamp, component_name, event))
```
manskx, line 52: ❤️

src/lightning_app/utilities/app_logs.py
```diff
@@ -0,0 +1,118 @@
...
+
+def _push_logevents_to_read_queue_callback(component_name: str, read_queue: queue.PriorityQueue):
+    """Pushes _LogEvents from websocket to read_queue.
+
+    Returns callback function used with `on_message_callback` of websocket.WebSocketApp.
+    """
+
+    def callback(ws_app: WebSocketApp, msg: str):
+        # We strongly trust that the contract on API will hold atm :D
+        event_dict = json.loads(msg)
```
manskx, line 44: nit: could you add an example of how the log message json looks like ?

manskx:
@adam-lightning We should add two things:
- An issue/ticket to create update the docs
- Update changelog

adam-lightning:
> @adam-lightning We should add two things:
> 
> * An issue/ticket to create update the docs
> * Update changelog

- Just create a ticket or also implement it before merging this PR ?
- I haven't found changelog in the existing files. Should I create one?

manskx:
> > @adam-lightning We should add two things:
> > 
> > * An issue/ticket to create update the docs
> > * Update changelog
> 
> @manskx
> 
> * Just create a ticket or also implement it before merging this PR ?
> * I haven't found changelog in the existing files. Should I create one?

Would be nice to update the docs in the same PR  :))) up to you

changelog is here: https://github.com/Lightning-AI/lightning/blob/35ec79ad1f75790767a6857637c9bbe052e47b58/src/lightning_app/CHANGELOG.md

src/lightning_app/cli/lightning_cli.py
adam-lightning: We can think about using `app_id` that's a good idea actually... 
It's `required=False` because later we do a check anyway, and for better UX we list all available applications.  Pushing check into the `click` validator would mess up the code ...

If we do not provide the `app_name` we get something like 
```
Error: You have not specified any LightningApp. Please select one of available: [brawny-cerf-2556, lightning/hackernews-app-0.0.2]
```

src/lightning_app/utilities/logs_socket_api.py
```diff
@@ -0,0 +1,91 @@
...
+        return token_resp.token
+
+    @staticmethod
+    def _socket_url(host: str, project_id: str, app_id: str, token: str, component: str) -> str:
+        return (
+            f"wss://{host}/v1/projects/{project_id}/appinstances/{app_id}/logs?"
+            f"token={token}&component={component}&follow=true"
+        )
+
+    def create_lightning_logs_socket(
```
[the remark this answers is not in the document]
adam-lightning: We agreed (at some point) that from the `lightning` side we should use only `lightning-cloud` as API spec, and everything above and custom, should be directly in lightning.

src/lightning_app/utilities/app_logs.py
adam-lightning: Surfacing exception message might just confuse the user imho.
My idea was that the exception that I encountered are mapped, and we surface a nice error to the user.
If we added the exception message we would "leak" the underlying logic which is never recommended.

src/lightning_app/utilities/app_logs.py
adam-lightning: I think that it's implicitly stated in the `_LogEvent` class. Including example directly might be also considered implementation leakage, so not sure if we want to include that explicitly here ;) 
```
{'message': '', 'timestamp': '', 'labels': {'app': '', 'container': '', 'filename': '', 'job': '', 'namespace': '', 'node_name': '', 'pod': '', 'stream': ''}}
```

src/lightning_app/cli/lightning_cli.py
adam-lightning: 👍

adam-lightning:
> > > @adam-lightning We should add two things:
> > > 
> > > * An issue/ticket to create update the docs
> > > * Update changelog
> > 
> > 
> > @manskx
> > 
> > * Just create a ticket or also implement it before merging this PR ?
> > * I haven't found changelog in the existing files. Should I create one?
> 
> Would be nice to update the docs in the same PR :))) up to you
> 
> changelog is here: https://github.com/Lightning-AI/lightning/blob/35ec79ad1f75790767a6857637c9bbe052e47b58/src/lightning_app/CHANGELOG.md

Updated the changelog. But I don't really know how / what to update in docs tbh. Do Yo have any idea @manskx  as to what to write and where? I think that adding CLI command and documenting it in CLI is good enough, right ?

manskx (approved):
Great work!

manskx:
> > > > @adam-lightning We should add two things:
> > > > 
> > > > * An issue/ticket to create update the docs
> > > > * Update changelog
> > > 
> > > 
> > > @manskx
> > > 
> > > * Just create a ticket or also implement it before merging this PR ?
> > > * I haven't found changelog in the existing files. Should I create one?
> > 
> > 
> > Would be nice to update the docs in the same PR :))) up to you
> > changelog is here: https://github.com/Lightning-AI/lightning/blob/35ec79ad1f75790767a6857637c9bbe052e47b58/src/lightning_app/CHANGELOG.md
> 
> Updated the changelog. But I don't really know how / what to update in docs tbh. Do Yo have any idea @manskx as to what to write and where? I think that adding CLI command and documenting it in CLI is good enough, right ?


I think we can create a GH issue to update the docs of this functionality.
cc: @tchaton @Felonious-Spellfire

Let's also keep an eye on the e2e CI after this gets merged.

src/lightning_app/cli/lightning_cli.py
```diff
@@ -49,10 +51,89 @@ def main():
 @click.group()
 @click.version_option(ver)
 def _main():
-    register_all_external_components()
+    # register_all_external_components()
```
tchaton, line 54: Why this change? I should belong to another PR.

src/lightning_app/cli/lightning_cli.py
```diff
@@ -49,10 +51,89 @@ def main():
...
     pass
 
 
+@_main.group()
+def show():
+    """Show given resource."""
+    pass
+
+
+@show.command()
```
tchaton, line 64: This doesn't work locally. So we need a mechanism to inform the users.

src/lightning_app/cli/lightning_cli.py
```diff
@@ -49,10 +51,89 @@ def main():
...
+@click.argument("app_name", required=False)
+@click.argument("components", nargs=-1, required=False)
+@click.option("-f", "--follow", required=False, is_flag=True, help="Wait for new logs, to exit use CTRL+C.")
+def logs(app_name: str, components: List[str], follow: bool) -> None:
+    """Show application logs. By default prints logs for all currently available components.
```
tchaton, line 69:
```suggestion
    """Show cloud application logs. By default prints logs for all currently available components.
```

src/lightning_app/cli/lightning_cli.py
```diff
@@ -49,10 +51,89 @@ def main():
...
+        app.name: app
+        for app in client.lightningapp_instance_service_list_lightningapp_instances(project.project_id).lightningapps
+    }
+
+    if not apps:
+        raise click.ClickException("Your app list is empty. Please run an application first.")
+
+    if not app_name:
+        raise click.ClickException(
+            f"You have not specified any LightningApp. Please select one of available: [{', '.join(apps.keys())}]"
```
tchaton, line 101:
```suggestion
            f"You have not specified any Lightning App. Please select one of available: [{', '.join(apps.keys())}]"
```

src/lightning_app/cli/lightning_cli.py
```diff
@@ -49,10 +51,89 @@ def main():
...
+    client = LightningClient()
+    project = _get_project(client)
+
+    apps = {
+        app.name: app
+        for app in client.lightningapp_instance_service_list_lightningapp_instances(project.project_id).lightningapps
+    }
+
+    if not apps:
+        raise click.ClickException("Your app list is empty. Please run an application first.")
```
tchaton, line 97:
```suggestion
        raise click.ClickException("You don't have any application in the cloud. Please, run an application first with `--cloud`.")
```

tchaton, line 97: Does this include stopped apps or only running ones?

src/lightning_app/cli/lightning_cli.py
```diff
@@ -49,10 +51,89 @@ def main():
...
+
+    if not app_name:
+        raise click.ClickException(
+            f"You have not specified any LightningApp. Please select one of available: [{', '.join(apps.keys())}]"
+        )
+
+    if app_name not in apps:
+        raise click.ClickException(
+            f"LightningApp '{app_name}' does not exist. Please select one of available: [{', '.join(apps.keys())}]"
```
tchaton, line 106:
```suggestion
            f"The Lightning App '{app_name}' does not exist. Please select one of following: [{', '.join(apps.keys())}]"
```

src/lightning_app/utilities/app_logs.py
```diff
@@ -0,0 +1,118 @@
...
+    #   run_forever() will run until we close() the connection from outside
+    log_threads = [Thread(target=work.run_forever) for work in log_sockets]
+
+    # Establish connection and begin pushing logs to the print queue
+    for th in log_threads:
+        th.start()
+
+    # Print logs from queue when log event is available
+    try:
+        while True:
```
tchaton, line 97: If I understand this properly, this doesn't strictly ensure the logs are in the correct order right ?

tchaton (approved):
LGTM once the comments are addressed.

src/lightning_app/utilities/app_logs.py
adam-lightning: Yes, not strictly, but works in practice good enough.

Because each component has logs from different websocket we need some way of funnelling them into one stream (stdout), idea here is to push from all WSs to one priority queue (with priority on timestamp) so it get's sorted "on-the-fly".

Not ideal - we can imagine that network latency might mess up output, but it's the best what we can do, while actually doing it "on-the-fly" rather than pulling all logs -> sorting -> printing - which would be time and memory consuming.
Also I think it might be slightly better than UI because of this priority queue, I sometimes observed that UI had messed up the order of logs.

src/lightning_app/cli/lightning_cli.py
adam-lightning: You are absolutely right. Mea culpa here 😓

src/lightning_app/cli/lightning_cli.py
adam-lightning: @tchaton  Actually, now I think about it. Maybe we could unify the behaviour of local and cloud app running? It would take some time but we have (almost) all peaces of the puzzle.
We could locally stream all logs to some output file (e.g. somewhere in `.lightning` directory) and read them with `show logs`.

So by default running locally would actually return control right after starting an app instead of hanging and printing output. 

```
lightning run app app.py [--cloud] 

lightning show logs --follow.  # print logs as they go -> we have no difference between behaviour locally and on cloud
```

What do you think? Anywho - something to think about :D

src/lightning_app/cli/lightning_cli.py
adam-lightning: non-deleted ones. So getting logs from stopped will be ok

src/lightning_app/utilities/app_logs.py
```diff
@@ -0,0 +1,118 @@
...
+
+def _app_logs_reader(
+    client: LightningClient, project_id: str, app_id: str, component_names: List[str], follow: bool
+) -> Iterator[Tuple[str, _LogEvent]]:
+
+    read_queue = queue.PriorityQueue()
+    logs_api_client = _LightningLogsSocketAPI(client.api_client)
+
+    # We will use a socket per component
+    log_sockets = [
```
awaelchli, line 76: This probably works well for some apps, and is a good start. I just want to point out that an app doesn't need to have a constant amount of "components". These could come and go during runtime of the app. So if you "follow" the logs, that's probably useful to know.

awaelchli (approved):
I like "show". For a different product, we had a CLI redesign planned (now on ice) where "show" was supposed to show information about many different types resources. It also fits well for logs. 

`lightning show [what] [name or id]`

src/lightning_app/cli/lightning_cli.py
tchaton: Yes, why not? Personally, I think it is a bit of overkill in the grand scheme of all the things to build, but I can imagine some users using this feature to isolate logs from a machine.

src/lightning_app/utilities/app_logs.py
tchaton: I think we could have a moving window buffer acting to sort the logs with a period of several seconds.

src/lightning_app/cli/lightning_cli.py
```diff
@@ -50,9 +53,89 @@ def main():
...
+    """Show cloud application logs. By default prints logs for all currently available components.
+
+    Example uses:
+
+        Print all application logs:
+
+            $ lightning show logs my-application
+
+
+        Print logs only from the root flow component:
```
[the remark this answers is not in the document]
tchaton: Done !

src/lightning_app/cli/lightning_cli.py
```diff
@@ -50,9 +53,89 @@ def main():
...
+
+            $ lightning show logs my-application flow
+
+
+        Print logs only from selected components:
```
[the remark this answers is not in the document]
tchaton: Done !

src/lightning_app/cli/lightning_cli.py
```diff
@@ -50,9 +53,89 @@ def main():
...
+
+            $ lightning show logs my-application root.comp_a root.comp_b
```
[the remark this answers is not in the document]
tchaton: Done !

src/lightning_app/utilities/app_logs.py
tchaton: Yes, this is definitely a skateboard version and needs to be improved upon.

PR #13638, opened 2022-07-13
Support setting the trainer reference recursively for ensembles

carmocca:
## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning/issues/13146

Changes:
- Recursively sets the Trainer reference for LightningModules
- Uses a weak reference for the Trainer
- Disambiguates the trainer attribute (optional) from the property getter (non-optional)
- ~~The same change was done to the Loop.trainer property for consistency~~ edit: it breaks the spawn queue
- Updates codebase accordingly

`model.trainer` will now raise a `RuntimeError` if it hasn't been set.

## Before submitting

## PR review

cc @borda @carmocca @justusschock @awaelchli @ananthsub @ninginthecloud @jjenniferdai @rohitgr7 @akihironitta


src/pytorch_lightning/core/module.py
```diff
@@ -172,6 +172,23 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
...
 
+    @property
+    def trainer(self) -> "pl.Trainer":
+        if self._trainer is None:
+            raise RuntimeError(f"{self.__class__.__qualname__} is not attached to a `Trainer`.")
+        return self._trainer
+
+    @trainer.setter
+    def trainer(self, trainer: "pl.Trainer") -> None:
+        if not isinstance(trainer, pl.Trainer):
```
justusschock, line 183:
```suggestion
        if not isinstance(trainer, pl.Trainer) and trainer is not None:
```
this would allow us to also recursively detach the trainer again in case we want to copy the model, pickle it (e.g. want to send it to another process)

justusschock (approved):
I like it!

src/pytorch_lightning/core/module.py
carmocca: @justusschock True. In that case, I'm leaning towards removing this check entirely and letting duck-typing happen.

What do you think? And should the Loop.trainer property have the same change?

justusschock: I still think it's good to have some check and not allow anything there. One more question: Do you think we should maybe make this a weakref? So that we don't actually have circular references?

For the Loop.trainer:
Is there ever a way where we do need to copy those or detach? From what I can think of, we don't meaning we are fine with just setting it, right?

carmocca: > Do you think we should maybe make this a weakref?

This will block pickling or deepcopying the LM, but it seems like that's already an issue considering the existence of the `_should_prevent_trainer_and_dataloaders_deepcopy` flag:

https://github.com/Lightning-AI/lightning/blob/feb8e7d344dae74d95dcacba95eaf75c2893cf3a/src/pytorch_lightning/core/module.py#L1966-L1967

Let me try setting the weakref to see if any tests complain.

> Is there ever a way where we do need to copy those or detach? From what I can think of, we don't meaning we are fine with just setting it, right?

Correct

src/pytorch_lightning/core/module.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         torch._C._log_api_usage_once(f"lightning.module.{self.__class__.__name__}")
 
         # pointer to the trainer object
-        self.trainer: Optional["pl.Trainer"] = None
+        self._trainer: Optional["pl.Trainer"] = None
```
carmocca, line 96: The idea is that we use the raw attribute if optional checks are desired, the getter otherwise.

Note that some areas of the codebase (saving, model summary, ...) can be used without a Trainer, so they need to use this attribute

src/pytorch_lightning/core/module.py
```diff
@@ -172,6 +173,21 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
         # multiple schedulers
         return lr_schedulers
 
+    @property
+    def trainer(self) -> "pl.Trainer":
```
carmocca, line 177: @awaelchli I'm pretty sure I broke the spawn launcher with this change:


```python
pytest trainer/properties/test_get_model.py::test_get_model_ddp_cpu
E         File "/home/carmocca/git/lightning/src/pytorch_lightning/trainer/trainer.py", line 2474, in training
E           return self.state.stage == RunningStage.TRAINING
E       AttributeError: 'Trainer' object has no attribute 'state'
```

But I'm not seeing the cause. Can you tell what I missed?

rohitgr7 (approved):

src/pytorch_lightning/core/module.py
awaelchli: @carmocca it looks like the trainer holding the loop and the trainer reference inside the fit loop are not the same. If you add
`assert self == self.fit_loop.trainer`
in line 1146 of trainer.py you get the assertion error. On master, this assertion holds true, so it must be something related to the setter/getter for trainer in the loops where it breaks.

awaelchli: If I remove the two lines

```py
        if not isinstance(trainer, weakref.ProxyTypes):
            trainer = weakref.proxy(trainer)
```
then the references are the same again and the error does not occur.

carmocca: Looks like the spawn pickling mechanism converts them to different objects, and since they don't share memory, the state is not set on the latter.

I'm going to undo the proxy change for the loops and see if it works for the LM at least

src/pytorch_lightning/core/module.py
```diff
@@ -172,6 +173,21 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
...
+        return self._trainer
+
+    @trainer.setter
+    def trainer(self, trainer: "pl.Trainer") -> None:
+        if not isinstance(trainer, weakref.ProxyTypes):
+            trainer = weakref.proxy(trainer)
+        self._trainer = trainer
+        for v in self.children():
+            if isinstance(v, LightningModule):
+                v.trainer = trainer
```
justusschock, line 189:
```suggestion
        for v in self.children():
            if isinstance(v, LightningModule):
                v.trainer = trainer
        if trainer is not None:
            trainer = weakref.proxy(trainer)
        self._trainer = trainer
       
```

very nit: This seems a bit cleaner to me (as you pass the core trainer reference to the children and they could theoretically do whatever they want with that). Also you spare one check and order shouldn't really matter here.
And IMO it does not make sense to have a weakref of `None`

justusschock (approved):
One very last nit (up to you!)

src/pytorch_lightning/core/module.py
carmocca: Looks good to me since `weakref.proxy(t) is weakref.proxy(t)` is True

otaj (approved):

src/pytorch_lightning/core/module.py
```diff
@@ -172,6 +173,21 @@ def lr_schedulers(self) -> Optional[Union[LRSchedulerTypeUnion, List[LRScheduler
...
+        if not self._running_torchscript and self._trainer is None:
+            raise RuntimeError(f"{self.__class__.__qualname__} is not attached to a `Trainer`.")
+        return self._trainer
+
+    @trainer.setter
+    def trainer(self, trainer: "pl.Trainer") -> None:
+        for v in self.children():
+            if isinstance(v, LightningModule):
+                v.trainer = trainer
+        if trainer is not None and not isinstance(trainer, weakref.ProxyTypes):
```
awaelchli, line 187: what is the reason for the None-check? Perhaps the input should be Optional[Trainer]?

awaelchli (approved):

src/pytorch_lightning/core/module.py
```diff
@@ -193,25 +209,25 @@ def example_input_array(self, example: Any) -> None:
     @property
     def current_epoch(self) -> int:
         """The current epoch in the ``Trainer``, or 0 if not attached."""
-        return self.trainer.current_epoch if self.trainer else 0
+        return self.trainer.current_epoch if self._trainer else 0
```
Borda, line 212: not sure why this changed?

Borda (approved):

src/pytorch_lightning/core/module.py
carmocca: > Disambiguates the trainer attribute (optional) from the property getter (non-optional)

Here we want to use the optional attribute, since `current_epoch` can be accessed without a Trainer. For example

```
my_model = MyModel()
print(my_model.current_epoch)
``

otaj:
It seems this PR introduced a failing test, which somehow haven't failed before.
https://github.com/Lightning-AI/lightning/actions/runs/3293585250/jobs/5434973295

The issue in the failing test is that the trainer instance is already garbage collected, which can happen with weakrefs.

cc @carmocca

carmocca:
Do you see it failing in different PRs? Do you suggest we remove the weakref, or that we ensure it doesn't get garbage collected in the test?

otaj:
I haven't noticed it in other PRs yet. Could be, that _something_ we introduced in the last week somewhat made it trigger, but I don't know have an idea of what it could be.

I think proper solution is to ensure it doesn't get garbage collected in the test, as I have a hard time imagining a situation in which a `Trainer` object gets garbage collected in real world use-case

carmocca:
I'm just not sure what I should change to fix it. This variable should already hold the trainer reference: https://github.com/Lightning-AI/lightning/blob/0fb31ed614e73be903f9d8b339247bae24440566/tests/tests_pytorch/utilities/test_parsing.py#L67

otaj:
Right, but that variable goes out of scope at the moment of returning from that function and is therefore free to get garbage-collected, since the only place where we have the reference is the weak reference. I think, what would help is to also return trainer instances from the `model_cases` function. I can do it myself.

Issue #11807, opened 2022-02-08
Prediction with custom batch sampler


stecklin:
## 🐛 Bug

When using a custom batch sampler for prediction, the following code gets executed:

https://github.com/PyTorchLightning/pytorch-lightning/blob/ab1c2ff23fd27ab6e5647e390313b569e32b61c5/pytorch_lightning/trainer/data_loading.py#L221-L226

This assumes the custom batch sampler has the same interface as [BatchSampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.BatchSampler). However, data loaders can take any [Sampler](https://pytorch.org/docs/stable/data.html#torch.utils.data.Sampler) as batch_sampler argument according to the docs: https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader

### To Reproduce

This is the custom sampler I was using:

```python
class MySampler(Sampler):
    def __init__(self, dataset):
        """
        Args:
            dataset (Dataset): PyTorch dataset.
        """
        self.dataset = dataset
        self.group_ids_to_sample_indices = self.get_group_ids_to_sample_indices()

    def get_group_ids_to_sample_indices(self):
        """ Group samples by id

        Returns:
            dict: group_id -> indices of samples with this id
        """
        group_ids_to_sample_indices = defaultdict(list)
        for i, sample in enumerate(self.dataset.samples):
            group_ids_to_sample_indices[sample["group_id"]].append(i)
        return group_ids_to_sample_indices

    def __iter__(self):
        group_ids = list(self.group_ids_to_sample_indices.keys())
        # Yield a batch of variable size with all samples for one group
        for group_id in group_ids:
            yield self.group_ids_to_sample_indices[group_id]

    def __len__(self):
        return len(self.group_ids_to_sample_indices)
```

### Expected behavior

Prediction runs without crashing. For training, validation and testing it already works.

### Environment

- PyTorch Lightning Version: 1.5.9
- PyTorch Version: 1.10.1

### Additional context

I did the following work around for now, but it feels very hacky:

```python
class MySampler(Sampler):
    def __init__(self, sampler=None, batch_size=None, drop_last=False, dataset=None):
        if sampler:
            self.dataset = sampler.data_source
        else:
            self.dataset = dataset
        self.group_ids_to_sample_indices = self.get_group_ids_to_sample_indices()
```

cc @tchaton @rohitgr7 @akihironitta @justusschock @awaelchli @ninginthecloud

[...]

PR #13640, opened 2022-07-13
Adds support for custom BatchSampler

### LINKED ISSUE #13007: Validation skipped when using custom batch sampler
## 🐛 Bug

Hi, I'm trying to use the batch sampler which was implemented as below,

```python
class BalancedBatchSampler(BatchSampler):
    def __init__(
        self, labels, n_classes_in_batch, n_samples_in_batch
    ):
        self.labels = labels
        self.label_set = list(set(self.labels))
        self.label_to_indices = {
            label: list(np.where(np.array(self.labels) == label)[0])
            for label in self.label_set
        }
        for l in self.label_set:
            np.random.shuffle(self.label_to_indices[l])
        self.used_label_indices_count = {label: 0 for label in self.label_set}

        self.n_dataset = len(self.labels)
        self.n_classes_in_batch = n_classes_in_batch
        self.n_samples_in_batch = n_samples_in_batch

        self.batch_size = self.n_samples_in_batch * self.n_classes_in_batch

    def __iter__(self):
        count = 0

        while count + self.batch_size < self.n_dataset:
            sampled_classes = np.random.choice(self.label_set, self.n_classes_in_batch, replace=False)
            indices = []

            for class_ in sampled_classes:
                pad_for_class = self.used_label_indices_count[class_]
                samples_for_class = self.label_to_indices[class_][pad_for_class: pad_for_class + self.n_samples_in_batch]
                indices += samples_for_class
                self.used_label_indices_count[class_] += self.n_samples_in_batch

                if self.used_label_indices_count[class_] + self.n_samples_in_batch > len(self.label_to_indices[class_]):
                    np.random.shuffle(self.label_to_indices[class_])
                    self.used_label_indices_count[class_] = 0

            yield indices
            count += self.batch_size

    def __len__(self):
        # step size in 1 epoch
        return self.n_dataset // self.batch_size
```

The sampler works well in the training stage of `trainer.fit()`, however, the validation stage has been skipped.

### To Reproduce

I confirmed that the behavior has not occurred when using `torch.utils.data.sampler.BatchSampler` as below,

```python
# train_dataloader+standard sampler * val_dataloader+standard sampler = validation not skipped
def train_dataloader(self):
    train_sampler = BatchSampler(RandomSampler(self.dataset_train), ...)
    return DataLoader(
        self.dataset_train,
        batch_sampler=train_sampler,
        ...
    )

def val_dataloader(self):
    train_sampler = BatchSampler(RandomSampler(self.dataset_val), ...)
    return DataLoader(
        self.dataset_val,
        batch_sampler=val_sampler,
        ...
    )

# train_dataloader+standard sampler * val_dataloader+custom sampler = validation not skipped
def train_dataloader(self):
    train_sampler = BatchSampler(...)
    return DataLoader(...)

def val_dataloader(self):
    val_sampler = BalancedBatchSampler(...)
    return DataLoader(...)

# train_dataloader+custom sampler * val_dataloader+standard sampler = validation skipped
def train_dataloader(self):
    train_sampler = BalancedBatchSampler(...)
    return DataLoader(...)

def val_dataloader(self):
    val_sampler = BatchSampler(...)
    return DataLoader(...)

# train_dataloader+custom sampler * val_dataloader+custom sampler = validation skipped
def train_dataloader(self):
    train_sampler = BalancedBatchSampler(...)
    return DataLoader(...)

def val_dataloader(self):
    val_sampler = BalancedBatchSampler(...)
    return DataLoader(...)
```

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.6.1
- PyTorch Version (e.g., 1.10): 1.11
- Python version (e.g., 3.9): 3.8.3
- OS (e.g., Linux): Linux
- CUDA/cuDNN version: 10.2

### Additional context

Unfortunately, I could not find the reason why the validation stage has been skipped, and what I'm missing in the implementation.
Is there proper practice to use a custom batch sampler?

cc @justusschock @awaelchli @ninginthecloud @rohitgr7

otaj:
## What does this PR do?

Adds support for custom `BatchSampler`

Fixes #13007 and #11807

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +322,44 @@ def _dataloader_init_kwargs_resolve_sampler(
     batch_sampler = getattr(dataloader, "batch_sampler")
     is_predicting = mode == RunningStage.PREDICTING
     # checking the batch sampler type is different than PyTorch default.
-    if batch_sampler is not None and (type(batch_sampler) is not BatchSampler or is_predicting):
-        batch_sampler = type(batch_sampler)(
-            sampler,
-            batch_size=batch_sampler.batch_size,
-            drop_last=(False if is_predicting else batch_sampler.drop_last),
-        )
+    batch_cls = type(batch_sampler)
```
carmocca, line 325:
```suggestion
    batch_sampler_cls = type(batch_sampler)
```
or
```suggestion
    sampler_cls = type(batch_sampler)
```

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +322,44 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+            try:
+                batch_sampler = batch_cls(
+                    sampler,
+                    batch_size=batch_sampler.batch_size,
+                    drop_last=(False if is_predicting else batch_sampler.drop_last),
+                )
+            except TypeError:
+                raise MisconfigurationException(
+                    "We tried to reinstantiate your custom batch sampler and failed. "
+                    "To mitigate this, either follow API of `BatchSampler` or instantiate "
```
carmocca, line 360:
```suggestion
                    "To mitigate this, either follow the API of `BatchSampler` or instantiate "
```

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +322,44 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+
+            batch_sampler = batch_cls(*args, **kwargs)
+        else:
+            try:
+                batch_sampler = batch_cls(
+                    sampler,
+                    batch_size=batch_sampler.batch_size,
+                    drop_last=(False if is_predicting else batch_sampler.drop_last),
+                )
+            except TypeError:
```
carmocca, line 357: This should check that the TypeError is a true positive as done in 

https://github.com/Lightning-AI/lightning/blob/daf7cec01e526c7503b99bfbf9e4c2048b8c8514/src/pytorch_lightning/utilities/data.py#L193-L202

src/pytorch_lightning/utilities/data.py
```diff
@@ -350,17 +382,36 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+
+    if inject_name in arg_names:
+        inject_index = arg_names.index(inject_name)
+        args = args[:inject_index] + [inject_object] + args[inject_index + 1 :]
+        return True, args, kwargs
+    elif inject_name in kwargs:
+        kwargs[inject_name] = inject_object
+        return True, args, kwargs
+
+    return False, args, kwargs
```
carmocca, line 401: What would be an example where we return False?

Shouldn't we include it as a new `kwarg` if the name isn't present already?

If the answer is no: this function should be called `_replace_value_in_saved`

src/pytorch_lightning/utilities/data.py
```diff
@@ -350,17 +382,36 @@ def _dataloader_init_kwargs_resolve_sampler(
     return {"sampler": sampler, "shuffle": False, "batch_sampler": None}
 
 
+def _inject_arg_to_saved(
+    args: List[Any], kwargs: Dict[str, Any], arg_names: List[str], inject_name: str, inject_object: Any
```
carmocca, line 386: super nit: I would put the inject argument key and value first in the signature

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +322,44 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+            args = list(batch_sampler.__pl_saved_args)
+            kwargs = batch_sampler.__pl_saved_kwargs
+            arg_names = batch_sampler.__pl_saved_arg_names
+
+            if is_predicting:
+                success, args, kwargs = _inject_arg_to_saved(args, kwargs, arg_names, "drop_last", False)
+                if not success:
+                    rank_zero_warn(
+                        "Trying to inject `drop_last=False` into batch sampler since you are predicting, however it "
+                        f"seems the class `{batch_cls}` does not support it. Your predictions might be incomplete. "
```
carmocca, line 337: The class is a class so this needs to use the qualname to avoid printing something like

`<class '__main__.Foo'>"`

```suggestion
                        f"seems the class `{batch_cls.__qualname__}` does not support it. Your predictions might be incomplete. "
```

Same comment applies to other usages of the class in a string

src/pytorch_lightning/utilities/data.py
```diff
@@ -371,18 +422,31 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
-        # We want to use the latest possible value for dataset argument (i.e. ideally what gets passed to DataLoader)
+        # We want to use the latest possible value for explicit arguments (i.e. ideally what gets passed to base class)
         # so that we can be sure, that it will not get changed anymore.
         # That is why we are setting this in every `__init__`
-        if "dataset" in param_names:
-            setattr(obj, "__dataset", args[param_names.index("dataset")])
-        elif "dataset" in kwargs:
-            setattr(obj, "__dataset", kwargs["dataset"])
+        if store_explicit_args is not None:
+            for explicit_arg in store_explicit_args:
```
carmocca, line 445: Is there any occurrence where we make use of the list?

src/pytorch_lightning/utilities/data.py
```diff
@@ -350,17 +382,36 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+
+    return False, args, kwargs
+
+
 def _auto_add_worker_init_fn(dataloader: DataLoader, rank: int) -> None:
     if int(os.environ.get("PL_SEED_WORKERS", 0)) and dataloader.worker_init_fn is None:
         dataloader.worker_init_fn = partial(pl_worker_init_function, rank=rank)
 
 
-def _wrap_dataloader_init(init: Callable) -> Callable:
```
carmocca, line 358: Do you think it would be useful to keep this function name, calling _wrap_init_method where `dataset`is passed as the explicit arg?

carmocca (commented):
This is getting really complex :joy:

Posting a partial review

src/pytorch_lightning/utilities/data.py
otaj: Done

src/pytorch_lightning/utilities/data.py
otaj: Done

src/pytorch_lightning/utilities/data.py
otaj: Done

src/pytorch_lightning/utilities/data.py
otaj: You're right the naming wasn't spot on, changed to your suggestion. False will be returned when we cannot find the argument in the argument list. For example, if we want to replace a sampler, but there is no sampler argument in `arg_names`, nor in `kwargs`

otaj: And yes, the answer is no :smile: We should not include it in `kwargs`, because the method might not support this particular argument

src/pytorch_lightning/utilities/data.py
otaj: I'm not sure I agree, but I don't have a strong preference either way, so changed according to your suggestion :smile:

src/pytorch_lightning/utilities/data.py
otaj: No. What happened was that I thought we might need some explicit args for the `BatchSampler` and had no idea how many, so changed it to list and then forgot about it to change it back :) I now made it a single string only, since there is no usecase where the list would be necessary, making it back to the list should not be a big change in the future if we ever need it

src/pytorch_lightning/utilities/data.py
otaj: Umm... no. I don't see an added value in keeping around multiple specialized functions, when they can be made one generic function while changing one or two arguments

otaj:
Thank you for the review! I addressed all your concerns :+1: 

Btw, regarding the complexity, it surprisingly wasn't too bad for me, since it's basically just logic from dataloader handling made generic

src/pytorch_lightning/utilities/data.py
```diff
@@ -371,18 +432,30 @@ def wrapper(obj: DataLoader, *args: Any, **kwargs: Any) -> None:
...
 
-        if not hasattr(obj, "__pl_dl_args"):
-            obj.__pl_dl_args = args
-            obj.__pl_dl_kwargs = kwargs
-            obj.__pl_dl_arg_names = param_names
+        default_kwargs = {
+            param.name: param.default
+            for param in params.values()
+            if param.name != "self"
+            and param.kind not in (param.VAR_POSITIONAL, param.VAR_KEYWORD)
```
carmocca, line 439: This is very similar to the tuple created in L428. Could we create this dict first and extract the tuple from the keys of this dict?

Then you could remove the filter `(param.name not in kwargs and param.name not in param_names)`

tests/tests_pytorch/utilities/test_data.py
```diff
@@ -311,14 +313,14 @@ def test_replace_dataloader_init_method(cls, args, kwargs, arg_names, dataset, c
         if isinstance(dataloader_value, torch.Tensor):
             assert dataloader_value is value
         else:
-            assert getattr(dataloader, key) == value
+            assert dataloader_value == value
```
carmocca, line 316: why does this use `==` vs `is` based on the type?

tests/tests_pytorch/utilities/test_data.py
```diff
@@ -328,7 +330,118 @@ def test_replace_dataloader_init_method(cls, args, kwargs, arg_names, dataset, c
...
+
+
+def test_replace_init_method_extra_kwargs():
```
carmocca, line 336: Was this case not covered by the existing dataloader tests?

tests/tests_pytorch/utilities/test_data.py
```diff
@@ -328,7 +330,118 @@ def test_replace_dataloader_init_method(cls, args, kwargs, arg_names, dataset, c
...
+
+@pytest.mark.parametrize("predicting", [True, False])
+def test_custom_batch_sampler(predicting):
+    class MyBatchSampler(BatchSampler):
+        def __init__(self, sampler, extra_arg, drop_last=True):
+            self.extra_arg = extra_arg
+            super().__init__(sampler, 10, drop_last)
+
+    with _replace_init_method(BatchSampler):
+        sampler = RandomSampler(range(10))
```
carmocca, line 358: This instantiation does not need to happen inside the context manager no? Same for the DataLoader.

Same comment applies to the other tests

carmocca (approved):
Great! Minor comments

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +322,54 @@ def _dataloader_init_kwargs_resolve_sampler(
...
     is_predicting = mode == RunningStage.PREDICTING
     # checking the batch sampler type is different than PyTorch default.
-    if batch_sampler is not None and (type(batch_sampler) is not BatchSampler or is_predicting):
-        batch_sampler = type(batch_sampler)(
-            sampler,
-            batch_size=batch_sampler.batch_size,
-            drop_last=(False if is_predicting else batch_sampler.drop_last),
-        )
+    batch_sampler_cls = type(batch_sampler)
+    if batch_sampler is not None and (batch_sampler_cls is not BatchSampler or is_predicting):
```
tchaton, line 326: Hey @otaj. I am confused here. Why do we need to re0-instantiate the BatchSampler of the users ?

This was done for data loaders because they block settings arguments once instantiated. But here, we can modify it in-place.

otaj: Hi @tchaton, fair question and to be honest, my best answer is "it was done like this when I came to this code". However, it might be better for custom batch samplers since they might do some setup in `__init__` based on received Sampler.

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +322,54 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+        if hasattr(batch_sampler, "__pl_saved_args"):
+            args = list(batch_sampler.__pl_saved_args)
```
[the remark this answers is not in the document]
otaj: Because I always forget, that you can concatenate tuples, for some reason, I think of them as more primitive than they actually are. :facepalm: 

Changed to tuples

src/pytorch_lightning/utilities/data.py
otaj: So, you're right about the fact, that it is very similar. I changed it according to your suggestion and honestly, it doesn't like like such an improvement to me. The only advantage I see is that now the intention of both of these structures is a little bit clearer.

tests/tests_pytorch/utilities/test_data.py
otaj: Surprisingly no. And they way existing tests were set up, it couldn't even be covered, because it checked whether `__pl_saved_kwargs == kwargs`, which were the ones passed to the test. So it did not do anything with the default kwargs at all

tests/tests_pytorch/utilities/test_data.py
otaj: Resolved

tests/tests_pytorch/utilities/test_data.py
otaj: Because checking tensor for equality using == requires `all()`. And since I am keeping the reference, because the data is random, it is easier to check for object equality. Btw, this was there from previous PR. Let me know, if you think I should change it to something else.

src/pytorch_lightning/utilities/data.py
carmocca: I like this version better. Thanks!

rohitgr7 (approved):
looks good! high level review.

minor comments.

carmocca:
@otaj There's a failing test. Also, @awaelchli would you kindly review?

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +323,54 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+
+            if is_predicting:
+                success, args, kwargs = _replace_value_in_saved_args("drop_last", False, args, kwargs, arg_names)
+                if not success:
+                    rank_zero_warn(
+                        f"Trying to inject `drop_last=False` into batch sampler since you are predicting, however it "
+                        f"seems the class `{batch_sampler_cls.__qualname__}` does not support it. "
+                        "Your predictions might be incomplete. To mitigate this, expose `drop_last` in the `__init__` "
+                        "method of your custom class."
+                    )
```
awaelchli, line 341: This is not something we do for the DataLoader I hope, right? Because there the drop_last argument is False by default. 

For the batch sampler, I don't fully understand the reasoning here. [The default value is not set](https://pytorch.org/docs/stable/data.html#torch.utils.data.BatchSampler), which means the user has to always provide the value explicitly. Why are we trying to change it then? 

So if my logic is correct here, I would strongly suggest not to interfere with the value for drop_last here, and only send out a warning.

src/pytorch_lightning/utilities/data.py
```diff
@@ -322,12 +323,54 @@ def _dataloader_init_kwargs_resolve_sampler(
...
+
+            success, args, kwargs = _replace_value_in_saved_args("sampler", sampler, args, kwargs, arg_names)
+            if not success:
+                raise MisconfigurationException(
+                    "Trying to inject modified sampler into the batch sampler; however, it seems the class "
+                    f"`{batch_sampler_cls.__qualname__}` does not support argument called sampler. To mitigate this, "
+                    "expose argument `sampler` in the `__init__` method of your custom class."
```
awaelchli, line 348:
```suggestion
                    "Trying to inject a modified sampler into the batch sampler; however, it seems the class "
                    f"`{batch_sampler_cls.__qualname__}` does not have an argument called `sampler.` To mitigate this, "
                    "expose an argument `sampler` in the `__init__` method of your custom class."
```

src/pytorch_lightning/utilities/data.py
awaelchli: Edit: I just saw that we already had this logic before
```
if batch_sampler is not None and (type(batch_sampler) is not BatchSampler or is_predicting):
        batch_sampler = type(batch_sampler)(
            sampler,
            batch_size=batch_sampler.batch_size,
            drop_last=(False if is_predicting else batch_sampler.drop_last),
        )
```
I disagree with this and will open a separate issue about it. So ignore my comment for now as it does not concern this PR.

awaelchli (approved):

tests/tests_pytorch/utilities/test_data.py
```diff
@@ -328,7 +330,123 @@ def test_replace_dataloader_init_method(cls, args, kwargs, arg_names, dataset, c
...
+        dataloader = LoaderSubclass(range(10))
+
+    assert dataloader.__pl_saved_args == (range(10),)
+    assert dataloader.__pl_saved_kwargs == {"batch_size": 10}
+    assert dataloader.__pl_saved_arg_names == ("dataset",)
+    assert dataloader.__dataset == range(10)
+
+
+@pytest.mark.parametrize("predicting", [True, False])
+def test_custom_batch_sampler(predicting):
```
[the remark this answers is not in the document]
otaj: Done. Let me know what you think

otaj:
@awaelchli thank you very much for the review! I hope I addressed everything

otaj:
> @otaj There's a failing test. Also, @awaelchli would you kindly review?

@carmocca that should be taken care of now. It was a simple issue of changing the text of exception raised not matching the pytest catching

carmocca:
IPU failure looks flaky. I've rerun twice, going for thrice

otaj:
> IPU failure looks flaky. I've rerun twice, going for thrice

And it failed again. Can we maybe mark the test as flaky, or set a bit lower threshold?

awaelchli:
I ran it twice too. Does not look flaky to me. The environment must have changed, we may start seeing it on other PRs too. There is also the possibility that it's related to the PR. Note that in the IPU strategy we use the `_get_dataloader_init_args_and_kwargs` that was modified here. Just pointing it out as a possibility.

otaj:
> Note that in the IPU strategy we use the `_get_dataloader_init_args_and_kwargs` that was modified here. Just pointing it out as a possibility.

You're absolutely right regarding this, however the code should be functionally the same for normal dataloaders. But, to be fair, it doesn't seem to fail on other PRs, so it's quite likely that it is indeed a regression in this PR.

Let's try to re-run once or twice more and if it is still failing, let's investigate more

otaj:
There is something fishy going on. If I try to modify the `test_optimization` to force it on cpu (change accelerator, remove `RunIf` and remove `dm.trainer` checks), I pass the test every single time, and comfortably. So it really seems the issue is introduced in this PR. :facepalm:

otaj:
@carmocca, @awaelchli I finally managed to properly debug the cause of the IPU regression. The problem was, that I was passing all default kwargs from `torch.utils.data.DataLoader` to `poptorch.DataLoader` which didn't play nicely with it. I don't think there's a substantial change from when you reviewed it (just split `__pl_saved_kwargs` into `__pl_saved_kwargs` and `__pl_saved_default_kwargs`).

psridhar-asapp:
Hey @otaj I am confused on how to update my sampler to fit this API.

We are dealing with audio data and we want to batch samples by sequence length. In order to do this, we have shape files that are CSVs mapping between each audio sample and its length. Our BatchSampler reads these files, sorts utterances by length, and then constructs fixed batches of similarly long utterances such that all batches are similar in size.

I am not sure what to move to a Sampler here. We could move the loading of this shapes file to the sampler backed by a new dataset, but we do not sample from the shapes file to build batches, we consume the entire file at once.

I think part of my doubt is a lack of understanding on what exactly you do to the Sampler passed into the BatchSampler. I could move the loading of the shapes file inside a Dataset, pass in a Sampler to this dataset as a hacky way of giving the dataset to the BatchSampler, and totally ignore the Sampler in my BatchSampler logic while just reading from its member Dataset, but I imagine this will not solve the problem I'm having.

otaj:
Hi @psridhar-asapp. First things first, this doesn't really *add* new API. It keeps the functionality as before, just know it allows your custom `BatchSampler` to have also more arguments, that don't appear in a signature of regular `BatchSampler`.

I remember we spoke on this topic in some other issue (I think it was #11807). I don't think we have a good resolution to it yet, honestly.

We don't really do *much* to your sampler. Everything important regarding `Sampler` and `DataLoader` is happening in this method (and of course, all the other methods it calls). https://github.com/Lightning-AI/lightning/blob/caaf35689c585fba1bb33b70243e7ddd2cf5d13c/src/pytorch_lightning/trainer/connectors/data_connector.py#L238 

Currently, it sets shuffling if it was set to None, and we're training, wraps it in a distributed wrapper and does a bit more of a magic in `_update_dataloader`, which is just juggling arguments. The most important bit of call tree in `_update_dataloader` is honestly this function https://github.com/Lightning-AI/lightning/blob/caaf35689c585fba1bb33b70243e7ddd2cf5d13c/src/pytorch_lightning/utilities/data.py#L315

That is where we wrap either `BatchSampler` or `Sampler` in structures used for prediction or fault tolerance. I _think_ (and _think_ is the operative word here) it might be possible to remove the `Sampler` from being passed around and you could be happy with your `BatchSampler` only solution, but I don't really have a clear vision about that yet and it will require non-trivial engineering work.

If you don't care about either of these things, you can just expose argument `sampler` in your custom `BatchSampler`, (which needs to be a proper subclass of it and instantiated in `*_dataloader` hook) and don't use that sampler at all. I'm sorry, I don't have a better solution for you

PR #13658, opened 2022-07-14
Add support for async checkpointing

### LINKED ISSUE #11561: Asynchronous checkpointing
## 🚀 Feature

The idea is to enable asynchronous model checkpointing during training.

### Motivation

I'm training thousands of small identical `gpytorch` models, each one is trained for 1000 epochs on a different 1d time series sample. The training on a GPU takes ~1/4 seconds of time. Saving the model at the end takes about 1.25 seconds, and this blocks the training of the next models and wastes GPU time.

### Pitch

Checkpointing currently blocks the training process and wastes GPU time. If implemented correctly, asynchronous (non-blocking) checkpointing could offer a significant speed up over time.

### Alternatives

Perhaps this doesn't need to be built into PyTorch-Lightning and can be implemented with asyncio.
If so, I would like to see an example.

### Additional context

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7

rohitgr7:
## What does this PR do?

Fixes #11561
## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7 @otaj @akihironitta


src/pytorch_lightning/utilities/cloud_io.py
```diff
@@ -24,6 +27,30 @@
...
+class ThreadQueue(threading.Thread):
+    def __init__(self, func: Callable, q: queue.Queue, interval: int = 5) -> None:
+        super().__init__(daemon=True)
+        self.func = func
+        self.q = q
+        self._close_thread = False
+        self._interval = interval
+
+    def run(self) -> None:
+        with _state_observer_lock:
```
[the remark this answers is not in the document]
tchaton: @rohitgr7 Here is an even cleaner way to handle the get from all the threads: https://github.com/SeldonIO/MLServer/blob/92e0a1aa01fe74bcb55d3d1ae2f783148a9e3151/mlserver/parallel/worker.py#L81.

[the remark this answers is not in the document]
rohitgr7: I used a lock here to ensure that it works correctly if the same filepath is used. if there are multiple checkpoints in the queue with the same filepath, we need to ensure that they get pulled out in a sequence, and all other threads should be blocked until then.

But just wondering, in such a case, multiple threads doesn't provide any value.

src/pytorch_lightning/plugins/io/torch_plugin.py
```diff
@@ -13,21 +13,52 @@
...
     """CheckpointIO that utilizes :func:`torch.save` and :func:`torch.load` to save and load checkpoints
-    respectively, common for most use cases."""
+    respectively, common for most use cases.
+
+    Args:
+        save_async: whether to save the checkpoint asynchronously or not.
+        num_threads: Number of threads to use for asynchronous checkpointing.
+    """
+
+    def __init__(self, save_async: bool = False, num_threads: Optional[int] = None):
```
carmocca, line 40: I'm not convinced that these should be flags on this plugin. IMO it should be a separate `AsyncTorchCheckpointIO` class. This will be very useful for code that does not support the async mechanism, because then it can check `if isisntance(AsyncTorchCheckpointIO)`

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -97,6 +97,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added Apple Silicon Support via `MPSAccelerator` ([#13123](https://github.com/PyTorchLightning/pytorch-lightning/pull/13123))
 
 
+- Added support for async checkpointing ([#13658](https://github.com/PyTorchLightning/pytorch-lightning/pull/13658))
```
carmocca, line 100: Reminder to add docs and experimental notices

src/pytorch_lightning/plugins/io/torch_plugin.py
```diff
@@ -13,21 +13,52 @@
...
+            )
+
+        if save_async:
+            rank_zero_info(
+                "With asynchronous checkpointing, it is recommended to set"
+                " `ModelCheckpoint(save_top_k=-1)` if your checkpoint interval is low."
+            )
+
+        self.queue = None
+        self.threads = None
```
awaelchli, line 53: I don't understand why we are using multiple threads. How does this speed up saving? Unless we are saving to a remote location by actually establishing independent connections, we can't scale IO like this just by having more writers. Or how does it work?
I highly recommend only having one thread that pulls from a queue (list) of checkpoints. This is easier to reason about and test as we can guarantee the order in which files are being saved.

src/pytorch_lightning/strategies/strategy.py
```diff
@@ -475,7 +475,7 @@ def on_predict_start(self) -> None:
 
     def on_train_end(self) -> None:
         """Called when train ends."""
-        pass
+        self.checkpoint_io.on_train_end()
```
awaelchli, line 478: consider moving this to teardown?

src/pytorch_lightning/utilities/cloud_io.py
```diff
@@ -24,6 +27,30 @@
 
 from pytorch_lightning.utilities.types import _PATH
 
+_state_observer_lock = threading.Lock()
```
awaelchli, line 30: This has nothing to do with the state observer lock in the app. Can we remove it, or rename it to something else?

src/pytorch_lightning/plugins/io/torch_plugin.py
rohitgr7: yep.. will update to single thread

rohitgr7:
after discussion with @carmocca 

the design
```py
class AsyncCheckpointIO(CheckpointIO):
    def __init__(self, checkpoint_io=None):

    @setter
    def checkpoint_io(...):
        ...

    def save_checkpoint(self, checkpoint, filepath)
        async_launch(self._checkpoint_io.save_checkpoint, checkpoint, filepath)
```
and in the Trainer, the user can just pass
```py
Trainer(plugins=[AsyncCheckpointIO()])
```
we will attach the appropriate checkpoint if not already attached. This will enable async checkpointing for existing plugins by providing async behavior without any code modification.

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    def __init__(self, checkpoint_io: Optional["CheckpointIO"] = None, interval: float = 2.0) -> None:
+        super().__init__()
+
+        self._checkpoint_io = checkpoint_io
+        self._thread = _ThreadQueue(q=queue.Queue(), interval=interval)
+        self._thread.start()
+
+    @property
```
otaj, line 39: How about moving this property to base `CheckpointIO`? The advantage would be to have the setter recursive and supporting multiple wrappers around each other such as `SplitCheckpointIO(AsyncCheckpointIO(HPUCheckpointIO()))`

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import Any, Dict, Optional
+
+from pytorch_lightning.plugins.io.checkpoint_plugin import CheckpointIO
+from pytorch_lightning.utilities.cloud_io import _ThreadQueue
+from pytorch_lightning.utilities.types import _PATH
+
+
+class AsyncCheckpointIO(CheckpointIO):
+    """CheckpointIO that utilizes :func:`torch.save` and :func:`torch.load` to save and load checkpoints
+    respectively, common for most use cases.
```
carmocca, line 25: You don't know that torch will be used. This should introduce what is the concept of async checkpointing instead.

src/pytorch_lightning/plugins/io/async_plugin.py
carmocca: I don't think this fits in the base class. Perhaps in a middle step between the base and this one. `_WrappingCheckpointIO`? Then `is_wrapper` would also be an `isinstance` check on this middle class

src/pytorch_lightning/plugins/io/async_plugin.py
rohitgr7: updated!

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -94,3 +98,41 @@ Custom Checkpoint IO Plugin
...
+**************************
+
+.. warning::
+
+    This is currently an experimental plugin/feature and API changes are to be expected.
+
+To enable saving the checkpoints asynchronously without blocking your training, you can configure
+:class:`~pytorch_lightning.plugins.io.async_plugin.AsyncCheckpointIO` plugin to ``Trainer``.
+
+It uses it's base ``CheckpointIO`` plugin's saving logic to save the checkpoint but performs the this operation asynchronously.
```
otaj, line 114:
```suggestion
It uses its base ``CheckpointIO`` plugin's saving logic to save the checkpoint but performs this operation asynchronously.
```

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -94,3 +98,41 @@ Custom Checkpoint IO Plugin
...
+   from pytorch_lightning.plugins.io import AsyncCheckpointIO
+
+
+   async_ckpt_io = AsyncCheckpointIO()
+   trainer = Trainer(plugins=[async_ckpt_io])
+   model = LitModel()
+   trainer.fit(model)
+
+
+If you want to plugin your own base ``CheckpointIO`` and want it be behave asynchronously, pass it while initializing ``AsyncCheckpointIO``.
```
otaj, line 128:
```suggestion
If you want the plugin to use your own base ``CheckpointIO`` and want the base to behave asynchronously, pass it as an argument while initializing ``AsyncCheckpointIO``.
```

otaj (approved):

src/pytorch_lightning/plugins/io/wrapper.py
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Optional
+
+from pytorch_lightning.plugins.io.checkpoint_plugin import CheckpointIO
+
+
+class _WrappingCheckpointIO:
```
otaj, line 19: It might be better if `_WrappingCheckpointIO` was a subclass of `CheckpointIO`, so that the inheritance is easier. Plus, it can have the `{save,load,remove}_checkpoint` methods, which would automatically check for `self._base_checkpoint_io_configured == True`. And the implementations of this class (such as `AsyncCheckpointIO`) could call `super().save_checkpoint(...)` instead of `self._checkpoint_io.save_checkpoint(...)` which is arguably more pythonic since it doesn't use private attributes

rohitgr7: I can still access it using `self.checkpoint_io.load_checkpoint()` since there's a `checkpoint_io` getter in `_WrappingCheckpointIO`. WDYT? I am hesitating to add `CheckpointIO` as the base because it's not actually a `CheckpointIO` class.

otaj: I think the main benefit from my point of view would be automatic check, whether `self._base_checkpoint_io_configured == True` and simplified inheritance (the implementations would not need to inherit from two classes but just one.)

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,75 @@
...
+    Args:
+        checkpoint_io: A checkpoint IO plugin that is used as the basis for async checkpointing.
+        interval: Sleep time between each queue check.
+    """
+
+    def __init__(self, checkpoint_io: Optional["CheckpointIO"] = None, interval: float = 2.0) -> None:
+        super().__init__(checkpoint_io)
+
+        self._thread = _ThreadQueue(q=queue.Queue(), interval=interval)
+        self._thread.start()
```
awaelchli, line 40: I recommend not starting the thread on init, but dynamically on the first call to save_checkpoint().
This allows us to use this with multiprocessing and don't need to worry about locks or threads not being picklable or forkable.

src/pytorch_lightning/plugins/io/wrapper.py
carmocca: I agree with Ota

tests/tests_pytorch/plugins/test_checkpoint_io_plugin.py
```diff
@@ -68,12 +78,76 @@ def test_checkpoint_plugin_called(tmpdir):
...
+
+    assert checkpoint_plugin.save_checkpoint.call_count == 3
+    assert checkpoint_plugin.remove_checkpoint.call_count == 1
+
+    base_ckpt_io = trainer.strategy.checkpoint_io.checkpoint_io
+    assert base_ckpt_io.save_checkpoint.call_count == 3
+    assert base_ckpt_io.remove_checkpoint.call_count == 1
+
+
+def test_multi_wrapped_checkpoint_io_initialization():
```
awaelchli, line 137: Please make sure that the thread(s) get joined when exiting the test.

src/pytorch_lightning/plugins/io/checkpoint_plugin.py
```diff
@@ -60,3 +60,14 @@ def remove_checkpoint(self, path: _PATH) -> None:
         Args:
             path: Path to checkpoint
         """
+
+    @property
+    def is_wrapper(self) -> bool:
+        """Property that tells whether a ``CheckpointIO`` is a wrapper around another ``CheckpointIO`` object."""
+        from pytorch_lightning.plugins.io.wrapper import _WrappingCheckpointIO
+
+        return isinstance(self, _WrappingCheckpointIO)
```
carmocca, line 69: IMO this property shouldn't exist here, and the places that call this should use `isinstance` directly

src/pytorch_lightning/utilities/cloud_io.py
```diff
@@ -25,6 +28,26 @@
 from pytorch_lightning.utilities.types import _PATH
 
 
+class _ThreadQueue(threading.Thread):
```
awaelchli, line 31: Let's use ThreadPoolExecutor, we don't need to reinvent the wheel :)

In the plugin do:
`executor = ThreadPoolExecutor(1)`

in save_checkpoint:
`executor.submit(func, args)`

awaelchli (changes requested):

src/pytorch_lightning/utilities/cloud_io.py
rohitgr7: But I need to join the queue and thread properly, can't find a way to do that using `ThreadPoolExecutor`.

tests/tests_pytorch/plugins/test_checkpoint_io_plugin.py
rohitgr7: they are on teardown.

src/pytorch_lightning/plugins/io/wrapper.py
rohitgr7: done!

src/pytorch_lightning/plugins/io/async_plugin.py
rohitgr7: done!

awaelchli:
Probably worth mentioning in the docs (for now) is that ANY error raised in the thread will not be raised in the main thread. 
So if you get an io error with torch.save for example, it won't be surfaced like ever.

Rabbit hole alert: A way to handle this would be to catch the exception in the function passed to the executor, and then add it to a shared variable. The main thread could then periodically check whether there was an error and raise it.

src/pytorch_lightning/utilities/cloud_io.py
awaelchli: You don't need to, because it will do so when the task finishes.

rohitgr7: will it make sure that the thread will run continuously until the queue is empty?

awaelchli: Yes, as long as there are items on the queue, the workers will do work.
https://superfastpython.com/threadpoolexecutor-in-python/#How_Does_ThreadPoolExecutor_Work_Internally

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -94,3 +98,41 @@ Custom Checkpoint IO Plugin
...
+
+.. code-block:: python
+
+   from pytorch_lightning.plugins.io import AsyncCheckpointIO
+
+
+   async_ckpt_io = AsyncCheckpointIO()
+   trainer = Trainer(plugins=[async_ckpt_io])
+   model = LitModel()
+   trainer.fit(model)
```
carmocca, line 125: nit: this doesn't seem relevant

```suggestion
```

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,62 @@
...
+    Args:
+        checkpoint_io: A checkpoint IO plugin that is used as the basis for async checkpointing.
+        interval: Sleep time between each queue check.
+    """
+
+    def __init__(self, checkpoint_io: Optional["CheckpointIO"] = None, interval: float = 2.0) -> None:
+        super().__init__(checkpoint_io)
+
+        self._executor = ThreadPoolExecutor(max_workers=1)
+        self._error = None
```
carmocca, line 38: Doesn't this need typing?

src/pytorch_lightning/plugins/io/wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, Dict, Optional
+
+from pytorch_lightning.plugins.io.checkpoint_plugin import CheckpointIO
+
+
+class _WrappingCheckpointIO(CheckpointIO):
+    """AsyncCheckpointIO enables saving the checkpoints asynchronously.
```
carmocca, line 20: This is copy-pasted

src/pytorch_lightning/plugins/io/wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+
+    .. warning::
+
+        This is currently an experimental plugin/feature and API changes are to be expected.
+
+    Args:
+        checkpoint_io: A checkpoint IO plugin that is used as the basis for async checkpointing.
+        interval: Sleep time between each queue check.
```
carmocca, line 28: Same. Not the args, incorrect descriptions

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,62 @@
...
+        self._executor.submit(_save_checkpoint, *args, **kwargs)
+
+        if self._error:
+            raise self._error
+
+    def remove_checkpoint(self, *args: Any, **kwargs: Any) -> None:
+        super().remove_checkpoint(*args, **kwargs)
+
+    def load_checkpoint(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
+        return super().load_checkpoint(*args, **kwargs)
```
carmocca, line 58: Why do you override these?

src/pytorch_lightning/plugins/io/wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+        self._base_checkpoint_io_configured: bool = False
+
+        if checkpoint_io is not None:
+            if isinstance(checkpoint_io, _WrappingCheckpointIO):
+                self._base_checkpoint_io_configured = checkpoint_io._base_checkpoint_io_configured
+            else:
+                self._base_checkpoint_io_configured = True
+
+    @property
+    def checkpoint_io(self):
```
carmocca, line 44: Typing?

src/pytorch_lightning/plugins/io/wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+        return self._checkpoint_io
+
+    @checkpoint_io.setter
+    def checkpoint_io(self, checkpoint_io: "CheckpointIO") -> None:
+        assert not isinstance(checkpoint_io, _WrappingCheckpointIO)
```
carmocca, line 49: Why is this a limitation?

src/pytorch_lightning/plugins/io/wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Args:
+        checkpoint_io: A checkpoint IO plugin that is used as the basis for async checkpointing.
+        interval: Sleep time between each queue check.
+    """
+
+    def __init__(self, checkpoint_io: Optional["CheckpointIO"] = None) -> None:
+        super().__init__()
+
+        self._checkpoint_io = checkpoint_io
```
carmocca, line 34: IMO this should not be an Optional attribute, meaning that if `None` was passed, we default to `TorchCheckpointIO`

src/pytorch_lightning/strategies/single_tpu.py
```diff
@@ -16,6 +16,7 @@
 
 import pytorch_lightning as pl
 from pytorch_lightning.plugins.io.checkpoint_plugin import CheckpointIO
+from pytorch_lightning.plugins.io.wrapper import _WrappingCheckpointIO
```
carmocca, line 19: Aren't these unused imports?

src/pytorch_lightning/plugins/io/wrapper.py
rohitgr7: let me revise the docs again.

src/pytorch_lightning/plugins/io/wrapper.py
rohitgr7: then it would be hard for us to tell which checkpoint io to use for eg. let's say when the user is using HPU, we can't tell whether it is passed by the user or is the default one unless we add some assertions or boolean flags...

Just wanted to keep it a little agnostic and let strategies decide the default one on their own.

src/pytorch_lightning/plugins/io/wrapper.py
rohitgr7: this is the base checkpoint, which is initialized by strategies internally and should be a base checkpoint, not a wrapper.

otaj: But it could easily be a wrapper, a user can set this. If it is a wrapper, then it makes sense to do something with that - set this wrapper and set original base as the base of the new wrapper

src/pytorch_lightning/plugins/io/wrapper.py
otaj: I have to agree with @rohitgr7 that None is a good default value to differentiate for the cases where the `checkpoint_io` hasn't been set explicitly and has to be set by strategies.

src/pytorch_lightning/plugins/io/wrapper.py
rohitgr7: some issue with my local env. I ran mypy to check any missing ones

src/pytorch_lightning/plugins/io/wrapper.py
rohitgr7: yeah they can.. by passing it to the init.

rohitgr7: this property is used by a strategy to set the defaults.

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -45,6 +45,10 @@ Built-in Checkpoint IO Plugins
        respectively, common for most use cases.
    * - :class:`~pytorch_lightning.plugins.io.XLACheckpointIO`
      - CheckpointIO that utilizes :func:`xm.save` to save checkpoints for TPU training strategies.
+   * - :class:`~pytorch_lightning.plugins.io.HPUCheckpointIO`
+     - CheckpointIO to save checkpoints for HPU training strategies.
+   * - :class:`~pytorch_lightning.plugins.io.AsyncCheckpointIO`
+     - ``AsyncCheckpointIO`` enables saving the checkpoints asynchronously using the ``ThreadPoolExecutor``.
```
awaelchli, line 51:
```suggestion
     - ``AsyncCheckpointIO`` enables saving the checkpoints asynchronously in a thread.
```
I would not mention the specific method we use. We are also only using a pool size of 1 as a special case.

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -94,3 +98,37 @@ Custom Checkpoint IO Plugin
...
+**************************
+
+.. warning::
+
+    This is currently an experimental plugin/feature and API changes are to be expected.
+
+To enable saving the checkpoints asynchronously without blocking your training, you can configure
+:class:`~pytorch_lightning.plugins.io.async_plugin.AsyncCheckpointIO` plugin to ``Trainer``.
+
+It uses its base ``CheckpointIO`` plugin's saving logic to save the checkpoint but performs this operation asynchronously.
```
awaelchli, line 114: I would not mention any "base" checkpoint io or wrapping logic here, because it's not needed in the code sample below. Only mention it in the second paragraph where it matters.

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+class AsyncCheckpointIO(_WrappingCheckpointIO):
+    """``AsyncCheckpointIO`` enables saving the checkpoints asynchronously using the ``ThreadPoolExecutor``.
+
+    .. warning::
+
+        This is currently an experimental plugin/feature and API changes are to be expected.
+
+    Args:
+        checkpoint_io: A checkpoint IO plugin that is used as the basis for async checkpointing.
+        max_workers: Number of threads.
```
awaelchli, line 31: I vote for not exposing number of threads.

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+            try:
+                assert self.checkpoint_io is not None
+                self.checkpoint_io.save_checkpoint(*args, **kwargs)
+            except BaseException as e:
+                self._error = e
+
+        self._executor.submit(_save_checkpoint, *args, **kwargs)
+
+        if self._error:
+            raise self._error
```
awaelchli, line 53: A comment is appropriate here that this only raises the error if an error was raised between the previous time save_checkpoint was called and now, because executor.submit is not blocking. 

And given that, you also need to raise in teardown, because technically the error could be raised in the very last checkpoint you save. 

Also, a super nit edge case, but it's not handling different errors from multiple threads. I still strongly suggest to keep it at 1 and not expose num_threads to user.

src/pytorch_lightning/plugins/io/checkpoint_plugin.py
```diff
@@ -43,12 +43,13 @@ def save_checkpoint(self, checkpoint: Dict[str, Any], path: _PATH, storage_optio
...
     @abstractmethod
-    def load_checkpoint(self, path: _PATH, storage_options: Optional[Any] = None) -> Dict[str, Any]:
+    def load_checkpoint(self, path: _PATH, map_location: Optional[Any] = None) -> Dict[str, Any]:
         """Load checkpoint from a path when resuming or loading ckpt for test/validate/predict stages.
 
         Args:
             path: Path to checkpoint
-            storage_options: Optional parameters when loading the model/training states.
+            map_location: a function, :class:`torch.device`, string or a dict specifying how to remap storage
+            locations.
```
awaelchli, line 52:
```suggestion
                locations.
```

awaelchli (approved):

docs/source-pytorch/common/checkpointing_expert.rst
```diff
@@ -94,3 +98,37 @@ Custom Checkpoint IO Plugin
 .. note::
 
     Some ``TrainingTypePlugins`` like ``DeepSpeedStrategy`` do not support custom ``CheckpointIO`` as checkpointing logic is not modifiable.
+
+
+**************************
+Asynchronous Checkpointing
```
tchaton, line 104: Did you do any benchmarking?

src/pytorch_lightning/plugins/io/async_plugin.py
```diff
@@ -0,0 +1,57 @@
...
+        def _save_checkpoint(*args: Any, **kwargs: Any) -> None:
+            try:
+                assert self.checkpoint_io is not None
+                self.checkpoint_io.save_checkpoint(*args, **kwargs)
+            except BaseException as e:
+                self._error = e
+
+        self._executor.submit(_save_checkpoint, *args, **kwargs)
+
+        if self._error:
```
tchaton, line 52: This might never be reached as the exception might come after the main thread pass over this.

src/pytorch_lightning/plugins/io/async_plugin.py
rohitgr7: any particular reason?

docs/source-pytorch/common/checkpointing_expert.rst
rohitgr7: very basic one, just on BoringModel() trained on 1000 epochs. It was around 30% faster when I added some sleep time while saving the checkpoint. Do you have any recommendations?

src/pytorch_lightning/plugins/io/async_plugin.py
rohitgr7: yep, I realized, but then the alternate option I could think of is when we are closing the thread, but raising an error there is might not be good since training is almost done and the error might not have been raised at the end within the thread.

rohitgr7: do you think we should raise the error at both the places? or somewhere else?

src/pytorch_lightning/plugins/io/async_plugin.py
rohitgr7: > Also, a super nit edge case, but it's not handling different errors from multiple threads. I still strongly suggest to keep it at 1 and not expose num_threads to user.

well then it's just any error from any thread. Anyway, the training is going to be interrupted.

src/pytorch_lightning/plugins/io/async_plugin.py
carmocca: I asked if we should expose it. Why not expose it?

carmocca (approved):

src/pytorch_lightning/plugins/io/async_plugin.py
awaelchli: I need to see a reason why using multiple threads us useful or faster. I can't think of any reason and nobody seems know too. It wouldn't make much sense right? If you are IO bound, using more threads that write in parallel isn't going to make it faster. 

So when an argument can only take one value, then why  have the argument exposed.

src/pytorch_lightning/plugins/io/async_plugin.py
awaelchli: But please consider add the error raising to the teardown, and adding the comment I suggested.

awaelchli:
Great work @rohitgr7!!

rohitgr7:
Thanks @awaelchli @carmocca @otaj for great reviews and discussion.

PR #13664, opened 2022-07-14
test: accelerator="auto" [RFC]

Borda:
## What does this PR do?

maximize testing, so in env with GPU always run all tests with GPU
otherwise running tests only on CPU is wasting resources...

## Before submitting

## PR review

## Did you have fun?

cc @borda @akihironitta


tests/tests_pytorch/utilities/test_cli.py
```diff
@@ -852,7 +852,7 @@ def after_tune(self):
 
 def test_lightning_cli_subcommands():
     subcommands = LightningCLI.subcommands()
-    trainer = Trainer()
+    trainer = Trainer(accelerator="auto")
```
carmocca, line 855: The Trainer does not run in this test, so this has no purpose

```suggestion
    trainer = Trainer()
```

tests/tests_pytorch/trainer/test_states.py
```diff
@@ -20,7 +20,7 @@
 
 def test_initialize_state():
     """Tests that state is INITIALIZING after Trainer creation."""
-    trainer = Trainer()
+    trainer = Trainer(accelerator="auto")
```
carmocca, line 23: This has no effect

```suggestion
    trainer = Trainer()
```

carmocca (commented):
If the `Trainer`s are not set to `fit/validate/tune`.... we should not set this flag.

Borda:
> If the `Trainer`s are not set to `fit/validate/tune`.... we should not set this flag.

yes, agree, this would need some more cleaning, but wanted to iterate the main idea

Borda:
@carmocca manually went throw all cases when we fall fit in our tests and if an appropriate added auto option...
-> mind check now again :rabbit:

carmocca:
A problem with this change is that CI will not tell you whether these are passing under a GPU job since we don't have a job that would test them after #13651. You should completely run the test suite in our cluster manually to verify. Otherwise users will see failures when they are run locally with a GPU machine.

Borda:
> A problem with this change is that CI will not tell you whether these are passing under a GPU job since we don't have a job that would test them after #13651. You should completely run the test suite in our cluster manually to verify. Otherwise users will see failures when they are run locally with a GPU machine.

We can rethink CI in following uodate, but at least running locally on GPU you will get wider/genelized testing...

Borda:
> I don't understand yet what the goal is.

At this moment, when you run all tests most are running on CPU and ignoring GPU, which is a very marginal use of PL users, they want to use GPU, so why we are testing mostly CPU configuration and then just believe that all would work also with GPU?
And even if you want to test on GPU you can't without this rewrite, in the other hand with setting GPU visibility you can very simply force run all on CPU even if you use a GPU machine (but not vice versa)...

> Recently @carmocca made it so that only tests annotated with `RunIf(gpus>0)` would run on the GPU CI. So adding auto to any of the tests here would not make a difference. These seem to be two conflicting directions.

yes, they won't run on CI, which can be changed later...

> Also, in general we would have to go through each test and check what its intentions are. For example, in configuration tests where we assert fixed values, e.g. explicitly the device, the test could fail if we just blindly added "auto" and it ran on e.g. GPU, TPU or IPU etc.

yes, that probably would need to be checked but it will be covered with them running all tests on GPU

rohitgr7:
Just some thoughts!

1. I understand we need to test everything on each device to ensure it won't break, but we never did that, and recently GPU CI time was reduced significantly, but now we are adding more tests to make it even higher (maybe?).
2. Did we see any issues related to these tests where it worked on CPU but failed on GPU/TPU?
3. Many of them are related to the model checkpointing, progress bar updates, callbacks (some of them), loops, etc., which I don't think will have any effect based on the device used.

carmocca:
I suggest we close this because of the limitations described in the comments above and because this will get implicitly tested the day we start defaulting to `auto` in the Trainer. Happy to continue the discussion if you feel differently.

PR #13685, opened 2022-07-16
fixes typing in pytorch_lightning/callbacks/stochastic_weight_avg.py

donlapark:
## What does this PR do?

Fixes typing in pytorch in `pytorch_lightning/callbacks/stochastic_weight_avg.py` from #13445.

Note:
* Line 109: there has been a known issue of `isinstance(..., Callable)` throwing an arg-type error (https://github.com/python/mypy/issues/6864).
* Line 127: If `self._swa_epoch_start` is a `float`, it is expected to be converted to `int` by a call to `on_fit_start()`. However, `mypy` does not see this and throws in a return-value error.
* Line 175: `torch.optim.swa_utils.SWALR` has not been type-annotated yet, so the type of the argument of `swa_lr` is automatically set to `float`.
* Line 246-258: `_BatchNorm.running_mean`, `_BatchNorm.running_var` and `_BatchNorm.num_batches_tracked` have types `Optional[Tensor]`. These attributes are initialized to be zero tensors via `self.reset_parameters()` but `mypy` does not see this.

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -106,7 +106,7 @@ def __init__(
         if wrong_type or wrong_float or wrong_list:
             raise MisconfigurationException("The `swa_lrs` should a positive float, or a list of positive floats")
 
-        if avg_fn is not None and not isinstance(avg_fn, Callable):
+        if avg_fn is not None and not isinstance(avg_fn, Callable):  # type: ignore[arg-type]
```
otaj, line 109: How about using a `callable` check? https://docs.python.org/3/library/functions.html#callable

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -118,27 +118,28 @@ def __init__(
...
-        self._model_contains_batch_norm = None
-        self._average_model = None
+        self._max_epochs: int
+        self._model_contains_batch_norm: bool
+        self._average_model = pl.LightningModule()
 
     @property
     def swa_start(self) -> int:
-        return max(self._swa_epoch_start - 1, 0)  # 0-based
+        return max(self._swa_epoch_start - 1, 0)  # type: ignore[return-value]
```
otaj, line 127: Please do not remove comments

otaj: And by the time it is used, it will already have the type properly resolved (no more float, but int only)

donlapark: My bad. The comment has been added back.

Borda (approved):

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
akihironitta: > And by the time it is used, it will already have the type properly resolved (no more float, but int only)

@otaj I thought so, too, but mypy seems to miss the value range check in `__init__`.

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -118,27 +118,28 @@ def __init__(
         self._annealing_strategy = annealing_strategy
         self._avg_fn = avg_fn or self.avg_fn
         self._device = device
-        self._model_contains_batch_norm = None
-        self._average_model = None
+        self._max_epochs: int
+        self._model_contains_batch_norm: bool
+        self._average_model = pl.LightningModule()
```
akihironitta, line 123: Instead, I think we could do:
```suggestion
        self._average_model: pl.LightningModule
```

donlapark: Maybe like this, just to be on the safe side?

```suggestion
        self._average_model: "pl.LightningModule"
```

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
otaj: it doesn't "miss" anything, the variable is a `Union[int, float]` in `__init__`, it only gets forced to `int` in `on_fit_start` https://github.com/Lightning-AI/lightning/blob/26361adce107234a858de7bd3a163bef43832ffb/src/pytorch_lightning/callbacks/stochastic_weight_avg.py#L149.

But it is first used by `on_train_epoch_start`, so that should be resolved by then. But, mypy doesn't know (and cannot know) about this order

donlapark: So what is the "right" way to resolve this? A simple fix would be put `# type: ignore` there or `assert isinstance(self._swa_epoch_start, int)`.

otaj: I'd go with assert

otaj: And check that the assert doesn't break any tests of course :)

kaushikb11 (approved):

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -229,35 +234,39 @@ def on_train_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
...
-                module.running_mean, device=pl_module.device, dtype=module.running_mean.dtype
+                module.running_mean,  # type: ignore[arg-type]
+                device=pl_module.device,
+                dtype=module.running_mean.dtype,  # type: ignore[union-attr]
             )
             module.running_var = torch.ones_like(
-                module.running_var, device=pl_module.device, dtype=module.running_var.dtype
+                module.running_var,  # type: ignore[arg-type]
+                device=pl_module.device,
+                dtype=module.running_var.dtype,  # type: ignore[union-attr]
```
carmocca, line 255: IMO these shouldn't have been ignored, but instead added assertions that `running_{mean,var}` are not None

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -229,35 +234,39 @@ def on_train_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule"):
...
             )
             self.momenta[module] = module.momentum
-            module.momentum = None
-            module.num_batches_tracked *= 0
+            module.momentum = None  # type: ignore[assignment]
+            module.num_batches_tracked *= 0  # type: ignore[assignment, operator]
```
carmocca, line 259: Same comment here about asserting that it's not None

src/pytorch_lightning/callbacks/stochastic_weight_avg.py
```diff
@@ -167,12 +169,15 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
             for lr, group in zip(self._swa_lrs, optimizer.param_groups):
                 group["initial_lr"] = lr
 
-            self._swa_scheduler = SWALR(
-                optimizer,
-                swa_lr=self._swa_lrs,
-                anneal_epochs=self._annealing_epochs,
-                anneal_strategy=self._annealing_strategy,
-                last_epoch=trainer.max_epochs if self._annealing_strategy == "cos" else -1,
+            self._swa_scheduler: _LRScheduler = cast(
```
carmocca, line 172: The type annotation is redundant when you cast

```suggestion
            self._swa_scheduler = cast(
```

carmocca (commented):
You can open a follow-up PR with the suggestions applied

PR #13689, opened 2022-07-17
Fix mypy errors attributed to `pytorch_lightning.loggers.comet`

JustinGoheen:
## What does this PR do?

Fix mypy errors attributed to `pytorch_lightning.loggers.comet` for issue #13445

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

PR #13692, opened 2022-07-17
Fix mypy errors attributed to `pytorch_lightning.loggers.neptune`

JustinGoheen:
## What does this PR do?

Fix mypy errors attributed to `pytorch_lightning.loggers.neptune` for issue #13445

## Before submitting

## PR review


JustinGoheen:
blocked by https://github.com/Lightning-AI/lightning/pull/13689 for call to `_add_prefix` addressed in this [diff](https://github.com/Lightning-AI/lightning/pull/13689/files/28d5ec17c0fad18e878a2e3a3b548f152168100d#diff-a71f79d3349c77d9f589c4a098b211eae8102e787eee74bc5e8166953a4a468e) of linked PR

otaj:
> blocked by #13689 for call to `_add_prefix` addressed in this [diff](https://github.com/Lightning-AI/lightning/pull/13689/files/28d5ec17c0fad18e878a2e3a3b548f152168100d#diff-a71f79d3349c77d9f589c4a098b211eae8102e787eee74bc5e8166953a4a468e) of linked PR

Hi @JustinGoheen, #13689 is already merged.

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -393,17 +393,17 @@ def _verify_input_arguments(
...
 
-    def __getstate__(self):
+    def __getstate__(self) -> Dict:
         state = self.__dict__.copy()
         # Run instance can't be pickled
         state["_run_instance"] = None
         return state
 
-    def __setstate__(self, state):
+    def __setstate__(self, state: Dict) -> None:
```
awaelchli, line 402:
```suggestion
    def __setstate__(self, state: Dict[str, Any]) -> None:
```

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -393,17 +393,17 @@ def _verify_input_arguments(
                 " you can't provide other neptune.init() parameters.\n"
             )
 
-    def __getstate__(self):
+    def __getstate__(self) -> Dict:
```
awaelchli, line 396:
```suggestion
    def __getstate__(self) -> Dict[str, Any]:
```

pyproject.toml
```diff
@@ -58,6 +58,10 @@ module = [
     "pytorch_lightning.demos.boring_classes",
     "pytorch_lightning.demos.mnist_datamodule",
     "pytorch_lightning.loggers.neptune",
+    "pytorch_lightning.loggers.mlflow",
+    "pytorch_lightning.loggers.tensorboard",
+    "pytorch_lightning.loggers.wandb",
+    "pytorch_lightning.profilers.advanced",
```
awaelchli, line 64:
```suggestion
```

awaelchli (approved):

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -334,7 +335,7 @@ def _neptune_init_args(self):
 
         return args
 
-    def _construct_path_with_prefix(self, *keys) -> str:
+    def _construct_path_with_prefix(self, *keys: str) -> str:
```
JustinGoheen, line 338: each call to this in the call hierarchy used a string. sanity checked the unpacking works with a str.

```python
LOGGER_JOIN_CHAR = "/"

def _construct_path_with_prefix(prefix, *keys: str) -> str:
    if prefix:
        return LOGGER_JOIN_CHAR.join([prefix, *keys])
    return LOGGER_JOIN_CHAR.join(keys)

_construct_path_with_prefix(None, "zed/quu")
Out: 'zed/quu'

_construct_path_with_prefix("root", "zed/quu")
Out: 'root/zed/quu'
```

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -618,18 +619,18 @@ def _dict_paths(cls, d: dict, path_in_build: str = None) -> Generator:
     @property
     def name(self) -> str:
         """Return the experiment name or 'offline-name' when exp is run in offline mode."""
-        return self._run_name
+        return self._run_name  # type: ignore[return-value]
```
JustinGoheen, line 622: `_run_name` is annotated with Optional[None] in \_\_init_\_\, and is then set to a string in `._retrieve_run_data`

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -618,18 +619,18 @@ def _dict_paths(cls, d: dict, path_in_build: str = None) -> Generator:
...
 
     @property
     def version(self) -> str:
         """Return the experiment version.
 
         It's Neptune Run's short_id
         """
-        return self._run_short_id
+        return self._run_short_id  # type: ignore[return-value]
```
JustinGoheen, line 630: `_run_sort_id` is annotated with Optional[None] in \_\_init_\_\, and is then set to a string in `._retrieve_run_data`.  mypy is incorrectly evaluating `_run_sort_id` as Optional[str]

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -649,11 +650,11 @@ def _signal_deprecated_api_usage(f_name, sample_code, raise_exception=False):
             raise ValueError("The function you've used is deprecated.\n" + msg_suffix)
 
     @rank_zero_only
-    def log_metric(self, metric_name: str, metric_value: Union[Tensor, float, str], step: Optional[int] = None):
+    def log_metric(self, metric_name: str, metric_value: Union[Tensor, float, str], step: Optional[int] = None) -> None:
         key = f"{self._prefix}/{metric_name}"
         self._signal_deprecated_api_usage("log_metric", f"logger.run['{key}'].log(42)")
         if torch.is_tensor(metric_value):
-            metric_value = metric_value.cpu().detach()
+            metric_value = metric_value.cpu().detach()  # type: ignore[union-attr]
```
JustinGoheen, line 657: types `float` and `str` cause the following error despite `if torch.is_tensor(metric_value):` checking the type before executing `metric_value.cpu().detach()`

```sh
 Item ...  of "Union[Tensor, float, str]" has no attribute "cpu"
```

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -602,8 +603,8 @@ def _get_full_model_name(model_path: str, checkpoint_callback: "ReferenceType[Ch
     @classmethod
     def _get_full_model_names_from_exp_structure(cls, exp_structure: dict, namespace: str) -> Set[str]:
         """Returns all paths to properties which were already logged in `namespace`"""
-        structure_keys = namespace.split(cls.LOGGER_JOIN_CHAR)
-        uploaded_models_dict = reduce(lambda d, k: d[k], [exp_structure, *structure_keys])
+        structure_keys: list = namespace.split(cls.LOGGER_JOIN_CHAR)
```
awaelchli, line 606:
```suggestion
        structure_keys: List[str] = namespace.split(cls.LOGGER_JOIN_CHAR)
```

src/pytorch_lightning/loggers/neptune.py
awaelchli: There is a check `torch.is_tensor(metric_value)` above, which means we know that it can only be a tensor. MyPy doesn't know this, but we can tell it this by 
`assert isinstance(metric_value, Tensor)`

IMO this should be done instead of the type ignore

JustinGoheen: I think future contributors/maintainers might see the assert and wonder why it was added after the is_tensor check, then remove it; whereas ignoring the error handles for mypy checks only, and does not introduce new code just to appease a mypy error.

In general, it would probably be good to set a convention on when, why, and how to change code or add ignore[union-attr] given the amount of optionals & unions in the code base.

src/pytorch_lightning/loggers/neptune.py
JustinGoheen: when I make this change locally, the following error occurs:

```sh
src/pytorch_lightning/loggers/neptune.py:607: error: List item 1 has incompatible type "List[str]"; expected "Dict[Any, Any]"  [list-item]
```

Given the above, I updated ln 607 to 

```python
uploaded_models_dict: Dict[str, List[str]] = reduce(lambda d, k: d[k], [exp_structure, *structure_keys])
```

however, mypy then complained:

```sh
src/pytorch_lightning/loggers/neptune.py:607: error: Incompatible return value type (got "List[str]", expected "Dict[str, List[str]]")  [return-value]
src/pytorch_lightning/loggers/neptune.py:607: error: Invalid index type "Dict[str, List[str]]" for "Dict[str, List[str]]"; expected type "str"  [index]
src/pytorch_lightning/loggers/neptune.py:607: error: List item 1 has incompatible type "List[str]"; expected "Dict[str, List[str]]"  [list-item]
```

Likely because the lambda in reduce is using recursion to create the Dict `uploaded_models_dict`.

No errors occur when I leave the code as:

```python
def _get_full_model_names_from_exp_structure(cls, exp_structure: dict, namespace: str) -> Set[str]:
    """Returns all paths to properties which were already logged in `namespace`"""
    structure_keys: List = namespace.split(cls.LOGGER_JOIN_CHAR)
    uploaded_models_dict: Dict = reduce(lambda d, k: d[k], [exp_structure, *structure_keys])
    return set(cls._dict_paths(uploaded_models_dict))
```

awaelchli: The reduction there is not compatible with the typing, so I changed it to a for loop. I'm not sure if it's "better" but this is the only way I found to make the types at least somewhat specific. Note, to properly type the dict there we would need some sort of recursive type definition, and I think that's not possible at the moment.

src/pytorch_lightning/loggers/neptune.py
awaelchli: Actually you are right, we should change it to 

`if isinstance(metric_value, Tensor):`

this is even better. Then the assert is not needed and the ignore can be dropped.

JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13692/commits/4f049a97ba3d32ccb18b51320b85efecd67ad452

src/pytorch_lightning/loggers/neptune.py
otaj: This should already be `str` by the time it's used, shouldn't it? If so, I'm voting for `assert` instead of `# type: ignore`

src/pytorch_lightning/loggers/neptune.py
otaj: This should already be `str` by the time it's used, shouldn't it? If so, I'm voting for `assert` instead of `# type: ignore`

JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13692/commits/575f150fdb8cd7406d3a04f43a35be1fb1e63618

src/pytorch_lightning/loggers/neptune.py
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13692/commits/575f150fdb8cd7406d3a04f43a35be1fb1e63618

otaj: So, it seems this broke couple tests. Something fishy going on, I will debug it next week.

JustinGoheen: would it be more appropriate to use

```python
    @property
    def name(self) -> Optional[str]:
        """Return the experiment name or 'offline-name' when exp is run in offline mode."""
        if self._run_name is not None:
            return self._run_name

    @property
    def version(self) -> Optional[str]:
        """Return the experiment version.

        It's Neptune Run's short_id
        """
        if self._run_short_id is not None:
            return self._run_short_id
```

this passes mypy checks; I can at least push the change so that the other CI/CD checks can be compared across the two commits

JustinGoheen: implemented the above in https://github.com/Lightning-AI/lightning/pull/13692/commits/987337f647be9014f74c2efde6f8efde879b0de2

src/pytorch_lightning/loggers/neptune.py
otaj: I took the liberty to update the code a bit to invoke `self.run` in order to fetch `self._run_name` and `self._run_short_id` to make the tests happy again

otaj (approved):

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -618,6 +619,9 @@ def _dict_paths(cls, d: dict, path_in_build: str = None) -> Generator:
     @property
     def name(self) -> str:
         """Return the experiment name or 'offline-name' when exp is run in offline mode."""
+        if self._run_name is None:
+            _ = self.run
+        assert self._run_name is not None
```
awaelchli, line 624: @otaj 
The design here was intentional. As you can see in the base class, the name and version properties are Optional. This is because most loggers are not pickle friendly and we want to discourage accessing them in the main process before processes get spawned. This is why we create the experiments lazily, only when it is first needed. Hence, run name and run id would be None here because the experiment was not created yet.

Does it make sense?

otaj: The idea makes sense. If the rest of the codebase is ok with`logger.name` being `None`, then so am I :joy: Thanks for the explanation.
I'm positive I saw a similar thing done in some other logger though, I'll have to look it up.

In the meantime, I'm reverting my two commits

awaelchli (approved):
Looks great!

src/pytorch_lightning/loggers/neptune.py
awaelchli: Thanks. I know it's suboptimal but it is the only design that generalized well across all loggers for multiprocessing. So far. I definitely agree it would be nice if we could have no optionals there.

src/pytorch_lightning/loggers/neptune.py
```diff
@@ -290,26 +288,27 @@ def __init__(
...
-    def _retrieve_run_data(self):
+    def _retrieve_run_data(self) -> None:
         try:
-            self._run_instance.wait()
-            self._run_short_id = self._run_instance["sys/id"].fetch()
-            self._run_name = self._run_instance["sys/name"].fetch()
+            if self._run_instance is not None:
+                self._run_instance.wait()
+                self._run_short_id = self._run_instance["sys/id"].fetch()
+                self._run_name = self._run_instance["sys/name"].fetch()
```
rohitgr7, line 304:
```suggestion
            assert self._run_instance is not None
            self._run_instance.wait()
            self._run_short_id = self._run_instance["sys/id"].fetch()
            self._run_name = self._run_instance["sys/name"].fetch()
```
I'd not change the logic.

rohitgr7 (approved):

PR #13738, opened 2022-07-19
Replace unwrapping logic in strategies

### LINKED ISSUE #8593: Avoid rewrapping LightningModules in plugins

### LINKED ISSUE #13500: FullyShardedDataParallel wrapped models not being unwrapped, leading to incorrect checkpoints.

### LINKED ISSUE #13502: Remove unwrapping logic in strategies in favor of a direct reference to the original module
## Proposed refactor

I propose to remove the fragile unwrapping logic in strategies in favor of a keeping a direct reference to the original module.

### Motivation

We currently have an unwrapping logic in our strategies that looks something like this:

```py
def unwrap_lightning_module(wrapped_model: nn.Module) -> "pl.LightningModule":
    model = wrapped_model
    if isinstance(model, (DistributedDataParallel, DataParallel)):
        model = unwrap_lightning_module(model.module)
    if isinstance(model, (_LightningModuleWrapperBase, _LightningPrecisionModuleWrapperBase)):
        model = unwrap_lightning_module(model.module)
    ...
```

On the strategy, we only store a reference `self.model` to the wrapped model, and we additionally have a property

```py
@property
def lightning_module(self):
    return unwrap_lightning_module(self.model)
```

That can return the original LightningModule. 
The basic unwrapping logic has been around since PL had support for DP and DDP. It makes little sense to have a function like this, because the user always gives us the pure LightningModule as input to `Trainer.fit`. For some reason we kept it and kept extending it, leading to a copy-paste every time a new strategy gets introduced. It also raises a few questions for users who want to build their own strategy. Overall, it's an unintuitive solution that seems overly complicated when one could just keep the original reference saved in addition to the wrapped model.

In LightningLite, we recently refactored our `_LiteModel` wrapper to do exactly this, keeping both the original and wrapped nn Module in order to provide pass-through access to attributes (#12597).

### Pitch

1. Remove all `unwrap_lightning_module` functions
2. Modify Strategy.connect() to just save a reference to the given model.
3. Make Strategy.lightning_module directly return this reference instead of calling the unwrap method(s)

There are no breaking changes. The result should be equivalent, but simpler, easier to test and thus less fragile.

### Additional context

We wanted to do this for a while, came on my radar again after seeing a contributor struggle with it in #13501.

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

cc @justusschock @awaelchli @rohitgr7 @akihironitta

awaelchli:
## What does this PR do?

Fixes #13502 (main RFC)
Fixes #8593
Fixes #13500

The PR looks big, but most complexity turns out to be in the deprecation. The rest of the changes are very thinly spread across multiple files, but are not complex by themselves. 

Before, you would do

```py
lightning_module = unwrap_lightning_module(self.model)
```

Now, directly access it:

```py
lightning_module = self.lightning_module
```

Terminology:
`self.model`: refers to a LightningModule that is potentially wrapped
`self.lightning_module`: is always the pure LightningModule as it gets passed to Trainer.fit by the user. This attribute is read-access only.

TODOs:
- [x] backward compatibility in wrappers/deprecation

No.

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @rohitgr7 @tchaton @akihironitta


carmocca (approved):
Great! Looks like it was a complex deprecation

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -55,6 +62,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed MPS device being unrecognized ([#13992](https://github.com/Lightning-AI/lightning/pull/13992))
 
 
+
```
carmocca, line 65:
```suggestion
```

src/pytorch_lightning/strategies/deepspeed.py
```diff
@@ -71,9 +71,13 @@ def remove_module_hooks(model: torch.nn.Module) -> None:
...
 class LightningDeepSpeedModule(_LightningModuleWrapperBase):
     def __init__(
-        self, pl_module: Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase], precision: Union[str, int]
+        self,
+        forward_module: Optional[Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]] = None,
+        pl_module: Optional[Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]] = None,
+        precision: Union[str, int] = 32,
     ) -> None:
-        super().__init__(pl_module)
+        self._validate_init_arguments(pl_module, forward_module)
```
carmocca, line 79: Beware of https://github.com/Lightning-AI/lightning/pull/14000 as it also deprecates this class (cc @rohitgr7)

src/pytorch_lightning/overrides/fairscale.py
```diff
@@ -11,27 +11,41 @@
...
 # limitations under the License.
+from typing import Optional, Union
+
 import torch.nn as nn
 
 import pytorch_lightning as pl
-from pytorch_lightning.overrides.base import _LightningModuleWrapperBase, unwrap_lightning_module
-from pytorch_lightning.utilities import _IS_WINDOWS, _module_available
-
-_FAIRSCALE_AVAILABLE = not _IS_WINDOWS and _module_available("fairscale.nn")
```
carmocca, line 20: What was the issue with keeping this variable here? https://github.com/Lightning-AI/lightning/issues/11826

carmocca:
Do these changes also suggest we should do some refactor with the `_LiteModule`? Now that the APIs match.

src/pytorch_lightning/strategies/sharded_spawn.py
```diff
@@ -19,18 +19,20 @@
...
-from pytorch_lightning.overrides.fairscale import _FAIRSCALE_AVAILABLE
+from pytorch_lightning.overrides.base import _LightningModuleWrapperBase
 from pytorch_lightning.strategies.ddp_spawn import DDPSpawnStrategy
 from pytorch_lightning.trainer.states import TrainerFn
+from pytorch_lightning.utilities import _IS_WINDOWS, _module_available
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.optimizer import optimizers_to_device
 from pytorch_lightning.utilities.rank_zero import rank_zero_only
 
+_FAIRSCALE_AVAILABLE = not _IS_WINDOWS and _module_available("fairscale.nn")
```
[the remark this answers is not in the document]
awaelchli: updated, thanks

src/pytorch_lightning/strategies/deepspeed.py
awaelchli: That's fine, we can adjust depending on what lands first. It will create a merge conflict that will notify us :)

src/pytorch_lightning/overrides/fairscale.py
awaelchli: Oh, I guess I was trying to mitigate a small code duplication. But we will address it anyway later. Reverted.

awaelchli:
> Do these changes also suggest we should do some refactor with the _LiteModule? Now that the APIs match.

I think this is doesn't have anything to do with the LiteModule. At the moment I don't see more refactors in this direction. But open to suggestions.

src/pytorch_lightning/strategies/sharded_spawn.py
```diff
@@ -40,11 +40,19 @@ class DDPSpawnShardedStrategy(DDPSpawnStrategy):
 
     strategy_name = "ddp_sharded_spawn"
 
+    def connect(self, model: "pl.LightningModule") -> None:
+        if not _FAIRSCALE_AVAILABLE:  # pragma: no cover
+            raise MisconfigurationException(
+                "`DDPSpawnShardedStrategy` requires `fairscale` to be installed."
+                " Install it by running `pip install fairscale`."
+            )
```
rohitgr7, line 48: this should be in init

src/pytorch_lightning/strategies/sharded.py
```diff
@@ -44,6 +42,14 @@ class DDPShardedStrategy(DDPStrategy):
     strategy_name = "ddp_sharded"
     _REDUCE_BUFFER_SIZE_DEFAULT: int = 2**23  # 8M
 
+    def connect(self, model: "pl.LightningModule") -> None:
+        if not _FAIRSCALE_AVAILABLE:  # pragma: no cover
+            raise MisconfigurationException(
+                "`DDPShardedStrategy` requires `fairscale` to be installed."
+                " Install it by running `pip install fairscale`."
+            )
```
rohitgr7, line 50: same here

src/pytorch_lightning/overrides/base.py
```diff
@@ -85,27 +104,53 @@ def forward(self, *inputs: Any, **kwargs: Any) -> Any:
...
+    def _validate_init_arguments(
+        cls,
+        pl_module: Optional[Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]] = None,
+        forward_module: Optional[Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]] = None,
+    ) -> None:
+        # TODO: In v1.10, remove this method and mark the forward_module init argument in all subclasses as required
+        if pl_module is not None:
+            rank_zero_deprecation(
+                f"The argument `pl_module` in `{cls.__name__}` is deprecated in v1.8 and will be removed in"
+                " v1.10. Please use `forward_module` instead."
```
rohitgr7, line 124:
```suggestion
                f"The argument `pl_module` in `{cls.__name__}` is deprecated in v1.8.0 and will be removed in"
                " v1.10.0. Please use `forward_module` instead."
```

src/pytorch_lightning/overrides/base.py
```diff
@@ -54,29 +55,47 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
         super().__init__()
-        self.module = pl_module
+        if not isinstance(forward_module, pl.LightningModule) and (
+            not isinstance(getattr(forward_module, "module", None), pl.LightningModule)
+        ):
+            raise ValueError(
+                "`forward_module` must be a `LightningModule` instance or have an attribute `.module` pointing to one,"
+                f" got: {forward_module.__class__.__qualname__}"
+            )
+        # TODO: In v1.10, remove the Optional type from forward_module and remove the assertion
```
rohitgr7, line 78: can remove the optional now by making an assertion in the child classes.

src/pytorch_lightning/overrides/base.py
```diff
@@ -85,27 +104,53 @@ def forward(self, *inputs: Any, **kwargs: Any) -> Any:
...
+        `LightningModule` directly through the strategy attribute `Strategy.lightning_module`.
+
     Raises:
         TypeError: If the unwrapping leads to a module that is not a LightningModule and that cannot be unwrapped
             further.
     """
+    if not _suppress_warning:
+        rank_zero_deprecation(
+            "The function `unwrap_lightning_module` is deprecated in v1.8 and will be removed in v1.10. Access the"
+            " `LightningModule` directly through the strategy attribute `Strategy.lightning_module`."
```
rohitgr7, line 145:
```suggestion
        The function ``unwrap_lightning_module`` is deprecated in v1.8.0 and will be removed in v1.10.0. Access the
        ``LightningModule`` directly through the strategy attribute ``Strategy.lightning_module``.

    Raises:
        TypeError: If the unwrapping leads to a module that is not a LightningModule and that cannot be unwrapped
            further.
    """
    if not _suppress_warning:
        rank_zero_deprecation(
            "The function `unwrap_lightning_module` is deprecated in v1.8.0 and will be removed in v1.10.0. Access the"
            " `LightningModule` directly through the strategy attribute `Strategy.lightning_module`."
```

src/pytorch_lightning/overrides/data_parallel.py
```diff
@@ -52,11 +52,23 @@ class LightningParallelModule(_LightningModuleWrapperBase):
...
     Args:
-        pl_module: the model to wrap
+        pl_module: The module to wrap. See description for `forward_module`.
+
+            .. deprecated:: v1.8
+                The argument `pl_module` is deprecated in v1.8 and will be removed in v1.10. Please use
+                `forward_module` instead.
+
+        forward_module: The module to wrap. If it's not a LightningModule, it must have an attribute ``.module``
+            pointing to a LightningModule reference.
```
rohitgr7, line 62:
```suggestion
                The argument ``pl_module`` is deprecated in v1.8.0 and will be removed in v1.10.0. Please use
                ``forward_module`` instead.

        forward_module: The module to wrap. If it's not a ``LightningModule``, it must have an attribute ``.module``
            pointing to a ``LightningModule`` reference.
```

src/pytorch_lightning/overrides/fairscale.py
```diff
@@ -11,27 +11,44 @@
...
+            forward_module: Optional[Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]] = None,
+            pl_module: Optional[Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]] = None,
+        ) -> None:
+            self._validate_init_arguments(pl_module, forward_module)
+            super().__init__(forward_module=(pl_module or forward_module))
 
     def unwrap_lightning_module_sharded(wrapped_model: nn.Module) -> "pl.LightningModule":
+        rank_zero_deprecation(
+            "The function `unwrap_lightning_module_sharded` is deprecated in v1.8 and will be removed in v1.10. Access"
+            " the `LightningModule` directly through the strategy attribute `Strategy.lightning_module`."
```
rohitgr7, line 45:
```suggestion
            "The function `unwrap_lightning_module_sharded` is deprecated in v1.8.0 and will be removed in v1.10.0. Access"
            "the `LightningModule` directly through the strategy attribute `Strategy.lightning_module`."
```

rohitgr7 (approved):

src/pytorch_lightning/overrides/fairscale.py
akihironitta: I think the whitespace is needed :eyes:

akihironitta (commented):
Thank you for the description! It helped me review the change :) LGTM!

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -22,12 +22,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
...
 ### Deprecated
 
 - Deprecated `amp_level` from `Trainer` in favour of passing it explictly via precision plugin ([#13898](https://github.com/Lightning-AI/lightning/pull/13898))
 
 
--
+- Deprecated the `unwrap_lightning_module` and `unwrap_lightning_module_sharded` utility functions in favor of accessing the unwrapped `LightningModule` on the strategy directly ([#13738](https://github.com/Lightning-AI/lightning/pull/13738))
+
+
+- Deprecated the `pl_module` argument in `LightningParallelModule`, `LightningDistributedModule`, `LightningShardedDataParallel`, `LightningBaguaModule` and `LightningDeepSpeedModule` wrapper classes ([#13738](https://github.com/Lightning-AI/lightning/pull/13738))
```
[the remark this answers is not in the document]
awaelchli: Yep, but still need to handle a special deprecation for the input argument I changed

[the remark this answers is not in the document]
rohitgr7: then should it be done for `LightningIPUModule`?

src/pytorch_lightning/overrides/base.py
awaelchli: This is the least ugliest way I found. I can't express what you want in mypy. It is not smart enough to resolve an assertion like this:

```
        assert not (pl_module is None and forward_module is None)
        reveal_type(forward_module)
```

`src/pytorch_lightning/overrides/data_parallel.py:72: note: Revealed type is "Union[pytorch_lightning.core.module.LightningModule, pytorch_lightning.overrides.base._LightningPrecisionModuleWrapperBase, None]"
`

I don't think this is worth investing time. It will be very easy to resolve this in 1.10 by just removing the code.

src/pytorch_lightning/strategies/sharded_spawn.py
awaelchli: You can see below, it was originally in the property. I'm moving this to connect as this was the closest to the property in terms of accessing the module. I won't move it to init to keep the behavior as compatible as possible.

src/pytorch_lightning/overrides/fairscale.py
awaelchli: @rohitgr7 why this. It is inconsistent in how we have done it so far.

src/pytorch_lightning/overrides/base.py
rohitgr7: cool!

src/pytorch_lightning/overrides/fairscale.py
rohitgr7: ah! I see. at some places we just don't add `z` of `x.y.z`, but I try to add the complete version. I think we should start adding a complete version since it is more explicit and tell users that the deprecation happened in `v1.8.0` and not `v.1.8.1` since recently we deprecated `LightningDeepSpeedModule` in `v1.7.1`.

src/pytorch_lightning/strategies/sharded_spawn.py
rohitgr7: okay, but it should be in init.. let's ignore this for now! will do a follow-up to move it.

rohitgr7 (approved):
If you could add comlpete version, it'd be nice.

else looks good.

src/pytorch_lightning/overrides/fairscale.py
awaelchli: That is great. I do not mind either way and always value consistency over a specific choice. It is a lifestyle I guess. Or a byproduct of taking too many math classes at uni 🤣 

I updated the versions, hopefully everywhere.

tchaton (approved):
LGTM !

PR #13782, opened 2022-07-21
Fix mypy errors attributed to `pytorch_lightning.callbacks.quantization`

krishnakalyan3:
Part of #13445


src/pytorch_lightning/callbacks/quantization.py
```diff
@@ -41,14 +41,14 @@
 
 
 def wrap_qat_forward_context(
-    quant_cb, model: "pl.LightningModule", func: Callable, trigger_condition: Optional[Union[Callable, int]] = None
+    quant_cb: Any, model: "pl.LightningModule", func: Callable, trigger_condition: Optional[Union[Callable, int]] = None
```
otaj, line 44: the type is not Any, but `QuantizationAwareTraining`

src/pytorch_lightning/callbacks/quantization.py
```diff
@@ -41,14 +41,17 @@
 
 
 def wrap_qat_forward_context(
-    quant_cb, model: "pl.LightningModule", func: Callable, trigger_condition: Optional[Union[Callable, int]] = None
+    quant_cb: Callback,
```
otaj, line 44: I don't think this can be any `Callback`

krishnakalyan3:
Thank you for all the reviews @otaj and for helping with this PR.

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -245,7 +245,7 @@ def _restore_quantization_callbacks(self) -> None:
             if state:
                 # The Quantization callbacks have a special method that must be called before restoring the weights
                 # of the model
-                callback._load_before_model(self.trainer.model, deepcopy(state))
+                callback._load_before_model(self.trainer.lightning_module, deepcopy(state))
```
otaj, line 248: Can someone with a bit more knowledge around quantization take a look whether it's correct? I suggested it because it seemed to make the most sense (and we were expecting `LightningModule` along the way anyway, but I am not 100% sure. cc @carmocca @awaelchli

otaj (approved):

src/pytorch_lightning/callbacks/quantization.py
```diff
@@ -181,7 +184,9 @@ def __init__(
             )
         self._observer_type = observer_type
 
-        if collect_quantization is not None and not isinstance(collect_quantization, (int, Callable)):
+        if collect_quantization is not None and not (
+            isinstance(collect_quantization, int) and callable(collect_quantization)
+        ):
```
otaj, line 189: Uh oh, @krishnakalyan3, I just noticed several failing tests, that are linked to the changes made here. Would you mind taking a look at them?

```suggestion
        if collect_quantization is not None and not (
            isinstance(collect_quantization, int) or callable(collect_quantization)
        ):
```

otaj: `isinstance` check against a tuple of types is equivalent to `or`, not `and`

krishnakalyan3: checking...

otaj (approved):

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
awaelchli: Yes it is correct, because _restore_quantization_callbacks gets called before the model gets wrapped in the strategy, so at this point in time, we have: `self.trainer.lightning_module == self.trainer.model`.

awaelchli (approved):

krishnakalyan3:
Thank you for all the feedback @otaj and @awaelchli

rohitgr7 (approved):

carmocca (approved):

krishnakalyan3:
Thank you for the reviews @rohitgr7 @carmocca @awaelchli @otaj

PR #13801, opened 2022-07-22
Docs/memory

oojo12:
## What does this PR do?

This PR updates the doc to note the 20GB disk size limit and specifies in the Cloud Compute table that Memory = RAM and memory in GPU column = GPU memory.

### DIFF
--- a/docs/source-app/core_api/lightning_work/index.rst
+++ b/docs/source-app/core_api/lightning_work/index.rst
@@ -2,7 +2,8 @@
 Lightning Work
 ##############
 
-**Audience:** Users who want to know how Lightning Work works under the hood 🤯.
+| **Audience:** Users who want to know how Lightning Work works under the hood 🤯.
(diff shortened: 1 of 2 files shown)


docs/source-app/core_api/lightning_work/index.rst
```diff
@@ -2,7 +2,8 @@
 Lightning Work
 ##############
 
-**Audience:** Users who want to know how Lightning Work works under the hood 🤯.
+| **Audience:** Users who want to know how Lightning Work works under the hood 🤯.
+| **Limitation:** At this time there is a 20GB disk size limit for Lightning Works.
```
awaelchli, line 6: There must be a typo. 20 GB would be too small.

awaelchli (approved):

docs/source-app/core_api/lightning_work/index.rst
oojo12: @hhsecond can you confirm or deny this?

PR #13806, opened 2022-07-23
Fix mypy errors attributed to `pytorch_lightning.trainer.connectors.data_connector.py`

JustinGoheen:
## What does this PR do?

Fix mypy errors attributed to `pytorch_lightning.trainer.connectors.data_connector.py` for issue #13445 

## Before submitting

## PR review


src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -453,14 +457,16 @@ def resolve_has_no_sequential_sampler(dataloader: DataLoader):
             )
 
             def replace_sampler(dataloader: DataLoader) -> DataLoader:
-                return _update_dataloader(dataloader, sampler=SequentialSampler(dataloader.dataset), mode=mode)
+                return _update_dataloader(
+                    dataloader, sampler=SequentialSampler(dataloader.dataset), mode=mode
+                )  # type: ignore[arg-type]
```
JustinGoheen, line 462: SequentialSampler is a PyTorch core method that requires arg 1 to be of type `Sized` however, `dataloader.dataset` is torch.utils.data.dataset, and is a `Generic[T_co]` type.

This appears to be an issue with PyTorch core's type hints.

src/pytorch_lightning/core/module.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._use_amp: bool = False
 
         # the precision used
-        self.precision: int = 32
+        self.precision: Union[int, str] = 32
```
JustinGoheen, line 102: added to resolve

```sh
src/pytorch_lightning/trainer/connectors/data_connector.py:148: error: Incompatible types in assignment (expression has type "Union[str, int]", variable has type "int")  [assignment]
```

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -347,7 +352,7 @@ def _reset_eval_dataloader(
             dataloaders = self._resolve_overfit_batches(dataloaders, mode)
 
         if not isinstance(dataloaders, list):
-            dataloaders = [dataloaders]
+            dataloaders = [dataloaders]  # type: ignore[list-item]
```
JustinGoheen, line 355: associated error ignored:

```sh
src/pytorch_lightning/trainer/connectors/data_connector.py:355: error: List item 0 has incompatible type "Union[DataLoader[Any], Collection[DataLoader[Any]]]"; expected "DataLoader[Any]"  [list-item]
```

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -14,7 +14,7 @@
 import multiprocessing
 import os
 from dataclasses import dataclass, field
-from typing import Any, Callable, Collection, List, Optional, Tuple, Union
+from typing import Any, Callable, Collection, Iterable, List, Optional, Tuple, TypeAlias, Union
```
otaj, line 17: Hi @JustinGoheen, all of the tests are now failing, as none of them are running on 3.10, which is the version when `TypeAlias` became available. https://docs.python.org/3/library/typing.html#typing.TypeAlias In order to fix it, please import `TypeAlias` from `typing_extensions`

JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13806/commits/87cb7a66d3ed03cdca5ed8ae60d211a7049e329e

otaj: Well, another error which made every test broke popped up -> `TypeAlias` is not subscriptable

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -141,7 +145,7 @@ def attach_data(
         # set local properties on the model
         self._copy_trainer_model_properties(model)
 
-    def _copy_trainer_model_properties(self, model: "pl.LightningModule") -> None:
+    def _copy_trainer_model_properties(self, model: TypeAlias["pl.LightningModule"]) -> None:
```
otaj, line 148: I don't think `TypeAlias` is appropriate - what are you trying to solve here?

src/pytorch_lightning/trainer/connectors/data_connector.py
otaj: This is not a proper way, how to resolve this. I'd rather see it resolved in methods generating this value (i.e. `self._resolve_overfit_batches` and `self._request_dataloader`)

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -489,7 +497,7 @@ class _DataLoaderSource:
     instance: Optional[Union[TRAIN_DATALOADERS, EVAL_DATALOADERS, "pl.LightningModule", "pl.LightningDataModule"]]
     name: str
 
-    def dataloader(self) -> Union[TRAIN_DATALOADERS, EVAL_DATALOADERS]:
+    def dataloader(self) -> Union[TypeAlias["instance"], TRAIN_DATALOADERS, EVAL_DATALOADERS]:
```
otaj, line 500: what is `TypeAlias["instance"]` supposed to be?

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -539,9 +547,9 @@ class _DataHookSelector:
         datamodule: A LightningDataModule
     """
 
-    model: "pl.LightningModule"
+    model: Optional["pl.LightningModule"]
     datamodule: Optional["pl.LightningDataModule"]
-    _valid_hooks: Tuple[str] = field(
+    _valid_hooks: Tuple[str, str, str] = field(
```
otaj, line 552: I don't think that limiting `_valid_hooks` to length 3 is a proper approach. It makes sense to make it variable length (https://docs.python.org/3/library/typing.html#typing.Tuple)

otaj (changes requested):

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: not able to find the original error in CI/CD checks; I removed locally and it is fine. Updated in https://github.com/Lightning-AI/lightning/pull/13806/commits/82ee684a0b76486eee08b0b9f7462762edf3ae51

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: without TypeAlias I get the following error:

```sh
src/pytorch_lightning/trainer/connectors/data_connector.py:500: error: Variable "pytorch_lightning.trainer.connectors.data_connector._DataLoaderSource.instance" is not valid as a type  [valid-type]
src/pytorch_lightning/trainer/connectors/data_connector.py:500: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
```

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: updated in https://github.com/Lightning-AI/lightning/pull/13806/commits/6020d0b3ec9afa56ee34757adea889163eac16bc

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: `_resolve_overfit_batches` is [used in trainer](https://github.com/Lightning-AI/lightning/blob/4e9fbdbe4013458ea33256f4e0b00cf0782e6064/src/pytorch_lightning/trainer/trainer.py#L1813) in a way that does not require a list be returned; updating `_resolve_overfit_batches` to fix the error here, would require that I alter code in trainer.

I tried `dataloaders = list(dataloaders)` out of curiosity, and that worked locally.

Updated in https://github.com/Lightning-AI/lightning/pull/13603/commits/ca4f1609731b64201c3fe184fe71a104fd15276d

src/pytorch_lightning/trainer/connectors/data_connector.py
otaj: The issue here seems that both of the types (`self.instance` as well as `self.dataloader()`) are valid, but mypy has a bit of a trouble here, since it thinks that it could happen, that `self.instance` is `pl.LightningModule` or `pl.LightningDataModule` and we return that value. Which is not true, as evidenced by the method, but mypy just doesn't understand it. In this case, I don't think there's an easy way to fix this and I'd go with `# type: ignore`

Any other opinions, @awaelchli @carmocca?

src/pytorch_lightning/trainer/connectors/data_connector.py
otaj: But `list(dataloaders) != [dataloaders]`. `list(iterable)` tries to convert that iterable into list (which would succeed for single dataloader, but is definitely not something we want here), whereas `[iterable]` creates a list with `iterable` as a first element. No wonder mypy stopped complaining, other tests had to start complaining.

otaj: Anyway, after taking a look that changing the methods is probably not feasible within this PR, I think in this case it makes sense to again, `# type: ignore`

otaj: Sorry about this

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: I created a constant `INSTANCE = Optional[Union[TRAIN_DATALOADERS, EVAL_DATALOADERS, "pl.LightningModule", "pl.LightningDataModule"]]` and updated the return to `Union[INSTANCE, TRAIN_DATALOADERS, EVAL_DATALOADERS]`; this works locally, updated in https://github.com/Lightning-AI/lightning/pull/13806/commits/1aebec4fe5f7cb291784e2164367c60a578b8f79

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: reverted in https://github.com/Lightning-AI/lightning/pull/13806/commits/c4039b1bc929aeb96049cfbd85e3b032f2166f14

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: to recreate the original error, I removed the return type in https://github.com/Lightning-AI/lightning/pull/13806/commits/b0cb0409a05469990f4805b953ed937f06707bd3

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: TypeAlias has been removed

src/pytorch_lightning/strategies/deepspeed.py
```diff
@@ -674,7 +674,7 @@ def _auto_select_batch_size(self) -> int:
             try:
                 train_dataloader = train_dl_source.dataloader()
                 if hasattr(train_dataloader, "batch_sampler"):
-                    batch_size = train_dataloader.batch_sampler.batch_size  # type: ignore[union-attr]
+                    batch_size = train_dataloader.batch_sampler.batch_size  # type: ignore[union-attr, assignment]
```
JustinGoheen, line 677: Once I added Instance as a return type in `data_connector._DataLoaderSource.dataloader`, this error was raised by mypy.

```sh
src/pytorch_lightning/strategies/deepspeed.py:677: error: Incompatible types in assignment (expression has type "Union[int, Any, Tensor, Module]", variable has type "int")  [assignment]
```

I traced back to `data_connector._DataLoaderSource.dataloader`, and there is no batch_sampler or batch_size in that specific method, or in the entire module.

`batch_sampler` and `batch_size` are attributes of torch.utils.data.DataLoader; given this is outside of the scope of PL, I opted to ignore the error.

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: reverted in https://github.com/Lightning-AI/lightning/pull/13806/commits/dfca1fe56087253e8fc3fb2bab52b7bdcce72377

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -486,10 +494,10 @@ class _DataLoaderSource:
             that returns the desired dataloader(s).
     """
 
-    instance: Optional[Union[TRAIN_DATALOADERS, EVAL_DATALOADERS, "pl.LightningModule", "pl.LightningDataModule"]]
+    instance: INSTANCE
     name: str
 
-    def dataloader(self) -> Union[TRAIN_DATALOADERS, EVAL_DATALOADERS]:
+    def dataloader(self) -> Union[INSTANCE, TRAIN_DATALOADERS, EVAL_DATALOADERS]:
```
otaj, line 500: The original types were correct, this method cannot return `LightningModule` or `LightningDataModule`. Either resolve it in the method (which I'm not really sure how feasible it is, since that method does a bit of magic, that `mypy` might not like) or add proper `# type: ignore`s, but don't extend the result type of this method

JustinGoheen: I opted to ignore after attempting to use self.is_defined to block mypy from evaluating return-value as a LightningModule.

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -45,6 +45,10 @@
 
 warning_cache = WarningCache()
 
+BATCH_DATALOADER = Union[Collection[DataLoader], List[DataLoader], DataLoader[Any]]
+REQUEST_DATALOADER = Union[DataLoader, List[DataLoader], BATCH_DATALOADER]
+INSTANCE = Optional[Union[TRAIN_DATALOADERS, EVAL_DATALOADERS, "pl.LightningModule", "pl.LightningDataModule"]]
```
otaj, line 50: Last nit: I'd rename this to something more descriptive

otaj (approved):

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: it's only used once now; so I just moved the annotation back to the appropriate class attr in https://github.com/Lightning-AI/lightning/pull/13806/commits/c2431aeb555bace222bbe470d8a60c381458f125

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -158,7 +161,7 @@ def attach_dataloaders(
         self.trainer.train_dataloader = None
         self.trainer.val_dataloaders = None
         self.trainer.test_dataloaders = None
-        self.trainer.predict_dataloaders = None
+        self.trainer.predict_dataloaders = None  # type: ignore[assignment]
```
rohitgr7, line 164: I think adding an entry here:
https://github.com/Lightning-AI/lightning/blob/ee4233b36cb70b19a13013371ada50875b429ef3/src/pytorch_lightning/trainer/trainer.py#L634
```py
self.predict_dataloaders = None
```

will help avoid `ignore[assignment]` here.

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -2240,7 +2240,7 @@ def is_global_zero(self) -> bool:
         return self.strategy.is_global_zero
 
     @property
-    def distributed_sampler_kwargs(self) -> Optional[dict]:
+    def distributed_sampler_kwargs(self) -> Mapping:
```
rohitgr7, line 2243:
```suggestion
    def distributed_sampler_kwargs(self) -> Optional[Dict[str, int]]:
```

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -497,7 +504,7 @@ def dataloader(self) -> Union[TRAIN_DATALOADERS, EVAL_DATALOADERS]:
         from pytorch_lightning import LightningDataModule, LightningModule  # prevent cyclic import
 
         if not self.name:
-            return self.instance
+            return self.instance  # type: ignore[return-value]
```
rohitgr7, line 507: do you know why we need to ignore here?

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -416,7 +421,7 @@ def _reset_eval_dataloader(
 
         return loader_num_batches, dataloaders
 
-    def _request_dataloader(self, stage: RunningStage) -> Union[DataLoader, List[DataLoader]]:
+    def _request_dataloader(self, stage: RunningStage) -> REQUEST_DATALOADER:
```
rohitgr7, line 424:
```suggestion
    def _request_dataloader(self, stage: RunningStage) -> TRAIN_DATALOADERS:
```
from `utilities/types.py`

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -435,10 +440,10 @@ def _request_dataloader(self, stage: RunningStage) -> Union[DataLoader, List[Dat
         return dataloader
 
     @staticmethod
-    def _resolve_overfit_batches(dataloaders: Collection[DataLoader], mode: RunningStage) -> Collection[DataLoader]:
+    def _resolve_overfit_batches(dataloaders: BATCH_DATALOADER, mode: RunningStage) -> BATCH_DATALOADER:
```
rohitgr7, line 443:
```suggestion
    def _resolve_overfit_batches(dataloaders: EVAL_DATALOADERS, mode: RunningStage) -> EVAL_DATALOADERS:
```
from `utilities/types.py`

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -45,6 +45,9 @@
 
 warning_cache = WarningCache()
 
+BATCH_DATALOADER = Union[Collection[DataLoader], List[DataLoader], DataLoader[Any]]
+REQUEST_DATALOADER = Union[DataLoader, List[DataLoader], BATCH_DATALOADER]
+
```
rohitgr7, line 50:
```suggestion
```

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: this suggestion causes several errors. the new type was made to address specific errors encountered.

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: https://github.com/Lightning-AI/lightning/pull/13806#discussion_r935491977

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: combining these suggestions with the other suggestions also causes additional errors

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: these are needed to resolve errors. each was created based on mypy error output.

src/pytorch_lightning/trainer/connectors/data_connector.py
JustinGoheen: doing this causes new errors 

```sh
src/pytorch_lightning/loops/dataloader/prediction_loop.py:63: error: Incompatible return value type (got "None", expected "Sequence[DataLoader[Any]]")  [return-value]
```

above error can be resolved with

```python
def dataloaders(self) -> Sequence[DataLoader]:
    """Returns all prediction dataloaders."""
    if self.trainer.predict_dataloaders:
        return self.trainer.predict_dataloaders
```

is it appropriate to annotate in Trainer, and update logic in a third file in order to remove an ignore?

src/pytorch_lightning/trainer/trainer.py
JustinGoheen: this causes

```sh
src/pytorch_lightning/trainer/connectors/data_connector.py:296: error: Argument after ** must be a mapping, not "Optional[Dict[str, int]]"  [arg-type]
```

src/pytorch_lightning/trainer/connectors/data_connector.py
rohitgr7: added a commit. removed 4 `ignores`

rohitgr7 (approved):

rohitgr7:
messed up somewhere :p
will check tomorrow!

JustinGoheen:
apologies, I'm no longer able to keep this PR up to date.

otaj:
@JustinGoheen, we're sorry to hear that. @rohitgr7, are you gonna push it over the finish line or should I? I noticed you started pushing some commits :zap:

rohitgr7:
@otaj it's done. Any things else needs to be done here? The CI is blocking the merge for now

otaj:
@rohitgr7, awesome, thank you! :tada:. I guess we just have to wait on codeowners and CI

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -498,18 +503,14 @@ def dataloader(self) -> Union[TRAIN_DATALOADERS, EVAL_DATALOADERS]:
 
         If the source is a module, the method with the corresponding :attr:`name` gets called.
         """
-        from pytorch_lightning import LightningDataModule, LightningModule  # prevent cyclic import
-
-        if not self.name:
-            return self.instance
```
carmocca, line 504: Is this removed on purpose?

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -439,10 +442,10 @@ def _request_dataloader(self, stage: RunningStage) -> Union[DataLoader, List[Dat
         return dataloader
 
     @staticmethod
-    def _resolve_overfit_batches(dataloaders: Collection[DataLoader], mode: RunningStage) -> Collection[DataLoader]:
+    def _resolve_overfit_batches(dataloaders: EVAL_DATALOADERS, mode: RunningStage) -> EVAL_DATALOADERS:
```
carmocca, line 445: Isn't this also used with training dataloaders?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -2231,7 +2231,7 @@ def is_global_zero(self) -> bool:
...
```
carmocca, line 2234: This would still be Optional

src/pytorch_lightning/trainer/connectors/data_connector.py
rohitgr7: this is covered by the default return at the end if `self.instance` is neither LM nor LDM.

src/pytorch_lightning/trainer/connectors/data_connector.py
rohitgr7: good catch! let me update

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -286,14 +286,19 @@ def _prepare_dataloader(
...
         return dataloader
 
-    def _resolve_sampler(self, dataloader: DataLoader, shuffle: bool, mode: Optional[RunningStage] = None) -> Sampler:
+    def _resolve_sampler(
+        self, dataloader: DataLoader, shuffle: bool, mode: Optional[RunningStage] = None
+    ) -> Union[Sampler, Iterable]:
         if self._requires_distributed_sampler(dataloader):
+            distributed_sampler_kwargs = self.trainer.distributed_sampler_kwargs
+            assert distributed_sampler_kwargs is not None
+            assert self.trainer.distributed_sampler_kwargs is not None
```
awaelchli, line 295:
```suggestion
```

awaelchli (approved):

src/pytorch_lightning/trainer/connectors/data_connector.py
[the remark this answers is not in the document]
otaj: I agree, but I can't find them anymore :joy: Marking as resolved so we can merge this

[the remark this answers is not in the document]
rohitgr7: looks like an old review before new updates

justusschock (approved):

carmocca (approved):

Borda (approved):

PR #11098, opened 2021-12-16
Fix TPU testing and collect all tests

awaelchli:
## What does this PR do?

Fixes #13720

Addresses a range of issues with the TPU CI:

- Now collects all tests that are marked with `RunIf(tpu=True)`. No longer do the tests get hardcoded and forgotten, leading to tests being added that never run.
- Removed the `@pl_multi_process` decorator on all tests. This decorator suppressed exceptions and assertion errors. Some tests were broken and outdated for a while and were never raising the errors due to this.
- Added standalone marker `RunIf(tpu=True, standalone=True)` for TPU tests as an alternative to the aforementioned `pl_multi_process`. The CI now runs standalone tests similar to the GPU test suite. This is necessary for example when we need to run with the single device tpu strategy that requires us to access the xla device in the main process. 
- Fixed tests that were affected by the aforementioned cause for failures. 

Note, after resolving the core issues, and pushing many commits to run the CI, I have not seen any flakiness and random behavior anymore. 

In a follow-up, we should update the CI to the latest torch_xla version.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

Part of #1 (it's a lie, this is just here to avoid noisy GitHub bot)

[...]

PR #13813, opened 2022-07-23
Fix mypy typing errors in pytorch_lightning/strategies/tpu_spawn.py

BongYang:
## What does this PR do?
Fix typing errors in pytorch_lightning/strategies/tpu_spawn.py in issue https://github.com/Lightning-AI/lightning/issues/13445. 

None

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -114,30 +114,32 @@ def _validate_patched_dataloaders(model: "pl.LightningModule") -> None:
             if not source.is_module():
                 TPUSpawnStrategy._validate_dataloader(source.instance)
```
BongYang, line 115: I have no idea to decide it.
source.instance can be `TRAIN_DATALOADERS = Union[
    DataLoader,
    Sequence[DataLoader],
    Sequence[Sequence[DataLoader]],
    Sequence[Dict[str, DataLoader]],
    Dict[str, DataLoader],
    Dict[str, Dict[str, DataLoader]],
    Dict[str, Sequence[DataLoader]]
]`
but _validate_dataloader can't support dictionary.

src/pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -147,7 +149,7 @@ def distributed_sampler_kwargs(self) -> Dict[str, int]:
     @property
     def is_distributed(self) -> bool:
         # HOST_WORLD_SIZE is None outside the xmp.spawn process
-        return os.getenv(xenv.HOST_WORLD_SIZE, None) and self.world_size != 1
+        return os.getenv(xenv.HOST_WORLD_SIZE, None) is not None and self.world_size != 1
```
awaelchli, line 152: Great! I also found this bug in #11098. A bit shorter and easier to read:
```suggestion
        return (xenv.HOST_WORLD_SIZE in os.environ) and self.world_size != 1
```

src/pytorch_lightning/strategies/tpu_spawn.py
awaelchli: @BongYang I think in this case we want to raise an error here if it is a dict. And then we can open an issue if we would like to support it in the future. What do you think?

Borda (approved):

otaj:
For some reason, `mypy` seems to have some trouble with `self._checkpoint_io` (line 79), which should not be happening. I would honestly consider it a mypy bug. However, there are also other failing tests

src/pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -196,20 +200,23 @@ def reduce(self, output, group: Optional[Any] = None, reduce_op: Optional[Union[
...
 
-    def _worker_setup(self, process_idx: int):
+    def _worker_setup(self, process_idx: int) -> None:
         reset_seed()
         self.set_world_ranks(process_idx)
         rank_zero_only.rank = self.global_rank
 
-    def validation_step(self, *args, **kwargs) -> Optional[STEP_OUTPUT]:
+    def validation_step(self, *args: Any, **kwargs: Any) -> Optional[STEP_OUTPUT]:
+        assert self.model
```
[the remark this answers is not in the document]
BongYang: done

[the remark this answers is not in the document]
awaelchli: @BongYang sorry, I posted this comment on the wrong PR. For the tpu strategy this does not apply. I reverted it. Thank you!

src/pytorch_lightning/strategies/tpu_spawn.py
awaelchli: I pushed a commit with a proposed change to make the validation match the type of the dataloaders well.

awaelchli (approved):

src/pytorch_lightning/strategies/tpu_spawn.py
BongYang: thank you!

src/pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -169,11 +174,11 @@ def configure_ddp(self) -> None:
     def model_to_device(self) -> None:
         self.model = self.wrapped_model.to(self.root_device)
 
-    def barrier(self, name: Optional[str] = None) -> None:
+    def barrier(self, name: Optional[str] = None, *args: Any, **kwargs: Any) -> None:
         if self.is_distributed:
             rendezvous(name)
 
-    def broadcast(self, obj: object, src: int = 0) -> object:
+    def broadcast(self, obj: object, src: int = 0) -> Any:
```
carmocca, line 181: Shouldn't `obj` and the return type match?

```suggestion
    def broadcast(self, obj: Any, src: int = 0) -> Any:
```

carmocca (commented):
The mypy job is failing

src/pytorch_lightning/strategies/tpu_spawn.py
awaelchli: Yes, we should change it to TBroadcast like in the other strategies

src/pytorch_lightning/strategies/tpu_spawn.py
```diff
@@ -96,17 +97,16 @@ def root_device(self) -> torch.device:
...
-        for dataloader in dataloaders:
+    def _validate_dataloader(dataloaders: Union[TRAIN_DATALOADERS, EVAL_DATALOADERS]) -> None:
+        def check_has_len(dataloader: DataLoader) -> None:
             if not has_len(dataloader):
                 raise MisconfigurationException(
                     "TPUs do not currently support IterableDataset objects, the dataset must implement `__len__`."
                     " HINT: You can mock the length on your dataset to bypass this MisconfigurationException."
                 )
 
+        apply_to_collection(dataloaders, dtype=object, wrong_dtype=(Sequence, Mapping), function=check_has_len)
```
rohitgr7, line 108: can you explain this change?

rohitgr7 (approved):

src/pytorch_lightning/strategies/tpu_spawn.py
awaelchli: Yes, I did this change to help out resolve a typing issue here. 
The entry point is  `TPUSpawnStrategy._validate_dataloader(source.instance)` we call below, and it takes `source.instance` which is a Union of a variety of collections of dataloaders. Not all these types were handled by the implementation of the validate_dataloader method previously, so I argue it's reasonable to udpate the implementation to cover all types (Sequence, Dict, ...). 

For this I used `apply_to_collection` to handle all collections. It may look strange that we use `wrong_type` here and `dtype=object`, but the intuitive solution of 
`apply_to_collection(dataloaders, dtype=DataLoader, function=check_has_len)` wouldn't work because the input can be dataloaders=DataLoader and that is not a collection.
Thus we invert the conditions.

Btw, there are tests that assert this error message is raised. We could extend the test to assert this for all supported collections.

otaj:
I took the liberty of solving the mypy errors regarding `self._checkpoint_io`

src/pytorch_lightning/strategies/tpu_spawn.py
BongYang: I updated it to TBroadcast. https://github.com/Lightning-AI/lightning/pull/13813/commits/ffbd23cf96abcaa910501217e48d52b53893536d

otaj (approved):

PR #13815, opened 2022-07-24
add lightning apps to info_packages of collect_env_details

luca-medeiros:
## What does this PR do?
Added `lightning` and `lightning_app` packages to [requirements/collect_env_details.py](https://github.com/Lightning-AI/lightning/blob/master/requirements/collect_env_details.py)

I added the apps with try exception because seems that they are not required/installed together with PyTorch Lightning. But the team could check that up. Either put all of them in a try/except or remove all and raise error when not installed.

Fixes https://github.com/Lightning-AI/lightning/issues/13747

None

## Before submitting

## PR review

## Did you have fun?

cc @awaelchli


awaelchli (approved):

Borda (approved):

Borda:
I believe this is not really needed for the Ap release...

carmocca (approved):

requirements/collect_env_details.py
```diff
@@ -56,6 +65,8 @@ def info_packages():
         "pyTorch_version": torch.__version__,
         "pyTorch_debug": torch.version.debug,
         "pytorch-lightning": pytorch_lightning.__version__,
+        "lightning": lightning.__version__ if "lightning" in sys.modules else None,
```
carmocca, line 68: Can be a follow-up, but I think this should be refactored to include all installed packages, with a solution similar to https://stackoverflow.com/a/69657737

manskx (approved):

requirements/collect_env_details.py
carmocca: Resolved this to merge. But feel free to do this contribution too @luca-medeiros!

luca-medeiros: That looks like a clean way to gather installed packages!

However, I'm not sure if I got what you meant completely.
Are we trying to show all the packages the user has installed or instead of checking Lightning in `sys.modules` we use the solution you shared to check if they exist in installed packages?
If we are talking about the first case, wouldn't it be a bit overwhelming to show all installed packages sometimes?

carmocca: Since this collection is meant for bug reports, the more information we have, the better it is.

We can always filter the list to keep only packages with `lightning` or `torch` in the name.

But what's most interesting about such change is that we don't need to manually import each package, which has issues like the one you fixed in this PR

luca-medeiros: Makes sense. Appreciate the clarification!
I will tackle that point and CC you in the PR.

PR #13846, opened 2022-07-26
Reset epoch progress with batch size scaler

### LINKED ISSUE #13696: scale_batch_size does not work anymore?
## 🐛 Bug

After an update of PyTorch Lightning, the batch size scaling of `trainer.tune` is not working anymore: It just exhausts the maximum configured trials without actually probing the training loop. This results in incorrect batch sizes that cause Cuda OOM errors.

After some code diving I think this is due to a bug in the method [`_run_power_scaling`](https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/tuner/batch_size_scaling.py#L131) that sets `trainer.fit_loop.global_step = 0` instead of `trainer.fit_loop.epoch_loop.global_step = 0`, which seems to be required after a [recent change](https://github.com/Lightning-AI/lightning/commit/aea96e45a416838265eff0056b56647d57f605ee#diff-a3f9154e7c08e7e6f2c703a4a483679193a2f4268cf1cf495d466a872e7cd5fdR167) to `FitLoop`.

Edit:

The script below works as expected with `pytorch-lightning==1.5.9`.

### To Reproduce

```python
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringDatamodule(LightningDataModule):
    def __init__(self, batch_size):
        super().__init__()
        self.batch_size = batch_size

    def train_dataloader(self):
        return DataLoader(RandomDataset(1_000, 999_999), batch_size=self.batch_size)

    def val_dataloader(self):
        return DataLoader(RandomDataset(1_000, 999_999), batch_size=self.batch_size)


    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(1_000, 10_000)
        self.batch_size = "unset"

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

        return torch.optim.SGD(self.parameters(), lr=0.1)


def run():
    model = BoringModel().to("cuda:0")
    trainer = Trainer(
            default_root_dir=os.getcwd(),
            limit_train_batches=1,
            limit_val_batches=1,
            limit_test_batches=1,
            num_sanity_val_steps=0,
            max_epochs=1,
            enable_model_summary=False,
            auto_scale_batch_size=True,
            gpus=1
    )

    # the model can be fitted with batch_size=1000 on a NVIDIA 2080Ti without OOM errors
    datamodule = BoringDatamodule(batch_size=1000)

    trainer.tune(model=model, datamodule=datamodule, scale_batch_size_kwargs={"init_val": 1_000, "max_trials": 10})

    print(f"fitting with {datamodule.batch_size=} (fails on a NVIDIA 2080Ti)")
    trainer.fit(model, datamodule=datamodule)


if __name__ == "__main__":
    run()
```

### Expected behavior

`trainer.tune` finds a batch size that does not provoke a Cuda OOM error.

### Environment

* CUDA:
        - GPU:
                - NVIDIA GeForce RTX 2080 Ti
        - available:         True
        - version:           11.3
* Packages:
        - numpy:             1.21.6
        - pyTorch_debug:     False
        - pyTorch_version:   1.12.0+cu113
        - pytorch-lightning: 1.6.4
        - tqdm:              4.64.0
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.8.10
        - version:           127-Ubuntu SMP Wed May 18 14:30:56 UTC 2022

cc @akihironitta @borda @rohitgr7

cschell:
## What does this PR do?

1. This adds a test to check for the bug reported in #13696
2. Fixes the reported bugs. I'm not entirely convinced by my fix, it seems like this can easily break again: I fixed the issue by calling the reset functions of two deeply nested objects. Ideally, there should be a method dedicated to reset the state of the Trainer so that `_run_binsearch_scaling` and `_run_power_scaling` don't have to know about the Trainer internals. However, it seems to work fine for now, and since there is a test now for that particular issue, any breaking changes should at least get noticed.

Fixes #13696 

## Before submitting

## PR review

- [X] Is this pull request ready for review? (if not, please submit in draft mode)
- [X] Check that all items from **Before submitting** are resolved
- [X] Make sure the title is self-explanatory and the description concisely explains the PR

## Did you have fun?

🦦


tests/tests_pytorch/tuner/test_scale_batch_size.py
```diff
@@ -140,6 +140,31 @@ def test_auto_scale_batch_size_trainer_arg(tmpdir, scale_arg):
...
+    model = CounterModel()
+    trainer = Trainer(default_root_dir=tmpdir, auto_scale_batch_size=True, max_epochs=1, accelerator="gpu", devices=1)
+    max_trials = 5
+    trainer.tune(model, scale_batch_size_kwargs={"max_trials": max_trials, "mode": scale_mode})
+
+    expected_minimum_number_of_passed_epochs = max_trials
+    actual_number_of_passed_epochs = model.num_completed_epochs
+    assert actual_number_of_passed_epochs >= expected_minimum_number_of_passed_epochs
+
+
```
rohitgr7, line 167: I think this test can be integrated with the last one (test_dataloader_reset_with_scale_batch_size) where you can Mock the calls of `on_train_epoch_end`.

rohitgr7 (commented):
will be fixed by #11089, but test could help :)

tests/tests_pytorch/tuner/test_scale_batch_size.py
cschell: Sure, I can do that, but I would argue that it is nicer to have them separated, since currently each of these tests has a dedicated purpose?

rohitgr7: yeah, but every `trainer.tune` call increases the CI time and it's fine to merge certain assertions if one of them is not doing any significant work.

rohitgr7 (approved):

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -166,7 +175,16 @@ def _run_binsearch_scaling(
...
-        trainer.fit_loop.global_step = 0  # reset after each try
+
+        # reset after each try
+        if model.automatic_optimization:
+            optim_progress = trainer.fit_loop.epoch_loop.batch_loop.optimizer_loop.optim_progress
+        else:
+            optim_progress = trainer.fit_loop.epoch_loop.batch_loop.manual_loop.optim_step_progress
+
+        optim_progress.reset()
+        trainer.fit_loop.epoch_progress.current.reset()
```
carmocca, line 186:
```suggestion
        trainer.fit_loop.epoch_progress.reset_on_run()
```

src/pytorch_lightning/tuner/batch_size_scaling.py
```diff
@@ -166,7 +175,16 @@ def _run_binsearch_scaling(
...
         garbage_collection_cuda()
-        trainer.fit_loop.global_step = 0  # reset after each try
+
+        # reset after each try
+        if model.automatic_optimization:
+            optim_progress = trainer.fit_loop.epoch_loop.batch_loop.optimizer_loop.optim_progress
+        else:
+            optim_progress = trainer.fit_loop.epoch_loop.batch_loop.manual_loop.optim_step_progress
+
+        optim_progress.reset()
```
carmocca, line 185: the optim loop should call `reset_on_run()`, and manual loop `reset()`. So it will need to be done inside the `if`

```suggestion
        if model.automatic_optimization:
            trainer.fit_loop.epoch_loop.batch_loop.optimizer_loop.optim_progress.reset_on_run()
        else:
            trainer.fit_loop.epoch_loop.batch_loop.manual_loop.optim_step_progress.reset()```

**Review by carmocca — APPROVED** [review]

File: src/pytorch_lightning/tuner/batch_size_scaling.py [hunk 4]
```diff
@@ -128,7 +128,16 @@ def _run_power_scaling(
...
+        trainer.fit_loop.epoch_progress.current.reset()
```
carmocca, line 139: Let's wrap this into a protected method to reuse the code

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: shouldn't both call `reset`? we need to make sure it's a fresh start when running a new batch.

src/pytorch_lightning/tuner/batch_size_scaling.py
rohitgr7: shouldn't be reset? we need to make sure it's a fresh start when running a new batch.

carmocca: That's what `reset_on_run` does, plus resets other state that should be reset.

rohitgr7: `reset_on_run` doesn't reset the total, right? which we should?

carmocca: Ah you changed it, the diff was doing just `current.reset()` before

awaelchli (approved):

justusschock (approved):

PR #13880, opened 2022-07-27
Cast on host instead of IPU when using `precision=16` 

### LINKED ISSUE #13828: Cast happens after copy in mixed precision training on IPU
## 🐛 Bug

During mixed precision training on IPU, the cast happens after the copy. When working in half precision on IPU it is preferred to cast to cast to half precision on the host.

### To Reproduce

<details>
<summary>
repro.py
</summary>

```python
import os
import torch
import argparse
import poptorch
import torch.nn.functional as F
from torchvision import datasets, models
from torchvision.transforms import ToTensor
from pytorch_lightning.strategies.ipu import IPUStrategy


class Model(pl.LightningModule):
    def __init__(self):
        super().__init__()
        self.model = models.resnet18(pretrained=False)

    def forward(self, x):
        return self.model(x)

        x, y = batch
        y_hat = self.model(x)
        loss = F.cross_entropy(y_hat, y)
        return loss

        return torch.optim.Adam(self.parameters(), lr=1e-3)


def main():
    batch_size = 32
    # data
    mnist = datasets.FakeData(
        size=100 * batch_size,
        image_size=(3, 28, 28),
        num_classes=10,
        transform=ToTensor(),
    )
    opts = poptorch.Options()
    train_loader = poptorch.DataLoader(
        options=opts,
        dataset=mnist,
        batch_size=batch_size,
        drop_last=True,
        mode=poptorch.DataLoaderMode.Async,
    )
    # model
    model = Model()
    # training
    trainer = pl.Trainer(
        accelerator="ipu",
        precision=args.precision,
        max_epochs=1,
        strategy=IPUStrategy(training_opts=opts),
    )
    trainer.fit(model, train_loader)


if __name__ == "__main__":
    parser.add_argument("--precision", choices=[16, 32], default=16, type=int)
    args = parser.parse_args()
    os.environ["POPLAR_ENGINE_OPTIONS"] =f'{{"autoReport.all":"true","autoReport.directory":"profile_fp{args.precision}"}}'
    main()
```

</details>

```bash
python repro.py --precision 16
python repro.py --precision 32
```

Look at the profiles in PopVision Graph Analyser and see that the copy is FP32 followed by a cast to FP16 on the IPU.

### Expected behavior

In the [documentation](https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#precision), the video tells us that the weights remain FP32 and the data is cast to the selected precision. When training with FP16 data on IPU, the cast should occur before copying the data to device. The screenshots below show that the data is FP32 during the copy when `precision=16`.

### Environment

* Packages:
        - numpy:             1.23.1
        - pyTorch_debug:     False
        - pyTorch_version:   1.10.0+cpu
        - pytorch-lightning: 1.6.5
        - tqdm:              4.64.0
* System:
        - OS:                Linux
        - architecture:
                - 64bit
                - ELF
        - processor:         x86_64
        - python:            3.8.10
        - version:           #124-Ubuntu SMP Thu Apr 14 19:46:19 UTC 2022

### Additional context

FP16 - left, FP32 - right

In this image you can see that the total amount of data received by the IPU for this batch is 301184B in both cases. If we divide by the size of the batch, it can be seen that the datatype in both copies is FP32. `301184 / (32*3*28*28) = 4B = 32b`:
image

cc @carmocca @justusschock @awaelchli @akihironitta @rohitgr7

HMellor:
## What does this PR do?

Removes the cast to half from the forward pass and adds it to the host side pre-processing.

In overloading `batch_to_device` it also prevents `data.to(device)`, which is not currently necessary for IPU.

Fixes #13828

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/strategies/ipu.py
```diff
@@ -272,6 +255,15 @@ def to_tensor(x):
...
         return args
 
+    @staticmethod
+    def batch_to(data: Tensor) -> Tensor:
+        return data.half()
+
+    def batch_to_device(self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0) -> Any:
+        if self.precision_plugin.precision in (PrecisionType.MIXED, PrecisionType.HALF):
+            batch = apply_to_collection(batch, (FloatTensor, torch.cuda.FloatTensor), function=self.batch_to)
+        return batch
```
rohitgr7, line 265: since this wasn't overridden before, was it sending it to the IPU device before using the super class method?

rohitgr7: also, now at what point does it send the batch to the IPU device?

HMellor: > since this wasn't overridden before, was it sending it to the IPU device before using the super class method?

This used to happen in the forward pass in the `LightningIPUModule`, which was too late. Because by this time the data was already on the IPU.

> also, now at what point does it send the batch to the IPU device?

This is handled by PopTorch via [StreamCopies](https://docs.graphcore.ai/projects/graph-analyser-userguide/en/latest/user-guide.html#streamcopy) and shouldn't be done explicitly using `data.to(device)`

rohitgr7: so I think we can keep the `LightningIPUModule` and override `batch_to_device` to just return the batch and not do anything. This way it will be cast to fp16 first and then moved to the device inside the `forward` method.

also can you add a test?

HMellor: It can't be done in `forward` because `forward` is traced and therefore all operations in forward are compiled and uploaded to the IPU.

Would a test that checks the type of the input to `LightningIPUModule.forward` be good?

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -42,25 +42,8 @@
 
 
 class LightningIPUModule(_LightningModuleWrapperBase):
```
rohitgr7, line 44: we can remove this? since it doesn't do anything now.

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -272,6 +255,15 @@ def to_tensor(x):
         args = apply_to_collection(args, dtype=(int, float), function=to_tensor)
         return args
 
+    @staticmethod
+    def batch_to(data: Tensor) -> Tensor:
```
rohitgr7, line 259: let's make this a local function inside `batch_to_device`?

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -272,6 +255,15 @@ def to_tensor(x):
...
+        return data.half()
+
+    def batch_to_device(self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0) -> Any:
```
rohitgr7, line 262: can you add a comment here?

src/pytorch_lightning/strategies/ipu.py
rohitgr7: I think yet, or if there's any existing test, you can add some assertions there.

rohitgr7: > The data.to(device) that is currently called doesn't actually do anything

that means `self.device`doesn't point to the actual IPU device?

src/pytorch_lightning/strategies/ipu.py
awaelchli: If the logic here is really not needed, I'm also in favor (removal with deprecation)

src/pytorch_lightning/strategies/ipu.py
HMellor: > that means `self.device` doesn't point to the actual IPU device?

I'm not sure what is supposed to have `device` as an attribute (it's not an attribute of the `IPUStrategy`), but the `device` argument of `batch_to_device` is `None` when running my reproducer.

src/pytorch_lightning/strategies/ipu.py
HMellor: I have removed `LightningIPUModule`, but I'm not sure what you're deprecation strategy is so please let me know if there's anything else I need to do

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -142,8 +120,7 @@ def setup(self, trainer: "pl.Trainer") -> None:
         self._optimizer_zero_grad_original = self.lightning_module.optimizer_zero_grad
         self._disable_zero_grad()
 
-        model = LightningIPUModule(self.lightning_module, self.precision_plugin.precision)
-        self.model = model
+        self.model = _LightningModuleWrapperBase(self.lightning_module)
```
rohitgr7, line 123: IPUs can only trace the `forward` method? not the `training_step/validation_step`, etc...?

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -41,28 +41,6 @@
     poptorch = None
 
 
-class LightningIPUModule(_LightningModuleWrapperBase):
```
rohitgr7, line 44: to deprecate the class, keep it and do:
```python
class LightningIPUModule(_LightningModuleWrapperBase):
    def __init__(self, ... same old args):
        rank_zero_deprecation('`LightningIPUModule` has been deprecated in v1.8.0 and will be removed in v2.0.0')
        ... same old code
```
and a test in `tests/tests_pytorch/deprecated_api/test_remove_2-0.py`.

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -272,6 +245,20 @@ def to_tensor(x):
...
+        # is moved to the device to prevent wasteful host->device copies.
+        if self.precision_plugin.precision in (PrecisionType.MIXED, PrecisionType.HALF):
+
+            def to_half(data: Tensor) -> Tensor:
+                return data.half()
+
+            batch = apply_to_collection(batch, (FloatTensor, torch.cuda.FloatTensor), function=to_half)
+        # We don't call `super().batch_to_device` because `data.to(device)` is not
+        # currently necessary for IPUs. The movement of data from host<->IPU is
+        # currently handled by PopTorch.
```
rohitgr7, line 259: if that's the case, we need to do a follow-up for users to let them know that some of the hooks won't work with IPUs, like `transfer_batch_to_device`, ...

src/pytorch_lightning/strategies/ipu.py
rohitgr7: `self.device` is a general attribute that has the value of current device. Gives the ability to users to create tensors anytime and move them to device even if they are not coming from the dataloaders. That means this is not possible when using IPUs, right?

one can't do
```py
def training_step(...):
    some_tensor = torch.randn(...).to(self.device)  # this won't move the newly created tensor to IPU?
```

HMellor: > That means this is not possible when using IPUs, right?

That is correct

src/pytorch_lightning/strategies/ipu.py
HMellor: The `forward` method of the `model` argument passed to `poptorch.{training/inference}Model` is currently traced. Before this PR, the `LightningIPUModule` was being passed as the `model` argument, so its `forward` was being traced. This was why the cast was happening on the IPU instead of the host.

src/pytorch_lightning/strategies/ipu.py
HMellor: Is this something you could do? I don't know the codebase well enough to know all the implications.

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -45,6 +46,7 @@ class LightningIPUModule(_LightningModuleWrapperBase):
     def __init__(
         self, pl_module: Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase], precision: Union[str, int]
     ) -> None:
+        rank_zero_deprecation("`LightningIPUModule` is deprecated in v1.8 and will be removed in v2.0.0")
```
awaelchli, line 49:
```suggestion
        rank_zero_deprecation("`LightningIPUModule` is deprecated in v1.8 and will be removed in v1.10")
```

tests/tests_pytorch/deprecated_api/test_remove_2-0.py
```diff
@@ -50,6 +51,13 @@ def test_v2_0_0_deprecated_ipus(_, monkeypatch):
         _ = Trainer(ipus=4)
 
 
+@mock.patch("pytorch_lightning.accelerators.ipu.IPUAccelerator.is_available", return_value=True)
+def test_v2_0_0_deprecated_lightning_ipu_module(_, monkeypatch):
+    monkeypatch.setattr(pytorch_lightning.strategies.ipu, "_IPU_AVAILABLE", True)
+    with pytest.deprecated_call(match=r"is deprecated in v1.8 and will be removed in v2.0."):
```
awaelchli, line 57:
```suggestion
    with pytest.deprecated_call(match=r"is deprecated in v1.8 and will be removed in v1.10."):
```

tests/tests_pytorch/accelerators/test_ipu.py
```diff
@@ -40,16 +40,19 @@
 
 class IPUModel(BoringModel):
     def training_step(self, batch, batch_idx):
+        assert self.precision == torch.finfo(batch.dtype).bits
```
awaelchli, line 43: These lines would fail without the fix in the strategy, correct?

awaelchli (approved):
Thank you for working on this!

src/pytorch_lightning/strategies/ipu.py
awaelchli: If we merge it as bugfix before 1.7 is release, we should decrement the numbers here. Waiting for reviews and decision.

carmocca: I propose we just give this a 1 version deprecation.

And I agree this can be considered a bug-fix. So let's assume it will be deprecated in `v1.7.x` and removed in `v1.8`.

Meaning we update these messages, cherry-pick into a bug-fix release after merging this PR, and then open a new PR removing the code from master.

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -113,6 +113,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- When training with `precision=16` on IPU, the cast has been moved off the IPU onto the host, making the copies from host to IPU cheaper ([#13880](https://github.com/Lightning-AI/lightning/pull/13880))
```
carmocca, line 116: Since we will be considering this change a bug-fix, please move this into the "Fixed" section

tests/tests_pytorch/deprecated_api/test_remove_2-0.py
carmocca: And also move it to the correct file

src/pytorch_lightning/strategies/ipu.py
HMellor: I've updated this to deprecate in 1.7 and remove in 1.8

carmocca (approved):

tests/tests_pytorch/accelerators/test_ipu.py
HMellor: It passes now, but I haven't verified that it would have failed before. We shouldn't merge until I have

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -272,6 +269,16 @@ def to_tensor(x):
         args = apply_to_collection(args, dtype=(int, float), function=to_tensor)
         return args
 
+    def batch_to_device(self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0) -> Any:
+        # This override is necessary because the cast must occur before the data
+        # is moved to the device to prevent wasteful host->device copies.
+        if self.precision_plugin.precision in (PrecisionType.MIXED, PrecisionType.HALF):
+            batch = apply_to_collection(batch, (FloatTensor, torch.cuda.FloatTensor), function=Tensor.half)
```
rohitgr7, line 276:
```suggestion
            batch = apply_to_collection(batch, Tensor, function=Tensor.half)
```
if we need to move all kinds of tensors for IPUs

tests/tests_pytorch/deprecated_api/test_remove_1-8.py
```diff
@@ -1006,6 +1007,13 @@ def test_trainer_config_ipus(monkeypatch, trainer_kwargs, expected_ipus):
         trainer.ipus == expected_ipus
 
 
+@mock.patch("pytorch_lightning.accelerators.ipu.IPUAccelerator.is_available", return_value=True)
+def test_v1_8_0_deprecated_lightning_ipu_module(_, monkeypatch):
+    monkeypatch.setattr(pytorch_lightning.strategies.ipu, "_IPU_AVAILABLE", True)
+    with pytest.deprecated_call(match=r"is deprecated in v1.7 and will be removed in v1.8."):
```
rohitgr7, line 1013:
```suggestion
    with pytest.deprecated_call(match=r"has been deprecated in v1.7.0 and will be removed in v1.8."):
```

src/pytorch_lightning/strategies/ipu.py
```diff
@@ -45,6 +46,7 @@ class LightningIPUModule(_LightningModuleWrapperBase):
     def __init__(
         self, pl_module: Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase], precision: Union[str, int]
     ) -> None:
+        rank_zero_deprecation("`LightningIPUModule` is deprecated in v1.7 and will be removed in v1.8.0")
```
rohitgr7, line 49:
```suggestion
        rank_zero_deprecation("`LightningIPUModule` has been deprecated in v1.7.0 and will be removed in v1.8.0")
```

rohitgr7 (approved):
LGTM

tests/tests_pytorch/accelerators/test_ipu.py
HMellor: Since `batch_to_device` is the last thing that happens before the tensors are _supposed_ to be on the device, the updated test should be valid. It fails before the PR and passes after.

src/pytorch_lightning/strategies/ipu.py
carmocca: Is this the change that caused https://github.com/Lightning-AI/lightning/pull/13983?

justusschock: that's part of it, but before e.g. DoubleTensors, BfloatTensors etc. wouldn't have been cast as well.

PR #13894, opened 2022-07-27
[CLI] add support to run app on a specific cluster

nicolai86:
Follow up PR for https://github.com/Lightning-AI/lightning/pull/13835.

This PR changes the run command to support a `--cluster-id` flag when `--cloud` is provided.
This way customers can run applications on their own BYOC compute clusters.

## Verification

- new unit tests
- updated existing unit tests
- working e2e test


awaelchli:
@nicolai86 I suggest you rebase this branch onto the other one, then change the base branch here in github to the other one to have the reviewers only see the diff. This way, not all changes need to be reviewed again. 

image

To request a review, click "ready for review" at the bottom. Code-owners will get a notification.

tests/tests_app_examples/test_v0_app.py
```diff
@@ -30,6 +30,42 @@ def test_v0_app_example():
...
+    with run_app_in_cloud(
+        os.path.join(_PROJECT_ROOT, "examples/app_v0"),
+        extra_args=["--cluster-id", os.environ.get("LIGHTNING_BYOC_CLUSTER_ID")],
+    ) as (
+        _,
+        view_page,
+        fetch_logs,
+    ):
+
+        def check_content(button_name, text_content):
```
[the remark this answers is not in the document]
nicolai86: done :+1:

src/lightning_app/cli/lightning_cli.py
```diff
@@ -121,6 +130,9 @@ def run():
 @run.command("app")
 @click.argument("file", type=click.Path(exists=True))
 @click.option("--cloud", type=bool, default=False, is_flag=True)
+@click.option(
+    "--cluster-id", type=str, default=None, help="Run Lightning App on a specific Lightning AI BYOC compute cluster"
```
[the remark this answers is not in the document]
nicolai86: @Felonious-Spellfire WDYT?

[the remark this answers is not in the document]
manskx: I think we can do something like:
"Run Lightning App on your own cluster. check <URL> to know more"

.lightningignore
```diff
@@ -0,0 +1 @@
+.git
```
manskx, line 1: Could this be removed  ?

src/lightning_app/CHANGELOG.md
```diff
@@ -10,7 +10,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Add support for `Lightning App Commands` through the `configure_commands` hook on the Lightning Flow and the `ClientCommand`  ([#13602](https://github.com/Lightning-AI/lightning/pull/13602))
 - Add support for Lightning AI BYOC cluster management ([#13835](https://github.com/Lightning-AI/lightning/pull/13835))
```
manskx, line 12: nit: Maybe we can update the changelog to include this PR also

.lightningignore
nicolai86: when I run the acceptance tests locally the entire lightning repository is packaged and uploaded to Lightning AI - maybe this is due to how I run things.
I'll remove it.

src/lightning_app/CHANGELOG.md
nicolai86: good point, added!

src/lightning_app/cli/lightning_cli.py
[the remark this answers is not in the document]
Felonious-Spellfire: @tchaton @nicolai86 

Linking to the doc would be great if we can do it.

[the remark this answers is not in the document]
nicolai86: Can we follow up on just the doc string in a separate PR? I don't know how to get a proper URL without https://github.com/Lightning-AI/lightning/pull/13976 being merged & released.

manskx (approved):

src/lightning_app/runners/cloud.py
```diff
@@ -157,14 +160,21 @@ def dispatch(
                 enable_app_server=app_spec.enable_app_server,
                 flow_servers=app_spec.flow_servers,
                 image_spec=app_spec.image_spec,
+                cluster_id=cluster_id,
```
hhsecond, line 163: I am not sure coupling app releases with a cluster is the right thing to do here. Apps releases are like new app versions that's available in the app store. The cluster coupling comes up when somebody tries to run this app version and hence should be an attribute of app instance (few lines below this)

src/lightning_app/cli/lightning_cli.py
```diff
@@ -142,7 +155,7 @@ def run_app(
     app_args: List[str],
 ):
     """Run an app from a file."""
-    _run_app(file, cloud, without_server, no_cache, name, blocking, open_ui, env)
+    _run_app(file, cloud, cluster_id, without_server, no_cache, name, blocking, open_ui, env)
```
hhsecond, line 158: can we raise if `cluster-id` is provided without `--cloud`

src/lightning_app/runners/cloud.py
```diff
@@ -108,6 +110,7 @@ def dispatch(
                 random_name = "".join(random.choice(string.ascii_lowercase) for _ in range(5))
                 spec = V1LightningworkSpec(
                     build_spec=build_spec,
+                    cluster_id=cluster_id,
```
hhsecond, line 113: `work.cloud_compute` already has a `clusters` attribute. We should check if there is a conflict between the values there and the `--cluster-id` provided and then either raise or continue with the priority one but that's probably a decision we need from Product

hhsecond (changes requested):
Thanks for the PR @nicolai86. Overall, the flow looks good

I have made few comments about some of the early design choices we made which conflicts with the design of this PR. We don't necessarily have to go with the original design but we need to come to a conscious consensus on how we should go about it

src/lightning_app/runners/cloud.py
hhsecond: I am not even sure why we have `cluster-id` in the release spec at all

src/lightning_app/cli/lightning_cli.py
nicolai86: done, and added unit tests for it.

src/lightning_app/runners/cloud.py
nicolai86: discussed internally - for now let's not rely on `work.cloud_compute.clusters` as we don't support the suggested functionality of distributing work between different clusters today

src/lightning_app/runners/cloud.py
nicolai86: From manual verification, the cluster didn't start on the selected cluster without passing this parameter to the release.

I think we need to circle back on our service design and omit this in the API if we don't need it. For now it's required to support this sadly.

src/lightning_app/runners/cloud.py
```diff
@@ -157,14 +160,21 @@ def dispatch(
...
             )
+            if cluster_id is not None:
+                self._ensure_cluster_project_binding(project.project_id, cluster_id)
+
             lightning_app_release = self.backend.client.lightningapp_v2_service_create_lightningapp_release(
                 project.project_id, lightning_app.id, release_body
             )
 
+            if cluster_id is not None:
+                print(f"running app on {lightning_app_release.cluster_id}")
```
tchaton, line 176: Let's use logger and provide a better message to the users.

tchaton (approved):
LGTM !

src/lightning_app/cli/lightning_cli.py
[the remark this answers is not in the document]
awaelchli: How does the cluster id relate to the cluster name given in the cloud compute? 

Also, @nicolai86 can you give us an example of such an ID? Is it human readable or just a long number?

awaelchli (approved):

src/lightning_app/runners/cloud.py
nicolai86: good point, changed to logger.

src/lightning_app/cli/lightning_cli.py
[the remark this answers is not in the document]
nicolai86: today the cluster id is the human readable name provided by the customer when the cluster is created.

so it's anything within `[a-zA-Z][a-zA-Z-]+`, in the control of the cluster creator.

ie for Lightning we use `litng-ai-03`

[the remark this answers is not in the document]
awaelchli: @tchaton That means we will remove the cluster argument from the CloudCompute?

PR #13895, opened 2022-07-28
Include `nn.Parameter` in ModelSummary

### LINKED ISSUE #12736: ModelSummary ignores nn.Parameter
## 🐛 Bug

I don't know if this is a bug of a feature, but currently ModelSummary only reports member `nn.Module`s and ignores `nn.Parameter`s that are directly created. I believe these parameters are an important part (and sometimes, a big part in terms of the number of weights) of the model so that they should be reported.

cc @borda @kaushikb11 @awaelchli @rohitgr7

yeonju7kim:
## What does this PR do?

When we defined torch.nn.parameter object in the model and then print ModelSummary for the model, the printed result doesn't show the parameter number of the torch.nn.parameter object.
Let me show you the example.

``` python
import torch

from pytorch_lightning.utilities.model_summary import ModelSummary

    def __init__(self):
        super().__init__()
        self.param = torch.nn.Parameter(torch.rand(2, 2))  # 2 * 2 = 4
        self.layer = torch.nn.Linear(2, 2)  # 2 * 2 + 2 = 6


if __name__ == "__main__":
    model = BoringModel()
    summary = ModelSummary(model)
    print(summary)
```

```
  | Name  | Type   | Params
---------------------------------
0 | param | Parameter | 4                        <- I add this line. 
1 | layer | Linear    | 6          
---------------------------------
10        Trainable params
0         Non-trainable params
10        Total params
0.000     Total estimated model params size (MB)
```

Fixes #12736

None

## Before submitting

## PR review

## Did you have fun?

Yes! It's my first contribution. I welcome your feedback!

### DIFF
--- a/src/pytorch_lightning/utilities/deepspeed_model_summary.py
+++ b/src/pytorch_lightning/utilities/deepspeed_model_summary.py
@@ -38,7 +38,8 @@ class DeepSpeedLayerSummary(LayerSummary):
         """Returns the number of parameters in this module."""
         if isinstance(self._module, torch.nn.Module):
             return sum(
-                deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+                deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters()
+            )
         elif isinstance(self._module, torch.nn.Parameter):
             return int(np.prod(self._module.shape))
         else:
@@ -53,6 +54,7 @@ class DeepSpeedLayerSummary(LayerSummary):
             return int(np.prod(self._module.shape))
         else:
             raise NotImplementedError
+
 
 class DeepSpeedSummary(ModelSummary):
     def summarize(self) -> Dict[str, DeepSpeedLayerSummary]:  # type: ignore[override]
--- a/src/pytorch_lightning/utilities/deepspeed_model_summary.py
+++ b/src/pytorch_lightning/utilities/deepspeed_model_summary.py
@@ -16,6 +16,7 @@
 from collections import OrderedDict
 from typing import Dict, List, Tuple
 
+import numpy as np
 import torch
 
 from pytorch_lightning.utilities.model_summary import (
@@ -35,13 +36,23 @@ class DeepSpeedLayerSummary(LayerSummary):
     @property
     def num_parameters(self) -> int:
         """Returns the number of parameters in this module."""
-        return sum(deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        if isinstance(self._module, torch.nn.Module):
+            return sum(
+                deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        elif isinstance(self._module, torch.nn.Parameter):
+            return int(np.prod(self._module.shape))
+        else:
+            raise NotImplementedError
 
     @property
     def average_shard_parameters(self) -> int:
         """Returns the number of parameters in this module."""
-        return sum(p.partitioned_size() if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
-
+        if isinstance(self._module, torch.nn.Module):
+            return sum(p.partitioned_size() if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        elif isinstance(self._module, torch.nn.Parameter):
+            return int(np.prod(self._module.shape))
+        else:
+            raise NotImplementedError
 
 class DeepSpeedSummary(ModelSummary):
     def summarize(self) -> Dict[str, DeepSpeedLayerSummary]:  # type: ignore[override]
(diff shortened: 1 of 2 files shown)


yeonju7kim:
I should fix mypy error first.

src/pytorch_lightning/utilities/deepspeed_model_summary.py
```diff
@@ -35,12 +36,24 @@ class DeepSpeedLayerSummary(LayerSummary):
     @property
     def num_parameters(self) -> int:
         """Returns the number of parameters in this module."""
-        return sum(deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        if isinstance(self._module, torch.nn.Module):
+            return sum(
+                deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters()
+            )
+        elif isinstance(self._module, torch.nn.Parameter):
+            return int(np.prod(self._module.shape))
```
justusschock, line 44: probably need deepspeed handling for this kind of parameter as well?

src/pytorch_lightning/utilities/deepspeed_model_summary.py
```diff
@@ -35,12 +36,24 @@ class DeepSpeedLayerSummary(LayerSummary):
...
+            raise NotImplementedError
 
     @property
     def average_shard_parameters(self) -> int:
         """Returns the number of parameters in this module."""
-        return sum(p.partitioned_size() if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        if isinstance(self._module, torch.nn.Module):
+            return sum(p.partitioned_size() if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        elif isinstance(self._module, torch.nn.Parameter):
+            return int(np.prod(self._module.shape))
```
justusschock, line 54: this doesn't look right. should also return either the partitioned_size or 0

src/pytorch_lightning/utilities/model_summary.py
```diff
@@ -241,7 +268,12 @@ def model_size(self) -> float:
         return self.total_parameters * self._precision_megabytes
 
     def summarize(self) -> Dict[str, LayerSummary]:
-        summary = OrderedDict((name, LayerSummary(module)) for name, module in self.named_modules)
+        all_parameter: List[Tuple[str, Union[nn.Module, nn.Parameter]]]
+        all_parameter = []
+        all_parameter.extend(self.named_modules)
+        all_parameter.extend(self.named_parameters)
```
justusschock, line 274: are you sure this isn't traversing recursively?

src/pytorch_lightning/utilities/deepspeed_model_summary.py
```diff
@@ -35,12 +36,24 @@ class DeepSpeedLayerSummary(LayerSummary):
     @property
     def num_parameters(self) -> int:
         """Returns the number of parameters in this module."""
-        return sum(deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        if isinstance(self._module, torch.nn.Module):
+            return sum(
+                deepspeed_param_size(p) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters()
+            )
+        elif isinstance(self._module, torch.nn.Parameter):
```
justusschock, line 43: I would handle this separately. The docstring says it handles the numver of parameters in **this module** so the module itself shouldn't be a nn.Parameter .

Also the class is named `LayerSummary`, to me that looks like directly registered parameters should be handled externally as well.

src/pytorch_lightning/utilities/model_summary.py
```diff
@@ -120,9 +126,13 @@ def layer_type(self) -> str:
     @property
     def num_parameters(self) -> int:
         """Returns the number of parameters in this module."""
-        return sum(
-            cast(int, np.prod(p.shape)) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters()
-        )
+        if isinstance(self._module, nn.Module):
+            return sum(int(np.prod(p.shape)) if not _is_lazy_weight_tensor(p) else 0 for p in self._module.parameters())
+        elif isinstance(self._module, nn.Parameter):
```
justusschock, line 131: Same thing. IMO this doesn't belong in this class.

justusschock (commented):
In general: What about buffers they may contribute to the memory footprint as well.

src/pytorch_lightning/utilities/model_summary.py
yeonju7kim: I used 'reverse', so it will not traverse recursively.

src/pytorch_lightning/utilities/deepspeed_model_summary.py
yeonju7kim: The class LayerSummary was originally the class that summarizes a single layer whose type was nn.Module.
However, not to ignore nn.Parameter I changed the type of layer to be nn.Module or nn.Parameter.
Because it can be nn.Module and nn.Parameter, I should write as above.

I think I should make another class that summarizes nn.Parameter and it will be class ParameterSummary.

PR #13903, opened 2022-07-28
Raise an error if `sync_dist` is set when logging torchmetrics

rohitgr7:
## What does this PR do?

sync_dist is ignored when `torchmetrics` instances are logged.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/trainer/connectors/logger_connector/result.py
+++ b/src/pytorch_lightning/trainer/connectors/logger_connector/result.py
@@ -452,6 +452,18 @@ class _ResultCollection(dict):
     ) -> None:
         """See :meth:`~pytorch_lightning.core.module.LightningModule.log`"""
         # no metrics should be logged with graphs
+        if (
+            sync_dist
+            and not isinstance(value, Tensor)
+            and (isinstance(value, Metric) or any(isinstance(v, Metric) for v in value.values()))
+        ):
+            raise MisconfigurationException(
+                "Setting `self.log(..., sync_dist=, reduce_fx=, sync_dist_group=)` won't have any effect if you are"
+                " logging a `torchmetrics` instance. Refer to"
+                " `https://torchmetrics.readthedocs.io/en/latest/pages/lightning.html` to configure distributed sync"
+                " across processed with `torchmetrics`."
+            )
+
         if not enable_graph:
             value = recursive_detach(value)
 
--- a/tests/tests_pytorch/trainer/logging_/test_logger_connector.py
+++ b/tests/tests_pytorch/trainer/logging_/test_logger_connector.py
@@ -700,3 +700,18 @@ def test_result_collection_no_batch_size_extraction():
     assert results["training_step.epoch_log_val"].value == log_val * batch_size
     assert results["training_step.epoch_log_val"].cumulated_batch_size == batch_size
     assert results["training_step.epoch_sum_log_val"].value == log_val
+
+
+def test_sync_dist_with_torchmetrics():
+    acc = Accuracy()
+    target = torch.tensor([0, 1, 2, 3])
+    preds = torch.tensor([0, 2, 1, 3])
+    acc.update(preds, target)
+
+    results = _ResultCollection(True)
+
+    with pytest.raises(
+        MisconfigurationException,
+        match=r"Setting `self.log\(..., sync_dist=, reduce_fx=, sync_dist_group=\)` won't have any effect",
+    ):
+        results.log("foo", "bar", acc, sync_dist=True)


rohitgr7:
An alternate solution (which I think is better) can be to set these values (`sync_dist`, `reduce_fx`, `sync_dist_group`) for `torchmetrics` instance internally with an info message without raising any such errors.

thoughts?

src/pytorch_lightning/trainer/connectors/logger_connector/result.py
```diff
@@ -452,6 +452,18 @@ def log(
     ) -> None:
         """See :meth:`~pytorch_lightning.core.module.LightningModule.log`"""
         # no metrics should be logged with graphs
+        if (
+            sync_dist
+            and not isinstance(value, Tensor)
+            and (isinstance(value, Metric) or any(isinstance(v, Metric) for v in value.values()))
+        ):
+            raise MisconfigurationException(
```
carmocca, line 460: I would personally put this in the `LightningModule.log` method where we have all the other validation checks

carmocca (commented):
> An alternate solution (which I think is better) can be to set these values (sync_dist, reduce_fx, sync_dist_group) for torchmetrics instance internally with an info message without raising any such errors. For this, we might have to make some changes to the Metric class.

Why is it better? Failing should be fine and it makes it very clear that they shouldn't be passed.

> we might have to make some changes to the Metric class

Why?

rohitgr7:
> > An alternate solution (which I think is better) can be to set these values (sync_dist, reduce_fx, sync_dist_group) for torchmetrics instance internally with an info message without raising any such errors. For this, we might have to make some changes to the Metric class.
> 
> Why is it better? Failing should be fine and it makes it very clear that they shouldn't be passed.
> 
> > we might have to make some changes to the Metric class
> 
> Why?

just for better UX. else the user has to take care of setting this at two different places in they are using both torchmetrics and tensors. Or even if they are using torchmetrics, it makes it easy for them since PL will handle it and it's their intention too. Just makes it easy for them since both are lightning products.

The changes that will be required are just the setters/getters for these arguments.

awaelchli:
I look at this from a different angle. 

The default for sync_dist in the log method is False. Which means when unset, it indicates that we will never communicate with different processes to compute the metric. When logging a TorchMetric however, this is clearly not the case. On the other hand, setting sync_dist=True explicitly and logging a TorchMetric does make sense, even if it doesn't trigger the exact same logic. It makes it appear that it is being "ignored", and raising an error here would prevent anyone to combine it with logging both a scalar and a TorchMetric. One can argue either way, I think it is up to us to determine whether this is considered a convenience or a misleading setting. 

If we decide to produce an error like proposed in this PR, we would additionally need to set the default from False to None and also error when it gets set to False.

rohitgr7:
> and raising an error here would prevent anyone to combine it with logging both a scalar and a TorchMetric.

@awaelchli 
not really, since they are checked per log value.
```py
self.log('some_val', scalar_val, sync_dist=True)  <- does not raise an error
self.log('some_tm_instance, torchmetric_instance, sync_dist=True)  <- raises an error
```
which means users should unset sync_dist for the second log. I can add the key name to make it more explicit. And this can easily be caused with `fast_dev_run` since we don't disable logging, but the loggers.

Also, the only reason I raised this PR is to avoid the silent behavior here and have seen concerns regarding this over slack/discussions.

awaelchli:
> not really, since they are checked per log value.

My comment was about log_dict, when different types of values get submitted in a dict. 

> If we decide to produce an error like proposed in this PR, we would additionally need to set the default from False to None and also error when it gets set to False.

Do you agree with this or not?

rohitgr7:
> My comment was about log_dict, when different types of values get submitted in a dict together.

yes, in that case, user has to separate them and make 2 calls. That's why I suggested this https://github.com/Lightning-AI/lightning/pull/13903#issuecomment-1197994979, which will end up with no errors.

> If we decide to produce an error like proposed in this PR, we would additionally need to set the default from False to None and also error when it gets set to False.

why do we need to raise an error when it's False? with DDP, we are not forcing them to sync always.

awaelchli:
> why do we need to raise an error when it's False? with DDP, we are not forcing them to sync always.

When using torchmetrics, it will sync whether you want it or not. Internally inside torchmetrics. [One can chose whether it happens additionally on step or not](https://torchmetrics.readthedocs.io/en/stable/pages/implement.html#torchmetrics.Metric). A user logging a metric and setting `sync_dist=False` is a contradiction, conceptually. 

Honestly, I wasn't sure in some of the statements I made earlier and had a hard time looking at the source code behind our result objects. I don't feel very confident, perhaps some of these checks are already done internally, but I wouldn't know where to look.

rohitgr7:
> When using torchmetrics, it will sync whether you want it or not.

True, sorry, I noticed this later. We can make it `None` by default.

> [One can chose whether it happens additionally on step or not](https://torchmetrics.readthedocs.io/en/stable/pages/implement.html#torchmetrics.Metric).

we can determine that based on `on_step` and `on_epoch` if `sync_dist!=None`:
- If either just `on_step=True` or just `on_epoch=True`, set the flag inside `torchmetrics` accordingly
- If `on_step=True` and `on_epoch=True`:
  - Set both the flags
  - Set only `sync_on_compute` to align it with how we handle scalar metrics.
  - Raise an error to ask user to set it by themselves, since we can't determine if they want to sync `on_step`, `on_epoch` or on both.

> perhaps some of these checks are already done internally, but I wouldn't know where to look.

if you are talking about:
- in case torchmetrics are logged, what happens when `sync_dist=True`? [It is ignored](https://github.com/Lightning-AI/lightning/blob/0883971ccb973be3f8d0495f442f80f253fd45bb/src/pytorch_lightning/trainer/connectors/logger_connector/result.py#L260-L267)
- what happens on step level if `sync_dist=True`. [Nothing](https://github.com/Lightning-AI/lightning/blob/0883971ccb973be3f8d0495f442f80f253fd45bb/src/pytorch_lightning/trainer/connectors/logger_connector/result.py#L525-L545).

carmocca:
This is a valuable improvement but low priority for the team. I created an issue to track this idea so that somebody can pick it up. Closing the PR.

PR #13918, opened 2022-07-28
Stop Lightning app from CLI

alecmerdler:
## What does this PR do?

Implements 'lightning stop app {id}'. This enables stopping
a Lightning app running in the cloud without using the UI.

#### Usage

1. Run any Lightning app in the cloud.
```
$ lightning run app --cloud app.py
```
2. Wait for app to start running and get its ID from the admin view URL.
3. Stop the app using the following command:
```
$ lightning stop app {id}
```

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_app/cli/lightning_cli.py
+++ b/src/lightning_app/cli/lightning_cli.py
@@ -19,11 +19,11 @@ from lightning_app.utilities.cli_helpers import (
     _format_input_env_variables,
     _retrieve_application_url_and_available_commands,
 )
+from lightning_app.utilities.cloud import _get_project, stop_app_cloud
 from lightning_app.utilities.install_components import register_all_external_components
 from lightning_app.utilities.login import Auth
-from lightning_app.utilities.state import headers_for
 from lightning_app.utilities.network import LightningClient
-from lightning_app.utilities.cloud import stop_app_cloud, _get_project
+from lightning_app.utilities.state import headers_for
 
 logger = logging.getLogger(__name__)
 
--- a/src/lightning_app/cli/lightning_cli.py
+++ b/src/lightning_app/cli/lightning_cli.py
@@ -22,6 +22,8 @@ from lightning_app.utilities.cli_helpers import (
 from lightning_app.utilities.install_components import register_all_external_components
 from lightning_app.utilities.login import Auth
 from lightning_app.utilities.state import headers_for
+from lightning_app.utilities.network import LightningClient
+from lightning_app.utilities.cloud import stop_app_cloud, _get_project
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/lightning_app/cli/lightning_cli.py
```diff
@@ -206,7 +208,24 @@ def stop():
     pass
 
 
-@_main.group(hidden=True)
+@stop.command("app")
+@click.argument("id", type=str)
+def stop_app(id: str):
```
alecmerdler, line 213: @awaelchli I could use some help with how to structure the logic for this command, specifically where to keep the API calls to the backend. Any suggestions?

awaelchli: @alecmerdler I don't know if I'm the best to ask. As we can see looking at the file structure in the cli folder, it is already a mess. Perhaps it is best to follow the pattern that emerged with the other commands for create, delete, list, etc. 

`lightning_cli_stop.py` contains the stop CLI subcommand definition
`cmd_stop.py` contains the api calls, error handling etc. 

What do you think?
cc @tchaton

src/lightning_app/cli/lightning_cli.py
```diff
@@ -206,7 +208,24 @@ def stop():
     pass
 
 
-@_main.group(hidden=True)
+@stop.command("app")
+@click.argument("id", type=str)
```
hhsecond, line 212: I am not sure if this is the right UX though. ID is not a memorizable string. So users need to get to the UI to copy the id to execute this command, which essentially defeat the purpose no?

Why wouldn't we use app name here?

src/lightning_app/cli/lightning_cli.py
alecmerdler: I agree, the CLI could use some cleanup. Your suggestion makes sense as a basic step to separate logic from presentation. I'll change the code to do that.

src/lightning_app/cli/lightning_cli.py
alecmerdler: The main reason to not use app name here is because the app name is not guaranteed to be unique. We are also adding the ability for the CLI to list apps, which could return their IDs to be used to delete a given app without using the UI.

hhsecond: The name is guarenteed to be unique under each project. We could raise a warning if user has multiple apps with same name but otherwise delete the one with the name provided?

hhsecond: Hmm, the BE List API requires a project ID to be passed always which makes the FE logic a bit complex. Perhaps this could be a feature for future

alecmerdler:
Bump

alecmerdler:
not stale

PR #13921, opened 2022-07-29
Snapshot selected globals and restore them in spawned process

### LINKED ISSUE #12685: Deterministic flags and other globals do not get transferred to spawned processes
## 🐛 Bug

When using the ddp-spawn strategy, global flags like the torch.use_deterministic_algorithms flags set by the Trainer do not get transferred to the worker processes. The reason is that torch gets re-imported when spawning the workers (naturally) and thus the globals are back to defaults.  

### To Reproduce

```python
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        assert torch.are_deterministic_algorithms_enabled()
        loss = self(batch).sum()
        return {"loss": loss}

def run():
    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        enable_model_summary=False,
        enable_progress_bar=False,
        strategy="ddp_spawn",
        devices=2,
        accelerator="cpu",
        deterministic=True,
    )


if __name__ == "__main__":
    run()

```

### Expected behavior

See assertion in the code. 

### Environment

PyTorch 1.10 and 1.11
PL 1.6 (master)

### Additional context

I dreamed about it last night (no kidding) and checked this morning and the bug turned out to be real.

This applies to all globals from torch, not just deterministic flags. We already handle this problem for the seed in the `DDPSpawnStrategy._worker_setup()` method.

awaelchli:
## What does this PR do?

Fixes #12685

The problem: Globals like torch deterministic flags or the seed get lost when spawning processes. This is caused because essentially the torch module gets re-imported in the child process, leading to globals being re-initialized. This is simply the nature of how `mp.spawn` works. This means it only affects ddp spawn (multiprocessing start method "spawn"). 

- Subprocess launch is not affected because all processes start from an identical state by design
- Fork is not affected because all children inherit the exact memory of the parent process

We can solve this by picking the globals we care about, save them, send them to the child process and restore them there before we launch the trainer run methods. Hence, this PR implements this directly in the launcher. 

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/strategies/ddp_spawn.py
```diff
@@ -175,7 +174,6 @@ def set_world_ranks(self, process_idx: int = 0) -> None:
         rank_zero_only.rank = self.cluster_environment.global_rank()
 
     def _worker_setup(self, process_idx: int) -> None:
-        reset_seed()
```
awaelchli, line 178: This is now handled in the launcher, so not necessary anymore. Would anyway not have been necessary when using fork.

akihironitta (approved):
LGTM!

carmocca (approved):

Borda (approved):

src/pytorch_lightning/strategies/launchers/multiprocessing.py
```diff
@@ -209,3 +223,53 @@ class _WorkerOutput(NamedTuple):
...
+
+            # in main process
+            snapshot = _GlobalStateSnapshot.capture()
+
+            # in worker process
+            snapshot.restore()
+    """
+
+    use_deterministic_algorithms: bool
+    use_deterministic_algorithms_warn_only: bool
```
rohitgr7, line 247: maybe capture these 2 in a dict? then you can pass it by checking the `torch_version` only once.

rohitgr7 (approved):

src/pytorch_lightning/strategies/launchers/multiprocessing.py
```diff
@@ -209,3 +223,53 @@ class _WorkerOutput(NamedTuple):
...
+        that we want to restore in a spawned worker process.
+        """
+        return cls(
+            use_deterministic_algorithms=torch.are_deterministic_algorithms_enabled(),
+            use_deterministic_algorithms_warn_only=torch.is_deterministic_algorithms_warn_only_enabled()
+            if _TORCH_GREATER_EQUAL_1_11
+            else False,
+            cudnn_benchmark=torch.backends.cudnn.benchmark,
+            rng_states=_collect_rng_states(),
+        )
```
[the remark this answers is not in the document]
awaelchli: done

src/pytorch_lightning/strategies/launchers/multiprocessing.py
awaelchli: Tried but makes it only more complicated, and requires to introduce new names. I prefer the explicit way as it is also easy to annotate with types.
The torch version check can be removed once our minimal version is 1.11.

tchaton (approved):
LGTM !

kaushikb11 (approved):

Issue #13124, opened 2022-05-21
Resuming from a mid-epoch checkpoint produces negative time estimates


fishbotics:
## 🐛 Bug

I have a checkpoint saved that I reloaded with `trainer(..., ckpt_path=...)` and then started training. The checkpoint was fairly early into the first epoch. Loading this epoch, I get some warnings and finally a super weird time estimate for training. The warnings are:

```
/usr/local/lib/python3.8/dist-packages/pytorch_lightning/callbacks/model_checkpoint.py:342: UserWarning: The dirpath has changed from '/workspace/mpinets_checkpoints/11dh7on3' to '/workspace/mpinets_checkpoints/<bound method DummyExperiment.nop of
 <pytorch_lightning.loggers.base.DummyExperiment object at 0x7fe38046ed60>>', therefore `best_model_score`, `kth_best_model_path`, `kth_value`, `last_model_path` and `best_k_models` won't be reloaded. Only `best_model_path` will be reloaded.
  warnings.warn(
```
and 
```
/usr/local/lib/python3.8/dist-packages/pytorch_lightning/loops/epoch/training_epoch_loop.py:135: UserWarning: You're resuming from a checkpoint that ended before the epoch ended. This can cause unreliable results if further training is done. Consider using an end-of-epoch checkpoint or enabling fault-tolerant training: https://pytorch-lightning.readthedocs.io/en/stable/advanced/fault_tolerant_training.html                       
```

And then finally, the training estimate comes back as negative. It looks like this :
```
Epoch 0:   0%|                                            | 11/63913 [05:54<-1:28:50, -34.17it/s, loss=0.00641, v_num=4rf6]
```

- PyTorch Lightning Version (e.g., 1.5.0): 1.6.3
- PyTorch Version (e.g., 1.10): 1.11
- Python version (e.g., 3.9): 3.8
- OS (e.g., Linux): Ubuntu 20.04
- CUDA/cuDNN version: 11.3
- GPU models and configuration: 8 A100s using ddp
- How you installed PyTorch (`conda`, `pip`, source): pip
- If compiling from source, the output of `torch.__config__.show()`:
- Any other relevant information:

### Additional context

cc @tchaton @rohitgr7 @awaelchli

awaelchli:
@fishbotics In your post here you included the warning message that tries to explain why this occurs:

[...]

PR #12889, opened 2022-04-26
Fix tqdm total and update

rohitgr7:
## What does this PR do?

Fixes #12875

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #13962, opened 2022-08-01
Fix resuming the tqdm progress bar 

awaelchli:
## What does this PR do?

The changes in #12889 have caused #13124. 

This PR reverts #12889 (except tests) and adds the fix to call `tqdm.initial = 0`, which prevents from tqdm incorrectly computing the estimates for ETA and iterations/s. All tests added in #12889 remain as they are not affected by reverting this. 

Since this is a visual bug, and the tqdm produces these estimates at runtime, I don't really know how to test this. Suggestions by reviewers are welcome!

<details>
  <summary>Repro example used to debug:</summary>
  

```py

import os
import shutil
from time import sleep

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        sleep(.1)
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        print()
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("val_loss", loss)
        print()

def run():
    train_data = DataLoader(RandomDataset(32, 32), batch_size=2)

    if os.path.exists("lightning_logs"):
        shutil.rmtree("lightning_logs")

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        max_epochs=1,
        enable_model_summary=False,
        enable_progress_bar=True,
        callbacks=ModelCheckpoint(monitor="train_loss", save_top_k=-1, every_n_train_steps=1),
    )
    trainer.fit(model, train_dataloaders=train_data, val_dataloaders=train_data)

    trainer = Trainer(
        default_root_dir=os.getcwd(),
        max_epochs=3,
        enable_model_summary=False,
        enable_progress_bar=True,
        callbacks=ModelCheckpoint(monitor="train_loss", save_top_k=-1, every_n_train_steps=1),
    )
    trainer.fit(
        model, train_dataloaders=train_data, ckpt_path="lightning_logs/version_0/checkpoints/epoch=0-step=3.ckpt"
    )


if __name__ == "__main__":
    run()

```

</details>

No.

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7 @awaelchli @akihironitta


src/pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -375,9 +379,7 @@ def convert_inf(x: Optional[Union[int, float]]) -> Optional[Union[int, float]]:
...
 
 
-def _update_n(bar: _tqdm, current: int, refresh_rate: int) -> None:
+def _update_n(bar: _tqdm, value: int) -> None:
     if not bar.disable:
-        total = bar.total
-        leftover = current % refresh_rate
-        advance = leftover if (current == total and leftover != 0) else refresh_rate
-        bar.update(advance)
+        bar.n = value
```
rohitgr7, line 384: do you think this is enough to make sure the time estimates are correctly updated?

last time I checked `update` does more than just setting the progress bar counter.
https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/tqdm/std.py#L1212-L1277

awaelchli: My understanding is that update() was never the problem, even in https://github.com/Lightning-AI/lightning/pull/12889. The problem is in resetting when a new epoch begins. The running estimate that the bar has needs to be reset, otherwise it becomes negative and it would take a few iterations for it to converge to the new value. The only way to achieve this I found is by setting initial = 0. 

[Here](https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/tqdm/std.py#L441-L445) is where the computation for the rate happens.
Right below that, the remaining time gets formatted based on the values computed in the rate, hence this is directly affected. 

Note that in your PR in #12889 the value for n is not correct after resuming. The progress bar starts at 0 even though it should start at the step where the checkpoint was saved. This is now fixed by reverting here.

Borda (approved):

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: I see. At that time, the negative estimates started coming after long training, but that fix introduced this issue where training is resumed.

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
```diff
@@ -254,12 +254,13 @@ def on_train_start(self, *_: Any) -> None:
     def on_train_epoch_start(self, trainer: "pl.Trainer", *_: Any) -> None:
         total_batches = self.total_batches_current_epoch
         self.main_progress_bar.reset(convert_inf(total_batches))
+        self.main_progress_bar.initial = 0
```
rohitgr7, line 257: maybe we should reinitialize it to`initial=reloaded value` which is `self.train_batch_idx + self._val_processed` for main progress bar.

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7, line 384: and keep this as it is? to make sure `miniters` is considered if set by user using tqdm options.
I don't know a lot about tqdm internals but manually setting just the value of `n` skips the update for `last_print_n` and `last_print_t`.

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
awaelchli: Setting it to that leads to negative values in the progress bar. 
After one epoch, these counters are outdated.

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
awaelchli: The previous way with update() would increase the bar by increments, but this is not correct when resuming with our counters. We can look for a better solution but this is the only one I found that worked.

rohitgr7: just want to make sure the old reported issues don't come up again since there's no way to test this in the CI to avoid regression. Will explore more on this.. let's merge this then for now.

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: I don't think so they will be outdated since they are reset.

rohitgr7 (approved):

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
awaelchli: Print the values and see for yourself. Doesn't get reset. Hence I can't use it as "initial value". 

image

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
awaelchli: Yes I saw the issues that the previous PR fixed. I checked [this](https://github.com/Lightning-AI/lightning/issues/12875) yesterday and it was also caused by not resetting tqdm.initial to 0.

src/pytorch_lightning/callbacks/progress/tqdm_progress.py
rohitgr7: ok.. I thought its batch_start hook since it's the case with others.. and there it's 0 after each epoch.

kaushikb11 (approved):

PR #13964, opened 2022-08-01
Fix an issue to avoid the impact of sanity check on `reload_dataloaders_every_n_epochs` for validation

### LINKED ISSUE #12145: check_val_every_n_epoch bug with list of dataloaders
## 🐛 Bug

If you have a list of dataloaders (used in a continual learning setting) and `check_val_every_n_epoch` in the trainer is not equal to 1 then `validation_step()` won't return the index of the dataloader currently being used. 

### To Reproduce

Dataloader for validation: 

```
def val_dataloader(self):
    return [DataLoader(ds, batch_size=self.batch_size, shuffle=True , num_workers=self.num_workers) for ds in self.test_datasets[:self.curr_index+1]]
```

Validation step: 
```
def validation_step(self, val_batch, batch_idx, dataloader_idx=None):
    if dataloader_idx:
        print("data loader index provided")
```

Run this in a training loop with `check_val_every_n_epoch=1` and compare to `check_val_every_n_epoch=5`. The latter will always have `dataloader_idx=None`. 

### Expected behavior

The dataloader_idx is always correctly provided independently of how often the validation is called. 

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): 1.5.4
- PyTorch Version (e.g., 1.10): 1.10.0          py3.8_cuda10.2_cudnn7.6.5_0
- Python version (e.g., 3.9): 3.8.3
- OS (e.g., Linux): Linux
- CUDA/cuDNN version: 10.1
- GPU models and configuration: v100

### Additional context

Thanks. 

cc @carmocca @justusschock @ananthsub @ninginthecloud @rohitgr7

rohitgr7:
## What does this PR do?

sanity_check was affecting the dataloader reload conditions for validation since we use the same dataloaders for validation too.

Found when working on #12145
This doesn't fix it because the behavior of `reload_dataloaders_every_n_epoch` is different for train and val. 

## Before submitting

## PR review

## Did you have fun?

Make sure you had fun coding 🙃 (painful)

cc @borda @justusschock @awaelchli @ninginthecloud @rohitgr7 @otaj @akihironitta


tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -996,17 +996,18 @@ def val_dataloader(self):
         assert train_reload_epochs == [0, 1, 2, 3, 4]
         assert val_reload_epochs == [0, 1, 2, 3, 4]
     elif n == 2:
-        assert train_reload_epochs == [0, 2, 4]
-        assert val_reload_epochs == [0, 2, 4]
+        assert train_reload_epochs == [0, 1, 3]
+        assert val_reload_epochs == [0, 1, 3]
```
awaelchli, line 1000: I'm a bit confused here. Shouldn't the sanity check also be seen in this list here, as an additional epoch 0? Perhaps we don't reload them when sanity check ran, and we just reuse them for the first real epoch. But then I don't understand why  [0, 2, 4] wasn't correct. My reasoning: Reloading every 2 epochs means skip every other. Since loading on epoch idx 0 is mandatory, we need to skip epoch idx 1 and then in epoch idx 2 we load again, and so on.

rohitgr7: > Perhaps we don't reload them when sanity check ran, and we just reuse them for the first real epoch.

yes.


I had a quick discussion with @carmocca offline regarding this.

Also: https://github.com/Lightning-AI/lightning/issues/12145#issuecomment-1196912756

but here it depends upon `check_val_every_n_epoch`. By default, it's `1`. But if you set `check_val_every_n_epoch=reload_dataloaders_every_n_epochs`, it will run at the 2nd, 4th epoch.. etc. The first `0` only runs the sanity check if `check_val_every_n_epoch=reload_dataloaders_every_n_epochs`. And that's how we have integrated other flags as well. For eg. `log_every_n_steps=2`, we log at 2nd, 4th step... etc.

Let me simplify the test and fix one case.

tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -983,67 +998,82 @@ def val_dataloader(self):
...
         # Sanity check at epoch 0 creates a validation dataloader, but validation is
         # checked (and in this case reloaded) every n epochs starting from epoch n-1
-        (3, [0, 2, 4, 6, 8], [0, 2, 5, 8]),
-        (5, [0, 2, 4, 6, 8], [0, 4, 9]),
+        (2, [0, 1, 3, 5, 7, 9], [1, 3, 5, 7, 9]),
+        (3, [0, 1, 3, 5, 7, 9], [2, 5, 8]),
+        (5, [0, 1, 3, 5, 7, 9], [4, 9]),
     ],
 )
 def test_dataloaders_load_every_n_epochs_infrequent_val(
```
rohitgr7, line 1032: I can merge this one and the above test to make it more robust... but this will create around 12 different cases. Will do it here if reviewers agree

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -399,6 +399,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed default `amp_level` for `DeepSpeedPrecisionPlugin` to `O2` ([#13897](https://github.com/PyTorchLightning/pytorch-lightning/pull/13897))
 
 
+- Fixed val dataloader reload conditions when `check_val_every_n_epochs` is set ([#13964](https://github.com/PyTorchLightning/pytorch-lightning/pull/13964))
```
rohitgr7, line 402:
```suggestion
- Fixed validation dataloader reload conditions when `check_val_every_n_epoch` is set ([#13964](https://github.com/PyTorchLightning/pytorch-lightning/pull/13964))
```

tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -983,67 +998,82 @@ def val_dataloader(self):
...
 
     trainer = Trainer(
         default_root_dir=tmpdir,
-        limit_train_batches=0.3,
-        limit_val_batches=0.3,
-        check_val_every_n_epoch=n,
+        limit_train_batches=1,
+        limit_val_batches=1,
+        check_val_every_n_epoch=check_val_every_n_epoch,
         reload_dataloaders_every_n_epochs=2,
```
rohitgr7, line 1066: weird that one is just `epoch` and another one is `epoch(s)`

tests/tests_pytorch/trainer/test_dataloaders.py
rohitgr7: I have made some updates and improved the tests to check the edge cases. The new logic aligns with your suggestion and also looks most optimal to me.

rohitgr7: but now it does not align with the rule other arguments like `log_every_n_steps` or `val_check_interval`.

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -61,13 +61,16 @@ def __init__(self, trainer: "pl.Trainer", multiple_trainloader_mode: str = "max_
     def _should_reload_train_dl(self) -> bool:
         """Check if train dataloader should be reloaded."""
         n_epochs = self.trainer.reload_dataloaders_every_n_epochs
-        return n_epochs and (self.trainer.current_epoch - self.trainer._last_train_dl_reload_epoch >= n_epochs)
```
rohitgr7, line 64: `_last_train_dl_reload_epoch` is totally useless since it doesn't do anything special

awaelchli: It was used to prevent a reload to happen twice at the same epoch. To guarantee that, for example, calling 

reset_train_dataloader()
reset_train_dataloader()

twice in a row would be a no-op for the second call.

tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -950,104 +950,87 @@ def test_dataloaders_load_only_once_no_sanity_check(tmpdir):
...
+        " train_reload_epochs_expect,val_reload_epochs_expect,val_step_epochs_expect"
+    ),
     [
-        # Sanity check at epoch 0 creates a validation dataloader, but validation is
-        # checked (and in this case reloaded) every n epochs starting from epoch n-1
-        (3, [0, 2, 4, 6, 8], [0, 2, 5, 8]),
-        (5, [0, 2, 4, 6, 8], [0, 4, 9]),
+        # general case where sanity check reloads the dataloaders for validation on current_epoch=0
+        (0, 1, 1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]),
+        (1, 1, 1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [1, 2, 3, 4, 5, 6, 7, 8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]),
```
awaelchli, line 961: Can you explain why there is one less reload when sanity check runs vs when it does not run?

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -61,13 +61,16 @@ def __init__(self, trainer: "pl.Trainer", multiple_trainloader_mode: str = "max_
...
+        return n_epochs and (self.trainer.current_epoch % n_epochs == 0)
 
     @property
     def _should_reload_val_dl(self) -> bool:
         """Check if validation dataloader should be reloaded."""
         n_epochs = self.trainer.reload_dataloaders_every_n_epochs
-        return n_epochs and (self.trainer.current_epoch - self.trainer._last_val_dl_reload_epoch >= n_epochs)
```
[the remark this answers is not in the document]
rohitgr7: > But this PR wants to reload also at the very beginning regardless

not really. It depends on the sanity check and `check_val_every_n_epoch`.

so everything is decided when `trainer.reset_val_dataloader` is called. This is called either during sanity check or during validation.

If a sanity check is happening:
- validation is enabled for the same epoch (in an ideal case 0), then `self.trainer._last_val_dl_reload_epoch` here will not be None here, and we don't have to reload the dataloaders again for that epoch.
- validation is not enabled for that epoch, then `self.trainer._last_val_dl_reload_epoch` will be None here during first validation and we have to reload the dataloaders regardless so once they are loaded, `_last_val_dl_reload_epoch` will get update inside `trainer.reset_val_dataloader`.

if sanity check is not happening, then this will be triggered only when `check_val_every_n_epoch` is satisfied, and we have to reload the dataloaders regardless, and again this will be updated during `trainer.reset_val_dataloader` for the next time.

tests/tests_pytorch/trainer/test_dataloaders.py
rohitgr7: because this is specific to just validation (without sanity check), so this ensures in case of `epoch=0`, we don't trigger `val_dataloader` again for validation.

src/pytorch_lightning/loops/dataloader/prediction_loop.py
```diff
@@ -60,7 +60,8 @@ def max_batches(self) -> List[int]:
     @property
     def dataloaders(self) -> Sequence[DataLoader]:
         """Returns all prediction dataloaders."""
-        return self.trainer.predict_dataloaders
+        dataloaders = self.trainer.predict_dataloaders
+        return [] if dataloaders is None else dataloaders
```
rohitgr7, line 64: mypy related

src/pytorch_lightning/loops/epoch/prediction_epoch_loop.py
```diff
@@ -162,8 +162,9 @@ def _get_batch_indices(self, dataloader_idx: int) -> List[List[int]]:
         """Returns a reference to the seen batch indices if the dataloader has a batch sampler wrapped by our
         :class:`~pytorch_lightning.overrides.distributed.IndexBatchSamplerWrapper`."""
         # the batch_sampler is not be defined in case of CombinedDataLoaders
+        assert self.trainer.predict_dataloaders
         batch_sampler = getattr(
-            self.trainer.predict_dataloaders[dataloader_idx],  # type: ignore[has-type]
+            self.trainer.predict_dataloaders[dataloader_idx],
```
rohitgr7, line 167: mypy related

src/pytorch_lightning/loops/fit_loop.py
```diff
@@ -210,7 +210,8 @@ def on_run_start(self) -> None:  # type: ignore[override]
 
         self.trainer.reset_train_dataloader(self.trainer.lightning_module)
         # reload the evaluation dataloaders too for proper display in the progress bar
-        self.epoch_loop.val_loop._reload_evaluation_dataloaders()
+        if self.epoch_loop._should_check_val_epoch():
+            self.epoch_loop.val_loop._reload_evaluation_dataloaders()
```
rohitgr7, line 214: reload for tqdm only if validation is supposed to happen during that epoch, else don't

src/pytorch_lightning/trainer/connectors/data_connector.py
rohitgr7: if that's the case, will revert this.

src/pytorch_lightning/trainer/connectors/data_connector.py
```diff
@@ -61,13 +61,16 @@ def __init__(self, trainer: "pl.Trainer", multiple_trainloader_mode: str = "max_
...
+        return n_epochs and (
+            self.trainer._last_val_dl_reload_epoch is None
+            or self.trainer.current_epoch - self.trainer._last_val_dl_reload_epoch >= n_epochs
+        )
```
rohitgr7, line 73: explained herE: https://github.com/Lightning-AI/lightning/pull/13964/files#r943493974

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -628,12 +628,11 @@ def _setup_on_init(self) -> None:
...
         self.num_test_batches = []
         self.num_val_batches = []
+        self.num_predict_batches = []
         self.test_dataloaders = None
         self.val_dataloaders = None
-        self._last_train_dl_reload_epoch = float("-inf")
-        self._last_val_dl_reload_epoch = float("-inf")
-
-        self.num_predict_batches = []
+        self.predict_dataloaders = None
```
rohitgr7, line 634: mypy related

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -628,12 +628,11 @@ def _setup_on_init(self) -> None:
...
+        self._last_val_dl_reload_epoch: Optional[int] = None
```
rohitgr7, line 635: we set this to None to make sure if validation is triggered, we reload the dataloaders regardless.

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1930,13 +1925,18 @@ def reset_val_dataloader(self, model: Optional["pl.LightningModule"] = None) ->
         has_step = is_overridden("validation_step", pl_module)
         enable_validation = self.limit_val_batches > 0
         if source.is_defined() and has_step and enable_validation:
+            # store epoch of dataloader reset for reload_dataloaders_every_n_epochs
+            # it should not reload again if it has already reloaded during sanity_check
+            if self.state.fn == TrainerFn.FITTING and (
+                (self.sanity_checking and self.fit_loop.epoch_loop._should_check_val_epoch())
+                or not self.sanity_checking
+            ):
+                self._last_val_dl_reload_epoch = self.current_epoch
```
rohitgr7, line 1934: if sanity check is not happening, this means this is triggered for validation and we should set this value. if sanity check is happening, we need to check if validation is going to happen in the same epoch or not. If not we should not set this value to avoid the effect of sanity check on `reload_dataloaders_every_n_epoch`, but if it will happen, we don't reload the dataloaders again and set this value to avoid the reload in the same epoch during validation.

tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -950,104 +950,87 @@ def test_dataloaders_load_only_once_no_sanity_check(tmpdir):
...
+        "num_sanity_val_steps, check_val_every_n_epoch, reload_dataloaders_every_n_epochs,"
+        " train_reload_epochs_expect,val_reload_epochs_expect,val_step_epochs_expect"
+    ),
     [
-        # Sanity check at epoch 0 creates a validation dataloader, but validation is
-        # checked (and in this case reloaded) every n epochs starting from epoch n-1
-        (3, [0, 2, 4, 6, 8], [0, 2, 5, 8]),
-        (5, [0, 2, 4, 6, 8], [0, 4, 9]),
+        # general case where sanity check reloads the dataloaders for validation on current_epoch=0
+        (0, 1, 1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]),
```
rohitgr7, line 960: note that: `val_reload_epochs_expect` and `val_step_epochs_expect` do not represent sanity check.
for sanity check it's separate within the test,

tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -950,104 +950,87 @@ def test_dataloaders_load_only_once_no_sanity_check(tmpdir):
...
+        (0, 3, 4, [0, 4, 8], [2, 8], [2, 5, 8]),
+        (1, 3, 4, [0, 4, 8], [2, 8], [2, 5, 8]),
+        # case where check_val_every_n_epoch > reload_dataloaders_every_n_epochs so expected val_reload_epoch
+        # and val_step_epoch will be same
+        (0, 2, 1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [1, 3, 5, 7, 9], [1, 3, 5, 7, 9]),
+        (1, 2, 1, [0, 1, 2, 3, 4, 5, 6, 7, 8, 9], [1, 3, 5, 7, 9], [1, 3, 5, 7, 9]),
+        (0, 3, 2, [0, 2, 4, 6, 8], [2, 5, 8], [2, 5, 8]),
+        (1, 3, 2, [0, 2, 4, 6, 8], [2, 5, 8], [2, 5, 8]),
+        (0, 5, 2, [0, 2, 4, 6, 8], [4, 9], [4, 9]),
+        (1, 5, 2, [0, 2, 4, 6, 8], [4, 9], [4, 9]),
```
rohitgr7, line 975: [this PR](https://github.com/Lightning-AI/lightning/issues/10948) introduced the link between `check_val_every_n_epoch` and `reload_dataloaders_every_n_epochs`. Here `reload_dataloaders_every_n_epochs` acts as a limit to check if this limit is crossed between consecutive validation. If `last validation epoch - current validation epoch >= reload_dataloaders_every_n_epochs`, then `check_val_every_n_epoch` takes the precedence.

src/pytorch_lightning/trainer/trainer.py
rohitgr7, line 635: this can mess up during restarting since it doesn't get saved and will be reset.. will do a follow-up to fix this

awaelchli (approved):

carmocca (approved):

Issue #13944, opened 2022-07-31
LightningFlow state increases indefinitely


belerico:
## 🐛 Bug

I don't know if this can be considered as a bug, but the LightningFlow state increases indefinitely.

### To Reproduce

```python
import subprocess
from typing import Any, Dict, Optional

import lightning as L
from pympler import asizeof
from pytorch_lightning.loggers import TensorBoardLogger


class TensorboardWork(L.LightningWork):
    """Tensorboard logger as a LightningWork

    Args:
        local_log_dir (str): where to save logs.
        **work_kwargs: additional arguments to pass to LightningWork.
    """

    def __init__(self, local_log_dir: str = "logs", **work_kwargs):
        super().__init__(**work_kwargs)
        self.local_log_dir = local_log_dir
        self._logger = TensorBoardLogger(self.local_log_dir)
        self._tensorboard_started = False

    @property
    def tensorboard_url(self) -> str:
        return self.url

    @property
    def tensorboard_log_dir(self) -> str:
        return self._logger.log_dir

    def run(self, episode_counter: int, metrics: Optional[Dict[str, Any]] = None):
        if not self._tensorboard_started:
            subprocess.Popen(
                [
                    "tensorboard",
                    "--logdir",
                    self.local_log_dir,
                    "--host",
                    self.host,
                    "--port",
                    str(self.port),
                ]
            )
            self._tensorboard_started = True
        if metrics is not None:
            self._logger.log_metrics(metrics, episode_counter)


class SimpleWork(L.LightningWork):
    def __init__(self, **work_kwargs):
        super().__init__(**work_kwargs)
        self.counter = 0

    def run(self, signal: int):
        self.counter += 1


class SimpleFlow(L.LightningFlow):
    def __init__(self):
        super().__init__()
        self.work0 = SimpleWork(parallel=True, cache_calls=True)
        self.work1 = SimpleWork(parallel=True, cache_calls=True)
        self.logger = TensorboardWork()

    def run(self):
        self.work0.run(self.work1.counter)
        if self.work0.counter > 0:
            self.work1.run(self.work0.counter)
        self.logger.run(self.work0.counter, {"State": asizeof.asizeof(self.state)})

    def configure_layout(self):
        tabs = [{"name": "Logs", "content": self.logger.tensorboard_url}]
        return tabs


if __name__ == "__main__":
    app = L.LightningApp(SimpleFlow())
```

### Expected behavior

[...]

PR #13970, opened 2022-08-01
Reduce state size

tchaton:
## What does this PR do?

This PR optimizes the size of the state. From benchmarking with the following example, the state seems to be 2 times smaller overall. 

```python
from lightning import LightningApp, LightningFlow, LightningWork
from lightning.app.runners import MultiProcessRuntime
from pympler import asizeof


class SizeWork(LightningWork):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.counter = 0

    def run(self, signal: int):
        self.counter += 1


class SizeFlow(LightningFlow):
    def __init__(self):
        super().__init__()
        self.work0 = SizeWork(parallel=True, cache_calls=True)
        self._state_sizes = {}

    def run(self):
        for idx in range(self.work0.counter + 2):
            self.work0.run(idx)

        self._state_sizes[self.work0.counter] = asizeof.asizeof(self.state)

        if self.work0.counter >= 20:
            self._exit()

if __name__ == "__main__":
    app = LightningApp(SizeFlow())
    MultiProcessRuntime(app, start_server=False).dispatch()
    print(app.root._state_sizes)
```  

```bash
python app.py
```

On master, we reached 0.1MB at counter 46.

With this PR, we reached 0.1MB at counter 95.

Parts of #13944

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7


src/lightning_app/utilities/proxies.py
```diff
@@ -372,20 +392,25 @@ def run_once(self):
...
         self.work._calls[call_hash]["statuses"].append(make_status(WorkStageStatus.RUNNING))
-        self.delta_queue.put(ComponentDelta(id=self.work_name, delta=Delta(DeepDiff(state, self.work.state))))
+        delta = Delta(DeepDiff(state, self.work.state))
+        self.delta_queue.put(ComponentDelta(id=self.work_name, delta=delta))
 
-        # 9. Start the state observer thread. It will look for state changes and send them back to the Flow
-        # The observer has to be initialized here, after the set_state call above so that the thread can start with
-        # the proper initial state of the work
-        self.state_observer.start()
+        state += delta
```
lantiga, line 402: I'm not sure I'm following. Why do we have to add `delta` to `state` here? `state` is a deep copy of `self.work.state` and `delta` was obtained by diffing `state` and `self.work.state` immediately after.

If we need to do this here (i.e. if we see deltas accumulating after 2 lines of code that we need to take into account) it probably means we need to find a different ordering for the operations.

src/lightning_app/utilities/proxies.py
```diff
@@ -372,20 +392,25 @@ def run_once(self):
...
 
         # 10. Unwrap the run method if wrapped.
         work_run = self.work.run
         if hasattr(work_run, "__wrapped__"):
             work_run = work_run.__wrapped__
 
+        # 11. Add the work state to the state observer and wait until it is ready.
+        self.state_observer._work_state.append(state)
+        while not self.state_observer._processing:
+            time.sleep(0.01)
```
lantiga, line 412: What if it is never ready? I'd add a timeout and generate an exception here.

tests/tests_app/core/test_lightning_app.py
```diff
@@ -884,3 +887,56 @@ def test_slow_flow():
...
+        for idx in range(self.work0.counter + 2):
+            self.work0.run(idx)
+
+        self._state_sizes[self.work0.counter] = asizeof.asizeof(self.state)
+
+        if self.work0.counter >= 20:
+            self._exit()
+
+
+def test_state_size_is_bounded():
```
lantiga, line 917: Not testing it's bounded, rather that it grows slowly (also see final review comment for proposal on how to make it effectively bounded or close to it).

lantiga (changes requested):
Looks good.

In general, this PR reduces the size of the state, but we could go one step further and keep the *circulating* state bounded.

If I'm not missing something, `calls` are essentially only useful for Flow, and not for Work, therefore we could avoid transmitting the whole `calls` down to Work (and up to FE).

If users want to display the status of calls, they may want to explicitly record those statuses in the circulating state from Flow.

Keeping call info in the flow only would make circulating state (which is what is causing the slowdown) considerably smaller and avoid it to grow. Then, when it's time to serialize, we would serialize call info together with Flow. This is just about avoiding to send call info around.

src/lightning_app/utilities/proxies.py
```diff
@@ -45,17 +45,13 @@ def unwrap(fn):
...
+def _send_data_to_caller_queue(work: "LightningWork", caller_queue: "BaseQueue", data: Dict, call_hash: str) -> Dict:
+
     if work._calls["latest_call_hash"] is None:
         work._calls["latest_call_hash"] = call_hash
 
     if call_hash not in work._calls:
         work._calls[call_hash] = {
-            "name": work_run.__name__,
-            "call_hash": call_hash,
-            "use_args": use_args,
```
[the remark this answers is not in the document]
tchaton: I meant to use it to track if arguments were passed, but this isn't useful and should be removed.

src/lightning_app/utilities/proxies.py
```diff
@@ -191,35 +190,51 @@ def run(self):
                 self.list.append(1)
     """
 
-    def __init__(self, work: "LightningWork", delta_queue: "BaseQueue", interval: float = 5) -> None:
+    def __init__(self, work, delta_queue: "BaseQueue", interval: float = 0.1) -> None:
```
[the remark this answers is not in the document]
tchaton: The proxy is waiting on the thread to change the processing state. I have selected a small interval not to make the tests slower, but there shouldn't be a race condition now. This is what this PR resolved. I could put 5 seconds, but the work would wait at max 10 secs (5 before, and 5 after) to ensure the thread has correctly sent its values. IMO, 1 seems a good default.

src/lightning_app/utilities/proxies.py
tchaton: We need to add the delta as we modified the state and this one is used to make a comparison at the end for the final delta. We ended up sending running status twice as this delta wasn't added.
This is mainly to avoid making another deepcopy as adding a delta is cheaper.

src/lightning_app/utilities/proxies.py
tchaton: Sounds good.

src/lightning_app/core/app.py
```diff
@@ -431,15 +431,7 @@ def _apply_restarting(self) -> bool:
...
-        work_finished_status = {}
-        for work in self.works:
-            work_finished_status[work.name] = False
-            for key in work._calls:
-                if key == "latest_call_hash":
-                    continue
-                fn_metadata = work._calls[key]
-                work_finished_status[work.name] = fn_metadata["name"] == "run" and "ret" in fn_metadata
-
+        work_finished_status = {work.name: len(work._calls) == 2 for work in self.works}
```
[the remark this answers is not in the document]
tchaton: Yes, let me add something more explicit

src/lightning_app/core/work.py
```diff
@@ -398,10 +400,10 @@ def __getattr__(self, item):
             return path
         return self.__getattribute__(item)
 
-    def _call_hash(self, fn, args, kwargs):
+    def _call_hash(self, fn, args, kwargs) -> str:
         hash_args = args[1:] if len(args) > 0 and args[0] == self else args
         call_obj = {"args": hash_args, "kwargs": kwargs}
-        return f"{fn.__name__}:{DeepHash(call_obj)[call_obj]}"
+        return str(DeepHash(call_obj)[call_obj])[:7]
```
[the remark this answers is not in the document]
tchaton: Yes, I will add a test.

tchaton:
> If I'm not missing something, `calls` are essentially only useful for Flow, and not for Work, therefore we could avoid transmitting the whole `calls` down to Work (and up to FE).

Hey @lantiga,

> If I'm not missing something, calls are essentially only useful for Flow, and not for Work, therefore we could avoid transmitting the whole calls down to Work (and up to FE).
This is already done this way right there: https://github.com/Lightning-AI/lightning/pull/13970/files#diff-5347e507417db0b69cdee291b23e64f0105cf382f743f8f2d7e60f308cf05b5dR65. I am popping the `call_hashes` before sending the state through the caller_queue. 
This made the responsiveness of the work much better compared to before.

src/lightning_app/utilities/proxies.py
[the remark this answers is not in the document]
lantiga: Great, let’s make it 1 second

src/lightning_app/utilities/proxies.py
lantiga: Sorry where do we modify the state between 394 and 399?

tests/tests_app/core/test_lightning_app.py
lantiga: Maybe a good test would also be looking at the size of the circulating state, to test that what we communicate is bounded.

lantiga:
> Hey @lantiga,
> 
> Quote: `If I'm not missing something, calls are essentially only useful for Flow, and not for Work, therefore we could avoid transmitting the whole calls down to Work (and up to FE).`
> 
> This is already done this way right there: https://github.com/Lightning-AI/lightning/pull/13970/files#diff-5347e507417db0b69cdee291b23e64f0105cf382f743f8f2d7e60f308cf05b5dR65. I am popping the `call_hashes` before sending the state through the caller_queue. This drastically increased the responsiveness of the work when I was doing my tests compared to master.

Ah, awesome. I didn’t catch that. Perfect, it would be good to add a comment and making it more prominent.

src/lightning_app/utilities/proxies.py
```diff
@@ -191,35 +190,51 @@ def run(self):
...
         super().__init__(daemon=True)
         self._work = work
         self._delta_queue = delta_queue
         self._interval = interval
         self._exit_event = Event()
         self._delta_memory = []
-        self._last_state = deepcopy(self._work.state)
```
awaelchli, line 201: What do the changes in the state observer have to do with the state size? All it should care about is sending deltas and keeping its own copy of the state updated. Would you mind breaking this out into a separate PR?

src/lightning_app/core/app.py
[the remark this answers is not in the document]
tchaton: Done !

src/lightning_app/core/work.py
```diff
@@ -457,8 +459,32 @@ def set_state(self, provided_state):
...
+
         self._changes = provided_state["changes"]
-        self._calls.update(provided_state["calls"])
+
+        if _is_flow_context():
+            if "call_hashes" not in provided_state["calls"]:
+                provided_state["calls"]["call_hashes"] = set()
+
+            for call_hash in list(provided_state["calls"]):
+                if call_hash == "latest_call_hash":
```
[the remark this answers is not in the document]
tchaton: I cleanup up a bit the code and added comments. This was the most compact format to keep track of everything.

src/lightning_app/core/work.py
```diff
@@ -457,8 +459,32 @@ def set_state(self, provided_state):
...
+                    continue
+
+                if "statuses" not in provided_state["calls"][call_hash]:
+                    continue
+
+                statuses = sorted(provided_state["calls"][call_hash]["statuses"], key=lambda x: x["timestamp"])
+
+                if statuses[-1]["stage"] == "succeeded":
+                    provided_state["calls"].pop(call_hash)
+                    provided_state["calls"]["call_hashes"].add(call_hash)
```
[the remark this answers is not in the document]
tchaton: Added documentation around it.

src/lightning_app/utilities/proxies.py
tchaton: Removed it. It wasn't required.

tests/tests_app/core/test_lightning_app.py
tchaton: Done !

src/lightning_app/utilities/proxies.py
tchaton: The StateObserver was wrongly implemented and sent duplicated deltas. I didn't manage to resolve this in an asynchronous manner, so I synchronized it with the main thread.

tchaton: Let me try to extract it to another PR.

src/lightning_app/core/work.py
```diff
@@ -216,14 +216,16 @@ def status(self) -> WorkStatus:
...
+            statuses = self._calls[call_hash].get("statuses", [])
             # deltas aren't necessarily coming in the expected order.
             statuses = sorted(statuses, key=lambda x: x["timestamp"])
             latest_status = statuses[-1]
-            if latest_status["reason"] == WorkFailureReasons.TIMEOUT:
+            if latest_status.get("reason") == WorkFailureReasons.TIMEOUT:
                 return self._aggregate_status_timeout(statuses)
             return WorkStatus(**latest_status)
+        elif call_hash in self._calls["call_hashes"]:
+            return WorkStatus(stage=WorkStageStatus.SUCCEEDED, timestamp=time.time())
```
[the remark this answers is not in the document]
tchaton: Yes, let me drop it entirely and set it to None. The timestamp is dropped to minimize the state size.

src/lightning_app/core/app.py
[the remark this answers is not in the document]
tchaton: Mind refreshing. I added some docstring and explanation.

src/lightning_app/runners/runtime.py
```diff
@@ -135,7 +135,11 @@ def dispatch(self, *args, **kwargs):
...
         if work.status.stage == WorkStageStatus.STOPPED:
             return
+
         latest_hash = work._calls["latest_call_hash"]
+
         if latest_hash is None:
             return
-        work._calls[latest_hash]["statuses"].append(make_status(WorkStageStatus.STOPPED))
+
+        if latest_hash in work._calls and "statuses" in work._calls[latest_hash]:
```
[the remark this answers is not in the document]
tchaton: Let's me clean this. Normally. no.

src/lightning_app/utilities/proxies.py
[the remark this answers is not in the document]
tchaton: Done !

src/lightning_app/core/work.py
```diff
@@ -457,8 +459,32 @@ def set_state(self, provided_state):
...
             setattr(self, k, v)
+
         self._changes = provided_state["changes"]
-        self._calls.update(provided_state["calls"])
+
+        if _is_flow_context():
+            if "call_hashes" not in provided_state["calls"]:
+                provided_state["calls"]["call_hashes"] = set()
+
+            for call_hash in list(provided_state["calls"]):
```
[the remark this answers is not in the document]
tchaton: Doen !

src/lightning_app/utilities/proxies.py
tchaton: @awaelchli Done! I am still not 100 % it is working as expected, but the Flow is trying its best to reconcile errors.

src/lightning_app/core/work.py
```diff
@@ -457,8 +482,35 @@ def set_state(self, provided_state):
...
+            # 2: Filter the statuses by timestamp
+            statuses = sorted(calls[call_hash]["statuses"], key=lambda x: x["timestamp"])
+
+            # If the latest status is succeeded, let's drop entirely
+            # the statuses collection associated to this call_hash
+            # and store the call_hash inside the `SUCCEEDED_CALL_HASHES` set.
+            # TODO: Store the statuses within a database in the future
+            if statuses[-1]["stage"] == WorkStageStatus.SUCCEEDED:
+                calls.pop(call_hash)
+                calls[CacheCallsKeys.SUCCEEDED_CALL_HASHES].add(call_hash)
```
manskx, line 511: @tchaton is this `SUCCEEDED_CALL_HASHES` getting cleaned ?

tests/tests_app/core/test_lightning_app.py
```diff
@@ -875,12 +878,44 @@ def test_slow_flow():
...
+
+        if self.work0.counter >= 20:
+            self._exit()
+
+
+def test_state_size_constant_growth():
+    app = LightningApp(SizeFlow())
+    MultiProcessRuntime(app, start_server=False).dispatch()
+    assert app.root._state_sizes[0] == 5904
+    assert app.root._state_sizes[20] <= 8120
```
manskx, line 921: The ideal test here will be to ensure the state size remains constant after thousands calls not only 20. We could mock the time.sleep() to make the test faster.

nit: it will be great to visualize it :D

src/lightning_app/utilities/proxies.py
```diff
@@ -85,15 +92,13 @@ def __post_init__(self):
...
-        use_args = len(kwargs) > 0 or (len(args) > 0 and not provided_none)
-
         self._validate_call_args(args, kwargs)
         args, kwargs = self._process_call_args(args, kwargs)
 
         call_hash = self.work._call_hash(self.work_run, args, kwargs)
-        entered = call_hash in self.work._calls
-        returned = entered and "ret" in self.work._calls[call_hash]
+        succeeded = call_hash in self.work._calls[CacheCallsKeys.SUCCEEDED_CALL_HASHES]
+        entered = call_hash in self.work._calls or succeeded
```
manskx, line 100: `self.work._calls[CacheCallsKeys.SUCCEEDED_CALL_HASHES]` should have a constant memory size. 

We want to answer the question that if we have seen this `call_hash` hash before or not, here we can use a bloom filter.

src/lightning_app/utilities/proxies.py
```diff
@@ -65,9 +67,14 @@ def _send_data_to_caller_queue(
     work._calls[call_hash]["statuses"].append(make_status(WorkStageStatus.PENDING))
 
     work_state = work.state
+    # There is no need to send the call_hashes to the work.
+    call_hashes = work_state["calls"].pop(CacheCallsKeys.SUCCEEDED_CALL_HASHES)
```
manskx, line 71: cool !!

tests/tests_app/core/test_lightning_app.py
tchaton: This is a slightly different problem.
Right now, the state stays constant assuming no other hash keys are added.

src/lightning_app/core/work.py
tchaton: No, this is the cache call hashes.

src/lightning_app/utilities/proxies.py
tchaton: This is growing as new keys can be added.

manskx: Why it should be growing ? the memory footprint of a running a lightning app should be constant. no ?

tchaton: I changed slightly the architecture. But the state is constant if nothing modifies it or no work is run with new arguments.

tests/tests_app/core/test_lightning_work.py
```diff
@@ -146,7 +146,7 @@ def run(self):
     class BlockingQueue(MockQueue):
         """A Mock for the file copier queues that keeps blocking until we want to end the thread."""
 
-        keep_blocking = True
+        keep_blocking = False
```
awaelchli, line 149: if you change that, then you may as well just go with the regular MockQueue :)

src/lightning_app/utilities/proxies.py
```diff
@@ -172,20 +179,15 @@ def sanitize(obj: Union[Path, Drive]) -> Union[Path, Dict]:
 
 class WorkStateObserver(Thread):
     """This thread runs alongside LightningWork and periodically checks for state changes.
-
     If the state changed from one interval to the next, it will compute the delta and add it to the queue which is
     connected to the Flow. This enables state changes to be captured that are not triggered through a setattr call.
-
```
awaelchli, line 178: Removing these lines will leave the docstring not rendering correctly in the html pages

awaelchli (approved):

tests/tests_app/core/test_lightning_work.py
tchaton: Yes, it is fine if the file copier terminates.

src/lightning_app/utilities/proxies.py
tchaton: Done !

manskx (approved):
This is a nice improvement 🚀 

I wouldn't close https://github.com/Lightning-AI/lightning/issues/13944 as completed yet, since the memory could still grow if calling the works with multiple arguments.

src/lightning_app/core/work.py
```diff
@@ -398,10 +419,10 @@ def __getattr__(self, item):
             return path
         return self.__getattribute__(item)
 
-    def _call_hash(self, fn, args, kwargs):
+    def _call_hash(self, fn, args, kwargs) -> str:
         hash_args = args[1:] if len(args) > 0 and args[0] == self else args
         call_obj = {"args": hash_args, "kwargs": kwargs}
-        return f"{fn.__name__}:{DeepHash(call_obj)[call_obj]}"
+        return str(DeepHash(call_obj)[call_obj])[:7]
```
lantiga, line 425: Can you document what is expected here in terms of string content, since we are only retaining the first characters?

src/lightning_app/utilities/proxies.py
```diff
@@ -394,47 +403,51 @@ def run_once(self):
...
+        # 17. DeepCopy the state and send the latest delta to the flow.
         # use the latest state as we have already sent delta
         # during its execution.
         # inform the task has completed
-        state = deepcopy(self.work.state)
+        reference_state = deepcopy(self.work.state)
         self.work._calls[call_hash]["statuses"].append(make_status(WorkStageStatus.SUCCEEDED))
         self.work._calls[call_hash]["ret"] = ret
-        self.delta_queue.put(ComponentDelta(id=self.work_name, delta=Delta(DeepDiff(state, self.work.state))))
+        self.delta_queue.put(ComponentDelta(id=self.work_name, delta=Delta(DeepDiff(reference_state, self.work.state))))
```
lantiga, line 440: I'm still failing to see why delta here would be different from zero, since we just deepcopy'ed and _calls is not part of the circulating state at this point.

lantiga (approved):
Approved. I added two comments to require further clarification.

src/lightning_app/utilities/proxies.py
tchaton: `_calls` is part of the circulating state, but only the section associated with the current request call hash.

PR #13992, opened 2022-08-02
Make MPSAccelerator platform check expect arm64

Birch-san:
Resolves https://github.com/Lightning-AI/lightning/issues/13991.

## What does this PR do?

Make MPSAccelerator's `platform.processor()` check expect arm64, which is what my M1 Max returns.

Otherwise it goes bang (see https://github.com/Lightning-AI/lightning/issues/13991).

If this code ever worked on any Mac, then this will break that Mac. I'm hoping a reviewer can reveal what device this worked on, or what testing it was originally subjected to. If it turns out it was never tested in this form, then this isn't necessarily a breaking change.

If the original "arm" platform check _does_ help in some situation, then this can be broadened to `platform.processor() in ('arm', 'arm64')`. This would be non-breaking, but my first preference is to see whether the simpler solution suffices.

## Before submitting

  - No, I ran a one-line program `Trainer(accelerator='mps')`.  
    I patched my mps.py in site-packages. Works with the patch, goes bang without it.
    I created this PR from the GitHub file editor. Haven't setup a whole dev environment for validating it.

## PR review

## Did you have fun?

It was alright


src/pytorch_lightning/accelerators/mps.py
```diff
@@ -24,7 +24,7 @@
 
 # For using the `MPSAccelerator`, user's machine should have `torch>=1.12`, Metal programming framework and
 # the ARM-based Apple Silicon processors.
-_MPS_AVAILABLE = _TORCH_GREATER_EQUAL_1_12 and torch.backends.mps.is_available() and platform.processor() == "arm"
+_MPS_AVAILABLE = _TORCH_GREATER_EQUAL_1_12 and torch.backends.mps.is_available() and platform.processor() == "arm64"
```
akihironitta, line 27: @Birch-san On my M1 MAX MBP 2021, `platform.processor()` is `"arm"`, so the current patch excludes the case. It seems the function just uses `uname -p` (https://github.com/python/cpython/blob/89f52293281b6efc4ef666ef25e677683821f4b9/Lib/platform.py#L762), so the following should work in both cases.

```suggestion
_MPS_AVAILABLE = _TORCH_GREATER_EQUAL_1_12 and torch.backends.mps.is_available() and "arm" in platform.processor()
```

akihironitta (changes requested):
Thank you for sending the fix!

src/pytorch_lightning/accelerators/mps.py
justusschock: It also returns "arm" on my machine (M1 Max MBP 2021 as well). Shouldn't the result then be consistent?

awaelchli: Returns "arm" for me. Doesn't seem to matter which Python version, I tried 3.8, 3.9, 3.10. 
I would go with `platform.processor() in ("arm", "arm64")` for now.

carmocca (approved):

src/pytorch_lightning/accelerators/mps.py
carmocca:
```suggestion
_MPS_AVAILABLE = _TORCH_GREATER_EQUAL_1_12 and torch.backends.mps.is_available() and platform.processor() in ("arm", "arm64")
```

justusschock (approved):

akihironitta (approved):

Birch-san:
thanks 🙂 yes, 

```python
platform.processor() in ("arm", "arm64")
```

seems like the right solution.

here's the reason for the discrepancy: `uname` from GNU coreutils returns `arm64`.

```
uname -p
arm64

uname --version
uname (GNU coreutils) 9.0
Packaged by https://NixOS.org
Copyright (C) 2021 Free Software Foundation, Inc.

which uname
/Users/birch/.nix-profile/bin/uname
```

If I change my `PATH` to just include binaries that ship with macOS, we get the system distribution of `uname`, which returns `arm`:

```
uname -p 
arm

which uname
/usr/bin/uname
```

PR #14002, opened 2022-08-03
(app) Add s3 drive type (1/2)

panos-is:
## What does this PR do?

~This PR adds a list of Drives to the app Work specification, this will enable apps to mount a given drive in the work and be able to access it through the file system (Using for example file system data loaders). This is an alternative way to work with drives than the existing access pattern and will later be expanded with additional options for optimizations.~

---

@rlizzo edits: This PR has been updated after feedback to basically only include the configuration for setting an s3 based Drive Type. A follow up PR will be made which allows us to inspect `Drive` objects attacked to each work attribute and then configure them based on the protocol (s3, etc). At the advice of @tchaton we have have removed the ability to use s3 based drives from the `Drive` `get/list/put/delete` method. If we want to enable this in the future, we will have a discussion on the API with @awaelchli and other interested parties. 

---

All the changes presented in this PR related to running in a cloud environment, when running locally, specifying work drives will not have any effect (However, you can mimic the cloud environment by placing of symlinking files to the same root folder).

Additionally, this PR updates the Drive spec:

1. Adding a new protocol `s3://` allowing the creation of drives directly from S3 buckets.
2. ~Adding a new, optional `optimization` field allowing users to specify optimizations for Drives when running on the cloud environment.~

This PR should be fully backwards compatible with existing code and the Work.drives field is optional. 

cc @borda

### DIFF
--- a/src/lightning_app/core/work.py
+++ b/src/lightning_app/core/work.py
@@ -52,7 +52,6 @@ class LightningWork(abc.ABC):
         cloud_build_config: Optional[BuildConfig] = None,
         cloud_compute: Optional[CloudCompute] = None,
         run_once: Optional[bool] = None,  # TODO: Remove run_once
-        drives: Optional[List[Drive]] = None,
     ):
         """LightningWork, or Work in short, is a building block for long-running jobs.
 
@@ -141,7 +140,6 @@ class LightningWork(abc.ABC):
         self._cloud_compute = cloud_compute or CloudCompute()
         self._backend: Optional[Backend] = None
         self._on_init_end()
-        self._drives = drives
 
     @property
     def url(self) -> str:
@@ -225,10 +223,6 @@ class LightningWork(abc.ABC):
 
     @property
     def cloud_compute(self) -> CloudCompute:
-        return self._cloud_compute
-
-    @property
-    def drives(self) -> List[Drive]:
         return self._cloud_compute
 
     @cloud_compute.setter
--- a/src/lightning_app/core/work.py
+++ b/src/lightning_app/core/work.py
@@ -52,6 +52,7 @@ class LightningWork(abc.ABC):
         cloud_build_config: Optional[BuildConfig] = None,
         cloud_compute: Optional[CloudCompute] = None,
         run_once: Optional[bool] = None,  # TODO: Remove run_once
+        drives: Optional[List[Drive]] = None,
     ):
         """LightningWork, or Work in short, is a building block for long-running jobs.
 
@@ -140,6 +141,7 @@ class LightningWork(abc.ABC):
         self._cloud_compute = cloud_compute or CloudCompute()
         self._backend: Optional[Backend] = None
         self._on_init_end()
+        self._drives = drives
 
     @property
     def url(self) -> str:
@@ -223,6 +225,10 @@ class LightningWork(abc.ABC):
 
     @property
     def cloud_compute(self) -> CloudCompute:
+        return self._cloud_compute
+
+    @property
+    def drives(self) -> List[Drive]:
         return self._cloud_compute
 
     @cloud_compute.setter
--- a/src/lightning_app/storage/drive.py
+++ b/src/lightning_app/storage/drive.py
@@ -51,8 +51,10 @@ class Drive:
             raise Exception(f"The id should be unique to identify your drive. Found `{self.id}`.")
 
         if optimization and optimization not in self.__PROTOCOLS_OPTIMIZATIONS__[self.protocol]:
-            raise Exception(f"This Drive protocol is not compatible with the {optimization} optimization. Available "
-                            f"optimizations for this protocol are: {self.__PROTOCOLS_OPTIMIZATIONS__[self.protocol]}")
+            raise Exception(
+                f"This Drive protocol is not compatible with the {optimization} optimization. Available "
+                f"optimizations for this protocol are: {self.__PROTOCOLS_OPTIMIZATIONS__[self.protocol]}"
+            )
 
         self.root_folder = pathlib.Path(root_folder).resolve() if root_folder else os.getcwd()
         if not os.path.isdir(self.root_folder):
--- a/src/lightning_app/storage/drive.py
+++ b/src/lightning_app/storage/drive.py
@@ -13,7 +13,8 @@ from lightning_app.utilities.component import _is_flow_context
 class Drive:
 
     __IDENTIFIER__ = "__drive__"
-    __PROTOCOLS__ = ["lit://"]
+    __PROTOCOLS__ = ["lit://", "s3://"]
+    __PROTOCOLS_OPTIMIZATIONS__ = {"lit://": [], "s3://": ["indexed-s3"]}
 
     def __init__(
         self,
@@ -21,6 +22,7 @@ class Drive:
         allow_duplicates: bool = False,
         component_name: Optional[str] = None,
         root_folder: Optional[str] = None,
+        optimization: Optional[str] = None,
     ):
         """The Drive object provides a shared space to write and read files from.
 
@@ -33,6 +35,8 @@ class Drive:
             component_name: The component name which owns this drive.
                 When not provided, it is automatically inferred by Lightning.
             root_folder: This is the folder from where the Drive perceives the data (e.g this acts as a mount dir).
+            optimization: Whether to optimize the drive with a given option. This also defines if the drive will be
+                mounted to the component and accessible via normal file system operations.
         """
         self.id = None
         for protocol in self.__PROTOCOLS__:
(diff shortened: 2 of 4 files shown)


src/lightning_app/core/work.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         cloud_build_config: Optional[BuildConfig] = None,
         cloud_compute: Optional[CloudCompute] = None,
         run_once: Optional[bool] = None,  # TODO: Remove run_once
+        drives: Optional[List[Drive]] = None,
```
tchaton, line 48: There is no need for this.

src/lightning_app/core/work.py
```diff
@@ -204,6 +206,10 @@ def cloud_build_config(self, build_config: BuildConfig) -> None:
     def cloud_compute(self) -> CloudCompute:
         return self._cloud_compute
 
+    @property
```
tchaton, line 209: Let's remove this.

src/lightning_app/core/work.py
```diff
@@ -119,6 +120,7 @@ def __init__(
         self._cloud_compute = cloud_compute or CloudCompute()
         self._backend: Optional[Backend] = None
         self._on_init_end()
+        self._drives = drives
```
tchaton, line 123: Same here.

src/lightning_app/core/work.py
tchaton: We simply need to list the drives attached to the work once the users call work.run.

src/lightning_app/storage/drive.py
```diff
@@ -46,6 +50,12 @@ def __init__(
         if "/" in self.id:
             raise Exception(f"The id should be unique to identify your drive. Found `{self.id}`.")
 
+        if optimization and optimization not in self.__PROTOCOLS_OPTIMIZATIONS__[self.protocol]:
```
tchaton, line 53: Add some tests ;)

src/lightning_app/storage/drive.py
```diff
@@ -33,6 +35,8 @@ def __init__(
             component_name: The component name which owns this drive.
                 When not provided, it is automatically inferred by Lightning.
             root_folder: This is the folder from where the Drive perceives the data (e.g this acts as a mount dir).
+            optimization: Whether to optimize the drive with a given option. This also defines if the drive will be
+                mounted to the component and accessible via normal file system operations.
```
awaelchli, line 39: Can the documentation here list what the possible values are? Or even better, have it part of the type annotation if possible. 

Another question I have is: Doesn't this leak too much infrastructure knowledge into the framework? As a user, why would I ever have optimization turned off? Can't we just have it enabled by default? If it's not just about turning thing on or off, how do I know which option to choose and when?

The docs should probably also mention that the value set here gets ignored when running locally. 
Not saying it should all be explained here in the api. Most of this can probably be clarified in user docs.

src/lightning_app/core/work.py
awaelchli: I agree. The drives is something the user manages by named attributes. Having them in a list is not useful IMO.

panos-is: Indeed. Didn't have the best read on the API, implementing the change now ;)

src/lightning_app/storage/drive.py
panos-is: It's tricky, but generally you can have multiple optimizations for a given protocol. This is not something that will happen immediately, but this is the general direction in which we're going. 

For example for `s3://` you could have `indexed-s3` (Fast, read-only, no cost), `direct-mount` (Slower, read-write, no cost) or `fsx` (Fastest, read-write, expensive) and these would all be expressed via protocol + optimization combinations. Similarly it is possible that the `indexed-s3` optimization will be made available for `lit://` drives to mount them to a work as a file system.

tchaton: @panos-is I think it would be better to abstract those within a table.

A user won't understand the difference between `indexed-s3`, etc. could be replaced 'default', 'fast', 'fastest'.

panos-is: @tchaton I don't really agree. We need to expose some of these ideas here otherwise it becomes very hard to figure out what these words really mean. We'll still have a default for each protocol, so it should be easy to utilize the base functionality without engaging with this at all.

tchaton: This would be documented within a table in the documentation and in the docstring. But I don't know what indexed-s3 means as a user. This doesn't provide value to me. But I understand optimization="fatest"

krishnakalyan3: That makes sense to me too. For example, we have something similar for our compute where abstract away the type of compute instance.

https://lightning.ai/lightning-docs/core_api/lightning_work/compute.html?highlight=compute

But ideally, we should have a lookup table within our documentation explaining what "fastest" would mean.

panos-is: I don't think this works like that, since different opts will have very different properties besides just speed (read-writedness, full or only forward sync etc) but for the time being we can just remove this and add a mount boolean flag and not have to resolve this discussion to unblock this.

src/lightning_app/core/work.py
rlizzo: removed

src/lightning_app/core/work.py
rlizzo: removed :)

src/lightning_app/core/work.py
rlizzo: removed

src/lightning_app/storage/drive.py
rlizzo: After conversing with @panos-is I devided to remove this. For now the only thing we need to know is that if an `s3://` drive is used, that will be automatically mounted. the optimization type is something we will work on later.

rlizzo:
Please see new PR description for updated changes here. I've taken over work here for @panos-is now that he is on-call and we brought this PR in a slightly different direction.

tchaton (approved):
LGTM !

rlizzo (approved):

lantiga (commented):
I don't think we should use the protocol (s3:// vs lit://) as an indication on how the drive should be mounted.
We should instead have an explicit flag for mounting, so we'll be able to mount non-s3:// locations (e.g. lit:// themselves, or GCP storage).

panos-is:
> I don't think we should use the protocol (s3:// vs lit://) as an indication on how the drive should be mounted. We should instead have an explicit flag for mounting, so we'll be able to mount non-s3:// locations (e.g. lit:// themselves, or GCP storage).

Eventually yes, but there's no particular reason to frontload this work here. Lets add these features when we get to them instead of adding a mount flag but only supporting `lit+false` and `s3+true`.

rlizzo:
@lantiga I've been going back and forth here, but I think I agree with @panos-is. When we have additional drive types and options it'll make sense to introduce more flags, but for right now the proposed solution is simple to explain/understand & doesn't require us to introduce breaking changes in the near future when we have a demonstrated need for additional argument / behaviors.

lantiga:
My need is about the now though (while designing the training app): how can we mount a lit:// drive? Should we surface its underlying s3 bucket?

Here's the use case: I use a notebook to download a dataset from somewhere (not s3) and I use a lit:// Drive to store the data. Then I want to mount that data to a Work. How do I do that?

Same goes with: I have a data-generating process that puts data into a lit:// Drive. Then I want to mount the Drive into a Work to run a training job.

Let's try to figure out a way to enable this because it's going to be a pretty common use case once we roll out the app for training.

src/lightning_app/storage/drive.py
```diff
@@ -142,6 +163,10 @@ def get(
         """
         if _is_flow_context():
             raise Exception("The flow isn't allowed to get files from a Drive.")
+        if self.protocol == "s3://":
+            raise PermissionError(
+                "S3 based drives cannot currently get files via this API. Did you mean to use `lit://` drives?"
+            )
```
awaelchli, line 169: @tchaton @rlizzo Why/how would one use a drive if all these basic operations are not supported. Can't put files and can't get files, what's the point xD

awaelchli (approved):

src/lightning_app/storage/drive.py
lantiga: This Drive type is only for mounting S3 buckets on the Work.

The original direction was to have the ability to mount drives on Work RO. Then to also allow users to mount S3 locations to Work RO as well.

My general disagreement is with calling them "S3 drives", I think it's an antipattern and it won't stand the test of time really quickly. I have already expressed it in a platform PR.

panos-is: These are mounted drives that will be visible as a file system at the `root_folder` location of the work. You can interact with the file system normally via any means with which you'd interact with files on the local machine. (ie. using dataloaders etc)

lantiga:
Any feedback on the strategy to enable mounting `lit://` drives RO from an API perspective?

A cheap way (in a future PR) would be in this case to snapshot a `lit://`drive to an `s3://` bucket and return the bucket location, so 
- it becomes immutable, so no problem if the `lit://` drive keeps changing if another Work writes to it, and we can safely mount it RO
- we have a `s3://` location we can use for mounting, aka  we can rely on the protocol to mount it (although I 100% do not like this approach - ok I've made my point)

All the above can happen behind the scenes, with the obvious disadvantage that we duplicate data, but that can be a clear caveat.

WDYT @panos-is @rlizzo

manskx (approved):

Borda (approved):

PR #13872, opened 2022-07-27
adding explain notes for requirements

Borda:
## What does this PR do?

Add clarification/explanation to each requirement file so we keep it in mind

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @rohitgr7

PR #14007, opened 2022-08-03
freeze requirements

Borda:
## What does this PR do?

Freeze some additional package requirements for the CI jobs
some details are in #13872
follow-up of #13998

## Before submitting

## PR review

## Did you have fun?


awaelchli (approved):
This looks great! So for apps we now have the same mechanism in setup.py for unpinning the versions?

Borda:
> This looks great! So for apps we now have the same mechanism in setup.py for unpinning the versions?

I believe we had it all the time repository-wide; we did not set the upper bound for the bot to update it...
https://github.com/Lightning-AI/lightning/blob/5f1cb913bb4cb9d3aac4ed27c4cab57e7a45a7d4/.github/dependabot.yml#L9

akihironitta (approved):

.github/workflows/code-checks.yml
```diff
@@ -34,7 +34,8 @@ jobs:
       run: |
         pip install torch==1.11 --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
         python ./requirements/pytorch/adjust-versions.py requirements/pytorch/extra.txt
-        pip install -r requirements/pytorch/devel.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
+        pip install -e . --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
```
carmocca, line 37: I think this change should be excluded from this PR as it changes the mypy configuration, hence the mypy failures.

It can be a follow-up

.github/workflows/code-checks.yml
```diff
@@ -34,7 +34,8 @@ jobs:
...
+        pip install "mypy==0.920"
```
carmocca, line 38: This shouldn't be pinned.

requirements/pytorch/test.txt
```diff
@@ -1,18 +1,16 @@
-coverage>=6.4
-codecov>=2.1
-pytest>=7.0
-pytest-cov
-pytest-forked
```
carmocca, line 5: @akihironitta Removed the pins a while back because we basically always want to use latest for these dependencies. I would keep it that way.

requirements/pytorch/test.txt
```diff
@@ -1,18 +1,16 @@
...
+coverage>=6.4, <=6.4.2
+codecov>=2.1, <=2.1.12
+pytest>=7.0, <=7.1.2
+pytest-cov <=3.0.0
+pytest-forked <=1.4.0
 pytest-rerunfailures>=10.2
-mypy>=0.920
```
carmocca, line 7: Why remove `mypy` from here? Having it here means dependabot will try to upgrade it. Also, it's useful for contributors who just install all test requirements.

carmocca (changes requested):

.github/workflows/code-checks.yml
Borda: Why not pin? Some versions may have a different interpretation, and we experienced it past...

.github/workflows/code-checks.yml
Borda:
```suggestion
        # todo: adjust requirements for both code-bases
        pip install -r requirements/pytorch/devel.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
```

requirements/pytorch/test.txt
Borda: Yes, `mypy` is used in a specific case and for such you do not need to install all testing requirements such as pytest with all plugins...

requirements/pytorch/test.txt
Borda: Well, but let's do it in a controlled way as we have now dependable so it sends updated...
otherwise you can say pretty much the same about all our dependencies that we want to use always the latest :rabbit:

.github/workflows/code-checks.yml
akihironitta: I think it should be pinned within one of the requirements files. Otherwise, it will lead to different versions between local dev envs and our CI.

requirements/pytorch/test.txt
carmocca: If mypy and flake8 are part of pre-commit and pre-commit is part of a CI job, I consider them testing requirements.

You don't want contributors to have to look inside the CI jobs to know what to install. IMO this fits here.

Additionally, there's the benefit of getting dependabot upgrades.

.github/workflows/code-checks.yml
```diff
@@ -34,7 +34,9 @@ jobs:
       run: |
         pip install torch==1.11 --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
         python ./requirements/pytorch/adjust-versions.py requirements/pytorch/extra.txt
+        # todo: adjust requirements for both code-bases
         pip install -r requirements/pytorch/devel.txt --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
+        pip install "mypy==0.920"
```
carmocca, line 39: The correct pin in master is:

```suggestion
        pip install "mypy==0.971"
```

requirements/pytorch/test.txt
Borda: yes, `flake8` is part of pre-commit, but Mypy not as if you run all files, it takes too much time, but in the normal scenario, it runs only on changed files, so shall we consider adding back `mypy` as a pre-commit hook?

carmocca: Sorry, my mistake.

`mypy` is not part of pre-commit, but still part of CI, so the testing requirements should also install it in my view

carmocca (approved):

PR #14049, opened 2022-08-05
Replace Slack url everywhere

edenlightning:
cc @borda @rohitgr7


yurijmikhalevich (approved):

rohitgr7:
qq: why is this done? I remember this was internally discussed to avoid incorrect redirects in case the URL expires.

awaelchli (approved):
If it is absolutely necessary, but I think it is still a step back. The links will get outdated quickly as we know from experience. 

If we had lightning.ai/slack or lightning.ai/community redirecting to the slack invite, that would be gold

Borda:
@edenlightning mind share reason, this link is permanent or will also expire in a few weeks?

Issue #13672, opened 2022-07-15
[Epic] Make Lightning App fully Serverless


tchaton:
## 🚀 Feature

In order to optimize cost at scale, it is important Lightning App becomes fully serverless.

Terminology: ``Serverless`` means there is a cost occurring when nothing is happing. This movement democratized by the cloud is known as ``Pay as you Go`` Services.

Making Apps Serverless is challenging but would enable more complex systems to be built at the fraction of the cost.

 ## Actions Items:
 
 ### Trainer, LightningModule
 - [] Implement Last Minute PL Fault Tolerance: This feature enables Fault Tolerance to be enabled last minute once receiving a SIGTERM signal, collecting all metadata, applying 1 extra baches and stop after generating an auto_restart checkpoints    

 
 ### Flow, Work
- [] Auto Restart: Works should be restartable. Introduce the self.restarting and checkpointing logic
- [] Add support for a PyTorchLightningPythonScript Component.
- [] Enable Flow portability. The flow should be restartable from one process to another, one machine to another
- [] Frontends Load Balancing: Frontend should be turned down when there aren't any incoming requests and restarted otherwise.
- [] State Arching: The state can grow quickly as an app is running and dynamically creating resources. The partial part of the state could be archived when not accessed for some time.
- [] Introduce Work Deletion: When work is deleted, resources should be stopped and the state removed from the App.  

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/Lightning-AI/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/Lightning-AI/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/Lightning-AI/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

PR #14076, opened 2022-08-07
(app) Run the flow only if the state has updated 1/2

tchaton:
## What does this PR do?

This PR introduces a breaking change where the flow will execute only if the state changes. This is required in order to move toward a serverless expression of the flow. This still triggers if the flow modifies its own state which we might want to disable in the future, but I believe this is required for now. 

Yes, this used to print infinitely on master, but now this will print only once as no state was modified. 

```python
class Flow(LightningFlow):
    def run(self):
        logger.info("Hello World")

app = LightningApp(Flow())
```

Parts of #13672

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/core/app.py
```diff
@@ -307,7 +308,10 @@ def maybe_apply_changes(self) -> bool:
         if not deltas:
             # When no deltas are received from the Rest API or work queues,
             # we need to check if the flow modified the state and populate changes.
-            if Delta(DeepDiff(self.last_state, self.state, verbose_level=2)).to_dict():
+            delta = Delta(DeepDiff(self.last_state, self.state, verbose_level=2))
+            difference = delta.to_dict()
```
lantiga, line 312: I would rework as:
```python
if not deltas:
    if DeepDiff(self.last_state, self.state, verbose_level=2):
        self.set_state(self.state)
        self._has_updated = True

    return False
```
since `DeepDiff` is truthy when there are differences and falsey otherwise.

tests/tests_app/core/test_lightning_flow.py
```diff
@@ -604,6 +605,7 @@ def run(self):
                 # TODO (tchaton) Optimize flow execution.
                 assert 4.0 < abs(time() - self._last_time) < 6.0
                 self._exit()
+        self.counter += 1
```
lantiga, line 608: I'm not sure why this is needed. The `SchedulerThread` is not reliant on the event loop iterating, so this should happen without state being forced to update.

Can you clarify @tchaton?

lantiga (changes requested):
Great that we're getting to this, really important.

It checks out, I've mostly got a question around the scheduler test.

src/lightning_app/core/app.py
```diff
@@ -352,17 +356,24 @@ def run_once(self):
...
+
         try:
             self.check_error_queue()
-            t0 = time()
-            self.root.run()
-            self._last_run_time = time() - t0
+            # Execute the flow only if:
+            # - There are state changes
+            # - It is the first execution of the flow
+            if self._has_updated or not self.first_execution:
```
hhsecond, line 366: why is there a negation on the "first execution" check? We shouldn't execute if it's not a first execution (and has_updated is False) right?
```
if self._has_updated or self.first_execution
```

src/lightning_app/core/app.py
```diff
@@ -352,17 +356,24 @@ def run_once(self):
...
+                self.root.run()
         except CacheMissException:
             self._on_cache_miss_exception()
         except (ExitAppException, KeyboardInterrupt):
             done = True
             self.stage = AppStage.STOPPING
 
+        self._last_run_time = time() - t0
+        self.first_execution = True
```
awaelchli, line 375: Why is this a public attribute and why do we need it? Wouldn't it be sufficient set `self._has_updated = True` in the beginning. If running is defined by state updates, we could say that the very first run is triggered by a state update.

src/lightning_app/core/app.py
tchaton: Done !

tests/tests_app/core/test_lightning_flow.py
tchaton: There were a bug in `_delta_to_app_state_delta`. Removed the counter.

src/lightning_app/core/app.py
tchaton: Switched off. Thanks.

src/lightning_app/utilities/app_helpers.py
```diff
@@ -322,8 +322,9 @@ def _delta_to_appstate_delta(root: "LightningFlow", component: "Component", delt
 
             delta_key_without_root = delta_key[4:]  # the first 4 chars are the word 'root', strip it
             new_key = new_prefix + delta_key_without_root
-            changed[new_key] = val
-            del changed[delta_key]
+            if new_key != delta_key:
```
tchaton, line 325: @lantiga This was the bug with the scheduler. The signal wasn't received as the entry was deleted.

awaelchli: I suggest to create a bugfix PR with these changes, to keep this PR atomic.

awaelchli (approved):

src/lightning_app/utilities/scheduler.py
```diff
@@ -15,7 +15,7 @@ class SchedulerThread(threading.Thread):
     def __init__(self, app) -> None:
         super().__init__(daemon=True)
         self._exit_event = threading.Event()
-        self._sleep_time = 0.5
+        self._sleep_time = 0.05
```
lantiga, line 18: This puts extra pressure on the flow. I think it's ok to have a scheduler with a granularity in the order of magnitude of a second.
Maybe we can have a flag that shortens this during testing to speed it up if that is the need?

tests/tests_app/core/test_lightning_app.py
```diff
@@ -920,3 +929,29 @@ def test_state_size_constant_growth():
...
+
+    def on_after_run_once(self):
+        self.counter += 1
+        if not self._has_updated and self.counter > 2:
+            return True
+        return super().on_after_run_once()
+
+
+def test_non_updated_flow(caplog):
+    """This tests validate the app can run 3 times and call the flow only once."""
```
lantiga, line 952: Nice

lantiga (commented):
Just a comment on the scheduler frequency. We should keep it at a second granularity (so 0.5 timeout is fine) and document it, and just patch it during testing to make tests faster.

src/lightning_app/core/app.py
```diff
@@ -307,9 +307,11 @@ def maybe_apply_changes(self) -> bool:
         if not deltas:
             # When no deltas are received from the Rest API or work queues,
             # we need to check if the flow modified the state and populate changes.
-            if Delta(DeepDiff(self.last_state, self.state, verbose_level=2)).to_dict():
+            deep_diff = DeepDiff(self.last_state, self.state, verbose_level=2)
```
manskx, line 310: Database updates won't trigger deltas?

manskx (approved):

src/lightning_app/core/app.py
lantiga: You mean in the future when we add support for databases?

manskx: Yes

lantiga: Good question :-) Possibly yes but it depends on the shape things take:
- if we have a database as a work component, then I would say it's up to the work to trigger an update by modifying a state variable when a transaction happens (or when it makes sense)
- if we have a database-backed state, then things definitely get more interesting, because the database is itself (part of) the state, so it should trigger a rerun automatically

src/lightning_app/utilities/scheduler.py
tchaton: Done !

lantiga (approved):

PR #14100, opened 2022-08-08
(app): Add load_state_dict and state_dict

tchaton:
## What does this PR do?

This PR introduces an API to reload dynamic component trees.

The current strategy is the following. We collect all the states. On reload, if a component state doesn't have its component, we give the responsibility to its closest parent to resolve it and then the cascade of instantiation is done. 

```python
class FlowReload(LightningFlow):

    def run(self):
        if not getattr(self, "w", None):
            self.w = WorkReload()
        self.w.run()

    def on_load_state_dict(self, children_states, extras) -> None:
        self.w = WorkReload()
```

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/core/flow.py
```diff
@@ -309,7 +309,7 @@ def get_all_children(self):
         self.get_all_children_(children)
         return children
 
-    def set_state(self, provided_state: Dict) -> None:
+    def set_state(self, provided_state: Dict, recurse: bool = True) -> None:
```
manskx, line 312: nit: can we add a docstring to explain the purpose of this flag

src/lightning_app/utilities/app_helpers.py
```diff
@@ -401,3 +401,116 @@ def default(self, obj: Any) -> Any:
...
+        elif isinstance(child, LightningWork):
+            # 8: Set the state on the work.
+            child.set_state(state)
+
+        else:
+            # 9: Collect any error messages.
+            error_messages.append((flow.name, child_name))
+
+
+def _load_state_dict(root_flow: "LightningFlow", state: Dict[str, Any], strict: bool = True) -> None:
```
manskx, line 466: where this is/will be called ?

src/lightning_app/core/flow.py
```diff
@@ -634,3 +636,10 @@ def my_remote_method(self, name):
             lightning my_command_name --args name=my_own_name
         """
         raise NotImplementedError
+
+    def on_save_state_dict(self) -> Optional[Dict[str, Any]]:
+        """Override to return any required metadata for reloading."""
+        return None
+
+    def on_load_state_dict(self, children_states: Dict[str, Any], extras: Optional[Dict[str, Any]] = None) -> None:
+        """Override to add your re-instantiation logic for dynamic components."""
```
manskx, line 645: I think we can find better names for these methods. I can't think of a name aTM :D

manskx (approved):

src/lightning_app/utilities/app_helpers.py
tchaton: This will be called in the runtime after setting up the app.

src/lightning_app/core/flow.py
```diff
@@ -634,3 +636,10 @@ def my_remote_method(self, name):
             lightning my_command_name --args name=my_own_name
         """
         raise NotImplementedError
+
+    def on_save_state_dict(self) -> Optional[Dict[str, Any]]:
+        """Override to return any required metadata for reloading."""
+        return None
```
awaelchli, line 642: @tchaton it is not anymore what we discussed. We can't name it `on_x` when it is required to return state. We need to call it something special, like `extra_state` or whatever to emphasize this is additional data not already captured in the attributes being saved.

src/lightning_app/core/flow.py
awaelchli: I think the docstring should explicitly say that here is the place to re-instantiate dynamic components. 
And it should also show how you are supposed to call super

src/lightning_app/core/flow.py
lantiga: It's not really `extra_state` IMO, this one needs to return the last resumable state (which may be the last, or not the last, at least not for work, depending on how the specific logic of the work).

I'm wondering what kind of metadata we need to include that we don't consider being in state. As in, state should be defined as what is to resume.

src/lightning_app/core/work.py
```diff
@@ -540,10 +540,6 @@ def _aggregate_status_timeout(self, statuses: List[Dict]) -> WorkStatus:
         status = {**timeout_statuses[-1], "timestamp": statuses[0]["timestamp"]}
         return WorkStatus(**status, count=len(timeout_statuses))
 
-    def load_state_dict(self, state):
```
lantiga, line 543: For confirmation: are we removing reloading from Work and expect this to only happen in the parent Flow?

What I'm concerned about is not really the loading, but the serializing. The work will need a way to produce its closest resumable state, which I'm expecting to be specific to the logic of the work itself in the general case.

src/lightning_app/utilities/app_helpers.py
```diff
@@ -401,3 +401,116 @@ def default(self, obj: Any) -> Any:
...
+
+def _state_dict(flow: "LightningFlow"):
+    state = {}
+    flows = [flow] + list(flow.flows.values())
+    for f in flows:
+        state[f.name] = f.state
+        extras = f.on_save_state_dict()
+        if extras:
+            state[f.name].update({"extras": extras})
+    for w in flow.works():
```
lantiga, line 414: Capturing state variables from work will not ensure that the state the work is exposing is resumable.

lantiga (commented):
A few comments on capturing a "resumable state" from Work.

src/lightning_app/core/work.py
tchaton: The plan is to have a `persist` and `restore` methods for the works as the state isn't fully part of the captured and a start_restart="latest|run_start". On reload, the runtime will select the right state and reload them.

lantiga: Ok, agreed

src/lightning_app/utilities/app_helpers.py
tchaton: Yes, this isn't the final code. Right now, this is mainly to ensure the logic works within the test. In the future, we would use the state of the work directly, not the one captured in the flow.

lantiga: Got it

src/lightning_app/utilities/app_helpers.py
```diff
@@ -402,3 +402,66 @@ def default(self, obj: Any) -> Any:
         if callable(getattr(obj, "__json__", None)):
             return obj.__json__()
         return json.JSONEncoder.default(self, obj)
+
+
+def _state_dict(flow: "LightningFlow"):
```
carmocca, line 407: I don't understand the use of these 2 helpers.

Which component calls these?
Why is this logic not part of `flow.state_dict`?

src/lightning_app/utilities/app_helpers.py
```diff
@@ -402,3 +402,66 @@ def default(self, obj: Any) -> Any:
...
+
+                    dynamic_children_state[flow_name].update({name.replace(parent_name + ".", ""): component_state})
+                    has_matched = True
+                    break
+            if has_matched:
+                break
+
+    # 5: Reload the flow states
+    for flow_name, flow in flow_map.items():
+        flow.load_state_dict(state.pop(flow_name), dynamic_children_state[flow_name], strict=strict)
```
ethanwharris, line 458: It may be that the flow in flow map does not have any `dynamic_children_state` right? So perhaps this should be:
```suggestion
        flow.load_state_dict(state.pop(flow_name), dynamic_children_state.get(flow_name, {}), strict=strict)
```

ethanwharris (approved):
Missing some context here as I guess this won't become usable yet as we don't have a pause / reload mechanism (maybe that could be reflected in the docstrings of the methods?).

Other than that the approach seems good, although you get some duplication by having to instantiate dynamic components in multiple places

src/lightning_app/core/flow.py
```diff
@@ -634,3 +643,81 @@ def my_remote_method(self, name):
             lightning my_command_name --args name=my_own_name
         """
         raise NotImplementedError
+
+    def state_dict(self):
+        """Returns the current flow state but not its children."""
```
awaelchli, line 648: I wouldn't use the same names as PyTorch. 

1. They don't do the same. Here the children are explicitly excluded
2. In PyTorch normally you don't override the state_dict and load_state_dict methods directly, instead you override [get_extra_state](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.set_extra_state) and set_extra_state. Here, it is the opposite: You are supposed to override them but not call them directly. 

For these reasons, I suggest to call them a different name. For example: `state_without_children` and `load_state`.

Borda:
@tchaton, mind checking the remaining comments?

PR #14102, opened 2022-08-08
Fix mypy errors attributed to `pytorch_lightning. strategies.sharded_spawn`

krishnakalyan3:
Part of https://github.com/Lightning-AI/lightning/issues/13445


rohitgr7 (approved):

otaj (approved):

Borda (approved):

src/pytorch_lightning/strategies/sharded_spawn.py
```diff
@@ -93,7 +96,7 @@ def block_backward_sync(self) -> Generator:
             yield None
 
     @rank_zero_only
-    def _optim_state_dict(self, optimizer):
+    def _optim_state_dict(self, optimizer: Optimizer) -> Dict[str, Tensor]:
```
awaelchli, line 99: [State dict ](https://github.com/pytorch/pytorch/blob/9213751970e4bb4eecd0b5dc6317740f017460a0/torch/optim/optimizer.py#L135)does not return a dict with tensors. 
The documentation [says](https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.state_dict.html#torch.optim.Optimizer.state_dict) "differs between optimizer classes." 
In their [type file](https://github.com/pytorch/pytorch/blob/master/torch/optim/optimizer.pyi), they annotate as Dict[str, Any]. 

Finally, also in the [fairscale OSS optimizer](https://fairscale.readthedocs.io/en/stable/api/optim/oss.html?highlight=OSS#fairscale.optim.OSS), which is used here concretely, they annotate as Dict[str, Any]. 


This is probably the best we can do.

awaelchli, line 99:
```suggestion
    def _optim_state_dict(self, optimizer: Optimizer) -> Dict[str, Any]:
```

src/pytorch_lightning/strategies/sharded_spawn.py
```diff
@@ -69,12 +71,13 @@ def _reinit_optimizers_with_oss(self, optimizers: List[Optimizer]) -> List["OSS"
...
     def _wrap_optimizers(self, optimizers: List[Optimizer]) -> List["OSS"]:
-        if self.model is not None and self.model.trainer.state.fn != TrainerFn.FITTING:
+        assert self.lightning_module
+        if self.model is not None and self.lightning_module.trainer.state.fn != TrainerFn.FITTING:
             return optimizers
 
         return self._reinit_optimizers_with_oss(optimizers)
 
-    def optimizer_state(self, optimizer: "OSS") -> Optional[dict]:
+    def optimizer_state(self, optimizer: "OSS") -> Dict[str, Tensor]:
```
awaelchli, line 80:
```suggestion
    def optimizer_state(self, optimizer: "OSS") -> Dict[str, Any]:
```

src/pytorch_lightning/overrides/base.py
```diff
@@ -54,7 +54,7 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
 
 
 class _LightningModuleWrapperBase(DeviceDtypeModuleMixin, torch.nn.Module):
-    def __init__(self, pl_module: Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase]) -> None:
+    def __init__(self, pl_module: Union["pl.LightningModule", _LightningPrecisionModuleWrapperBase, nn.Module]) -> None:
```
awaelchli, line 57: Adding nn.Module here is not appropriate. I will take a look

src/pytorch_lightning/strategies/sharded_spawn.py
krishnakalyan3: Thanks for the suggestions. That makes sense.

awaelchli (approved):

PR #14145, opened 2022-08-10
Update defaults for WandbLogger's run name and project name

### LINKED ISSUE #14054: WandbLogger : use random name instead of project as default name
## Proposed refactor
Use the wandb behavior (creating a random name) if no name is passed for the run or just let it `None` and wandb will deal with creating the name. 

### Motivation
I got 20 runs with all the same name which is not easy to navigate through 

### Pitch
Use a random name or name with a date or anything that would distinguish two runs. 

### Additional context
It's here: https://github.com/Lightning-AI/lightning/blob/9d02ad761c6c9360cba96e339d340c0cc2e572a8/src/pytorch_lightning/loggers/wandb.py#L300

cc @tchaton @rohitgr7 @awaelchli @morganmcg1 @borisdayma @scottire @manangoel99

awaelchli:
## What does this PR do?

Fixes #14054

Script to test the behavior if you want to play with it:
```py
import os

import torch

from pytorch_lightning.loggers import WandbLogger


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        print(self.trainer.logger.save_dir)
        print(self.trainer.logger.name)
        print(self.trainer.logger.version)

def run():

    model = BoringModel()
    logger = WandbLogger(project="this")
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        enable_model_summary=False,
        logger=logger,
    )


if __name__ == "__main__":
    run()

```

Folder structure:
image

Wandb online record:
image

Yes, but they are necessary to satisfy the users. 
- name: used to return the wandb.Run.name, but now returns the project name.
- Since name can't be None, we had to use a default name for the project name, and setting it to "lightning_logs" makes it more consistent with the other loggers. 

## Before submitting

## PR review

## Did you have fun?

cc @tchaton @rohitgr7 @borda @justusschock @awaelchli @morganmcg1 @borisdayma @scottire @manangoel99 @akihironitta


src/pytorch_lightning/loggers/wandb.py
```diff
@@ -450,13 +451,13 @@ def save_dir(self) -> Optional[str]:
 
     @property
     def name(self) -> Optional[str]:
```
rohitgr7, line 453: I think it's a little off now since `name` argument and property are different. Instead, should we add/rename `project_name`?

rohitgr7 (approved):

src/pytorch_lightning/loggers/wandb.py
awaelchli: I can't fix it in another way. The name here is tied to what is being required by the checkpointing logic, as explained in #14188. 

The only other way would be to hack an "isinstance check" in the ModelCheckpoint class and change the behavior for just wandb, but considering that #14188 is looking promising, I want to avoid this at all cost. 

Besides, it is also the same for the default logger, TensorBorad, where name does not represent the individual run's name, but the name of the entire "project'.

Borda (approved):

src/pytorch_lightning/loggers/wandb.py
rohitgr7: #14188 looks good, but considering the logger individually, this is a little inconsistent since we have both `name` and `project_name` as arguments. I think we should have both these properties,  or which ever is required after #14188.

> Besides, it is also the same for the default logger, TensorBorad, where name does not represent the individual run's name, but the name of the entire "project'.

a wandb user might not care about other loggers at all.

awaelchli: There will always be terminology in the 3rd party apis that does not mean the same thing in PL. The name property is on the base interface and I can't change it in just one logger. The discussion needs to be offloaded, I cannot address this here. If you want to add a project name property to the WandbLogger, feel free to do that. But at the moment, the implementation of "name" can't change, not until we have https://github.com/Lightning-AI/lightning/issues/14188

rohitgr7: will create an issue to discuss this.

justusschock (approved):

awaelchli:
gatekeeper rerun

PR #14151, opened 2022-08-10
Fix saving hyperparameters in a composition where parent is not a LM or LDM

### LINKED ISSUE #13181: Unexpected behavior of .save_hyperparameters() for same parameter name

awaelchli:
## What does this PR do?

Fixes #13181
Closes: #13202

Look at this code:

```py
    class Child(LightningModule):
        def __init__(self, same_arg):
            super().__init__()
            self.save_hyperparameters()

    class Parent: 
        def __init__(self, same_arg="parent_default", other_arg="other"):
            self.child = Child(same_arg="cocofruit")

    parent = Parent()
    print(parent.child.hparams)
```

Note: There is a LightningModule composed inside of a non-LightningModule class. 

**Q: At the call `self.save_hyperparameters()`, which parameters do we excpect to get saved?**
A: Only the ones from the child LightningModule. But on master, it collects also the ones from the parent class. 

**Q: But Adrian, collecting more parameters is always better. The more the better!**
A: No. 1) You could coincidentally have two arguments with the same name in two different classes. So there is a collision and in master, one overwrites the other. Bad! 2) We save hyperparameters not just for the sake of saving them. We want to save exactly the ones needed to reconstruct the object when loading `ChildInComposition.load_from_checkpoint`. No more, no less.

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @ninginthecloud @rohitgr7 @otaj @akihironitta

### DIFF
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -70,7 +70,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed epoch-end logging results not being reset after the end of the epoch ([#14061](https://github.com/Lightning-AI/lightning/pull/14061))
 
 
-- Fix saving hyperparameters in a composition where the parent class is not a `LightningModule` or `LightningDataModule` ([#14151](https://github.com/Lightning-AI/lightning/pull/14151))
+- Fixed saving hyperparameters in a composition where the parent class is not a `LightningModule` or `LightningDataModule` ([#14151](https://github.com/Lightning-AI/lightning/pull/14151))
 
 
 
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -70,6 +70,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed epoch-end logging results not being reset after the end of the epoch ([#14061](https://github.com/Lightning-AI/lightning/pull/14061))
 
 
+- Fix saving hyperparameters in a composition where the parent class is not a `LightningModule` or `LightningDataModule` ([#14151](https://github.com/Lightning-AI/lightning/pull/14151))
+
+
+
 ## [1.7.1] - 2022-08-09
 
 ### Fixed
--- a/src/pytorch_lightning/utilities/parsing.py
+++ b/src/pytorch_lightning/utilities/parsing.py
@@ -162,7 +162,10 @@ def get_init_args(frame: types.FrameType) -> Dict[str, Any]:
 
 
 def collect_init_args(
-    frame: types.FrameType, path_args: List[Dict[str, Any]], inside: bool = False
+    frame: types.FrameType,
+    path_args: List[Dict[str, Any]],
+    inside: bool = False,
+    classes: Tuple[Type, ...] = (object,),
 ) -> List[Dict[str, Any]]:
     """Recursively collects the arguments passed to the child constructors in the inheritance tree.
 
@@ -170,6 +173,7 @@ def collect_init_args(
         frame: the current stack frame
         path_args: a list of dictionaries containing the constructor args in all parent classes
         inside: track if we are inside inheritance path, avoid terminating too soon
+        classes: the classes in which to inspect the frames
 
     Return:
           A list of dictionaries where each dictionary contains the arguments passed to the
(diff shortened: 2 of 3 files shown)


src/pytorch_lightning/utilities/parsing.py
```diff
@@ -181,13 +185,13 @@ def collect_init_args(
     if not isinstance(frame.f_back, types.FrameType):
         return path_args
 
-    if "__class__" in local_vars:
+    if "__class__" in local_vars and issubclass(local_vars["__class__"], classes):
```
awaelchli, line 188: Reviewers: This function `collect_init_args` is recursive, it will go up the stack frames until it hits a class's `__init__`. But this init could be inside of another init frame. On master, we would just continue iterating. Here, I'm making sure that we are only doing shenanigans when the init belongs to a subclass of LM or LDM.

src/pytorch_lightning/utilities/parsing.py
```diff
@@ -162,14 +162,18 @@ def get_init_args(frame: types.FrameType) -> Dict[str, Any]:
 
 
 def collect_init_args(
-    frame: types.FrameType, path_args: List[Dict[str, Any]], inside: bool = False
+    frame: types.FrameType,
+    path_args: List[Dict[str, Any]],
+    inside: bool = False,
+    classes: Tuple[Type, ...] = (object,),
```
carmocca, line 168: Initializing the tuple to contain a class looks like a code smell to me.

You could also do:

```py
classes = (,)
...

if "__class__" in local_vars and not classes or issubclass(local_vars["__class__"], classes):
    ...
```

carmocca (approved):

src/pytorch_lightning/utilities/parsing.py
awaelchli: What about it is a code smell? 
An empty tuple expresses that there are no classes we should search for, but on the contrary, we interpret it as looking for all of them. The special casing only makes it more difficult to understand the code now.

carmocca: I guess it's not really a problem because (1) tuples are immutable and (2) it contains a class type which is a singleton anyways. But in general this pattern is discouraged because all function calls will share the same instance

```python
def foo(arg=tuple()):
    return arg

a = foo()
b = foo()
print(a is b)
```

So it's up to you. I prefer the explicit check but feel free to revert. It's got my approval both ways :smile:

otaj (approved):

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -70,6 +70,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed epoch-end logging results not being reset after the end of the epoch ([#14061](https://github.com/Lightning-AI/lightning/pull/14061))
 
 
+- Fix saving hyperparameters in a composition where the parent class is not a `LightningModule` or `LightningDataModule` ([#14151](https://github.com/Lightning-AI/lightning/pull/14151))
```
rohitgr7, line 73:
```suggestion
- Fixed saving hyperparameters in a composition where the parent class is not a `LightningModule` or `LightningDataModule` ([#14151](https://github.com/Lightning-AI/lightning/pull/14151))
```

rohitgr7 (approved):

PR #14170, opened 2022-08-11
Remove skipping logic in favor of path filtering

carmocca:
## What does this PR do?

If the check groups work, we no longer need to run empty jobs for the green checkmark. We can now use the path filtering mechanism

Addresses https://github.com/Lightning-AI/lightning/pull/13661#pullrequestreview-1039323533

A follow-up PR could add further filtering to other jobs

None

### DIFF
--- a/.github/checkgroup.yml
+++ b/.github/checkgroup.yml
@@ -101,7 +101,6 @@ subprojects:
   - id: "lightning_app"
     paths:
       - ".github/workflows/ci-app*.yml"
-      - "examples/app_**"
       - "requirements/app/**"
       - "src/lightning_app/**"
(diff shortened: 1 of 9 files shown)


.github/checkgroup.yml
```diff
@@ -101,7 +101,6 @@ subprojects:
   - id: "lightning_app"
     paths:
       - ".github/workflows/ci-app*.yml"
-      - "examples/app_**"
```
carmocca, line 104: This is duplicated

Borda (approved):

tchaton (approved):
LGTM !

PR #11867, opened 2022-02-11
Support sharded optimizers outside of DDP sharded strategy

DuYicong515:
## What does this PR do?
Fixes #6387 

### Does your PR introduce any breaking changes? If yes, please list them.
This check was removed. 
https://github.com/PyTorchLightning/pytorch-lightning/blob/6f22b3623c28028026b3cb8bb534c1ebca9c5ac8/pytorch_lightning/strategies/sharded.py#L88-L90

 However, the check was also removed in sharded_spawn,py, also the optimizer type hint is "OSS", this might not happen. 
 https://github.com/PyTorchLightning/pytorch-lightning/blob/6f22b3623c28028026b3cb8bb534c1ebca9c5ac8/pytorch_lightning/strategies/sharded_spawn.py#L72-L75

The original PR [link](https://github.com/PyTorchLightning/pytorch-lightning/commit/da6dbc8d1d128cf783d7151b012a5502bbd52bf5#diff-dd216bb90001809b0fa26c62c8261069ffcf18eff5278ee5a62b5e362ba1930fL41-L43) that removed the check
 

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [N/A] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [N/A] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

[...]

PR #14208, opened 2022-08-15
Support sharded optimizer state dumping outside of sharded strategies

### LINKED ISSUE #6387: Support for sharded optimizers when dumping checkpoints outside of the DDP sharded training type plugin
## 🐛 Bug

Using fairscale distributed optimizers without DDP sharded leads to crashes or inconsistent stae in the trainer when checkpointing. This will also occur with PyTorch's latest prototype version of the ZeroRedundancy optimizer. @SeanNaren 

Imagine this scenario:
- Someone wraps their optimizer with Fairscale OSS inside their lightning module. The user *does not* use the DDP sharded plugin.
- At checkpoint time, when the trainer dumps the checkpoint dict, it looks up the optimizer state
- The optimizer state goes through the training type plugin
- The training type plugin calls `optimizer.state_dict()` - For fairscale/pytorch distributed optimizers, we need to consolidate the state dict on a rank. Afterwards, we should look up the state dict only from that rank.

One could add a callback which implements `on_save_checkpoint` to call `consolidate_state_dict()` on the optimizer across all ranks. However, the trainer calls `state_dict` on all ranks, leading to the exception here: https://github.com/facebookresearch/fairscale/blob/1204c7cf54ec301d46a0d3f3fd703da6b306f8f5/fairscale/optim/oss.py#L354-L358

This error occurs only on non-zero ranks. As a result, the error in checkpointing is compounded by the comment here around the exception handling logic for training and its interaction with checkpointing: https://github.com/PyTorchLightning/pytorch-lightning/issues/6343#issuecomment-792159048

Proposal to fix:
- We upstream this optimizer state on all training type plugins: https://github.com/PyTorchLightning/pytorch-lightning/blob/966184a452c773a691b063195c817f3aed899f16/pytorch_lightning/plugins/training_type/sharded.py#L56-L68
- This way, until PyTorch has a consistent interface for distributed/sharded optimizers, we check if the optimizer is fairscale OSS/PyTorch ZeroRedundancy. If so, we call `consolidate_state_dict` on all ranks, and then fetch the optimizer state only from rank 0. 
- plugins for FSDP or fancier things can continue to override this to handle the optimizer states however they need to. 
- If the optimizer is neither of these, we use the existing flow

cc @awaelchli @ananthsub @ninginthecloud @rohitgr7 @tchaton @akihironitta @blefaudeux

awaelchli:
## What does this PR do?

Fixes #6387
Redo of  #11867

Motivation: User wants to checkpoint the sharded optimizer outside of ddp/sharded strategy, e.g. not using it. 

This PR implements it exactly as proposed in #6387 and #11867. However, this leaks fairscale specific logic into the base strategy. We have some options to mitigate the issue:

**Option 0:**
Do not care about this. Move forward with this PR as is.

**Option A:**
Only apply the current approach to the native sharded optimizer from torch. For Fairscale, the user is still forced to use the dedicated strategy and we will keep the logic for this one in the Fairscale sharded strategy. 

**Option B:** 
Instead of moving it to the top into the base strategy, move it to the parallel strategy. This means you will have to use at least a strategy where parallel execution is assumed. Otherwise, the logic stays the same.

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @rohitgr7 @akihironitta


src/pytorch_lightning/strategies/strategy.py
```diff
@@ -170,6 +174,19 @@ def optimizer_state(self, optimizer: Optimizer) -> Dict[str, Tensor]:
...
+            optimizer = optimizer._optimizer
+
+        if _FAIRSCALE_AVAILABLE:
+            from fairscale.optim import OSS
+
+        if (_TORCH_GREATER_EQUAL_1_10 and isinstance(optimizer, ZeroRedundancyOptimizer)) or (
+            _FAIRSCALE_AVAILABLE and isinstance(optimizer, OSS)
+        ):
+            optimizer.consolidate_state_dict()
+            return optimizer.state_dict() if self.is_global_zero else {}
```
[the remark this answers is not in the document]
awaelchli: I completely agree. I was assigned the issue in the sprint planning, where it was decided that we complete the issue. 

> Would return optimizer.state_dict() if self.is_global_zero else {} be valid in all cases? We only checkpoint at rank0 anyways, right?

I don't think so. It should normally be returned on all ranks for consistency, and then the checkpoint_io/strategy would decide what to do. For example in deepspeed, each rank saves a shard of weights.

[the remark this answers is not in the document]
justusschock: Could we do it by a separate checkpoint callback or an io plugin?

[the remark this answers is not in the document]
awaelchli: This consolidation doesn't fit well into the checkpoint callback, and the io plugins are meant to deal with IO only. 
I like your solution with `hasattr(optimizer, 'consolidate_state_dict')` and updated the implementation.

justusschock (approved):

src/pytorch_lightning/strategies/strategy.py
```diff
@@ -170,6 +170,16 @@ def optimizer_state(self, optimizer: Optimizer) -> Dict[str, Tensor]:
...
         Allows for syncing/collating optimizer state from processes in custom plugins.
         """
+        if isinstance(optimizer, LightningOptimizer):
+            optimizer = optimizer._optimizer
+
+        if hasattr(optimizer, "consolidate_state_dict"):
+            # there are optimizers like Fairscale's OSS or PyTorch's ZeroRedundancyOptimizer that shard their
+            # states, and to avoid OOM we consolidate the full state on rank 0 only
+            optimizer.consolidate_state_dict()
+            return optimizer.state_dict() if self.is_global_zero else {}
```
carmocca, line 180: In the deleted code, `rank_zero_only` was used which checks for the local rank, whereas this checks for the global rank. Is this intended?

tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -252,3 +258,50 @@ def test_ddp_strategy_set_timeout(mock_init_process_group):
...
+    trainer = Trainer(accelerator="gpu", devices=2, strategy=strategy, max_epochs=2)
+
+    trainer.fit(model)
+
+    checkpoint_path = os.path.join(tmpdir, "model.pt")
+    trainer.save_checkpoint(checkpoint_path)
+    saved_model = BoringModel.load_from_checkpoint(checkpoint_path)
+
+    # Assert model parameters are identical after loading
+    for ddp_param, shard_param in zip(model.parameters(), saved_model.parameters()):
```
carmocca, line 283: nit: why are these variables called ddp/shard param?

```suggestion
    for trained_param, loaded_param in zip(model.parameters(), saved_model.parameters()):
```

This comment applies to the other tests too

carmocca (approved):

tests/tests_pytorch/strategies/test_ddp_strategy.py
awaelchli: yep! updated all occurrences

src/pytorch_lightning/strategies/strategy.py
awaelchli: `rank_zero_only` means global rank in general. 
The special case is, when running on single node, `local_rank == global_rank`
Does that answer your question?

We could also decorate with rank_zero_only here, but the problem with that is it returns None on rank > 0. But I think it makes way more sense to return the empty dict.

tests/tests_pytorch/strategies/test_sharded_strategy.py
```diff
@@ -314,3 +315,27 @@ def test_block_backward_sync():
...
+        base_optimizer = torch.optim.SGD(self.layer.parameters(), lr=0.1)
+        return OSS(params=base_optimizer.param_groups, optim=type(base_optimizer), **base_optimizer.defaults)
+
+
+@RunIf(min_cuda_gpus=2, skip_windows=True, fairscale=True)
+@pytest.mark.parametrize("strategy", (pytest.param("ddp_sharded", marks=RunIf(standalone=True)), "ddp_sharded_spawn"))
+def test_ddp_sharded_strategy_checkpoint_multi_gpu_fairscale_optimizer(tmpdir, strategy):
+    """Test to ensure that checkpoint is saved correctly when using fairscale optimizers."""
+    model = BoringFairScaleOptimizerModel()
+    trainer = Trainer(accelerator="gpu", devices=2, strategy=strategy, max_epochs=2)
```
rohitgr7, line 331: does it need 2 complete epochs?

tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -252,3 +258,50 @@ def test_ddp_strategy_set_timeout(mock_init_process_group):
...
+    def configure_optimizers(self):
+        return ZeroRedundancyOptimizer(self.layer.parameters(), optimizer_class=torch.optim.Adam, lr=0.1)
+
+
+@RunIf(min_cuda_gpus=2, skip_windows=True, min_torch="1.10")
+@pytest.mark.parametrize("strategy", (pytest.param("ddp", marks=RunIf(standalone=True)), "ddp_spawn"))
+def test_ddp_strategy_checkpoint_zero_redundancy_optimizer(tmpdir, strategy):
+    """Test to ensure that checkpoint is saved correctly when using zero redundancy optimizer."""
+    model = BoringZeroRedundancyOptimizerModel()
+    trainer = Trainer(accelerator="gpu", devices=2, strategy=strategy, max_epochs=2)
```
rohitgr7, line 297: same here

tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -252,3 +258,50 @@ def test_ddp_strategy_set_timeout(mock_init_process_group):
...
+        base_optimizer = torch.optim.SGD(self.layer.parameters(), lr=0.1)
+        return OSS(params=base_optimizer.param_groups, optim=type(base_optimizer), **base_optimizer.defaults)
+
+
+@RunIf(min_cuda_gpus=2, skip_windows=True, fairscale=True)
+@pytest.mark.parametrize("strategy", (pytest.param("ddp", marks=RunIf(standalone=True)), "ddp_spawn"))
+def test_ddp_strategy_checkpoint_multi_gpu_fairscale_optimizer(tmpdir, strategy):
+    """Test to ensure that checkpoint is saved correctly when using faircale optimizer."""
+    model = BoringFairScaleOptimizerModel()
+    trainer = Trainer(accelerator="gpu", devices=2, strategy=strategy, max_epochs=2)
```
rohitgr7, line 274: same here

rohitgr7 (commented):
I think this might be breaking the strategy encapsulation, and now since PyTorch copied Zero from deepspeed, we need to provide special support for it.

Why not let users override the `Strategy` if they are mixing the configuration? or use deepspeed stage 1 if they want to use Zero or some special wrapper?

Also, I remember @otaj is working on checkpoint-related part which will let users choose how the checkpoint is created. So maybe that could be a better way to handle such configuration?

tests/tests_pytorch/strategies/test_sharded_strategy.py
awaelchli: Yes, probably not. Copied the tests from the old PR. But max_steps=1 should do. Thanks

awaelchli:
@rohitgr7 

> I think this might be breaking the strategy encapsulation, and now since PyTorch copied Zero from deepspeed, we need to provide special support for it.

I agree, this does not fit well into our strategy design. It is also not my personal opinion that we should do it that way. But in sprint planning it was determined that it is important to work on this issue and I was assigned to it. So I am going to complete the task regardless. 



> Also, I remember @otaj is working on checkpoint-related part which will let users choose how the checkpoint is created. So maybe that could be a better way to handle such configuration?

It is not the responsibility of the checkpoint callback to know WHAT to save. Therefore it won't belong there.

rohitgr7:
> But in sprint planning it was determined that it is important to work on this issue and I was assigned to it. 

can we discuss this again today or in the retro?

> It is not the responsibility of the checkpoint callback to know WHAT to save. Therefore it won't belong there.

I didn't say checkpoint callback.

awaelchli:
Should this PR and the related issue be closed?

- 👀 Yes, close it
- 👍 No, move forward with this solution

carmocca:
This PR resolves an existing issue in our codebase. Whether the patch is perfect design-wise should be secondary to the improvement in stability. The latter can be improved with time or on a follow-up as designs mature, especially if there's no clear alternative at the moment. My vote is :+1:

Borda (approved):

PR #14212, opened 2022-08-15
Allowed setting attributes on `DataLoader` and `BatchSampler` when instantiated inside `*_dataloader` hooks

otaj:
## What does this PR do?

Since #12981 and #13640, the new mechanism for re-instantiating `DataLoader` and `BatchSampler`s was adopted in case it was necessary to modify some of their behavior. While in general, this allowed for greater flexibility, it also introduced a subtle bug, that it was impossible to set attributes on these classes that would survive the re-instantiation process. To illustrate, let's take a look at the example below:

```python
loader = DataLoader(...)
loader.my_attribute = 10 # <- this will not survive re-instantiation
```

In order to fix it, this PR patches `__setattr__` and `__delattr__` methods, which serve as tape recorders, recording all the `__setattr__` and `__delattr__` calls happening on the object while in the proper context manager. During re-instantiation, all of these calls are replayed on top of the newly created object.

It also replaces all calls of `setattr(obj, <internal_pl_name>, value)` to `object.__setattr__(obj, <internal_pl_name>, value)` , especially within these contexts in order to not run into `RecursionError`s

It was discovered together with @krshrimali while debugging Flash CI

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @ninginthecloud @rohitgr7 @otaj

### DIFF
--- a/src/pytorch_lightning/utilities/data.py
+++ b/src/pytorch_lightning/utilities/data.py
@@ -37,7 +37,7 @@ from pytorch_lightning.overrides.distributed import IndexBatchSamplerWrapper
 from pytorch_lightning.trainer.states import RunningStage
 from pytorch_lightning.utilities.apply_func import _is_dataclass_instance
 from pytorch_lightning.utilities.auto_restart import CaptureIterableDataset, CaptureMapDataset, FastForwardSampler
-from pytorch_lightning.utilities.enums import _FaultTolerantMode
+from pytorch_lightning.utilities.enums import _FaultTolerantMode, LightningEnum
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
 from pytorch_lightning.utilities.meta import _get_all_subclasses
 from pytorch_lightning.utilities.rank_zero import rank_zero_warn
@@ -47,6 +47,18 @@ from pytorch_lightning.utilities.warnings import WarningCache
 BType = Union[Tensor, str, Mapping[Any, "BType"], Iterable["BType"]]
 
 warning_cache = WarningCache()
+
+
+class _WrapAttrTag(LightningEnum):
+    SET = "set"
+    DEL = "del"
+
+    def __call__(self, *args):
+        if self == self.SET:
+            fn = setattr
+        else:
+            fn = delattr
+        return fn(*args)
 
 
 def _extract_batch_size(batch: BType) -> Generator[int, None, None]:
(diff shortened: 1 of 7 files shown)


awaelchli:
This is a huge rabbit hole, what are we doing 🤣 🤣 
What's next? Let's also record all method calls on the dataloader and replaying them? They could have side effects outside the dataloader 🤯  Where will we stop? 😅

awaelchli (approved):

otaj:
I fully agree, @awaelchli, that we're digging a rabbit hole here. And, to be fair, I was also thinking of recording the calls :joy: 

But, this was sadly a real issue, experienced by Flash and I'd rather have this solved for everyone and somewhat nicely, than telling @krshrimali hacks to make it work without this PR :joy:

krshrimali (approved):
Thanks, @otaj for the PR! 🚀 LGTM

rohitgr7 (approved):

src/pytorch_lightning/utilities/data.py
```diff
@@ -475,45 +503,82 @@ def wrapper(obj: Any, *args: Any, **kwargs: Any) -> None:
...
 
         if not hasattr(obj, "__pl_saved_args"):
-            obj.__pl_saved_args = args
-            obj.__pl_saved_kwargs = kwargs
-            obj.__pl_saved_arg_names = param_names
-            obj.__pl_saved_default_kwargs = default_kwargs
+            object.__setattr__(obj, "__pl_saved_args", args)
+            object.__setattr__(obj, "__pl_saved_kwargs", kwargs)
+            object.__setattr__(obj, "__pl_saved_arg_names", param_names)
+            object.__setattr__(obj, "__pl_saved_default_kwargs", default_kwargs)
```
justusschock, line 509: do we need to call this ecplicitly? can't we do `setattr(obj, "__pl_saved_args", args)`

justusschock (approved):

src/pytorch_lightning/utilities/data.py
otaj: I'd almost say that's a philosophical question :joy:

If for some reason a user overwrites their `__setattr__` (for example to mimic frozen class), then calling normal `setattr(obj, ..., ...)` would not work. So, that and running into `RecursionError`s while testing were the reasons why I decided to use `object.__setattr__(obj, ..., ...)` everywhere in the case of these internal attributes

justusschock: Makes sense, I was just wondering if it was an issue with class binding or whether this was for other reasons :) Because I actually prefer not to call magic methods directly if possible, but in this case it makes sense :)

src/pytorch_lightning/utilities/data.py
```diff
@@ -49,6 +49,18 @@
...
 
 
+class _WrapAttrTag(LightningEnum):
+    SET = "set"
+    DEL = "del"
+
+    def __call__(self, *args):
+        if self == self.SET:
+            fn = setattr
+        else:
```
Borda, line 59: elif?

src/pytorch_lightning/utilities/data.py
```diff
@@ -450,6 +442,40 @@ def _auto_add_worker_init_fn(dataloader: DataLoader, rank: int) -> None:
         dataloader.worker_init_fn = partial(pl_worker_init_function, rank=rank)
 
 
+def _reinstantiate_wrapped_cls(orig_object: Any, *args: Any, explicit_cls: Optional[Type] = None, **kwargs: Any) -> Any:
+    if explicit_cls is None:
+        constructor = type(orig_object)
+    else:
+        constructor = explicit_cls
```
Borda, line 449:
```suggestion
    constructor = type(orig_object) if explicit_cls is None else explicit_cls
```

Borda (approved):
stretched my knowledge a bit, but so far I can say, LGTM

carmocca (approved):
Did not review. Will check it in the future. Approving to unblock 👍

src/pytorch_lightning/utilities/data.py
otaj: Not needed since it's an `Enum` with only two possible values. It cannot be constructed with any other value

PR #14224, opened 2022-08-16
Add ColossalAI strategy

ver217:
## What does this PR do?

Fixes https://github.com/hpcaitech/ColossalAI/issues/1330
Fixes https://github.com/Lightning-AI/lightning/issues/12733

Add ColossalAI strategy which supports ZeRO-DP with chunk-based memory management.

No.

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,275 @@
...
+import pytorch_lightning as pl
+from pytorch_lightning.accelerators.cuda import CUDAAccelerator
+from pytorch_lightning.overrides.base import _LightningModuleWrapperBase, _LightningPrecisionModuleWrapperBase
+from pytorch_lightning.plugins.precision import ColossalAIPrecisionPlugin
+from pytorch_lightning.strategies.ddp import DDPStrategy
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.imports import _RequirementAvailable
+from pytorch_lightning.utilities.types import STEP_OUTPUT
+
+_COLOSSALAI_AVAILABLE = _RequirementAvailable("colossalai")
```
rohitgr7, line 14:
```suggestion
from lightning_utilities.core.imports import RequirementCache
from pytorch_lightning.utilities.types import STEP_OUTPUT

_COLOSSALAI_AVAILABLE = RequirementCache("colossalai")
```

rohitgr7: this will fix the CI and run tests

1SAA: Thanks. I've corrected it.

rohitgr7: thanks! looks like there are a few more utilities that were moved to `lightning_utilities.core` recently that is causing this issue and blocking ci tests. Can you try running them locally using PL master and updating the imports? I can't do it since I don't have edit access to their branch.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,275 @@
...
+    ) -> None:
+        if not _COLOSSALAI_AVAILABLE:
+            raise MisconfigurationException(
+                "To use the `ColossalAIStrategy`, please install `colossalai` first. "
+                "Download `colossalai` by consulting https://colossalai.org/download."
+            )
+
+        accelerator = CUDAAccelerator()
+        precision_plugin = ColossalAIPrecisionPlugin()
+        super().__init__(accelerator=accelerator, precision_plugin=precision_plugin)
```
rohitgr7, line 127: this should expect
```py
accelerator
parallel_devices
cluster_environment
checkpoint_io
precision_plugin
```
and should be passed to super class.

also accelerator is configured by accelerator flag inside Trainer so no explicit assignment is required.
although there should be a misconfiguration exception here that it can't work with other accelerators except cuda.

same for precision plugin. Default to `ColossalAIPrecisionPlugin` but users should be able to customize and pass their own.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,275 @@
...
+
+        Returns: Model parallel context.
+        """
+        return ModelShardedContext()
+
+    def setup_precision_plugin(self) -> None:
+        super().setup_precision_plugin()
+        is_training = self.lightning_module.trainer and self.lightning_module.trainer.training
+        if is_training:
+            assert len(self.optimizers) == 1, "ColossalAIStrategy only supports single Optimizer now."
```
rohitgr7, line 177: this should be a `MisconfigurationException`.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,275 @@
...
+
+class ModelShardedContext(ColoInitContext):
+    def _post_init_method(self, module: torch.nn.Module, *args, **kwargs):
+        if getattr(module, "_colossalai_module", False) is True:
+            return
+        super()._post_init_method(module, *args, **kwargs)
+        module._colossalai_module = True
+
+
+class ColossalAIStrategy(DDPStrategy):
```
rohitgr7, line 36: also need to add it to the registry so that users can pass `Trainer(strategy='collosalai')`.
Check deepspeed `register_strategies` for reference.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,60 @@
...
+import pytorch_lightning as pl
+from pytorch_lightning.plugins.precision.precision_plugin import PrecisionPlugin
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.warnings import WarningCache
+
+warning_cache = WarningCache()
+
+
+class ColossalAIPrecisionPlugin(PrecisionPlugin):
```
rohitgr7, line 15: it should accept precision as an argument which should be setup automatically inside accelerator_connector when users pass `Trainer(precision=16, strategy='collosalai')`.

Also does it support:
```
bf16
64-bit precision
```
?

if not, a MisconfigurationException should be raised for others.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,275 @@
...
+
+    It configures accelerator and precision, and you should not configure them when initializing ``Trainer``.
+    CUDA is essential for this strategy. Please make sure CUDA is available.
+
+    Example::
+
+        class GLUETransformer(LightningModule):
+            ...
+            def configure_sharded_model(self) -> None:
+                self.model = BertForSequenceClassification.from_pretrained('bert-base-uncased')
```
rohitgr7, line 48: this is an example of manual sharding. Does it support auto-sharding?

rohitgr7 (commented):
Great work so far!
High-level review.

Are you planning to add docs in this PR or another?

src/pytorch_lightning/strategies/colossalai.py
1SAA: I have tried run tests locally, but I failed. I just set the environment variable `PL_RUN_STANDALONE_TESTS` to 1 and directly used command `pytest` to run all the tests. But the test went endlessly. I wonder what is the correct way to run the tests.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,275 @@
...
+        trainer = Trainer(..., strategy=ColossalAIStrategy())
+
+    Args:
+        use_chunk (bool, optional): Whether to use chunk-based memory management.
+            It can speed up training, but slightly more memory will be used. Defaults to True.
+        chunk_size (Optional[int], optional): The size of a chunk.
+            It will be ignored when ``use_chunk=False``.
+            If it's None, a best chunk size will be searched out based on ``chunk_search_range``,
+            ``chunk_search_n_grids`` and ``min_chunk_size``.
+            Defaults to None.
```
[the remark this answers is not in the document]
1SAA: OK.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Right.

src/pytorch_lightning/strategies/colossalai.py
1SAA: It is not an example of manual sharding. The fact is that we require users to initialize their models in the context `ColoInitContext`, which is wrapped as the return value of `model_sharded_context` function in the strategy. As a result, we set the model attribute of `GLUETransformer` as None in its `__init__` function, and use `configure_sharded_model` function where the model is actually initialized. 
Although Colossalai strategy always make the model auto-sharded, users do not need to care about the distributed storage of the model. We hide all sharding details away from users. 
In sum, we wrote like this to force users initializing their models in `ColoInitContext`. Do you have any idea to realize it in a more explicit way?

src/pytorch_lightning/strategies/colossalai.py
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Sure.

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: this is the correct way to run them, but for now the issues are coming from import so you can run the ones that are not standalone and we can later check the CI with correct imports.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -13,7 +14,15 @@
 
 
 class ColossalAIPrecisionPlugin(PrecisionPlugin):
-    def __init__(self) -> None:
+    def __init__(self, precision: Union[str, int] = 16, amp_level: Optional[str] = None) -> None:
```
rohitgr7, line 17: we can remove amp_level since we are going to deprecate it soon, so let's not add it.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -1,17 +1,19 @@
...
-from pytorch_lightning.plugins.precision import ColossalAIPrecisionPlugin
+from pytorch_lightning.plugins.environments.cluster_environment import ClusterEnvironment
+from pytorch_lightning.plugins.io.checkpoint_plugin import CheckpointIO
+from pytorch_lightning.plugins.precision import ColossalAIPrecisionPlugin, PrecisionPlugin
 from pytorch_lightning.strategies.ddp import DDPStrategy
 from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.imports import _RequirementAvailable
 from pytorch_lightning.utilities.types import STEP_OUTPUT
 
-_COLOSSALAI_AVAILABLE = RequirementCache("colossalai")
```
rohitgr7: why not using `RequirementCache`?

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: if users send a LightningModule initialized on CPU containing some layers inside already and not overriding the `configure_sharded_model`, is it possible for colossalai to initialize it somehow without any context, with some sort of wrapper? Something we recently added to [fully sharded native recently](https://github.com/Lightning-AI/lightning/pull/14252).

.github/checkgroup.yml
```diff
@@ -1,20 +1,18 @@
 custom_service_name: "Lightning CI required checker"
+# For security reasons, configuration is only loaded from the repository's default branch,
```
rohitgr7, line 2: can you rebase with master? maybe you picked some unrelated changes here.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            return self.model(*args, **kwargs)
+
+    def predict_step(self, *args: Any, **kwargs: Any) -> STEP_OUTPUT:
+        assert self.model is not None
+        with self.precision_plugin.predict_step_context():
+            return self.model(*args, **kwargs)
+
+    @classmethod
+    def register_strategies(cls, strategy_registry: Dict) -> None:
+        strategy_registry.register("colossalai", cls, description="Default Colossalai Strategy")
```
rohitgr7, line 294:
```suggestion
        strategy_registry.register("colossalai", cls, description="Default ColossalAI Strategy")
```

.github/checkgroup.yml
1SAA: We just merged the latest master branch to our code. The file `.github/checkgroup.yml` is the same as the parallel file in the latest master branch. I don't know why this problem is occured.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Currently, we don't support transforming models by some wrappers. It is necessary for users to initialize models with the context `ColoInitContext`.

.github/checkgroup.yml
1SAA: The problem is solved when I rebased my branch to the master branch.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Done.

src/pytorch_lightning/plugins/precision/colossalai.py
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Since the previous version didn't merge with the master branch, I have to use the old version of RequirementAvail to run the test. I have changed it to RequirementCache now.

src/pytorch_lightning/plugins/precision/colossalai.py
1SAA: Done.

rohitgr7:
hey @ver217 @1SAA 
Great work here so far. I think we are close. Can you fix the missing imports? Then we can check it on the CI. Currently there's an error: https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=98164&view=logs&j=3f274fac-2e11-54ca-487e-194c91f3ae9f&t=fd22e4d0-068c-5651-f467-789f79e7bb7b

you can try running some tests locally to fix the import issues.

I don't have edit access here, so I can't push any commits.

also, do you plan to add the docs in this PR or in a follow-up?

1SAA:
Hi @rohitgr7

Thank you for your review. I've fixed the import issue. As for the docs, we've written the docstring in `ColossalAIStrategy`. We don't know what else should we do in order to add docs. Could you tell me what else we can do for the docs?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+                child.to(self.root_device)
+
+    @property
+    def lightning_module(self) -> Optional["pl.LightningModule"]:
+        if isinstance(self.model, ZeroDDP):
+            return self.model.module.lightning_module
+        return super().lightning_module
+
+    def teardown(self) -> None:
+        return
```
rohitgr7, line 263: it doesn't work with ddp.teardown?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def optimizer_step(self, optimizer, opt_idx: int, closure, model=None, **kwargs: Any) -> Any:
+        model = model or self.lightning_module
+        return self.precision_plugin.optimizer_step(model, optimizer, opt_idx, closure, **kwargs)
+
+    def lightning_module_state_dict(self):
+        return self.model.state_dict()
```
rohitgr7, line 270: does it return the state dict of the complete model or just the sharded model?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+
+        accumulation_scheduler = trainer.accumulation_scheduler
+        if accumulation_scheduler.epochs != [0]:
+            raise MisconfigurationException(
+                "Colossalai currently does not support different `accumulate_grad_batches` at different epochs."
+            )
+        self.setup_optimizers(trainer)
+
+        if not isinstance(self.precision_plugin, ColossalAIPrecisionPlugin):
+            raise MisconfigurationException("`ColossalaiStrategy` is only compatible with `ColossalAIPrecisionPlugin`.")
```
rohitgr7, line 239: can you move all these exceptions at the top in this method?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+
+    @property
+    def root_device(self) -> torch.device:
+        if self.parallel_devices is not None:
+            return self.parallel_devices[self.local_rank]
+        return get_current_device()
+
+    def model_to_device(self) -> None:
+        pl_module = self.lightning_module
+        pl_module._device = self.root_device
```
rohitgr7, line 251:
```suggestion
```
already done in `setup` above

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+    def clip_grad_by_value(self, optimizer: Optimizer, clip_val: Union[int, float]) -> None:
+        raise MisconfigurationException("clip_grad_by_value is not supported by `Colossalai`")
+
+    def optimizer_step(self, model, optimizer, optimizer_idx: int, closure, **kwargs: Any) -> Any:
+        closure_result = closure()
+        self._after_closure(model, optimizer, optimizer_idx)
+        skipped_backward = closure_result is None
+        if isinstance(model, pl.LightningModule) and model.automatic_optimization and skipped_backward:
+            raise MisconfigurationException(
+                "Skipping backward by returning `None` from your `training_step` is not supported by `Colossalai`"
```
awaelchli, line 54: Which spelling do we use? ColossalAI, or Colossal AI?

src/pytorch_lightning/plugins/precision/colossalai.py
awaelchli: Conceptual question here: It looks like we have overridden this precision plugin just so we can send out warnings and errors. A particular reason why it isn't handled within the strategy directly? Also, this leaves no room for customization, so I suggest to mark the precision plugin protected, signaling we don't intend a user to modify it. 
What do you think? cc @rohitgr7

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+    from colossalai.nn.parallel import ZeroDDP
+    from colossalai.tensor import ProcessGroup
+    from colossalai.utils import get_current_device
+    from colossalai.utils.model.colo_init_context import ColoInitContext
+    from colossalai.zero import ZeroOptimizer
+
+
+class ColossalAIStrategy(DDPStrategy):
+    """ColossalAI strategy. It only supports single optimizer which must be  `colossalai.nn.optimizer.CPUAdam`_ or
+    `colossalai.nn.optimizer.HybridAdam`_ now. You must initialize your model in ``configure_sharded_model()``.
```
awaelchli, line 32: > You must initialize your model in ``configure_sharded_model()``

I think what is meant here is the creation of the layers (and parameters) has to go into the special hook. Maybe we want to clarify that? "Initialization" is a broad term.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            ...
+            def configure_sharded_model(self) -> None:
+                self.model = BertForSequenceClassification.from_pretrained('bert-base-uncased')
+            def on_load_checkpoint(self, checkpoint) -> None:
+                if not hasattr(self, 'model'):
+                    self.configure_sharded_model()
+        trainer = Trainer(..., strategy=ColossalAIStrategy())
+
+    Args:
+        use_chunk (bool, optional): Whether to use chunk-based memory management.
```
awaelchli, line 49: We normally don't add the types in the docstring, as they get auto-generated in the docs from the type annotation in the signature. I would remove them

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,263 @@
...
+
+if _COLOSSALAI_AVAILABLE:
+    from colossalai.nn.optimizer import HybridAdam
+
+
+def test_invalid_colosalai(monkeypatch):
+    import pytorch_lightning.strategies.colossalai as colossal_strategy
+
+    monkeypatch.setattr(colossal_strategy, "_COLOSSALAI_AVAILABLE", False)
+    with pytest.raises(MisconfigurationException):
```
awaelchli, line 28: Let's add the match= here for the message?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+        class GLUETransformer(LightningModule):
+            ...
+            def configure_sharded_model(self) -> None:
+                self.model = BertForSequenceClassification.from_pretrained('bert-base-uncased')
+            def on_load_checkpoint(self, checkpoint) -> None:
+                if not hasattr(self, 'model'):
+                    self.configure_sharded_model()
+        trainer = Trainer(..., strategy=ColossalAIStrategy())
+
+    Args:
```
awaelchli, line 48: Are the defaults for these args chosen based on what the defaults are in ColossalAI?
Would the average user just run with strategy="colossal", or is the average user expected to tune certain parameters? If the latter, let's make sure the frequently used ones are at the top? Maybe it's already the case, sorry, just making sure.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+    from colossalai.logging import disable_existing_loggers, get_dist_logger
+    from colossalai.nn.optimizer import CPUAdam, HybridAdam
+    from colossalai.nn.parallel import ZeroDDP
+    from colossalai.tensor import ProcessGroup
+    from colossalai.utils import get_current_device
+    from colossalai.utils.model.colo_init_context import ColoInitContext
+    from colossalai.zero import ZeroOptimizer
+
+
+class ColossalAIStrategy(DDPStrategy):
```
awaelchli, line 30: Important consideration: Is Colossal in anyway related to torch ddp? If not, please consider inheriting from the ParallelStrategy base class directly!

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            )
+        self.setup_optimizers(trainer)
+
+        if not isinstance(self.precision_plugin, ColossalAIPrecisionPlugin):
+            raise MisconfigurationException("`ColossalaiStrategy` is only compatible with `ColossalAIPrecisionPlugin`.")
+        self.setup_precision_plugin()
+        self.model_to_device()
+
+    @property
+    def root_device(self) -> torch.device:
```
awaelchli, line 244: I strongly suggest to move all properties to the top after init, followed by public methods. This is the organization we want to follow (some strategies don't do that yet)

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            disable_existing_loggers()
+            gpc.init_global_dist(
+                rank=self.global_rank,
+                world_size=self.world_size,
+                backend="nccl",
+                host=self.cluster_environment.main_address,
+                port=self.cluster_environment.main_port,
+            )
+            gpc.set_device(self.local_rank)
+        self.process_group = ProcessGroup()
```
awaelchli, line 160: this attribute isn't declared in init. 
Please note my comment earlier about overriding DDPStrategy. DDP has it's own attribute called _process_group_backend. Another reason why you might want to consider inheriting from the base class, rather than from DDP.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            self.model = pl_module._colossalai_zero[0]
+        if is_training:
+            self.optimizers = [
+                ZeroOptimizer(optimizer, self.model, gpu_margin_mem_ratio=self.gpu_margin_mem_ratio, **self.amp_kwargs)
+            ]
+
+    def setup(self, trainer: "pl.Trainer") -> None:
+        if not isinstance(self.accelerator, CUDAAccelerator):
+            raise MisconfigurationException(
+                f"Colossalai is only supported on GPU but `{self.accelerator.__class__.__name__}` is used."
```
awaelchli, line 219: Double check spelling for Colossalai?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            )
+        self.accelerator.setup(trainer)
+
+        assert self.lightning_module is not None
+        self.lightning_module._device = self.root_device
+
+        if trainer.accumulate_grad_batches > 1:
+            raise MisconfigurationException(
+                "Colossalai does not support gradient accumulation now. " "please set `accumulate_grad_batches` to 1."
```
awaelchli, line 228: There is a problem with the string here:
```suggestion
                "Colossalai does not support gradient accumulation now. Please set `accumulate_grad_batches` to 1."
```

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+    def root_device(self) -> torch.device:
+        if self.parallel_devices is not None:
+            return self.parallel_devices[self.local_rank]
+        return get_current_device()
+
+    def model_to_device(self) -> None:
+        pl_module = self.lightning_module
+        pl_module._device = self.root_device
+        for child in pl_module.modules():
+            if child is not pl_module and getattr(child, "_colossalai_module", None) is not True:
```
awaelchli, line 253:
```suggestion
            if child is not pl_module and not getattr(child, "_colossalai_module", None):
```
Maybe I'm missing something here. What is the value of `child._colossalai_module` if it's defined?

src/pytorch_lightning/strategies/colossalai.py
awaelchli, line 270: Here, `self.model` is most likely ZeroDDP. But does its state dict really return the state of the inner module? Can you clarify? 
We expect the state dict to contain the parameter names of LM, no prefixes for "module" etc.

Another consideration here is: Is the state dict the same on all ranks? How does colossal consolidate it? If it doesn't do it, or can't, let's document it here?

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,263 @@
...
+        default_root_dir=tmpdir,
+        accelerator="gpu",
+        devices=1,
+        precision=16,
+        strategy="colossalai",
+        max_epochs=1,
+        accumulate_grad_batches={0: 1, 4: 2, 8: 3},
+    )
+
+    with pytest.raises(MisconfigurationException):
```
awaelchli, line 102: what is the expected error here? Let's add the match="..."

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,263 @@
...
+class ModelParallelBoringModel(BoringModel):
+    def __init__(self):
+        super().__init__()
+        self.layer = None
+
+    def configure_sharded_model(self) -> None:
+        self.layer = torch.nn.Linear(32, 2)
+
+    def on_load_checkpoint(self, checkpoint: Dict[str, Any]) -> None:
+        self.configure_sharded_model()
```
awaelchli, line 57: If the strategy configures `restore_checkpoint_after_setup = True` (it's a property), then this wouldn't be necessary I believe. The trainer would call configure_sharded_model() and then load the weights after that.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+    Example::
+
+        class GLUETransformer(LightningModule):
+            ...
+            def configure_sharded_model(self) -> None:
+                self.model = BertForSequenceClassification.from_pretrained('bert-base-uncased')
+            def on_load_checkpoint(self, checkpoint) -> None:
+                if not hasattr(self, 'model'):
+                    self.configure_sharded_model()
+        trainer = Trainer(..., strategy=ColossalAIStrategy())
```
rohitgr7, line 46:
```suggestion
        class GLUETransformer(LightningModule):
            ...

            def configure_sharded_model(self) -> None:
                self.model = BertForSequenceClassification.from_pretrained("bert-base-uncased")

        trainer = Trainer(..., strategy="colossalai")
```
why `on_load_checkpoint` is required?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+        super().setup_precision_plugin()
+        is_training = self.lightning_module.trainer and self.lightning_module.trainer.training
+        if is_training:
+            if len(self.optimizers) > 1:
+                raise MisconfigurationException("ColossalAIStrategy only supports single Optimizer now.")
+            optimizer = self.optimizers[0]
+            assert isinstance(optimizer, (CPUAdam, HybridAdam)), (
+                "ColossalAIStrategy only supports colossalai.nn.optimizer.CPUAdam "
+                "and colossalai.nn.optimizer.HybridAdam."
+            )
```
awaelchli, line 189: Let's make this a regular runtime error, not an assertion. An assertion must always hold true, but here we are dealing with user input.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,263 @@
...
+        default_root_dir=tmpdir,
+        accelerator="gpu",
+        devices=1,
+        precision=16,
+        strategy="colossalai",
+        enable_progress_bar=False,
+        enable_model_summary=False,
+    )
+    with pytest.raises(
+        AssertionError,
```
awaelchli, line 120: See my comment earlier about raising an assertion error in the code

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,263 @@
...
+        strategy="colossalai",
+        track_grad_norm=2,
+        enable_progress_bar=False,
+        enable_model_summary=False,
+    )
+    from pytorch_lightning.plugins.precision.colossalai import warning_cache
+
+    with pytest.warns(UserWarning, match="will be ignored since ColossalAI handles the backward"):
+        trainer.fit(model)
+    print(warning_cache)
```
awaelchli, line 149: Should we remove the print?

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,263 @@
...
+    dm = ClassifDataModule()
+    model = ModelParallelClassificationModel()
+    trainer = Trainer(
+        default_root_dir=tmpdir, accelerator="gpu", devices=2, precision=16, strategy="colossalai", max_epochs=1
+    )
+    trainer.test(model, datamodule=dm)
+
+
+@RunIf(min_cuda_gpus=2, standalone=True, colossalai=True)
+def test_multi_gpu_model_colossalai_fit_test(tmpdir):
```
awaelchli, line 247: If we have this test, let's kill the other two? Or what is the value of having them separate?

rohitgr7 (commented):
For the docs you need to update these files under docs/source-pytorch
```py
./advanced/model_parallel.rst
./extensions/strategy.rst
./guides/speed.rst
./api_references.rst
```
in the `model_parallel.rst` you can add the extensive docs of its usage and some internals with references as a user guide. Also, I'd also suggest adding/linking the benchmark results too, which you showed in the meeting, and more if you have done more experiments.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            If it's None, a best chunk size will be searched out based on ``chunk_search_range``,
+            ``chunk_search_n_grids`` and ``min_chunk_size``.
+        enable_distributed_storage (bool, optional): Whether to storage model in a distributed manner.
+            It reduces memory from 1 to 1/N, but it may slow down training.
+        placement_policy (str, optional): It can be "cpu", "cuda" and "auto".
+            If it's "cpu", parameters, gradients and optimizer states will be offloaded to CPU,
+            which means min CUDA memory will be used.
+            If it's "cuda", they won't be offloaded, which means max CUDA memory will be used. It's the fastest.
+            If it's "auto", they are moving dynamically based on CPU and CUDA memory usage.
+            It will utilize heterogeneous memory space evenly and well.
```
rohitgr7, line 62: can you add them as bullet points?

awaelchli (commented):
This looks great! Thank you for implementing the strategy.
Sorry for the many comments and questions in my review, most of them are just about making sure the things are consistent with the other implementations. This was the main focus of my review.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+            )
+        optimizer.step()
+
+    def _track_grad_norm(self, trainer: "pl.Trainer") -> None:
+        if trainer.track_grad_norm == -1:
+            return
+        # the gradients are not available in the model due to gradient partitioning in zero stage >= 2
```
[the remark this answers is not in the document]
1SAA: Yeah.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+from pytorch_lightning.utilities.enums import PrecisionType
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.model_helpers import is_overridden
+
+warning_cache = WarningCache()
+
+
+class ColossalAIPrecisionPlugin(PrecisionPlugin):
+    def __init__(self, precision: Union[str, int] = 16) -> None:
+        if not (precision == PrecisionType.HALF):
```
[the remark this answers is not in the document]
1SAA: Since the `PrecisionType` only reload the operator `==`, it is not correct here to use the operator `!=`.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+            raise MisconfigurationException(
+                f"`Trainer(strategy='colossalai', precision={precision})` is not supported. `precision` must be 16"
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+            )
+        super().__init__()
+        self.precision = 16
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/plugins/precision/__init__.py
```diff
@@ -39,4 +40,5 @@
     "ShardedNativeMixedPrecisionPlugin",
     "TPUPrecisionPlugin",
     "TPUBf16PrecisionPlugin",
+    "ColossalAIPrecisionPlugin",
```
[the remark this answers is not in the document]
1SAA: OK.

src/pytorch_lightning/strategies/colossalai.py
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            The number of steps to increase loss scale when no overflow occurs.
+        hysteresis (int, optional): The number of overflows before decreasing loss scale.
+        max_scale (float, optional): The maximum dynamic loss scaling value.
+
+    .. _colossalai.nn.optimizer.CPUAdam:
+        https://colossalai.readthedocs.io/en/latest/colossalai/colossalai.nn.optimizer.cpu_adam.html
+    .. _colossalai.nn.optimizer.HybridAdam:
+        https://colossalai.readthedocs.io/en/latest/colossalai/colossalai.nn.optimizer.hybrid_adam.html
+    """
+
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+        min_scale (float, optional): The minimum dynamic loss scaling value.
+        growth_factor (float, optional): The multiplication factor for increasing loss scale.
+        backoff_factor (float, optional): The multiplication factor for decreasing loss scale.
+        growth_interval (int, optional):
+            The number of steps to increase loss scale when no overflow occurs.
+        hysteresis (int, optional): The number of overflows before decreasing loss scale.
+        max_scale (float, optional): The maximum dynamic loss scaling value.
+
+    .. _colossalai.nn.optimizer.CPUAdam:
+        https://colossalai.readthedocs.io/en/latest/colossalai/colossalai.nn.optimizer.cpu_adam.html
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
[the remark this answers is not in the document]
1SAA: Right.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
+        parallel_devices: Optional[List[torch.device]] = None,
+        cluster_environment: Optional[ClusterEnvironment] = None,
+        checkpoint_io: Optional[CheckpointIO] = None,
+        precision_plugin: Optional[PrecisionPlugin] = None,
+    ) -> None:
+        if not _COLOSSALAI_AVAILABLE:
+            raise MisconfigurationException(
+                "To use the `ColossalAIStrategy`, please install `colossalai` first. "
+                "Download `colossalai` by consulting https://colossalai.org/download."
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+                module._colossalai_module = True
+
+        return ModelShardedContext()
+
+    def setup_precision_plugin(self) -> None:
+        super().setup_precision_plugin()
+        is_training = self.lightning_module.trainer and self.lightning_module.trainer.training
+        if is_training:
+            if len(self.optimizers) > 1:
+                raise MisconfigurationException("ColossalAIStrategy only supports single Optimizer now.")
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,294 @@
...
+            pl_module._colossalai_zero = [self.model]
+        else:
+            self.model = pl_module._colossalai_zero[0]
+        if is_training:
+            self.optimizers = [
+                ZeroOptimizer(optimizer, self.model, gpu_margin_mem_ratio=self.gpu_margin_mem_ratio, **self.amp_kwargs)
+            ]
+
+    def setup(self, trainer: "pl.Trainer") -> None:
+        if not isinstance(self.accelerator, CUDAAccelerator):
```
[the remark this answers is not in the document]
1SAA: Currently, we don't have devices withch contain m1. So we have not tested ColossalAI in m1.

src/pytorch_lightning/strategies/colossalai.py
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
1SAA: It returns the complete model just as the function in `torch`. But all the tensors are detached from the parameters.

src/pytorch_lightning/strategies/colossalai.py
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Done.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Maybe only line `pl_module._device = self.root_device` should be deleted?

src/pytorch_lightning/strategies/colossalai.py
1SAA: Done.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def backward(
+        self,
+        model: "pl.LightningModule",
+        closure_loss: Tensor,
+        optimizer: Optional[Optimizer],
+        optimizer_idx: Optional[int],
```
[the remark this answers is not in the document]
1SAA: Done.

src/pytorch_lightning/plugins/precision/colossalai.py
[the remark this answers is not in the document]
1SAA: Done.

1SAA: ColossalAI. And I have changed all names to it.

src/pytorch_lightning/strategies/colossalai.py
1SAA: In the end, I removed the assignment in the `steup` function.

src/pytorch_lightning/strategies/colossalai.py
1SAA: When users want to load checkpoint, their models may not ready yet. So we need to help them initailize models.

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: you should not remove it from setup since we need to make sure that the device is connected before we initialize the optimizers.

1SAA: OK. But somehow it is overwritten, I have to set its value again.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
1SAA: yes.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
1SAA: I saw that there are two similar tests in the file `test_ddp.py`. So we wrote like that.

src/pytorch_lightning/strategies/colossalai.py
1SAA: I've changed it.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
1SAA: Additional message is added.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Types are removed.

src/pytorch_lightning/strategies/colossalai.py
1SAA: I've changed it to `trainer = Trainer(..., accelerator="gpu", precision=16, strategy="colossalai")` which is the current way we are using.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Properties are moved up.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Fixed now.

src/pytorch_lightning/strategies/colossalai.py
1SAA: I have removed this attribute.

src/pytorch_lightning/strategies/colossalai.py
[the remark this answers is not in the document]
rohitgr7: you need to add it outside the doctrings

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,336 @@
...
+
+        max_scale: The maximum dynamic loss scaling value.
+
+    .. _colossalai.nn.optimizer.CPUAdam:
+        https://colossalai.readthedocs.io/en/latest/colossalai/colossalai.nn.optimizer.cpu_adam.html
+
+    .. _colossalai.nn.optimizer.HybridAdam:
+        https://colossalai.readthedocs.io/en/latest/colossalai/colossalai.nn.optimizer.hybrid_adam.html
+
+        strategy_name = "colossalai"
```
rohitgr7, line 106:
```suggestion
```
you need to add it outside the doctrines. It's a class variable

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: okay.. will check

tests/tests_pytorch/strategies/test_colossalai_strategy.py
rohitgr7: yes, we can kill the other two. We already have extensive tests, and this PR only adds another strategy.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,269 @@
...
+    model = ModelParallelClassificationModel()
+    trainer = Trainer(
+        default_root_dir=tmpdir,
+        accelerator="gpu",
+        devices=2,
+        precision=16,
+        strategy=ColossalAIStrategy(initial_scale=32),
+        max_epochs=1,
+    )
+    trainer.fit(model, datamodule=dm)
```
rohitgr7, line 265: can you assert train and val accuracy? also is it possible to limit the batches here, keeping the accuracy benchmark the same?

rohitgr7 (commented):
Great!

few more comments left to address above.
and docs as mentioned here: https://github.com/Lightning-AI/lightning/pull/14224#pullrequestreview-1106210324

1SAA:
Hi @rohitgr7 

After our discussion, we decide to add the docs in another PR. Could we just merge this PR first?

src/pytorch_lightning/strategies/colossalai.py
1SAA: Done.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
1SAA: ok.

tests/tests_pytorch/strategies/test_colossalai_strategy.py
1SAA: Thanks. I've configured it to True now.

src/pytorch_lightning/strategies/colossalai.py
1SAA: Yes. Colossalai is not designed to frequently set up and detroy the distributed environment, it assumed that users only need to initialize the distributed envionment in the start of their scripts and exit the distributed environment in the end. It keeps a global distributed configuration and ddp.teardown will damage this configuration.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,65 @@
...
+            raise MisconfigurationException(
+                "Skipping backward by returning `None` from your `training_step` is not supported by `Colossalai`"
+            )
+        optimizer.step()
+
+    def _track_grad_norm(self, trainer: "pl.Trainer") -> None:
+        if trainer.track_grad_norm == -1:
+            return
+        # the gradients are not available in the model due to gradient partitioning in zero stage >= 2
+        warning_cache.warn(
```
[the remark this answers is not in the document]
carmocca: No for the same reason as https://github.com/Lightning-AI/lightning/pull/14224/files#r969925050

rohitgr7:
okay.. looks like colossalai tests are not running and are skipped since it's not being installed on the CI. Installation instructions are here: `pip install colossalai==0.1.10+torch1.12cu11.3`.
@1SAA can you add this to https://github.com/Lightning-AI/lightning/blob/master/requirements/pytorch/strategies.txt?

src/pytorch_lightning/plugins/precision/colossalai.py
rohitgr7: I think we should keep this because the optimizer handles the `backward` call and `clip_grad_by_norm`. Users should be able to provide a specific plugin for customization and still use this strategy, but it won't' be possible if we have these restrictions in the strategy itself.

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: I am fine either way, but this shouldn't block the PR since a few more strategies use DDPStrategy as the base. We can refactor it later to flatten it.

1SAA:
Hi @rohitgr7 

The package link is added and `HybridAdam` can be imported after the installation. But I still have a problem in my local tests. I passed all my tests "standalone". Here, I test a single test funtion by setting the `standalone` parameter to False and `PL_RUN_STANDALONE_TESTS` to 0. Local tests seem can't run properly when `PL_RUN_STANDALONE_TESTS=1`. I tried to run the file `test_ddp.py` and it just ran forever. It seems the tests do not run really standalone, some deadlock problem is likely occured. I wonder is it normal to have deadlock in local tests and how the really testing plantform handle standalone tests.

.azure/gpu-tests.yml
```diff
@@ -91,6 +91,7 @@ jobs:
         set -e
         python -c "fname = 'requirements/pytorch/strategies.txt' ; lines = [line for line in open(fname).readlines() if 'horovod' not in line] ; open(fname, 'w').writelines(lines)"
         python -c "fname = 'requirements/pytorch/strategies.txt' ; lines = [line for line in open(fname).readlines() if 'bagua' not in line] ; open(fname, 'w').writelines(lines)"
+        python -c "fname = 'requirements/pytorch/strategies.txt' ; lines = [line for line in open(fname).readlines() if 'colossalai' not in line] ; open(fname, 'w').writelines(lines)"
```
otaj, line 94: we need to do the same filtering for building `Dockerfile`s

requirements/pytorch/strategies.txt
```diff
@@ -1,6 +1,7 @@
 # NOTE: the upper bound for the package version is only set for CI stability, and it is dropped while installing this package
 #  in case you want to preserve/enforce restrictions on the latest compatible version, add "strict" as an in-line comment
 
+colossalai==0.1.10
```
otaj, line 4: I highly doubt this is the only possible version, we don't want to pin package version to the particular version

rohitgr7: yeah, i'll address this later once GPU ci is green

1SAA: but it is the only possible version, since versions older than 0.1.10 have different apis and 0.1.10 is the newest version. we can just set `colossalai>=0.1.10`.

otaj: Yes, `colossalai>=0.1.10` is absolutely fine with me, but hard pin to a particular version is just weird.

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,78 @@
...
+        **kwargs: Any,
+    ) -> None:
+        assert optimizer is not None
+        optimizer.backward(tensor)
+
+    def clip_grad_by_norm(self, optimizer: Optimizer, clip_val: Union[int, float]) -> None:
+        optimizer.clip_grad_norm(None, clip_val)
+
+    def clip_grad_by_value(self, optimizer: Optimizer, clip_val: Union[int, float]) -> None:
+        raise MisconfigurationException("`clip_grad_by_value` is not supported by `ColossalAI`")
```
awaelchli, line 52:
```suggestion
        raise NotImplementedError("`clip_grad_by_value` is not supported by `ColossalAI`")
```
or runtimeerror

awaelchli (approved):

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,78 @@
...
+        precision: Half precision (16).
+
+    Raises:
+        MisconfigurationException:
+            If precison is not 16.
+    """
+
+    def __init__(self, precision: Union[str, int] = 16) -> None:
+        if not (precision == PrecisionType.HALF):
+            raise MisconfigurationException(
```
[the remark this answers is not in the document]
rohitgr7: I did a few.. wasn't sure about others. I believe there's an issue around it

dockers/base-cuda/Dockerfile
```diff
@@ -136,6 +136,13 @@ RUN \
     if [[ "$CUDA_VERSION_MM" = "$CUDA_VERSION_BAGUA" ]]; then python -c "import bagua_core; bagua_core.install_deps()"; fi && \
     python -c "import bagua; print(bagua.__version__)"
 
+RUN \
+    # install ColossalAI
+    CUDA_VERSION_MM_COLOSSALAI=$(python -c "import torch ; print(''.join(map(str, torch.version.cuda)))") && \
+    CUDA_VERSION_COLOSSALAI=$(python -c "print([ver for ver in [11.3, 11.1] if $CUDA_VERSION_MM_COLOSSALAI >= ver][0])") && \
+    pip install "colossalai==0.1.10+torch1.12cu${CUDA_VERSION_COLOSSALAI}" --find-links https://release.colossalai.org && \
```
otaj, line 143: Please, also parametrize torch version. I know we are using currently 1.12, but it would be ideal to set and forget, so when the update to 1.13 happens, we don't have to change it here

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -48,6 +48,8 @@
     from colossalai.utils import get_current_device
     from colossalai.utils.model.colo_init_context import ColoInitContext
     from colossalai.zero import ZeroOptimizer
+else:
+    ColoInitContext = None
```
otaj, line 52:
```suggestion
    ColoInitContext = Any
```
Whenever you are defining a symbol to be used as a type, use `Any` instead of `None` so we don't run into errors such as https://github.com/Lightning-AI/lightning/pull/14809

src/pytorch_lightning/plugins/precision/colossalai.py
```diff
@@ -0,0 +1,90 @@
...
+        self,
+        tensor: Tensor,
+        model: "pl.LightningModule",
+        optimizer: Optional[Steppable],
+        optimizer_idx: Optional[int],
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        assert optimizer is not None
+        optimizer.backward(tensor)
```
carmocca, line 58: What class will optimizer be here that it supports backward?

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,449 @@
...
+        growth_factor: float = 2,
+        backoff_factor: float = 0.5,
+        growth_interval: int = 1000,
+        hysteresis: int = 2,
+        max_scale: float = 2**32,
+        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
+        parallel_devices: Optional[List[torch.device]] = None,
+        cluster_environment: Optional[ClusterEnvironment] = None,
+        checkpoint_io: Optional[CheckpointIO] = None,
+        precision_plugin: Optional[PrecisionPlugin] = None,
```
carmocca, line 153: This should be annotated with the new precision plugin introduced

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,449 @@
...
+        chunk_search_n_grids: int = 1024,
+        min_chunk_size: Optional[int] = None,
+        initial_scale: float = 2**32,
+        min_scale: float = 1,
+        growth_factor: float = 2,
+        backoff_factor: float = 0.5,
+        growth_interval: int = 1000,
+        hysteresis: int = 2,
+        max_scale: float = 2**32,
+        accelerator: Optional["pl.accelerators.accelerator.Accelerator"] = None,
```
carmocca, line 149:
```suggestion
        accelerator: Optional["pl.accelerators.Accelerator"] = None,
```

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,449 @@
...
+from pytorch_lightning.strategies.ddp import DDPStrategy
+from pytorch_lightning.strategies.strategy import TBroadcast
+from pytorch_lightning.trainer.states import TrainerFn
+from pytorch_lightning.utilities.enums import PrecisionType
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.types import STEP_OUTPUT
+
+_COLOSSALAI_AVAILABLE = RequirementCache("colossalai")
+if _COLOSSALAI_AVAILABLE:
```
carmocca, line 39: All of these should be moved to local imports in the spirit of #12786

src/pytorch_lightning/plugins/precision/colossalai.py
rohitgr7: CPUAdam or HybridAdam, there's a check while initializing the optimizers from LM.

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: but there are too many. Should I move all of them or the ones that are used only once?

dockers/base-cuda/Dockerfile
rohitgr7: done

dockers/base-cuda/Dockerfile
```diff
@@ -136,6 +138,14 @@ RUN \
     if [[ "$CUDA_VERSION_MM" = "$CUDA_VERSION_BAGUA" ]]; then python -c "import bagua_core; bagua_core.install_deps()"; fi && \
     python -c "import bagua; print(bagua.__version__)"
 
+RUN \
+    # install ColossalAI
+    PYTORCH_VERSION_COLOSSALAI=$(python -c "import torch; print(torch.__version__.split('+')[0][:4])") && \
+    CUDA_VERSION_MM_COLOSSALAI=$(python -c "import torch ; print(''.join(map(str, torch.version.cuda)))") && \
+    CUDA_VERSION_COLOSSALAI=$(python -c "print([ver for ver in [11.3, 11.1] if $CUDA_VERSION_MM_COLOSSALAI >= ver][0])") && \
+    pip install "colossalai==0.1.10+torch${PYTORCH_VERSION_COLOSSALAI}cu${CUDA_VERSION_COLOSSALAI}" --find-links https://release.colossalai.org && \
+    python -c "import colosaalai; print(colossalai.__version__)"
```
otaj, line 147:
```suggestion
    python -c "import colossalai; print(colossalai.__version__)"
```

dockers/base-cuda/Dockerfile
```diff
@@ -88,6 +88,8 @@ RUN \
     python ./requirements/pytorch/adjust-versions.py requirements/pytorch/base.txt ${PYTORCH_VERSION} && \
     python ./requirements/pytorch/adjust-versions.py requirements/pytorch/extra.txt ${PYTORCH_VERSION} && \
     python ./requirements/pytorch/adjust-versions.py requirements/pytorch/examples.txt ${PYTORCH_VERSION} && \
+    # remove colossalai from requirements since they are installed separately
```
otaj, line 91: All the changes made to this `Dockerfile` also need to be made to `dockers/release/Dockerfile`, `build-pl` tests are failing because of that

rohitgr7: argggh.

otaj:
Dockers are hopefully fixed. Will check in the morning

otaj:
@rohitgr7, dockers are building. Docker image build of `build-pl` is slightly wrong, because it bases itself on the result of `build-cuda` images available in dockerhub, which is not available there yet, but it's not an issue, since result of `build-cuda` will be pushed to dockerhub on merge to master, so once that happens, everything will work out as it should.

However, there is one GPU test, that is consistently failing, and I think it was written by @awaelchli, do you think you can take a look there, please?

otaj:
So, I found the issue with the failing test. In #14984 we started making sure, that the fork process is not poisoned by CUDA calls. However, just importing `colossalai` caused a call to `torch.cuda.num_devices()` (through calling `torch_amp`). The fix is to guard any `colossalai` imports by our context manager `_patch_cuda_is_available`. I'm not really sure, if this is the nicest possible solution, but hey, at least it works.

cc @carmocca to give a decision on whether it's okay enough solution

src/pytorch_lightning/strategies/colossalai.py
carmocca: All

otaj: Done

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
+# Copyright The PyTorch Lightning team.
```
carmocca, line 1: The filename should drop the `_strategy` suffix

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+        lr_scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=1)
+        return [optimizer], [lr_scheduler]
+
+
+class ModelParallelBoringModelNoSchedulers(ModelParallelBoringModel):
+    def configure_optimizers(self):
+        return HybridAdam(self.layer.parameters(), lr=1e-3)
+
+
+@RunIf(min_cuda_gpus=1, standalone=True, colossalai=True)
```
carmocca, line 84: Can we test this error without launching a standalone test? This is very expensive to run in CI

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+        enable_progress_bar=False,
+        enable_model_summary=False,
+        gradient_clip_val=1.0,
+        gradient_clip_algorithm="value",
+    )
+    with pytest.raises(NotImplementedError, match="`clip_grad_by_value` is not supported by `ColossalAI`"):
+        trainer.fit(model)
+
+
+@RunIf(min_cuda_gpus=1, standalone=True, colossalai=True)
```
carmocca, line 103: Can we test this error without launching a standalone test? This is very expensive to run in CI

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+    )
+
+    with pytest.raises(
+        ValueError,
+        match="ColossalAI currently does not support different `accumulate_grad_batches` at different epochs.",
+    ):
+        trainer.fit(model)
+
+
+@RunIf(min_cuda_gpus=1, standalone=True, colossalai=True)
```
carmocca, line 123: Can we test this error without launching a standalone test? This is very expensive to run in CI

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+        default_root_dir=tmpdir,
+        accelerator="gpu",
+        devices=1,
+        precision=16,
+        strategy="colossalai",
+        enable_progress_bar=False,
+        enable_model_summary=False,
+    )
+    with pytest.raises(
+        MisconfigurationException,
```
carmocca, line 137: Should be a ValueError

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+        match="`ColossalAIStrategy` only supports `colossalai.nn.optimizer.CPUAdam` "
+        "and `colossalai.nn.optimizer.HybridAdam` as its optimizer.",
+    ):
+        trainer.fit(model)
+
+
+@RunIf(min_cuda_gpus=1, standalone=True, colossalai=True)
```
carmocca, line 144: Can we test this error without launching a standalone test? This is very expensive to run in CI

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+        self.log("test_acc", self.test_acc(logits, y), prog_bar=True, sync_dist=True)
+
+    def predict_step(self, batch, batch_idx):
+        x, _ = batch
+        return self.forward(x)
+
+
+@RunIf(min_cuda_gpus=2, standalone=True, colossalai=True)
+def test_multi_gpu_checkpointing(tmpdir):
+    seed_everything(4321)
```
carmocca, line 244: I don't think seeding is relevant for this test

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+        MisconfigurationException,
+        match="To use the `ColossalAIStrategy`, please install `colossalai` first. "
+        "Download `colossalai` by consulting `https://colossalai.org/download`.",
+    ):
+        ColossalAIStrategy()
+
+
+@RunIf(colossalai=True)
+def test_colossalai_strategy_with_trainer_by_instance(tmpdir):
+    trainer = Trainer(fast_dev_run=True, default_root_dir=tmpdir, precision=16, strategy=ColossalAIStrategy())
```
carmocca, line 51: Trainer doesn't start

```suggestion
def test_colossalai_strategy_with_trainer_by_instance():
    trainer = Trainer(precision=16, strategy=ColossalAIStrategy())
```

tests/tests_pytorch/strategies/test_colossalai_strategy.py
```diff
@@ -0,0 +1,291 @@
...
+
+    assert isinstance(trainer.strategy, ColossalAIStrategy)
+    assert isinstance(trainer.strategy.precision_plugin, ColossalAIPrecisionPlugin)
+
+
+@RunIf(colossalai=True)
+def test_colossalai_strategy_with_trainer_by_string(tmpdir):
+    trainer = Trainer(fast_dev_run=True, default_root_dir=tmpdir, precision=16, strategy="colossalai")
```
carmocca, line 59:
```suggestion
def test_colossalai_strategy_with_trainer_by_string():
    trainer = Trainer(precision=16, strategy="colossalai")
```

src/pytorch_lightning/strategies/colossalai.py
```diff
@@ -0,0 +1,477 @@
...
+from pytorch_lightning.strategies.strategy import TBroadcast
+from pytorch_lightning.trainer.states import TrainerFn
+from pytorch_lightning.utilities.enums import PrecisionType
+from pytorch_lightning.utilities.exceptions import MisconfigurationException
+from pytorch_lightning.utilities.model_helpers import is_overridden
+from pytorch_lightning.utilities.types import STEP_OUTPUT
+
+_COLOSSALAI_AVAILABLE = RequirementCache("colossalai")
+if TYPE_CHECKING:
+    if _COLOSSALAI_AVAILABLE:
```
carmocca, line 41: Why not merge these 2?

otaj: Done

tests/tests_pytorch/strategies/test_colossalai_strategy.py
otaj: We have mix of both in our codebase. But sure

rohitgr7 (approved):

src/pytorch_lightning/strategies/colossalai.py
rohitgr7: as discussed, we will do this ref in future.

otaj (approved):

PR #14298, opened 2022-08-18
Add back support for logging in the gradient clipping hooks

carmocca:
## What does this PR do?

Reverts unintentional breaking change from https://github.com/Lightning-AI/lightning/pull/14069

Reported in Slack by @ericharper 

None

## Before submitting

## PR review

cc @borda @carmocca @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy

### DIFF
--- a/src/pytorch_lightning/trainer/connectors/logger_connector/fx_validator.py
+++ b/src/pytorch_lightning/trainer/connectors/logger_connector/fx_validator.py
@@ -44,8 +44,13 @@ class _FxValidator:
             allowed_on_step=(False, True), allowed_on_epoch=(False, True), default_on_step=True, default_on_epoch=False
         ),
         "lr_scheduler_step": None,
-        "configure_gradient_clipping": None,
-        "clip_gradients": None,
+        # should match `optimizer_step`
+        "configure_gradient_clipping": _LogOptions(
+            allowed_on_step=(False, True), allowed_on_epoch=(False, True), default_on_step=True, default_on_epoch=False
+        ),
+        "clip_gradients": _LogOptions(
+            allowed_on_step=(False, True), allowed_on_epoch=(False, True), default_on_step=True, default_on_epoch=False
+        ),
         "on_before_zero_grad": _LogOptions(
             allowed_on_step=(False, True), allowed_on_epoch=(False, True), default_on_step=True, default_on_epoch=False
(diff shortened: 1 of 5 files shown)


src/pytorch_lightning/trainer/connectors/logger_connector/fx_validator.py
```diff
@@ -44,8 +44,13 @@ class _LogOptions(TypedDict):
             allowed_on_step=(False, True), allowed_on_epoch=(False, True), default_on_step=True, default_on_epoch=False
         ),
         "lr_scheduler_step": None,
-        "configure_gradient_clipping": None,
-        "clip_gradients": None,
+        # should match `optimizer_step`
+        "configure_gradient_clipping": _LogOptions(
+            allowed_on_step=(False, True), allowed_on_epoch=(False, True), default_on_step=True, default_on_epoch=False
+        ),
+        "clip_gradients": _LogOptions(
```
carmocca, line 51: This hook is not meant to be overridden, so we technically shouldn't support `log`ging in it. However, since it's called directly from `configure_gradient_clipping`, we don't have a place to reliably disable logging in it.

Borda (approved):

src/pytorch_lightning/trainer/connectors/logger_connector/fx_validator.py
rohitgr7: I don't think we need to add logging for this one because this is called by user without any override and logging func name will remain `configure_gradient_clipping`.

rohitgr7: the reason I added this here is just for the sake of the test case.

awaelchli (approved):

src/pytorch_lightning/trainer/connectors/logger_connector/fx_validator.py
carmocca: As I described in my previous message, I couldn't find a reliable way to disable logging for this hook. It would need to be done in the `configure_gradient_clipping` hook but it wouldn't work for users that override it

rohitgr7 (approved):

justusschock (approved):

PR #14303, opened 2022-08-18
Improve filtering of Lightning app logs

alecmerdler:
## What does this PR do?

Adds filtering of internal log messages using the delimiters
that separate different sections of the logs.

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_app/utilities/app_logs.py
+++ b/src/lightning_app/utilities/app_logs.py
@@ -2,7 +2,7 @@ import json
 import queue
 import sys
 from dataclasses import dataclass
-from datetime import datetime, timedelta
+from datetime import datetime
 from json import JSONDecodeError
 from threading import Thread
 from typing import Callable, Iterator, List, Optional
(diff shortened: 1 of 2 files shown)


src/lightning_app/utilities/app_logs.py
```diff
@@ -109,18 +143,20 @@ def _app_logs_reader(
...
 
-            if start_timestamp and log_event.timestamp > start_timestamp:
+            if component_logs_started[log_event.component_name] and not is_internal_log(log_event.message):
                 yield log_event
+                continue
+
+            section = LogSection.USER_RUN_FLOW if log_event.component_name == "flow" else LogSection.USER_RUN_WORK
+
+            if log_delimiter_begin_for(section) in log_event.message:
+                component_logs_started[log_event.component_name] = True
```
manskx, line 159: If the app is restarted, will the logs contain the `delimiter_begin` multiple times ?

alecmerdler: Right now, `lightning show logs` doesn't behave the same as the UI, which clears the logs after every app restart. We could do this in the CLI as well by passing `start` query parameter to the logs endpoint. We could include that in this PR or in a follow up.

nicolai86:
can we add individual `lightning_app` files to our mypy config so we reduce the missing typing debt in that area over time?

alecmerdler:
@nicolai86 That's a good idea. I'm less familiar with MyPy tooling, so someone else would be better to implement that.

src/lightning_app/utilities/app_logs.py
manskx: I think the logs should be cleared after every app stop/start, let's do it in a separate PR or this one if you prefer.

Maybe with an additional flag, we could show all logs in the future :D 
`lightning show logs --all`

src/lightning_app/utilities/app_logs.py
```diff
@@ -40,6 +40,41 @@ def __gt__(self, other: "_LogEvent") -> bool:
         return self.timestamp > other.timestamp
 
 
+class LogSection:
+    """Delimiters used to separate sections of Lightning App logs."""
+
+    USER_RUN_FLOW = "USER_RUN_FLOW"
```
hhsecond, line 46: We should probably show the user dependency installation as well

alecmerdler: @hhsecond Getting the CLI to have complete parity with how the UI displays logs is a bigger task than this pull request was intended to cover.

src/lightning_app/utilities/app_logs.py
nohalon: I agree that the logs should get cleared after every app start/stop. can we add that to this PR

src/lightning_app/utilities/app_logs.py
nohalon: @tchaton

tchaton: Hey, @alecmerdler We should do this on the platform side. It doesn't make sense to filter the logs with the private indications in an open-source project.

nohalon:
Let's handle this kind of thing from the platform side. Chatted with @tchaton about it and I agree with him that this needs to be passed properly from the platform

src/lightning_app/utilities/app_logs.py
alecmerdler: @tchaton This logic is just copied from what we do in the UI, which any dedicated person could reverse-engineer. It's significantly more work to completely filter internal things from the external logs API and also have our nice app startup progress indicator. This was intended to be a :bicyclist: solution that would last us a few months until we had time to implement the next iteration.

PR #14319, opened 2022-08-19
Fix device parser logic to avoid creating CUDA context

### LINKED ISSUE #14292: Actor hangs with `ray` using PL v>=1.7
When invoking a training from within a Ray actor, the training process hangs. The reason for this is that Ray hangs (happens on multiple Ubuntu machines) when forking a process the way you did when calling `device_parser.num_cuda_devices`.  This doesn't happen on v1.6.5.

To sum-up (before delving into the details), the hanging occurs because of the following change between versions:
In v1.6.5 the `gpu` (changed to `cuda` in the newer version) accelerator just calls `torch.cuda.device_count() > 0` directly, while in 1.7.0 and above, this function `torch.cuda.device_count` is called in a forked process  (`utilities/device_parser.py`, line 346, in `num_cuda_devices`): 
```
with multiprocessing.get_context("fork").Pool(1) as pool:
    return pool.apply(torch.cuda.device_count)
```

The way I deduced the problem is by listening to system-signals and printing the traceback when getting `SIGUSR1` within the hanged process. The traceback we obtained was (starting after calling `pl.Trainer(...)`)

```
File "<__CONDA__ENV__>/lib/python3.9/site-packages/pytorch_lightning/utilities/argparse.py", line 345, in insert_env_defaults
  return fn(self, **kwargs)
File "<__CONDA__ENV__>/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 433, in __init__
  self._accelerator_connector = AcceleratorConnector(
File "<__CONDA__ENV__>/lib/python3.9/site-packages/pytorch_lightning/trainer/connectors/accelerator_connector.py", line 212, in __init__
  self._set_parallel_devices_and_init_accelerator()
File "<__CONDA__ENV__>/lib/python3.9/site-packages/pytorch_lightning/trainer/connectors/accelerator_connector.py", line 525, in _set_parallel_devices_and_init_accelerator
  if not self.accelerator.is_available():
File "<__CONDA__ENV__>/lib/python3.9/site-packages/pytorch_lightning/accelerators/cuda.py", line 91, in is_available
  return device_parser.num_cuda_devices() > 0
File "<__CONDA__ENV__>/lib/python3.9/site-packages/pytorch_lightning/utilities/device_parser.py", line 346, in num_cuda_devices
  return pool.apply(torch.cuda.device_count)
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/pool.py", line 736, in __exit__
  self.terminate()
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/pool.py", line 654, in terminate
  self._terminate()
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/util.py", line 224, in __call__
  res = self._callback(*self._args, **self._kwargs)
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/pool.py", line 729, in _terminate_pool
  p.join()
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/process.py", line 149, in join
  res = self._popen.wait(timeout)
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/popen_fork.py", line 43, in wait
  return self.poll(os.WNOHANG if timeout == 0.0 else 0)
File "<__CONDA__ENV__>/lib/python3.9/multiprocessing/popen_fork.py", line 27, in poll
  pid, sts = os.waitpid(self.pid, flag)
```

The hanging is reproduced by forking from within a Ray worker:
```python
import ray
import multiprocessing

@ray.remote
def f():
    with multiprocessing.get_context("fork").Pool(1) as pool:
        return pool.apply(int)

print(ray.get(f.remote()))
```

Thanks!

awaelchli:
## What does this PR do?

Fixes #14292 

Discussed several solutions offline with @carmocca and @justusschock:
- Rearrange logic in AcceleratorConnector: not feasible
- Use a different start method to launch the process in which we evaluate the cuda function: too costly
- Detect whether we are running inside of ray process: Not possible

Opened issue on torch side: https://github.com/pytorch/pytorch/issues/83973

The current solution seems to be the only one that allows us to have support for ddp-fork and at the same time be compatible with other frameworks. Ray users will have to set `PL_DISABLE_FORK=1` for now. 

## Before submitting

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7 @justusschock @kaushikb11 @awaelchli @akihironitta


src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -340,7 +341,7 @@ def num_cuda_devices() -> int:
     Unlike :func:`torch.cuda.device_count`, this function will do its best not to create a CUDA context for fork
     support, if the platform allows it.
     """
-    if "fork" not in torch.multiprocessing.get_all_start_methods():
+    if "fork" not in torch.multiprocessing.get_all_start_methods() or _is_forking_disabled():
```
justusschock, line 344: shouldn't this `if "fork" not in torch.multiprocessing.get_all_start_methods()` not be part of `_is_forking_disabled`? Doesn't matter if it's disabled due to the env variable or because it's not supported by the OS

justusschock (approved):

src/pytorch_lightning/utilities/device_parser.py
awaelchli: Maybe it could, but we also have this kind of check in the accelerator connector 

```py
        if strategy_flag in _DDP_FORK_ALIASES and "fork" not in torch.multiprocessing.get_all_start_methods():
            raise ValueError(
                f"You selected `Trainer(strategy='{strategy_flag}')` but process forking is not supported on this"
                f" platform. We recommed `Trainer(strategy='ddp_spawn')` instead."
            )
```

for a specific user message about platform support. So in other words, at the moment "fork not supported by platform" and "fork disabled" are two different error messages.

carmocca (approved):

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for saving sharded optimizer state dict outside of `DDPShardedStrategy` ([#14208](https://github.com/PyTorchLightning/pytorch-lightning/pull/14208))
 
 
+- Added an environment variable `PL_DISABLE_FORK` that can be used to disable all forking in the Trainer ([#14319](https://github.com/Lightning-AI/lightning/issues/14319))
```
rohitgr7, line 21: can you add it to the docs?

rohitgr7 (approved):

src/pytorch_lightning/CHANGELOG.md
rohitgr7: oh! nvm.
there's an error

awaelchli: I don't want to. This is just a workaround until https://github.com/pytorch/pytorch/issues/83973 lands. And we will open an issue in ray lightning to set this variable.

justusschock: @awaelchli I still would prefer it to be mentioned somewhere in the docs as even when the issue in torch lands, this would only be valid for the most recent torch version meaning that we may need to go with the current approach for a while (until all supported versions of torch are released past this issue).

Borda (approved):

src/pytorch_lightning/CHANGELOG.md
awaelchli: Yes, we are aware that this will have to stay for a while. But every user is a potential ray user and so documenting this environment variable in the "introduction" is not exactly great. It also can't exactly be documented in the ddp-fork section because the purpose is different. Then the only other solution is to document ALL environment variables in Lightning in one place. This will unnecessarily hold back this PR.

justusschock: Agreed. I think documenting ALL environment variables together with their purpose is something that would definitely be helpful but also definitely goes beyond this PR :)

PR #14334, opened 2022-08-21
[CLI] Adding opportunity to see basic cluster logs

dmitsf:
## What does this PR do?

The PR introduces an opportunity to see basic BYOC logs in CLI: creation/any reconciliation/deletion. 
That's needed mainly for debug aims, sometimes we need to know what happens to BYOC cluster.

The PR adds a command to inspect the logs:

```lightning show cluster logs [OPTIONS] CLUSTER_NAME```

The list of the possible options:
```
Options:
  --from TEXT      The starting timestamp to query cluster logs from.
  --to TEXT        The end timestamp / relative time increment to query logs
                   for. This is ignored when following logs (with
                   -f/--follow).
  --limit INTEGER  The max number of log lines returned.
  -f, --follow     Wait for new logs, to exit use CTRL+C.
  --help           Show help message and exit.
```

Some examples:

```
      Print cluster logs:

          $ lightning show cluster logs my-cluster

      Print cluster logs and wait for new logs:

          $ lightning show cluster logs my-cluster --follow

      Print cluster logs, from 48 hours ago to now:

          $ lightning show cluster logs my-cluster --from "48 hours ago"

      Print cluster logs, 10 most recent lines:

          $ lightning show cluster logs my-cluster --limit 10
```

Live example 1:

image

Live example 2:

Screen Shot 2022-08-21 at 6 44 53 PM

No, it doesn't.

## Before submitting

- [x]  Did you write any new necessary tests? (not for typos and docs)

## PR review

## Did you have fun?

cc @borda @tchaton @rohitgr7

### CODE REVIEW

File: src/lightning_app/utilities/cli_helpers.py [hunk 1]
```diff
@@ -119,3 +121,15 @@ def _retrieve_application_url_and_available_commands(app_id_or_name_or_url: Opti
                     raise Exception(f"The server didn't process the request properly. Found {resp.json()}")
                 return lightningapp.status.url, _extract_command_from_openapi(resp.json())
     return None, None
+
+
+def _arrow_time_callback(_ctx, _param, value, arw_now=arrow.utcnow()) -> arrow.Arrow:
```
nicolai86, line 126: can we make sure that any new code added to `src/lightning_app` is mypy compliant? 
Add the files to the [mypy configuration](https://github.com/Lightning-AI/lightning/blob/master/pyproject.toml#L27) and fix all the missing type annotations.

src/lightning_app/utilities/log_helpers.py
```diff
@@ -0,0 +1,32 @@
...
+
+# A general error callback for log reading, prints most common types of possible errors.
+def _error_callback(ws_app: WebSocketApp, error: Exception):
+    errors = {
+        KeyError: "Malformed log message, missing key",
+        JSONDecodeError: "Malformed log message",
+        TypeError: "Malformed log format",
+        ValueError: "Malformed date format",
+    }
+    print(f"Error while reading logs ({errors.get(type(error), 'Unknown')}), {error}", file=sys.stderr)
```
nicolai86, line 31: can we pass the logger down to use `logger.error` instead of print here?

src/lightning_app/utilities/logs_socket_api.py
```diff
@@ -93,3 +97,76 @@ def print_log_msg(ws_app, msg):
         )
 
         return WebSocketApp(socket_url, on_message=on_message_callback, on_error=on_error_callback)
+
+
+class _ClusterLogsSocketAPI(_LogsSocketAPI):
```
nicolai86, line 102: please add tests for the actual log consumption.

src/lightning_app/utilities/cli_helpers.py
```diff
@@ -119,3 +121,15 @@ def _retrieve_application_url_and_available_commands(app_id_or_name_or_url: Opti
...
+    try:
+        return arw_now.dehumanize(value)
+    except:
+        pass
```
Borda, line 130: Shall here be some warning at least...

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    log_thread = Thread(target=log_socket.run_forever)
+
+    # Establish connection and begin pushing logs to the print queue
+    log_thread.start()
+
+    # Print logs from queue when log event is available
+    try:
+        while True:
```
Borda, line 91: Will this ever get out with the infinite loop? with raised exception?

src/lightning_app/cli/lightning_cli.py
```diff
@@ -141,6 +145,93 @@ def logs(app_name: str, components: List[str], follow: bool) -> None:
...
+@click.option("--limit", default=1000, help="The max number of log lines returned.")
+@click.option("-f", "--follow", required=False, is_flag=True, help="Wait for new logs, to exit use CTRL+C.")
+def cluster_logs(cluster_name: str, to_time: arrow.Arrow, from_time: arrow.Arrow, limit: int, follow: bool) -> None:
+    """Show cluster logs.
+
+    Example uses:
+
+        Print cluster logs:
+
+            $ lightning show cluster logs my-cluster
```
awaelchli, line 180: Do you think we could converge to

```
lightning show [what] [name or id]
```

in the future? This was our vision initially. The id or name of the resource would identify whether it is an app, cluster, whatever (from looking up in database). So instead of

```
lightning show cluster logs my-cluster
```

we would have

```
lightning show logs my-cluster
```

but you could also have

```
lightning show logs my-app
```

The arguments --from and --to generalize to all of them too.

src/lightning_app/cli/cmd_clusters.py
```diff
@@ -93,10 +93,14 @@ def create(
...
-    def list(self):
+    def get_clusters(self):
         resp = self.api_client.cluster_service_list_clusters(phase_not_in=[V1ClusterState.DELETED])
+        return resp.clusters
+
+    def list(self):
+        clusters = self.get_clusters()
         console = Console()
-        console.print(ClusterList(resp.clusters).as_table())
+        console.print(ClusterList(clusters).as_table())
```
adam-lightning, line 103: I'd either remove the `get_clusters` function and merge it into `list()` - if not used anywhere else. 
Or (probably better) make the `get_clusters` return `ClusterList` object - this would reduce passing between functions OpenAPI objects - and use our internal objects for representation.

src/lightning_app/cli/lightning_cli.py
```diff
@@ -141,6 +145,93 @@ def logs(app_name: str, components: List[str], follow: bool) -> None:
...
+        limit=limit,
+        follow=follow,
+    )
+
+    colors = {"error": "red", "warn": "yellow", "info": "green"}
+
+    for log_event in log_reader:
+        date = log_event.timestamp.strftime("%m/%d/%Y %H:%M:%S")
+        color = colors.get(log_event.labels.level, "green")
+        rich.print(f"[{color}]{log_event.labels.level:5}[/{color}] {date} {log_event.message.rstrip()}")
```
adam-lightning, line 232: Probably we should unify printing with application logs - i.e use the same format and colors.

src/lightning_app/utilities/cli_helpers.py
adam-lightning, line 126: Just asking: do we have UTC on all timestamps in all system? Or this translation might cause confusion in some cases?

src/lightning_app/cli/lightning_cli.py
adam-lightning: Agree with @awaelchli - If possible we should show logs by resource id. So one `show logs <res>` would exists, instead of separate "per-resource-type".
I could help out with that functionality if needed, let me know ;)

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -0,0 +1,108 @@
...
+import dateutil.parser
+from websocket import WebSocketApp
+
+from lightning_app.utilities.log_helpers import _error_callback, _OrderedLogEntry
+from lightning_app.utilities.logs_socket_api import _ClusterLogsSocketAPI
+from lightning_app.utilities.network import LightningClient
+
+
+@dataclass
+class _ClusterLogEventLabels:
```
adam-lightning, line 17: Have we thought about unifying that with AppLogs?  I think it should be doable ;D

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if "message" in event_dict:
+            message = event_dict["message"]
+            timestamp = dateutil.parser.isoparse(event_dict["timestamp"])
+            event = _ClusterLogEvent(
+                message=message,
+                timestamp=timestamp,
+                labels=labels,
+            )
+            read_queue.put(event)
```
adam-lightning, line 55: We can (if I remember correctly) push tuple (ts, event) and queue will take weight from first argument (ts) 
So there will be no need to implement < and <= operators for `_ClusterLogEvent` :)
Also inheritance on those Event classes seems odd (code smell)

adam-lightning (changes requested):

src/lightning_app/utilities/cli_helpers.py
dmitsf: We have a common Exception raising below, but sure, having it here will be better!

src/lightning_app/cli/lightning_cli.py
dmitsf: I think this can be a good next step, sure! For this stage, `lightning show cluster logs my-cluster` is the approved option, so I think it will be good to leave it untouched here.

src/lightning_app/utilities/cli_helpers.py
dmitsf: Yes, we use UTC, and this is a good point :) But the confusion is a well-known for many cases.

src/lightning_app/utilities/cluster_logs.py
dmitsf: I general - yes, but we have to do quite a lot backend changes :)

src/lightning_app/cli/lightning_cli.py
dmitsf: I think it would be good to keep color differentiating for log levels. We inspect only a single cluster here, i.e. no need to highlight different components.
On the other format parts, we try to keep the date and message in the same form as app logs have 
(In app logs, we have:
```        
rich.print(f"[{color}]{log_event.component_name}[/{color}] {date} {log_event.message}")
```
)

dmitsf: And I keep the same field width for log level to avoid "dancing lines" :)

src/lightning_app/utilities/cluster_logs.py
dmitsf: Sure, nice idea! I think yes, but maybe it should be unified with app logs :) Maybe can be a subject of a subsequent PR to don't touch too many things here.

src/lightning_app/cli/cmd_clusters.py
dmitsf: Done!

src/lightning_app/utilities/cluster_logs.py
dmitsf: Yes, we will break the loop in a case of exception - the same as for app logs

src/lightning_app/utilities/cluster_logs.py
adam-lightning: 👍

src/lightning_app/utilities/log_helpers.py
dmitsf: Done!

adam-lightning (approved):
I believe that solution is good enough for current circumstances that we can go with it. 
Tho I'd like to see the plan for further improvements as to not lose track of it.

dmitsf:
> I believe that solution is good enough for current circumstances that we can go with it. Tho I'd like to see the plan for further improvements as to not lose track of it.

Since the main thing is a cluster log command format, probably we should raise this discussion to a product level. Also, I think unifying apps and cluster logs is also important.

src/lightning_app/cli/lightning_cli.py
dmitsf: (Probably we should raise this discussion to a product level in the future)

nicolai86 (approved):

src/lightning_app/cli/lightning_cli.py
hhsecond: The concern is this will "never" get addressed and we'll end up having an inconsistent CLI

src/lightning_app/cli/lightning_cli.py
```diff
@@ -141,6 +145,93 @@ def logs(app_name: str, components: List[str], follow: bool) -> None:
...
+
+    if cluster_name not in clusters:
+        raise click.ClickException(
+            f"The cluster '{cluster_name}' does not exist."
+            f" Please select one of the following: [{', '.join(clusters.keys())}]"
+        )
+
+    log_reader = _cluster_logs_reader(
+        client=client,
+        cluster_id=clusters[cluster_name],
```
hhsecond, line 220: is it `cluster_id` or `cluster_name`?

src/lightning_app/utilities/cli_helpers.py
```diff
@@ -119,3 +121,16 @@ def _retrieve_application_url_and_available_commands(app_id_or_name_or_url: Opti
...
+
+
+def _arrow_time_callback(
+    _ctx: "click.core.Context", _param: "click.core.Option", value: str, arw_now=arrow.utcnow()
+) -> arrow.Arrow:
+    try:
+        return arw_now.dehumanize(value)
+    except ValueError:
+        raise click.ClickException(f"cannot parse time {value}")
+    try:
```
hhsecond, line 133: this code is unreachable

tests/tests_clusters/test_cluster_logs.py
```diff
@@ -0,0 +1,103 @@
...
+    cluster_name = os.environ.get("LIGHTNING_CLOUD_CLUSTER_NAME" "" "")
+    with run_cli(
+        [
+            "show",
+            "cluster",
+            "logs",
+            cluster_name,
+        ]
+    ) as (stdout, stderr):
+        assert "Error while reading logs" in stdout, f"stdout: {stdout}\nstderr: {stderr}"
```
manskx, line 103: These tests will be skipped always unless `LIGHTNING_CLOUD_CLUSTER_NAME` is set. do you plan to add it ? or when do you want to run these tests ?

src/lightning_app/cli/lightning_cli.py
```diff
@@ -141,6 +145,93 @@ def logs(app_name: str, components: List[str], follow: bool) -> None:
...
+
+
+@cluster.command(name="logs")
+@click.argument("cluster_name", required=True)
+@click.option(
+    "--from",
+    "from_time",
+    default="24 hours ago",
+    help="The starting timestamp to query cluster logs from.",
+    callback=_arrow_time_callback,
```
manskx, line 161: Have you tried to use a timestatmp here not only `dehumanize ()` ?

src/lightning_app/utilities/cli_helpers.py
manskx: this callback is very limited to human friendly formats, e.g. `2 days ago` , `in a month` but won't work in case of a normal timestamp. `Mon Sep 08 16:41:45 2014` .

You can just parse the value in case of an exception.

src/lightning_app/cli/lightning_cli.py
dmitsf: TBH, I'm not a fan of the idea where we need to guess what type the `resource_name` in `lightning show logs <resource_name>` has, it can be confusing for users and for our logic too (although we can define type in almost all cases, we can face confusions e.g. if we have app and cluster named `test`, and should restrict the same names for resources). Maybe we should also consider something like `lightning show logs app <app_name>` or `lightning show logs cluster <cluster_name>`.

tests/tests_clusters/test_cluster_logs.py
dmitsf: No, it's not supposed to be checked to CI: we have to have a permanent BYOC to do that.

src/lightning_app/utilities/cli_helpers.py
dmitsf: Added a possibility of parsing ISO-8601 (default in `arrow.get`)
To parse `Mon Sep 08 16:41:45 2014` we have to have a bit more of logic, not sure we need it.

src/lightning_app/cli/lightning_cli.py
dmitsf: In the CLI command user should set `cluster_name`, but in API, we use a `cluster_id` to request logs.

Borda:
> unblock, did not review

hi, pls have look, if you have some time... :pray:

PR #14341, opened 2022-08-22
Raise an error when resuming training with Apex

awaelchli:
## What does this PR do?

Alternative approach to  #12389

Fixes https://github.com/Lightning-AI/lightning/issues/11488

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @carmocca @awaelchli @akihironitta @rohitgr7


src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -97,6 +97,3 @@ def optimizer_step(
 
     def state_dict(self) -> Dict[str, Any]:
         return amp.state_dict()
-
-    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
-        amp.load_state_dict(state_dict)
```
awaelchli, line 102: By deleting this, we do nothing now (super implementation) for all stages by default.

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -209,6 +209,10 @@ def restore_training_state(self) -> None:
     def restore_precision_plugin_state(self) -> None:
         """Restore the precision plugin state from the pre-loaded checkpoint."""
         prec_plugin = self.trainer.precision_plugin
+
+        if isinstance(prec_plugin, ApexMixedPrecisionPlugin) and self.trainer.state.fn == TrainerFn.FITTING:
+            raise RuntimeError("Resuming training with APEX is currently not supported.")
```
awaelchli, line 214: This only matters in training. We can still resume via trainer.test(), but optimizer states don't matter there.

tests/tests_pytorch/models/test_amp.py
```diff
@@ -211,3 +211,24 @@ def configure_optimizers(self):
...
     assert isinstance(trainer.lr_scheduler_configs[0].scheduler.optimizer, optim.Adam)
     assert isinstance(trainer.lr_scheduler_configs[1].scheduler.optimizer, optim.SGD)
+
+
+@RunIf(min_cuda_gpus=1, amp_apex=True)
+def test_amp_with_apex_reload(tmpdir):
+    model = BoringModel()
+    trainer = Trainer(
+        default_root_dir=tmpdir,
+        max_steps=5,
```
carmocca, line 221: Why that many steps?

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -209,6 +209,10 @@ def restore_training_state(self) -> None:
     def restore_precision_plugin_state(self) -> None:
         """Restore the precision plugin state from the pre-loaded checkpoint."""
         prec_plugin = self.trainer.precision_plugin
+
+        if isinstance(prec_plugin, ApexMixedPrecisionPlugin) and self.trainer.state.fn == TrainerFn.FITTING:
```
carmocca, line 213: What about a `Trainer` without apex enabled loaded with a checkpoint that has apex state? That doesn't raise. Should it?

carmocca (approved):

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
justusschock, line 213: Why not raise this error within the `ApexMixedPrecisionPlugin`? That way it would be encapsulated within the responsible plugin.

justusschock (approved):

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
awaelchli: Because inside there, we don't know whether the trainer is training or testing. 
I don't know if I can move it inside like you said. There would still have to be an isinstance check somewhere to set this special attribute.

justusschock: I imagined it to be something like this:


```python
def load_state_dict(self, state_dict):
    self._state_dict_loaded = True
    return super().load_state_dict(state_dict)

def optimizer_step(self, *args, **kwargs): # args and kwargs for simplicity, ofc keep real signature
    # in here we definitely know, that we are training
   if self._state_dict_loaded:
        raise MisconfigurationException(...)
   ... # continue with usual optimizer step.

```

Wouldn't that work?

awaelchli: I understand what you mean now. Yes, I definitely like this more!

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -83,6 +84,8 @@ def optimizer_step(
         closure: Callable[[], Any],
         **kwargs: Any,
     ) -> Any:
+        if self._state_dict_loaded:
+            raise RuntimeError("Resuming training with APEX is currently not supported.")
```
rohitgr7, line 88: should we suggest not using apex while configuring the Trainer to avoid this?

rohitgr7 (approved):

src/pytorch_lightning/plugins/precision/apex_amp.py
awaelchli: How would you show this error message during Trainer configuration? What would be the condition?

rohitgr7:
```py
raise RuntimeError("Resuming training with APEX is currently not supported. Please configure your `Trainer(amp_backend=None)` while resuming.")
```

rohitgr7: oh! my comment might have confused you. I meant let users know how they can update the Trainer configuration to resume without apex :)

Issue #14012, opened 2022-08-03
Update tests that mock cuda availability and extend it for mps 


awaelchli:
## Proposed refactor

We have tests that fail locally on MPS supported devices because their mocking of CUDA does not extend to MPS. After #13947 landed, the availability check actually works now and we need to update some tests. These are mainly the tests that have this combination of settings:

- has `accelerator="gpu"` (which includes both cuda and mps)
- has function mocks like `torch.cuda.is_available` as True or `torch.cuda.device_count > 0`. 

An example of such a test:

https://github.com/Lightning-AI/lightning/blob/e6a8283e9cd9df53fb661c64bbf2037e1391a16d/tests/tests_pytorch/trainer/connectors/test_accelerator_connector.py#L245-L262

### Motivation

Let MPS users run the tests locally (me).

### Pitch

Extend the mocking to MPS where applicable. In some tests it may make sense to actually only test for cuda, in which case we should change `accelerator="gpu"` to `"cuda"`. Some other tests may need to be skipped entirely on MPS. 

Note that simply slapping a `RunIf(min_cuda=x)` on top of these tests is not an option, as for most of the tests we want to run them with mocks on the CPU. 

### Additional context

Rather do this sooner than later. Right now, I can't differentiate from tests failing because of bugs in my branch vs. them failing due to this. 

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/Lightning-AI/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/Lightning-AI/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/Lightning-AI/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging PyTorch Lightning, Transformers, and Hydra.

cc @borda @akihironitta @justusschock

JustinGoheen:
[...]

PR #14368, opened 2022-08-23
Disable non blocking to device with MPS

### LINKED ISSUE #13285: MPS Inf/Nan Loss
I am encountering a bug, namely some of my neural network's targets are corrupted using the M1 GPU. This does not happen on CPU. Specifically, some targets are set to large values (~ -2+e25), resulting in inf/nan loss.

I have isolated the behavior by stepping through the code and verifying at which steps the targets are still in the range (-1,1) as intended. The corrupted values first occur in the **batch** argument of **validation_step()**. I implemented a custom collate_fn to verify that nothing is wrong with my datasets and dataloaders. The correct targets leave the collate_fn, but then some of them appear corrupted within validation_step().

As I wrote above, the corrupted targets appear only when using the M1 GPU, whereas on the CPU everything works correctly. I could investigate this issue further if someone could tell me what happens between collate_fn and validation_step, i.e. where else I should step through the code to identify the source of corruption.

_Originally posted by @gloryVine in https://github.com/PyTorchLightning/pytorch-lightning/issues/13102#issuecomment-1154574207_

cc @akihironitta @justusschock

j0rd1smit:
## What does this PR do?
This PR ensures that the race condition bug in Pytorch (https://github.com/pytorch/pytorch/issues/83015) does not affect lightning. As discussed with @justusschock in https://github.com/Lightning-AI/lightning/issues/13285 this PR disables non-blocking moves to MPS devices since it can result in tensors with different values after the moves to the MPS device. To verify it works now, the test case `test_data_is_not_changed_after_move_to_mps_device` has been added.

Fixes #13285

No.

## Before submitting

## PR review

## Did you have fun?
Yes, this was my first ever PR 😀. It was quite fun to do. However, since it is my first time, I hope I did everything OK. If not, let me know. I'm happy to learn.


src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -40,6 +41,10 @@
 
 
 _CPU_DEVICES = ("cpu", torch.device("cpu"))
+if _TORCH_GREATER_EQUAL_1_12 and torch.backends.mps.is_available() and platform.processor() in ("arm", "arm64"):
```
carmocca, line 44: You can import and use the same check in `src/pytorch_lightning/accelerators/mps.py`

tests/tests_pytorch/accelerators/test_mps.py
```diff
@@ -162,3 +162,14 @@ def to(self, *args, **kwargs):
 
     assert batch.text.type() == "torch.mps.LongTensor"
     assert batch.label.type() == "torch.mps.LongTensor"
+
+
+@RunIf(mps=True)
+def test_data_is_not_changed_after_move_to_mps_device():
+    trainer = Trainer(accelerator="mps", devices=1)
```
carmocca, line 169: I suggest unit-testing the `move_data_to_device` function only, instead of instantiating an entire trainer

tests/tests_pytorch/accelerators/test_mps.py
```diff
@@ -162,3 +162,14 @@ def to(self, *args, **kwargs):
...
+    x = torch.zeros([10, 10])
+    device = torch.device("mps")
+
+    for _ in range(1000):
```
carmocca, line 173: I'm hesitant of the robustness of this test. It might be better to not add a test at all actually.

src/pytorch_lightning/utilities/apply_func.py
j0rd1smit: Thanks for the review.

Good point. I'm fixing this now. However, the simple approach of importing `MPSAccelerator.is_available()` or `_MPS_AVAILABLE` does not work because import `pytorch_lightning.accelerators.mps` creates a circular import:
```ImportError: cannot import name '_TPU_AVAILABLE' from partially initialized module 'pytorch_lightning.utilities'`. 
I tried to fix this by moving `_MPS_AVAILABLE`. I guess  `pytorch_lightning.utilities.imports`  is the right place for this then, but I'm not sure. What do you think?

File: tests/tests_pytorch/accelerators/test_mps.py (hunk 3 above)
Comment (j0rd1smit) [reply tests/tests_pytorch/accelerators/test_mps.py:173]: Yeah you probably right. Creating a test case for a race bug like this is rather hard and will always be flaky. I can create a test case to verify that `.to` is called without `non_blocking`. However this will require some mocking or patching. So I guess that is the reason this test case does not exist for the `cpu` case. So I guess you are right. It might be better to not add a test case for this.

File: src/pytorch_lightning/utilities/apply_func.py (hunk 1 above)
Comment (carmocca) [reply src/pytorch_lightning/utilities/apply_func.py]: I was able to resolve the circular dependency in https://github.com/Lightning-AI/lightning/pull/14368/commits/02220a711bbfb693a5b295a2304d9cd3033b8875

We prefer keeping the availability checks in the files they are used: https://github.com/Lightning-AI/lightning/issues/11826. The ones in imports just haven't been updated.

**Review by carmocca — APPROVED** [review]
Somebody will need to manually run the CI on an MPS device

File: src/pytorch_lightning/utilities/apply_func.py [hunk 4]
```diff
@@ -39,6 +40,10 @@
 
 
 _CPU_DEVICES = ("cpu", torch.device("cpu"))
+if _MPS_AVAILABLE:
+    _MPS_DEVICES = ("mps", torch.device("mps"))
```
justusschock, line 44:
```suggestion
    _MPS_DEVICES = ("mps", torch.device("mps"), torch.device("mps:0"))
```

This is what pytorch creates most of the time and unfortunately, they are not the same

src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -39,6 +40,10 @@
...
+else:
+    _MPS_DEVICES = ("mps",)
```
justusschock, line 46:
```suggestion
_BLOCKING_DEVICE_TYPES = ('cpu', 'mps')
```

src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -342,7 +347,8 @@ def batch_to(data: Any) -> Any:
 
         kwargs = {}
         # Don't issue non-blocking transfers to CPU
-        if isinstance(data, Tensor) and device not in _CPU_DEVICES:
+        # Same with MPS due to a race condition bug: https://github.com/pytorch/pytorch/issues/83015
+        if isinstance(data, Tensor) and device not in _CPU_DEVICES and device not in _MPS_DEVICES:
```
justusschock, line 351:
```suggestion
        if isinstance(device, str):
             device = torch.device(device)
             
        if isinstance(data, Tensor) and device.type not in _BLOCKING_DEVICES:
```

src/pytorch_lightning/utilities/apply_func.py
justusschock: I prefer my other suggestion below :) If we don't use that one, we have to do this though :)

justusschock (approved):

awaelchli:
> Somebody will need to manually run the CI on an MPS device

@carmocca Easier said than done :) See https://github.com/Lightning-AI/lightning/issues/14012
But perhaps we can run some selected tests related to device transfers

src/pytorch_lightning/utilities/device_parser.py
```diff
@@ -109,6 +108,8 @@ def parse_gpu_ids(
     gpus = _normalize_parse_gpu_input_to_list(gpus, include_cuda=include_cuda, include_mps=include_mps)
     if not gpus:
         raise MisconfigurationException("GPUs requested but none are available.")
+    from pytorch_lightning.plugins.environments import TorchElasticEnvironment
```
awaelchli, line 111: This change should be reverted

awaelchli (approved):
Could you also add a changelog entry? :)

justusschock:
> @carmocca Easier said than done :) See #14012 But perhaps we can run some selected tests related to device transfers

All mps-specific tests are passing for me. Except for one unrelated to this PR for which I will open an issue right away :)

src/pytorch_lightning/utilities/apply_func.py
j0rd1smit: Oh, I quite like this solution. It is a bit more elegant. As a side benefit, we then also do **not** need to make these import changes as proposed by @carmocca. So should I undo these import changes, or should I leave them in this PR?

src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -322,6 +322,9 @@ def move_data_to_device(batch: Any, device: Union[str, torch.device]) -> Any:
         - :class:`torch.device`
     """
 
+    if isinstance(device, str):
+        device = torch.device(device)
```
justusschock, line 326:
```suggestion
        device = torch.device(device)
        
    assert isinstance(device, torch.device)
```

j0rd1smit:
The test `trainer/logging_/test_logger_connector.py::test_result_collection_on_tensor_with_mean_reduction` now fails because it creates ` _ResultCollection(True)` instead of ` _ResultCollection(True, device="cpu")` ([code link](https://github.com/Lightning-AI/lightning/blob/34f98836fb452674f43f66babb2325ec17e8e192/tests/tests_pytorch/trainer/logging_/test_logger_connector.py#L538)). Therefore the `_ResultCollection.device == None`. The original implementation of `move_data_to_device` implicitly worked with `device=None` even though it had the type `device: Union[str, torch.device]` ([code link](https://github.com/Lightning-AI/lightning/blob/34f98836fb452674f43f66babb2325ec17e8e192/src/pytorch_lightning/utilities/apply_func.py#L308)).

So I can fix the `test_result_collection_on_tensor_with_mean_reduction` test case, but the implicit behavior for `device=None` will no longer work. Is this ok since it has the type `device: Union[str, torch.device]`? On the other hand, `tensor.to(device=None)` is supported by torch as a no op. So maybe we should support it here? Or should this be a separate PR?

If we decided not to support None in `move_data_to_device` then we would also have a look at `_ResultCollection` since `None` is the default value for its device ([code link](https://github.com/Lightning-AI/lightning/blob/34f98836fb452674f43f66babb2325ec17e8e192/src/pytorch_lightning/trainer/connectors/logger_connector/result.py#L396)).

justusschock:
@j0rd1smit seems like you fixed it. I think it's reasonable to update the type, but I'd do so in another PR as this might require further typing changes for mypy. If you want to tackle that as well, feel free to do so :)

j0rd1smit:
@justusschock agreed that should be a separate PR. So guess this PR is ready.

akihironitta (approved):

src/pytorch_lightning/utilities/apply_func.py
```diff
@@ -342,7 +345,8 @@ def batch_to(data: Any) -> Any:
 
         kwargs = {}
         # Don't issue non-blocking transfers to CPU
-        if isinstance(data, Tensor) and device not in _CPU_DEVICES:
+        # Same with MPS due to a race condition bug: https://github.com/pytorch/pytorch/issues/83015
+        if isinstance(data, Tensor) and isinstance(device, torch.device) and device.type not in _BLOCKING_DEVICE_TYPES:
```
awaelchli, line 349:
```suggestion
        if isinstance(data, Tensor) and device.type not in _BLOCKING_DEVICE_TYPES:
```
`isinstance(device, torch.device)` is always true, given the new lines above that were added.

awaelchli: but mypy does not understand this, so has to be kept probably

carmocca: No because `None` is also passed internally: https://github.com/Lightning-AI/lightning/pull/14368/#issuecomment-1225825455

PR #14381, opened 2022-08-24
testing pydantic v1.10.0a1

samuelcolvin:
### TO NOT MERGE

This is pull request is just to test pydantic v1.10, see https://github.com/pydantic/pydantic/discussions/4359


samuelcolvin:
Any idea if this relates to changes in pydantic? `master` seems to be failing too but with a different error?

PR #14458, opened 2022-08-30
Grid 10410 very large limits say 10000 dont work

nmiculinic:
## What does this PR do?

Fixes https://linear.app/lightning-ai/issue/GRID-10410/very-large-limits-say-10000-dont-work-error-400 partially. 

Today if you try to fetch gridv1.Cluster logs in the CLI and want more than 5000 lines it'll brake due to internal limitations. This PR implements batching, thus we're fetching 5000 by 5000 log lines from the API server. 

No

## Before submitting

## PR review

## Did you have fun?

Not really, though this has to get done.


src/lightning_app/cli/lightning_cli.py
```diff
@@ -216,21 +217,24 @@ def cluster_logs(cluster_name: str, to_time: arrow.Arrow, from_time: arrow.Arrow
...
+        colors = {"error": "red", "warn": "yellow", "info": "green"}
 
-    for log_event in log_reader:
-        date = log_event.timestamp.strftime("%m/%d/%Y %H:%M:%S")
-        color = colors.get(log_event.labels.level, "green")
-        rich.print(f"[{color}]{log_event.labels.level:5}[/{color}] {date} {log_event.message.rstrip()}")
+        for log_event in log_reader:
+            date = log_event.timestamp.strftime("%m/%d/%Y %H:%M:%S")
+            color = colors.get(log_event.labels.level, "green")
+            rich.print(f"[{color}]{log_event.labels.level:5}[/{color}] {date} {log_event.message.rstrip()}")
```
dmitsf, line 235: I see two potential things here:
- `try` block seems to be too large, it includes `logs_api_client` initialization and log reading loop; although I see we already have similar stories it the code (e.g. socket reading loop), maybe we can split it and make more specific exception processing blocks?
- we don't specify `Exception` type while we read logs in the loop. Does it hang while logs reading if a user uses `--follow` and tries to break the loop? Maybe we can specify potential exception types/ or at least process `KeyboardInterrupt` separately to correct break the loop?

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -88,10 +100,46 @@ def _cluster_logs_reader(
 
     # Print logs from queue when log event is available
     try:
+        last_start_time = start
+        items_remaining_in_batch = batch_size
```
dmitsf, line 104: Maybe these lines can be moved from `try` block.

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -39,35 +39,47 @@ def _push_log_events_to_read_queue_callback(read_queue: queue.PriorityQueue):
...
+            read_queue.put(ev)
 
     return callback
 
 
+def _parse_log_event(msg: str) -> List[_ClusterLogEvent]:
+    # We strongly trust that the contract on API will hold atm :D
+    event_dict = json.loads(msg)
+    log_entry_labels = {}
+    allowed_log_entry_labels = {x.name for x in dataclasses.fields(_ClusterLogEventLabels)}
```
dmitsf, line 53: I think, since we don't use most of the labels, now we can:
-  clean up fields in `_ClusterLogEventLabels` - it seems we use only `labels.level`?

or:
- get rid of `_ClusterLogEventLabels` class at all and simplify the things, define labels attribute of `_ClusterLogEvent` class as `field` (but in this case we lose opportunity to access to the log labels as an attribute via `.` (`log_entry.labels.level`), we should use `log_entry.labels['level']`):

```
from dataclasses import dataclass, field

@dataclass
class _ClusterLogEvent(_OrderedLogEntry):
    labels: field(default_factory=dict)
```

nmiculinic: See PR changes, we can still access things via `.` and have type annotations for common things :)

src/lightning_app/cli/lightning_cli.py
nmiculinic: > we don't specify Exception type while we read logs in the loop. Does it hang while logs reading if a user uses --follow and tries to break the loop? Maybe we can specify potential exception types/ or at least process KeyboardInterrupt separately to correct break the loop

It used to hang before this PR, now this is fixed as well :0

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -43,59 +46,95 @@ def _push_log_events_to_read_queue_callback(read_queue: queue.PriorityQueue):
...
-        if "message" in event_dict:
-            message = event_dict["message"]
-            timestamp = dateutil.parser.isoparse(event_dict["timestamp"])
-            event = _ClusterLogEvent(
-                message=message,
-                timestamp=timestamp,
-                labels=labels,
-            )
-            read_queue.put(event)
+    def callback(_: WebSocketApp, msg: str):
```
tchaton, line 49: This isn't the same code, you are returning log_events and it doesn't filter the private logs.

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -43,59 +46,95 @@ def _push_log_events_to_read_queue_callback(read_queue: queue.PriorityQueue):
...
+        )
+        log_events.append(event)
+    return log_events
+
+
 def _cluster_logs_reader(
-    client: LightningClient,
+    logs_api_client: _ClusterLogsSocketAPI,
     cluster_id: str,
-    start: arrow.Arrow,
```
tchaton, line 67: Let's remove arrow dep.

src/lightning_app/cli/lightning_cli.py
```diff
@@ -216,21 +217,24 @@ def cluster_logs(cluster_name: str, to_time: arrow.Arrow, from_time: arrow.Arrow
...
 
-    log_reader = _cluster_logs_reader(
-        client=client,
-        cluster_id=clusters[cluster_name],
-        start=from_time.int_timestamp,
-        end=to_time.int_timestamp,
-        limit=limit,
-        follow=follow,
-    )
+    try:
```
tchaton, line 220: Would be better to understand why this happens.

src/lightning_app/utilities/cluster_logs.py
nmiculinic: I'm not following; this is the same code just refactored?

src/lightning_app/utilities/cluster_logs.py
nmiculinic: This can be in follow up PR, this one is focused on fixes

src/lightning_app/cli/lightning_cli.py
nmiculinic: What do you mean? What would be actionable to do here exactly?

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -13,23 +13,26 @@
...
-    identifier: Optional[str] = None
-    issuer: Optional[str] = None
-    error: Optional[str] = None
-    errorVerbose: Optional[str] = None
-    dir: Optional[str] = None
-    bucket: Optional[str] = None
-    prefix: Optional[str] = None
-    loki_s3: Optional[str] = None
+
+    def __init__(self, **kwargs):
```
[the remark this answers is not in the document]
nmiculinic: Thanks, great suggestion!

src/lightning_app/cli/lightning_cli.py
tchaton: I meant it would be good to understand why `Error while reading logs` is triggered instead of catching it. Logs are key and it needs to be reliable.

src/lightning_app/utilities/cluster_logs.py
tchaton: Oh, my bad, I thought this was the app_logs callback. I guess we don't have private logs for our BYOC clusters.

src/lightning_app/cli/lightning_cli.py
tchaton: I guess the stack trace in debug mode.

tchaton (approved):
LGTM once last comments are addressed.

src/lightning_app/cli/lightning_cli.py
nmiculinic: @tchaton ok

FYI lightning_cloud package uses `DEBUG` envvar, so  this means we're using 2 envvars in the same CLI

nmiculinic: Done; reusing value from lighting_cloud package to keep those things in sync, though it's being imported in the constants

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -2,34 +2,22 @@
...
 
-import arrow
 import dateutil.parser
 from websocket import WebSocketApp
 
-from lightning_app.utilities.log_helpers import _error_callback, _OrderedLogEntry
+from lightning_app.utilities.log_helpers import _OrderedLogEntry
 from lightning_app.utilities.logs_socket_api import _ClusterLogsSocketAPI
-from lightning_app.utilities.network import LightningClient
+from pytorch_lightning.utilities import AttributeDict
```
manskx, line 12: This is a dependency on the `pytorch_lightning` package, which means we cannot use the app package separately.

manskx: @Borda @tchaton is this fine ?

nmiculinic: Ohh...good point! 

What do you recommend I do in this case?

manskx: @nmiculinic for now, we shouldn't use anything form `pytorch_lightning` maybe just duplicate it in the app utilities.

dmitsf (approved):

src/lightning_app/utilities/cluster_logs.py
nmiculinic: @tchaton what do you recommend to use? Maybe copy&paste the AttributeDict from the pytorch lightning?

Borda: Copy the code to App or move the class to utilities...
cc: @carmocca

carmocca: Yes, it could go into https://github.com/Lightning-AI/utilities/tree/main/src/lightning_utilities/core

awaelchli (approved):

nmiculinic:
This is blocked on prod release which support nanosecond precision timestamps for logs

src/lightning_app/utilities/cluster_logs.py
```diff
@@ -1,100 +1,127 @@
...
+    end: Optional[float],
     limit: int,
     follow: bool,
-    on_error_callback: Optional[Callable] = None,
+    batch_size: int = 5000,
 ) -> Iterator[_ClusterLogEvent]:
-
-    logs_api_client = _ClusterLogsSocketAPI(client.api_client)
-    read_queue = queue.PriorityQueue()
+    read_queue = queue.Queue(maxsize=2 * batch_size)
```
dmitsf, line 70: Just a question: does the queue limiting make sense here?

dmitsf (approved):
🚀

PR #14483, opened 2022-09-01
Fix fault tolerant training in cases where num_workers > 0

### LINKED ISSUE #12285: Resuming with fault-tolerant training fails when DataLoader has num_workers>1
## 🐛 Bug

Resuming with fault-tolerant training fails when DataLoader has `num_workers`>1:
```
  ...
  File "~/pytorch-lightning/pytorch_lightning/utilities/auto_restart.py", line 533, in _reload_dataloader_state_dict_automatic_map_dataset
    dataloader.dataset.load_state_dict(
  File "~/pytorch-lightning/pytorch_lightning/utilities/auto_restart.py", line 261, in load_state_dict
    self._cached_state_dict = _rotate_worker_indices(deepcopy(state_dict), latest_worker_id, num_workers)
  File "~/pytorch-lightning/pytorch_lightning/utilities/auto_restart.py", line 615, in _rotate_worker_indices
    raise MisconfigurationException("The `state` should contain `num_workers - 1` values.")
pytorch_lightning.utilities.exceptions.MisconfigurationException: The `state` should contain `num_workers - 1` values.
```

### To Reproduce
Change 
https://github.com/PyTorchLightning/pytorch-lightning/blob/d31126c3310af62d76ffb87368ae1e4cc5bbac93/pl_examples/fault_tolerant/automatic.py#L96-L97
to
```
return DataLoader(RandomGetItemDataset(3, 1), num_workers=2)
```

### Environment

- PyTorch Lightning Version (e.g., 1.5.0): master ([d31126c](https://github.com/PyTorchLightning/pytorch-lightning/commit/d31126c3310af62d76ffb87368ae1e4cc5bbac93))

cc @carmocca @justusschock @awaelchli @ninginthecloud @akihironitta

jhoareau:
## What does this PR do?

Fixes #12285

This fix assumes that the state of all DataLoader workers is the same, which should be the case following my understanding, but please correct me if I'm wrong. If that assumption is wrong, the solution would probably rather take place in `_collect_states_on_rank_zero_over_collection`, where we currently gather states only on rank 0, but we would need to collect state for all ranks.

In my understanding, including in the distributed setting, each worker might look at a different shard of data, but each process shares the same memory space apart from the rank, which should be naturally be populated by Pytorch at instanciation time, unless `rank` is set manually by the user at the sampler level. I do not cover this specific case, but this should already unblock the linked issue.
The corresponding fix was tested on our own codebase with success.

The fix allowed to re-enable a currently disabled test in `test_auto_restart.py`.

## Before submitting

## PR review

### DIFF
--- a/src/pytorch_lightning/utilities/auto_restart.py
+++ b/src/pytorch_lightning/utilities/auto_restart.py
@@ -194,12 +194,18 @@ class MergedIteratorState:
     @property
     def sampler_states(self) -> Dict[int, Any]:
         """Returns the merged sampler states for all worker processes."""
-        return {0: self.state[k].sampler_state[0] for k in self.state.keys()}
+        if self.represent_map_dataset:
+            return {k: self.state[k].sampler_state[0] for k in self.state.keys()}
+        else:
+            return {sampler_name: {k: value.sampler_state[k] for k, value in self.state[sampler_name].items()} for sampler_name in self.state.keys()}
 
     @property
     def dataset_states(self) -> Dict[int, Any]:
-        """Returns the merged dataset states for all worker processes."""
-        return {k: self.state[k].dataset_state[k] for k in self.state.keys()}
+        """Returns the merged sampler states for all worker processes."""
+        if self.represent_map_dataset:
+            return {k: self.state[k].dataset_state[k] for k in self.state.keys()}
+        else:
+            return {sampler_name: {k: value.dataset_state[k] for k, value in self.state[sampler_name].items()} for sampler_name in self.state.keys()}
 
     @classmethod
     def from_state_dict(cls, state_dict) -> "MergedIteratorState":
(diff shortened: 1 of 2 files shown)


tests/tests_pytorch/utilities/test_auto_restart.py
```diff
@@ -1278,8 +1278,7 @@ def test_collect_states_with_collection():
     assert generated == [{"a": {0: state}, "b": [{"a": {0: state}}]}]
 
 
-# FIXME(@tchaton): >0 num_workers failing
-@pytest.mark.parametrize("num_workers", [0, pytest.param(2, marks=[RunIf(slow=True), pytest.mark.xfail()])])
+@pytest.mark.parametrize("num_workers", [0, 2])
```
carmocca, line 1281: Needs to keep the slow mark

```suggestion
@pytest.mark.parametrize("num_workers", [0, pytest.param(2, marks=RunIf(slow=True))])
```

carmocca (commented):
cc @thomas and @awaelchli for review as authors

tests/tests_pytorch/utilities/test_auto_restart.py
jhoareau: Thank you!

src/pytorch_lightning/utilities/auto_restart.py
```diff
@@ -601,10 +601,11 @@ def _rotate_worker_indices(state: Dict[int, Any], latest_worker_id: int, num_wor
     on."""
     if num_workers == 0:
         return state
+    # the state of the first worker is the same as the state of all workers.
+    if len(state) == 1:
+        state = {i: state[0] for i in range(num_workers)}
```
awaelchli, line 606: > This fix assumes that the state of all DataLoader workers is the same, which should be the case following my understanding, but please correct me if I'm wrong.

So, the state can be different in each worker because each worker usually gets a different seed. And you can see that e.g. on line 255 we collect the random state in the getitem, which runs inside each worker process.

This is my quick immediate answer here, but I need to spend more time now studying your fix (and the actual problem)

awaelchli, line 606: > This fix assumes that the state of all DataLoader workers is the same, which should be the case following my understanding, but please correct me if I'm wrong.

So, the state can be different in each worker because each worker usually gets a different seed. And you can see that e.g. on line 255 we collect the random state in the getitem, which runs inside each worker process.

This is my quick immediate answer here, but I need to spend more time now studying your fix (and the actual problem)

jhoareau: Hi @awaelchli, thanks for looking into it.

I can confirm that you seem to be right, in https://github.com/pytorch/pytorch/blob/master/torch/utils/data/_utils/worker.py#L217 each worker gets the same base seed, but adds the worker_id as an offset.
In this case, my fix would break that behaviour, and not resuming the training correctly. I'll have another go at a fix, because I think we can also fix this by collecting states for all ranks instead of just rank 0. I'll convert this PR back to a draft.

jhoareau: I've investigated some more and found out that we are saving the states for all workers correctly, we just don't load it correctly, so the fix might not be too hard anyway in the end.

jhoareau:
Hi @awaelchli @carmocca @tchaton I have managed a fix that works only for the automatic fault-tolerance + map-style dataset case. I do not have the capacity to look into it further for the other cases. You are welcome to take over my PR in order to implement a more proper fix.

PR #14451, opened 2022-08-30
Standalone Lite CI setup

awaelchli:
## What does this PR do?

Foundation for moving Lite to a standalone package.

With this PR, we set up:
- The package
- The CI running tests on the package in separate jobs

Next PRs will move the actual code and replace the placeholder tests cases.

### Does your PR introduce any breaking changes? If yes, please list them.

No.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @carmocca @akihironitta @borda

PR #14502, opened 2022-09-02
Use a standalone test symlink for Lite

carmocca:
## What does this PR do?

Part of #14451

Also part of the standalone config was missing.

None

cc @carmocca @akihironitta @borda @justusschock @awaelchli


.azure/gpu-tests-lite.yml
```diff
@@ -78,7 +80,6 @@ jobs:
     - bash: bash run_standalone_tests.sh
       workingDirectory: tests/tests_lite
       env:
-        PL_USE_MOCKED_MNIST: "1"
```
carmocca, line 81: I don't think Lite will use the MNIST dataset

tests/tests_lite/conftest.py
```diff
@@ -0,0 +1,52 @@
+import os
+from typing import List
+
+import pytest
+
+
+def pytest_collection_modifyitems(items: List[pytest.Function], config: pytest.Config) -> None:
```
carmocca, line 7: This was missed, so the standalone test step is running all tests currently.

tests/tests_lite/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+    """An adaptation of `tests/tests_pytorch/conftest.py::pytest_collection_modifyitems`"""
+    initial_size = len(items)
+    conditions = []
+    filtered, skipped = 0, 0
+
+    options = dict(
+        standalone="PL_RUN_STANDALONE_TESTS",
+        min_cuda_gpus="PL_RUN_CUDA_TESTS",
```
carmocca, line 15: The only change is that I removed the slow test keyword

tests/tests_lite/conftest.py
```diff
@@ -0,0 +1,52 @@
+import os
+from typing import List
+
+import pytest
```
carmocca, line 4: @awaelchli just as a heads up, you will probably need some of the other items we have in `conftest` for tests to work properly.

awaelchli: Yes, I included these already in #14492

awaelchli (approved):

akihironitta (approved):

PR #14508, opened 2022-09-03
Introduce `Logger.experiment_dir`

### LINKED ISSUE #14188: Introduce `Logger.experiment_dir`
## 🚀 Feature

Introduce a property/method `experiment_dir` on the abstract logger interface. All loggers must implement this and return a path to where their files are being saved. 

### Motivation

The Trainer currently determines where checkpoints are saved like so:
```py
os.path.join(logger.save_dir, str(logger.name), logger.version, "checkpoints")
```

As you can see, it assumes that there is a rigid folder structure `save_dir/name/version`. Every logger has to follow this structure.
However, at least one logger (wandb) struggles with this, because it has an additional concept of "project" that does not fit well within this layout (https://github.com/Lightning-AI/lightning/issues/14054). Furthermore, the "name" may change AFTER the files were written, so there is a disconnect between the folder structure and what is represented "online" in the wandb ui. 

One observation is that the trainer actually never cares about any of the subfolders. All it needs is the full path. This is why I believe, if we let the logger decide how to assemble the path, we will avoid many issues and disagreements with 3rd party experiment management. 

We are struggling with making all loggers agree to a standardized layout, so maybe we shouldn't try to do that, at least not in every aspect of how the loggers work (https://github.com/Lightning-AI/lightning/issues/12028#issuecomment-1088325894). 

### Pitch

Introduce

```py
def experiment_dir(self) -> str:
    """Returns the (relative) path to where logs and artifacts get saved for the current version."""

```

This should probably be a relative path, so that the Trainer can prepend a root dir if needed (default cwd). 
An implementation of a logger could then look like this:

```py
def experiment_dir(self):
    return os.path.join(self.save_dir, self.name, self.version)
```

And the checkpointing in the Trainer would save files to

```py
os.path.join(logger.experiment_dir(), "checkpoints") 
```
for example.

### Implications

1. After introducing experiment dir, we could consider deprecating `save_dir` and `log_dir` or remove it from the base interface.
2. The code here in `Trainer.log_dir` would probably get simplified: https://github.com/Lightning-AI/lightning/blob/fe9e5d55bf7991ba36b76d6adae9075b93dfcaa0/src/pytorch_lightning/trainer/trainer.py#L2208-L2219
3. The code in the checkpoint callback would get simplified too: https://github.com/Lightning-AI/lightning/blob/fe9e5d55bf7991ba36b76d6adae9075b93dfcaa0/src/pytorch_lightning/callbacks/model_checkpoint.py#L595-L608
4. It would immediately address these issues, as the LightningCLI can write to the experiment dir without ambiguity: #14162, #12748

### Alternatives

Struggle

### Additional context

Discussions in:
https://github.com/Lightning-AI/lightning/issues/14054

Current blockers: 
#12177

cc @borda @tchaton @justusschock @awaelchli @edward-io @ananthsub @rohitgr7 @kamil-kaczmarek @Raalsky @Blaizzy

tshu-w:
"## What does this PR do?

Introduce the property first.

Part of #14188

Next steps:
1. Change `Trainer.log_dir` to `experiment_dir` and `ModelCheckpoint` Implementation #14640,
2. Update implementations of some Loggers, and maybe drop some unnecessary properties(`name`, `version`, etc)
3. Introduce `Trainer.experiment_dir` and deprecate `Trainer.log_dir`

When the `Logger.save_dir` is `None` (DummyLogger, CometLogger(online), Deprecated LoggerCollection, etc), the `Trainer.log_dir` is `None` either, and the checkpoints directory is `default_root_dir/name/version`.
https://github.com/Lightning-AI/lightning/blob/b33e3ef18aa8f50dd905e3762aaaa3117aa65705/src/pytorch_lightning/callbacks/model_checkpoint.py#L594-L604

Since there are many Loggers that can have a `save_dir` of None, I prefer to make `logger.experiment_dir` return a relative path. 
cc @awaelchli 

## Before submitting

## PR review

## Did you have fun?

"


carmocca:
If this is going to be sequenced so the properties are introduced first, and then used. I would like that you open the follow-up PR already so that we can see that things worked as expected (CI passed).

tshu-w:
> If this is going to be sequenced so the properties are introduced first, and then used. I would like that you open the follow-up PR already so that we can see that things worked as expected (CI passed).

Sure, I can open the following PR first. Should I make this to draft PR first.

justusschock:
Closed in favor of #14640

PR #14562, opened 2022-09-06
Rohit's App codebase overview

rohitgr7:
## What does this PR do?

Ignore the CI stuff. I did that to avoid running any tests for now

## Before submitting

## PR review

## Did you have fun?


src/lightning_app/core/flow.py
```diff
@@ -93,6 +93,7 @@ def __init__(self):
         """
         from lightning_app.runners.backends.backend import Backend
 
+        # Question: What is this state?
         self._state = set()
```
tchaton, line 97: This is the internal variables used to know if an attribute is part of the state. We are keeping only the attribute name.

src/lightning_app/core/flow.py
```diff
@@ -149,6 +150,8 @@ def __setattr__(self, name, value):
             elif isinstance(value, LightningWork):
                 self._works.add(name)
                 _set_child_name(self, value, name)
+                # Question: I do not understand, why are we removing it from the state
```
tchaton, line 153: We aren't removing it from the state. We are simply naming the component as {parent_name}.{attribute_name}.

src/lightning_app/core/flow.py
```diff
@@ -149,6 +150,8 @@ def __setattr__(self, name, value):
...
+                # and not doing that when value is a structure?
```
tchaton, line 154: The structure is responsible for naming its own children.

src/lightning_app/core/flow.py
```diff
@@ -180,6 +183,8 @@ def __setattr__(self, name, value):
             elif _is_json_serializable(value):
                 self._state.add(name)
 
+                # Question: Just wondering, should we even allow users to even change the values to
```
tchaton, line 186: In theory yes, we would need to create a BaseType users can rely upon to create special behaviours with the state like Payload for example.

src/lightning_app/core/flow.py
```diff
@@ -351,7 +355,7 @@ def schedule(
 
         .. code-block:: python
 
-            from lightning_app import LightningFlow
+            from lightning import LightningFlow
```
tchaton, line 358: We are in the lightning_app docs.

src/lightning_app/utilities/proxies.py
```diff
@@ -123,6 +123,7 @@ def __call__(self, *args, **kwargs):
                     # overriding the return value for next loop iteration.
                     _send_data_to_caller_queue(self.work, self.caller_queue, data, call_hash)
         if not self.parallel:
+            # Question: Didn't understand this exception. Can you improve it?
```
tchaton, line 126: Here is how it works.

When running a work with `parallel=False`, we can't simply continue the flow execution and therefore when the work hasn't completed.
In order to enable this, we raise an exception and re-run the flow from the beginning.

src/lightning_app/core/flow.py
rohitgr7: yes, that is happening inside `_set_child_name`, but I guess the name is removed from the state when an existing attribute is set to Flow/Work. But I don't think this is happening when it's set to any of the structures.

PR #14567, opened 2022-09-06
Use TorchVision's Multi-weight Support and Model Registration API on Lightning

datumbox:
Hey Lightning team,

Thanks a lot for all your work on this project. 

I'm raising this PR to raise awareness of two important API changes that happened on TorchVision. In v0.13 TorchVision released its [Multi-weight support API](https://pytorch.org/blog/introducing-torchvision-new-multi-weight-support-api/) which allowed us to deploy updated weights for the majority of the models. This replaced the `pretrained=True` parameter with a `weights=something` (`something` can be a string or an enum). In addition to this, TorchVision has recently introduced a new [Model Registration API](https://pytorch.org/blog/easily-list-and-initialize-models-with-new-apis-in-torchvision/) which allows the easier listing of models and weights. Its aim is to eliminate the somehow hacky idiom of using `torchvision.models.__dict__` to retrieve models and weights. This is expected to be released on v0.14.

I would love to hear your thoughts on the proposed changes. For full transparency, though the Multi-weight API is stable, the Model Registration is currently considered Beta. We've intentionally marked it as such to ensure that we can act upon the feedback of our partners and users. We are currently collecting feedback at https://github.com/pytorch/vision/issues/6365 and hoping to promote to stable before TorchVision v0.14. If you have any feedback or feature requests, we would be happy to discuss more. 

Please note that it's likely that you might decide not to merge this PR and that's totally OK. Perhaps this can be used as an example for future PRs. Finally, apologies if I put things on the wrong place or use idioms you don't favour,  I'm very new to Lightning's code-base.

Looking forward to your feedback.


awaelchli:
@datumbox If this is the recommended way by torchvision moving forward, I think we should go ahead with your changes :)
In the future, once our minimum torch version pushes also the minimum torchvision compatible version over the deprecation end, we can then remove again the backward compatibility.

src/pytorch_lightning/utilities/imports.py
```diff
@@ -42,6 +42,7 @@
 _RICH_AVAILABLE = package_available("rich") and compare_version("rich", operator.ge, "10.2.2")
 _TORCH_QUANTIZE_AVAILABLE = bool([eg for eg in torch.backends.quantized.supported_engines if eg != "none"])
 _TORCHVISION_AVAILABLE = package_available("torchvision")
+_TORCHVISION_GREATER_EQUAL_0_14 = compare_version("torchvision", operator.ge, "0.14.0")
```
awaelchli, line 45: Could we move this to the file where it is used directly? We try to move towards more localized definitions, see #11826.

awaelchli (approved):

src/pytorch_lightning/utilities/imports.py
datumbox: Thanks a lot for the feedback. I've moved `_TORCHVISION_GREATER_EQUAL_0_14` to `model_helpers.py`. Happy to make any other changes needed. :)

awaelchli: Great, thank you!

carmocca (approved):

Borda (approved):

datumbox:
@awaelchli @carmocca @Borda Thanks a lot for your help on this.

PR #14583, opened 2022-09-07
[CLI] Move storage from app prefix to project/app prefix

dmitsf:
## What does this PR do?

We decided to move app storage prefix from `app_id` to `project_id/app_id`.
This PR makes the change. The environment variable `LIGHTNING_STORAGE_PATH` will be set inside the workflow container, and we will take the storage path from this environment variable.
If it's not specified, it will use the previous logic for defining app storage prefix.

Yes! App storage prefix will be different for new apps.
@rlizzo @panos-is @rusenask please take a look should we check do we have something from previously created app storages in the new logic, e.g. here: https://github.com/Lightning-AI/lightning/blob/master/src/lightning_app/storage/orchestrator.py#L90:L92 ?

We will deprecate old-way created storages in ~1 month.

## Before submitting

## PR review

## Did you have fun?


src/lightning_app/storage/path.py
```diff
@@ -402,6 +403,36 @@ def path_to_work_artifact(path: Union[Path, pathlib.Path, str], work: "Lightning
...
+    """
+    return shared_storage_path_legacy() / "artifacts" / work.name
+
+
+def path_to_work_artifact_legacy(path: Union[Path, pathlib.Path, str], work: "LightningWork") -> pathlib.Path:
+    """This is a legacy version of path_to_work_artifact to support app storages with old prefixes for some time.
+
+    It will be deprecated soon. It duplicates the logic to make the deprecation easier.
+    """
+    return artifacts_path_legacy(work) / pathlib.Path(*pathlib.Path(path).absolute().parts[1:])
```
dmitsf, line 433: I think it will be better to have separate legacy functions, WDYT?

dmitsf: To make further removal easier

rlizzo:
> should we check do we have something from previously created app storages in the new logic, e.g. here: https://github.com/Lightning-AI/lightning/blob/master/src/lightning_app/storage/orchestrator.py#L90:L92 ?

I think for this we can actually get rid of that conditional and the old way of doing things entirely. This code would only ever run on newly created apps, not currently running apps that don't know the about the environment variable. 

We just need to start injecting the env var on the cloud (before this is merged so that e2e tests can pass) and call it a day :)

Borda (approved):

src/lightning_app/storage/path.py
```diff
@@ -385,6 +385,10 @@ def shared_storage_path() -> pathlib.Path:
     ``SHARED_MOUNT_DIRECTORY`` environment variable. In the cloud, the shared path will point to a S3 bucket. All Works
     have access to this shared dropbox.
     """
+    storage_path = os.getenv("LIGHTNING_STORAGE_PATH", "")
+    if storage_path != "":
+        return pathlib.Path(storage_path)
+
     bucket_name = os.getenv("LIGHTNING_BUCKET_NAME", "")
     app_id = os.getenv("LIGHTNING_CLOUD_APP_ID", "")
```
awaelchli, line 393: After the change now, when will this code path be triggered?

Maybe what that means is that from now on, `LIGHTNING_STORAGE_PATH` has the bucket name and app id already combined. In this case, I'm wondering why this code would still be needed? Maybe add a TODO here that once the transitioning phase is over, we want to remove it.

awaelchli (approved):
Could we add a tiny test case (or modify existing one) just to make sure we understand what the exact value of `LIGHTNING_STORAGE_PATH` would look like and how it's returned from this function.

awaelchli:
@dmitsf Just double checking, is the milestone correct, as this is a new addition/breaking change rather than a bugfix?

src/lightning_app/storage/path.py
dmitsf: Yes, a good point!
I'll add the test and the comment. We assume `LIGHTNING_STORAGE_PATH` will be passed from BE side, but we'd like to leave the old way for some time for compatibility reasons.

dmitsf:
> @dmitsf Just double checking, is the milestone correct, as this is a new addition/breaking change rather than a bugfix?

Thanks! Changed to `0.7`.

src/lightning_app/storage/path.py
dmitsf: We assume this will be activated by the BE release with corresponding BE change (setting the envvar to the container). In practice, we can expect it on Sept. 19-21th.

tchaton: Yes, let's add a TODO to remove the previous logic

tchaton (approved):
LGTM !

PR #14294, opened 2022-08-18
Set path filters in favor of the required-job action

carmocca:
## What does this PR do?

I added the [check-group](https://github.com/tianhaoz95/check-group) code into our Probot in https://github.com/carmocca/probot in https://github.com/Lightning-AI/lightning/pull/14621

Before this is merged, we need to **unmark all jobs as required but check-group, WIP, pre-commit, check-schema**

Closes #14507
Closes #14565 

### Does your PR introduce any breaking changes? If yes, please list them.

None

PR #14587, opened 2022-09-07
Update checkgroup config

carmocca:
## What does this PR do?

Part of #14294

Need to merge this into master to continue the above, as it can only read the config in the default branch (master)

This has no impact on CI atm

None

### DIFF
--- a/.github/checkgroup.yml
+++ b/.github/checkgroup.yml
@@ -3,6 +3,9 @@ custom_service_name: "Lightning CI required checker"
 # changes made in pull requests from different branches or forks are ignored. This means that changes to this file
 # will only be used after they are merged.
 subprojects:
+
+  # SECTION: pytorch_lightning
+
   - id: "CI: CircleCI"
     paths:
       - ".circleci/**"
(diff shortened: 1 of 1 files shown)


.github/checkgroup.yml
```diff
@@ -41,6 +41,25 @@ subprojects:
...
+      - "pl-slow (windows-2022, 3.7, 1.11)"
+      - "test-on-tpus"
+
+  - id: "lightning_lite"
+    paths:
+      - "requirements/lite/**"
+      - "src/lightning_lite/**"
+      - "tests/tests_lite/**"
+      - "setup.cfg"  # includes pytest config
+    checks:
```
Borda, line 62: I think there shall be also all major Pl tests as PL is now depending on Lite

carmocca: True. I'll do that in a follow-up because that would require updating the path filters for the PL jobs and lite is still WIP

awaelchli (approved):
I thought we undid checkgroup for a reason?

carmocca:
It's under development, but the config file can only be read off the default branch (master), so I need this merged to continue.

PR #14603, opened 2022-09-08
Use the `pull_request_target` workflow event

carmocca:
## What does this PR do?

Docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
This is so that forks can access the secrets.

I think this needs to be merged to work. This would explain why probot-cc and circleci did not run

### DIFF
--- a/.github/workflows/probot-auto-cc.yml
+++ b/.github/workflows/probot-auto-cc.yml
@@ -3,7 +3,7 @@ name: Probot
 on:
   issues:
     types: [labeled]
-  pull_request:
+  pull_request_target:
     types: [labeled, ready_for_review]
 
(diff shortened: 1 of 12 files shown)


.github/workflows/probot-auto-cc.yml
```diff
@@ -3,7 +3,7 @@ name: Probot
 on:
   issues:
     types: [labeled]
-  pull_request:
+  pull_request_target:
```
awaelchli, line 6: why is it pull request target here but just pull_request in other files?

awaelchli (approved):

Borda (approved):

PR #14626, opened 2022-09-09
Make the SLURM Preemption/Timeout Signal Configurable

### LINKED ISSUE #14610: Make the Slurm Signal Configurable
## 🚀 Feature

Lightning currently assumes that the slurm preemption signal is set to USR1, it would be nice to add an option to SLURMEnvironment to change this.

### Motivation

I use submitit to handle slurm submissions and they recently moved away from USR1 because of an NCCL conflict.

This breaks the requeuing in pytorch lightning. 

See also https://github.com/facebookincubator/submitit/issues/1709

### Pitch

`SLURMEnvironment` can be trivially extended to account for this with something like 

```
def __init__(self, auto_requeue: bool = True, signal: str = "USR1") -> None:
        super().__init__()
        self.auto_requeue = auto_requeue
        self.signal = signal
```

then `SignalConnector` ~ln 60 can be changed to something like:

```
sig = getattr(signal, "SIG" + environment.signal, None)
self._register_signal(sig, HandlersCompose(sigusr1_handlers))
```

cc @borda @awaelchli

Queuecumber:
## What does this PR do?

This PR adds a flag to the `SLURMEnvironment` class which allows users to change which signal will be watched for preemtion/timeout. 

Although SIGUSR1 is fairly standard, some libraries (e.g., submitit) which submit to SLURM use a different signal or users may have another compatibility issue with SIGUSR1 and therefore require the use of another signal. 

Fixes #14610 

None

## Before submitting

## PR review

## Did you have fun?


Queuecumber:
@awaelchli This is ready for review, I have two failing tests but I'm not sure whats causing it

Queuecumber:
One thing to discuss: 

The way I implemented this if the user specifies an invalid signal, that emits a warning message and the SignalConnector falls back to using SIGUSR1 

This could also be a fatal error of some sort (either exception and crash or error log and no signal listening)

docs/source-pytorch/clouds/cluster_advanced.rst
```diff
@@ -88,6 +88,18 @@ To train a model using multiple nodes, do the following:
 
         #SBATCH --signal=SIGUSR1@90
 
+    You can change this signal if your environment requires the use of a different one, for example
+
+    .. code-block:: bash
+
+        #SBATCH --signal=SIGHUP@90
+
+    Then, when you make your trainer, pass the ` signal` option to the :class:`~pytorch_lightning.plugins.environments.slurm_environment.SLURMEnvironment` plugin::
```
awaelchli, line 97:
```suggestion
    Then, when you make your trainer, pass the `signal` option to the :class:`~pytorch_lightning.plugins.environments.slurm_environment.SLURMEnvironment` plugin:
```

src/pytorch_lightning/plugins/environments/slurm_environment.py
```diff
@@ -30,9 +30,10 @@ class SLURMEnvironment(ClusterEnvironment):
             rescheduled gets determined by the owner of this plugin.
     """
 
-    def __init__(self, auto_requeue: bool = True) -> None:
+    def __init__(self, auto_requeue: bool = True, signal: str = "USR1") -> None:
```
awaelchli, line 33: Some considerations:
- We could support int here, i.e., signal: int = signal.SIGUSR1
- We could support both, i.e., signal: int | str = signal.SIGUSR1
- Is excluding the SIG prefix intuitive? Intuitively, I would have passed in `signal.SIGUSR1` or `"SIGUSR1"`. 

No strong opinions here, maybe I'm overthinking this.

src/pytorch_lightning/plugins/environments/slurm_environment.py
```diff
@@ -30,9 +30,10 @@ class SLURMEnvironment(ClusterEnvironment):
             rescheduled gets determined by the owner of this plugin.
```
awaelchli, line 30: Would you mind also adding the docstring for the new signal argument? Thanks :)

src/pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -51,19 +51,28 @@ def register_signal_handlers(self) -> None:
...
+        # Windows seems to have signal incompatibilities
         if not self._is_on_windows():
-            if sigusr1_handlers and not self._has_already_handler(signal.SIGUSR1):
-                self._register_signal(signal.SIGUSR1, HandlersCompose(sigusr1_handlers))
+            if isinstance(environment, SLURMEnvironment):
+                sigusr = getattr(signal, "SIG" + environment.signal, None)
+
+                if sigusr is None:
+                    log.warning(f"Unrecognized signal name SIG{environment.signal}, assuming SIGUSR1")
+                    sigusr = signal.SIGUSR1
```
awaelchli, line 64: I suggest to move this validation to the SLURMEnvironment directly. What do you think?
Then, we can assume that SLURMEnvironment.signal will always return a good one.

src/pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -51,19 +51,28 @@ def register_signal_handlers(self) -> None:
...
+            sigusr_handlers.append(self.slurm_sigusr_handler_fn)
             sigterm_handlers.append(self.sigterm_handler_fn)
 
-        # signal.SIGUSR1 doesn't seem available on windows
+        # Windows seems to have signal incompatibilities
         if not self._is_on_windows():
-            if sigusr1_handlers and not self._has_already_handler(signal.SIGUSR1):
-                self._register_signal(signal.SIGUSR1, HandlersCompose(sigusr1_handlers))
+            if isinstance(environment, SLURMEnvironment):
+                sigusr = getattr(signal, "SIG" + environment.signal, None)
```
awaelchli, line 60: See my considerations above. I think this logic could be simplified, by direct access via environment.signal, and the default (sigusr = signal.SIGUSR1) should also be set in the SLURMEnvironment IMO.

awaelchli (commented):
Overall approach and implementation looks good to me. Minor comments. Thanks for this contribution, and the testing on the cluster!

awaelchli:
> The way I implemented this if the user specifies an invalid signal, we emit a warning message and the SignalConnector falls back to using SIGUSR1

I think we should immediately check this in the init of SLURMEnvironment. Did you choose a warning for a reason? I.e., long waiting job in the queue and erroring out can be annoying to the user? If there is a concern, I think that's fine. Otherwise I would error out, as warnings could be easily overlooked or lost in logs.

awaelchli:
> I have two failing tests but I'm not sure whats causing it

Will look at it once we have resolved the discussions here. Probably a difference in the signal library between Python versions.

src/pytorch_lightning/plugins/environments/slurm_environment.py
Queuecumber: Yeah completely forgot that

src/pytorch_lightning/plugins/environments/slurm_environment.py
Queuecumber: I think going straight for `signal: int = signal.SIGUSR1` makes sense actually, and this simplifies the code quite a bit so less room for error, I'll make that change. I don't see any particular advantage for the string parameter.

tests/tests_pytorch/trainer/connectors/test_signal_connector.py
```diff
@@ -90,14 +90,37 @@ def test_auto_requeue_flag(auto_requeue):
...
     else:
         assert signal.getsignal(signal.SIGTERM) is signal.SIG_DFL
         assert signal.getsignal(signal.SIGUSR1) is signal.SIG_DFL
 
     connector.teardown()
 
 
+@RunIf(skip_windows=True)
+@pytest.mark.parametrize("auto_requeue", (True, False))
+@pytest.mark.parametrize("requeue_signal", [(signal.SIGUSR1, signal.SIGUSR2, signal.SIGHUP)])
```
carmocca, line 103: I wouldn't duplicate the test above just for checking this parametrization. The difference is only a single line, why not merge the tests?

carmocca (approved):

tests/tests_pytorch/trainer/connectors/test_signal_connector.py
Queuecumber: Initially it was to make sure the default argument was working as expected but that's not as important with the updated interface from the above discussions

docs/source-pytorch/clouds/cluster_advanced.rst
```diff
@@ -88,6 +88,18 @@ To train a model using multiple nodes, do the following:
 
         #SBATCH --signal=SIGUSR1@90
 
+    You can change this signal if your environment requires the use of a different one, for example
```
awaelchli, line 91: This step by step guide here is to show how to get started quickly with default settings. So I think it's better not to talk about customization here directly. 

I would move this new paragraph to the detail description in the section "Enable auto wall-time resubmitions" below. What do you think?

awaelchli (approved):

tests/tests_pytorch/trainer/connectors/test_signal_connector.py
```diff
@@ -78,8 +78,9 @@ def training_step(self, batch, batch_idx):
 
 @RunIf(skip_windows=True)
 @pytest.mark.parametrize("auto_requeue", (True, False))
-def test_auto_requeue_flag(auto_requeue):
-    trainer = Trainer(plugins=[SLURMEnvironment(auto_requeue=auto_requeue)])
+@pytest.mark.parametrize("requeue_signal", [(signal.SIGUSR1, signal.SIGUSR2, signal.SIGHUP)])
```
awaelchli, line 81:
```suggestion
@pytest.mark.parametrize("requeue_signal", [signal.SIGUSR1, signal.SIGUSR2, signal.SIGHUP])
```

This will fix the "expected int, not tuple" test error

Queuecumber: Oops missed that

justusschock (approved):

Queuecumber:
This is rebased on latest master and ready to go unless there is more feedback

src/pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -51,19 +51,24 @@ def register_signal_handlers(self) -> None:
...
 
-        # signal.SIGUSR1 doesn't seem available on windows
+        # Windows seems to have signal incompatibilities
         if not self._is_on_windows():
-            if sigusr1_handlers and not self._has_already_handler(signal.SIGUSR1):
-                self._register_signal(signal.SIGUSR1, HandlersCompose(sigusr1_handlers))
+            if isinstance(environment, SLURMEnvironment):
+                sigusr = environment.requeue_signal
+            else:
+                sigusr = signal.SIGUSR1
```
Borda, line 62:
```suggestion
            sigusr = environment.requeue_signal if isinstance(environment, SLURMEnvironment) elsesignal.SIGUSR1
```

Borda (approved):

Borda:
@awaelchli seems to be ready... :)

awaelchli:
Earlier I promised I would take care of the failing Windows tests: https://github.com/Lightning-AI/lightning/actions/runs/3039218876/jobs/4893893890

Which is what I'm doing now. Otherwise we can't merge with failing tests.

Queuecumber:
Ok I understand now, because python on windows doesn't even define those constants for the signals

rohitgr7 (approved):

awaelchli:
@Queuecumber Looks like we got the Windows cases figured out. 
Great work, thanks for contributing this improvement!

Queuecumber:
Thanks for merging, do you guys know what the ETA is for it showing up in an installable package (pre-release or full release)

I'm eager to use this in semi-production

awaelchli:
This feature will go in to 1.8, we are targeting end of September / beginning October. 

To use this immediately, you can install the build from the master branch: 
pip install https://github.com/Lightning-AI/lightning/archive/refs/heads/master.zip -U

PR #14629, opened 2022-09-09
Add easy access to `state_dict` in Lite module wrapper

### LINKED ISSUE #14571: Simplify state dict access on model wrapper in Lite
### Motivation

Currently, the user has to be aware of whether their model is wrapped or not. To save a checkpoint, they need to call
model.module to access the state dict. Example code:

```py
import torch

from pytorch_lightning.lite import LightningLite


class Lite(LightningLite):
    def run(self):
        model = torch.nn.Linear(2, 2)
        model = self.setup(model)
        print(model.state_dict().keys())

        # Current:
        # _forward_module.module.weight, _forward_module.module.bias
        # _original_module.weight, _original_module.bias

        # Expected:
        # weight, bias

        # Note: .module gives the expected keys
        print(model.module.state_dict().keys())


if __name__ == "__main__":
    Lite(accelerator="cpu", strategy="ddp", devices=2).run()

```

As you can see, not only do we get the prefixes on the keys, but also duplicate weight references, because internally we save a reference to the original module in addition to the ddp wrapper. 

This could also be seen as unintended behavior, and be fixed in a bug release. 

### Pitch

Implement the `state_dict` and `load_state_dict` methods on the wrapper, and redirect the call to the original module. 

### Alternatives

Keep as is, but make it clear in the docs that user needs to call `model.module` to access the original model.

awaelchli:
## What does this PR do?

Fixes #14571

## Breaking change

This is a breaking change for anyone who had to work-around the saving and loading with `model.module`, or mangling with the raw state-dict keys of `model.state_dict()`.

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/lite/wrappers.py
```diff
@@ -115,6 +117,26 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
...
+
+    @overload
+    def state_dict(self, *, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
+
+    def state_dict(
+        self, *args: Any, destination: Optional[T_destination] = None, prefix: str = "", keep_vars: bool = False
+    ) -> Optional[T_destination]:
+        return self._original_module.state_dict(
+            destination=destination,  # type: ignore[type-var]
```
awaelchli, line 132: The type hints in nn.Module are unfortunately a bit messed up

src/pytorch_lightning/lite/wrappers.py
```diff
@@ -115,6 +117,26 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
...
+    def state_dict(self, *, destination: T_destination, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
+
+    @overload
+    def state_dict(self, *, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
+
+    def state_dict(
+        self, *args: Any, destination: Optional[T_destination] = None, prefix: str = "", keep_vars: bool = False
+    ) -> Optional[T_destination]:
```
carmocca, line 130: Why optional?

src/pytorch_lightning/lite/wrappers.py
```diff
@@ -115,6 +117,26 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
...
+        return self._original_module.state_dict(
+            destination=destination,  # type: ignore[type-var]
+            prefix=prefix,
+            keep_vars=keep_vars,
+        )
+
+    def load_state_dict(self, state_dict: Mapping[str, Any], strict: bool = True) -> Any:
```
carmocca, line 137: Doesnt this always return None?

carmocca (approved):

src/pytorch_lightning/lite/wrappers.py
awaelchli: If I don't, mypy infers it can be optional. I'm just trying to override torch's method, which doesn't use mypy. If you have a better suggestion, please modify it.

src/pytorch_lightning/lite/wrappers.py
awaelchli: See nn.Module implementation, which has no types (torch 1.12). I'm adding what mypy infers. Please edit directly if you have another solution.

src/pytorch_lightning/lite/wrappers.py
justusschock: In torch they just don't provide any typing (including return types) as the types should be clear from overloads: https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L1374

justusschock (approved):

Borda (approved):

src/pytorch_lightning/lite/wrappers.py
rohitgr7: it returns a NamedTuple with missing keys

rohitgr7 (approved):

Borda:
@awaelchli, mind checking/resolving the remaining conversations? :otter:

src/pytorch_lightning/lite/wrappers.py
carmocca: Okay. Not sure how to fix this so resolving

src/pytorch_lightning/lite/wrappers.py
```diff
@@ -115,6 +118,26 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
...
+    @overload
+    def state_dict(self, *, destination: T_destination, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
+
+    @overload
+    def state_dict(self, *, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
+
+    def state_dict(
+        self, *args: Any, destination: Optional[T_destination] = None, prefix: str = "", keep_vars: bool = False
```
carmocca, line 130: I would remove `*args` since you aren't passing them. They are deprecated upstream anyways

```suggestion
        self, destination: Optional[T_destination] = None, prefix: str = "", keep_vars: bool = False
```

src/pytorch_lightning/lite/wrappers.py
```diff
@@ -115,6 +118,26 @@ def _convert_float_tensor(t: Tensor) -> Tensor:
         output = apply_to_collection(output, function=_convert_float_tensor, dtype=Tensor)
         return output
 
+    @overload
+    def state_dict(self, *, destination: T_destination, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
+
+    @overload
+    def state_dict(self, *, prefix: str = ..., keep_vars: bool = ...) -> T_destination:
+        ...
```
carmocca, line 127: Why do we define these overloads? I tried removing them and doesn't seem like anything changes

justusschock: this does help a lot with IDE completion and should enable us to remove the typing of the method's implementation at all as this already defines typing.

awaelchli: I'm sorry, I can't remove the types @justusschock, our mypy says: Missing type annotation for function etc.

> Why do we define these overloads? I tried removing them and doesn't seem like anything changes

@carmocca I get this: 

```
src/pytorch_lightning/lite/wrappers.py:121: error: Signature of "state_dict" incompatible with supertype "Module"  [override]
src/pytorch_lightning/lite/wrappers.py:121: note:      Superclass:
src/pytorch_lightning/lite/wrappers.py:121: note:          @overload
src/pytorch_lightning/lite/wrappers.py:121: note:          def [T_destination <: Dict[str, Any]] state_dict(self, *, destination: T_destination, prefix: str = ..., keep_vars: bool = ...) -> T_destination
src/pytorch_lightning/lite/wrappers.py:121: note:          @overload
src/pytorch_lightning/lite/wrappers.py:121: note:          def state_dict(self, *, prefix: str = ..., keep_vars: bool = ...) -> Dict[str, Any]
src/pytorch_lightning/lite/wrappers.py:121: note:      Subclass:
src/pytorch_lightning/lite/wrappers.py:121: note:          def [T_destination <: Dict[str, Any]] state_dict(self, destination: Optional[T_destination] = ..., prefix: str = ..., keep_vars: bool = ...) -> Optional[T_destination]

```

@carmocca @justusschock feel free to push any changes regarding typing, as long as it is valid. I don't care too much about the typing details here.

Issue #12690, opened 2022-04-10
CUDA kernel error for BaguaStrategy with algorithm="async"


awaelchli:
### To Reproduce

```py
import os

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer
from pytorch_lightning.strategies import BaguaStrategy


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)


def run():
    train_data = DataLoader(RandomDataset(32, 64), batch_size=2)

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        enable_model_summary=False,
        accelerator="gpu",
        devices=2,
        strategy=BaguaStrategy(algorithm="async")
    )
    trainer.fit(model, train_dataloaders=train_data)


if __name__ == "__main__":
    run()

```

**algorithm="gradient_all_reduce": no error**
**algorithm="decentralized": no error**

**algorithm="async":**
```
Failed: Cuda error kernels/bagua_kernels.cu:628 'no kernel image is available for execution on the device
```

**algorithm="bytegrad":**
```
Failed: Cuda error kernels/bagua_kernels.cu:285 'invalid device function'
```

**algorithm="low_precision_decentralized":**
```
Failed: Cuda error kernels/bagua_kernels.cu:597 'no kernel image is available for execution on the device'
```

### Expected behavior

No error.

### Environment

```bash
* CUDA:
        - GPU:
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
                - NVIDIA GeForce RTX 3090
[...]

PR #14632, opened 2022-09-09
CI: Use self-hosted Azure GPU runners

Borda:
## What does this PR do?

image

## Before submitting

## PR review

## Did you have fun?


.azure/gpu-tests-lite.yml
```diff
@@ -41,12 +41,12 @@ jobs:
     timeoutInMinutes: "20"
```
carmocca, line 41: Why add an empty README.md file?

.azure/gpu-tests-lite.yml
```diff
@@ -41,12 +41,12 @@ jobs:
...
     # how much time to give 'run always even if cancelled tasks' before stopping them
     cancelTimeoutInMinutes: "2"
-    pool: azure-jirka-spot
+    pool: lit-rtx-3090
     container:
       image: "pytorchlightning/pytorch_lightning:base-cuda-py3.9-torch1.12-cuda11.6.1"
       # default shm size is 64m. Increase it to avoid:
       # 'Error while creating shared memory: unhandled system error, NCCL version 2.7.8'
-      options: "--runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all --shm-size=512m"
+      options: "--runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=${Agent.Name##*_} --shm-size=2gb"
```
carmocca, line 49: What does this VISIBLE_DEVICES change do?

.azure/start.sh
```diff
@@ -5,6 +5,15 @@
 
 set -e
 
+# export all args as env variables
```
carmocca, line 8: Why is this PR mixing HPU changes with the GPU changes?

.azure/gpu-tests-lite.yml
Borda: I am trying to have up to 4 agents per machine where each is using a different GPU, so:
- agent 0: `CUDA_VISIBLE_DEVICES=0,1`
- agent 1: `CUDA_VISIBLE_DEVICES=2,3`
- agent 2: `CUDA_VISIBLE_DEVICES=4,5`
- agent 3: `CUDA_VISIBLE_DEVICES=6,7`

awaelchli:
I'm not super informed about this, just a random suggestion:

Instead of switching immediately to the self-hosted runner, we could add is as a second job first, running alongside the current one. This way we can observe its stability without impacting development when an unexpected issue occurs. Inspired by what Carlos did with the probot, letting us test and observe it first. 

What do you think?

Borda:
> Instead of switching immediately to the self-hosted runner, we could add is as a second job first, running alongside the current one. This way we can observe its stability without impacting development when an unexpected issue occurs. Inspired by what Carlos did with the probot, letting us test and observe it first.

Unfortunately, a pool can't be mixed with a scale-set and self-hosted... :otter:
BUT, with this merge, some (only new PRs or update from the master) will use this new pool so not all will be switched at one so we can observe it meantime... :koala: 
Also, we do not delete the actual pool so with a simple pool name we can roll back any time :rabbit:

.azure/gpu-tests-lite.yml
Borda: it was a trigger

Borda:
Not sure why the stand-alone check is failing when all particular tests seem to be passing :/
https://dev.azure.com/Lightning-AI/lightning/_build/results?buildId=99723&view=logs&j=8657dba9-5789-5a03-e474-09d653a3c932&t=258d7da4-3c1c-50bf-1a46-94c8219b4dd8
maybe @carmocca would know?

carmocca (approved):

.azure/gpu-tests-lite.yml
```diff
@@ -41,12 +41,14 @@ jobs:
     timeoutInMinutes: "20"
     # how much time to give 'run always even if cancelled tasks' before stopping them
     cancelTimeoutInMinutes: "2"
-    pool: azure-jirka-spot
+    pool: lit-rtx-3090
```
carmocca, line 44: Lite GPU tests are failing

.azure/gpu-tests.yml
```diff
@@ -90,28 +96,21 @@ jobs:
...
-      displayName: 'Get legacy checkpoints'
-      condition: eq(variables['continue'], '1')
-
-    - bash: python -m coverage run --source pytorch_lightning -m pytest
-      workingDirectory: src/pytorch_lightning
-      displayName: 'Testing: PyTorch doctests'
-      condition: eq(variables['continue'], '1')
-
-    - bash: python -m coverage run --source pytorch_lightning -m pytest --ignore benchmarks -v --junitxml=$(Build.StagingDirectory)/test-results.xml --durations=50
+    - bash: python -m pytest --no-header strategies/test_bagua_strategy.py::test_async_algorithm
```
carmocca, line 107: This Bagua test is segfaulting in the 3090s cc @wangraying

Borda:
@awaelchli, this is the related issue with Bagua #12690?

awaelchli:
Yes @Borda

wangraying:
Add @BinhangYuan @woqidaideshi

.azure/gpu-tests.yml
```diff
@@ -88,10 +92,20 @@ jobs:
       displayName: 'Install dependencies'
       condition: eq(variables['continue'], '1')
 
+    # TODO
```
carmocca, line 95: Instead of completely skipping Bagua installation, we can skip the problematic tests if we are in the newer GPU architecture

Borda: Good point, so do we have some safe tests if the tests would run... like try/catch, so it ll be a rather general solution regardless of the machine you run it on?

carmocca: Try catch won't work since segfaults just mess everything up.

I suggest skipping the tests instead

```python
    if torch.cuda.get_device_capability()[0] >= 8:
        pytest.skip("Bagua does not support this architecture")
```

Borda: I see; I am rather thinking instead of hardcode test just for one CUDA capacity to have in the CI flow check that the installed version is compatible given HW, which would yield in something like env variable, and this will be parsed in tests

```bash
BAGUA_GPU_COMPATIBLE=$(python -c "print(int(torch.cuda.get_device_capability()[0] < 8))")
```
and in code
```py
if not bool(os.environ.get("BAGUA_GPU_COMPATIBLE", 1)):
        pytest.skip("Bagua does not support this architecture")
```

carmocca: The Bagua test file can define this condition and all tests use it. Does t need to be an environment variable

Borda: Well, my point is the separate environment from code... so the model situation is that in a few weeks the compile Bagua with this GPU capacity, and we need to edit tests, and I would rather update the CI pipeline... not sure if I describe it well :(

justusschock: wouldn't it then work to simply skip installing bagua in those envs (as it wouldn't be valid anyways)? The tests should only be run if bagua is installed

carmocca: @justusschock AFAIK not all Bagua algorithms fail in the new architecture. Not installing Bagua means the strategy is now completely untested

> and I would rather update the CI pipeline

What's the issue with updating a test attribute? I find it confusing defining it as an environment variable. It's a leaky way to do it.

Borda: I see; I am just thinking if there was runtime to determine if the used HW is compatible with the installed Bagua version...
As you said, the compatibility raises segmentation error, see, we could have a trivial script to run in a separate CI step, and if it passes, we will run all Bagua, and if it fails, we skip all...
But you are right that atm would mean no Bagua will be tested
Also not sure if the CI step failed would propagate as complete workflow fail :(

justusschock: @carmocca I get that, but if it is segfaulting due to the gpu architecture, I wouldn't trust the results with other algorithms as well and might expect this to be flaky at some point.

carmocca: If we gets worse, we can always choose to leave it untested. But having some tests at least checks that the strategy runs at all.

tests/tests_pytorch/run_standalone_tests.sh
```diff
@@ -12,18 +12,18 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-set -e
+set -ex
 # THIS FILE ASSUMES IT IS RUN INSIDE THE tests/tests_<package> DIRECTORY
 
 # Batch size for testing: Determines how many standalone test invocations run in parallel
 # It can be set through the env variable PL_STANDALONE_TESTS_BATCH_SIZE and defaults to 6 if not set
-test_batch_size="${PL_STANDALONE_TESTS_BATCH_SIZE:-6}"
+test_batch_size="${PL_STANDALONE_TESTS_BATCH_SIZE:-1}"
```
carmocca, line 20:
```suggestion
test_batch_size="${PL_STANDALONE_TESTS_BATCH_SIZE:-6}"
```

tests/tests_pytorch/run_standalone_tests.sh
```diff
@@ -12,18 +12,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-set -e
+set -ex
```
carmocca, line 15:
```suggestion
set -e
```

.azure/gpu-tests.yml
```diff
@@ -106,18 +110,27 @@ jobs:
         FREEZE_REQUIREMENTS: 1
       displayName: 'Install dependencies'
 
+    - bash: |
+        CUDA_VERSION_MM=$(python -c "import torch ; print(''.join(map(str, torch.version.cuda.split('.')[:2])))")
+        CUDA_VERSION_BAGUA=$(python -c "print([ver for ver in [116,113,111,102] if $CUDA_VERSION_MM >= ver][0])")
+        pip install "bagua-cuda$CUDA_VERSION_BAGUA"
+        pip list | grep bagua
+      displayName: 'Install Bagua'
+      condition: eq(variables['continue'], '1')
```
carmocca, line 119: This has been removed

```suggestion
```

.azure/gpu-tests.yml
```diff
@@ -106,18 +110,27 @@ jobs:
         FREEZE_REQUIREMENTS: 1
       displayName: 'Install dependencies'
 
+    - bash: |
+        CUDA_VERSION_MM=$(python -c "import torch ; print(''.join(map(str, torch.version.cuda.split('.')[:2])))")
```
carmocca, line 114: Unused

```suggestion
```

otaj (approved):

Borda:
argument `--runtime=nvidia` was deprecated in favor of `--gpus=all`
see: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/4585

.azure/gpu-tests.yml
```diff
@@ -112,16 +120,17 @@ jobs:
...
+        python -c "import torch ; mgpu = torch.cuda.device_count() ; assert mgpu == 2, f'GPU: {mgpu}'"
         python requirements/pytorch/check-avail-strategies.py
         python requirements/pytorch/check-avail-extras.py
       displayName: 'Env details'
 
     - bash: bash .actions/pull_legacy_checkpoints.sh
       displayName: 'Get legacy checkpoints'
 
-    - bash: python -m coverage run --source pytorch_lightning -m pytest
+    - bash: python -m coverage run --source pytorch_lightning -m pytest .
```
carmocca, line 133: nit: this is the same

```suggestion
    - bash: python -m coverage run --source pytorch_lightning -m pytest
```

carmocca (approved):

.azure/gpu-tests-lite.yml
```diff
@@ -41,12 +41,16 @@ jobs:
...
+    variables:
+      DEVICES: $( python -c 'print("$(Agent.Name)".split("_")[-1])' )
     container:
       image: "pytorchlightning/pytorch_lightning:base-cuda-py3.9-torch1.12-cuda11.6.1"
       # default shm size is 64m. Increase it to avoid:
       # 'Error while creating shared memory: unhandled system error, NCCL version 2.7.8'
-      options: "--runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all --shm-size=512m"
+      # argument `--runtime=nvidia` was deprecated in favor of `--gpus=all`
+      # see: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/4585
```
carmocca, line 52:
```suggestion
```

.azure/gpu-tests.yml
```diff
@@ -67,12 +67,16 @@ jobs:
...
+    pool: lit-rtx-3090
+    variables:
+      DEVICES: $( python -c 'print("$(Agent.Name)".split("_")[-1])' )
     container:
       image: $(image)
       # default shm size is 64m. Increase it to avoid:
       # 'Error while creating shared memory: unhandled system error, NCCL version 2.7.8'
-      options: "--runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all --shm-size=512m"
+      # argument `--runtime=nvidia` was deprecated in favor of `--gpus=all`
+      # see: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/4585
```
carmocca, line 78: We don't need to mention this in the code

```suggestion
```

PR #14638, opened 2022-09-10
simplify storage import 

williamFalcon:
silly that this used to fail... makes documentation and code overly verbose.

```python
import lightning as L

L.app.storage.Drive(...)
```
Also, pip install was failing... so had to add this docs.txt file @Borda


Borda:
> Also, pip install was failing... so had to add this docs.txt file

How did it failed? 
Also why not use requirements/app/docs.txt?

PR #14640, opened 2022-09-11
Deprecated `Trainer.log_dir` with `Trainer.experiment_dir`

### LINKED ISSUE #14188: Introduce `Logger.experiment_dir`

tshu-w:
## What does this PR do?

Fixes #14188
Includes and closes #14508

Yes, make `Trainer.log_dir` to use `Logger.experiment_dir` and ModelCheckpoint use `Trainer.log_dir`  directly.

Next steps:
2. Update implementations of some Loggers, and maybe drop some unnecessary properties(`name`, `version`, etc)
3. Update trainer doc

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -46,7 +46,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a warning when the model passed to `LightningLite.setup()` does not have all parameters on the same device ([#14822](https://github.com/Lightning-AI/lightning/pull/14822))
 
 
-- Introduce `Logger.experiement_dir` ([#14188](https://github.com/Lightning-AI/lightning/pull/14508))
+- Introduce `Logger.experiement_dir` ([#14508](https://github.com/Lightning-AI/lightning/pull/14508))
 
 
 ### Changed
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -46,6 +46,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a warning when the model passed to `LightningLite.setup()` does not have all parameters on the same device ([#14822](https://github.com/Lightning-AI/lightning/pull/14822))
 
 
+- Introduce `Logger.experiement_dir` ([#14188](https://github.com/Lightning-AI/lightning/pull/14508))
+
 
 ### Changed
 
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -138,6 +138,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   * Deprecated the `pytorch_lightning.utilities.device_parser.parse_hpus` in favor of `pytorch_lightning.accelerators.hpu.parse_hpus`
 
 
+- Deprecated `Trainer.log_dir` in favor of `Trainer.experiement_dir` ([#14640](https://github.com/Lightning-AI/lightning/pull/14640))
+
+
 ### Removed
 
 - Removed the deprecated `Trainer.training_type_plugin` property in favor of `Trainer.strategy` ([#14011](https://github.com/Lightning-AI/lightning/pull/14011))
--- a/src/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/src/pytorch_lightning/callbacks/model_checkpoint.py
@@ -576,8 +576,7 @@ class ModelCheckpoint(Checkpoint):
         determine where to save checkpoints. The path for saving weights is set in this priority:
 
         1.  The ``ModelCheckpoint``'s ``dirpath`` if passed in
-        2.  The ``Logger``'s ``log_dir`` if the trainer has loggers
-        3.  The ``Trainer``'s ``default_root_dir`` if the trainer has no loggers
+        2.  The ``Trainer``'s ``experiment_dir``
 
         The path gets extended with subdirectory "checkpoints".
         """
@@ -585,7 +584,7 @@ class ModelCheckpoint(Checkpoint):
             # short circuit if dirpath was passed to ModelCheckpoint
             return
 
-        ckpt_path = os.path.join(trainer.log_dir, "checkpoints")
+        ckpt_path = os.path.join(trainer.experiment_dir, "checkpoints")
         ckpt_path = trainer.strategy.broadcast(ckpt_path)
 
         self.dirpath = ckpt_path
--- a/src/pytorch_lightning/callbacks/model_checkpoint.py
+++ b/src/pytorch_lightning/callbacks/model_checkpoint.py
@@ -585,21 +585,7 @@ class ModelCheckpoint(Checkpoint):
             # short circuit if dirpath was passed to ModelCheckpoint
             return
 
-        if len(trainer.loggers) > 0:
-            if trainer.loggers[0].save_dir is not None:
-                save_dir = trainer.loggers[0].save_dir
-            else:
-                save_dir = trainer.default_root_dir
-            save_dir = trainer.loggers[0].save_dir or trainer.default_root_dir
-            name = trainer.loggers[0].name
-            version = trainer.loggers[0].version
-            version = version if isinstance(version, str) else f"version_{version}"
-
-            ckpt_path = os.path.join(save_dir, str(name), version, "checkpoints")
-        else:
-            # if no loggers, use default_root_dir
-            ckpt_path = os.path.join(trainer.default_root_dir, "checkpoints")
-
+        ckpt_path = os.path.join(trainer.log_dir, "checkpoints")
         ckpt_path = trainer.strategy.broadcast(ckpt_path)
 
         self.dirpath = ckpt_path
--- a/src/pytorch_lightning/trainer/trainer.py
+++ b/src/pytorch_lightning/trainer/trainer.py
@@ -1865,10 +1865,7 @@ class Trainer(
     @property
     def log_dir(self) -> Optional[str]:
         if len(self.loggers) > 0:
-            if isinstance(self.loggers[0], TensorBoardLogger):
-                dirpath = self.loggers[0].log_dir
-            else:
-                dirpath = self.loggers[0].save_dir
+            dirpath = self.loggers[0].experiment_dir
         else:
             dirpath = self.default_root_dir
 
--- a/src/pytorch_lightning/trainer/trainer.py
+++ b/src/pytorch_lightning/trainer/trainer.py
@@ -279,7 +279,7 @@ class Trainer(
 
             logger: Logger (or iterable collection of loggers) for experiment tracking. A ``True`` value uses
                 the default ``TensorBoardLogger``. ``False`` will disable logging. If multiple loggers are
-                provided, local files (checkpoints, profiler traces, etc.) are saved in the ``log_dir`` of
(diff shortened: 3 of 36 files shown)


carmocca:
@tshu-w CI is red, have you checked the failing tests?

src/pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -591,20 +591,9 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
-            name = trainer.loggers[0].name
-            version = trainer.loggers[0].version
-            version = version if isinstance(version, str) else f"version_{version}"
-
-            ckpt_path = os.path.join(save_dir, str(name), version, "checkpoints")
         else:
-            # if no loggers, use default_root_dir
-            ckpt_path = os.path.join(trainer.default_root_dir, "checkpoints")
+            log_dir = trainer.log_dir if trainer.log_dir is not None else trainer.default_root_dir
+            ckpt_path = os.path.join(log_dir, "checkpoints")
```
awaelchli, line 596: We might want to introduce `Trainer.experiment_dir` for this purpose here, then deprecate the old `Trainer.log_dir`. For users who access trainer.log_dir in their code. 

What do you think?

tshu-w:
> @tshu-w CI is red, have you checked the failing tests?

I have fixed most of them ;)

I don't know how to fix the last one, if I add `print(logger.experiment)` between `fit` and `test`, the test pass on my side. https://github.com/Lightning-AI/lightning/blob/024e7b8204846ead790cf5a098e4a1ca61fcb171/tests/tests_pytorch/loggers/test_all.py#L119-L141

src/pytorch_lightning/callbacks/model_checkpoint.py
tshu-w: Good idea, but I find that this will bring many modifications, should I start another PR or follow this one?

awaelchli: Do you mean because trainer.log_dir is used in many places? I think that's fine. If you think it's too much, you could only do the changes in the source code and do the docs updates in a separate PR mabye.

tshu-w: I see, will look at it. BTW, it will be much appreciated if you could check the last failed test (`loggers/test_all.py::test_loggers_fit_test_all[CometLogger]`) when you are free, I'm not sure why it failed.

awaelchli: I gave it a try. Not 100% sure if my fix is correct, but if I'm looking at it correctly, it's just because the removed logic in model checkpoint caused the test to first access version before experiment, and because we mock a lot of things it didn't like that :)

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -40,6 +40,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `WandbLogger.download_artifact` and `WandbLogger.use_artifact` for managing artifacts with Weights and Biases ([#14551](https://github.com/Lightning-AI/lightning/issues/14551))
 
 
+- Introduce `Logger.experiement_dir` ([#14188](https://github.com/Lightning-AI/lightning/pull/14508))
+
+
```
Borda, line 45:
```suggestion
```

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -72,6 +75,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - In Lightning Lite, state-dict access to the module wrapper now gets passed through to the original module reference ([#14629](https://github.com/Lightning-AI/lightning/pull/14629))
 
 
+- Improved the error messaging when passing `Trainer.method(model, x_dataloader=None)` with no module-method implementations available ([#14614](https://github.com/Lightning-AI/lightning/pull/14614))
+
+
```
Borda, line 80:
```suggestion
```

src/pytorch_lightning/callbacks/model_checkpoint.py
tshu-w: Thx @awaelchli, I'm even less sure because I'm not familiar with the Trainer part yet, so I didn't modify the test directly. 

Also I have submitted the code to introduce `Trainer.experiment_dir`.

Borda:
@carmocca @awaelchli, mind checking this one? :otter:

tshu-w:
> @Borda I believe we are quite close to the 1.8 release and my capacity is already used up at the moment. In order to finish #14188 completely, we need to properly sequence all (three?) PRs that @tshu-w started.

I've submitted two PRs so far, and this one follows (or includes) https://github.com/Lightning-AI/lightning/pull/14508, so if there's no problem, we can just merge this one.

I'll check the failing test about docs this weekend.

tshu-w:
This PR is ready to review now.

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1863,18 +1862,27 @@ def model(self, model: torch.nn.Module) -> None:
...
+                save_dir = self.default_root_dir
+            dirpath = os.path.join(save_dir, self.loggers[0].experiment_dir)
         else:
             dirpath = self.default_root_dir
 
         dirpath = self.strategy.broadcast(dirpath)
         return dirpath
 
+    @property
+    def log_dir(self) -> str:
```
carmocca, line 1879: Where did we discuss that `log_dir` would be deprecated?

carmocca: If they have different purposes, then I don't think `log_dir` should be deprecated. There would be uses cases for each.

If they have the same purpose and just differ in implementation, then we could just change it without a deprecation and keep the `log_dir` name.
In the past, we've done changes to `log_dir` without going through a deprecation.

awaelchli: @carmocca The reasoning is the following:

I think we should make `log_dir` do what it does for our default logger, TensorBoard, meaning returning the full log dir `root / name / version`. This would mean calling log_dir after this PR is the same for these users, no breaking change.

BUT, the consideration for deprecation is only relevant for everyone else who doesn't use TB. Here, we have the option to 
- breaking change, no warning, just change it
- forward deprecation: warn that the behavior will change in the future. Problem: User can't do anything to avoid the warning
- Deprecate the entire log_dir: Same as above, but user can change to trainer.experiment_dir to avoid the deprecation warning

awaelchli: There is also the cheap alternative to masquerade it as a fix, i.e., pushing the change as "Fixed: Trainer.log_dir is now consistently returning the versioned log dir for all loggers".

carmocca: Thanks for the context. I think you should choose.

My only suggestion is that this is hard to understand. We would have `log_dir`, `experiment_dir`, `save_dir`, `default_root_dir` ... (any other dir?) and the relationship is not clear from just the names.
If the solution can avoid introducing a new name, I would vote for it. `log_dir` is for the `Trainer`, but would it make sense to have it also be the name for `Logger.log_dir`?

tshu-w: Hi, any update on it?
I vote for deprecate the entire `log_dir` or  keep `log_dir`

Borda:
@carmocca, what is missing here? :rabbit:

carmocca:
We need to decide the resolution to the conversation above

src/pytorch_lightning/trainer/trainer.py
carmocca: I lean towards the latter. @awaelchli what's your opinion? cc @justusschock too

awaelchli: @carmocca It is not just about picking the name. The main challenge is how to deliver the improved way without major breaking changes. We can imagine that logging and managing experiments is a core Lightning experience, so it will affect many users. Introducing experiment dir was meant to minimize that (existing code still works). I would also love to keep the same name log_dir, but how are we going to do it???

justusschock: I would also try to keep the naming as unified as possible and go with `Logger.log_dir`.

tshu-w:
Hi there, any updates on the resolution? I noticed that @awaelchli  submitted a pull request at https://github.com/Lightning-AI/lightning/pull/16663, but closed it.

Borda:
ping @awaelchli @justusschock @carmocca ^^

lantiga:
I’m against breaking changes in this case

tshu-w:
I think https://github.com/Lightning-AI/lightning/issues/14188 still needs to be resolved. If breaking changes are not supported, can we consider the other two solutions mentioned in https://github.com/Lightning-AI/lightning/pull/14640#discussion_r981731009?

lantiga:
100% with you there @tshu-w
What solution of the two would you recommend?

tshu-w:
Both are fine for me, it's up to you to discuss and decide.😃

PR #14660, opened 2022-09-12
CI: install more OS

Borda:
## What does this PR do?

adding more OS for install CI

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/.github/workflows/ci-pkg-install.yml
+++ b/.github/workflows/ci-pkg-install.yml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: true
       max-parallel: 1
       matrix:
-        os: [ubuntu-20.04, macOS-11, windows-2022]
+        os: [ubuntu-20.04, ubuntu-22.04, macOS-11, macOS-12, windows-2022]
         pkg: ["app", "lite", "pytorch"]
         python-version: [3.8]  # , 3.9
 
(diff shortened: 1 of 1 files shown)


carmocca:
Why not replace the old ones? Do you expect that one would fail and not the other?

Borda:
> Why not replace the old ones?

they are not too old, and people are still using them such as Ubuntu 20.04 is still major

> Do you expect that one would fail and not the other?

don't know, but this is a reaction to @williamFalcon issue last week where he had a problem installing `lightning` on his Apple M1... but probably the issue is related to the new chair arch - AMD not the OS :confused:

akihironitta (approved):
(ignore me if I'm wrong but) IMHO, variation in the Python version can impact more than in the OS version, so if we really need to extend the coverage, I would start with adding more Python versions.

Borda:
> (ignore me if I'm wrong but) IMHO, variation in the Python version can impact more than in the OS version, so if we really need to extend the coverage, I would start with adding more Python versions.

You are right that the Python version is also important; in such case we can consider adding tests for py3.10 as I would assume that changes are rather liners so testing the min and max version shall be sufficient :otter:

carmocca:
I would rather not add more CI stress (jobs) if they won't provide relevant coverage. As you said:

>  but probably the issue is related to the new chair arch - AMD not the OS

williamFalcon:
pip install lightning is failing on various python versions and machine types… we definitely need tests for this.

carmocca:
My point is that if the current additions (ubuntu-22.04, macOS-12) covered the issue experienced the jobs would be failing. But they aren't meaning they aren't testing anything meaningful at the moment.

carmocca:
We cannot test mac Mx chips because there's no cloud provider, but we can try adding a job for 3.10

awaelchli (approved):

akihironitta:
> My point is that if the current additions (ubuntu-22.04, macOS-12) covered the issue experienced the jobs would be failing. But they aren't meaning they aren't testing anything meaningful at the moment.

(I approved this change, but) I now totally agree with you.

PR #14681, opened 2022-09-13
Standalone Lite: DataParallel Strategy

awaelchli:
## What does this PR do?

Adds the DP strategy for Lite. It is very similar to the one in PL. It will be one of the easiest to "migrate" when we later make PL depend on the implementation in Lite.

## Before submitting

## PR review

## Did you have fun?


justusschock (approved):

src/lightning_lite/strategies/dp.py
```diff
@@ -0,0 +1,100 @@
...
+    @property
+    def world_size(self) -> int:
+        return 1
+
+    def setup_module(self, module: Module) -> DataParallel:
+        """Wraps the given model into a :class:`~torch.nn.parallel.DataParallel` module."""
+        return DataParallel(module=module, device_ids=self.parallel_devices)
+
+    def module_to_device(self, module: Module) -> None:
+        module.to(self.root_device)
```
carmocca, line 74: Noticing now that these are `module` instead of `model` and that the APIs are slightly different here vs in PL. Will they be reconcilliated eventually?

src/lightning_lite/strategies/dp.py
```diff
@@ -0,0 +1,100 @@
...
+    def local_rank(self) -> int:
+        return 0
+
+    @property
+    def node_rank(self) -> int:
+        return 0
+
+    @property
+    def world_size(self) -> int:
+        return 1
```
carmocca, line 67: Technically these are not necessary since `cluster_environment=None` so the parent's implementations are enough

```suggestion
```

src/lightning_lite/strategies/dp.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def setup_module(self, module: Module) -> DataParallel:
+        """Wraps the given model into a :class:`~torch.nn.parallel.DataParallel` module."""
+        return DataParallel(module=module, device_ids=self.parallel_devices)
+
+    def module_to_device(self, module: Module) -> None:
+        module.to(self.root_device)
+
+    def batch_to_device(self, batch: Any, device: Optional[torch.device] = None, dataloader_idx: int = 0) -> Any:
```
carmocca, line 76: Lite should remove the dataloader_idx? The base class does not have it

```suggestion
    def batch_to_device(self, batch: Any, device: Optional[torch.device] = None) -> Any:
```

carmocca (approved):

src/lightning_lite/strategies/dp.py
awaelchli: Thank you! Small mishaps like these happen, until we have the full integration with all tests running. Fingers crossed!

src/lightning_lite/strategies/dp.py
awaelchli: Yes, purposefully named module for clarity. 
PL will eventually also call this method, and feed into it the LightningModule (or wrapped LM if applicable). 

I felt comfortable doing the name change, especially because we needed to do another change anyway: The module as input argument. This is missing in PL's version    ` def model_to_device(self) -> None:`

Borda (approved):

PR #14692, opened 2022-09-13
Standalone Lite: Connector

awaelchli:
## What does this PR do?

Adds the Connector (in PL called AcceleratorConnector) for Lite. The responsibility remains the same as in PL:
It selects the accelerator, configures plugins, selects the strategy, and validates the user inputs regarding all these.

- The logic is all based on the PL AcceleratorConnector
- All code was copied over, then all code that does not work with Lite was deleted (e.g., apex configuration, horovod logic, layer sync logic, etc.)
- Misconfiguration was replaced with Python error types
- No validation logic was removed or changed. Error messages that mentioned Trainer now mention Lite. 
- All tests in `tests/tests_pytorch/trainer/connectors/test_accelerator_connector.py` were copied to Lite and re-written.

## Before submitting

## PR review

## Did you have fun?

cc @tchaton @rohitgr7 @justusschock @awaelchli @carmocca @kaushikb11 @ninginthecloud


src/lightning_lite/connector.py
```diff
@@ -0,0 +1,616 @@
...
+        B. Strategy > Accelerator/precision/plugins
+    """
+    def __init__(
+        self,
+        accelerator: Optional[Union[str, Accelerator]] = None,
+        strategy: Optional[Union[str, Strategy]] = None,
+        devices: Optional[Union[List[int], str, int]] = None,
+        num_nodes: int = 1,
+        precision: Union[int, str] = 32,
+        amp_type: str = "native",  # TODO(lite): Remove this unused argument and adjust logic below
```
awaelchli, line 108: TODO

src/lightning_lite/connector.py
```diff
@@ -0,0 +1,596 @@
...
+)
+from lightning_lite.strategies.ddp_spawn import _DDP_FORK_ALIASES
+from lightning_lite.utilities import _StrategyType, device_parser, rank_zero_deprecation, rank_zero_info, rank_zero_warn
+from lightning_lite.utilities.imports import _HPU_AVAILABLE, _IPU_AVAILABLE, _IS_INTERACTIVE, _TPU_AVAILABLE
+
+_PLUGIN = Union[Strategy, Precision, ClusterEnvironment, CheckpointIO]
+_PLUGIN_INPUT = Union[_PLUGIN, str]
+
+
+class _Connector:
```
justusschock, line 63: why is this class protected? IMO it shouldn't be, just it's instance in Lite should be protected.

src/lightning_lite/connector.py
```diff
@@ -0,0 +1,596 @@
...
+    def __init__(
+        self,
+        accelerator: Optional[Union[str, Accelerator]] = None,
+        strategy: Optional[Union[str, Strategy]] = None,
+        devices: Optional[Union[List[int], str, int]] = None,
+        num_nodes: int = 1,
+        precision: Union[int, str] = 32,
+        plugins: Optional[Union[_PLUGIN_INPUT, List[_PLUGIN_INPUT]]] = None,
+        tpu_cores: Optional[Union[List[int], str, int]] = None,  # deprecated
+        gpus: Optional[Union[List[int], str, int]] = None,  # deprecated
```
justusschock, line 103: do we really want to carry over this deprecation? Having a newly designed package would be the perfect time of getting rid of stuff like this.

src/lightning_lite/connector.py
```diff
@@ -0,0 +1,596 @@
...
+    ) -> None:
+        # 1. Parsing flags
```
justusschock, line 105: Maybe not in this PR, but later on: Should we eventually split out several connectors? One for plugins, one for strategies, one for accelerators? 

The logic is already contained within separated functions and would maybe allow further customization (which might also come handy when extending this later for PL) as well as a less bloated general connector.

justusschock (approved):
LGTM, just some general questions

src/lightning_lite/connector.py
awaelchli: My reasoning here was simply that we want to communicate this is not an internal class that a user should customize, override or use in anyway outside of Lite. I'm fine either way, @carmocca for second opinion.

src/lightning_lite/connector.py
awaelchli: It would be nice not to, but it was already in Lite beforehand. So I see two options:

A) In the new package, delete these. The Lite class that will stay in PL for backward compatibility will have additional logic to handle this (using the old connector?). 
B) We carry it over but shorten the deprecation period from 2.0 to something earlier (is this breaking our promises??)

src/lightning_lite/connector.py
awaelchli: Sounds nice in theory. I don't know if it would work though with the current complex logic, as there can be relationships between strategies and plugins that need validation to reference these and rely on a particular order of configuration.

src/lightning_lite/connector.py
justusschock: I'm not saying that they should be public in general, I'm just saying that they should be public w.r.t the namespace of this file.

src/lightning_lite/connector.py
justusschock: I think it would, as we already encapsulated it into the current six steps. We just need to figure out, what to pass to which function.

src/lightning_lite/connector.py
justusschock: I would go for A), have a tiny wrapper that basically only translates `tpu_cores=X` to `accelerator='tpu', devices=X` and `gpus=X` to `accelerator='gpu', devices=X`. This should be a minimal helper which could then be reused for PL

src/lightning_lite/connector.py
awaelchli: Yes why not, let's try to do it as a follow up, as time and priorities allow :)

src/lightning_lite/connector.py
carmocca: Since breaking changes and deprecations are such a large issue, I prefer to be safe and prefix everything that somebody could ever think of importing.

What you say makes some sense but it's impossible to communicate to the users other than through the documentation

Nowadays nobody looks at imports, as the IDE usually manages them. Nobody checks that they imported:
`from lightning_lite.connector import Connector` (supposedly protected externally, public internally)
instead of
`from lightning_lite import Connector` (public externally and internally)

justusschock: I do :D

but fine by me :)

src/lightning_lite/connector.py
awaelchli: Waiting to implement this in case other reviewers want to comment first.

PR #14703, opened 2022-09-14
Introduce Upload File endpoint

tchaton:
## What does this PR do?

This PR enables us to get rid of the File Server to upload files to the Shared Storage and reduce cost of the Training Studio Cloud App.

Furthermore, I made some tests and it seems the async mechanism enables to keep fetching the Rest API while uploading a file.

https://user-images.githubusercontent.com/12861981/190347318-1796392d-f496-486b-9c2b-12986952b85b.mov

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?


src/lightning_app/core/api.py
```diff
@@ -234,6 +236,29 @@ async def post_state(
...
+            drive = Drive(
+                "lit://uploaded_files",
+                component_name="file_server",
+                allow_duplicates=True,
+                root_folder=tmp,
+            )
+            tmp_file = os.path.join(tmp, filename)
+
+            with open(tmp_file, "wb") as f:
+                while content := await uploaded_file.read(10240):
```
awaelchli, line 253: What is the magic number? Adding a comment might help

src/lightning_app/core/api.py
```diff
@@ -234,6 +236,29 @@ async def post_state(
...
+                    f.write(content)
+
+            drive.put(filename)
+    finally:
+        uploaded_file.file.close()
```
awaelchli, line 258: Here we are closing the file, but where do we open it? What if an unopened file gets passed in?

tests/tests_app/core/test_lightning_api.py
```diff
@@ -446,6 +447,27 @@ def test_configure_api():
         f"http://localhost:{APP_SERVER_PORT}/api/v1/request", data=InputRequestModel(name="hello").json()
     )
     assert response.json() == {"name": "hello", "counter": 1}
+
+    with open(__file__, "rb") as f:
```
awaelchli, line 451: I find it strange to use the file we are testing in. Maybe use a tempfile instead?

tests/tests_app/core/test_lightning_api.py
```diff
@@ -446,6 +447,27 @@ def test_configure_api():
...
+        response = requests.put(
+            f"http://localhost:{APP_SERVER_PORT}/api/v1/upload_file/example.txt", files={"uploaded_file": f}
```
awaelchli, line 453: We also need to test when example.txt is omitted (the filename gets taken from the object)

awaelchli (approved):

awaelchli:
Is it needed to add docs anywhere for this?

ethanwharris (approved):

src/lightning_app/core/api.py
Borda: Is this 3.8+ syntax?

src/lightning_app/core/api.py
tchaton: I think this is a FastAPI requirement.

tchaton: Removed.

tests/tests_app/core/test_lightning_api.py
tchaton: I tried but didn't manage to do so. I thought it didn't manage more.

tests/tests_app/core/test_lightning_api.py
tchaton: I removed it. It seems it doesn't work.

kaushikb11 (approved):

src/lightning_app/core/api.py
```diff
@@ -234,6 +236,29 @@ async def post_state(
...
+        )
+        tmp_file = os.path.join(tmp, filename)
+
+        with open(tmp_file, "wb") as f:
+            done = False
+            while not done:
+                # Note: The 8192 number doesn't have a strong reason.
+                content = await uploaded_file.read(8192)
+                f.write(content)
+                done = content == b""
```
kaushikb11, line 256: Should we catch the exception if there's one while reading the file and add a helpful message?

tchaton: I will do when I had to others

tchaton: But it is async, so it is catched already.

src/lightning_app/core/api.py
```diff
@@ -234,6 +236,29 @@ async def post_state(
...
+
+        drive.put(filename)
```
hhsecond, line 258: should we do `drive.put` in a different process? This might block server thread and since we are using only a single thread for the whole app, it might have effect on healthz, app server end point etc

hhsecond (approved):
looks good. Just one question

src/lightning_app/core/api.py
tchaton: I think it is fine for now. We can always optimize it if needed. I tried with the Training Studio App and it didn't seem to affect the server as shown in the video above.

PR #14712, opened 2022-09-15
Tune the checkgroup config

carmocca:
## What does this PR do?

Minor tweaks to the checkgroup config.

None

### DIFF
--- a/.github/checkgroup.yml
+++ b/.github/checkgroup.yml
@@ -31,6 +31,7 @@ subprojects:
       - "setup.cfg"  # includes pytest config
       - ".github/workflows/docs-*.yml"
     checks:
+      # Note: updates here should be applied ot the lightning_lite group
       - "pl-conda (3.8, 1.10)"
       - "pl-conda (3.8, 1.9)"
       - "pl-conda (3.9, 1.11)"
@@ -54,7 +55,8 @@ subprojects:
       - "pl-slow (macOS-11, 3.7, 1.11)"
       - "pl-slow (ubuntu-20.04, 3.7, 1.11)"
       - "pl-slow (windows-2022, 3.7, 1.11)"
-      - "test-on-tpus"
+      # TODO: since this job cannot run on forks, it cannot be required or it will block all PL PRs from forks
+      #- "test-on-tpus"
 
   - id: "pytorch_lightning: Conda"
     paths:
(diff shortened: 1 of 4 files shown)


.github/checkgroup.yml
```diff
@@ -163,8 +165,25 @@ subprojects:
...
+      - "pl-cpu (macOS-11, 3.7, oldest, stable)"
+      - "pl-cpu (ubuntu-20.04, 3.10, latest, stable)"
+      - "pl-cpu (ubuntu-20.04, 3.7, latest, stable)"
+      - "pl-cpu (ubuntu-20.04, 3.7, oldest, stable)"
+      - "pl-cpu (windows-2022, 3.10, latest, stable)"
+      - "pl-cpu (windows-2022, 3.7, latest, stable)"
+      - "pl-cpu (windows-2022, 3.7, oldest, stable)"
+      - "make-doctest (pytorch)"
+      - "make-html (pytorch)"
+      - "pytorch-lightning (GPUs)"
```
carmocca, line 182: Lite does not support HPU and IPU yet, so I haven't added their checks as required for Lite changes

Please confirm if that's okay @awaelchli. Maybe you want them to still run in each Lite PR for sanity

awaelchli: Hmm, we have utilities in Lite that are used by HPU and or IPU on the PL side. If this code would change in a bad way, we wouldn't see these checks fail on the PR, right? The risk is small, so I don't mind. On the other hand, IPU jobs are quite fast, so adding them wouldn't hurt I guess (except when the queue gets long)?

carmocca: The risk is small. If it was up to me, I would not add them until they break and we regret not having added them :stuck_out_tongue: 

I'll leave the decision up to you. Not adding them will mean faster development as you'd have less CI overhead.

awaelchli (approved):

.github/checkgroup.yml
```diff
@@ -31,6 +31,7 @@ subprojects:
       - "setup.cfg"  # includes pytest config
       - ".github/workflows/docs-*.yml"
     checks:
+      # Note: updates here should be applied ot the lightning_lite group
```
otaj, line 34:
```suggestion
      # Note: updates here should be applied to the lightning_lite group
```

otaj (approved):

Borda (approved):

PR #14717, opened 2022-09-15
Drop gatekeeper CI checks

carmocca:
## What does this PR do?

Drop the gatekeeper checks. With proper code-owners we shouldn't need it.
The job is not marked as required
The path filters are not correct.

Discussed with Jirka offline

None


otaj (approved):

Borda:
> Discussed with Jirka offline

I think we both were online on slack :otter:

Borda (approved):

PR #12069, opened 2022-02-23
Unify checks

justusschock:
## What does this PR do?
fixes  #10821 

Generalizes Type-Checks

### Does your PR introduce any breaking changes? If yes, please list them.
Yes, deprecates `trainer.amp_backend` and `MixedPrecisionPlugin.backend`

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda

PR #14792, opened 2022-09-19
Introduce primitives for input/output dtype conversion in Lite Precision

awaelchli:
## What does this PR do?

The Precision plugin exposes API to convert model input and outputs to the appropriate precision dtype. 
Note that this is just for Lite for now, as precision integration in PL is still ongoing. 

Follow up to #14815

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @awaelchli @rohitgr7 @carmocca @akihironitta


src/lightning_lite/plugins/precision/deepspeed.py
```diff
@@ -53,18 +56,23 @@ def __init__(self, precision: Union[str, int], amp_type: str, amp_level: Optiona
 
             amp_level = amp_level or "O2"
 
-        supported_precision = (PrecisionType.HALF, PrecisionType.FLOAT, PrecisionType.BFLOAT)
-        if precision not in supported_precision:
```
awaelchli, line 57: This was a bug, right? Comparing the precision (int) with the enum strings would never work here, i.e., the if-condition would be false.

src/lightning_lite/wrappers.py
```diff
@@ -96,26 +96,12 @@ def module(self) -> nn.Module:
...
         """Casts all inputs to the right precision and handles autocast for operations in the module forward
         method."""
-        precision = self._precision_plugin.precision
-        precision_to_type = {
-            "bf16": torch.bfloat16,
-            16: torch.float16,
-            32: torch.float32,
-            64: torch.float64,
-        }
-        # TODO: let the precision plugin handle the conversion
```
awaelchli, line 106: This is the TODO we want to resolve in this PR.

src/pytorch_lightning/strategies/utils.py
```diff
@@ -34,3 +37,13 @@ def _call_register_strategies(registry: _StrategyRegistry, base_module: str) ->
     for _, mod in getmembers(module, isclass):
         if issubclass(mod, Strategy) and _is_register_method_overridden(mod, Strategy, "register_strategies"):
             mod.register_strategies(registry)
+
+
+def _fp_to_half(tensor: torch.Tensor, precision: PrecisionType) -> torch.Tensor:
```
awaelchli, line 42: Moves back to where it came from.

src/lightning_lite/plugins/precision/deepspeed.py
carmocca: No. Our enums are magic :sparkles: and convert to str in the equality check: https://github.com/Lightning-AI/utilities/blob/86e9e48e453ae9690448a0d3cef320d3dc4bd777/src/lightning_utilities/core/enums.py#L23

src/lightning_lite/plugins/precision/deepspeed.py
```diff
@@ -53,18 +56,23 @@ def __init__(self, precision: Union[str, int], amp_type: str, amp_level: Optiona
...
+        supported_precision = ("16", "32", "bf16")
+        if str(precision) not in supported_precision:
             raise ValueError(
                 f"`precision={precision!r})` is not supported in DeepSpeed."
-                f" `precision` must be one of: {(x.value for x in supported_precision)}."
+                f" `precision` must be one of: {', '.join(supported_precision)}."
             )
 
         super().__init__()
         self.precision = precision
```
carmocca, line 67: If you stop using the enums to check. Then IMO this should be saved as a string to avoid potential issues with checking `str == int`

`self.precision = str(self.precision)`

And in the dictionaries: `{"bf16": torch.bfloat16, "16": torch.float16, "32": torch.float32}`

carmocca (approved):
Merged #14823

src/lightning_lite/plugins/precision/precision.py
```diff
@@ -34,6 +36,20 @@ def forward_context(self) -> Generator[None, None, None]:
         """A contextmanager for managing model forward/training_step/evaluation_step/predict_step."""
         yield
 
+    def convert_input(self, data: Tensor) -> Tensor:
+        """Convert model inputs (forward) to the floating point precision type of this plugin.
+
+        This is a no-op for tensors that are not of floating-point type or already have the desired type.
+        """
+        return _convert_fp_tensor(data, torch.float32)
```
carmocca, line 44: Couldn't we do this in the base

```suggestion
        precision_to_type = {"bf16": torch.bfloat16, 16: torch.float16, 32: torch.float32}
        dst_type = precision_to_type[self.precision]
        return _convert_fp_tensor(data, dst_type)
```

so that none of the subclasses need to override it?

src/lightning_lite/plugins/precision/deepspeed.py
awaelchli: Yes, I had the impression we decided to move away from the enums a while ago, although I don't remember all the arguments. 

I agree it would be good to only have the string type. I need to check where the precision attribute is used, connector etc.

src/lightning_lite/plugins/precision/precision.py
awaelchli: Ideally, I want to mark theses interfaces abstract to force all of them to implement it. But I can't as this would bring more breaking changes on the PL side. 

Since some plugins can only manage a subset of precision types, I think it would be wrong if the super implementation could handle more than allowed.

src/pytorch_lightning/strategies/utils.py
awaelchli: Todo @awaelchli 
Remove after Carlos changes

src/lightning_lite/plugins/precision/deepspeed.py
awaelchli: That explains it .Thanks

src/lightning_lite/plugins/precision/deepspeed.py
justusschock: See #12069 which is related

src/lightning_lite/plugins/precision/precision.py
justusschock: I prefer having that as abstract interface.

src/lightning_lite/plugins/precision/precision.py
```diff
@@ -34,6 +36,20 @@ def forward_context(self) -> Generator[None, None, None]:
...
+
+    def convert_output(self, data: Tensor) -> Tensor:
```
justusschock, line 46: what if the input was not of default type? Will we still convert the output to default type or should that rather be converted to the inputs type?

src/lightning_lite/plugins/precision/deepspeed.py
justusschock: @carmocca @awaelchli while on it and while adding a new package, should we also integrate https://github.com/Lightning-AI/lightning/issues/9956#issuecomment-1207246337 to avoid the deprecation cycle here?

src/lightning_lite/plugins/precision/precision.py
awaelchli: Good point Justus. Maybe we should not convert the output at all, or do the conversion back to input as you said? For example, using double precision, we would downgrade to lower precision (32) but maybe user needs the double for further loss calculations. 

In both these cases, we wouldn't need a convert_output() method and would handle it directly in the lite forward wrapper.

awaelchli: In any case, right now the implementation is in parity to what was before. But no saying the previous way was more correct.

src/lightning_lite/plugins/precision/precision.py
awaelchli: Some options:

1. Current
2. Add abstract now: Small breaking change for custom plugins (who makes them?). Plus, when #14798 merges, we need to provide the implementations for PL plugins
3. Add `TODO(lite): make abstract` to delay it

src/lightning_lite/connector.py
```diff
@@ -58,6 +59,7 @@
 
 _PLUGIN = Union[Precision, ClusterEnvironment, CheckpointIO]
 _PLUGIN_INPUT = Union[_PLUGIN, str]
+_PRECISION_INPUT = Literal[16, 32, 64, "bf16"]
```
carmocca, line 62: I would also include `"16", "32", "64"`

src/lightning_lite/plugins/precision/deepspeed.py
```diff
@@ -43,7 +46,7 @@ class DeepSpeedPrecision(Precision):
             If unsupported ``precision`` is provided.
     """
 
-    def __init__(self, precision: Union[str, int], amp_type: str, amp_level: Optional[str] = None) -> None:
+    def __init__(self, precision: Literal[16, 32, "bf16"], amp_type: str, amp_level: Optional[str] = None) -> None:
```
carmocca, line 49: Same suggestion

src/lightning_lite/plugins/precision/deepspeed.py
carmocca: > I had the impression we decided to move away from the enums a while ago, although I don't remember all the arguments.

The argument is that they are not easy to extend.

Let's continue the discussion about planning for #9956 offline.

> I agree it would be good to only have the string type. I need to check where the precision attribute is used, connector etc.

From this comment I assume you still need to set it as a str

src/lightning_lite/plugins/precision/precision.py
carmocca: If the PL classes will not define it. It's boilerplate-y to mark it as abstract

src/lightning_lite/plugins/precision/precision.py
carmocca: The safest thing IMO would be to convert to the original data precision. Otherwise not do this at all.

Can you create an issue about this for tracking?

src/lightning_lite/plugins/precision/deepspeed.py
awaelchli: Regarding this and the other comments, I spent like an hour on this and it's spreading like a cancer. I have to break it out because the mypy is just forcing me to do it everywhere. 
#14857

src/lightning_lite/connector.py
awaelchli: Please check my comment below. If I include string alternatives here, now I need to update the comparisons in the connector, which then leads to more typing updates, which then leads to etc. etc.

src/lightning_lite/plugins/precision/precision.py
awaelchli: Created issue #14859

Removed the convert_output method. We can introduce it later should we have a decision. Right now, I'll just leave the output as it was handled previously.

src/lightning_lite/plugins/precision/deepspeed.py
carmocca: If it's becoming a larger unnecessary refactor, you could also revert back to Union[str, int] and still use the `PrecisionType` magic enums for equalities

awaelchli: yes, reverted. using old logic now. can be re-done in https://github.com/Lightning-AI/lightning/pull/14857

carmocca (approved):
Re-approving after the latest changes

tests/tests_lite/plugins/precision/test_native_amp_integration.py
```diff
@@ -67,6 +67,6 @@ def after_backward(self, model):
     ],
 )
 def test_native_mixed_precision(accelerator, precision, expected_dtype):
-    lite = NativeMixedPrecisionBoringLite(accelerator=accelerator, precision=16)
+    lite = NativeMixedPrecisionBoringLite(accelerator=accelerator, precision=precision)
```
awaelchli, line 70: Including a small fix here. This was failing locally on the rtx card where bf16_cuda is available (not here in CI). And the test was obviously meant to take the parametrized argument as input.

otaj (approved):

PR #12828, opened 2022-04-20
Fixed encoding issues on terminals that do not support unicode characters

alvitawa:
Fixes https://github.com/PyTorchLightning/pytorch-lightning/issues/12823

PR #14795, opened 2022-09-20
Fix icon encoding in PL App pretty-print

### LINKED ISSUE #14226: Icons in PL app CLI output broken in some terminals
### To Reproduce

```
lightning init pl-app train.py
```

image

### Expected behavior

Icons get displayed or hidden

The fix would be similar as in [https://github.com/PyTorchLightning/pytorch-lightning/blob/dd475183227644a8d22dca3[…]fb0a9b2c4/pytorch_lightning/loops/dataloader/evaluation_loop.py](https://github.com/PyTorchLightning/pytorch-lightning/blob/dd475183227644a8d22dca3[%E2%80%A6]fb0a9b2c4/pytorch_lightning/loops/dataloader/evaluation_loop.py)

### Environment

This is a copy of an old [linear ticket](https://linear.app/lightning-ai/issue/LAI2-9250/the-icons-in-the-summary-dont-get-displayed-correctly) from user testing.

christopher-nguyen-re:
## What does this PR do?

This PR aims to fix broken CLI output for some terminals. I added a test in tests/tests_app/cli/test_cli.py but am unsure if it is an effective one. I will need to verify that this works on terminals addressed by @awaelchli.

Fixes #14226 

## Before submitting

## PR review

## Did you have fun?

Yes! This was a fun first experience.


src/lightning_app/cli/cmd_pl_init.py
```diff
@@ -148,8 +149,17 @@ def print_pretty_report(
         text_pathname.stylize(f"link file://{path}")
         text_pathname.append(f" {padding} {help_text}", "blue")
 
-        icon = "📂 " if path.is_dir() else "📄 "
-        tree.add(Text(icon) + text_pathname)
+        icon = ""
+        try:
+            if path.is_dir():
+                icon = "📂 ".encode()
```
awaelchli, line 155: Small suggestion: You could make this into a small helper function like this:

```py
def _encode_icon(icon: str) -> str:
    try:
        return "📂 ".encode()
    except UnicodeEncodeError:
         return b"  "
```
and then call it here in the print_pretty_report(). 

This would allow you to more easily test it, by calling the _encode_icon() function directly in your test.

awaelchli, line 155:
```suggestion
                icon = "📂 ".encode(sys.stdout.encoding)
```
This is needed, otherwise it won't fix the issue. 

The issue occurs when for example Python is encoding the string in utf8 but the terminal is set to display ascii.

src/lightning_app/cli/cmd_pl_init.py
```diff
@@ -148,8 +149,17 @@ def print_pretty_report(
...
+            else:
+                icon = "📄 ".encode()
```
awaelchli, line 157: Instead of encoding and decoding it again, let's just test whether it can be encoded or not. 
See the implementation here: https://github.com/Lightning-AI/lightning/blob/2622989b108368813fdd3850b002fd1ad69988c2/src/pytorch_lightning/loops/dataloader/evaluation_loop.py#L379-L385

You can create a small utility function:

```py
_can_encode_icon(icon: str) -> bool:
    try:
        icon.encode(sys.stdout.encoding)
        return True
    except UnicodeEncodeError:
        return False
```

This would then be easier to test as well.

christopher-nguyen-re: I have implemented the utility function and changed the encoding to include `sys.stdout.encoding`. Could you explain how to set up the terminal to display ascii for the purpose of testing? I am unable to replicate the bug.

awaelchli (approved):
This is great! 
I improved the test, take a look.

src/lightning_app/cli/cmd_pl_init.py
awaelchli: Great! Thanks. 
The issue I reported didn't occur to me directly, I just reported it from an internal user testing. I believe it was @carmocca and the terminal was the standard Ubuntu terminal.

carmocca: https://github.com/Lightning-AI/lightning/issues/12823 was the reported issue and #12828 has the fix proposed here with its test. In case you want inspiration

awaelchli: I think the test already does the same now. Testing against anything that is not utf works I guess

carmocca (approved):

src/lightning_app/cli/cmd_pl_init.py
christopher-nguyen-re: Is there anything else I should change to this? If not, what are the next steps I need to do?

awaelchli: I think the issue is fixed, great work. The next step is review: At the bottom of the page you see that code-owner review is required from one of the maintainers listed there. After the missing approval is given, the PR will be merged. 
Thanks for your patience!

tchaton (approved):
Looks awesome ! Great contribution 1

ethanwharris (approved):

PR #14735, opened 2022-09-16
Standalone Lite: Move Lite classes

awaelchli:
## What does this PR do?

Follow up to #14726.
**This PR should be REBASE MERGED**

Moves the LightningLite class and wrappers to the standalone package. 
There are two commits:

- Moving the files/classes to the lite package.
- Add backward compatibility: PL has a subclass implementation in `src/pytorch_lightning/lite` with:
    - Conversion logic that turns `gpus=x` and `tpu_cores=x` into `devices=x`
    - Conversion logic that turns `plugins=PLPrecisionPlugin` into `plugins=LitePrecisionPlugin`
    - Conversion logic that turns `strategy=PLStrategy` into `strategy=LiteStrategy`

Please review the two commits separately. 

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @borda @justusschock @awaelchli @rohitgr7 @carmocca @akihironitta

PR #14798, opened 2022-09-20
Integrate Lite Precision into PL

awaelchli:
## What does this PR do?

Follow up to #14735
The PL base class PrecisionPlugin is now based on the Lite Precision plugin. 

## Before submitting

## PR review

## Did you have fun?

cc @borda @justusschock @carmocca @awaelchli @akihironitta


src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -97,7 +87,7 @@ def optimizer_step(
         self._after_closure(model, optimizer, optimizer_idx)
         skipped_backward = closure_result is None
         # in manual optimization, the closure does not return a value
-        if not isinstance(model, pl.LightningModule) or not model.automatic_optimization or not skipped_backward:
```
awaelchli, line 100: Removed several isinstance checks in the plugins. These were here to separate code path for Lite and PL, but we no longer have them shared after #14735. The code path is now completely isolated.

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -43,7 +42,7 @@ def __init__(self, amp_level: str = "O2") -> None:
         self._connected = False
         self._state_dict_loaded = False
 
-    def main_params(self, optimizer: Optimizer) -> _PARAMETERS:
+    def get_main_params(self, optimizer: Optimizer) -> _PARAMETERS:
```
awaelchli, line 45: I should probably revert his api change in Lite so we don't have to deprecate/rename it

src/pytorch_lightning/plugins/precision/deepspeed.py
```diff
@@ -97,24 +86,19 @@ def backward(
                 "You have overridden the `LightningModule.backward` hook but it will be ignored since DeepSpeed handles"
                 " the backward logic internally."
             )
+        optimizer, optimizer_idx, *args = args
         deepspeed_engine: "deepspeed.DeepSpeedEngine" = model.trainer.model
-        deepspeed_engine.backward(closure_loss, *args, **kwargs)
-
-    def _run_backward(
```
awaelchli, line 103: Removed run_backward in all plugins. This is dead code now after #14735, and calls into the Lite specific backward implementation.

src/pytorch_lightning/plugins/precision/mixed.py
```diff
@@ -1,26 +0,0 @@
-# Copyright The PyTorch Lightning team.
```
awaelchli, line 1: TODO: should we deprecate or drop it immediately?

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -55,35 +54,26 @@ def dispatch(self, trainer: "pl.Trainer") -> None:
...
         """Run before precision plugin executes backward.
 
         Args:
+            tensor: the loss value obtained from the closure
             model: the model to be optimized
-            closure_loss: the loss value obtained from the closure
-            optimizer: current optimizer being used. ``None`` if using manual optimization
-            optimizer_idx: the index of the current optimizer. ``None`` if using manual optimization
         """
+        optimizer, optimizer_idx, *args = args
```
carmocca, line 64: Optimizer idx is unused

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -55,35 +54,26 @@ def dispatch(self, trainer: "pl.Trainer") -> None:
...
-    def backward(
-        self,
-        model: "pl.LightningModule",
-        closure_loss: Tensor,
-        optimizer: Optional[Optimizer],
-        optimizer_idx: Optional[int],
-        *args: Any,
-        **kwargs: Any,
-    ) -> None:
+    def backward(self, tensor: Tensor, model: "pl.LightningModule", *args: Any, **kwargs: Any) -> None:
```
carmocca, line 57: `closure_` does not apply to Lite necessarily, but why not keep it as `loss` instead of just `tensor`?

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -55,35 +54,26 @@ def dispatch(self, trainer: "pl.Trainer") -> None:
...
         self,
-        model: Optional[Union["pl.LightningModule", Module]],
         optimizer: Optimizer,
-        optimizer_idx: int,
-        closure: Callable[[], Any],
+        model: Optional["pl.LightningModule"] = None,
         **kwargs: Any,
     ) -> Any:
+        optimizer_idx = kwargs.pop("optimizer_idx")
+        closure = kwargs.pop("closure")
```
carmocca, line 76: I don't get the reason for doing this. How is this different to having them explicitly defined in the signature?

src/pytorch_lightning/plugins/precision/deepspeed.py
```diff
@@ -97,24 +86,19 @@ def backward(
                 "You have overridden the `LightningModule.backward` hook but it will be ignored since DeepSpeed handles"
                 " the backward logic internally."
             )
+        optimizer, optimizer_idx, *args = args
```
carmocca, line 89: These are unused

src/pytorch_lightning/plugins/precision/mixed.py
carmocca: There seems to be no direct public usage of it: https://grep.app/search?q=%28MixedPrecision%29

src/pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -21,175 +21,90 @@
...
     def connect(
         self, model: Module, optimizers: List[Optimizer], lr_schedulers: List[Any]
     ) -> Tuple[Module, List[Optimizer], List[Any]]:
         """Connects this plugin to the accelerator and the training process."""
         return model, optimizers, lr_schedulers
 
-    def pre_backward(self, model: "pl.LightningModule", closure_loss: Tensor) -> Tensor:
-        """Run before precision plugin executes backward.
+    @contextlib.contextmanager
+    def train_step_context(self) -> Generator[None, None, None]:
```
carmocca, line 45: Moving these has messed up the diff. Is this intended?

src/pytorch_lightning/plugins/precision/apex_amp.py
carmocca: Ah okay. This is just to avoid the `override` mypy complaint

src/pytorch_lightning/plugins/precision/deepspeed.py
awaelchli: I prefer this way to make it explicit what the two arguments we are unpacking do.

src/pytorch_lightning/plugins/precision/apex_amp.py
awaelchli: it is used below in the super call

src/pytorch_lightning/plugins/precision/precision_plugin.py
awaelchli: It is intentional, to make the order or methods consistent across plugins

src/lightning_lite/plugins/precision/precision.py
```diff
@@ -67,7 +67,7 @@ def optimizer_step(
         """Hook to run the optimizer step."""
         return optimizer.step(**kwargs)
 
-    def get_main_params(self, optimizer: Optimizer) -> _PARAMETERS:
+    def main_params(self, optimizer: Optimizer) -> _PARAMETERS:
```
awaelchli, line 70: Don't like the name, but reverting it to avoid breaking change (although this only exists for apex which nobody uses).

src/pytorch_lightning/plugins/precision/apex_amp.py
awaelchli: In Lite, you can call backward on anything that is a tensor. It doesn't have to be a scalar value like in PL that represents a loss (it could be a score that needs to be maximized). As long as it has a graph attached. In PL, one would have to override LM.backward() for this I suppose.

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -55,35 +54,28 @@ def dispatch(self, trainer: "pl.Trainer") -> None:
...
         """Run before precision plugin executes backward.
 
         Args:
+            tensor: the loss value obtained from the closure
             model: the model to be optimized
-            closure_loss: the loss value obtained from the closure
-            optimizer: current optimizer being used. ``None`` if using manual optimization
-            optimizer_idx: the index of the current optimizer. ``None`` if using manual optimization
         """
+        optimizer, optimizer_idx, *args = args
```
justusschock, line 70: can't the optimizer and index be passed as kwargs? Does the lite class take care of this so that they are always passed as args here?

To be safe we would need to do something like

```python
if "optimzer" in kwargs:
    optimizer = kwargs.pop("optimizer")
else:
   optimizer, *args = args

if "optimizer_idx" in kwargs:
    optimizer_idx = kwargs.pop("optimizer_idx")
else:
    optimizer_idx, *args = args
```

src/pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -30,7 +30,7 @@
     from torch.cuda.amp import autocast as old_autocast
 
 
-class NativeMixedPrecisionPlugin(MixedPrecisionPlugin):
+class NativeMixedPrecisionPlugin(PrecisionPlugin):
```
justusschock, line 33: Is there a plan to merge these with the lite plugins, build ontop of the lite plugins?

justusschock (approved):

src/pytorch_lightning/plugins/precision/apex_amp.py
awaelchli: The strategy calls the precision plugin like this:

        self.precision_plugin.backward(closure_loss, self.lightning_module, optimizer, optimizer_idx, *args, **kwargs)

This is why they are positional. There is a contract there that this is the way to  pass them in. I wanted to minimize breaking change.

awaelchli: > Does the lite class take care of this so that they are always passed as args here?

The lite class doesn't have a definition for optimizer or idx.

src/pytorch_lightning/plugins/precision/native_amp.py
awaelchli: Right now they can't because the base class in PL has PL-specific methods (e.g. connect). I don't see that changing very soon. The same also in strategies.

src/pytorch_lightning/plugins/precision/apex_amp.py
```diff
@@ -97,7 +89,7 @@ def optimizer_step(
         self._after_closure(model, optimizer, optimizer_idx)
         skipped_backward = closure_result is None
         # in manual optimization, the closure does not return a value
-        if not isinstance(model, pl.LightningModule) or not model.automatic_optimization or not skipped_backward:
+        if not model.automatic_optimization or not skipped_backward:
```
tchaton, line 92: Neat !

src/pytorch_lightning/plugins/precision/deepspeed.py
```diff
@@ -97,24 +93,19 @@ def backward(
                 "You have overridden the `LightningModule.backward` hook but it will be ignored since DeepSpeed handles"
                 " the backward logic internally."
             )
+        _, _, *args = args
```
tchaton, line 96: Not sure fan of those args unpacking. This is easy to forget and break stuff.

src/pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -57,27 +57,20 @@ def __init__(
         self.device = device
         self.scaler = scaler
 
-    def pre_backward(self, model: "pl.LightningModule", closure_loss: Tensor) -> Tensor:
-        if self.scaler is not None:
-            closure_loss = self.scaler.scale(closure_loss)
-        return super().pre_backward(model, closure_loss)
-
-    def _run_backward(self, tensor: Tensor, model: Optional[Module], *args: Any, **kwargs: Any) -> None:
+    def pre_backward(self, tensor: Tensor, module: "pl.LightningModule") -> Tensor:  # type: ignore[override]
```
tchaton, line 60: There are some inconsistency with module and model associated with the `LightningModule`. Let's use model everywhere no ?

src/pytorch_lightning/plugins/precision/native_amp.py
```diff
@@ -57,27 +57,20 @@ def __init__(
...
-    def optimizer_step(
-        self,
-        model: Optional[Union["pl.LightningModule", Module]],
-        optimizer: Optimizer,
-        optimizer_idx: int,
-        closure: Callable[[], Any],
-        **kwargs: Any,
-    ) -> Any:
+    def optimizer_step(self, optimizer: Steppable, **kwargs: Any) -> Any:
+        optimizer_idx = kwargs.pop("optimizer_idx")
```
tchaton, line 66: Why don't you expose them in the method signature ?

src/pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -21,97 +21,59 @@
 from torch.optim import Optimizer
 
 import pytorch_lightning as pl
-from lightning_lite.utilities.types import _PARAMETERS
+from lightning_lite.plugins import Precision as _Precision
```
tchaton, line 24:
```suggestion
from lightning_lite.plugins import Precision as LitePrecision
```

src/pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -21,97 +21,59 @@
...
-    def post_backward(self, model: "pl.LightningModule", closure_loss: Tensor) -> Tensor:
-        """Run after precision plugin executes backward.
+        optimizer, optimizer_idx, *args = args
+        model.backward(tensor, optimizer, optimizer_idx, *args, **kwargs)
 
-        Args:
-            model: the model to be optimized
-            closure_loss: the loss value obtained from the closure
-        """
+    def post_backward(self, tensor: Tensor, module: "pl.LightningModule") -> Tensor:  # type: ignore[override]
```
tchaton, line 68: same here. Might be better to use model.

tchaton (approved):
LGTM ! Minor comments.

src/pytorch_lightning/plugins/precision/deepspeed.py
awaelchli: what should I do instead?

src/pytorch_lightning/plugins/precision/native_amp.py
awaelchli: LightningModule is a module, and in Lite we have nn.Module. This is the name that is consistent.

src/pytorch_lightning/plugins/precision/native_amp.py
awaelchli: In Lite, there is no optimizer_idx

src/pytorch_lightning/plugins/precision/deepspeed.py
justusschock: I prefer not to leak them to Lite as they aren't required there.

carmocca (approved):

PR #14810, opened 2022-09-20
Remove the model argument from Lite's `optimizer_step` via structural typing

carmocca:
## What does this PR do?

See title

None

cc @borda @justusschock @awaelchli @rohitgr7 @carmocca

### DIFF
--- a/src/lightning_lite/utilities/types.py
+++ b/src/lightning_lite/utilities/types.py
@@ -83,5 +83,5 @@ class ReduceLROnPlateau(_Stateful[str], Protocol):
 class Steppable(Protocol):
     """To structurally type ``optimizer.step()``"""
 
-    def step(self, closure: Optional[Callable[[], float]] = ...) -> Optional[float]:
+    def step(self, closure: Optional[Callable[[], float]] = ..., **kwargs: Any) -> Optional[float]:
         ...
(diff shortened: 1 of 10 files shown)


src/lightning_lite/utilities/types.py
```diff
@@ -77,3 +77,11 @@ def __init__(
 
     def step(self, metrics: Union[float, int, Tensor], epoch: Optional[int] = None) -> None:
         ...
+
+
+@runtime_checkable
+class Steppable(Protocol):
+    """To structurally type ``optimizer.step()``"""
+
+    def step(self, closure: Optional[Callable[[], float]] = ...) -> Optional[float]:
```
awaelchli, line 86: what about the kwargs? we need to be able to pass them down

awaelchli:
@carmocca I took care of updating the tests

awaelchli (approved):

carmocca:
> @carmocca I took care of updating the tests

Thanks! I moved the Steppable assertion into the test so we don't need to mock the engine during the "check step called" tests

justusschock (approved):

src/lightning_lite/utilities/types.py
justusschock: could we do this with overloads so that still one of the "options" would match the optimizer?

awaelchli: @carmocca ok, so we would need to put the type ignore in the usages where we pass kwargs i guess

awaelchli: Feel free to resolve my part of this conversation or maybe try what Justus said.

PR #14831, opened 2022-09-21
Add support for custom cloud compute configurations for Flows

nicolai86:
## What does this PR do?

This PR exposes the custom cloud compute configuration we support on the lightning AI platform as part of the lightning framework.
It depends on an update lightning-cloud package, which is why this PR is pending right now.

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/core/app.py
```diff
@@ -45,6 +46,7 @@ class LightningApp:
     def __init__(
         self,
         root: "lightning_app.LightningFlow",
+        flow_compute_config: Optional["lightning_app.CloudCompute"] = None,
```
awaelchli, line 49: This is great. Two minor comments:

- We call the class "CloudCompute", so I would avoid CloudConfig or config in the name of arguments, or in the docstring explanation. 
- The argument in work is called simply "cloud_compute". For the app here, I propose "flow_cloud_compute" or "cloud_compute" as the input argument names.

nicolai86: agreed. renamed to `flow_cloud_compute`.

src/lightning_app/core/app.py
```diff
@@ -45,6 +46,7 @@ class LightningApp:
     def __init__(
         self,
         root: "lightning_app.LightningFlow",
+        flow_cloud_compute: Optional["lightning_app.CloudCompute"] = None,
```
tchaton, line 49: Let's call it `default_cloud_compute`. I think it is more accurate.

src/lightning_app/runners/cloud.py
```diff
@@ -177,12 +178,18 @@ def dispatch(
             frontend_spec = V1Flowserver(name=flow_name)
             frontend_specs.append(frontend_spec)
 
+        flow_compute_config_name = "flow-lite"
```
tchaton, line 181: Let's just call it default.

src/lightning_app/core/app.py
awaelchli: This was also my suggestion above, but the logic that sets this as default also on the Work is missing. It would have to be done either here or in a different PR. But we are close to a release, so this should be considered carefully I think.

src/lightning_app/core/app.py
```diff
@@ -79,6 +82,10 @@ def __init__(
 
         _validate_root_flow(root)
         self._root = root
+        self.default_cloud_compute = default_cloud_compute or lightning_app.CloudCompute()
+
+        if self.default_cloud_compute.preemptible:
+            raise Exception("The default Cloud Compute can't be preemptible.")
```
awaelchli, line 88: Note: it is being removed here:
https://github.com/Lightning-AI/lightning/pull/14852/

src/lightning_app/core/app.py
```diff
@@ -45,6 +46,7 @@ class LightningApp:
     def __init__(
         self,
         root: "lightning_app.LightningFlow",
+        default_cloud_compute: Optional["lightning_app.CloudCompute"] = None,
```
hhsecond, line 49: should this be named `default_**`. What does `default` indicate which is not clear from just using `cloud_compute`?

src/lightning_app/core/app.py
```diff
@@ -60,6 +62,7 @@ def __init__(
         Arguments:
             root: The root LightningFlow component, that defines all the app's nested components, running infinitely.
                 It must define a `run()` method that the app can call.
+            default_cloud_compute: The default Compute Compute used for flow, Rest API and frontend's.
```
hhsecond, line 65:
```suggestion
            default_cloud_compute: The default Cloud Compute used for flow, Rest API and frontend's.
```

src/lightning_app/core/app.py
hhsecond: @tchaton @awaelchli

awaelchli: @tchaton chose this name. As I mentioned, there would be logic needed that applies the value given here to all Works that have no user-provided compute configuration. If this logic is not intended to be added, then we should not call it `default_...`, and `flow_cloud_compute` instead as proposed above.

nicolai86: @tchaton @awaelchli any objections if I revert the change to `flow_cloud_compute` and we address the default behaviour separately?

awaelchli: No objection. @tchaton ?

lantiga: I like `flow_cloud_compute`, it’s more specific compared to `default`

nicolai86: :+1: renamed back to `flow_cloud_compute`.

src/lightning_app/core/app.py
```diff
@@ -64,6 +66,7 @@ def __init__(
         Arguments:
             root: The root LightningFlow component, that defines all the app's nested components, running infinitely.
                 It must define a `run()` method that the app can call.
+            flow_cloud_compute: The default Cloud Compute used for flow, Rest API and frontend's.
```
rohitgr7, line 69: there can be flows within root flows, so this on is only for root flow?
also wondering why is it not part of Flow construct?

src/lightning_app/core/app.py
```diff
@@ -91,6 +94,10 @@ def __init__(
         self.root_path = root_path  # when running behind a proxy
         _validate_root_flow(root)
         self._root = root
+        self.flow_cloud_compute = flow_cloud_compute or lightning_app.CloudCompute()
+
+        if self.flow_cloud_compute.preemptible:
```
rohitgr7, line 99: also fyi: https://github.com/Lightning-AI/lightning/pull/14852/files

src/lightning_app/core/app.py
awaelchli: All flows always runs in the same process, on the same machine. Configuring different cloud compute for different flow components is not valid. Therefore, it gets defined on the app.

awaelchli:
> It depends on an update lightning-cloud package, which is why this PR is pending right now.

Is this resolved? Don't you need to bump the version in this PR?

awaelchli (approved):
> It depends on an update lightning-cloud package, which is why this PR is pending right now.

Is this resolved? Don't you need to bump the version in this PR?

Furthermore, will you update the user guide in our docs?

nicolai86:
> > It depends on an update lightning-cloud package, which is why this PR is pending right now.
> 
> Is this resolved? Don't you need to bump the version in this PR?

This is resolved - there were other PRs that bumped the version past what I needed for this PR, hence the change is no longer present.

> Furthermore, will you update the user guide in our docs?

I thought we auto-generate the docs from the code. Can you point me to what I need to add?
\cc @awaelchli

awaelchli:
> I thought we auto-generate the docs from the code. Can you point me to what I need to add?

Yes, you added the necessary API docs to the classes/functions etc. which is great! In addition, with every feature we usually have docs with user guide, i.e., this document would explain on a high level what it is used for, when to use it, with examples and notes. 

For cloud compute specifically, this is documented in `docs/source-app/core_api/lightning_work/compute_content.rst`. There is not much content there, but with additions like the one in this PR I believe it is important that we keep this updated, especially now that the story slightly changes: No longer is it just the work that takes a CloudCompute.

nicolai86:
thanks @awaelchli , I've added basic documentation for Flows & CloudCompute.
It's minimal because today, on lightning managed cloud you can't configure too much - we need to extend this for BYOC functionality which allows much more freedom eventually.

src/lightning_app/runners/cloud.py
```diff
@@ -197,6 +198,11 @@ def dispatch(
             flow_servers=frontend_specs,
             desired_state=V1LightningappInstanceState.RUNNING,
             env=v1_env_vars,
+            user_requested_flow_compute_config=V1UserRequestedFlowComputeConfig(
```
hhsecond, line 201: There are other arguments in the CloudCompute class which we are not using here. Perhaps raise error or a warning if those are set. Like the user is specifying a `cluster` in the compute config but is pointed to run on a different cluster etc

hhsecond (approved):

src/lightning_app/runners/cloud.py
nicolai86: this was in the code at some point, but as raised [here](https://github.com/Lightning-AI/lightning/pull/14831#discussion_r993556833) it was removed due to the PR that deletes the properties.

awaelchli:
@nicolai86 I fixed the docs build for you (hopefully). 
@edenlightning @edenafek could you also take a look at the docs here?

src/lightning_app/core/app.py
nicolai86: this change was reverted because it lead to more discussions

tchaton: I think this requires a final API decision. Flow doesn't seem right as it runs the flow, rest api, frontends and in the future the works.
I agree default is mis-leading and I don't think it is a good candidate.

I think `main_cloud_compute` or `root_cloud_compute` would fit better.

tchaton (approved):
LGTM !

PR #14882, opened 2022-09-24
Show Chinese characters correctly in a simple way

### LINKED ISSUE #14837: When test, the printed log does not show the full text of the metrics, when named in Chinese
### First check

- [X] I'm sure this is a bug.
- [X] I've added a descriptive title to this bug.
- [X] I've provided clear instructions on how to reproduce the bug.
- [X] I've added a code sample.
- [X] I've provided any other important info that is required.

### Bug description

Hi,

I am dealing with a very messy segmentation problem, which requires me to log the per-category accuracy and iou at test time. I used self.log(category, metrics) to do this job at "on_test_epoch_end". However, the log does not show the full metrics text, and shorten rest of it to "..." instead, when the metric name is in Chinese. I know that in the file pytorch_lightning.loops.dataloader.evaluation_loop, by line 340, it setting the max_length based on the longest string. However, my longest Chinese metric is 22 words long, and I checked that "max_length" is 25, but the printed metrics still hides part of the name.

Screenshot from 2022-09-22 14-43-52

 

### How to reproduce the bug

```python
def on_test_epoch_end(self):
    self.log("车道线-虚减速让行线_清晰_树荫遮挡_acc", 0.0121)
```

### Error messages and logs

```

# Error messages and logs here please

```

### Important info

```

#- Lightning Component (e.g. Trainer, LightningModule, LightningApp, LightningWork, LightningFlow): LightningModule
#- PyTorch Lightning Version (e.g., 1.5.0): 1.7.6
#- PyTorch Version (e.g., 1.10): 1.12.1
#- Python version (e.g., 3.9): 3.9
#- OS (e.g., Linux): Ubuntu 20.04 in English
#- CUDA/cuDNN version: 11.6
#- GPU models and configuration: 3060Ti
#- How you installed Lightning(`conda`, `pip`, source): pip
#- Running environment of LightningApp (e.g. local, cloud): local

```

### More info

_No response_

heng-yuwen:
## What does this PR do?

Fixes #14837 to show the full text of Chinese characters in metrics.

## Before submitting

## PR review

## Did you have fun?


awaelchli (approved):

src/pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -334,7 +334,9 @@ def _print_results(results: List[_OUT_DICT], stage: str) -> None:
 
         # fallback is useful for testing of printed output
         term_size = shutil.get_terminal_size(fallback=(120, 30)).columns or 120
-        max_length = int(min(max(len(max(metrics_strs, key=len)), len(max(headers, key=len)), 25), term_size / 2))
+        # encode as gbk, which treat Chinese characters as twice the width of alphabet characters.
+        matrics_encoded = [metric.encode("gbk") for metric in metrics_strs]
```
justusschock, line 338: Can we please use an encoding that is not specialized on chinese characters? GBK specializes on chinese characters and performs worse for others. I'd prefer something like UTF-8, UTF-16 or even UTF-32 (depending on what we want to support) instead

justusschock, line 338:
```suggestion
        metrics_encoded = [metric.encode("gbk") for metric in metrics_strs]
```

src/pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -334,7 +334,9 @@ def _print_results(results: List[_OUT_DICT], stage: str) -> None:
...
+        max_length = int(min(max(len(max(matrics_encoded, key=len)), len(max(headers, key=len)), 25), term_size / 2))
```
justusschock, line 339:
```suggestion
        max_length = int(min(max(len(max(metrics_encoded, key=len)), len(max(headers, key=len)), 25), term_size / 2))
```

justusschock (changes requested):

carmocca (commented):
Can you add a test? Maybe adding to https://github.com/Lightning-AI/lightning/blob/d2e4e7e003dd2c71bbbd9c5afafc00d0ee19898a/tests/tests_pytorch/trainer/logging_/test_eval_loop_logging.py#L894

Also update the CHANGELOG with a "Fixed" entry

src/pytorch_lightning/loops/dataloader/evaluation_loop.py
123mutourener: Yes you are right, gbk makes sure Chinese and English characters work as expected. However, I didn’t try other languages. I will have a look and see if other encodings can make it work.

heng-yuwen: Hi @justusschock I searched for methods to handle other languages, but it seems that it is not a trivial task. 

I found this repo useful: https://github.com/fedora-infra/kitchen/blob/73e09ff3ac6e453f1e6f307cb2a48474b7435f82/kitchen3/kitchen/text/display.py#L335

It seems that it can properly calculate the character width for Chinese, Japanese, Korean and Spanish at least. I haven't tested other languages, but I am not sure if it is necessary to support more languages.

justusschock: @heng-yuwen thanks for looking into this. I do not quite understand the difficulty here.
From looking at the utf-8 encoding table [here](https://www.utf8-chartable.de/unicode-utf8-table.pl) it seems that utf-8 does support encoding for chinese, english and most other languages (including cyrillic alphabets and others as well).
Why can't we just use this?

heng-yuwen: @justusschock The problem is not about encoding, but about how to get the printed character width. If you try to encode the Chinese character as utf-8, the len() function gives the character number instead of width.

heng-yuwen: Also, if you are interested, gbk encodes Chinese characters with 2 bytes, which is also the length for most Chinese characters. That's why the length of gbk encoding can make Chinese characters look fine when printed. For other languages, it needs extra dedicated way to deal with the length.

carmocca: If you write a test that shows the bug/issue, then we will understand the problem better so that we can find a reliable solution across languages

justusschock:
Closing for now.  Let's continue the discussion in the related issue.

PR #14904, opened 2022-09-27
Make internal torchscript check a class attribute

carmocca:
## What does this PR do?

3rd time is the charm.
Reported in Slack.

None

cc @borda @carmocca @justusschock @awaelchli @ananthsub @ninginthecloud @jjenniferdai @rohitgr7 @akihironitta

### DIFF
--- a/src/pytorch_lightning/core/module.py
+++ b/src/pytorch_lightning/core/module.py
@@ -1878,34 +1878,36 @@ class LightningModule(
         mode = self.training
 
         LightningModule._running_torchscript = True
+        try:
+            if method == "script":
+                torchscript_module = torch.jit.script(self.eval(), **kwargs)
+            elif method == "trace":
+                # if no example inputs are provided, try to see if model has example_input_array set
+                if example_inputs is None:
+                    if self.example_input_array is None:
+                        raise ValueError(
+                            "Choosing method=`trace` requires either `example_inputs`"
+                            " or `model.example_input_array` to be defined."
+                        )
+                    example_inputs = self.example_input_array
 
-        if method == "script":
-            torchscript_module = torch.jit.script(self.eval(), **kwargs)
-        elif method == "trace":
-            # if no example inputs are provided, try to see if model has example_input_array set
-            if example_inputs is None:
-                if self.example_input_array is None:
-                    raise ValueError(
-                        "Choosing method=`trace` requires either `example_inputs`"
-                        " or `model.example_input_array` to be defined."
-                    )
-                example_inputs = self.example_input_array
+                # automatically send example inputs to the right device and use trace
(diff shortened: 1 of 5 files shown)


src/pytorch_lightning/core/module.py
```diff
@@ -1889,7 +1877,7 @@ def to_torchscript(
         """
         mode = self.training
 
-        self._running_torchscript = True
+        LightningModule._running_torchscript = True
```
justusschock, line 1880: you should still be able to use `self` here as long as this was declared a classattribute before, no?

If you derive a class from it and only set it for `LightningModule`, is it also set for the derived class?

carmocca: > you should still be able to use self here as long as this was declared a classattribute before, no?

I thought so too, but it does not work. Not sure why

> If you derive a class from it and only set it for LightningModule, is it also set for the derived class?


I believe the modified test covers this

awaelchli (approved):

src/pytorch_lightning/core/module.py
awaelchli: Maybe you missed an instance attribute? Because if setattr is invoked, the value of Class.attr and instance.attr would diverge.

justusschock (approved):

src/pytorch_lightning/core/module.py
carmocca: I don't think I did. It might be caused by the torchscript implementation

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -242,9 +242,11 @@ def log_graph(self, model: "pl.LightningModule", input_array: Optional[Tensor] =
             if input_array is not None:
                 input_array = model._on_before_batch_transfer(input_array)
                 input_array = model._apply_batch_transfer_handler(input_array)
-                model._running_torchscript = True
-                self.experiment.add_graph(model, input_array)
-                model._running_torchscript = False
+                pl.LightningModule._running_torchscript = True
+                try:
+                    self.experiment.add_graph(model, input_array)
```
awaelchli, line 247: @carmocca should we go with a context manager? exit gets called when an exception occurs inside, I think it would look slightly cleaner.

carmocca: I would agree if this was something user facing, but defining a context manager class and instantiating it etc just to set this attribute seems overkill to me.

No strong opinion though, I can do it if you prefer

carmocca: Done

awaelchli (approved):

tchaton (approved):
LGTM !

PR #14907, opened 2022-09-27
Find last checkpoints on restart

### LINKED ISSUE #14466: Last saved checkpoint not resumed when using trainer.fit(..., ckpt_path='last')
## 🐛 Bug

Last saved checkpoint (by setting save_last=True in ModelCheckpoint callback) is not automatically resumed when specifying ckpt_path='last' argument in trainer.fit method. 

From the related code (https://github.com/Lightning-AI/lightning/blob/1.7.3/src/pytorch_lightning/trainer/trainer.py#L1419), I see that the last model path is taken from the "last_model_path" attribute of the ModelCheckpoint callback object, but its value is always an empty string at the moment of resuming the latest checkpoint.

One workaround is to manually set this attribute just after creating the callback, but I don't know if this was the intended behaviour or if there is another (and possibly better) way to do all this.

### To Reproduce

Run this code a first time and stop it after some epochs have been completed. Then run again. The second run does not resume the latest saved checkpoint but starts the training again from the first epoch.

```python
import os

import torch
from time import sleep


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

    def on_validation_epoch_end(self) -> None:
        sleep(1)

def run():

    model = BoringModel()
    trainer = Trainer(
        callbacks=ModelCheckpoint(monitor='valid_loss', mode='min', filename='bestLoss', save_last=True),
        accelerator='cpu',
        logger=None,
        default_root_dir=os.getcwd(),
        max_epochs=1000)

    # fit
    trainer.fit(model, train_dataloaders=train_data, val_dataloaders=val_data, ckpt_path='last')


if __name__ == "__main__":
    run()

```

### Expected behavior

Resume training from the last saved checkpoint when using ckpt_path="last" option in trainer.fit

### Environment
<details>
  <summary>Details</summary>

        * CUDA:
                - GPU:
                        - Quadro T1000
                - available:         True
                - version:           10.2
        * Lightning:
                - pytorch-lightning: 1.7.3
                - torch:             1.12.1
                - torchmetrics:      0.9.3
        * Packages:
                - absl-py:           1.2.0
                - aiohttp:           3.8.1
                - aiosignal:         1.2.0
                - async-timeout:     4.0.2
                - attrs:             22.1.0
                - cachetools:        5.2.0
                - certifi:           2022.6.15
                - charset-normalizer: 2.1.1
                - frozenlist:        1.3.1
                - fsspec:            2022.7.1
                - google-auth:       2.11.0
                - google-auth-oauthlib: 0.4.6
                - grpcio:            1.48.0
                - idna:              3.3
                - importlib-metadata: 4.12.0
                - markdown:          3.4.1
                - markupsafe:        2.1.1
                - multidict:         6.0.2
                - numpy:             1.23.2
                - oauthlib:          3.2.0
                - packaging:         21.3
                - pip:               22.2.2
                - protobuf:          3.19.4
                - pyasn1:            0.4.8
                - pyasn1-modules:    0.2.8
                - pydeprecate:       0.3.2
                - pyparsing:         3.0.9
                - pytorch-lightning: 1.7.3
                - pyyaml:            6.0
                - requests:          2.28.1
                - requests-oauthlib: 1.3.1
                - rsa:               4.9
                - setuptools:        63.3.0
                - six:               1.16.0
                - tensorboard:       2.10.0
                - tensorboard-data-server: 0.6.1
                - tensorboard-plugin-wit: 1.8.1
                - torch:             1.12.1
                - torchmetrics:      0.9.3
                - tqdm:              4.64.0
                - typing-extensions: 4.3.0
                - urllib3:           1.26.12
                - werkzeug:          2.2.2
                - wheel:             0.37.1
                - yarl:              1.8.1
                - zipp:              3.8.1
        * System:
                - OS:                Linux
                - architecture:
                        - 64bit
                        - ELF
                - processor:         x86_64
                - python:            3.8.10
                - version:           #49~20.04.1-Ubuntu SMP Thu Aug 4 19:15:44 UTC 2022
</details>

### Additional context

cc @justusschock @awaelchli @ninginthecloud @rohitgr7 @otaj @akihironitta

otaj:
## What does this PR do?

Fixes #14466

Although #12816 enabled support for passing `"last"` as an argument for reloading checkpoint, reloading happened only for FT checkpoints and checkpoints that were created within the same run.

This PR adds support for reloading last checkpoint on restart of the run.

Note, that this makes `"best"` and `"last"` keywords divergent in their behavior - `"best"` doesn't reload the best available checkpoint when starting a new run for training.

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7 @otaj @carmocca @jjenniferdai @akihironitta


src/pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -583,6 +583,8 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
         """
         if self.dirpath is not None:
             # short circuit if dirpath was passed to ModelCheckpoint
+            if broadcast:
+                self.dirpath = trainer.strategy.broadcast(self.dirpath)
```
awaelchli, line 587: This is too dangerous for my taste. It can lead to unwanted hangs when dirpath is None on one rank and not None on another. Even if it's not the case today, it's annoying to debug if you run into it. Our best practice in Lightning is to never run collectives under a conditional unless you can guarantee that every rank enters it. 

I suggest to keep the broadcasting outside of `__resolve_ckpt_dir`, i.e., wherever it is needed, call

```
dir = __resolve_ckpt_dir()
dir = broadcast(dir)
```
and otherwise not. 

What do you think?

otaj: That absolutely makes sense. Will make the required changes

otaj: Done. Thank you!

src/pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -255,7 +255,9 @@ def state_key(self) -> str:
         )
 
     def setup(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", stage: str) -> None:
-        self.__resolve_ckpt_dir(trainer)
+        dirpath = self.__resolve_ckpt_dir(trainer)
+        dirpath = trainer.strategy.broadcast(dirpath)
+        self.dirpath = dirpath
         assert self.dirpath is not None
```
carmocca, line 261: Not necessary now

```suggestion
```

src/pytorch_lightning/callbacks/model_checkpoint.py
```diff
@@ -600,9 +602,14 @@ def __resolve_ckpt_dir(self, trainer: "pl.Trainer") -> None:
...
 
-        ckpt_path = trainer.strategy.broadcast(ckpt_path)
+        return ckpt_path
 
-        self.dirpath = ckpt_path
+    def _find_last_checkpoints(self, trainer: "pl.Trainer") -> List[str]:
+        # find all checkpoints in the folder
+        ckpt_path = self.__resolve_ckpt_dir(trainer)
+        if self._fs.exists(ckpt_path):
+            return [os.path.normpath(p) for p in self._fs.ls(ckpt_path, detail=False) if self.CHECKPOINT_NAME_LAST in p]
```
carmocca, line 611: Why do you normpath?

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -161,9 +161,10 @@ def _set_ckpt_path(
             ckpt_path = getattr(self.trainer.checkpoint_callback, "best_model_path", None)
 
         if ckpt_path == "last":
-            candidates = [getattr(ft, "ckpt_path", None) for ft in ft_checkpoints] + [
-                getattr(cb, "last_model_path", None) for cb in self.trainer.checkpoint_callbacks
-            ]
+            candidates = [getattr(ft, "ckpt_path", None) for ft in ft_checkpoints]
```
carmocca, line 164: Should all of these be sets?

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -161,9 +161,10 @@ def _set_ckpt_path(
...
+            for callback in self.trainer.checkpoint_callbacks:
+                if hasattr(callback, "_find_last_checkpoints"):
```
carmocca, line 166: For customization, we'd want to have this as something the users can subclass and override with some guarantees. I guess this is why you chose to use `hasattr`.

But I would check `isinstance(ModelCheckpoint)` instead in this PR

src/pytorch_lightning/callbacks/model_checkpoint.py
otaj: Because otherwise tests fail on Windows, because `self._fs.ls` uses `posixpath.separator` on local filesystem regardless of platform

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
otaj: Good idea, done

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
otaj: Yeah, that makes sense since we probably cannot expect people to implement this in the same way as us if they subclass a checkpoint. Done

carmocca (commented):
LGTM. Waiting on a test though!

awaelchli:
I found some tests related to this in `test_trainer.py`, `test_checkpoint_path_input_last` etc.. Perhaps one could modify one of these or follow a similar pattern as these tests.

otaj:
@awaelchli, @carmocca there is a test now. Btw, writing it helped me uncover a small bug with the code, see d3a3b35

tests/tests_pytorch/trainer/test_trainer.py
```diff
@@ -776,6 +776,40 @@ def test_checkpoint_path_input_last(tmpdir, ckpt_path, save_last, fn):
...
 
 
+def test_checkpoint_find_last(tmpdir):
+    """Test that the last checkpoint is found correctly."""
+    model = BoringModel()
+    mc = ModelCheckpoint(save_last=True)
+    trainer = Trainer(
+        max_epochs=1,
+        limit_train_batches=1,
+        limit_val_batches=1,
```
carmocca, line 786: Probably does not need validation

```suggestion
        limit_val_batches=0,
```

tests/tests_pytorch/trainer/test_trainer.py
```diff
@@ -776,6 +776,40 @@ def test_checkpoint_path_input_last(tmpdir, ckpt_path, save_last, fn):
...
+    )
+    assert trainer.ckpt_path is None
+    trainer.fit(model)
+
+    model = BoringModel()
+    mc = ModelCheckpoint()
+    trainer = Trainer(
+        max_epochs=1,
+        limit_train_batches=1,
+        limit_val_batches=1,
```
carmocca, line 801: Same comment

tests/tests_pytorch/trainer/test_trainer.py
```diff
@@ -776,6 +776,40 @@ def test_checkpoint_path_input_last(tmpdir, ckpt_path, save_last, fn):
...
+        enable_model_summary=False,
+        enable_progress_bar=False,
+        logger=False,
+        default_root_dir=tmpdir,
+        callbacks=[mc],
+    )
+    assert trainer.ckpt_path is None
+    trainer.fit(model)
+
+    model = BoringModel()
```
carmocca, line 796: You probably can get away with reusing the previous instance

```suggestion
```

tests/tests_pytorch/trainer/test_trainer.py
```diff
@@ -776,6 +776,40 @@ def test_checkpoint_path_input_last(tmpdir, ckpt_path, save_last, fn):
...
+    mc = ModelCheckpoint()
+    trainer = Trainer(
+        max_epochs=1,
+        limit_train_batches=1,
+        limit_val_batches=1,
+        enable_model_summary=False,
+        enable_progress_bar=False,
+        logger=False,
+        default_root_dir=tmpdir,
+        callbacks=[mc],
```
carmocca, line 806: This does not require pasing the checkpoint callback again, no?

tests/tests_pytorch/trainer/test_trainer.py
otaj: I want to be sure we're simulating restarting the run as closely as possible.

tests/tests_pytorch/trainer/test_trainer.py
otaj: It does - we are finding last checkpoints as a method on `ModelCheckpoint` instance

carmocca (approved):

awaelchli (approved):
Great!

PR #14910, opened 2022-09-27
Prepare CI to run on 3090s

carmocca:
## What does this PR do?

Part of https://github.com/Lightning-AI/lightning/pull/14632

None

### DIFF
--- a/tests/tests_pytorch/strategies/test_bagua_strategy.py
+++ b/tests/tests_pytorch/strategies/test_bagua_strategy.py
@@ -45,6 +45,10 @@ def test_bagua_default(tmpdir):
     assert isinstance(trainer.strategy, BaguaStrategy)
 
 
+@pytest.mark.skipif(
+    torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 8,
+    reason="Async does not support this CUDA architecture",
+)
 @RunIf(min_cuda_gpus=2, standalone=True, bagua=True)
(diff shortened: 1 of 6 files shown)


tests/tests_pytorch/strategies/test_bagua_strategy.py
```diff
@@ -45,6 +45,10 @@ def test_bagua_default(tmpdir):
     assert isinstance(trainer.strategy, BaguaStrategy)
 
 
+@pytest.mark.skipif(
+    torch.cuda.is_available() and torch.cuda.get_device_capability()[0] >= 8,
+    reason="Async does not support this CUDA architecture",
+)
```
awaelchli, line 51: 👍

awaelchli (approved):

akihironitta (approved):

tchaton (approved):
LGTM !

tests/tests_pytorch/strategies/test_bagua_strategy.py
otaj: I know it's already merged, but can we maybe have generic `RunIf` for these kind of things? I imagine this can pop up more frequently. If we're confident, such a test is going to be only one, then disregard it, but if there's going to be more of those, then `RunIf` would be much nicer.

carmocca: I believe it's the only instance. We can add it later if more pop up

PR #14915, opened 2022-09-27
feat: option to add custom meta tags to the UI container

pritamsoni-hsr:
## What does this PR do?

feature to allow users to add custom meta tags directly to the lightning app.

under the hood, it writes these tags to `index.html` as the last step in the LightningApp constructor.

# Limitation

if a user runs multiple apps on a single machine at the same time the latest tags will be visible in all the apps.

image should be a valid url, rn users cannot provide a local filesystem path.

need to update docs 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/requirements/app/base.txt
+++ b/requirements/app/base.txt
@@ -8,3 +8,4 @@ croniter>=1.3.0, <1.4.0  # strict; TODO: for now until we find something more ro
 traitlets>=5.3.0, <=5.4.0
 arrow>=1.2.0, <1.2.4
 lightning-utilities==0.3.*
+beautifulsoup4<=4.8.2
--- a/src/lightning_app/core/app.py
+++ b/src/lightning_app/core/app.py
@@ -18,12 +18,14 @@ from lightning_app.core.constants import (
     DEBUG_ENABLED,
     FLOW_DURATION_SAMPLES,
     FLOW_DURATION_THRESHOLD,
+    FRONTEND_DIR,
     STATE_ACCUMULATE_WAIT,
 )
 from lightning_app.core.queues import BaseQueue, SingleProcessQueue
 from lightning_app.frontend import Frontend
 from lightning_app.storage import Drive, Path
 from lightning_app.storage.path import storage_root_dir
+from lightning_app.utilities import frontend
 from lightning_app.utilities.app_helpers import _delta_to_app_state_delta, _LightningAppRef, Logger
 from lightning_app.utilities.commands.base import _process_requests
 from lightning_app.utilities.component import _convert_paths_after_init, _validate_root_flow
@@ -46,6 +48,7 @@ class LightningApp:
         self,
         root: "lightning_app.LightningFlow",
         debug: bool = False,
+        info: frontend.AppInfo = None,
     ):
         """The Lightning App, or App in short runs a tree of one or more components that interact to create end-to-end
         applications. There are two kinds of components: :class:`~lightning_app.core.flow.LightningFlow` and
@@ -62,6 +65,8 @@ class LightningApp:
                 It must define a `run()` method that the app can call.
             debug: Whether to activate the Lightning Logger debug mode.
                 This can be helpful when reporting bugs on Lightning repo.
+            info: Provide additional info about the app which will be used to update html title,
+                description and image meta tags and specify any additional tags as list of html strings.
 
         .. doctest::
 
@@ -132,6 +137,10 @@ class LightningApp:
             _console.setLevel(logging.DEBUG)
 
         logger.debug(f"ENV: {os.environ}")
+
+        # update index.html,
+        # this should happen once for all apps before the ui server starts running.
+        frontend.update_index_file_with_info(FRONTEND_DIR, info=info)
 
     def get_component_by_name(self, component_name: str):
         """Returns the instance corresponding to the given component name."""
--- a/src/lightning_app/utilities/frontend.py
+++ b/src/lightning_app/utilities/frontend.py
@@ -1,0 +1,61 @@
+from dataclasses import dataclass
+from typing import List, Optional
+
+from bs4 import BeautifulSoup
+
+
+@dataclass
+class AppInfo:
+    title: Optional[str] = None
+    favicon: Optional[str] = None
+    description: Optional[str] = None
+    image: Optional[str] = None
+    # ensure the meta tags are correct or the UI might fail to load.
+    meta_tags: Optional[List[str]] = None
+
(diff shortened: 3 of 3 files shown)


src/lightning_app/core/app.py
```diff
@@ -133,6 +136,9 @@ def __init__(
 
         logger.debug(f"ENV: {os.environ}")
 
+        # update index.html
```
alecmerdler, line 139: Make a comment on why this should be in the constructor (it should only be written once during the app lifecycle).

alecmerdler (approved):

requirements/app/base.txt
```diff
@@ -8,3 +8,4 @@ croniter>=1.3.0, <1.4.0  # strict; TODO: for now until we find something more ro
 traitlets>=5.3.0, <=5.4.0
 arrow>=1.2.0, <1.2.4
 lightning-utilities==0.3.*
+beautifulsoup4==4.8.2
```
tchaton, line 11: Why this is needed? Let's avoid adding new dependencies.

src/lightning_app/utilities/add_meta.py
```diff
@@ -0,0 +1,61 @@
...
+
+    with original_file.open() as f:
+        original = f.read()
+
+    with entry_file.open("w") as f:
+        f.write(_get_updated_content(original=original, info=info))
+
+
+def _get_updated_content(original: str, info: AppInfo) -> str:
+    soup = BeautifulSoup(original, "html.parser")
```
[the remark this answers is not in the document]
pritamsoni-hsr: with a parser it's very much clean, we can use regex but there is a high chance that we'll miss some edge cases, that's why a parser is good to use here.

also, we'll soon be expanding extra things that users can pass.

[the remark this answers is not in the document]
tchaton: Not a huge fan as any extra dependency makes maintainability harder.

Fine for now.

ethanwharris (approved):
LGTM :smiley:

kaushikb11 (approved):
LGTM!

requirements/app/base.txt
awaelchli, line 11:
```suggestion
beautifulsoup4<=4.8.2
```
It's important we don't pin. The pip dependency resolver is not exactly flexible. It is unlikely that for the changes here you require the exact version on the dot, and no other one works.

src/lightning_app/utilities/add_meta.py
```diff
@@ -0,0 +1,61 @@
+from dataclasses import dataclass
+from typing import List, Optional
+
+from bs4 import BeautifulSoup
+
+
+@dataclass
+class AppInfo:
```
awaelchli, line 8: Originally, the idea was to have that in the .lightning file that is stored alongside the app. Today we save only the name of the app there. 

Are we sure about splitting this up?

src/lightning_app/core/app.py
```diff
@@ -46,6 +48,7 @@ def __init__(
         self,
         root: "lightning_app.LightningFlow",
         debug: bool = False,
+        info: add_meta.AppInfo = None,
```
awaelchli, line 51: This argument is missing in the docstring.

awaelchli (commented):
I share same concerns with Thomas about the dependency. Have we looked at the code, are we happy with how it is maintained, how heavy/light it is etc?

src/lightning_app/core/app.py
nohalon: yeah @pritamsoni-hsr can you add `into` to the arguments on line 68

src/lightning_app/utilities/add_meta.py
manskx: Agree with @awaelchli, we already have `AppConfig` object: https://github.com/Lightning-AI/lightning/blob/7dbd038a9455f3c67da3fe19ef579058feef51b2/src/lightning_app/utilities/packaging/app_config.py#L13 which is the `.lightning` file and can be used to contain configurations or meta data of the app

src/lightning_app/core/app.py
```diff
@@ -18,12 +18,14 @@
...
     FLOW_DURATION_SAMPLES,
     FLOW_DURATION_THRESHOLD,
+    FRONTEND_DIR,
     STATE_ACCUMULATE_WAIT,
 )
 from lightning_app.core.queues import BaseQueue, SingleProcessQueue
 from lightning_app.frontend import Frontend
 from lightning_app.storage import Drive, Path
 from lightning_app.storage.path import storage_root_dir
+from lightning_app.utilities import add_meta
```
manskx, line 28: Maybe `add_meta` can be renamed to something else?

src/lightning_app/utilities/add_meta.py
pritamsoni-hsr: this sort of API requirement came from exec team.

src/lightning_app/core/app.py
```diff
@@ -133,6 +136,10 @@ def __init__(
 
         logger.debug(f"ENV: {os.environ}")
 
+        # update index.html,
+        # this should happen once for all apps before the ui server starts running.
+        add_meta.update_index_file_with_info(FRONTEND_DIR, info=info)
```
manskx, line 141: This means that the default `index.html` will be changed for all apps, is this intended?

src/lightning_app/core/app.py
pritamsoni-hsr: yeah I added this now and marked this in PR description.

src/lightning_app/core/app.py
pritamsoni-hsr: frontend.py

pritamsoni-hsr:
bs4 is around 2MB in size, @awaelchli. I think we should just use it instead of writing parsers from scratch.

src/lightning_app/core/app.py
manskx: Imagine these two apps:
app_1.py
```python

app = L.LightningApp(info=AppInfo(name="Im a cool app"))

```
the title of the HTML will be "Im a cool app"

app_2.py
```python

app = L.LightningApp()

```

if you run app_1 first then app_2, will app_2 have "Im a cool app" title ?

tchaton (approved):
Approving to unblock releasing!

src/lightning_app/core/app.py
pritamsoni-hsr: no, it'll be reverted to original.

PR #14926, opened 2022-09-28
Fix TPU test CI

carmocca:
## What does this PR do?

- Properly set the PR number
- Clean up logs
- Make fetching faster

Part of #14550 

None

cc @carmocca @akihironitta @borda @kaushikb11 @rohitgr7


awaelchli (approved):
You added Lite 🤩

awaelchli:
I need to look into the pickle issue ... We can backtrack changes we made to the multiprocess launcher recently, hmmm.

src/pytorch_lightning/strategies/launchers/xla.py
```diff
@@ -139,30 +138,3 @@ def _collect_rank_zero_results(self, trainer: "pl.Trainer", results: Any) -> Opt
...
-    nprocs: Optional[int] = None,
-    join: bool = True,
-    daemon: bool = False,
-    start_method: str = "spawn",
-) -> Optional[ProcessContext]:
-    """Wraps the :func:`torch_xla.distributed.xla_multiprocessing.spawn` with added teardown logic for the worker
-    processes."""
-
-    @wraps(fn)
-    def wrapped(rank: int, *_args: Any) -> None:
```
carmocca, line 156: Removed to fix

```py
E       _pickle.PicklingError: Can't pickle <function _XLALauncher._wrapping_function at 0x7f079bf74b90>: it's not the same object as pytorch_lightning.strategies.launchers.xla._XLALauncher._wrapping_function
```

tests/tests_pytorch/models/test_tpu.py
```diff
@@ -366,20 +324,23 @@ def test_if_test_works_with_checkpoint_false(tmpdir):
     assert trainer.state.finished, f"Training failed with {trainer.state}"
 
 
-@RunIf(tpu=True)
-def test_tpu_sync_dist():
-    """Test tpu spawn sync dist operation."""
+def xla_launch(function):
+    strategy = TPUSpawnStrategy(parallel_devices=list(range(8)))
+    launcher = _XLALauncher(strategy=strategy)
+    return launcher.launch(function, strategy)
```
awaelchli, line 330: @carmocca each "function" here must probably call `.setup_environment()` before you can call any distributed collectives. You are seeing the error about `Accessing the XLA device before processes have spawned is not allowed.` because the test is trying to access strategy.device before `.setup_environment()` is called. 

In Trainer, this is done right after processes get spawned, and in Lite this is done in the function that wraps around the run() method.

carmocca: Why would the launcher not take care of this?

awaelchli: I think you are getting confused. 
I can see that you called setup_environment on the main process with the trainer changes. This calls torch.cuda.set_device on the main process, which is not correct. The error message is well justified in this case. 

I would definitely call setup_environment in your special test manually. Otherwise you just end up implementing https://github.com/Lightning-AI/lightning/pull/11073

tests/tests_pytorch/models/test_tpu.py
```diff
@@ -366,20 +325,32 @@ def test_if_test_works_with_checkpoint_false(tmpdir):
     assert trainer.state.finished, f"Training failed with {trainer.state}"
 
 
-@RunIf(tpu=True)
-def test_tpu_sync_dist():
-    """Test tpu spawn sync dist operation."""
+def wrap_launch_function(fn, strategy, *args, **kwargs):
+    # FIXME: i think this requires https://github.com/Lightning-AI/lightning/pull/11073
```
carmocca, line 329: I'm getting:

```python
  File "/root/miniconda3/envs/lightning/lib/python3.7/site-packages/torch_xla/distributed/xla_multiprocessing.py", line 330, in _mp_start_fn
    _start_fn(index, pf_cfg, fn, args)
  File "/root/miniconda3/envs/lightning/lib/python3.7/site-packages/torch_xla/distributed/xla_multiprocessing.py", line 324, in _start_fn
    fn(gindex, *args)
  File "/home/runner/work/lightning/src/pytorch_lightning/strategies/launchers/xla.py", line 101, in _wrapping_function
    results = function(*args, **kwargs)
  File "/home/runner/work/lightning/tests/tests_pytorch/models/test_tpu.py", line 331, in wrap_launch_function
    strategy.setup_environment()
  File "/home/runner/work/lightning/src/pytorch_lightning/strategies/strategy.py", line 130, in setup_environment
    assert self.accelerator
```

tests/tests_pytorch/accelerators/test_tpu.py
```diff
@@ -66,14 +69,16 @@ def test_resume_training_on_cpu(tmpdir):
...
 def test_if_test_works_after_train(tmpdir):
     """Ensure that .test() works after .fit()"""
-
-    # Train a model on TPU
     model = BoringModel()
     trainer = Trainer(max_epochs=1, accelerator="tpu", devices=8, default_root_dir=tmpdir, fast_dev_run=True)
-    trainer.fit(model)
-    assert len(trainer.test(model)) == 1
+    # https://github.com/pytorch/xla/issues/1666
+    with pytest.raises(ProcessRaisedException, match="OSError: libmkl_intel_lp64"):
```
carmocca, line 78: I'm hoping that unpinning `mkl` will resolve this after the new docker image is published

src/pytorch_lightning/strategies/launchers/multiprocessing.py
```diff
@@ -97,7 +98,9 @@ def launch(self, function: Callable, *args: Any, trainer: Optional["pl.Trainer"]
         # The default cluster environment in Lightning chooses a random free port number
         # This needs to be done in the main process here before starting processes to ensure each rank will connect
         # through the same port
-        os.environ["MASTER_PORT"] = str(self._strategy.cluster_environment.main_port)
+        cluster_environment = self._strategy.cluster_environment or LightningEnvironment()
```
awaelchli, line 101: this would actually never happen with the Trainer/Lite. 

`_MultiProcessingLauncher` is not meant to be used with arbitrary things as input. It assumes that the strategy given here is properly set up by the Connector. It is why these launchers are not public API.

tests/tests_pytorch/models/test_tpu.py
```diff
@@ -366,20 +325,32 @@ def test_if_test_works_with_checkpoint_false(tmpdir):
...
+
 
-    def test_sync_dist(rank):
-        trainer = Trainer(accelerator="tpu", devices=8)
-        trainer.strategy._launched = True
+def xla_launch(fn):
+    strategy = TPUSpawnStrategy(parallel_devices=list(range(8)))
+    launcher = _XLALauncher(strategy=strategy)
+    wrapped = partial(wrap_launch_function, fn, strategy)
+    return launcher.launch(wrapped, strategy)
```
awaelchli, line 340: You need to set the accelerator here. 
```suggestion
    strategy = TPUSpawnStrategy(parallel_devices=list(range(8)), accelerator=TPUAccelerator())
    launcher = _XLALauncher(strategy=strategy)
    wrapped = partial(wrap_launch_function, fn, strategy)
    return launcher.launch(wrapped, strategy)
```

src/pytorch_lightning/strategies/launchers/xla.py
```diff
@@ -139,30 +135,3 @@ def _collect_rank_zero_results(self, trainer: "pl.Trainer", results: Any) -> Opt
         self.add_to_queue(trainer, extra)
 
         return _WorkerOutput(best_model_path, weights_path, trainer.state, results, extra)
-
-
-def _save_spawn(
```
awaelchli, line 144: So the conclusion was that this save_spawn was somehow the reason for the original pickling issue we discussed on this pr?

I'm fine with having the alternative of
`_rank_teardown(process_idx)`, but the benefit of the save_spawn function was that if you used it, it would guarantee you with the correct teardown and you don't forget to put special teardown logic in your own function.

tests/tests_pytorch/models/test_tpu.py
awaelchli: See: https://github.com/Lightning-AI/lightning/pull/14926#discussion_r983551597

src/pytorch_lightning/strategies/launchers/xla.py
carmocca: I believe it was. At least the error disappear after removing it.

I agree with your second point. I'm going to open a separate PR proposing to use the launchers in our tests to launch

src/pytorch_lightning/strategies/launchers/multiprocessing.py
carmocca: I think it's good to allow using the launchers like this. Makes launching processes super easy and simple. I am going to move this into a separate PR though.

carmocca: Synced in a call: dropping this change, hopefully we can do this one day with lazy init.

tests/tests_pytorch/models/test_tpu.py
carmocca: Done. I've accepted that we dont have lazy init yet.

dockers/tpu-tests/tpu_test_cases.jsonnet
```diff
@@ -20,29 +20,41 @@ local tputests = base.BaseTest {
...
-      pip install -e .[test]
+
+      echo "--- Fetch the SHA's changes ---"
+      git clone --single-branch --depth 1 https://github.com/Lightning-AI/lightning.git /home/runner/work/lightning
+      cd home/runner/work/lightning
+      git fetch origin --depth 1 pull/{PR_NUMBER}/head:test/{PR_NUMBER}
+      git -c advice.detachedHead=false checkout {SHA}
+
+      echo "--- Install PL ---"
+      pip install mkl-2022.1.0  # failed list: 2021.4.0, 2022.2.0
```
carmocca, line 36: FIXME: debugging different versions... :skull:

carmocca:
Debugging recap:

`mkl==2021.4.0`, the version we use in CI currently, does have the `.so` inside
```bash
find / -name 'libmkl*' | grep libmkl_intel_lp64.so.1
/root/miniconda3/envs/lightning/lib/libmkl_intel_lp64.so.1
```
and `python -c "import torch"` works.
yet the `OSError` still appears in the spawn test:
```python
  File "<string>", line 1, in <module>
  File "/root/miniconda3/envs/lightning/lib/python3.7/multiprocessing/spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "/root/miniconda3/envs/lightning/lib/python3.7/multiprocessing/spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
  File "/root/miniconda3/envs/lightning/lib/python3.7/site-packages/torch/__init__.py", line 201, in <module>
    _load_global_deps()
  File "/root/miniconda3/envs/lightning/lib/python3.7/site-packages/torch/__init__.py", line 154, in _load_global_deps
    ctypes.CDLL(lib_path, mode=ctypes.RTLD_GLOBAL)
  File "/root/miniconda3/envs/lightning/lib/python3.7/ctypes/__init__.py", line 364, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libmkl_intel_lp64.so.1: cannot open shared object file: No such file or directory
  File "<string>", line 1, in <module>
  File "/root/miniconda3/envs/lightning/lib/python3.7/multiprocessing/spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "/root/miniconda3/envs/lightning/lib/python3.7/multiprocessing/spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
  File "/root/miniconda3/envs/lightning/lib/python3.7/site-packages/torch/__init__.py", line 201, in <module>
    _load_global_deps()
  File "/root/miniconda3/envs/lightning/lib/python3.7/site-packages/torch/__init__.py", line 154, in _load_global_deps
    ctypes.CDLL(lib_path, mode=ctypes.RTLD_GLOBAL)
  File "/root/miniconda3/envs/lightning/lib/python3.7/ctypes/__init__.py", line 364, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libmkl_intel_lp64.so.1: cannot open shared object file: No such file or directory
```

Lastest `mkl` doesn't even provide the dll (find doesn't show it). Just `python -c "import torch"` triggers the `OSError`

So I don't think we can switch change the mkl version.

`torch_xla` installs `mkl-include` in their CI: https://github.com/pytorch/xla/blob/3eddd035a1d6270b6b07902ffeb84f2af486196e/scripts/build_torch_wheels.sh#L182 but I didn't see a difference when doing the same.

Worst case we just merge this with the skipped TPU test.

awaelchli:
@carmocca I honestly think skipping that one test that is provoking this error is totally fine, as it is not very critical. Maybe one more thing one could do is clone the branch into Colab and run some tests there just as a sanity check manually. But I wouldn't spend too much time on it.

carmocca:
It worries me a bit because the tests just fast-dev-run trains and tests a BoringModel, which is the simplest integration test we could have. We might have to skip more tests, I just don't know yet because getting one run to happen is really difficult, given the availability.

carmocca:
I've run TPUs 10 times and cannot get them to provide a machine once. I propose we merge this as it is to at least fix the launching code and continue fixing tests separately.

awaelchli (approved):

otaj (approved):

PR #14927, opened 2022-09-28
An instance of SaveConfigCallback should only save the config once

mauvilsa:
## What does this PR do?

When using `LightningCLI(.., run=False)` it is very common to call multiple `Trainer` methods, e.g. `fit` then `test`. In the second method call, the `SaveConfigCallback` attempts to save the config again, which by default would fail because it refuses to overwrite. This makes it necessary to have `save_config_overwrite=True` even though one might not want this. This can be observed in some places, for example https://github.com/Lightning-AI/lightning/issues/1715#issuecomment-1207380816. 

This change is so that a single instance of `SaveConfigCallback` only attempts to save the config once, avoiding the problem described before. I would consider this to be a bug fix.

None

## Before submitting

## PR review

## Did you have fun?


awaelchli:
I think I agree with the changes. However, I don't understand this sentence:

> This makes it necessary to have save_config_overwrite=True even though one might not want this.

In order to reproduce, I think it makes sense to have run=False, overwrite=False, because we want to save the config of the initial fit/test. If we were to overwrite, we would end up with the config of the last ever called fit/test, but that would make it not reproducible.

The logic in the code of this PR however makes sense to me, and I think it implements what I mean.

awaelchli (approved):

carmocca (approved):

tchaton (approved):
LGTM !

otaj (approved):

src/pytorch_lightning/cli.py
```diff
@@ -238,6 +242,7 @@ def setup(self, trainer: Trainer, pl_module: LightningModule, stage: str) -> Non
             self.parser.save(
                 self.config, config_path, skip_none=False, overwrite=self.overwrite, multifile=self.multifile
             )
+            self.already_saved = True
```
carmocca, line 245: The GPU job is getting canceled because an example is hanging. It's caused by this PR because it happened 3 times and I haven't seen it in any other PR.

This attribute only gets set on global zero

awaelchli: Perhaps a `self.already_saved = trainer.strategy.broadcast(self.already_saved)` will do here, at the end of the setup()?

awaelchli: It succeeded. Resolve the conversation to auto merge it if you are happy :)

mauvilsa: I am fine with the change, though I don't understand how the broadcast works. Could it be instead a single assignment `self.already_saved = trainer.strategy.broadcast(True)` inside the `if trainer.is_global_zero:`?

awaelchli: Broadcast sends the value passed in from processes 0 to all other processes. In order for this to work, all processes need to reach the broadcast call. Here we can use it so that all processes end up with the same value for already_saved. This matters when we have multiple fit calls, and enter setup again subsequently. If the already_saved value wasn't the same for all processes, and say after the first setup already_saved was True on rank 0 but false on rank 1, then in the next call to setup(), process 0 would immediately exit the setup while process 1 would execute the body of it. But process 1 will hang, because it will hit the line `file_exists = trainer.strategy.broadcast(file_exists)`, and wait there indefinitely for process 0 to join. 

Given that the only time already_saved is modified is in the if condition above, then one could also do this:

```diff
index 981372cc2..5eea3d97d 100644
--- a/src/pytorch_lightning/cli.py
+++ b/src/pytorch_lightning/cli.py
@@ -242,10 +242,9 @@ class SaveConfigCallback(Callback):
             self.parser.save(
                 self.config, config_path, skip_none=False, overwrite=self.overwrite, multifile=self.multifile
             )
-            self.already_saved = True

-        # broadcast so that all ranks are in sync on future calls to .setup()
-        self.already_saved = trainer.strategy.broadcast(self.already_saved)
+        self.already_saved = True
+        trainer.strategy.barrier()

```

> Could it be instead a single assignment self.already_saved = trainer.strategy.broadcast(True) inside the if trainer.is_global_zero:?

No. As mentioned briefly above, all processes need to enter the call to broadcast. Furthermore, broadcasting a constant would be redundant, as it would be equivalent to just setting self.already_saved = True on all ranks. 

I hope this helps answer the question.

mauvilsa: Thank you for the explanation. I am fine with any solution. Do you think that the `barrier` case makes the code more understandable?

awaelchli: IMO both are equally readable and understandable. The barrier version is a tiny bit in the lead in terms of performance/sync-overhead, because the broadcast has to convert the bool to a tensor and then back to a bool.

PR #14057, opened 2022-08-06
CI: Update `base-ipu` docker image

akihironitta:
## What does this PR do?
- Resolves timeout in master due to very slow dependency resolution by conda. (from `>100min` to `<10min`)
- Most of the change comes from the `base-cuda` dockerfile.

### Does your PR introduce any breaking changes? If yes, please list them.
None

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [n/a] Did you make sure to **update the documentation** with your changes? (if necessary)
- [n/a] Did you write any **new necessary tests**? (not for typos and docs)
- [n/a] Did you verify new and **existing tests pass** locally with your changes?
- [n/a] Did you list all the **breaking changes** introduced by this pull request?
- [n/a] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @carmocca @akihironitta @borda

PR #14934, opened 2022-09-29
Improve building times of IPU docker image

otaj:
## What does this PR do?

IPU docker image build quite often fails because it is too slow. This PR rearranges couple things to improve build times, whereas the image should remain mostly the same

### DIFF
--- a/dockers/base-ipu/Dockerfile
+++ b/dockers/base-ipu/Dockerfile
@@ -18,6 +18,7 @@ LABEL maintainer="Lightning-AI <https://github.com/Lightning-AI>"
 
 ARG PYTHON_VERSION=3.9
 ARG CONDA_VERSION=4.9.2
+ARG PYTORCH_VERSION=1.9
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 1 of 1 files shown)


carmocca (approved):

kaushikb11 (approved):

dockers/base-ipu/Dockerfile
```diff
@@ -56,13 +57,12 @@ ENV \
 
 COPY environment.yml environment.yml
 
-RUN conda create -y --name $CONDA_ENV python=${PYTHON_VERSION} pytorch=${PYTORCH_VERSION} cudatoolkit=${CUDA_VERSION} -c pytorch && \
-    conda init bash && \
+RUN conda init bash && \
```
justusschock, line 60: one general question: why don't we start with a miniconda image in the first place?

justusschock (approved):

dockers/base-ipu/Dockerfile
otaj: No idea. I can refactor the image to it, but that's not work for this PR, in this one I tried to just improve building times while keeping the image as similar to the original as possible

justusschock: agreed, but this would further reduce building time on our end :)

carmocca: Resolving to unblock master. We can consider further reductions separately

akihironitta:
Thank you @otaj! Will follow up in my PR #14057 to further reduce the build time.

PR #14962, opened 2022-09-30
Remove unused gcsfs dependency

carmocca:
## What does this PR do?

Closes #14788

None


otaj (approved):

tests/tests_pytorch/utilities/test_remote_filesystem.py
```diff
@@ -1,106 +0,0 @@
...
-
-import fsspec
-import pytest
-
-from pytorch_lightning import Trainer
-from pytorch_lightning.callbacks import ModelCheckpoint
-from pytorch_lightning.demos.boring_classes import BoringModel
-from pytorch_lightning.loggers import TensorBoardLogger
-
-GCS_BUCKET_PATH = os.getenv("GCS_BUCKET_PATH", None)
```
awaelchli, line 24: have we considered adding the bucket path to CI and moving the requirement to test.txt?

carmocca: I mentioned that as a different option in https://github.com/Lightning-AI/lightning/pull/14788#discussion_r984104347.

I lean towards removal because we have no logic specific to gcs in the codebase. This is all encapsulated by `fsspec`, and we have tests for our integration with it. I assume fsspec has gcs specific tests.

awaelchli: Totally fine with me, just making sure we're not removing something that was needed

awaelchli (approved):

Issue #9234, opened 2021-08-31
Create Mixin for block backward sync and deprecate block_backward_sync from Parallel Plugin


four4fish:
## Proposed refactoring or deprecation

Discussed in https://github.com/PyTorchLightning/pytorch-lightning/pull/9101

### Motivation

Parallel plugin should be generic and self-contained. Move reference to DistributedDataParallel from parallel block_backward_sync

### Pitch

Discussed in https://github.com/PyTorchLightning/pytorch-lightning/pull/9101
@ananthsub  suggested create interface for this trait
`class PluginWithBlockBackwardSync(ABC):

    @contextmanager
    @abstractmethod
    def block_backward_sync(self) -> Generator:`

### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

<sub>

- [**Metrics**](https://github.com/PyTorchLightning/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Flash**](https://github.com/PyTorchLightning/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, finetuning and solving problems with deep learning

- [**Bolts**](https://github.com/PyTorchLightning/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks and more for research and production with PyTorch Lightning and PyTorch

- [**Lightning Transformers**](https://github.com/PyTorchLightning/lightning-transformers): Flexible interface for high performance research using SOTA Transformers leveraging Pytorch Lightning, Transformers, and Hydra.

</sub>

PR #14952, opened 2022-09-29
Merge DDPStrategy and DDPSpawnStrategy in Lite

awaelchli:
## What does this PR do?

Fixes #14707 

### Does your PR introduce any breaking changes? If yes, please list them.

Yes.

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @borda @justusschock @awaelchli @carmocca

PR #14966, opened 2022-09-30
Efficient gradient accumulation in LightningLite

### LINKED ISSUE #14713: Provide a method to skip gradient synchronization in LightningLite
## 🚀 Feature

Provide an easy to use method to skip the gradient synchronization.

### Motivation

Skipping the synchronization of gradients in `.backward()` is useful when accumulating gradients in the training loop, cutting down communication overhead for DDP and the sort. 

### Pitch

Introduce a context manager like we have internally in PL:

```py
Lite.skip_gradient_sync(model)
Lite.skip_backward_sync(model)
Lite.block_gradient_sync(model)
Lite.block_backward_sync(model)
Lite.enable_gradient_sync(model, True|False)
Lite.disable_gradient_sync(model, True|False)
Lite.no_sync(model, enabled=True|False)
Lite.no_sync(model, enabled=True|False)
```

Usage in the training loop:

```py

num_accumulation_steps = 8
for step, (x, y) in enumerate(train_dataloader):
    
    with self.no_sync(model, step % num_accumulation_steps != 0)
        out = model(x)
        loss = loss_fn(out, y)
        self.backward(loss)

    if step % num_accumulation_steps == 0:
        # apply accumulated gradients
        optimizer.step()
        optimizer.zero_grad()

```

### Additional context

awaelchli:
## What does this PR do?

Fixes #14713
Also sets the stage for #9234

```py

# Accumulate gradient 8 batches at a time
is_accumulating = batch_idx % 8 != 0

with self.skip_backward_sync(model, enabled=is_accumulating):
    output = model(input)
    loss = ...
    self.backward(loss)
    ...

if not is_accumulating:
    optimizer.step()
    optimizer.zero_grad()
```

None.

## Before submitting

## PR review

## Did you have fun?


src/lightning_lite/lite.py
```diff
@@ -345,6 +345,49 @@ def all_gather(
...
+        if not isinstance(module, _LiteModule):
+            raise TypeError(
+                "You need to set up the model first before you can call `self.skip_backward_sync()`:"
+                " `self.setup(model, ...)`"
+            )
+        if self._strategy._backward_sync_control is None:
+            raise TypeError(
+                f"The `{self._strategy.__class__.__name__}` does not support skipping the gradient synchronization."
+                f" Remove `.skip_backward_sync()` from your code or choose a differnt strategy."
+            )
```
awaelchli, line 381: I think it makes sense to error out, given that not all strategies support this. However, this also means that the code is not fully strategy agnostic, switching between accelerators and strategies will require you to change the code. 

Alternatively this could be a warning. Then the context manager becomes a no-op and will emit a warning that the strategy does not support skipping sync. 

- 🚀 Error (current)
- 🎉 Warning

src/lightning_lite/strategies/ddp_spawn.py
```diff
@@ -26,6 +26,7 @@
 from lightning_lite.plugins.environments.cluster_environment import ClusterEnvironment
 from lightning_lite.plugins.io.checkpoint_plugin import CheckpointIO
 from lightning_lite.plugins.precision import Precision
+from lightning_lite.strategies.ddp import _DDPBackwardSyncControl
```
awaelchli, line 29: Note: The two strategies will merge in #14952

src/lightning_lite/lite.py
justusschock: TBH, I don't think we should error out and I think that for some strategies it makes sense to warn if there is no support for this, for others however, like the single device (and yes, we should support it there as a no-op to be fully agnostic!) it does not make sense to warn as there is no backward sync happening anyways.

So I guess I propose the following:

- Only warn if a backward sync is happening and cannot be prevented and have a no-op
- Have a no-op if prevention is not necessary but still no backward sync happens
- Have the actual implementation if required

src/lightning_lite/lite.py
```diff
@@ -345,6 +345,49 @@ def all_gather(
     def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
         return self._strategy.broadcast(obj, src=src)
 
+    @contextmanager
+    def skip_backward_sync(self, module: _LiteModule, enabled: bool = True) -> Generator:
```
justusschock, line 349: I'd rename this to `no_backward_sync` or just `no_sync` as done in torch. This would allow use to include this for single device strategies and have a no-op if necessary even though we technically don't skip it. This is also more aligned with torch api as `torch.no_grad`

justusschock (approved):
ltgm in general few comments regarding naming and errors

src/lightning_lite/lite.py
awaelchli: How would you actually implement these distinctions though? Requiring a strategy to implement the interface just to warn seems like unnecessary boilerplate.

src/lightning_lite/lite.py
awaelchli: no_backward_sync is fine with me. no_sync is proposed in https://github.com/Lightning-AI/lightning/issues/14713 too but I've stayed away from it because the expression is too general. I think it would be better to have the name specifically point out that this applies to syncing gradients from backward, and nothing else.

src/lightning_lite/lite.py
```diff
@@ -345,6 +345,49 @@ def all_gather(
...
+                skip.
+        """
+
+        if not isinstance(module, _LiteModule):
+            raise TypeError(
+                "You need to set up the model first before you can call `self.skip_backward_sync()`:"
+                " `self.setup(model, ...)`"
+            )
+        if self._strategy._backward_sync_control is None:
+            raise TypeError(
```
carmocca, line 378: I'm not sure this should be an exception. This means changing strategies might require code changes. I suggest relaxing it to a warning

src/lightning_lite/strategies/ddp.py
```diff
@@ -178,3 +180,18 @@ def _determine_ddp_device_ids(self) -> Optional[List[int]]:
...
+        """Blocks gradient synchronization inside the
+        :class:`~torch.nn.parallel.distributed.DistributedDataParallel` wrapper."""
+        if not isinstance(module, DistributedDataParallel):
+            raise TypeError(
+                "Blocking backward sync is only possible if the module passed to"
+                f" `{self.__class__.__name__}.skip_backward_sync` is wrapped in `DistributedDataParallel`."
+                f" Got: {module.__class__.__name__}."
+            )
+        with module.no_sync():  # type: ignore[operator]
+            yield None
```
carmocca, line 197: nit:
```suggestion
            yield
```

src/lightning_lite/lite.py
carmocca: +1 to `no_backward_sync`

carmocca (approved):

src/lightning_lite/lite.py
carmocca: Just saw the discussion above

src/lightning_lite/lite.py
carmocca: Can't we just check whether it's a parallel strategy in addition `sync_control is None`?

src/lightning_lite/lite.py
```diff
@@ -345,6 +345,49 @@ def all_gather(
...
+                f"The `{self._strategy.__class__.__name__}` does not support skipping the gradient synchronization."
+                f" Remove `.skip_backward_sync()` from your code or choose a differnt strategy."
```
carmocca, line 380:
```suggestion
                f" Remove `.skip_backward_sync()` from your code or choose a different strategy."
```

src/lightning_lite/lite.py
awaelchli: @carmocca @justusschock in your preferred choice for no_backward_sync, have you considered the semantics wrt. the enabled=True/False argument. 

One can argue that no_backward_sync(enabled=True) is confusing as it is not clear whether it is doing the sync or not without reading the documentation. It is the same in torch with no_grad, I know. But just so that you are aware of it.

carmocca: I thought about it, but prefer to be consistently inconsistent with what torch does.

src/lightning_lite/lite.py
awaelchli: I made is no-op for single device and warn otherwise. But mypy had a problem with it, so I ignored it.

docs/source-pytorch/starter/lightning_lite.rst
```diff
@@ -725,3 +725,29 @@ the data is written to disk.
...
+        loss = ...
+        self.backward(loss)
+        ...
+
+    # Step the optimizer every 8 batches
+    if not is_accumulating:
+        optimizer.step()
+        optimizer.zero_grad()
+
+Not all strategies support it. Both the model's `.forward()` and the `self.backward()` call need to run under
```
tchaton, line 752: Can you list which strategies don't support it ? I think DeepSpeed might work with this one no ?

tchaton (approved):
LGTM !

docs/source-pytorch/starter/lightning_lite.rst
awaelchli: I can, the list will probably get outdated but we will try to keep it updated. 
My understanding is that [DeepSpeed handles accumulation internally](https://ghpages-test.readthedocs.io/en/latest/config.html#deepspeed.config.BatchConfig.gradient_accumulation_steps) in the DeepSpeedEngine (you configure it through the config dict/file), and thus we don't have influence on it.

PR #14972, opened 2022-10-02
feat: allow root path to run the app on `/path`

pritamsoni-hsr:
## What does this PR do?

allow apps to be served from a custom path when running app behind a proxy.

by default apps are served from `https://<app_id>.litng-ai-03.litng.ai`, this PR adds a feature so apps can be served from 'https://mydomain.com/<my_app>'.

if you want to build custom UI, make sure all internal links are prefixed with `window.app_prefix`

- this works only behind a proxy and not on the `api_id.litng-ai-03.litng.ai/<custom_path>` and this is because of uvicorn's root path param.

## Limitations 
relative links inside the app will break, so user should use `window.app_prefix` in their JS, there is no warning around that atm.

## breaking changes
none

## Before submitting

## PR review

## Did you have fun?


hhsecond:
> relative links inside the app will break, so user should use window.app_prefix in their JS, there is no warning around that atm.
Can this be updated in the docs?

src/lightning_app/core/app.py
```diff
@@ -67,6 +68,9 @@ def __init__(
                 This can be helpful when reporting bugs on Lightning repo.
             info: Provide additional info about the app which will be used to update html title,
                 description and image meta tags and specify any additional tags as list of html strings.
+            root_path: Set this to `/path` if you want to run your app behind a proxy at `/path` leave empty for "/".
```
hhsecond, line 71: can you clarify what's proxy in this context? Our users might not know what would that be? Perhaps an example?

Also, in the next line, it is mentioned that the app won't start but not clear what should be done for it to actually work!

src/lightning_app/utilities/frontend.py
```diff
@@ -27,19 +32,29 @@ def update_index_file_with_info(ui_root: str, info: AppInfo = None) -> None:
...
-    if not info:
-        return
+    if info:
+        with original_file.open() as f:
+            original = f.read()
 
-    original = ""
+        with entry_file.open("w") as f:
+            f.write(
+                rewrite_static_with_root_path(_get_updated_content(original=original, root_path=root_path, info=info))
```
hhsecond, line 41: do we have to pipe this call? Can we `rewrite_static_with_root_path` happens inside `_get_updated_content`.

And probably not even a function, just a section inside `_get_updated_content`
?

src/lightning_app/utilities/frontend.py
```diff
@@ -14,10 +14,15 @@ class AppInfo:
     meta_tags: Optional[List[str]] = None
 
 
-def update_index_file_with_info(ui_root: str, info: AppInfo = None) -> None:
+def update_index_file_with_info_and_root_path(
```
hhsecond, line 17: perhaps just `update_index_file`?

hhsecond: the arguments says what's being used for updating it

hhsecond (commented):
Some minor nits and clarification on the docs

tchaton (approved):
LGTM!
@pritamsoni-hsr Do you think it would make sense to extract those metadata directly into a private file under a `.lightning` folder?

Borda:
> Do you think it would make sense to extract those metadata directly into a private file under a `.lightning` folder?

let's leave it s separate feature :)

src/lightning_app/utilities/frontend.py
kaushikb11: good point!

kaushikb11: updated!

src/lightning_app/core/app.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         root: "lightning_app.LightningFlow",
         debug: bool = False,
         info: frontend.AppInfo = None,
+        root_path: str = "",
```
awaelchli, line 52: the text below says that for customization, I should set it to "/path". By extension, shouldn't this then default to "/"?

src/lightning_app/core/app.py
awaelchli: I agree that the second sentence should probably point to some resource for further details :)

src/lightning_app/frontend/web.py
```diff
@@ -20,6 +20,7 @@ class StaticWebFrontend(Frontend):
 
     Arguments:
         serve_dir: A local directory to serve files from. This directory should at least contain a file `index.html`.
+                root_path: A path prefix when routing traffic from behind a proxy at `<root_path>`
```
awaelchli, line 23: malformated

src/lightning_app/frontend/frontend.py
```diff
@@ -15,21 +15,21 @@ def __init__(self) -> None:
...
+    def start_server(self, host: str, port: int, root_path: str = "") -> None:
         """Start the process that serves the UI at the given hostname and port number.
 
         Arguments:
             host: The hostname where the UI will be served. This gets determined by the dispatcher (e.g., cloud),
                 but defaults to localhost when running locally.
             port: The port number where the UI will be served. This gets determined by the dispatcher, which by default
                 chooses any free port when running locally.
-
+            root_path: root_path for the server if app in exposed via a proxy at `/<root_path>`
```
awaelchli, line 26:
```suggestion
            root_path: root_path for the server if app in exposed via a proxy at `/<root_path>`
            
```
You need to leave an empty line

awaelchli: Can we include an example like `https://domain.com/<root_path>/my_page` to visualize what it means.

tests/tests_app/frontend/test_web.py
```diff
@@ -54,17 +55,21 @@ def test_start_server_through_function(uvicorn_mock, tmpdir, monkeypatch):
     FastAPIMock.get.return_value = FastAPIGetDecoratorMock
     monkeypatch.setattr(lightning_app.frontend.web, "FastAPI", MagicMock(return_value=FastAPIMock))
 
-    lightning_app.frontend.web.start_server(serve_dir=tmpdir, host="myhost", port=1000, path="/test-flow")
-    uvicorn_mock.run.assert_called_once_with(app=ANY, host="myhost", port=1000, log_config=ANY)
-    FastAPIMock.mount.assert_called_once_with("/test-flow", ANY, name="static")
+    root_path = "/base"
```
awaelchli, line 58: could you parametrize the test with `root_path=("/", "/base")` please?

src/lightning_app/utilities/frontend.py
kaushikb11: updated!

src/lightning_app/core/app.py
kaushikb11: uvicorn and fastapi both have empty string as default:
https://github.com/encode/uvicorn/blob/master/uvicorn/main.py#L248
https://github.com/tiangolo/fastapi/blob/master/fastapi/applications.py#L77

tests/tests_app/frontend/test_web.py
kaushikb11: updated!

ethanwharris (approved):

src/lightning_app/core/app.py
kaushikb11: @pritamsoni-hsr

manskx (approved):

src/lightning_app/core/app.py
pritamsoni-hsr: can we add this url? https://www.fortinet.com/resources/cyberglossary/proxy-server

pritamsoni-hsr:
@tchaton we could, but I think maintaining it in the app is easiest, we can later serialize and deserialize `.lightning` file

kaushikb11 (approved):

PR #14992, opened 2022-10-04
Launch options for Lightning Lite

awaelchli:
## What does this PR do?

Subclassing and overriding `LightningLite.run()` is no longer required. You can create and use the `LightningLite` object directly:

```py
import torch.distributed
from lightning_lite import LightningLite

lite = LightningLite()
print("launched", lite.global_rank)
assert torch.distributed.is_initialized()
lite.barrier()
print("end")
```
To launch the script, a special CLI needs to be used:

```
python -m lightning_lite.cli example.py --accelerator cuda --devices 2 --precision bf16
```
The command above will eventually become `lightning run ...` in the future.

Alternatively, one can launch programmatically:

```py
lite = LightningLite(accelerator="cuda", devices=2)
lite.launch()
...
```

In Jupyter notebooks (Colab, Kaggle etc.) you would launch like this:

```py

# Notebook cell 1
def training_function(lite):
    lite.setup(model) 
    ...

# Notebook cell 2
lite = LightningLite(accelerator="cuda", devices=2)
lite.launch(training_function)
```

## Before submitting

## PR review

## Did you have fun?


justusschock (approved):

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,146 @@
...
+        type=str,
+        default="cpu",
+        choices=("cpu", "gpu", "cuda", "mps", "tpu"),
+        help="The hardware accelerator to run on.",
+    )
+    parser.add_argument(
+        "--strategy",
+        type=str,
+        default=None,
+        choices=(None, "ddp", "dp", "deepspeed"),
```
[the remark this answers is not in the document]
awaelchli: I added it to the top of the file in a constant, but in general the cli can't support the same set of strategy inputs as the Lite class itself (right  now). Example: ddp_spawn

src/lightning_lite/lite.py
```diff
@@ -450,3 +480,11 @@ def _validate_setup_dataloaders(dataloaders: Sequence[DataLoader]) -> None:
 
         if any(not isinstance(dl, DataLoader) for dl in dataloaders):
             raise TypeError("Only PyTorch DataLoader are currently supported in `setup_dataloaders`.")
+
+
+def _is_using_cli() -> bool:
+    return "LT_ACCELERATOR" in os.environ
```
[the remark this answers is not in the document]
awaelchli: Done!

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,146 @@
...
+from argparse import ArgumentParser, Namespace
+from typing import List, Tuple
+
+import torch.distributed.run as torchrun
+
+from lightning_lite.accelerators import CPUAccelerator, CUDAAccelerator, MPSAccelerator
+
+
+def _parse_args() -> Tuple[Namespace, List[str]]:
+    parser = ArgumentParser()
```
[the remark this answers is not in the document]
awaelchli: Done!

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+        "--devices",
+        type=str,
+        default="1",
+        help=(
+            "Number of devices to run on (``int``), which devices to run on (``list`` or ``str``), or ``'auto'``."
+            " The value applies per node."
+        ),
+    )
+    parser.add_argument(
+        "--num-nodes",
```
carmocca, line 55: nit: I would prefer the `--num_nodes` pattern (underscore instead of dash), as it matches arguments in Python.

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+            "Double precision (``64``), full precision (``32``), half precision (``16``) or bfloat16 precision"
+            " (``'bf16'``)"
+        ),
+    )
+
+    args, script_args = parser.parse_known_args()
+    return args, script_args
+
+
+def _set_env_variables(args: Namespace) -> None:
```
carmocca, line 96: Do you think these should be popped after the launched function finishes?

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+    os.environ["LT_CLI_USED"] = "1"
+    os.environ["LT_ACCELERATOR"] = str(args.accelerator)
+    if args.strategy is not None:
+        os.environ["LT_STRATEGY"] = str(args.strategy)
+    os.environ["LT_DEVICES"] = str(args.devices)
+    os.environ["LT_NUM_NODES"] = str(args.num_nodes)
+    os.environ["LT_PRECISION"] = str(args.precision)
+
+
+def _get_num_processes(accelerator: str, devices: str) -> int:
```
carmocca, line 110: Should we move this to the connector? Might be useful in the future

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+    """Parse the `devices` argument to determine how many processes need to be launched on the current machine."""
+    if accelerator in ("cuda", "gpu"):
+        parsed_devices = CUDAAccelerator.parse_devices(devices)
+    elif accelerator in ("mps", "gpu"):
```
carmocca, line 114: If "gpu" is passed, it will always use the CUDAAccelerator. I think it needs to manually call `parse_gpu_ids`

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+        parsed_devices = MPSAccelerator.parse_devices(devices)
+    elif accelerator == "tpu":
+        raise ValueError("Launching processes for TPU through the CLI is not supported.")
```
carmocca, line 117: Is this a real limitation or a we-haven't-done-it-yet limitation?

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+        return CPUAccelerator.parse_devices(devices)
+    return len(parsed_devices) if parsed_devices is not None else 0
+
+
+def _torchrun_launch(args: Namespace, script_args: List[str]) -> None:
+    """This will invoke `torchrun` programmatically to launch the given script in new processes."""
+    if args.strategy == "dp":
+        num_processes = 1
+    else:
+        num_processes = _get_num_processes(args.accelerator, args.devices)
```
carmocca, line 128: (Question, not for this PR) Some strategies define `strategy.num_processes`. Do you think it should be moved to the accelerator?

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+
+    torchrun_args = []
```
carmocca, line 130: nit: this could be initialized with the flags below.

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,151 @@
...
+    torchrun_args.extend(["--nproc_per_node", str(num_processes)])
+    torchrun_args.extend(["--nnodes", str(args.num_nodes)])
+    torchrun_args.extend(["--node_rank", str(args.node_rank)])
+    torchrun_args.extend(["--master_addr", args.main_address])
+    torchrun_args.extend(["--master_port", str(args.main_port)])
+    torchrun_args.append(args.script)
+    torchrun_args.extend(script_args)
+
+    # set a good default number of threads for OMP to avoid warnings being emitted to the user
+    os.environ.setdefault("OMP_NUM_THREADS", str(max(1, (os.cpu_count() or 1) // num_processes)))
```
carmocca, line 140: Why do we need `max`?

```suggestion
    os.environ.setdefault("OMP_NUM_THREADS", str((os.cpu_count() or 1) // num_processes))
```

src/lightning_lite/connector.py
```diff
@@ -523,6 +531,27 @@ def _lazy_init_strategy(self) -> None:
                 f" found {self.strategy.__class__.__name__}."
             )
 
+    @staticmethod
```
carmocca, line 534: Is this static so it can be overridden? Move outside the class instead? Maybe even define in the `cli.py` file so the definition and parsing is together

carmocca (approved):

src/lightning_lite/cli.py
awaelchli: The division with num_processes can lead to a number smaller than 1, which gets floored to 0. max(1, x) clips the value to >=1, so OMP_NUM_THREADS is at least 1.

src/lightning_lite/connector.py
awaelchli: It belongs to this class, because only this class has custom logic in the init that corresponds to the implementation in this function.

src/lightning_lite/cli.py
awaelchli: Good point. Actually, we should probably expect the _ vs. - to be a common typo. The issue is, if we only support the num_nodes (underscore) version, and user types num-nodes (hypen) by accident, the argument does not get parsed by the cli and passed down as script args. This silent behavior, in general, is worrying me a bit tbh.

src/lightning_lite/cli.py
awaelchli: It is a slightly different meaning, so no I don't recommend it. 

Accelerator parses the devices. With many strategies, num_devices = num_processes per node. But for example for DP, that's not the case. For DP, we always have 1 process but can support N devices.

src/lightning_lite/cli.py
awaelchli: Yeah, so with the CLI we can't really support the MultiprocessingLauncher.

It should be possible to launch processes the other way for XLA/TPU too, but we don't have this launcher in Trainer. Before we add support for this, we should verify it in the console on a tpu machine. I don't think this is a priority though.

src/lightning_lite/cli.py
awaelchli: sure, we can make it a context manager that restores the env on exit

src/lightning_lite/CHANGELOG.md
```diff
@@ -10,7 +10,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
--
+### Added
```
Borda, line 13:
```suggestion
```

src/lightning_lite/cli.py
awaelchli: Any suggestion how to write the existing tests in test_cli in a way that is compatible with your suggestion? If the env variables don't survive after calling the cli, I'm not sure how to assert them.

src/lightning_lite/cli.py
awaelchli: updated and added test

src/lightning_lite/cli.py
awaelchli: I honestly don't see how. The parsing here is very specific.

src/lightning_lite/cli.py
otaj: `assert "var" not in os.environ` should do the trick, right?

awaelchli: That would be tautological if we delete the var after running the command. I want to test the opposite. See for example any test with name `test_cli_env_vars_*`.

tchaton:
@awaelchli Looks great and I really like `lightning run lite ...` approach.

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,177 @@
...
+from lightning_lite.accelerators import CPUAccelerator, CUDAAccelerator, MPSAccelerator
+from lightning_lite.utilities.device_parser import _parse_gpu_ids
+from lightning_lite.utilities.imports import _IS_WINDOWS, _TORCH_GREATER_EQUAL_1_13
+
+# torchrun in PyTorch 1.13.0 has a bug on the Windows platform and is thus not importable:
+# https://github.com/pytorch/pytorch/issues/85427
+if not (_IS_WINDOWS and _TORCH_GREATER_EQUAL_1_13):
+    import torch.distributed.run as torchrun
+else:
+    torchrun = None  # type: ignore[assignment]
```
carmocca, line 28: Can you make this a local import in `_torchrun_launch` so we don't need to check the `if` again? Also avoids setting a `None` default

src/lightning_lite/cli.py
```diff
@@ -0,0 +1,177 @@
...
+        "--nproc_per_node",
+        str(num_processes),
+        "--nnodes",
+        str(args.num_nodes),
+        "--node_rank",
+        str(args.node_rank),
+        "--master_addr",
+        args.main_address,
+        "--master_port",
+        str(args.main_port),
```
carmocca, line 160: nit: this is more readable

```suggestion
        f"--nproc_per_node={num_processes}",
        f"--nnodes={args.num_nodes}",
        f"--node_rank={args.node_rank}",
        f"--master_addr={args.main_address}",
        f"--master_port={args.main_port}",
```

src/lightning_lite/lite.py
```diff
@@ -496,3 +524,11 @@ def _validate_setup_dataloaders(dataloaders: Sequence[DataLoader]) -> None:
 
         if any(not isinstance(dl, DataLoader) for dl in dataloaders):
             raise TypeError("Only PyTorch DataLoader are currently supported in `setup_dataloaders`.")
+
+
+def _is_using_cli() -> bool:
+    return bool(int(os.environ.get("LT_CLI_USED", "0")))
```
carmocca, line 530:
```suggestion
    return bool(os.environ.get("LT_CLI_USED", False))
```

carmocca (approved):

src/lightning_lite/lite.py
awaelchli: No, this is problematic. 
`bool("0") is True`

src/lightning_lite/cli.py
awaelchli: It would make the mocking of torchrun in tests incredibly hard. Shouldn't we be able to remove this code again once pytorch fixes the issue in one of their 1.13.x release? https://github.com/pytorch/pytorch/pull/88250

awaelchli: I'm struggling to refactor all tests for the local import

carmocca: Then put the import and error into an utility fn that returns torchrun and mock this fn

src/lightning_lite/cli.py
awaelchli: I think I will leave this as future work. I'm not concerned about this right now because the expected behavior is that the CLI exits when the called program exits. It won't invoke any other programs, so it is ok to leave the environ as it is.

src/lightning_lite/cli.py
awaelchli: It eventually worked with monkeypatch. I hope it's ok now

PR #14994, opened 2022-10-04
update optimizer typing

justusschock:
## What does this PR do?

Updates structural optimizer typing to the actual list of things we expect.

### DIFF
--- a/src/lightning_lite/utilities/types.py
+++ b/src/lightning_lite/utilities/types.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterator, List, Optional, TypeVar, Union
+from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, TypeVar, Union
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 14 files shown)


src/lightning_lite/utilities/types.py
```diff
@@ -80,9 +80,19 @@ def step(self, metrics: Union[float, int, Tensor], epoch: Optional[int] = None)
...
 
 @runtime_checkable
-class Steppable(Protocol):
-    """To structurally type ``optimizer.step()``"""
+class Optimizable(Protocol):
+    """To structurally type ``optimizer``"""
+
+    param_groups: List[Dict[Any, Any]]
+    defaults: Dict[Any, Any]
+    state: Dict[Any, Any]
```
awaelchli, line 88: Does this apply to DeepspeedEngine?

tests/tests_lite/plugins/precision/test_deepspeed.py
```diff
@@ -50,15 +50,15 @@ def test_deepspeed_precision_backward():
 
 
 @RunIf(deepspeed=True)
-def test_deepspeed_engine_is_steppable():
-    """Test that the ``DeepSpeedEngine`` conforms to the Steppable API.
+def test_deepspeed_engine_is_Optimizable():
```
awaelchli, line 53:
```suggestion
def test_deepspeed_engine_is_optimizable():
```

awaelchli (approved):

src/lightning_lite/utilities/types.py
```diff
@@ -80,9 +80,19 @@ def step(self, metrics: Union[float, int, Tensor], epoch: Optional[int] = None)
 
 
 @runtime_checkable
-class Steppable(Protocol):
```
carmocca, line 83: I suggest we keep Steppable. We can use inheritance for protocols

tests/tests_lite/plugins/precision/test_deepspeed.py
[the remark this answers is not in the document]
justusschock: done

src/lightning_lite/utilities/types.py
justusschock: done

src/lightning_lite/utilities/types.py
justusschock: reverted for deepspeed

carmocca (approved):

PR #14995, opened 2022-10-04
Add Database Component

tchaton:
## What does this PR do?

This component provides a simple way to add a SQLite database to your app using the SQLModel dependency.

This component was created from the work done in the Lightning [HPO repo](https://github.com/Lightning-AI/lightning-hpo) and it was re-used in the Echo App.

```python
from sqlmodel import Field, SQLModel

from lightning import LightningApp, LightningFlow
from lightning_app.components.database import Database, DatabaseClient


class CounterModel(SQLModel, table=True):
    __table_args__ = {"extend_existing": True}

    id: int = Field(default=None, primary_key=True)
    count: int


class Flow(LightningFlow):
    def __init__(self):
        super().__init__()
        self.db = Database(models=[CounterModel])
        self._client = None
        self.counter = 0

    def run(self):
        self.db.run()

        if not self.db.alive():
            return

        if self.counter == 0:
            self._client = DatabaseClient(model=CounterModel, db_url=self.db.url)
            self._client.reset_database()

        assert self._client

        rows = self._client.select_all()

        print(f"{self.counter}: {rows}")

        if not rows:
            self._client.insert(CounterModel(count=0))
        else:
            row: CounterModel = rows[0]
            row.count += 1
            self._client.update(row)

        if self.counter >= 100:
            row: CounterModel = rows[0]
            self._client.delete(row)
            self._client.delete_database()
            self._exit()

        self.counter += 1


app = LightningApp(Flow())
```

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/components/database/server.py
```diff
@@ -0,0 +1,196 @@
...
+logger = Logger(__name__)
+engine = None
+
+
+class DatabaseType:
+    FLOW = "flow"
+    WORK = "work"
+
+
+class Database(LightningFlow):
```
manskx, line 35: Should this component be part of lightning or as a separate component to be used in the lightning gallery? I would prefer to have it separately similar to the Redis component or a postgress in the future.

tchaton: I would prefer to keep it in the framework right now. I don't think the 1 package 1 component is really scalable.

tchaton: Furthermore, the Echo App copied the code over. I want to encourage folks to use core components and build upon them.

manskx: Is it something that has been discussed and agreed to put similar components inside the framework ?

The component doesn't work out of the box, it needs to install its dependencies.

src/lightning_app/components/database/server.py
```diff
@@ -0,0 +1,138 @@
...
+        self.debug = debug
+        self._models = models if isinstance(models, list) else [models]
+        self._client = None
+
+    def run(self) -> None:
+        app = FastAPI()
+
+        create_database(self.db_filename, self._models, self.debug)
+        models = {m.__name__: m for m in self._models}
+        app.post("/select_all/")(partial(general_select_all, models=models))
```
lantiga, line 113: This looks great, the fundamental hold-up I have is that we are exposing raw database methods to the outside world. This may be ok for a trusted environment, but it's not ok for an untrusted one.
We should restrict the origin of the requests to allow requests from the app itself, it should be easily doable.

lantiga (changes requested):

src/lightning_app/components/database/server.py
tchaton: Yes, we chatted with @lantiga in Bergamo about it.
Overall, core abstractions should live within the framework as it pushes for re-utilisation.

Good examples are ServeGradio and TracerPythonScript. They are being re-used in lot of codebases.

src/lightning_app/components/database/server.py
tchaton: Add a security mechanism with a token. There is a period of time where the token can be accessed from the AppState until the work remove it from its state.

src/lightning_app/components/database/server.py
```diff
@@ -103,19 +95,20 @@ def run(self):
         self.db_filename = db_filename
         self.debug = debug
         self._models = models if isinstance(models, list) else [models]
-        self._client = None
+        self.token = token
```
lantiga, line 98: this should be `self._token`, so it's not exposed to the global state

src/lightning_app/components/database/utilities.py
```diff
@@ -124,78 +123,106 @@ def get_primary_key(model_type: Type[SQLModel]) -> str:
...
-        session.commit()
+class SelectAll:
+    def __init__(self, models, token):
+        print(models, token)
+        self.models = models
+        self.token = token
+
+    def __call__(self, data: Dict):
+        if self.token and data["token"] != self.token:
+            raise Exception("The request doesn't have the correct token !")
```
lantiga, line 160: Should we raise a standard `HTTPException` from FastAPI (returning 401, Unauthorized) instead?
In any case, I would reword as "Unauthorized request".

lantiga (changes requested):
Minor change (take a state variable to private)

src/lightning_app/components/database/server.py
tchaton: Done !

lantiga (approved):
LGTM!

src/lightning_app/components/database/client.py
```diff
@@ -0,0 +1,77 @@
...
+class DatabaseClient:
+    def __init__(self, db_url: str, token: str, model: Optional[T] = None) -> None:
+        self.db_url = db_url
+        self.model = model
+        self.token = token
+        self._session = None
+
+    def select_all(self, model: Optional[Type[T]] = None) -> List[T]:
+        cls = model if model else self.model
+        assert cls
```
justusschock, line 44: are these asserts  necessary? Generally asserts within the code should be avoided if possible

src/lightning_app/components/database/server.py
```diff
@@ -0,0 +1,156 @@
...
+
+                    self.counter += 1
+
+            app = LightningApp(Flow())
+        """
+        super().__init__(parallel=True, cloud_build_config=BuildConfig(["sqlmodel"]))
+        self.db_filename = db_filename
+        self.debug = debug
+        self._models = models if isinstance(models, list) else [models]
+        self.drive = Drive("lit://database")
```
justusschock, line 117: should this be hardcoded? What if there are multiple databases?

justusschock (approved):

src/lightning_app/components/database/client.py
tchaton: Yes, better if no model has been provided in either init nor select.

src/lightning_app/components/database/server.py
tchaton: Good catch! I added support for them to be stored in an isolated way. Yes, it is fine for the name to be hardcoded as it is only a folder name.

manskx (approved):

src/lightning_app/components/database/server.py
manskx: awesome, thanks !

Borda (commented):
Can we add documentation?

tchaton:
> Can we add documentation?

@Borda Done !

src/lightning_app/components/database/server.py
```diff
@@ -0,0 +1,159 @@
...
+        use_localhost = "LIGHTNING_APP_STATE_URL" not in os.environ
+        if use_localhost:
+            return self.url
+        if self.internal_ip != "":
+            return f"http://{self.internal_ip}:{self.port}"
+        return self.internal_ip
+
+    def on_exit(self):
+        self.drive.put(self.db_filename)
+        print("Stored the database to the Drive.")
```
awaelchli, line 159: Is this the "new" way of persisting files? 
So far the designed approach was to use Path() on the attribute, and Lightning would persist these files automatically.

src/lightning_app/components/database/utilities.py
```diff
@@ -0,0 +1,248 @@
...
+    from sqlalchemy.inspection import inspect as sqlalchemy_inspect
+    from sqlmodel import JSON, select, Session, SQLModel, TypeDecorator
+
+logger = Logger(__name__)
+engine = None
+
+T = TypeVar("T")
+
+
+# Taken from https://github.com/tiangolo/sqlmodel/issues/63#issuecomment-1081555082
```
awaelchli, line 24: I think it is great to take inspiration from there for the core logic. However, by copying 1:1 we are inheriting very bad practices. The code should be simplified to be more readable and testable.

src/lightning_app/components/database/utilities.py
```diff
@@ -0,0 +1,248 @@
...
+def get_primary_key(model_type: Type[SQLModel]) -> str:
+    primary_keys = sqlalchemy_inspect(model_type).primary_key
+
+    if len(primary_keys) != 1:
+        raise ValueError(f"The model {model_type.__name__} should have a single primary key field.")
+
+    return primary_keys[0].name
+
+
+class GeneralModel(BaseModel):
```
awaelchli, line 124: These classes are public apis yet have no documentation. Let's add it? Or add minimal documentation and make them protected/internal APIs?

src/lightning_app/utilities/serializable.py
```diff
@@ -0,0 +1,9 @@
+from typing import Any, Dict
+
+from typing_extensions import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class Hashable(Protocol):
+    def to_dict(self) -> Dict[str, Any]:
+        ...
```
awaelchli, line 9: in the other libraries we organize these under utilities/types.py. We should probably move this there

awaelchli (approved):

src/lightning_app/components/database/server.py
```diff
@@ -0,0 +1,159 @@
...
+from lightning_app.storage import Drive
+from lightning_app.utilities.app_helpers import Logger
+from lightning_app.utilities.imports import _is_sqlmodel_available
+
+if _is_sqlmodel_available():
+    from sqlmodel import SQLModel
+
+
+logger = Logger(__name__)
+engine = None
```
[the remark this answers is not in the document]
tchaton: Thanks !

src/lightning_app/components/database/server.py
```diff
@@ -0,0 +1,159 @@
...
+        about your application.
+
+        The provided models are SQLModel tables
+
+        Arguments:
+            models: A SQLModel or a list of SQLModels table to be added to the database.
+            db_filename: The name of the SQLite database.
+            debug: Whether to run the database in debug mode.
+            mode: Whether the database should be running within the flow or dedicated work.
+            token: Token used to protect the database access. Ensure you don't expose it through the App State.
```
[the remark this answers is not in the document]
tchaton: Removed !

src/lightning_app/components/database/server.py
tchaton: Yes, Path would have been fine but I didn't implement the restore mechanism. So I decided to give it a go with Drive instead.

src/lightning_app/components/database/utilities.py
tchaton: Moved to private.

src/lightning_app/components/database/utilities.py
tchaton: Yes, unfortunately. This code is slightly magic to me too. I roughly understand how it works and I tested it works in practice.

PR #15001, opened 2022-10-04
[Lite] precision_plugin -> precision

carmocca:
## What does this PR do?

Rename argument as Lite dropped the `_plugin` suffix

None


justusschock:
@carmocca I feel this could easily be confused with the precision flag in the main API classes and people trying to pass precision=16 in there.

src/lightning_lite/connector.py
```diff
@@ -63,8 +63,7 @@
 
 
 class _Connector:
-    """The Connector parses several Lite arguments and instantiates the Strategy including other components such as
-    the Accelerator and Precision plugins.
+    """The Connector parses several Lite arguments and instantiates the Strategy including it's owned components.
```
awaelchli, line 66:
```suggestion
    """The Connector parses several Lite arguments and instantiates the Strategy including its owned components.
```
What was wrong about the previous description?
To be clear, the precision class still formally remains a plugin, the same way the cluster environment is a plugin.

src/lightning_lite/connector.py
```diff
@@ -113,8 +112,8 @@ def __init__(
         # For devices: Assign gpus, ipus, etc. to the accelerator flag and devices flag
         self._strategy_flag: Optional[Union[Strategy, str]] = None
         self._accelerator_flag: Optional[Union[Accelerator, str]] = None
-        self._precision_flag: Optional[_PRECISION_INPUT] = None
-        self._precision_plugin_flag: Optional[Precision] = None
+        self._precision_input: Optional[_PRECISION_INPUT] = None
+        self._precision_instance: Optional[Precision] = None
```
awaelchli, line 116: I wouldn't diverge from the accelerator connector class unless absolutely necessary. When we need to make changes, or fixes, we likely need to update both classes (today), and inconsistencies will make it only more difficult.

src/pytorch_lightning/lite/lite.py
```diff
@@ -268,20 +267,20 @@ def _to_lite_strategy(strategy: PLStrategy) -> LiteStrategy:
...
             checkpoint_io=strategy.checkpoint_io,
-            precision_plugin=_to_lite_precision_plugin(strategy.precision_plugin),
+            precision=_to_lite_precision(strategy.precision_plugin),
         )
 
 
-def _to_lite_precision_plugin(plugin: Optional[PLPrecisionPlugin]) -> LitePrecision:
-    """Re-instantiates a PL-PrecisionPlugin as the corresponding Lite-Precision plugin."""
+def _to_lite_precision(plugin: Optional[PLPrecisionPlugin]) -> LitePrecision:
+    """Re-instantiates a PL-PrecisionPlugin as the corresponding Lite precision."""
```
awaelchli, line 283: NIT: In plain text, it is still totally ok to refer to it as a plugin (see my comment above)

awaelchli (approved):

src/lightning_lite/connector.py
carmocca: I always prefer avoiding describing the code in docstrings, as they often go stale otherwise.

src/lightning_lite/connector.py
carmocca: I think they will inevitably diverge if we are not going to port API changes to the Trainer and its flags that we will do to Lite. This will be a challenge in the coming months, but we don't know what will be the end result. For this reason, I wouldn't block changes to Lite just because we don't know what will happen in PL.

carmocca:
> I feel this could easily be confused with the precision flag in the main API classes and people trying to pass precision=16 in there.

True, but if we do #9956 then we would probably rename the inner one as `dtype` or `value` or `argument`

justusschock (approved):

src/lightning_lite/connector.py
awaelchli: I would not hold back on anything that simplifies or does meaningful refactors. But simple name changes on internal flags just makes it harder when it doesn't have to be.

PR #11531, opened 2022-01-18
Replace custom `AllGather` implementation

carmocca:
## What does this PR do?

Remove custom AllGatherGrad implementation in favor of torch.distributed's

Fixes #10445
Continues #10504

### Does your PR introduce any breaking changes? If yes, please list them.

None

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)**? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

cc @borda @awaelchli @carmocca @justusschock @rohitgr7 @akihironitta @tchaton

PR #15003, opened 2022-10-05
Make gradients available for all_gather on TPU

stekiri:
## What does this PR do?

Support autograd for all_gather on TPU using [`torch_xla.core.functions.all_gather`](https://pytorch.org/xla/release/1.12/index.html#torch_xla.core.functions.all_gather)

Fixes https://github.com/Lightning-AI/lightning/issues/6295

None

## Before submitting

## PR review

## Did you have fun?


src/lightning_lite/strategies/xla.py
```diff
@@ -155,20 +155,26 @@ def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
...
+            sync_grads: flag that allows users to synchronize gradients for the all_gather operation
         Return:
             A tensor of shape (world_size, batch, ...)
         """
         if isinstance(tensor, Tensor) and tensor.dim() == 0:
             tensor = tensor.unsqueeze(0)
-        import torch_xla.core.xla_model as xm
 
-        return xm.all_gather(tensor)
+        if sync_grads:
```
carmocca, line 170: Is there any disavantage to always using `xf.all_gather`? If no, then let's just switch the function

stekiri: I assume there's a certain communication cost that comes with returning the gradients, even though I cannot say if this actually a problem when running with TPUs.

However, if someone calls the method with `sync_grad=False`, he probably expects that the gradients are not returned, but `xf.all_gather` would still return them. That's why I have the feeling that a switch between the two functions could be better, I saw [this code for the DDP strategy](https://github.com/Lightning-AI/lightning/blob/6f16e46bdb397bf1898e4d7d9050492924177abd/src/lightning_lite/utilities/distributed.py#L201) which is doing it similarly.

carmocca: I expect both functions to be equal when run under a `no_grad` context manager. For example, `torch.distributed`does not offer different implementations whether grads are supported or not.

In fact, looking at the implementation: https://github.com/pytorch/xla/pull/2086/files, it seems like they just copied PyTorch's implementation but using xla's collective ops.

Note that the linked code for `AllGatherGrad` is outdated: #11531

stekiri: Thanks for your explanation, seems like I didn't really understand your initial comment 😅 

Just to make sure that I got it right this time, are you suggesting to do it like in the following snippet? 
```
with nullcontext() if sync_grads else torch.no_grad():
    return xf.all_gather(tensor)
```

carmocca: No, just `xf.all_gather(tensor)`, without the context manager. That will be set elsewhere

stekiri: Sorry, I still don't get it, what do you mean exactly when you say it "will be set elsewhere"?

Do you suggest that if a pytorch-lightning user does not want gradients, he should call it like this:
```
with torch.no_grad():
    self.all_gather(some_tensor)
```
If that is what you meant, I agree it's more straight-forward. The only small problem is that the `sync_grads` parameter won't be taken into account (which may surprise the user). Even though, I have to say that I don't think that it is very intuitive that the default of `sync_grads` is `False` because I assume users prefer to have the gradients (and with your proposal, it would be very easy to achieve.

Having said that I am totally fine to simply use `xf.all_gather(tensor)`, I could update the docstring that explains that the `sync_grads` param is not taken into account to make users aware that the behavior is a bit different from the DDP strategy.

carmocca: > he should call it like this:

The trainer sets the context manager already: https://github.com/Lightning-AI/lightning/blob/24c26f7db298e7fd83e4a37459a46d9043763b0e/src/pytorch_lightning/trainer/trainer.py#L2225-L2238


> I could update the docstring that explains that the sync_grads param is not taken into account to make users aware that the behavior is a bit different from the DDP strategy.

Yes, let's do that!

The parameter is a bit of a legacy one so it's fine to have it not do anything.

tests/tests_pytorch/accelerators/test_tpu.py
```diff
@@ -323,3 +323,63 @@ def test_trainer_config_device_ids(devices, expected_device_ids):
     trainer = Trainer(accelerator="tpu", devices=devices)
     assert trainer.device_ids == expected_device_ids
     assert trainer.num_devices == len(expected_device_ids)
+
+
+@RunIf(tpu=True, standalone=True)
+@mock.patch.dict(os.environ, os.environ.copy(), clear=True)
+def test_all_gather_sync_grads(tmpdir):
```
awaelchli, line 330: I recommend combining the two tests into one. Spawning processes is expensive, and here I see too little justification for it to be two separate ones. 

We will also need unit tests for the xla strategy in lightning_lite. I'm happy to help adding these if desired. 
(I suggest we use the xla_launch utility from tests/tests_lite/strategies/test_xla.py)

src/lightning_lite/strategies/xla.py
```diff
@@ -155,20 +155,26 @@ def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
...
+            import torch_xla.core.functions as xf
+
+            return xf.all_gather(tensor)
+        else:
+            import torch_xla.core.xla_model as xm
+
+            return xm.all_gather(tensor)
```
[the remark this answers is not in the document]
stekiri: Also fine by me :) Do you think that this solution should be preferred over the proposal of @carmocca to only use `xf.all_gather(tensor)` in both cases?

tests/tests_pytorch/accelerators/test_tpu.py
stekiri: Alright, I'll combine them and have a look at the test for lightning_lite. I'll ping you if I need a hand with it.

tests/tests_lite/strategies/test_xla.py
```diff
@@ -113,3 +114,17 @@ def test_xla_validate_unsupported_iterable_dataloaders(_, dataloader, monkeypatc
 
     with pytest.raises(TypeError, match="TPUs do not currently support"):
         XLAStrategy().process_dataloader(dataloader)
+
+
+def tpu_all_gather_fn(strategy):
+    for sync_grads in [True, False]:
+        tensor = torch.tensor(1, device=strategy.root_device)
+        result = strategy.all_gather(tensor, sync_grads=sync_grads)
+        assert result.sum() == 8
```
awaelchli, line 123: How about we do a backward and show gradients gets synced?

```suggestion
    tensor = torch.tensor(1, device=strategy.root_device)
    result = strategy.all_gather(tensor, sync_grads=False)
    assert result.sum() == 8
    
    tensor = torch.tensor(1.0, device=strategy.root_device , requires_grad=True)
    result = strategy.all_gather(tensor, sync_grads=False)
    result.sum().backward()
    assert torch.equal(tensor.grad, torch.tensor(1.0))
```

tests/tests_pytorch/accelerators/test_tpu.py
```diff
@@ -323,3 +323,39 @@ def test_trainer_config_device_ids(devices, expected_device_ids):
     trainer = Trainer(accelerator="tpu", devices=devices)
     assert trainer.device_ids == expected_device_ids
     assert trainer.num_devices == len(expected_device_ids)
+
+
+@RunIf(tpu=True, standalone=True)
```
awaelchli, line 328:
```suggestion
@RunIf(tpu=True)
```
with devices=8, the standalone mark is not necessary.

tests/tests_lite/strategies/test_xla.py
stekiri: Good idea, I totally agree, this test is a lot more meaningful.

justusschock (approved):

tests/tests_lite/strategies/test_xla.py
stekiri: The TPU tests are failing, it was already the case before the last changes. Do you know how I can access the full output logs of the test? I only see:

```
Waiting for job to finish: ..........................................................Failed:1
Done waiting. Job status code: 1
Error: Process completed with exit code 1.
```

You got a suggestion how to run the TPU tests before committing changes? Maybe you already got a Colab notebook ready somewhere? 😁

carmocca (changes requested):
https://github.com/Lightning-AI/lightning/pull/15349 needs to be merged first

stekiri:
@carmocca, I ran my new tests successfully on Colab using a TPU runtime. However, the `test-on-tpus` run in this PR keeps failing, I seem to hit the [100 minutes timeout](https://github.com/Lightning-AI/lightning/blob/master/dockers/tpu-tests/tpu_test_cases.jsonnet#L11) and there seem to be a problem with [the logs](https://github.com/Lightning-AI/lightning/actions/runs/3377042270/jobs/5605497932#step:10:54): `error: You must be logged in to the server (Unauthorized)`.

tests/tests_pytorch/accelerators/test_tpu.py
```diff
@@ -323,3 +323,44 @@ def test_trainer_config_device_ids(devices, expected_device_ids):
     trainer = Trainer(accelerator="tpu", devices=devices)
     assert trainer.device_ids == expected_device_ids
     assert trainer.num_devices == len(expected_device_ids)
+
+
+@RunIf(tpu=True)
+@mock.patch.dict(os.environ, os.environ.copy(), clear=True)
+def test_all_gather(tmpdir):
```
carmocca, line 330: There's no value in adding this test, as this is all covered by the one you added in the other file

carmocca (approved):
You can ignore the "Unauthorized" error. If the job timed out it's because there's no hardware availability at this moment.

tests/tests_pytorch/accelerators/test_tpu.py
stekiri: This test is for the code modifs in the `TPUSpawnStrategy` of `pytorch_lightning` whereas the other test is for the code modifs in the `XLAStrategy` of `lightning_lite`. It concerns different classes, therefore, there should be a test for both execution paths even though the test code is the same. Do you want to imply that internally only one of those strategies are used?

It looks like the `TPUSpawnStrategy` of `pytorch_lightning` and the `XLAStrategy` of `lightning_lite` are very much the same. As I don't know much about the internal usages, I'm not well placed to say whether they can be entirely merged. Surely, efforts could be made to avoid redundancy in these two classes. However, this may best be done in a separate ticket to avoid polluting this PR.

awaelchli: I generally agree with @carmocca. 
The test in the other file for lite is great, but even in that one, one could argue that it is going too far in that it is testing the correctness of `xf.all_gather`. The real test for us we care about is whether the "wrapper" method calls the underlying function correctly (with the right arguments).

Regarding the two classes:
The current state is that the two classes are independent, yes. The idea however is that in the future, the tpu strategy in pytorch_lightning will use most of the code in the other strategy in lightning_lite. 

I am ok with adding these two "slow" tests. But remember, launching processes is extremely expensive. 
If more operations get the "grad" support, we should extend this test instead of launching new ones.

awaelchli:
@stekiri We haven't forgotten this PR. Some more patience is required, sorry about this. We have some issue getting the TPU CI to work on GitHub here #15788.

Issue #14986, opened 2022-10-03
Single Node distributed training with multiple GPUs works but fails on multi-node


martinez-zacharya:
### Bug description

I'm able to train on multiple GPUs on one node, but it fails when trying to do multi-node. I'm using SLURM to submit my jobs. Thank you ahead of time!

### How to reproduce the bug

```python
import os
import torch
from torch.nn import functional as F
from torch.utils.data import DataLoader
from torchvision.datasets import MNIST
import torchvision.transforms as transforms
from pytorch_lightning import Trainer
import pytorch_lightning as pl


class CoolModel(pl.LightningModule):

    def __init__(self):
        super(CoolModel, self).__init__()
        self.l1 = torch.nn.Linear(28 * 28, 10)

    def forward(self, x):
        return torch.relu(self.l1(x.view(x.size(0), -1)))

    def my_loss(self, y_hat, y):
        return F.cross_entropy(y_hat, y)

    def training_step(self, batch, batch_nb):
        x, y = batch
        y_hat = self.forward(x)
        return {'loss': self.my_loss(y_hat, y)}

    def validation_step(self, batch, batch_nb):
        x, y = batch
        y_hat = self.forward(x)
        return {'val_loss': self.my_loss(y_hat, y)}

    def validation_end(self, outputs):
        avg_loss = torch.stack([x['val_loss'] for x in outputs]).mean()
        return {'avg_val_loss': avg_loss}

    def configure_optimizers(self):
        return [torch.optim.Adam(self.parameters(), lr=0.02)]

    def train_dataloader(self):
        return DataLoader(MNIST(os.getcwd(), train=True, download=True, transform=transforms.ToTensor()), batch_size=32)

    def val_dataloader(self):
        return DataLoader(MNIST(os.getcwd(), train=True, download=True, transform=transforms.ToTensor()), batch_size=32)

    def test_dataloader(self):
        return DataLoader(MNIST(os.getcwd(), train=True, download=True, transform=transforms.ToTensor()), batch_size=32)




model = CoolModel()


trainer = Trainer(max_epochs=1, gpus=4, num_nodes=2, accelerator='gpu', strategy='fsdp')

trainer.fit(model)
```

```shellscript
#!/bin/bash
#SBATCH --time=1:00:00   # walltime
#SBATCH --ntasks=2   # number of processor cores (i.e. tasks)
#SBATCH --ntasks-per-node=4
#SBATCH --nodes=2 # number of nodes
[...]

Issue #14078, opened 2022-08-07
RFC: Remove `num_nodes` Trainer argument and infer world size from cluster environment directly


awaelchli:
## 🚀 Feature

Remove the redundant `num_nodes` Trainer argument. Knowing the number of nodes is not required, and the world size is provided by the cluster environment anyway.  

### Motivation

Users have always struggled getting multi-node training to work because there exists many points of failures in the configurations. One of them is setting the number of devices and nodes, where in Lightning you are required to set `devices=n` and `num_nodes=m` to match the `world_size=n*m` set by the cluster or launcher (torchrun, slurm, etc.). When num_nodes is not set correctly, it leads to too few processes joining the group and thus the program gets stuck. 

We periodically receive issues and questions where it becomes clear users forget to set this values correctly (#13804, #14022, #10098, 
#8707, #7429,  #6206). But technically, Lightning never requires to know the number of nodes, only the world size. This value can be read from the ClusterEnvironment directly. 

The conclusion: `num_nodes` in the Trainer is redundant. 

Another benefit of not having to specify the number of nodes is that the cluster can dynamically allocate nodes and devices based on available resources. In the end, a user who wants to train on 8 GPUs does not care if it is over 2 nodes of 4 GPUs each or over 4 nodes with 2 GPUs each. 

Historically, `num_nodes` was introduced before the ClusterEnvironment abstraction became the primary way of collecting information about the cluster. 

### Pitch

Deprecate the redundant `num_nodes` Trainer argument.

Pros:
- One less thing for users to get wrong
- One less step to explain in the docs
- Code simplifies
- It is no longer required to know the number of nodes ahead of time

Cons:
- None currently known

### Alternatives

Keep it, but validate num_nodes setting against cluster environment (#10107)
### Additional context

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/Lightning-AI/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/Lightning-AI/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

[...]

PR #15011, opened 2022-10-05
Validate SRUN variables when launching in SLURM

### LINKED ISSUE #10150: Validate environment variables in SLURMEnvironment and warn user about incompatible settings
## 🚀 Feature

Check slurm environment settings and print warnings if needed.

### Motivation

If SLURM srun variables are set incorrectly, the processes can hang and the user will not know why.

Examples:
- https://github.com/PyTorchLightning/pytorch-lightning/issues/4612#issuecomment-949937617
- #15141

### Pitch

In the SLURM cluster environment, check for example:

```py
if os.environ.get("SLURM_CPUS_PER_TASK") > 1:
    warn(
        "You asked SLURM for multiple CPUs but we are not sure your machine has multiple CPUs. "
        "If your process hangs, consider setting --cpus-per-task=1"
    )

```
### Alternatives

Do nothing. Users will keep submitting issues :) 

### Additional context

<sub>

</sub>

cc @borda @awaelchli

awaelchli:
## What does this PR do?

Addresses suggestion in #14986, frequent user errors that result in their SLURM jobs hanging. By adding warnings and errors, we can prevent these frequent questions and give feedback directly at runtime.

Fixes #10150

No intended breaking changes:
- Lightning SLURM users were always meant to use the --ntasks-per-node, otherwise their processes would hang. Hence, the error is appropriate.
- The warning about srun not being used is very unlikely going to be a false positive, because the `srun` program is only installed on SLURM clusters. However, we won't make it an error just in case there is another program called `srun` on your system unrelated to SLURM.

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @carmocca @justusschock


src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +144,19 @@ def resolve_root_node_address(nodes: str) -> str:
...
+        Right now, we only check for the most commond user errors. See `the srun docs
+        <https://slurm.schedmd.com/srun.html>`_ for a complete list of supported srun variables.
+        """
+
+        ntasks = int(os.environ.get("SLURM_NTASKS", "1"))
+        if ntasks > 1:
+            raise rank_zero_warn(
+                f"You set `--ntasks={ntasks}` in your SLURM bash script, but this variable is not supported. Please use"
+                f" `--ntasks-per-node={ntasks}` instead."
+            )
```
awaelchli, line 162: @Queuecumber Sorry for pinging you, but as a slurm user you might know this. 
1. Should this be an error? For now I made it a warning as I am not 100% sure if I break someone's code.
2. Would it be useful to also provide a warning/error when SLURM_CPUS_PER_TASK is set, with an error similar as in #10150?

Queuecumber: I won't claim any special knowledge of slurm outside whatever I need to run my experiments, but my understanding is that `ntasks` and `ntasks-per-node` can both be set and then `ntasks-per-node` functions as a maximum. Why you would want to do that and frankly what that even means I don't know but technically it's "allowed". So in that case I suppose a loud warning is best so that the one person who uses both those options and lightning doesn't have their code crash. 

That said, it may not be the best idea to commit to slurm validation. 

I always viewed the slurm multiprocessing detection/checkpointing/requeueing support in lightning as nice-haves with a limited scope and I think that fits nicely with the rest of what lightning does in terms of making your life easier without being too invasive. For example, lightning isn't writing the sbatch file and submitting the job for you; that would be out of scope.

It may be worth discussing if this kind of validation, too, is out of scope. I won't say that I think it definitely is but it might be worth discussing it internally. Validating all the permutations of slurm options and which ones conflict and how they change based on context is extremely complex and there are other options besides these that people may want validated. The other way to do it is to just go all-in and implement something like submitit which creates the sbatch file based on the trainer config, and that way you have complete control of what options are being used so you wouldn't need any validation logic.

Just something to think about

awaelchli: Greatly appreciate the quick response!

>  but my understanding is that ntasks and ntasks-per-node can both be set and then ntasks-per-node functions as a maximum

I would say yes, that's also my understanding that this is allowed in general. But I don't see that being useful or compatible with Lightning Trainer at the moment. Right now, there is no elasticity and we require each node to run the same number of processes. I don't see that changing anytime soon. Since that number set there must also be exactly the same as in the Trainer(devices=x), I don't see how setting both `ntasks-per-node` and `ntasks` to different values is valid. An example of this mistake can be seen here: #14986, and similar questions have been asked in the past and on slack. 

> Validating all the permutations of slurm options and which ones conflict and how they change based on context is extremely complex and there are other options besides these that people may want validated. 

Yeah, I don't think that's feasible. That's why I wrote in the comment in the code that we only intend to catch the most frequent user mistakes. 

> The other way to do it is to just go all-in and implement something like submitit which creates the sbatch file based on the trainer config, and that way you have complete control of what options are being used so you wouldn't need any validation logic.

Out of scope of Lightning I think. I would rather invest in a smooth interaction with submitit and Lightning.

Queuecumber: In that case I think crashing the job does make sense

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -89,6 +96,7 @@ def detect() -> bool:
         This will then avoid the detection of ``SLURMEnvironment`` and another environment can be detected
         automatically.
         """
+        SLURMEnvironment._validate_srun_used()
```
justusschock, line 99: isn't it sufficient to only run this on the environment's init? Why would you want to run it also on detection? If this is used in the end, it will run in init anyways and otherwise it doesn't make sense to validate IMO

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +149,36 @@ def resolve_root_node_address(nodes: str) -> str:
...
+    @staticmethod
+    def _validate_srun_used() -> None:
+        """Checks if the `srun` command is available and used.
+
+        Parallel jobs (multi-GPU, multi-node) in SLURM are launched by prepending `srun` in front of the Python command.
+        Not doing so will result in processes hanging, which is a frequent user error. Lightning will emit a warning if
+        `srun` is found but not used.
+        """
+        srun_exists = subprocess.call(["which", "srun"]) == 0
+        srun_used = "SLURM_NTASKS" in os.environ
```
justusschock, line 162:
```suggestion
        srun_used = SLURMEnvironment.detect()
```

when not calling in `detect` (see my comment above), we could reuse this in case the implementation might change :)

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +149,36 @@ def resolve_root_node_address(nodes: str) -> str:
...
+        hint = " ".join(["srun", os.path.basename(sys.executable), *sys.argv])[:64]
```
justusschock, line 163: why do we limit this to 64 chars?

justusschock (approved):

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: No, I think not. That would render the check useless. The SLURM environment variables are only properly set if srun is used. The dependencies are:
srun used -> environment variables set -> detect() returns True

So, when in the AcceleratorConnector we do 

```
for env in environments:
    env.detect()
```

we want that detect() checks if srun is available but not used. If it is available and not used, a different environment gets selected because SLURMEnvironment.detect() = False and the process hangs, but the warning gets emitted and the user can see it. 

On the other hand, we also call the check in init in case user explicitly instantiates and passes the plugin.

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: This would result in infinite recursion, since we need to check it in detect() itself too.

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: In case the python command is long (maybe many hyperparameters) we truncate it so the warning message remains readable. 64 is an arbitrary choice and the author likes powers of 2.

src/lightning_lite/plugins/environments/slurm_environment.py
carmocca: srun_used could become an utility fn in this file to reuse the implementation, where  both detect and this one call it.

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +149,36 @@ def resolve_root_node_address(nodes: str) -> str:
         nodes = re.sub(r"\[(.*?)[,-].*\]", "\\1", nodes)  # Take the first node of every node range
         nodes = re.sub(r"\[(.*?)\]", "\\1", nodes)  # handle special case where node range is single number
         return nodes.split(" ")[0].split(",")[0]
+
+    @staticmethod
```
carmocca, line 153: I suggest making this a utility function in this file. Unless you have a specific reason to make it static.

src/lightning_lite/plugins/environments/slurm_environment.py
carmocca, line 163: The hint should only be constructed if the `if` below is True

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +149,36 @@ def resolve_root_node_address(nodes: str) -> str:
...
+        """Checks if the `srun` command is available and used.
+
+        Parallel jobs (multi-GPU, multi-node) in SLURM are launched by prepending `srun` in front of the Python command.
+        Not doing so will result in processes hanging, which is a frequent user error. Lightning will emit a warning if
+        `srun` is found but not used.
+        """
+        srun_exists = subprocess.call(["which", "srun"]) == 0
+        srun_used = "SLURM_NTASKS" in os.environ
+        hint = " ".join(["srun", os.path.basename(sys.executable), *sys.argv])[:64]
+        if srun_exists and not srun_used:
```
carmocca, line 164: Should we also check if `srun` is in sys.argv? Or too brittle?

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +149,36 @@ def resolve_root_node_address(nodes: str) -> str:
...
+            rank_zero_warn(
+                "The `srun` command is available on your system but is not used. HINT: If your intention is to run"
+                f" Lightning on SLURM, prepend your python command with `srun` like so: {hint} ...",
+                category=PossibleUserWarning,
+            )
+
+    @staticmethod
```
carmocca, line 171: Same suggestion

src/lightning_lite/plugins/environments/slurm_environment.py
```diff
@@ -141,3 +149,36 @@ def resolve_root_node_address(nodes: str) -> str:
...
+    def _validate_srun_variables() -> None:
+        """Checks for conflicting or incorrectly set variables set through `srun` and raises a useful error
+        message.
+
+        Right now, we only check for the most common user errors. See `the srun docs
+        <https://slurm.schedmd.com/srun.html>`_ for a complete list of supported srun variables.
+        """
+        ntasks = int(os.environ.get("SLURM_NTASKS", "1"))
+        if ntasks > 1 and "SLURM_NTASKS_PER_NODE" not in os.environ:
```
carmocca, line 180: Why is ntasks okay if `SLURM_NTASKS_PER_NODE` is also set?

The error isn't clear, it says it's not supported but technically it is if both are set? The tests modified below have both set

carmocca (approved):

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: The reason is that this is not really a utility function that should be used anywhere else. It only makes sense in the context of this SLURM class and therefore is a staticmethod.

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: srun won't be in argv as argv only stores what is passed to python afaik. Anything before the python command won't show up.

src/lightning_lite/plugins/environments/slurm_environment.py
carmocca: This is mostly a stylistic choice, but so far we've been choosing module (aka utility) functions over static methods unless inheritance or a shared interface is desired. Not a strong suggestion though

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: NTASKS is equivalent to WORLD_SIZE and should be there always. The key here is that the user instructs SLURM that it should start a fixed number of processes per node, and not distribute the processes arbitrarily. It is important that Lighting has the same number of processes on each node (the same limitation applies to torchrun/torchelastic). This is what we are requiring here.

In the future this might change (#14078) but currently it is a requirement.

src/lightning_lite/plugins/environments/slurm_environment.py
awaelchli: I did this, although I find it adds an indirection that makes the code less readable.

PR #15016, opened 2022-10-06
Introduce base collective and main subclasses

carmocca:
## What does this PR do?

Part 1 of https://github.com/Lightning-AI/lightning/pull/14996

For now, the classes have not been integrated. Also not exposed from the `lite.plugins`

None

cc @borda @awaelchli @rohitgr7 @akihironitta @carmocca @justusschock


src/lightning_lite/plugins/collectives/__init__.py
```diff
@@ -0,0 +1,9 @@
+from lightning_lite.plugins.collectives.collective import Collective
+from lightning_lite.plugins.collectives.single_device_collective import SingleDeviceCollective
+from lightning_lite.plugins.collectives.torch_collective import TorchCollective
+
+__all__ = [
+    "Collective",
+    "TorchCollective",
+    "SingleDeviceCollective",
+]
```
justusschock, line 9: should we omit this as well for now to make it less discoverable?

carmocca: If an user is already importing `lite.plugins.collectives`, they've been discovered

src/lightning_lite/plugins/collectives/single_device_collective.py
```diff
@@ -0,0 +1,110 @@
...
+        *_: Any,
+        **__: Any,
+    ) -> None:
+        pass
+
+    @staticmethod
+    def init_group(
+        **kwargs: Any,
+    ) -> CollectibleGroup:
+        return object()  # type: ignore[return-value]
```
carmocca, line 102: I wonder if this should raise an error instead. It should never be used if there's no multiple devices

tests/tests_lite/plugins/collectives/test_torch_collective.py
```diff
@@ -15,64 +21,72 @@
...
+            marks=RunIf(max_torch="1.10"),
+        ),
+        pytest.param(
             "broadcast_object_list",
             {"object_list": [PASSED_OBJECT], "src": 0, "device": torch.device("cpu")},
             "object_list",
+            marks=RunIf(min_torch="1.10"),
         ),
-        (
+        pytest.param(
```
carmocca, line 87: If it's nto necessasry lets remove

src/lightning_lite/plugins/collectives/single_device_collective.py
```diff
@@ -0,0 +1,69 @@
...
+    def all_gather(self, tensor_list: List[torch.Tensor], tensor: torch.Tensor, **__: Any) -> List[torch.Tensor]:
+        return [tensor]
+
+    def gather(self, tensor: torch.Tensor, *_: Any, **__: Any) -> Optional[List[torch.Tensor]]:
+        return [tensor]
+
+    def scatter(  # type: ignore[override]
+        self,
+        tensor: torch.Tensor,
+        scatter_list: List[torch.Tensor],  # it doesn't make sense to have a None here for a single device
```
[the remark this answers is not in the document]
otaj: it's `Optional` in the base class, because it can be `None` on the ranks, that are not `src`. However, on the single device, there doesn't really exist a rank, which is not `src`, therefore it cannot be `Optional`

tests/tests_lite/plugins/collectives/test_torch_collective.py
```diff
@@ -10,9 +11,7 @@
 from lightning_lite.plugins.environments import LightningEnvironment
 from lightning_lite.strategies import DDPStrategy
 from lightning_lite.strategies.launchers.multiprocessing import _MultiProcessingLauncher
-from lightning_lite.utilities.imports import _TORCH_GREATER_EQUAL_1_11, _TORCH_GREATER_EQUAL_1_12
-
-torch_test_assert_close = torch.testing.assert_close if _TORCH_GREATER_EQUAL_1_12 else torch.testing.assert_allclose
```
carmocca: @otaj why did you remove this? isn't this needed to support <1.12?

src/lightning_lite/plugins/collectives/torch_collective.py
```diff
@@ -114,11 +114,11 @@ def monitored_barrier(self, timeout: Optional[datetime.timedelta] = None, wait_a
...
     ) -> None:
-        if main_address is not None:
-            os.environ["MASTER_ADDR"] = main_address
-        if main_port is not None:
-            os.environ["MASTER_PORT"] = str(main_port)
         if not dist.is_initialized():
+            if main_address is not None:
+                os.environ["MASTER_ADDR"] = main_address
+            if main_port is not None:
+                os.environ["MASTER_PORT"] = str(main_port)
```
carmocca, line 121: We should pop these in teardown

tests/tests_lite/plugins/collectives/test_torch_collective.py
```diff
@@ -153,67 +152,99 @@ def test_repeated_create_and_destroy():
...
+    with mock.patch("torch.distributed.init_process_group"), mock.patch(
         "torch.distributed.new_group"
     ) as new_mock, mock.patch("torch.distributed.destroy_process_group") as destroy_mock:
         collective = TorchCollective(instantiate_group=True)
         collective.teardown()
         collective.create_group()
         collective.teardown()
-    assert init_mock.call_count == 1
     assert new_mock.call_count == 2
     assert destroy_mock.call_count == 2
```
carmocca, line 169: Isn't this test redundant since we have the fixture that asserts the same?

src/lightning_lite/plugins/collectives/collective.py
```diff
@@ -30,6 +30,10 @@ def rank(self) -> int:
     def world_size(self) -> int:
         ...
 
+    @property
+    def is_member(self) -> bool:
+        return self.rank >= 0
```
carmocca, line 35: IMO we shouldn't add this, this logic should be in whoever manages the collectives

tests/tests_lite/plugins/collectives/test_torch_collective.py
```diff
@@ -153,67 +152,99 @@ def test_repeated_create_and_destroy():
...
+
+    if strategy.global_rank in (0, 1):
+        tensor = torch.tensor([strategy.global_rank])
+
+        # How do I call all_reduce with the collective object on the left group?
+        left_collective.all_reduce(tensor)
+        if left_collective.rank == 0:
+            assert tensor == 1
+
+    torch.distributed.barrier()
```
carmocca, line 231: Do we strictly need this barrier?

tests/tests_lite/plugins/collectives/test_torch_collective.py
otaj: Nope, `torch.testing.assert_close` is available since 1.9 https://github.com/pytorch/pytorch/issues/61844

src/lightning_lite/plugins/collectives/collective.py
```diff
@@ -0,0 +1,137 @@
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+import torch
+from typing_extensions import Self
+
+from lightning_lite.utilities.types import CollectibleGroup
+
+
+class Collective(ABC):
```
awaelchli, line 10: A simple docstring would be nice, something like:

A thin wrapper around torch process group for backend agnostic collective operations.

src/lightning_lite/plugins/collectives/collective.py
```diff
@@ -0,0 +1,137 @@
...
+            self.create_group()
+
+    @property
+    @abstractmethod
+    def rank(self) -> int:
+        ...
+
+    @property
+    @abstractmethod
+    def world_size(self) -> int:
```
awaelchli, line 24:
```suggestion
    def size(self) -> int:
```
or
```suggestion
    def group_size(self) -> int:
```
to be compliant with the semantics of torch groups? Thoughts?

tests/tests_lite/plugins/collectives/test_torch_collective.py
```diff
@@ -0,0 +1,237 @@
...
+        # check we can create_group again
+        collective.create_group().teardown()
+
+
+def collective_launch(fn, parallel_devices, num_groups=1):
+    strategy = DDPSpawnStrategy(
+        accelerator=CPUAccelerator(), parallel_devices=parallel_devices, cluster_environment=LightningEnvironment()
+    )
+    launcher = _MultiProcessingLauncher(strategy=strategy)
+    collectives = [TorchCollective() for _ in range(num_groups)]
```
awaelchli, line 173: TorchCollective() without specifying ranks=..., should it default to all ranks? We probably can't given the design, but would be worth checking what torch's create_group() does

awaelchli (approved):

tests/tests_lite/plugins/collectives/test_torch_collective.py
otaj: I think, it will default to all ranks on `TorchCollective` by default at this point and for any other collectives we decide to add, we can try to mimic this behavior as much as possible.

src/lightning_lite/strategies/fairscale.py
```diff
@@ -17,12 +17,14 @@
...
-from torch.distributed.constants import default_pg_timeout
 from torch.nn import Module
 from torch.optim import Optimizer
 
 from lightning_lite.accelerators import Accelerator
-from lightning_lite.plugins import CheckpointIO, ClusterEnvironment, Precision
+from lightning_lite.plugins.collectives.torch_collective import default_pg_timeout
+from lightning_lite.plugins.environments.cluster_environment import ClusterEnvironment
+from lightning_lite.plugins.io.checkpoint_plugin import CheckpointIO
+from lightning_lite.plugins.precision.precision import Precision
```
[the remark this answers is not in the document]
carmocca: Inside, we use the lowest possible to improve dependency resolution

Outside, the highest. For future compatibility

src/lightning_lite/plugins/collectives/collective.py
```diff
@@ -0,0 +1,146 @@
...
+from typing import Any, List, Optional
+
+import torch
+from typing_extensions import Self
+
+from lightning_lite.utilities.types import CollectibleGroup
+
+
+class Collective(ABC):
+    def __init__(self, instantiate_group: bool = False, **group_kwargs: Any) -> None:
```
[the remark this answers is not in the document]
carmocca: Removed these arguments

otaj (approved):

tests/tests_lite/plugins/collectives/test_single_device_collective.py
```diff
@@ -0,0 +1,34 @@
...
+
+    with pytest.raises(RuntimeError, match=r"SingleDeviceCollective` does not own a group"):
+        _ = collective.group
+
+    with mock.patch(
+        "lightning_lite.plugins.collectives.single_device_collective.SingleDeviceCollective.init_group"
+    ) as init_mock, mock.patch(
+        "lightning_lite.plugins.collectives.single_device_collective.SingleDeviceCollective.new_group"
+    ) as new_mock:
+        collective.create_group(init_kwargs={"arg2": 13, "arg3": 10}, group_kwargs={"arg1": 15, "arg3": 10})
```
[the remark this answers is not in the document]
carmocca: These are class methods, so it's not so easy. We'd need to cleanup after

tests/tests_lite/plugins/collectives/test_torch_collective.py
```diff
@@ -0,0 +1,219 @@
...
+
+    destroy_mock.assert_called_once_with(init_mock.return_value)
+    assert collective._group is None
+
+
+@RunIf(distributed=True)
+def test_init_and_new_group():
+    with mock.patch("torch.distributed.init_process_group") as init_mock, mock.patch(
+        "torch.distributed.new_group"
+    ) as new_mock, mock.patch("torch.distributed.destroy_process_group") as destroy_mock:
```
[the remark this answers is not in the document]
carmocca: Test was removed

src/lightning_lite/plugins/collectives/collective.py
carmocca: Done

src/lightning_lite/plugins/collectives/collective.py
carmocca: torch uses `get_world_size`

src/lightning_lite/plugins/collectives/single_device_collective.py
```diff
@@ -0,0 +1,81 @@
...
+class SingleDeviceCollective(Collective):
+    @property
+    def rank(self) -> int:
+        return 0
+
+    @property
+    def world_size(self) -> int:
+        return 1
+
+    def broadcast(self, tensor: torch.Tensor, *_: Any, **__: Any) -> torch.Tensor:
```
justusschock, line 18: should we name them args and kwargs like usual or do we say since they should not be used we don't care?

src/lightning_lite/plugins/collectives/single_device_collective.py
```diff
@@ -0,0 +1,81 @@
...
+    def all_reduce(self, tensor: torch.Tensor, *_: Any, **__: Any) -> torch.Tensor:
+        return tensor
+
+    def reduce(self, tensor: torch.Tensor, *_: Any, **__: Any) -> torch.Tensor:
+        return tensor
+
+    def all_gather(self, tensor_list: List[torch.Tensor], tensor: torch.Tensor, **__: Any) -> List[torch.Tensor]:
+        return [tensor]
+
+    def gather(self, tensor: torch.Tensor, *_: Any, **__: Any) -> Optional[List[torch.Tensor]]:
```
justusschock, line 30: why is the return type optional here when the tensor is not?

src/lightning_lite/plugins/collectives/torch_collective.py
```diff
@@ -0,0 +1,169 @@
...
+        op = self._convert_to_native_op(op)
+        dist.reduce(tensor, dst, op=op, group=self.group)
+        return tensor
+
+    def all_gather(self, tensor_list: List[torch.Tensor], tensor: torch.Tensor) -> List[torch.Tensor]:
+        dist.all_gather(tensor_list, tensor, group=self.group)
+        return tensor_list
+
+    def gather(
+        self, tensor: torch.Tensor, gather_list: Optional[List[torch.Tensor]] = None, dst: int = 0
```
justusschock, line 54: with an optional `gather_list` the API is not same. Can we do anything around that?
Same concern about returntype as above

src/lightning_lite/plugins/collectives/torch_collective.py
```diff
@@ -0,0 +1,169 @@
...
+    def barrier(self, device_ids: Optional[List[int]] = None) -> None:
+        dist.barrier(group=self.group, device_ids=device_ids)
+
+    def monitored_barrier(self, timeout: Optional[datetime.timedelta] = None, wait_all_ranks: bool = False) -> None:
+        dist.monitored_barrier(group=self.group, timeout=timeout, wait_all_ranks=wait_all_ranks)
+
+    def setup(
+        self, main_address: Optional[str] = None, main_port: Optional[str] = None, **kwargs: Any
+    ) -> Self:  # type: ignore[valid-type]
+        if not self.is_initialized():
```
justusschock, line 119: why do we need to set those? Once we initialized the group they should not matter anymore

justusschock (approved):

src/lightning_lite/plugins/collectives/single_device_collective.py
carmocca: since this has a `dst` in torch, the ranks not in the group return None

src/lightning_lite/plugins/collectives/single_device_collective.py
otaj: that is exactly the reason, why they are `_` and `__` -> to indicate, they are very much not used

src/lightning_lite/plugins/collectives/single_device_collective.py
otaj: Resolved

PR #15042, opened 2022-10-08
Resolve interactions between CUDA tests

carmocca:
## What does this PR do?

Resolves local failures. I am on 1.13

cc @borda

### DIFF
--- a/src/lightning_lite/accelerators/cuda.py
+++ b/src/lightning_lite/accelerators/cuda.py
@@ -82,7 +82,7 @@ def _get_all_available_cuda_gpus() -> List[int]:
 def _patch_cuda_is_available() -> Generator:
     """Context manager that safely patches :func:`torch.cuda.is_available` with its NVML-based version if
     possible."""
-    if hasattr(torch._C, "_cuda_getDeviceCount") and _device_count_nvml() >= 0:
+    if hasattr(torch._C, "_cuda_getDeviceCount") and _device_count_nvml() >= 0 and not _TORCH_GREATER_EQUAL_1_13:
         # we can safely patch is_available if both torch has CUDA compiled and the NVML count is succeeding
         # otherwise, patching is_available could lead to attribute errors or infinite recursion
         orig_check = torch.cuda.is_available
--- a/tests/tests_lite/accelerators/test_cuda.py
+++ b/tests/tests_lite/accelerators/test_cuda.py
@@ -61,3 +61,4 @@ def test_num_cuda_devices_without_nvml(*_):
     num_cuda_devices.cache_clear()
     assert is_cuda_available()
     assert num_cuda_devices() == 100
+    num_cuda_devices.cache_clear()


src/lightning_lite/accelerators/cuda.py
```diff
@@ -82,7 +82,7 @@ def _get_all_available_cuda_gpus() -> List[int]:
 def _patch_cuda_is_available() -> Generator:
     """Context manager that safely patches :func:`torch.cuda.is_available` with its NVML-based version if
     possible."""
-    if hasattr(torch._C, "_cuda_getDeviceCount") and _device_count_nvml() >= 0:
+    if hasattr(torch._C, "_cuda_getDeviceCount") and _device_count_nvml() >= 0 and not _TORCH_GREATER_EQUAL_1_13:
```
carmocca, line 85: This was causing a `RecursionError` when these two ran together

`pytest tests/tests_lite/accelerators/test_cuda.py tests/tests_lite/plugins/precision/test_native_amp.py`

tests/tests_lite/accelerators/test_cuda.py
```diff
@@ -61,3 +61,4 @@ def test_num_cuda_devices_without_nvml(*_):
     num_cuda_devices.cache_clear()
     assert is_cuda_available()
     assert num_cuda_devices() == 100
+    num_cuda_devices.cache_clear()
```
carmocca, line 64: This was causing

`E       RuntimeError: CUDA error: invalid device ordinal` with the command:

`pytest tests/tests_lite/accelerators/test_cuda.py tests/tests_lite/plugins/precision/test_native_amp_integration.py`

carmocca:
cc @speediedan

awaelchli (approved):

justusschock (approved):

src/lightning_lite/accelerators/cuda.py
speediedan: Hmm, I'm unable to reproduce the recursion issue with `torch` 1.13 and the latest lightning commits (I also saw no errors with the 1.14 nightly btw): 

```bash
git rev-parse HEAD && python -c "import torch;import pytorch_lightning as pl;print(torch.__version__);print(pl.__version__);"
ca3c4e7f0723b0d173a63e4cc11bebb414de9b4b
1.13.0
1.8.0rc0
$ pytest tests/tests_lite/accelerators/test_cuda.py tests/tests_lite/plugins/precision/test_native_amp.py
...
collected 15 items                                                                                                                                                                          
...
tests/tests_lite/accelerators/test_cuda.py ........                                                                                                                                   [ 53%]
tests/tests_lite/plugins/precision/test_native_amp.py .......                                                                                                                         [100%]
============================================================================== 15 passed,
...
```

Adding the `not _TORCH_GREATER_EQUAL_1_13` condition will still lead to bad forks even in 1.13 at the moment. The `torch.cuda.is_available` patch this block performs is still required with 1.13 (unless PyTorch PR [#85951](https://github.com/pytorch/pytorch/pull/85951) gets merged before 1.13 which we can't count on) so I think it'd be best to find another way around the recursion if possible. Since I can't reproduce on my end, would you mind posting the relevant traceback?

awaelchli: I'm sorry, I accidentally resolved this conversation and it auto-merged!

carmocca: @speediedan For the `RecursionError` to appear, you need to revert the change in this file but keep the added `cache_clear` in `test_cuda.py`.

Maybe let's assume the patch to PyTorch won't make it and convert these to 1_14 already. wdyt?

PR #15043, opened 2022-10-08
Enable some CPU tests on draft mode

carmocca:
## What does this PR do?

See title. The list and parameterizations of jobs is be unchanged.

### DIFF
--- a/.github/workflows/ci-pytorch-test-full.yml
+++ b/.github/workflows/ci-pytorch-test-full.yml
@@ -25,7 +25,7 @@ jobs:
 
   pl-cpu:
     runs-on: ${{ matrix.os }}
-    if: github.event.pull_request.draft == false
+    if: ${{ github.event.pull_request.draft == false || matrix.draft == true }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-pytorch-test-full.yml
```diff
@@ -36,17 +36,17 @@ jobs:
           - {os: "ubuntu-20.04", python-version: "3.8", pytorch-version: "1.10"}  # ex-conda
           - {os: "ubuntu-20.04", python-version: "3.9", pytorch-version: "1.11"}  # ex-conda
           - {os: "ubuntu-20.04", python-version: "3.9", pytorch-version: "1.12"}  # ex-conda
-          - {os: "ubuntu-20.04", python-version: "3.10", pytorch-version: "1.12"}
+          - {os: "ubuntu-20.04", python-version: "3.10", pytorch-version: "1.12", draft: "true"}
           - {os: "macos-11", python-version: "3.7", pytorch-version: "1.9", requires: "oldest"}
```
carmocca, line 40: @akihironitta unrelated question. why do we not set `requires` for all jobs?

carmocca:
Looks like this is not supported at the moment: https://github.com/actions/runner/issues/1985

Borda (commented):
you would need to make the workflow abstract and call it in two jobs, one running with any other running without draft

.github/workflows/ci-pytorch-test-full.yml
akihironitta: @carmocca I'm sorry I missed your message here. `requires: "latest"` wasn't used in any steps' conditions in the workflow, so I only kept `requires: "oldest"`. We can add those back here (or remove them in other workflows, too) for consistency.

carmocca: It's okay. Now I see that it's simpler to omit it

PR #15058, opened 2022-10-10
Standardize Lite's filenames

carmocca:
## What does this PR do?

Lite's naming scheme is inconsistent:
- Accelerator, Strategy and Precision do not include the suffix. For example, we use `cpu.py` instead of `cpu_accelerator.py`
- ClusterEnvironment and Collective do include it. For example `lsf_environment.py` and `single_device_collective.py`
- CheckpointIO has `_plugin`. For example `torch_plugin.py`

This PR removes the suffix from all filenames **with the exception** of `torch_collective.py` and `torch_checkpoint_io.py` so they are not mistaken for actual `torch`.

None


otaj:
Can we resolve these inconsistencies the other way around, i.e. putting suffix to **ALL** files? That way, although the names will be quite long, we won't have to have exceptions to the rule.

carmocca:
I don't mind either way. Suffix just has the annoyance of longer and repeating imports. Let's vote in your comment

src/lightning_lite/plugins/__init__.py
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from lightning_lite.plugins.environments.cluster_environment import ClusterEnvironment
-from lightning_lite.plugins.io.checkpoint_plugin import CheckpointIO
-from lightning_lite.plugins.io.torch_plugin import TorchCheckpointIO
-from lightning_lite.plugins.io.xla_plugin import XLACheckpointIO
+from lightning_lite.plugins.io.checkpoint_io import CheckpointIO
+from lightning_lite.plugins.io.torch_checkpoint_io import TorchCheckpointIO
```
awaelchli, line 16: We could do `torch_` the same way we do for `logging_` in trainer. This wouldn't affect users really, because they are supposed to import from the higher plugin level.

awaelchli (approved):

src/lightning_lite/plugins/__init__.py
Borda: also we could reduce `_io` as it is the module just before

Borda, line 16:
```suggestion
from lightning_lite.plugins.io.checkpointing import CheckpointIO
from lightning_lite.plugins.io.torch_checkpointing import TorchCheckpointIO
```

carmocca: That doesn't match the class names

otaj: I agree with @awaelchli's proposal.

src/lightning_lite/plugins/__init__.py
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from lightning_lite.plugins.environments.cluster_environment import ClusterEnvironment
-from lightning_lite.plugins.io.checkpoint_plugin import CheckpointIO
-from lightning_lite.plugins.io.torch_plugin import TorchCheckpointIO
-from lightning_lite.plugins.io.xla_plugin import XLACheckpointIO
+from lightning_lite.plugins.io.checkpoint_io import CheckpointIO
+from lightning_lite.plugins.io.torch_ import TorchCheckpointIO
```
tchaton, line 16: Maybe torch_io, xla_io to avoid the weird torch_ ?

tchaton (approved):
LGTM !

justusschock (approved):

src/lightning_lite/plugins/__init__.py
carmocca: @awaelchli, I see you thumbs-upped this comment, but underscore was your suggestion: https://github.com/Lightning-AI/lightning/pull/15058#discussion_r991509209

I don't mind either option. This can be changed later anyways

awaelchli: I think in this case it's a good suggestion. Either way is fine with me.

carmocca: Doing this in a different PR

justusschock: I don't like this tbh this is again a inconsistency. If we really want to get rid of the weird `torch_`, I'd rather rename it to `pytorch`, but not add a inconsistency again just for this case (which is even weirder IMO)

carmocca: This disagreement is why I chose to do this in a separate PR. Please :+1: or :-1: #15189 and I'll close or proceed based on the results.

justusschock: yes, just wanted to present an alternative :D

PR #15095, opened 2022-10-12
CI: add pytest for LAI

Borda:
## What does this PR do?

adding pytest for the full `lightning` package

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @tchaton @rohitgr7 @carmocca @akihironitta @borda

PR #15105, opened 2022-10-12
LAI: creating mirror package

Borda:
## What does this PR do?

repacing meta-package by mirror-package
interested in cross imports use copy source files
Requires (?) #15095
Closes #15122
Closes #15297

## Before submitting

## PR review

## Did you have fun?

cc @justusschock @awaelchli @rohitgr7 @tchaton @borda


awaelchli (approved):

.github/workflows/ci-pkg-install.yml
```diff
@@ -17,25 +17,14 @@ defaults:
...
     strategy:
-      fail-fast: true
-      max-parallel: 1
+      fail-fast: false
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04, macOS-11, macOS-12, windows-2022]
-        pkg: ["app", "lite", "pytorch"]
-        python-version: [3.8]  # , 3.9
+        pkg: ["app", "lite", "pytorch", ""]
+        python-version: ["3.7" , "3.10"]  #
```
Borda, line 27:
```suggestion
        python-version: ["3.7" , "3.10"]
```

.actions/setup_tools.py
```diff
@@ -394,74 +178,33 @@ def parse_version_from_file(pkg_root: str) -> str:
...
-        # clean to many empty lines
-        body = [ln for ln, _group in groupby(body)]
-        # drop duplicated lines
-        body = prune_duplicate_lines(body)
-        # compose the target file name
+        with open(py_file, encoding="utf-8") as fo:
+            py = fo.readlines()
+        for n2, n1 in pkg_lut.items():
+            for i, ln in enumerate(py):
+                py[i] = re.sub(rf"(?!_){n1}(?!_)", f"lightning.{n2}", ln)
```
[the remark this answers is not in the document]
otaj: there are instances in app package, where variable is called `lightning_app` or `pytorch_lightning`, this would not guard against it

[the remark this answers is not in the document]
Borda: see: https://github.com/Lightning-AI/lightning/pull/15207#discussion_r1000366666
and the 5 cases I have edited in 0a17aba13e6087dbd0cc5cf932236491e7d38763

Borda:
It seems all tests for Lite work on stand-alone as well as monolithic package
For PL test with the monolithic package, it executes and fails on distributed, which I believe shall be skipped
and for App it works with the monolithic package, but one failed test with the stand-alone pkg.

.actions/setup_tools.py
```diff
@@ -394,74 +178,63 @@ def parse_version_from_file(pkg_root: str) -> str:
...
+     'delete_cloud_lightning_apps',
+     'from lightning.app import',
+     'lightning_apps = []',
+     'lightning.app is ours',
+     'def _lightning_app():',
+     ':class:`~lightning.app.core.flow.LightningFlow`']
+    """
+    for n2, n1 in pkg_lut.items():
+        for i, ln in enumerate(lines):
+            lines[i] = re.sub(rf"([^_]|^){n1}([^_\w]|$)", rf"\1lightning.{n2}\2", ln)
```
justusschock, line 203: can you add a comment for what this regex does so that we don't need to look it up every time?

.actions/setup_tools.py
```diff
@@ -394,74 +178,63 @@ def parse_version_from_file(pkg_root: str) -> str:
...
+    return lines
 
-    As validation run in termnal: `flake8 src/lightning/ --ignore E402,F401,E501`
 
-    >>> create_meta_package(os.path.join(_PROJECT_ROOT, "src"))
-    """
+def copy_adjusted_modules(src_folder: str, pkg_name: str, lit_name: str, pkg_lut: dict) -> None:
```
justusschock, line 207: couldn't we just import this from the assistant.py? Should be exactly the same.

justusschock (approved):

.actions/setup_tools.py
Borda: I belive it is covered by the example in docstring :otter:

.actions/setup_tools.py
Borda: assistant is not distributed with the package, so in contrast, we may think about importing `setup_tools` in assistant...

justusschock: fine with me as well. Just saying that we don't want the same logic twice unless absolutely necessary.

Borda: Agree, would you be fine to leave this unification in separate PR?
```suggestion
# TODO: unify usage with assistant function, such that import this function in there
def copy_adjusted_modules(src_folder: str, pkg_name: str, lit_name: str, pkg_lut: dict) -> None:
```

Borda:
> and for App it works with the monolithic package, but one failed test with the stand-alone pkg.
> 
> ```
> E       lightning_app.utilities.exceptions.MisconfigurationException: The provided file /home/runner/work/lightning/lightning/examples/app_v0/app.py does not contain a LightningApp. Instantiate your app at the module level like so: `app = LightningApp(flow, ...)`
> ```

shall be fixed by #15260
cc: @otaj @manskx

.actions/setup_tools.py
justusschock: definitely!

.github/actions/pkg-check/action.yml
```diff
@@ -5,6 +5,7 @@ inputs:
   pkg-name:
     description: package name inside lightning.*
     required: true
+    default: "lightning"
```
Borda, line 8:
```suggestion
    default: ""
```

.github/workflows/ci-app-examples.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, macOS-11, windows-2022]
-        pkg-name: ["app"]  # TODO: add "LAI" # "LAI" installs the monolithic package
-        python-version: [3.8]
+        pkg-name: ["lightning", "app"]  # the "lightning" is for installing monolithic package
```
carmocca, line 35: I strongly suggest we add parametrizations separately, as I did in #15279

requirements/app/test.txt
```diff
@@ -3,7 +3,7 @@ codecov>=2.1, <=2.1.12
 pytest>=7.0, <=7.1.2
 pytest-timeout <=2.1.0
 pytest-cov <=3.0.0
-playwright==1.27.1
+playwright>=1.27.1, <1.27.2
```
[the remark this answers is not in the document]
Borda:
```suggestion
playwright==1.27.1
```

.github/workflows/ci-pkg-install.yml
```diff
@@ -17,141 +17,40 @@ defaults:
...
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install pytest extensions
+        # TODO: for mono-package create & load requirements/test.txt
+        if: ${{ matrix.pkg != '' }}
+        run: grep 'pytest>' requirements/${{ matrix.pkg }}/test.txt | xargs -0 pip install
+
       - name: DocTests actions
```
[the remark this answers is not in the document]
Borda: not really, no other parametrization checks that the CLI behave as expected

.github/workflows/ci-app-examples.yml
carmocca: You can revert the added parametrizations once CI is green here

.github/workflows/ci-pkg-install.yml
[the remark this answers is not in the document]
carmocca: You can add the `lighting --version` check to the other workflow. Launching a separate job just for that is overkill

And the doctests are already tested in the other one.

.github/workflows/ci-app-examples.yml
Borda: but why?

.github/workflows/ci-pkg-install.yml
```diff
@@ -17,141 +17,40 @@ defaults:
...
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
-      max-parallel: 1
+      fail-fast: false
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04, macOS-11, macOS-12, windows-2022]
-        pkg: ["app", "lite", "pytorch"]
-        python-version: [3.8]  # , 3.9
+        pkg: ["app", "lite", "pytorch", ""]
```
carmocca, line 26: Can this use "lightning" as everywhere else?

.github/workflows/ci-pkg-install.yml
```diff
@@ -17,141 +17,40 @@ defaults:
...
-  install-standalone:
-    needs: init-temp
+  install-pkg:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
-      max-parallel: 1
+      fail-fast: false
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04, macOS-11, macOS-12, windows-2022]
```
carmocca, line 25: Why do we need multiple OS versions?

.github/workflows/ci-pkg-install.yml
```diff
@@ -17,141 +17,40 @@ defaults:
...
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install pytest extensions
+        # TODO: for mono-package create & load requirements/test.txt
+        if: ${{ matrix.pkg != '' }}
+        run: grep 'pytest>' requirements/${{ matrix.pkg }}/test.txt | xargs -0 pip install
+
```
carmocca, line 39: Why do we need pytest extensions? This job only tests setup-tools and checks the lightning version

.actions/setup_tools.py
```diff
@@ -394,74 +178,64 @@ def parse_version_from_file(pkg_root: str) -> str:
...
-        if ln.lstrip() not in body_ or ln.lstrip() in (")", ""):
-            body_.append(ln)
-    return body_
-
-
-def create_meta_package(src_folder: str, pkg_name: str = "pytorch_lightning", lit_name: str = "pytorch"):
-    """Parse the real python package and for each module create a mirroe version with repalcing all function and
-    class implementations by cross-imports to the true package.
+def _replace_imports_in_file(lines: List[str], pkg_lut: Dict[str, str]) -> List[str]:
+    """REplace imports of standlone package to lightning.
```
carmocca, line 182:
```suggestion
    """Replace imports of standalone package to lightning.
```

.azure/app-cloud-e2e.yml
```diff
@@ -133,6 +133,14 @@ jobs:
       condition: eq(variables['name'], 'quick_start')
       displayName: 'Install Quick Start'
 
+    - bash: python -m pip install -q -r .actions/requirements.txt
+      displayName: 'Install assistant dependencies'
+
+    - bash: |
+        python .actions/assistant.py copy_replace_imports --source_dir="./examples" --source_import="lightning.app" --target_import="lightning_app"
+        python .actions/assistant.py copy_replace_imports --source_dir="./examples" --source_import="lightning" --target_import="lightning_app"
```
carmocca, line 141: Why do we want the standalone imports here instead of the `lightning` alternatives?

.github/actions/pkg-check/action.yml
```diff
@@ -5,6 +5,7 @@ inputs:
   pkg-name:
     description: package name inside lightning.*
     required: true
+    default: ""
```
carmocca, line 8: should default be lightning?

.github/workflows/ci-app-examples.yml
carmocca: Because I would like for us to really think about what jobs we will be adding, and adding them to checkgroup etc, but this PR is already pretty long, I don't want it to get huge.

.github/workflows/ci-app-tests.yml
```diff
@@ -64,9 +64,11 @@ jobs:
           pip list
 
       - name: DocTests App
-        if: ${{ matrix.pkg-name != 'LAI' }}
         working-directory: src
-        run: python -m pytest lightning_app --ignore-glob="**/cli/*-template/**"
+        run: |
+          pip install pytest -q
```
carmocca, line 69: Why does this install pytest here? Shouldn't the "Install dependencies" below (L84) be merged with "Install package" above (L59)?

.github/workflows/ci-lite-test-full.yml
```diff
@@ -83,9 +83,11 @@ jobs:
         pip list
 
     - name: DocTests Lite
-      if: ${{ matrix.pkg-name != 'LAI' }}
       working-directory: src
-      run: python -m pytest lightning_lite
+      run: |
+        pip install pytest -q
```
carmocca, line 88: Same question here

.github/workflows/ci-pytorch-test-full.yml
```diff
@@ -111,13 +111,15 @@ jobs:
...
+        pip install -e . --upgrade --find-links ${TORCH_URL}
         pip list
 
     - name: DocTests PL
-      if: ${{ matrix.pkg-name != 'LAI' }}
-      working-directory: src
-      run: python -m pytest pytorch_lightning
+      working-directory: ./src
+      run: |
+        pip install pytest -q
```
carmocca, line 120: Same question here

docs/source-lit/conf.py
```diff
@@ -398,13 +398,13 @@ def find_source():
 from typing import Optional
 
 import torch
-import pytorch_lightning as pl
+import lightning.pytorch as pl
```
carmocca, line 401: Why are we updating the docs import? The rest of the docs don't use the newer imports, no?

Borda: this docs `source-lit` is specific as they will refer only to `lightning` package

.github/workflows/ci-app-tests.yml
Borda: Not really, we want to have a test if package is fine without installing al extras, so for a user who would just install PL/App it works, which was the case with Apps that it was not true and was crashing

.github/workflows/ci-lite-test-full.yml
Borda: Will mark this as resolved as it has the same question/answer, so we can keep just one sample :)

.github/workflows/ci-pytorch-test-full.yml
Borda: Will mark this as resolved as it has the same question/answer, so we can keep just one sample :)

.github/actions/pkg-check/action.yml
Borda: it can be anything except `app|lite|pytorch` so why add noise to have defaulted, in fact, default is empty as most users shall install straight without setting any env. variables

.azure/app-cloud-e2e.yml
Borda: because we are testing App as standalone which means that `lightning` does not exists

.github/workflows/ci-app-examples.yml
Borda: So are you suggesting making these tests run once and then dropping CI for its codebase and praying that all will still work even though we won't be testing it?
cc: @otaj @justusschock

.github/workflows/ci-pkg-install.yml
Borda: This is not an edit of this PR :otter:

.github/workflows/ci-pkg-install.yml
Borda:
```suggestion
```
good question, but it was there already, not addition of this PR

docs/source-lit/conf.py
Borda: but more important with this package you won't get `lightning_app` nor `pytorch_lightning` anymore so in such case your package will be completely broken

.azure/app-cloud-e2e.yml
carmocca: Okay. I guess this will change when we parametrize this file

.github/workflows/ci-app-examples.yml
carmocca: We often sequence pieces of work where a fix is pushed before the tests are added.

In fact, previous PRs of this line of work already merged worked without tests being ready.

I'm not saying that the changes shouldn't be tested, just that it would be a better experience for reviewers if we sequence it more finely.

.github/workflows/ci-pkg-install.yml
carmocca: I thought they were the same. What is the difference?

.github/workflows/ci-pkg-install.yml
carmocca: I was just asking, didn't say you needed to modify this here.

.github/workflows/ci-lite-test-full.yml
carmocca: Repeating the question can be helpful in case you do one change, it ensures you don't miss out any other occurrences. It's not meant as spam.

.github/workflows/ci-app-tests.yml
carmocca: What does that have to do with doctests? Users would not run them

Maybe you want to do that instead in `test-pkg`: https://github.com/Lightning-AI/lightning/pull/15105/files#diff-f7f8305e901571c1ad55ba41890b6c6bb21aa4a1ba94b61ee77239860fe0bf1bR20 since we have it already

.github/workflows/ci-lite-test-full.yml
Borda: that is good point, and I ma fine with it, but lets keep conversation just with one

.github/workflows/ci-pkg-install.yml
Borda: it was added a few weeks ago as a reaction to @williamFalcon concerns that we can't install lightning on some machines such as M1 with py3.8

.github/workflows/ci-pkg-install.yml
Borda: They have the effect, but from practical point of view, almost no user will have set this env variable by default to "lightning"

.github/workflows/ci-app-examples.yml
Borda: > We often sequence pieces of work where a fix is pushed before the tests are added.

really? that not good, that it is against the best practice

> In fact, previous PRs of this line of work already merged worked without tests being ready.

I don't think so; this CI was added before this PR, or are you referring to fixes in the package/test for PL, which was done on green status?

.github/workflows/ci-app-tests.yml
Borda: yes we can run doctests in installation workflow...

Borda:
@carmocca, regarding the parameterization, are you fine with replacing adding the full matrix with adding a few configurations in the `include` matrix section? :rabbit:

.github/workflows/ci-app-examples.yml
Borda: pruned the jobs count here f1630e0786a04952bb9f1377645391165246aaba

tests/tests_pytorch/loops/test_training_loop.py
```diff
@@ -140,6 +141,7 @@ def validation_step(self, *args):
     assert model.validation_called_at == (0, 5)
 
 
+@RunIf(standalone=True)
 def test_fit_loop_done_log_messages(caplog):
```
awaelchli, line 145: why is this being marked as standalone? Let's undo this.

tests/tests_pytorch/loops/test_training_loop.py
```diff
@@ -210,6 +212,7 @@ def training_step_end(self, outputs):
         (4, 10, 4, True, True, True),
     ],
 )
+@RunIf(standalone=True)
```
awaelchli, line 215: same comment

tests/tests_app_examples/test_argparse.py
```diff
@@ -10,12 +10,13 @@ def test_app_argparse_example():
...
 
     command_line = [
-        os.path.join(os.path.dirname(os.path.dirname(_PACKAGE_ROOT)), "examples/app_argparse/app.py"),
+        os.path.join(_PROJECT_ROOT, "examples/app_argparse/app.py"),
         "--app_args",
         "--use_gpu",
         "--without-server",
     ]
     result = application_testing(command_line=command_line)
+    print(result.stdout)
```
awaelchli, line 19: debugging leftovers?

docs/source-lit/conf.py
```diff
@@ -398,20 +398,20 @@ def find_source():
...
     _TORCH_GREATER_EQUAL_1_10,
 )
-from pytorch_lightning.loggers.neptune import _NEPTUNE_AVAILABLE
-from pytorch_lightning.loggers.comet import _COMET_AVAILABLE
-from pytorch_lightning.loggers.mlflow import _MLFLOW_AVAILABLE
-from pytorch_lightning.loggers.wandb import _WANDB_AVAILABLE
+from pytorch.lightning.loggers.neptune import _NEPTUNE_AVAILABLE
+from pytorch.lightning.loggers.comet import _COMET_AVAILABLE
+from pytorch.lightning.loggers.mlflow import _MLFLOW_AVAILABLE
+from pytorch.lightning.loggers.wandb import _WANDB_AVAILABLE
```
awaelchli, line 415: This is in reverse. Can we change it to lightning.pytorch?

Borda: great catCT!!

tests/tests_app_examples/test_argparse.py
Borda:
```suggestion
```

tests/tests_pytorch/loops/test_training_loop.py
Borda: @justusschock :otter:

tests/tests_pytorch/loops/test_training_loop.py
```diff
@@ -140,6 +141,7 @@ def validation_step(self, *args):
     assert model.validation_called_at == (0, 5)
 
 
+@RunIf(standalone=True)
```
Borda, line 144:
```suggestion
```

tests/tests_pytorch/loops/test_training_loop.py
Borda:
```suggestion
```

tests/tests_pytorch/loops/test_training_loop.py
justusschock: this is because it would fail otherwise due to importing the checkpoints and thereby importing PL as standalone.

#15309 is supposed to fix this.

awaelchli: Ok, let's land this PR https://github.com/Lightning-AI/lightning/pull/15309 and see if it fixes it!

carmocca (commented):
Very nice! Almost there.

I updated the checkgroup config with the new and changed jobs

.github/workflows/ci-pkg-install.yml
```diff
@@ -51,112 +40,24 @@ jobs:
...
-
-      - uses: ./.github/actions/pkg-install
-        with:
-          pkg-name: "lightning"
-          pip-flags: "-U --pre --find-links ../pypi/"
-
-      - name: Run CLI
-        run: python -m lightning --version
+          pip list
+          scope=$(python -c "lut = {'app': 'lightning_app', 'lite': 'lightning_lite', 'pytorch': 'pytorch_lightning'} ; print(lut.get('${{matrix.pkg}}', 'lightning'))")
```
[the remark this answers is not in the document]
Borda: look-up-table

tests/tests_pytorch/conftest.py
```diff
@@ -191,11 +191,23 @@ def caplog(caplog):
     """
     import logging
 
-    lightning_logger = logging.getLogger("pytorch_lightning")
-    propagate = lightning_logger.propagate
-    lightning_logger.propagate = True
+    root_logger = logging.getLogger()
```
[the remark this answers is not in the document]
justusschock: not strictly, but it is a more complete logger update than the one before where the propagation was not always fixed correctly. We should be able to revert it if you prefer minimal changes though

.github/workflows/ci-app-tests.yml
```diff
@@ -80,6 +70,15 @@ jobs:
           key: ${{ runner.os }}-pip-py${{ matrix.python-version }}-${{ matrix.pkg-name }}-${{ matrix.requires }}-${{ hashFiles('requirements/app/base.txt') }}
           restore-keys: ${{ runner.os }}-pip-py${{ matrix.python-version }}-${{ matrix.pkg-name }}-${{ matrix.requires }}-
 
+      - name: switch PyTorch URL
```
tchaton, line 73:
```suggestion
      - name: Switch PyTorch URL
```

.github/workflows/ci-lite-test-full.yml
```diff
@@ -101,13 +96,18 @@ jobs:
...
+      env:
+        PACKAGE_NAME: ${{ matrix.pkg-name }}
       run: |
-        pip install -r requirements/lite/devel.txt --upgrade --find-links ${TORCH_URL}
+        pip install -e . -r requirements/lite/devel.txt --upgrade --find-links ${TORCH_URL}
         pip list
 
     - name: Adjust tests
-      if: ${{ matrix.pkg-name == 'LAI' }}
+      if: ${{ matrix.pkg-name == 'lightning' }}
```
tchaton, line 110: Same comment as above.

.github/workflows/ci-pkg-install.yml
tchaton: @Borda Let's use lightning everywhere. The CI is already confusing enough for new users.

setup.py
```diff
@@ -25,12 +25,13 @@
...
 
      - for `pytorch-lightning` use `export PACKAGE_NAME=pytorch ; pip install .`
+     - for `lightning-lite` use `export PACKAGE_NAME=lite ; pip install .`
      - for `lightning-app` use `export PACKAGE_NAME=app ; pip install .`
 
 3. Building packages as sdist or binary wheel and installing or publish to PyPI afterwords you use command
     `python setup.py sdist` or `python setup.py bdist_wheel` accordingly.
    In case you want to build just a particular package you would use exporting env. variable as above:
-   `export PACKAGE_NAME=pytorch|app ; python setup.py sdist bdist_wheel`
+   `export PACKAGE_NAME=pytorch|app|lite ; python setup.py sdist bdist_wheel`
```
tchaton, line 34:
```suggestion
   `export PACKAGE_NAME=pytorch|app|lite|lightning ; python setup.py sdist bdist_wheel`
```

.github/workflows/ci-app-examples.yml
```diff
@@ -89,11 +94,17 @@ jobs:
         run: pip install -e .
 
       - name: Adjust tests
-        if: ${{ matrix.pkg-name == 'LAI' }}
+        if: ${{ matrix.pkg-name == 'lightning' }}
         run: python .actions/assistant.py copy_replace_imports --source_dir="./tests" --source_import="lightning_app" --target_import="lightning.app"
 
+      - name: Adjust examples
```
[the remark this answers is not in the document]
manskx: ideally, we want to run the same tests with `lightning` and `lightning_app` to make sure both packages work

.github/workflows/ci-app-examples.yml
```diff
@@ -89,11 +94,17 @@ jobs:
         run: pip install -e .
 
       - name: Adjust tests
-        if: ${{ matrix.pkg-name == 'LAI' }}
+        if: ${{ matrix.pkg-name == 'lightning' }}
```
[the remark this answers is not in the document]
akihironitta: There're a few `pkg-name: "lightning"` at L40-L42.

.github/actions/pkg-check/action.yml
```diff
@@ -21,10 +22,7 @@ runs:
   - name: Source check
     env:
       PACKAGE_NAME: ${{ inputs.pkg-name }}
-    run: |
-      python setup.py check --metadata --strict
-      # TODO: fix reimporting (F811) which is aftefact after rplacing var assigne with import in meta package
-      flake8 src/lightning/ --ignore E402,F401,E501,W391,E303,F811
+    run: python setup.py check --metadata --strict
```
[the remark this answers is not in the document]
Borda: we do not create anything new here, so it is already check with the original codebase

setup.py
Borda: no really, `lightning` does not have any effect

tests/tests_pytorch/conftest.py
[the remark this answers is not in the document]
carmocca: I don't understand why we need it. Propagation only needs to be fixed for the root logger, no? 

What would fail if this is reverted?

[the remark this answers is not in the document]
justusschock: nothing would fail, but propagation setting on the root logger would only work if sub-loggers have not yet been created.

But again: For this PR it isn't needed, so here we can revert it if you prefer :)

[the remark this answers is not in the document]
carmocca: How could a sub-logger be created without the root first? If somebody manually did the following?

```py
getLogger("pytorch_lightning.foo.bar")
import pytorch_lightning  # will configure the pl logger
```

Up to you. If you think it's important that we keep it, then I'm fine with it.

.github/checkgroup.yml
```diff
@@ -241,18 +279,24 @@ subprojects:
...
-      - "app-examples (macOS-11, app, 3.8, latest)"
-      - "app-examples (macOS-11, app, 3.8, oldest)"
-      - "app-examples (ubuntu-20.04, app, 3.8, latest)"
-      - "app-examples (ubuntu-20.04, app, 3.8, oldest)"
-      - "app-examples (windows-2022, app, 3.8, latest)"
-      - "app-examples (windows-2022, app, 3.8, oldest)"
+      - "app-examples (macOS-11, app, 3.9, latest)"
+      - "app-examples (macOS-11, app, 3.9, oldest)"
+      - "app-examples (ubuntu-20.04, app, 3.9, latest)"
+      - "app-examples (ubuntu-20.04, app, 3.9, oldest)"
```
otaj, line 285: I know it's not really within scope of this PR, but can we have the same machines everywhere? Currently, we have `windows-2022`, `ubuntu-20.04`, `ubuntu-22.04`, `macOS-11`, `macOS-12`

tests/tests_pytorch/conftest.py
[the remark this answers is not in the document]
justusschock: If you create a sub-logger with propagate = False and only turn it on at the root logger, then it still won't propagate and this way it will :)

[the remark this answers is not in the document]
justusschock: I'd prefer to keep it.

otaj (approved):

.github/checkgroup.yml
Borda: I would say it is good to have a reasonable mix, as a kind of sampling over a large configuration space... but I would say that all shall be used for testing if the package can be installed :otter:

carmocca: I would always use latest since we have no code that depends on the specific os version. But yes, this would be proposed in a separate PR.

.actions/setup_tools.py
Borda: > The change should be quite simple, and it avoids adding all this code, to the best of my knowledge, this would be the diff

I think it is the issue with the diff tool as we removed 6 functions completely and wrote two new, but the diff is trying to find some "random parity and then it shows wild diff :/"

carmocca: My point is that doing what @justusschock proposed at the top should be easy, just need to load the assistant and call it as I showed above. Do you see any other issues?

manskx (approved):

.github/workflows/ci-lite-tests.yml
```diff
@@ -72,21 +82,6 @@ jobs:
...
-      run: python -c "print('TORCH_URL=https://download.pytorch.org/whl/' + str('test/cpu/torch_test.html' if '${{matrix.release}}' == 'pre' else 'cpu/torch_stable.html'))" >> $GITHUB_ENV
-
-    - name: Install package
-      env:
-        PACKAGE_NAME: ${{ matrix.pkg-name }}
-      run: |
-        pip install -e . pytest --upgrade --find-links ${TORCH_URL}
-        pip list
-
-    - name: DocTests Lite
```
carmocca, line 85: Is this doctest step removed on purpose?

Borda:
pending on PL fixes in #15329

carmocca (dismissed):

.actions/setup_tools.py
carmocca: I'll resolve to unblock

Borda:
@carmocca @justusschock adding back the standalone run to the 3 PL tests... :innocent:

Borda:
dropping building older CUDA images with  olderPT version because of Apex, see: #15345

.github/checkgroup.yml
```diff
@@ -165,24 +181,35 @@ subprojects:
...
       - "pl-cpu (macOS-11, pytorch, 3.7, 1.9, oldest)"
       - "pl-cpu (macOS-11, pytorch, 3.10, 1.13, pre)"
-      - "pl-cpu (ubuntu-20.04, pytorch, 3.8, 1.11)"
+      - "pl-cpu (ubuntu-20.04, pytorch, 3.8, 1.10)"
+      - "pl-cpu (ubuntu-20.04, pytorch, 3.9, 1.11)"
+      - "pl-cpu (ubuntu-20.04, pytorch, 3.10, 1.11)"
       - "pl-cpu (ubuntu-20.04, pytorch, 3.10, 1.12)"
       - "pl-cpu (ubuntu-20.04, pytorch, 3.7, 1.9, oldest)"
       - "pl-cpu (ubuntu-20.04, pytorch, 3.9, 1.13, pre)"
       - "pl-cpu (windows-2022, pytorch, 3.8, 1.9)"
```
carmocca, line 201: You removed this job from the workflow

```suggestion
```

carmocca:
The checkgroup.yml is broken after the master merge. Please be very careful with the update as errors there can block master. If you are unsure, let me make the updates.

.github/workflows/ci-lite-tests.yml
```diff
@@ -54,8 +54,12 @@ jobs:
           - {os: "macOS-11", pkg-name: "lite", python-version: "3.10", pytorch-version: "1.13", release: "pre"}
           - {os: "ubuntu-20.04", pkg-name: "lite", python-version: "3.9", pytorch-version: "1.13", release: "pre"}
           - {os: "windows-2022", pkg-name: "lite", python-version: "3.8", pytorch-version: "1.13", release: "pre"}
+          # "lightning" installs the monolithic package
+          - {os: "macOS-11", pkg-name: "lightning", python-version: "3.8", pytorch-version: "1.12"}
+          - {os: "ubuntu-20.04", pkg-name: "lightning", python-version: "3.8", pytorch-version: "1.12"}
+          - {os: "windows-2022", pkg-name: "lightning", python-version: "3.8", pytorch-version: "1.12"}
 
-    timeout-minutes: 10
+    timeout-minutes: 15
```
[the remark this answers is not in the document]
otaj: I've noticed, that 10 minutes is sometimes not enough to install everything on all instances. I didn't do the change, but I would propose the same one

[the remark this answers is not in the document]
Borda: yes, install om Mac sometimes takes too long

Borda:
> Please be very careful with the update as errors there can block master. If you are unsure, let me make the updates.

sorry, and for sure will do :) from now on :rabbit:

Borda:
as we are going to drop support for PT 19 in #15347 I am dropping testing for oldes and monolitic pkg

tests/tests_pytorch/loops/test_training_loop.py
awaelchli: https://github.com/Lightning-AI/lightning/pull/15309 is merged, can we try again @Borda?

```suggestion
def test_fit_loop_done_log_messages(caplog):
```

tests/tests_pytorch/loops/test_training_loop.py
awaelchli:
```suggestion
```

awaelchli (approved):

awaelchli (approved):

carmocca:
> as we are going to drop support for PT 19 in https://github.com/Lightning-AI/lightning/pull/15347 I am dropping testing for oldes and monolitic pkg

I suggest we keep them in this PR in case this is merged and #15347 does not make it in time before the release (maybe because it gets blocked or PyTorch delays its release). So I would revert your latest commit

Borda:
> I suggest we keep them in this PR in case this is merged and #15347 does not make it in time before the release

so how about keeping them but not setting them as required for now and @justusschock is checking them...

carmocca:
Okay. Fine with me

PR #15110, opened 2022-10-12
Force NVML-based CUDA check in PyTorch 1.14+

awaelchli:
## What does this PR do?

Follow up to https://github.com/pytorch/pytorch/pull/85951 which landed in pytorch master and will appear in 1.14. 

cc @borda @justusschock @awaelchli @kaushikb11 @akihironitta @rohitgr7 @speediedan @carmocca 

## Before submitting

## PR review

## Did you have fun?


src/lightning_lite/accelerators/cuda.py
```diff
@@ -102,11 +102,12 @@ def num_cuda_devices() -> int:
     Unlike :func:`torch.cuda.device_count`, this function does its best not to create a CUDA context for fork support,
     if the platform allows it.
     """
-    if _TORCH_GREATER_EQUAL_1_13:
+    if _TORCH_GREATER_EQUAL_1_14:
+        # We set `PYTORCH_NVML_BASED_CUDA_CHECK=1` in pytorch_lightning.__init__.py
```
awaelchli, line 106:
```suggestion
        # We set `PYTORCH_NVML_BASED_CUDA_CHECK=1` in lightning_lite.__init__.py
```

src/lightning_lite/__init__.py
```diff
@@ -13,6 +14,12 @@
     _logger.propagate = False
 
 
+# In PyTorch 1.14+, setting this variable will force `torch.cuda.is_available()` and `torch.cuda.device_count()`
+# to use an NVML-based implementation that doesn't poison forks.
+# https://github.com/pytorch/pytorch/issues/83973
+os.environ["PYTORCH_NVML_BASED_CUDA_CHECK"] = "1"
```
carmocca, line 20: I would suggest moving this into the `cuda.py` file. It does get imported by default anyways.

I guess you chose to set it on import in case users make calls before the Trainer starts, additionally, I would also set it on setup and pop it on teardown.

awaelchli: I want this to happen before anything else runs in Lightning, so that we don't shoot ourselves in the foot in the future when we import something that checks `torch.cuda.*`. 

This is an env variable for PyTorch, we don't need to tear it down. And we also don't wan to tear it down between Trainer calls for example, because if you poison a fork after trainer.fit(), you won't be able to run trainer.test() afterwards (ddp_fork in a notebook)

carmocca:
Why is this still a draft? We want this for 1.8, correct?

carmocca:
Won't people that use PyTorch 1.13 and install PL 1.8 have issues with fork until we update the version check?

awaelchli:
Yes, I guess for the version check that's good to have.

carmocca (approved):
The changes LGTM. Failures look unrelated. Maybe Lite master is broken

justusschock (approved):

jzazo:
Will there be a way to disable this? When running a pytest test, this check sometimes takes ~9s, adding a lot of overhead to the test suite. I think there are situations when you know you don't want to run with GPU, and you would like this disabled. What do you think?

awaelchli:
@jzazo Which version of Lightning and which version of PyTorch is your pytest running with?

jzazo:
pytorch_lightning 3.8.6
pytest 7.2.0
torch 1.13.1

jzazo:
I am analyzing the time with `pyinstrument`. I was observing two days ago ~9s delay when the trainer was initialized, but I can no longer reproduce that number. I deleted and reinstalled the environment, maybe things changed. If I can reproduce it again I will let you know.

If you have comments, explain. Should it be really fast ~ms?

awaelchli:
I don't know why it should be slower by this much. Looking back, in 1.8.6 we are using this code to determine the device count:

https://github.com/Lightning-AI/lightning/blob/caa3329b79965f7712e49bd9f8933f7a4686c49b/src/lightning_lite/accelerators/cuda.py#L106-L112

Since you are using torch 1.13.1, we are falling back to the default torch implementation:

```py
    if _TORCH_GREATER_EQUAL_1_13:
```

If you are seeing this again, it could be worth trying to set `os.environ["PYTORCH_NVML_BASED_CUDA_CHECK"] = "0"` in a fixture in your test suite.

PR #15169, opened 2022-10-18
Expose sharding context manager in Lite

awaelchli:
## What does this PR do?

API not final. Suggestions welcome!

After #14992, the context manager for sharding model layers won't be applied automatically anymore when avoiding the subclassing/run method. As a follow-up, we expose the context manager so the user can apply it. Right now, only DeepSpeed uses it, but we will have FSDP soon too: #14898 

No.

## Before submitting

## PR review

## Did you have fun?


src/lightning_lite/lite.py
```diff
@@ -345,6 +345,24 @@ def all_gather(
     def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
         return self._strategy.broadcast(obj, src=src)
 
+    # TODO: Find intuitive name
+    @contextmanager
+    def create_sharded_model(self) -> Generator:
```
awaelchli, line 350: Many names are floating around, all of them are kinda meh.

```py
with sharding():

with model_parallel():

with create_sharded_model():

with sharded_model()
```

Looking for suggestions.

Borda: I like 4)

src/lightning_lite/strategies/deepspeed.py
```diff
@@ -340,8 +342,6 @@ def module_sharded_context(self) -> Generator[None, None, None]:
             model_parallel_context = deepspeed.zero.Init(
                 remote_device=self.remote_device, pin_memory=True, config_dict_or_path=self.config, dtype=dtype
             )
-        else:
```
carmocca, line 343: Same suggestion, if the else stays then we avoid creating the nullcontext

carmocca (approved):

src/lightning_lite/strategies/strategy.py
```diff
@@ -314,3 +304,18 @@ def no_backward_sync(self, module: Module) -> Generator:
 
         This is a context manager. It is only effective if it wraps a call to `.backward()`.
         """
+
+
+class _Sharded(ABC):
```
justusschock, line 309: can we make this a protocol instead? Then strategy authors don't need to inherit from a private baseclass but instancechecks would still work within our codebase.

tests/tests_lite/test_lite.py
```diff
@@ -541,3 +542,18 @@ def run(self):
         TypeError, match=escape("Overriding `LightningLite.run()` and launching from the CLI is not allowed")
     ):
         LiteWithRun()
+
+
+def test_module_sharding_context():
```
justusschock, line 547: if we decide to make `_Sharded` a protocol, can we add some tests that our sharded strategies actually pass the instance tests of this protocol (once added and use a dummy class in the meantime)?

justusschock (approved):
LGTM in general. One comment on polymorphism in implementation

src/lightning_lite/strategies/strategy.py
awaelchli: My intention here was to make it very clear what the different interfaces are that a strategy exposes. Today, it is just "a bunch of methods that you can call". With the protocol, it is not as clear, because the interface wouldn't be listed as parent in the class definition. 

Also, this is just protected here for now. We are the only strategy authors at the moment. If I had made it public, someone would have requested to make it not public.

justusschock: To me it's less about public or private.

To me this sounds like you want to check if a strategy implements a certain thing and this is what Protocols are for in python. 

With abstract classes, you force strategies to implement this as soon as they inherit from it (which also is a valid approach).

So I guess it comes down to the following statement:
Personally, I don't like multiple baseclasses as long as there is an alternative (which here we have, as  it is just about API definition/check and not about actually providing functionality).

I am fine either way (which is why I also approved), that's just a preference. So feel free to resolve it if that's the only blocker left :)

awaelchli: I agree that multiple inheritance is not the right choice in many cases, and can often be avoided for a better design. In this case, I believe it does have a valid and strong use case. The difference to "just adding a base class" is that here it is specifically purely an interface (all methods are abstract). No implementation is provided. I would definitely not do this if I had to provide an implementation to the `_Sharded` method(s).

> With abstract classes, you force strategies to implement this as soon as they inherit from it (which also is a valid approach).

This is 100% what I was going for. My requirements that guided the choice here were:
- We want to go towards flattening the strategy inheritance hierarchy
- When flattening, avoid adding methods to the base strategy class while they only apply to certain strategy types
- Provide better documentation which methods are required to be implemented and which ones are optional.

I believe only the current approach can satisfy all these requirements.

Does @carmocca have any strong opinions about this? Otherwise I would keep it.

carmocca: I mostly agree with Adrian. It's true we could go the protocol way here and things would work, but why would we when we control all these implementations. So far we've only used protocols to fill the gaps in the designs of PyTorch, which we don't control

PR #15173, opened 2022-10-18
Check if GPU isn't being used by another process

mogwai:
## What does this PR do?

Adds a small check to see if the GPU isn't being used by any other torch process. 

This allows Trainer.auto_select_gpus to work as expected.

Suprised no one caught this :rofl: 

No

## Before submitting

### DIFF
--- a/src/pytorch_lightning/CHANGELOG.md
+++ b/src/pytorch_lightning/CHANGELOG.md
@@ -35,7 +35,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a more descriptive error message when attempting to fork processes with pre-initialized CUDA context ([#14709](https://github.com/Lightning-AI/lightning/pull/14709))
 - Added support for custom parameters in subclasses of `SaveConfigCallback` ([#14998](https://github.com/Lightning-AI/lightning/pull/14998))
 - Added `inference_mode` flag to Trainer to let users enable/disable inference mode during evaluation ([#15034](https://github.com/Lightning-AI/lightning/pull/15034))
-- Added fix to auto_select_gpus to make it choose gpu's without any memory allocated to them. ([#15173](https://github.com/Lightning-AI/lightning/pull/15173))
 
 
 ### Changed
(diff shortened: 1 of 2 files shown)


src/pytorch_lightning/tuner/auto_gpu_select.py
```diff
@@ -55,6 +55,9 @@ def pick_single_gpu(exclude_gpus: List[int]) -> int:
     for i in range(num_cuda_devices()):
         if i in exclude_gpus:
             continue
+        # Check if the device has memory used by another torch process
+        if torch.cuda.memory_usage(f"cuda:{i}") > 0:
+            continue
```
awaelchli, line 60: This check is redundant if the GPU is in exclusive mode, as only one process is allowed to run on it. That's why our check for placing a tensor on the device is enough to test for exclusivity. As the docs for memory_usage say, the check there is based on some sampling of ~1sec which is not fully reliable but we could probably still add it here. 

In addition, since you have been using/wanting to use this feature, would you mind checking out #13079 and comment on it with your thoughts? Thanks!

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a more descriptive error message when attempting to fork processes with pre-initialized CUDA context ([#14709](https://github.com/Lightning-AI/lightning/pull/14709))
 - Added support for custom parameters in subclasses of `SaveConfigCallback` ([#14998](https://github.com/Lightning-AI/lightning/pull/14998))
 - Added `inference_mode` flag to Trainer to let users enable/disable inference mode during evaluation ([#15034](https://github.com/Lightning-AI/lightning/pull/15034))
+- Added fix to auto_select_gpus to make it choose gpu's without any memory allocated to them. ([#15173](https://github.com/Lightning-AI/lightning/pull/15173))
```
awaelchli, line 38: Let's move this to the "Changed" section.

awaelchli (approved):
I would say that this is probably ok to add but also not ideal. 

There are definitely situations where we can easily trigger race conditions in which the check will not allocate the "expected" GPUs. One should not expect to be able to launch a batch of experiments at the exact same time and have them all allocate different GPUs "automatically". I hope this is clear to the user, and I hope we don't advertise this flag further as the automatic way of "allocating resources".

src/pytorch_lightning/tuner/auto_gpu_select.py
```diff
@@ -55,16 +55,19 @@ def pick_single_gpu(exclude_gpus: List[int]) -> int:
     for i in range(num_cuda_devices()):
         if i in exclude_gpus:
             continue
+        device = torch.device(f"cuda:{i}")
+        # Check if the device has memory used by another torch process
+        if torch.cuda.memory_usage(device) > 0 and torch.cuda.utilization(device):
```
justusschock, line 60: This won't work as expected. I just simulated this by having a first torch process creating a random tensor on gpu. In nvidia-smi you can clearly see the occupied memory.

When starting a second process though and retrieving these variables I get:

```
>>> import torch                                                                                                                                                                      │         
>>> torch.cuda.memory_usage('cuda:0')                                                                                                                                                 
0                                                                                                                                                                                     │         
>>> torch.cuda.memory_reserved('cuda:0')                                                                                                                                              
0
```
indicating that torch only returns the stats for the current process.

Furthermore this has an additional dependency on pynvml.

justusschock (changes requested):

src/pytorch_lightning/tuner/auto_gpu_select.py
mogwai: Hmm thanks @justusschock let me confirm this and try to figure out if there is a way we can do this with torch

mogwai: Yeah can confirm this. 

This very odd an not what you'd expect from torch?

PR #14967, opened 2022-10-01
FSDP (native) support for LightningLite

awaelchli:
## What does this PR do?

Adds support for `strategy="fsdp"` in LightningLite.

Fixes #14898

TODOs:
- [x] no sync after #14966  (see also [this](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.no_sync) and [this](https://github.com/pytorch/xla/blob/44104e7e8dd703b844ec16dd053520e24965e24d/torch_xla/distributed/fsdp/xla_fully_sharded_data_parallel.py#L667))

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @borda @carmocca @justusschock @awaelchli

PR #15185, opened 2022-10-19
Support individual setup of model and optimizer in Lite

awaelchli:
## What does this PR do?

This PR is providing new APIs for setting up model and optimizers to enable FSDP in #14967. 
The design considerations were discussed [here](https://www.notion.so/lightningai/Lightning-Lite-Setup-for-FSDP-1a70be1a258a4aae8d92bfbdb8a42a27).

**Before:**

Model and optimizers had to be set up jointly. 
```py
model = ...
optimizer = Adam(model.parameters())
model, optimizer = self.setup(model, optimizer)
```

**Now:**

Jointly setting up model and optimizer is still possible.
```py
model = ...
optimizer = Adam(model.parameters())
model, optimizer = self.setup(model, optimizer)
```
For some strategies, this is the only way possible (DeepSpeed).

However, there are strategies (FSDP) that require to set up the model first before the optimizer can be created:
```py
model = ...
model = self.setup_model(model)

optimizer = Adam(model.parameters())
optimizer = self.setup_optimizers(optimizer)
```

This PR introduces these two new APIs `setup_model` and `setup_optimizers` and the error messaging for which strategies support which setup procedure.

No.

## Before submitting

## PR review

## Did you have fun?


src/lightning_lite/lite.py
```diff
@@ -489,10 +554,37 @@ def _validate_setup(model: nn.Module, optimizers: Sequence[Optimizer]) -> None:
         if any(isinstance(opt, _LiteOptimizer) for opt in optimizers):
             raise ValueError("An optimizer should be passed only once to the `setup` method.")
 
+    def _validate_setup_model(self, model: nn.Module) -> None:
```
awaelchli, line 557: Error messaging here for built-in strategies provide guidance to users who need to change their code to call the setup methods in the right way.

src/lightning_lite/lite.py
```diff
@@ -159,19 +166,77 @@ def setup(
...
         self._models_setup += 1
+
         if optimizers:
-            # join both types in a list for API convenience
-            return [model] + optimizers
+            # join both types in a tuple for API convenience
+            return tuple((model, *optimizers))
+        return model
+
+    def setup_model(self, model: nn.Module, move_to_device: bool = True) -> _LiteModule:
```
carmocca, line 190: Since we are introducing these specific blocks, I think we should rename `def setup()` to `def setup_module_and_optimizers` to avoid confusion from users who don't know which one to use

src/lightning_lite/lite.py
```diff
@@ -159,19 +166,77 @@ def setup(
             model = self._move_model_to_device(model=model, optimizers=list(optimizers))
 
         # Let accelerator/plugin wrap and connect the models and optimizers
-        model, optimizers = self._strategy.setup_module_and_optimizers(model, list(optimizers))
+        if optimizers:
+            model, optimizers = self._strategy.setup_module_and_optimizers(  # type: ignore[assignment]
```
carmocca, line 170: This uses `module` but the user API is `setup_model`. Use only `module`?

src/lightning_lite/lite.py
```diff
@@ -159,19 +166,77 @@ def setup(
             model = self._move_model_to_device(model=model, optimizers=list(optimizers))
 
         # Let accelerator/plugin wrap and connect the models and optimizers
-        model, optimizers = self._strategy.setup_module_and_optimizers(model, list(optimizers))
+        if optimizers:
```
carmocca, line 169: If `setup` becomes `setup_module_and_optimizers`, then this `if` can be removed and we can reslve the typing mess

carmocca (commented):
Posting some basic design questions before doing the full review.
Since we are discarding polymorphism, I would not allow the option of `setup` vs the separate `setup_*` options. Having these redundant options is confusing for newcomers

src/lightning_lite/lite.py
awaelchli: No, that's too long. Just setup is ok. Since it does not specify anything in the name, the intuition is that it can setup multiple things which it does. And error messaging will guide the user eventually.

carmocca: @lantiga Can you clarify whether you meant that:

A) `setup()` kept as-is (polymorphic). We introduce `setup_model` and `setup_optimizer` for the cases where one needs to initialize them separately.
B) `setup` renamed to `setup_model_and_optimizer` (explicit) where both have to be passed. We introduce `setup_model` and `setup_optimizer` for the cases where one needs to initialize them separately.

This is in reference to your comment:

> I don’t like the polymorphic setup method. I think we should be explicit:
setup will work for all strategies that don’t need to wrap the submodules.

PR currently implements A). My question is if we should remove all polymorphism and do B)

awaelchli: A) should be the preferred way IMO, as this covers the vast majority of use cases. The name `setup_model_and_optimizer` is too long and not as easy to remember as just setup().

src/lightning_lite/lite.py
awaelchli: Internally in the strategy we want to use "module" to differentiate between the old way of referring to "model" as the wrapped LightningModule. But the user facing API in Lite shouldn't care about that, because it never had to care about this distinction.

carmocca: I would have them match to avoid confusion from those who snoop inside. But not a strong opinion.

Borda (approved):
1. The API looks good to me :otter: 
2. why are you updating `_notebooks` head?
3. shall we add also docs update for this?

awaelchli:
This PR is a preparation for #14898. I think it will make most sense to add the docs there for full context.

src/lightning_lite/lite.py
justusschock: I agree with @carmocca on this. Especially if lite is meant for people who want to read the code there shouldn't be confusion be these simple naming conventions from "legacy PL"

src/lightning_lite/lite.py
justusschock: I agree with @awaelchli here; it is just way more convenient to call `.setup`

justusschock (approved):

src/lightning_lite/lite.py
awaelchli: I updated the relevant parts that this PR touches in [8220c0e](https://github.com/Lightning-AI/lightning/pull/15185/commits/8220c0ed1ea30c736f8bd4b6a4290b3378898804), but not everything. I want to keep the PR as focused as possible so it does not distract from what it is meant to do. Happy to receive a PR that updates the rest of "model" occurrences.

src/lightning_lite/lite.py
```diff
@@ -517,10 +582,37 @@ def _validate_setup(model: nn.Module, optimizers: Sequence[Optimizer]) -> None:
         if any(isinstance(opt, _LiteOptimizer) for opt in optimizers):
             raise ValueError("An optimizer should be passed only once to the `setup` method.")
 
+    def _validate_setup_model(self, model: nn.Module) -> None:
+        if isinstance(model, _LiteModule):
+            raise ValueError("A model should be passed only once to the `setup_model` method.")
+
+        if isinstance(self._strategy, (DDPShardedStrategy, DDPSpawnShardedStrategy)):
```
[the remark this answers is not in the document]
awaelchli: It may seems so right now, but actually the _Sharded type does not determine whether a strategy supports being setup in a particular order. You see, there are two examples

- DDPShardedStrategy: Needs joint setup of optimizer and model
- FSDPStrategy: Needs setup model first before optimizer

both are of type _Sharded :)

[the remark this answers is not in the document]
justusschock: should we then have a flag on the strategy indicating that so that we don't have to maintain that list here?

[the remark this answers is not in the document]
awaelchli: For now this should be fine. I would avoid adding new attributes on the strategies unless absolutely necessary.

The strategies anyway need to handle this in the respective methods by producing an error if you call the wrong methods(s), e.g., FSDP gives an error if you call `strategy.setup_module_and_optimizer()`. But the error message will reference the API of the strategy itself. That's why I added custom error messages here in Lite so that we can reference the lite.x usage to better inform the user what they should do.

PR #15122, opened 2022-10-13
LAI [wip] 

Borda:
## What does this PR do?

Fixes #\<issue_number>

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [ ] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #15195, opened 2022-10-19
Resolve collectives test issues

carmocca:
## What does this PR do?

Part of #15122 

None

cc @carmocca @justusschock @awaelchli @borda @otaj


src/lightning_lite/plugins/collectives/torch_.py
```diff
@@ -100,6 +100,8 @@ def scatter_object_list(
         return scatter_object_output_list
 
     def barrier(self, device_ids: Optional[List[int]] = None) -> None:
+        if dist._rank_not_in_group(self.group):
+            return
```
carmocca, line 104: Without this, we would need to check the global rank before calling `barrier` as it was done in https://github.com/Lightning-AI/lightning/pull/15195/commits/baa6f91fa8758e23ea4f3e70d12917d0028d5de7 to avoid the warning:
```python
UserWarning: Running barrier on global rank 0 which does not belong to the given group.
```

src/lightning_lite/plugins/collectives/torch_.py
```diff
@@ -133,6 +135,12 @@ def setup(
 
     def teardown(self) -> Self:  # type: ignore[valid-type]
         super().teardown()
+        default_group = dist.GroupMember.WORLD
+        if default_group is not None:  # ensures it hasn't been destroyed already
+            group_map = dist.distributed_c10d._pg_map
+            if len(group_map) == 1 and default_group in group_map:  # ensures that only the default group is left
+                self.destroy_group(default_group)
```
carmocca, line 142: This will destroy the default process group (created internally by `init_process_group`) only if it hasn't been destroyed already and it's the last group leftover

src/lightning_lite/plugins/collectives/torch_.py
```diff
@@ -100,6 +100,8 @@ def scatter_object_list(
         return scatter_object_output_list
 
     def barrier(self, device_ids: Optional[List[int]] = None) -> None:
+        if dist._rank_not_in_group(self.group):
```
carmocca, line 103: If we want to avoid protected API:

```suggestion
        if group == dist.GroupMember.NON_GROUP_MEMBER
```

src/lightning_lite/plugins/collectives/torch_.py
```diff
@@ -133,6 +135,12 @@ def setup(
 
     def teardown(self) -> Self:  # type: ignore[valid-type]
         super().teardown()
+        default_group = dist.GroupMember.WORLD
```
carmocca, line 138: FIXME: this should also check if we were the ones to set up the default group

src/lightning_lite/plugins/collectives/torch_.py
justusschock: I think this is a general issue for any compute if collectives are called on the wrong rank and adding this check can't hurt at all, right? It would make it a lot easier that user's don't have to pay attention on which rank to call which collectives but to always call them on any rank and it doesn't do wrong stuff (or anything at all) unless you're on a correct rank.

src/lightning_lite/plugins/collectives/torch_.py
justusschock: +1 for this

justusschock (approved):
Thanks a lot for looking into this! Any idea why it did not fail before?

otaj:
> Thanks a lot for looking into this! Any idea why it did not fail before?

One of my ideas is that in #15122 we tested it on python 3.8 and in normal CI, we test this on 3.10. Could be different enough

src/lightning_lite/plugins/collectives/torch_.py
otaj: Please!

src/lightning_lite/plugins/collectives/torch_.py
awaelchli: Let's not manipulate the default group. A collective operates only on one group and should not teardown another one. I have been puzzled from the beginning why we are managing the default group inside the collectives.

tests/tests_lite/plugins/collectives/test_torch_.py
```diff
@@ -226,18 +234,21 @@ def test_collectives_distributed(n):
...
+    with check_destroy_group():  # the fixture does not work inside the launched function, use it manually
+        left_collective.create_group(ranks=[0, 1])
+        right_collective.create_group(ranks=[1, 2])
 
-    if strategy.global_rank in (0, 1):
-        tensor = torch.tensor(strategy.global_rank)
-        left_collective.all_reduce(tensor)
-        assert tensor == 1
-    right_collective.barrier()
-    if right_collective.rank >= 0:
```
awaelchli, line 237: the problem was also this line, which should have been `rank in (1, 2)` as per my original demo script :)

tests/tests_lite/plugins/collectives/test_torch_.py
```diff
@@ -226,18 +234,21 @@ def test_collectives_distributed(n):
...
         tensor = torch.tensor(strategy.global_rank)
-        right_collective.all_reduce(tensor)
-        assert tensor == 3
+        if strategy.global_rank in (0, 1):
+            tensor = left_collective.all_reduce(tensor)
+            assert tensor == 1
+        right_collective.barrier()  # avoids deadlock for global rank 1
```
awaelchli, line 245: the comment here is correct. the reason behind it is explained here in the warning in the pytorch docs in case you were wondering:
https://pytorch.org/docs/stable/distributed.html#groups

src/lightning_lite/plugins/collectives/torch_.py
carmocca: I don't think it hurts for `barrier` specifically. But perhaps the warning is useful to uncover issues in the code for the other collectives.

Note that if we dont' have this, it still works correctly, it just raises the warning.

src/lightning_lite/plugins/collectives/torch_.py
carmocca: `collective.setup()` will call `torch.distributed.init_process_group()` which internally manipulates the default group.

This is why I added its destruction to `collective.teardown()`

The only alternative here is that this class doesn't call `init_process_group` for the user at all, then we would either create a separate utility/class for that or just call this boilerplate manually somewhere else

tests/tests_lite/plugins/collectives/test_torch_.py
carmocca: Thank you! That reference is very useful. Although it says it applies to NCCL, when this test uses Gloo. Maybe this limitation applies to both and the docs just don't mention it

src/lightning_lite/plugins/collectives/torch_.py
justusschock: I prefer to have it the current way.

Just for the record: We could also think about not destroying the default group ever as this would be handled automatically upon the process termination. 
And if we don't send anything over that connection, the only downside of having the group inplace would be an open socket on each process, right?

src/lightning_lite/plugins/collectives/torch_.py
justusschock: here it does, maybe somewhere else it doesn't... 

But I think that this would be more consistent in general anyways. Meaning that if we create something we also destroy it if no longer needed

awaelchli (approved):

PR #15208, opened 2022-10-20
[LAI] Rename instances of "lightning_app" and "pytorch_lightning"

otaj:
Part of #15122 

When we will be creating full package and replace all occurrences of strings `lightning_lite`, `lightning_app` and `pytorch_lightning`, we need to make sure we don't replace these occurrences within variable/function names as these would produce syntax errors

cc @justusschock @awaelchli @rohitgr7 @borda

### DIFF
--- a/.azure/app-cloud-e2e.yml
+++ b/.azure/app-cloud-e2e.yml
@@ -168,7 +168,7 @@ jobs:
       artifact: $(name)
 
     - bash: |
-        time python -c "from lightning.app import testing; testing.delete_cloud_lightningapps()"
+        time python -c "from lightning.app import testing; testing.delete_cloud_lightning_apps()"
       condition: always()
       env:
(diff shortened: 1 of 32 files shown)


Borda (changes requested):
This imply significant APPI change, anyway why this is needed when this is already solved in https://github.com/Lightning-AI/lightning/pull/15105/files#r1000350794

Borda (approved):

tchaton (approved):
LGTM !

manskx (approved):

manskx:
How we will deal with imports that contain either "lightning_app" or "pytorch_lightning" ?

Borda:
> How we will deal with imports that contain either "lightning_app" or "pytorch_lightning" ?

what do you mean? the imports for `lightning.XXX` shall be corrected in the following PR #15105

PR #15235, opened 2022-10-21
[App] Optimise Queues Calls

tchaton:
## What does this PR do?

Currently, the Boring App makes 1.2k queue calls per minute. This PR reduces this by some factors.

Here is the current repartition.

```python
ORCHESTRATOR_COPY_RESPONSE_root.dict.dst_w: 54
ORCHESTRATOR_COPY_REQUEST_root.dict.src_w: 182
API_RESPONSE_QUEUE: 131
ORCHESTRATOR_COPY_RESPONSE_root.dict.src_w: 54
ORCHESTRATOR_REQUEST_root.dict.dst_w: 54
API_DELTA_QUEUE: 57
DELTA_QUEUE: 57
ORCHESTRATOR_COPY_REQUEST_root.dict.dst_w: 182
API_STATE_PUBLISH_QUEUE: 131
ERROR_QUEUE: 58
ORCHESTRATOR_REQUEST_root.dict.src_w: 53
CALLER_QUEUE_root.dict.src_w: 183
```

Here is the new repartition:

TODO @hhsecond 

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/core/constants.py
```diff
@@ -35,7 +35,7 @@ def get_lightning_cloud_url() -> str:
 REDIS_QUEUES_READ_DEFAULT_TIMEOUT = 0.005
 
 HTTP_QUEUE_URL = os.getenv("LIGHTNING_HTTP_QUEUE_URL", "http://localhost:9801")
-HTTP_QUEUE_REFRESH_INTERVAL = float(os.getenv("LIGHTNING_HTTP_QUEUE_REFRESH_INTERVAL", "0.1"))
+HTTP_QUEUE_REFRESH_INTERVAL = float(os.getenv("LIGHTNING_HTTP_QUEUE_REFRESH_INTERVAL", "1"))
```
lantiga, line 38: Not clear to me what the practical implication of setting this to 1 is. 1 second is a noticeable lag so we should be careful here

src/lightning_app/storage/orchestrator.py
```diff
@@ -48,10 +48,10 @@ def __init__(
         self.response_queues = response_queues
         self.copy_request_queues = copy_request_queues
         self.copy_response_queues = copy_response_queues
-        self.waiting_for_response: Set[str] = set()
+        self.waiting_for_response: Dict[str, str] = {}
         self._validate_queues()
         self._exit_event = threading.Event()
-        self._sleep_time = 0.1
+        self._sleep_time = 5
```
lantiga, line 54: I think a sleep time of 5 seconds for storage updates is a bit too long. 2 seconds?

src/lightning_app/core/app.py
```diff
@@ -292,43 +292,39 @@ def _collect_deltas_from_ui_and_work_queues(self) -> t.List[t.Union[Delta, APIRe
...
-                )  # TODO: rename
-                if delta_from_api:
-                    if isinstance(delta_from_api, DeltaRequest):
-                        delta_from_api = delta_from_api.delta
-                    deltas.append(delta_from_api)
+            if self.delta_queue and keep_fetching:
+                # TODO: Fetch all available deltas at once to reduce queue calls.
+                delta: t.Optional[
+                    t.Union[DeltaRequest, APIRequest, CommandRequest, ComponentDelta]
+                ] = self.get_state_changed_from_queue(self.delta_queue)
```
lantiga, line 305: Here we are calling get from redis repeatedly: can’t we just call/expose `LRANGE` on redis so we can directly get all deltas currently on the queue in one call?
@hhsecond

src/lightning_app/core/app.py
```diff
@@ -292,43 +292,39 @@ def _collect_deltas_from_ui_and_work_queues(self) -> t.List[t.Union[Delta, APIRe
...
-                    if work:
-                        new_work_delta = _delta_to_app_state_delta(self.root, work, deepcopy(component_output.delta))
-                        deltas.append(new_work_delta)
-                else:
-                    should_get_component_output = False
+                    keep_fetching = False
 
-            # if both queues were found empties, should break the while loop.
-            if not should_get_delta_from_api and not should_get_component_output:
+            if keep_fetching:
```
lantiga, line 327: isn’t this 
```python
if not keep_fetching:
    break
```
?

lantiga (changes requested):
A few points, one for @hhsecond too

src/lightning_app/core/app.py
hhsecond: That would break the semantics of queue's `push`/`pop` behaviour but as far as we know that's the contract, it should be alright!

src/lightning_app/core/app.py
hhsecond: or just break from the above line where we set `keep_fetching = False`

src/lightning_app/core/app.py
hhsecond: but probably tricky because it's two different data structures - go fetching an array of python picked binaries (go arrays) and then client is python. We could introduce some common interface in between, like protobuf, perhaps

src/lightning_app/core/constants.py
tchaton: Only the works with the `caller_queue` are waiting infinitely.  Delaying the work to start 1 later in the cloud should be fine. This would reduce drastically the number of calls.

src/lightning_app/storage/orchestrator.py
tchaton: Yes, I think both are fine. But moved to 2.

src/lightning_app/core/app.py
tchaton: Yes, this would be the best.

src/lightning_app/core/app.py
tchaton: I removed it completly.

src/lightning_app/core/app.py
lantiga: Ok after a quick huddle we decided to create an extra endpoint and send back the whole queue as a JSON payload of  Base64 encoded deltas.

tchaton: Yes, let's do this in another PR as this is related to adding a new method to the base abstraction.

lantiga (approved):
Looks good!

ethanwharris (approved):

hhsecond (approved):

src/lightning_app/core/app.py
```diff
@@ -298,44 +298,35 @@ def _collect_deltas_from_ui_and_work_queues(self) -> List[Union[Delta, APIReques
...
 
-        should_get_delta_from_api = True
-        should_get_component_output = True
         deltas = []
+        api_or_command_request_deltas = []
         t0 = time()
 
         while (time() - t0) < self.state_accumulate_wait:
 
-            if self.api_delta_queue and should_get_delta_from_api:
```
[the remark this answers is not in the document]
tchaton: I merged both queues into a single one. Same behaviour, less queue calls.

src/lightning_app/storage/orchestrator.py
```diff
@@ -151,21 +152,26 @@ def run_once(self, work_name: str) -> None:
...
-            )
-            destination = response.destination
-            assert response.source == work_name
-            response_queue = self.response_queues[destination]
-            response_queue.put(response)
-            # the request has been processed, allow new requests to come in for the destination work
-            self.waiting_for_response.remove(destination)
+        # Check the current work is within the sources.
+        # It is possible to have multiple destination targeting
+        # the same source concurrently.
```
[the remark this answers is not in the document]
tchaton: I am currently storing `self.waiting_for_response[work_name] = request.source`.

Therefore, if you have 2 destinations work targetting the same sources, `self.waiting_for_response.values()` would contain twice that source. 

This is what I meant to say there. 

Does it make sense ?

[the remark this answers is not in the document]
awaelchli: How is it possible that there can be multiple requests from the same work. On the work side, we block when waiting for a response. 
I am not convinced the logic here is correct. There must be some misunderstanding.

awaelchli (approved):

PR #15237, opened 2022-10-21
Introduce checkpoint migration

awaelchli:
## What does this PR do?

Part of #9396
Continues https://github.com/Lightning-AI/lightning/pull/8558

- When Lightning loads checkpoints, the migration gets applied automatically. Instructions are printed how to upgrade the checkpoint files permanently.
- This PR does not add any new migrations. These will be added in follow up PRs. 
- The only migration function we have at the moment is the one that was in upgrade_checkpoint.py (refactored)
- A warning is raised when attempting to load a checkpoint that is newer than the current Lightning version. This is optional, but I added this to signal "migration" here means we only support going forward in time, not back. Please do not ask me to add support for going back in time.

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli @ananthsub @ninginthecloud @rohitgr7 @otaj


src/pytorch_lightning/utilities/migration/migrations.py
```diff
@@ -0,0 +1,69 @@
...
+    Commit: a5d1176
+    """
+    keys_mapping = {
+        "checkpoint_callback_best_model_score": (ModelCheckpoint, "best_model_score"),
+        "checkpoint_callback_best_model_path": (ModelCheckpoint, "best_model_path"),
+        "checkpoint_callback_best": (ModelCheckpoint, "best_model_score"),
+        "early_stop_callback_wait": (EarlyStopping, "wait_count"),
+        "early_stop_callback_patience": (EarlyStopping, "patience"),
+    }
+    checkpoint["callbacks"] = checkpoint.get("callbacks") or {}
```
carmocca, line 60: Should this add the key if it didn't exist and none of the key mappings apply?

tests/tests_pytorch/utilities/test_upgrade_checkpoint.py
```diff
@@ -42,8 +41,8 @@
...
 def test_upgrade_checkpoint(tmpdir, old_checkpoint, new_checkpoint):
-    filepath = os.path.join(tmpdir, "model.ckpt")
-    torch.save(old_checkpoint, filepath)
-    upgrade_checkpoint(filepath)
-    updated_checkpoint = torch.load(filepath)
+    _set_version(old_checkpoint, "0.9.0")
+    _set_version(new_checkpoint, pl.__version__)
+    updated_checkpoint, _ = migrate_checkpoint(old_checkpoint)
     assert updated_checkpoint == new_checkpoint
+    assert _get_version(updated_checkpoint) == pl.__version__
```
carmocca, line 48: Can you also assert `old_checkpoint` to ensure it's modified by reference?

src/pytorch_lightning/utilities/migration/utils.py
```diff
@@ -0,0 +1,133 @@
...
+        exc_type: Optional[Type[BaseException]],
+        exc_value: Optional[BaseException],
+        exc_traceback: Optional[TracebackType],
+    ) -> None:
+        if hasattr(pl.utilities.argparse, "_gpus_arg_default"):
+            delattr(pl.utilities.argparse, "_gpus_arg_default")
+        del sys.modules["pytorch_lightning.utilities.argparse_utils"]
+
+
+def _pl_migrate_checkpoint(checkpoint: _CHECKPOINT, checkpoint_path: _PATH) -> _CHECKPOINT:
```
carmocca, line 98: I'm a bit confused about the difference between this function and `migrate_checkpoint`. Can you clarify when which would be used? Could they be refactored so one calls the other?

carmocca (commented):
Overall LGTM!

src/pytorch_lightning/utilities/migration/migrations.py
```diff
@@ -0,0 +1,69 @@
...
+
+from typing import Any, Callable, Dict, List
+
+from pytorch_lightning.callbacks.early_stopping import EarlyStopping
+from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
+
+_CHECKPOINT = Dict[str, Any]
+
+
+def migration_index() -> Dict[str, List[Callable[[_CHECKPOINT], _CHECKPOINT]]]:
```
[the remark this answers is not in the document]
awaelchli: No. Can be protected. 

Forgot to mention, this is a function simply so that it can be at the top close to the description. If it was a constant, I'd have to put it at the bottom of the file.

src/pytorch_lightning/utilities/migration/migrations.py
```diff
@@ -0,0 +1,69 @@
...
+"""Contains migration functions to upgrade legacy checkpoints to the format of the current Lightning version.
+
+When Lightning loads a checkpoint, these migrations will be applied on the loaded checkpoint dictionary sequentially,
+see :func:`~pytorch_lightning.utilities.migration.utils.migrate_checkpoint`.
+
+How to add a new migration?
+
+1. Create a new function with a descriptive name and docstring that explains the details of this migration. Include
+   version information as well as the specific commit or PR where the breaking change happened.
+2. Add the function to the `migration_index()` below. The key in the index is the version of Lightning in which the
```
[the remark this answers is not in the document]
awaelchli: Yes. Should I explicitly state it here that it is for developers?

src/pytorch_lightning/utilities/migration/utils.py
awaelchli: `upgrade_checkpoint.py` (the cli) uses the main `migrate_checkpoint` function. 
PL Trainer uses the `_pl_migrate_checkpoint` function, because it logs a message that is specific to the automatic ugprade. We don't want this message to appear for the upgrade script.

awaelchli: > Could they be refactored so one calls the other?

It already is like this. The `_pl_migrate_checkpoint` calls `migrate_checkpoint`

src/pytorch_lightning/utilities/migration/migrations.py
awaelchli: It was like this already in the 2+ year old code. I think it is safe to just keep it like that.

src/pytorch_lightning/utilities/migration/utils.py
```diff
@@ -0,0 +1,133 @@
...
+    return checkpoint
+
+
+def _get_version(checkpoint: _CHECKPOINT) -> str:
+    """Get the version of a Lightning checkpoint."""
+    return checkpoint["pytorch-lightning_version"]
+
+
+def _set_version(checkpoint: _CHECKPOINT, version: str) -> None:
+    """Set the version of a Lightning checkpoint."""
```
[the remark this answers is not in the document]
awaelchli: Hmm, that's probably nice to have. What should we name it? `pytorch-lightning_version_before_upgrade`?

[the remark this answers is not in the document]
justusschock: +1 for this or it might be hard to reproduce issues with this. 

What about `legacy_pytorch-lightning_version` ? Slightly shorter

src/pytorch_lightning/utilities/migration/migrations.py
```diff
@@ -0,0 +1,69 @@
...
+
+from typing import Any, Callable, Dict, List
+
+from pytorch_lightning.callbacks.early_stopping import EarlyStopping
+from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
+
+_CHECKPOINT = Dict[str, Any]
+
+
+def _migration_index() -> Dict[str, List[Callable[[_CHECKPOINT], _CHECKPOINT]]]:
```
justusschock, line 40: can we actually make this a registry? If the user has a custom callback/strategy/plugin/any component returning a different state, it would be easier if they can just add it to the registry (only handling their stuff) and still call our utilities

awaelchli: Thanks for the suggestion. Until now, the users could do that in the on_load_checkpoint hooks. I would hold off on this feature for now until we know users actually need it. So far, we don't have any issues for it.

justusschock: probably because they just adopted their code internally to work with the new and the legacy way without actually upgrading the checkpoint and also because we did not provide a mechanism for this

src/pytorch_lightning/utilities/migration/utils.py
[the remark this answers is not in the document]
awaelchli: Done in 9ff93506efe304ccd1a02456ce244d822913ad3d

src/pytorch_lightning/utilities/migration/migrations.py
```diff
@@ -0,0 +1,69 @@
+# Copyright The PyTorch Lightning team.
```
Borda, line 1: shall this rather be:
`src/pytorch_lightning/utilities/migrations/migration.py`

awaelchli: I named it plural because this file is mean to hold all the migration functions that apply. No strong opinion, happy to change

src/pytorch_lightning/utilities/migration/migrations.py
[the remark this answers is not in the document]
carmocca: I would suggest so.

src/pytorch_lightning/core/saving.py
```diff
@@ -156,6 +158,11 @@ def _load_from_checkpoint(
     with pl_legacy_patch():
         checkpoint = pl_load(checkpoint_path, map_location=map_location)
 
+    # convert legacy checkpoints to the new format
+    checkpoint = _pl_migrate_checkpoint(
+        checkpoint, checkpoint_path=(checkpoint_path if isinstance(checkpoint_path, (str, Path)) else None)
```
carmocca, line 163: The isisntance is to pass None when `checkpoint_path` is an IO type? Shouldnt't that also be supported?

src/pytorch_lightning/utilities/migration/utils.py
```diff
@@ -0,0 +1,140 @@
...
+    for migration_version, migration_functions in index.items():
+        if not _should_upgrade(checkpoint, migration_version):
+            continue
+        for migration_function in migration_functions:
+            checkpoint = migration_function(checkpoint)
+
+        applied_migrations[migration_version] = [fn.__name__ for fn in migration_functions]
+
+    if ckpt_version != pl.__version__:
+        _set_legacy_version(checkpoint, _get_version(checkpoint))
```
carmocca, line 59:
```suggestion
        _set_legacy_version(checkpoint, ckpt_version)
```

src/pytorch_lightning/utilities/migration/utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def _set_version(checkpoint: _CHECKPOINT, version: str) -> None:
+    """Set the version of a Lightning checkpoint."""
+    checkpoint["pytorch-lightning_version"] = version
+
+
+def _set_legacy_version(checkpoint: _CHECKPOINT, version: str) -> None:
+    """Set the legacy version of a Lightning checkpoint."""
+    checkpoint["legacy_pytorch-lightning_version"] = version
```
carmocca, line 135: I think there should be logic that ensures this only gets written ever once, or that the version here matches the existing one (if exists)

carmocca (approved):

src/pytorch_lightning/core/saving.py
carmocca: Can't we say the same? We don't have IO specific logic, the annotation is there because it's in `torch.load`

awaelchli: But if checkpoint_path is bytes, we will dump a bunch of bytes into the info message. What is the usefulness of that?

carmocca: I see. Resolving

src/pytorch_lightning/utilities/migration/utils.py
awaelchli: Addressed. can you check the newly added test in 40353c3

carmocca: Thanks

PR #15213, opened 2022-10-20
CI: testing monolotic package

Borda:
## What does this PR do?

extending CI workflows to test particular domain even for monolithic package
Needs #15207
Closes #15095 
Closes #15209

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @carmocca @akihironitta @tchaton @rohitgr7

PR #15282, opened 2022-10-24
Remove examples and loggers from develop dependencies

carmocca:
## What does this PR do?

We've always mocked loggers in our CI, meaning that the dependencies weren't installed at all.

#15213 accidentally changed this to install `devel`, which includes loggers.

This PR reverts that by removing the loggers installation instead of reverting using `devel`.

We can remove the `loggers.txt` files because they are never installed in CI.

Fixes the current blocker in PL:

```python
E                   ValueError: path is on mount 'c:', start on mount 'D:'
```
seen in https://github.com/Lightning-AI/lightning/actions/runs/3314622131/jobs/5474147614

This fix is caused by an old version incompatibility with Windows. If we want to revert the existing decision of mocking loggers and instead install them, then we could just upgrade the minimum WandB version.

Removes the `pytorch_lightning[loggers]` installation option

cc @borda @carmocca @akihironitta @tchaton @rohitgr7 @justusschock


requirements/pytorch/devel.txt
```diff
@@ -4,11 +4,5 @@
 # install all extra dependencies for full package testing
 -r ./extra.txt
 
-# install all loggers for full package testing
```
Borda, line 7: can we rather have another file ci.txt with these simplified options?

requirements/pytorch/loggers.txt
```diff
@@ -1,8 +0,0 @@
-# NOTE: the upper bound for the package version is only set for CI stability, and it is dropped while installing this package
-#  in case you want to preserve/enforce restrictions on the latest compatible version, add "strict" as an in-line comment
-
-# all supported loggers
-neptune-client>=0.10.0, <0.16.10
```
Borda, line 5: so we never test loggers?

carmocca: We've never tested loggers

requirements/pytorch/devel.txt
carmocca: What would you call it?

requirements/pytorch/loggers.txt
Borda: Can we have some integration test for this or remove loggers completely from the codebase :(

requirements/pytorch/devel.txt
carmocca: And in any case, why have it if we don't use it?

And for users, why would any user want to install all loggers?

requirements/pytorch/loggers.txt
carmocca: We've been fine without installing the loggers in CI, the tests all use mocks. We don't need to remove them

tchaton (approved):
LGTM !

requirements/pytorch/loggers.txt
Borda: but we do not test it with real package, right so how we know it still works?

src/pytorch_lightning/__setup__.py
```diff
@@ -33,14 +33,13 @@ def _prepare_extras() -> Dict[str, Any]:
     extras = {
         # 'docs': load_requirements(file_name='docs.txt'),
         "examples": setup_tools.load_requirements(file_name="examples.txt", **common_args),
-        "loggers": setup_tools.load_requirements(file_name="loggers.txt", **common_args),
```
Borda, line 36: we shall keep this option

Borda (changes requested):

src/pytorch_lightning/__setup__.py
```diff
@@ -33,14 +33,13 @@ def _prepare_extras() -> Dict[str, Any]:
...
         "extra": setup_tools.load_requirements(file_name="extra.txt", **common_args),
         "strategies": setup_tools.load_requirements(file_name="strategies.txt", **common_args),
         "test": setup_tools.load_requirements(file_name="test.txt", **common_args),
     }
     for req in parse_requirements(extras["strategies"]):
         extras[req.key] = [str(req)]
-    extras["dev"] = extras["extra"] + extras["loggers"] + extras["test"]
+    extras["dev"] = extras["extra"] + extras["test"]
     extras["all"] = extras["dev"] + extras["examples"] + extras["strategies"]  # + extras['docs']
```
Borda, line 43:
```suggestion
    extras["all"] = extras["dev"]+ extras["loggers"] + extras["examples"] + extras["strategies"]  # + extras['docs']
```

src/pytorch_lightning/__setup__.py
carmocca: What for?

requirements/pytorch/loggers.txt
awaelchli: In Lightning, a logger is a wrapper around a third-party logger. Obviously it is not on us to test the third-party behaves correctly. However, it is our responsibility to test that our wrapper calls the third-party objects correctly. This is sufficient for the vast majority of testing cases.

src/pytorch_lightning/__setup__.py
carmocca: When would an user want this?

requirements/pytorch/loggers.txt
```diff
@@ -1,8 +1,5 @@
...
-comet-ml>=3.1.12, <3.31.16
-mlflow>=1.0.0, <1.29.0
-wandb>=0.10.22, <0.13.4
+# all supported loggers. we don't install them in CI but are here as a reference of the supported loggers
```
carmocca, line 1: I've added back this file for the doc's mocks and just as a brief reference of which loggers we support, but I don't think it makes sense to expose this as `pytorch_lightning[loggers]` if we don't use it since there's no reason for users to install all the loggers, normally the'd just install the one they use

src/pytorch_lightning/__setup__.py
carmocca: We don't need to because we don't do anything special other than instantiating our own Logger classes. There's no version that would be incompatible for that in our docs build

awaelchli (approved):

otaj (approved):

otaj:
Btw, I believe the problem which this PR is somewhat trying to solve is the fact we have very inconsistent naming between requirement files and `extra_requires` keys https://github.com/Lightning-AI/lightning/issues/15210

docs/source-pytorch/visualize/supported_exp_managers.rst
```diff
@@ -8,7 +8,8 @@ To use `Comet.ml <https://www.comet.ml/site/>`_ first install the comet package:
 
 Configure the logger and pass it to the :class:`~pytorch_lightning.trainer.trainer.Trainer`:
 
-.. code-block:: python
+.. testcode::
```
Borda, line 11: Why this when we are dropping loggers, so they are never installed?

Borda (approved):

docs/source-pytorch/visualize/supported_exp_managers.rst
carmocca: Docs build was failing. Making it a testcode means we can use skipif

requirements/pytorch/loggers.info
```diff
@@ -0,0 +1,5 @@
+# all supported loggers. this list is here as a reference, but they are not installed in CI
+neptune-client
+comet-ml
+mlflow>
+wandb>
```
otaj, line 5:
```suggestion
mlflow
wandb
```

requirements/pytorch/devel.txt
```diff
@@ -4,11 +4,5 @@
...
--r ./loggers.txt
-
 # extended list of dependencies for development and run lint and tests
 -r ./test.txt
-
-# install all extra dependencies for running examples
--r ./examples.txt
```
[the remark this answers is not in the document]
Borda: But `devel.txt` is meant for developers, not for CI; that we import it in CI is just our simplification, and no developer shall have limited experience just because of convenience... :)

docs/source-pytorch/visualize/supported_exp_managers.rst
Borda: so far I know `.. code-block:: python` are not tested...

carmocca: I think so. Maybe I misjudged where the error was coming from. If you want, I can revert the testcode in a follow-up

requirements/pytorch/devel.txt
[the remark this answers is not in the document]
carmocca: Our examples use many different requiremenets that are not included in this file (nvidia dali for example). Developers do not need to install all of them to develop PL, especially if they are untested.

PR #15301, opened 2022-10-25
Added a check to validate that wrapped FSDP models are used while initializing optimizers

### LINKED ISSUE #14900: Tensors not on the same device when using FSDP auto-wrapping
## Bug

When using the FSDP with defaults and auto-wrapping, there is an error in forward saying tensors and weights are not on the same device. 

### To Reproduce

```py

import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

def run():

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        accelerator='gpu',
        enable_model_summary=False,
        precision=16,
        strategy='fsdp_native'
    )


if __name__ == "__main__":
    run()

```

Error:

```
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cpu and cuda:0! (when checking argument for argument mat1 in method wrapper_addmm)
```

### Additional context

Reported by user kavya on [Slack](https://pytorch-lightning.slack.com/archives/CRBLFHY79/p1664201840388219). 
Comment:

>  however fairscale fsdp strategy seems to work fine with mixed precision. any additional steps necessary for fsdp_native with torch==1.12.1+cu113 and pytorch-lightning==1.7.7

rohitgr7:
## What does this PR do?

Closes #14900

## Before submitting

## PR review

## Did you have fun?

cc @borda


Borda (approved):

src/pytorch_lightning/strategies/fully_sharded_native.py
```diff
@@ -212,6 +212,12 @@ def _setup_model(self, model: torch.nn.Module) -> FullyShardedDataParallel:
             del self.kwargs["auto_wrap_policy"]
 
         log.detail(f"setting up FSDP model with device id: {self.root_device.index}, kwargs: {self.kwargs}")
+
+        rank_zero_info(
```
awaelchli, line 216: @rohitgr7 
One last todo I have in my PR here #14967 is to validate that the optimizer is referencing the weights of the wrapped model. Once I have that code, I will let you know. We will have to do it here too. I believe it is better to only warn if necessary. In fact, we should error out, because if the optimizer is referencing the unwrapped model weights, the .step() will not update the right weights.

rohitgr7: > referencing the weights of the wrapped model.

I checked it, for the unwrapped model you get `ValueError: optimizer got an empty parameter list` so not sure if it's possible to detect it.

awaelchli: Here is the code that I added in Lite FSDP (https://github.com/Lightning-AI/lightning/pull/14967):
Every optimizer that references parameters in a FSDP model must reference parameters of type FlatParameter. If that's not the case, the user must have messed up the order. 

 ```py
   def setup_optimizer(self, optimizer: Optimizer) -> Optimizer:
        """Set up an optimizer for a model wrapped with FSDP.
        This setup method doesn't modify the optimizer or wrap the optimizer. The only thing it currently does is
        verify that the optimizer was created after the model was wrapped with `:meth:setup_module` with a reference
        to the flattened parameters.
        """
        from torch.distributed.fsdp import FlatParameter

        num_groups = len(optimizer.param_groups)
        if num_groups > 1:
            raise ValueError(
                "An optimizer used with an FSDP model does not support multiple param groups."
                f" Found {num_groups} parameter groups."
            )

        if any(isinstance(param, FlatParameter) for param in optimizer.param_groups[0]["params"]):
            return optimizer

        raise ValueError(
            "The optimizer does not seem to reference any FSDP parameters. HINT: Make sure to create the optimizer"
            " after setting up the model."
        )
```

Comment (awaelchli) [reply src/pytorch_lightning/strategies/fully_sharded_native.py:216]: If you want, we can make the validation code a utility function that can be shared between Lite and PL

Comment (rohitgr7) [reply src/pytorch_lightning/strategies/fully_sharded_native.py:216]: in lite if you initialize optimizer with old model (unwrapped), it doesn't raise any error?

Comment (awaelchli) [reply src/pytorch_lightning/strategies/fully_sharded_native.py:216]: of course it will, that's the whole point of this message. see the test cases

Comment (rohitgr7) [reply src/pytorch_lightning/strategies/fully_sharded_native.py:216]: oh I see. I guess you are proposing an extension here to check whether optimizer is initialized with wrapped models. My motivation for the PR was to just let users know that since with unwrapped model an error is raised.

yes, we should use this common utility. We should do it for both native and fairscale. You can merge you PR first and then I'll continue here :)

qq: what if users use the unwrapped model and they get `ValueError: optimizer got an empty parameter list` with no clue what to do? Do you think we should keep this info message?

File: tests/tests_pytorch/strategies/test_ddp_fully_sharded_native.py [hunk 2]
```diff
@@ -18,46 +18,6 @@
     from torch.distributed.fsdp.wrap import wrap
 
 
-def custom_auto_wrap_policy(
```
rohitgr7, line 21: these are just moved.. utils first and then tests.

tests/tests_pytorch/strategies/test_ddp_fully_sharded_native.py
```diff
@@ -214,35 +248,15 @@ def test_fully_sharded_native_strategy_checkpoint_multi_gpus(tmpdir, model, stra
     _run_multiple_stages(trainer, model)
 
 
-def _run_multiple_stages(trainer, model, model_path: Optional[str] = None):
```
rohitgr7, line 217: these are just moved.. utils first and then tests.

tests/tests_pytorch/strategies/test_ddp_fully_sharded_with_full_state_dict.py
```diff
@@ -171,51 +216,6 @@ def test_fully_sharded_strategy_checkpoint_multi_gpus(tmpdir, model, strategy):
     _run_multiple_stages(trainer, model)
 
 
-def _assert_save_equality(trainer, ckpt_path, cls=TestFSDPModelManualWrapped):
```
rohitgr7, line 174: these are just moved.. utils first and then tests.

tests/tests_pytorch/strategies/test_ddp_fully_sharded_with_full_state_dict.py
```diff
@@ -254,3 +254,17 @@ def test_fsdp_rewrap_limitation(tmpdir):
 
     with pytest.raises(MisconfigurationException, match="Using the same instance of model .* not supported"):
         trainer.test(model)
+
+
+@RunIf(min_cuda_gpus=1, skip_windows=True, standalone=True, fairscale_fully_sharded=True)
+def test_invalid_parameters_in_optimizer(tmpdir):
```
rohitgr7, line 260: new test

tests/tests_pytorch/strategies/test_ddp_fully_sharded_native.py
```diff
@@ -214,35 +248,15 @@ def test_fully_sharded_native_strategy_checkpoint_multi_gpus(tmpdir, model, stra
...
-    trainer.predict(model)  # model is wrapped, will not call `configure_sharded_model`
-
-    # provide model path, will create a new unwrapped model and load and then call `configure_shared_model` to wrap
-    trainer.predict(ckpt_path=model_path)
-
-
-def _assert_save_equality(trainer, ckpt_path, cls=TestFSDPModel):
-    # Use FullySharded to get the state dict for the sake of comparison
-    model_state_dict = trainer.strategy.lightning_module_state_dict()
+@RunIf(min_cuda_gpus=1, skip_windows=True, standalone=True, min_torch="1.12")
```
rohitgr7, line 251: new test

awaelchli (approved):
Just the rank zero info message should be removed IMO, the rest LGTM.

src/pytorch_lightning/strategies/fully_sharded_native.py
```diff
@@ -215,6 +217,12 @@ def _setup_model(self, model: torch.nn.Module) -> FullyShardedDataParallel:
...
+            "When using PyTorch FSDP auto-wrap, make sure to initalize your model using trainer else"
+            " you will get an error.\ntorch.optim.Optimizer(self.trainer.model.parameters(), ...)"
+        )
```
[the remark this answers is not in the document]
rohitgr7: if users do
```py
    Adam(self.parameters())
    # or
    Adam(self.any_layer_defined_in_init.parameters())
```
they get `ValueError: optimizer got an empty parameter list`, which might confuse them, so I thought of keeping the message. Note that this is an info not a warning.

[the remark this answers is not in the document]
awaelchli: Then this needs more work. The point of adding the validation logic was to prevent this.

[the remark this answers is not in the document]
rohitgr7: looking forward to see how this can be done!

awaelchli:
I think the meta package or something in the CI using it is broken. Changes in lightning_lite do not get reflected in the PL installation, as we can see [here](https://github.com/Lightning-AI/lightning/actions/runs/3371618011/jobs/5594088502).

awaelchli:
Blocked by #15469

src/pytorch_lightning/strategies/fully_sharded_native.py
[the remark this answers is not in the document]
rohitgr7: clever! thank you. Added another commit to raise other kind of errors

src/pytorch_lightning/strategies/fully_sharded.py
```diff
@@ -191,16 +190,27 @@ def setup(self, trainer: "pl.Trainer") -> None:
...
         self.setup_precision_plugin()
 
+    def setup_optimizers(self, trainer: "pl.Trainer") -> None:
+        invalid_params_error = False
+        try:
+            super().setup_optimizers(trainer)
+        except ValueError as e:
+            if "optimizer got an empty parameter list" not in str(e):
+                raise
+            invalid_params_error = True
```
carmocca, line 200: You don't need this flag, the block below can be indented into this except block

And then use `raise ValueError(...) from e`

src/pytorch_lightning/strategies/fully_sharded_native.py
```diff
@@ -255,6 +257,22 @@ def setup(self, trainer: "pl.Trainer") -> None:
...
+    def setup_optimizers(self, trainer: "pl.Trainer") -> None:
+        invalid_params_error = False
+        try:
+            super().setup_optimizers(trainer)
+        except ValueError as e:
+            if "optimizer got an empty parameter list" not in str(e):
+                raise
+            invalid_params_error = True
+
+        if invalid_params_error or any(not _optimizer_has_flat_params(optimizer) for optimizer in self.optimizers):
```
carmocca, line 269: Same suggestion

src/pytorch_lightning/strategies/fully_sharded.py
rohitgr7: actually, we do.. even if the try block is successful, there's another condition where this error will be raised.

carmocca: Sorry, thought that had an `and`, not an `or`

carmocca (approved):

carmocca (commented):
There's a docs blocker which I believe is caused by #15489

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 -
 
 
+- Added a check to validate that wrapped FSDP models are used while initializing optimizers ([#15319](https://github.com/Lightning-AI/lightning/pull/15319))
```
akihironitta, line 20:
```suggestion
- Added a check to validate that wrapped FSDP models are used while initializing optimizers ([#15301](https://github.com/Lightning-AI/lightning/pull/15301))
```

akihironitta (approved):

PR #15253, opened 2022-10-22
Enhance `reduce_boolean_decision` to accommodate `any`-analogous semantics expected by `EarlyStopping` Callback

speediedan:
## What does this PR do?

Fixes #15252 

When ``EarlyStopping`` is used in a distributed context, early stopping conditions may be met in some processes before others (especially when the ``divergence_threshold`` is sufficiently low).

Per the described intended behavior of ``EarlyStopping``:
https://github.com/Lightning-AI/lightning/blob/be1eb5e86d07fe22b53a59184089aac569875117/src/pytorch_lightning/callbacks/early_stopping.py#L204-L205
 , all training processes should be stopped when an ``EarlyStopping`` threshold is reached in **any** process. The current behavior of ``reduce_boolean_decision`` is to only return ``True`` when **all** input process decisions are ``True``:
https://github.com/Lightning-AI/lightning/blob/be1eb5e86d07fe22b53a59184089aac569875117/src/lightning_lite/strategies/parallel.py#L88-L92

Though this issue can be avoided when logging the monitored metric with ``sync_dist=True``, since that configuration is not mandatory, ``reduce_boolean_decision`` should be adapted to behave as the ``EarlyStopping`` callback expects. 

This PR maintains the current ``reduce_boolean_decision`` behavior by default, but enhances the function to accommodate ``any``-analogous semantics as expected by the ``EarlyStopping`` callback. This PR also includes an additional test to validate the aforementioned new behavior resolves the issue described.

### Does your PR introduce any breaking changes? If yes, please list them.

None

## Before submitting

- [X] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [X] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [X] Did you make sure to **update the documentation** with your changes? (if necessary)
- [X] Did you write any **new necessary tests**? (not for typos and docs)
- [X] Did you verify new and **existing tests pass** locally with your changes?
- [X] Did you list all the **breaking changes** introduced by this pull request?
- [X] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
[...]

PR #15304, opened 2022-10-25
Support DDP with LRFinder

### LINKED ISSUE #14624: Automatic LR finder for DDP
## 🚀 Feature

Currently, `auto_lr_find` is only supported for a single GPU. It would be good to have this support for a DDP environment. This is mentioned in the Lightning docs [here](https://pytorch-lightning.readthedocs.io/en/stable/advanced/training_tricks.html?highlight=LEARNING%20RATE%20FINDER#learning-rate-finder).

### Motivation

The motivation is quite simple. A lot of people use multi-GPU environments to train their models. Having an automatic LR finder for this will help them.

### Pitch

None.

### Alternatives

None.

### Additional context

Discussed this issue already in the PL Slack channel with @rohitgr7, @krshrimali and @awaelchli.

rohitgr7:
## What does this PR do?

Fixes #14624

## Before submitting

## PR review

## Did you have fun?

cc @borda @akihironitta @rohitgr7


src/pytorch_lightning/tuner/lr_finder.py
```diff
@@ -218,6 +223,7 @@ def lr_find(
     # Save initial model, that is loaded after learning rate is found
     ckpt_path = os.path.join(trainer.default_root_dir, f".lr_find_{uuid.uuid4()}.ckpt")
     trainer.save_checkpoint(ckpt_path)
+    ckpt_path = trainer.strategy.broadcast(ckpt_path)
```
rohitgr7, line 226:
```suggestion
    ckpt_path = trainer.strategy.broadcast(ckpt_path)
    trainer.save_checkpoint(ckpt_path)
```

tests/tests_pytorch/tuner/test_lr_finder.py
```diff
@@ -434,3 +435,25 @@ def test_if_lr_finder_callback_already_configured():
...
+        default_root_dir=tmpdir,
+        max_epochs=1,
+        strategy="ddp",
+        devices=2,
+        accelerator="cpu",
+    )
+
+    trainer.tuner.lr_find(model, datamodule=dm, update_attr=True, num_training=20)
+    lr = trainer.lightning_module.lr
+    lr = trainer.strategy.broadcast(lr)
```
awaelchli, line 457: The tuner should guarantee that the learning rate suggestion (and all the picked learning rate through the scheduler) are the same on all ranks. So the test shouldn't broadcast this. Instead, it should all_gather and assert that all LR's are the same value.

tests/tests_pytorch/tuner/test_lr_finder.py
```diff
@@ -434,3 +435,25 @@ def test_if_lr_finder_callback_already_configured():
 
     with pytest.raises(MisconfigurationException, match="Trainer is already configured with a .* callback"):
         trainer.tune(model)
+
+
+def test_lr_finder_with_ddp(tmpdir):
```
awaelchli, line 440: this has to be made a standalone test.

awaelchli (commented):
Thanks for working on it. I know it is WIP, but wanted to give some feedback for the test. Hope it's useful.

tests/tests_pytorch/tuner/test_lr_finder.py
rohitgr7: they can't be the same since it is determined from the lr vs loss curve, and even though lr values will be the same, loss can't be on each GPU process.

awaelchli: Okay. And what about the learning rates at the intermediate steps? I couldn't find where we are synchronizing this.

src/pytorch_lightning/tuner/lr_finder.py
```diff
@@ -380,7 +388,7 @@ def on_train_batch_end(
         # Check if we diverging
         if self.early_stop_threshold is not None:
             if current_step > 1 and smoothed_loss > self.early_stop_threshold * self.best_loss:
-                trainer.fit_loop.max_steps = current_step  # stop signal
+                trainer.should_stop = True  # stop signal
```
awaelchli, line 391: We need some mechanism that all ranks make this decision simultaneously. Either guarantee that the if condition is always the same in all ranks, or reduce the boolean decision across all ranks. I couldn't see where we are doing this. See also #15253

tests/tests_pytorch/tuner/test_lr_finder.py
rohitgr7: learning rate at intermediate steps will be the same.. only losses will be different.

src/pytorch_lightning/tuner/lr_finder.py
rohitgr7: good point!
since we rely on lr of rank zero.. let's broadcast?
I am not sure what's the best way here. If we do all reduce, then it's not good to just broadcast the lr at the end of lr_Find, we should reduce that well too.

tests/tests_pytorch/tuner/test_lr_finder.py
awaelchli: sorry, the learning rates will be the same because of the scheduler, I agree. What I meant was that the tracked losses will be different and thus also the final suggested learning rate.

src/pytorch_lightning/tuner/lr_finder.py
awaelchli: Yean, I think doing a broadcast here is definitely one solution you could do. For now, that should be ok. The tuner is anyway only a heuristic, we don't need to be super sophisticated here with the stopping criterion. 


FYI:

> If we do all reduce, then it's not good to just broadcast the lr at the end of lr_Find, we should reduce that well too.

That wouldn't be the same. Reducing the boolean decision here means either
- set stop True if any of the ranks want to stop
- set stop True if all of the ranks want to stop

reduce_boolean_decision does the latter, but after https://github.com/Lightning-AI/lightning/pull/15253 we could have the former too.

rohitgr7: then let's do broadcast when any of the ranks stop and broadcast it's learning rate

docs/source-pytorch/advanced/training_tricks.rst
```diff
@@ -270,7 +270,12 @@ initial learning rate.
 .. warning::
 
     For the moment, this feature only works with models having a single optimizer.
-    LR Finder support for DDP and any of its variations is not implemented yet. It is coming soon.
+
+
+.. note::
+
+    With DDP since each process runs in isolation, which ever process gets the suggest learning rate first, that learning rate
+    will be broadcasted to all other processes.
```
awaelchli, line 278:
```suggestion
    With DDP: Since each process runs in isolation, whichever process gets the suggest learning rate first, that learning rate
    will be broadcasted to all other processes.
```

src/pytorch_lightning/tuner/lr_finder.py
```diff
@@ -380,10 +392,20 @@ def on_train_batch_end(
...
                 if self.progress_bar:
                     self.progress_bar.close()
 
+        if trainer.should_stop:
```
awaelchli, line 399: can you add a todo that this should be replaced with reduce_boolean_decision (any) after #15253 lands?

src/pytorch_lightning/tuner/lr_finder.py
```diff
@@ -380,10 +392,20 @@ def on_train_batch_end(
...
+            self.should_stop_rank = int(
+                trainer.strategy.reduce(
+                    torch.tensor(trainer.global_rank, device=pl_module.device), reduce_op="max"
+                ).item()
```
awaelchli, line 403: I don't understand how this can work. The result of this will always be the "highest" rank: `max(0, 1, 2, ...., n) = n`

Let's simplify this (see also my comment above about broadcast):

I suggest to reduce using sum. Then, if any item is 1 (should_stop), the sum will be > 0. And you won't need the extra broadcast below because every rank already will have the same number. You can just set 

trainer.should_stop = ...reduce(...).item() > 0

Then, when stopped, you broadcast the "results" from rank 0 to all others. This should be good.

src/pytorch_lightning/tuner/lr_finder.py
```diff
@@ -345,6 +356,7 @@ def __init__(
         self.best_loss = 0.0
         self.progress_bar_refresh_rate = progress_bar_refresh_rate
         self.progress_bar = None
+        self.should_stop_rank = 0
```
awaelchli, line 359: I vote for not doing it this way. This makes it complicated where I believe it doesn't need to be. At any point, an arbitrary number of processes can decide to "should_stop=True". The choice here to choose "one of them" is as arbitrary as choosing just rank 0. We should just let rank 0 return the results IMO. See again my suggestion here: https://github.com/Lightning-AI/lightning/pull/15304/files#r1009288887

src/pytorch_lightning/tuner/lr_finder.py
rohitgr7: I am trying to broadcast the rank which is triggering the should_stop. If rank 1 triggers should_stop, rank 1 results should be broadcasted. If rank 1 and 2 triggers should_stop together (there's a barrier), we populate results from any of those ranks.

src/pytorch_lightning/tuner/lr_finder.py
rohitgr7: the possibility is that rank 0 might not have good results to generate a relevant lr suggestion.

awaelchli: We don't care about that. User needs to "tune" the tuner anyway, as this can happen on a single-device configuration too!

docs/source-pytorch/advanced/training_tricks.rst
```diff
@@ -270,7 +270,13 @@ initial learning rate.
...
 
     For the moment, this feature only works with models having a single optimizer.
-    LR Finder support for DDP and any of its variations is not implemented yet. It is coming soon.
+
+
+.. note::
+
+    With DDP: Since all the processes run in isolation, only process with ``global_rank=0`` will make the decision to stop the
+    learning rate finder and broadcast its results to all other ranks. That means, at the end of LR finder, each process will be running with
+    the learning rate found on ``global_rank==0``.
```
awaelchli, line 279:
```suggestion
    the learning rate found on ``global_rank=0``.
```

awaelchli (approved):
Now this looks pretty good! Simple, effective, and easy to reason about!

Borda (approved):

PR #15309, opened 2022-10-25
add cloudio pickle patching for unified package

justusschock:
## What does this PR do?
Part of #15105 

In legacy versions of PL we pickled callbacks for checkpointing. Loading these checkpoints now results in importing PL even for the unified package (or failing if PL isn't present).

This unpickler patch redirects all imports from PL to the `lightning.pytorch` package if necessary.

Fixes #\<issue_number>

None

## Before submitting

## PR review

## Did you have fun?

cc @borda


justusschock:
should be ready for review now

justusschock:
@carmocca the cleaning of sys modules we talked about caused a lot of other tests to crash. So this is the solution I came up with instead.

tests/tests_pytorch/graveyard/test_cloud_io.py
```diff
@@ -0,0 +1,69 @@
...
+LEGACY_CHECKPOINTS_PATH = os.path.join(_PATH_LEGACY, "checkpoints")
+CHECKPOINT_EXTENSION = ".ckpt"
+# load list of all back compatible versions
+with open(os.path.join(_PROJECT_ROOT, "legacy", "back-compatible-versions.txt")) as fp:
+    LEGACY_BACK_COMPATIBLE_PL_VERSIONS = [ln.strip() for ln in fp.readlines()]
+
+
+@pytest.mark.parametrize("pl_version", LEGACY_BACK_COMPATIBLE_PL_VERSIONS)
+@pytest.mark.skipif(
+    not "pytorch_" + "lightning" in sys.modules, reason="This test is only relevant for the standalone package"
```
awaelchli, line 21: the string addition can be collapsed in a single string, right?

src/pytorch_lightning/_graveyard/cloud_io.py
```diff
@@ -0,0 +1,23 @@
...
+        if module.startswith("pytorch_" + "lightning"):
+            # for the standalone package this won't do anything,
+            # for the unified mirror package it will redirect the imports
+
+            old_module = deepcopy(module)
+            module = "pytorch_lightning" + module[len("pytorch_" + "lightning") :]
+
+            # this warning won't trigger for standalone as these imports are identical
+            if module != old_module:
+                warnings.warn(f"Redirecting import of {module}.{name} to pytorch_lightning.{name}")
```
awaelchli, line 18: What is the motivation for showing this to the user? Maybe we could say what they should do to avoid the warning?

awaelchli (approved):
Thanks looking into this probably very hard to debug issue! 
I hope we're not looking into the entrance of a rabbit hole 😅

src/pytorch_lightning/_graveyard/cloud_io.py
Borda: note that in a mono-package it will be replaced to be
```
warnings.warn(f"Redirecting import of {module}.{name} to lightning.pytorch.{name}")
```

tests/tests_pytorch/graveyard/test_cloud_io.py
Borda: Yes and no, we want to preserve this even in the mono-package and the setup is running a naive replacement, so to prevent it we need to be creative...

Borda (approved):

tests/tests_pytorch/graveyard/test_cloud_io.py
justusschock: It could, but then this would be replaced when converting the tests for the unified mirror package which we don't want. So this is kind of hacky to prevent this

src/pytorch_lightning/_graveyard/cloud_io.py
justusschock: They cannot do much to prevent this. The issue is that we did pickle entire classes in old checkpoints which results in pytorch_lightning being imported (or failure to load the checkpoint at all if PL is not present standalone). This warning is just to be on the safe side, that users will be informed what we are doing. They also should only experience this while loading legacy checkpoints

justusschock:
@awaelchli I would like to wait for @carmocca to review as I initially discussed this with him

Borda:
> @awaelchli I would like to wait for @carmocca to review as I initially discussed this with him

just for the record, if this works correctly we can revert 0fd1be6baef34c011357e02f170f26e78dbe87ef

src/pytorch_lightning/_graveyard/cloud_io.py
```diff
@@ -0,0 +1,23 @@
...
+from typing import Any
+
+
+class RedirectingUnpickler(pickle.Unpickler):
+    def find_class(self, module: str, name: str) -> Any:
+        if module.startswith("pytorch_" + "lightning"):
+            # for the standalone package this won't do anything,
+            # for the unified mirror package it will redirect the imports
+
+            old_module = deepcopy(module)
```
carmocca, line 13: Why are we deepcopying if it's a plain string?

src/pytorch_lightning/_graveyard/cloud_io.py
carmocca: Isn't the current way printing the same twice?

```suggestion
                warnings.warn(f"Redirecting import of {old_module}.{name} to {module}.{name}")
```

src/pytorch_lightning/_graveyard/cloud_io.py
```diff
@@ -0,0 +1,23 @@
...
+import warnings
+from copy import deepcopy
+from typing import Any
+
+
+class RedirectingUnpickler(pickle.Unpickler):
+    def find_class(self, module: str, name: str) -> Any:
+        if module.startswith("pytorch_" + "lightning"):
+            # for the standalone package this won't do anything,
+            # for the unified mirror package it will redirect the imports
```
carmocca, line 12: Have you considered cases where the user actually has both installed? Would there be any implications from this change?

Perhaps we should also not run this code if both are available

src/pytorch_lightning/_graveyard/__init__.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import pytorch_lightning._graveyard.callbacks
+import pytorch_lightning._graveyard.cloud_io
```
carmocca, line 16: Why did you choose `cloud_io`? This has nothing todo with our `src/lightning_lite/utilities/cloud_io.py` module, no?

I would suggest
```suggestion
import pytorch_lightning._graveyard.legacy_import_unpickler
```

Or something along those lines

(remember the test name too)

src/pytorch_lightning/_graveyard/cloud_io.py
```diff
@@ -0,0 +1,23 @@
...
+            module = "pytorch_lightning" + module[len("pytorch_" + "lightning") :]
+
+            # this warning won't trigger for standalone as these imports are identical
+            if module != old_module:
+                warnings.warn(f"Redirecting import of {module}.{name} to pytorch_lightning.{name}")
+
+        return super().find_class(module, name)
+
+
+pickle.Unpickler = RedirectingUnpickler  # type: ignore
```
carmocca, line 21: (not for this PR) I'm not sure the graveyard is the right place for this code.

Could it be added as a migration piece in the future? ref: #15237 

If we did, then maybe we can restore this `pickle.Unpickler` override back to the original

carmocca (approved):

src/pytorch_lightning/_graveyard/__init__.py
justusschock: Fine with me as well. I chose `cloud_io` since our basic load function is within the `cloud_io` package.

src/pytorch_lightning/_graveyard/cloud_io.py
justusschock: I haven't considered this case so far. But in that case it would depend which package is used to load the checkpoint. I.e when loaded with `lightning.pytorch` I'd argue that this should be executed but if you load with `pytorch_lightning` it should not. And this is also the way it behaves.

src/pytorch_lightning/_graveyard/cloud_io.py
justusschock: This is a leftover, where `module` was changed in place. Should be safe to delete now.

src/pytorch_lightning/_graveyard/cloud_io.py
justusschock: Exactly my thoughts. I didn't want to interfer with existing checkpoint PRs however. I chose graveyard because it is for legacy checkpoints (we don't support pickling anymore) and graveyard was where we had patching of modules anyways.

I'd prefer to move it to checkpointing after #15237 .

otaj (approved):
This is awesome, @justusschock, thanks for working on this!

tests/tests_pytorch/graveyard/test_legacy_import_unpickler.py
```diff
@@ -0,0 +1,66 @@
+import glob
+import os
+import sys
+
+import pytest
+import torch
+
+import pytorch_lightning  # noqa: F401
+from tests.tests_pytorch.helpers.utils import no_warning_call
```
otaj, line 9:
```suggestion
from tests_pytorch.helpers.utils import no_warning_call
```

tests/tests_pytorch/graveyard/test_legacy_import_unpickler.py
```diff
@@ -0,0 +1,66 @@
...
+def test_imports_standalone(pl_version: str):
+    assert any(
+        key.startswith("pytorch_" + "lightning") for key in sys.modules.keys()
+    ), "Imported PL, so it has to be in sys.modules"
+    path_legacy = os.path.join(LEGACY_CHECKPOINTS_PATH, pl_version)
+    path_ckpts = sorted(glob.glob(os.path.join(path_legacy, f"*{CHECKPOINT_EXTENSION}")))
+    assert path_ckpts, f'No checkpoints found in folder "{path_legacy}"'
+    path_ckpt = path_ckpts[-1]
+
+    with no_warning_call():
```
akihironitta, line 30: It's better to specify which warning exactly?
```suggestion
    with no_warning_call(match="Redirecting import of*"):
```

akihironitta (approved):

tchaton (approved):
LGTM !

carmocca:
Bad news. This is failing in #15105: https://github.com/Lightning-AI/lightning/actions/runs/3328686095/jobs/5504966109

justusschock:
@carmocca I'm on it!

Issue #11361, opened 2022-01-07
CUDA Out of memory error during checkpoint saving when using DDPShardedPlugin


villmow:
## 🐛 Bug

In a multigpu setup (8xA100) I try to use parameter sharding using the `DDPShardedPlugin`. However, it fails after validation when trying to save the checkpoint with the following stacktrace. When I start the same script with `DDPPlugin(find_unused_parameters=False)` it works like a charm and saves a checkpoint.

```
Traceback (most recent call last):                                                                                                                                                                                                                                                
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 685, in _call_and_handle_interrupt                                                                                                                  
    return trainer_fn(*args, **kwargs)                                                                                                                                                                                                                                            
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 777, in _fit_impl                                                                                                                                   
    self._run(model, ckpt_path=ckpt_path)                                                                                                                                                                                                                                         
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1199, in _run                                                                                                                                       
    self._dispatch()                                                                                                                                                                                                                                                              
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1279, in _dispatch                                                                                                                                  
    self.training_type_plugin.start_training(self)                                                                                                                                                                                                                                
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/plugins/training_type/training_type_plugin.py", line 202, in start_training                                                                                                   
    self._results = trainer.run_stage()                                                                                                                                                                                                                                           
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1289, in run_stage                                                                                                                                  
    return self._run_train()                                                                                                                                                                                                                                                      
  File "/miniconda3/envs/cenv-39/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1319, in _run_train                                                                                                                                 
    self.fit_loop.run()                                                                                                                                                                                                                                                           
[...]

PR #15317, opened 2022-10-25
Add test to verify that lowering gpus on restart works with sharded spawn

### LINKED ISSUE #14485: Can sharded be resumed with a different number of devices?
## 🚀 Feature

Does this work?

```
model = BoringModel()
trainer = Trainer(strategy="ddp_sharded_spawn", fast_dev_run=True, gpus=2)
trainer.fit(model)

checkpoint_path = os.path.join(tmpdir, "model.pt")
trainer.save_checkpoint(checkpoint_path)

model = BoringModel()
trainer = Trainer(strategy="ddp_sharded_spawn", fast_dev_run=True, gpus=1)
trainer.fit(model, ckpt_path=checkpoint_path)
```

### Motivation

We had a legacy test skipped in our CI mentioning that this is unsupported.

### Pitch

We should check, and raise an appropriate error if that's the case.

### Additional context

https://github.com/Lightning-AI/lightning/pull/14476/files/38e10ba837dad423ecaa52f300609083de379e19#r960749044

rohitgr7:
## What does this PR do?

Fixes #14485

## Before submitting

## PR review

## Did you have fun?


tests/tests_pytorch/strategies/test_sharded_strategy.py
```diff
@@ -18,6 +20,37 @@
...
+            self._is_equal(actual, expected) for actual, expected in zip(self.state_dict(), self.old_model_state_dict)
+        )
+
+        for optimizer, state in zip(self.trainer.optimizers, self.old_optimizer_states):
+            optimizer.consolidate_state_dict()
+            self._is_equal(optimizer.state_dict(), state)
+
+    def _is_equal(self, a, b):
+        if isinstance(a, torch.Tensor):
+            return torch.all(torch.eq(a, b))
```
justusschock, line 46:
```suggestion
            return torch.allclose(a, b)
```

otherwise this might fail due to minor numeric inaccuracies. `allclose` == `eq` for integer tensors and `allclose` == `a - b < eps` for floating point :)

tests/tests_pytorch/strategies/test_sharded_strategy.py
```diff
@@ -18,6 +20,37 @@
...
+
+        if isinstance(a, Mapping):
```
justusschock, line 48: would it be easier to use `apply_to_collections` here with `torch.allclose` on every `Tensor`?

justusschock (approved):

tests/tests_pytorch/strategies/test_sharded_strategy.py
```diff
@@ -18,6 +20,37 @@
...
+        return optimizer
+
+
+class CheckModelRestore(ModelWithAdamOptimizer):
+    def __init__(self, old_model_state_dict, old_optimizer_states):
+        super().__init__()
+        self.old_model_state_dict = old_model_state_dict
+        self.old_optimizer_states = old_optimizer_states
+
+    def on_train_start(self):
```
awaelchli, line 35: We discussed this briefly in the past offline in general. 

We will have to provide a standardized way to do this in Lightning. Not necessarily automatically on resuming, but at least by providing utilities in some form to do this, not just for Zero1, but for all types of sharded state dicts and checkpoints. 

I will look into making a GitHub issue.

tests/tests_pytorch/strategies/test_sharded_strategy.py
```diff
@@ -18,6 +20,37 @@
...
+        assert all(
+            self._is_equal(actual, expected) for actual, expected in zip(self.state_dict(), self.old_model_state_dict)
+        )
+
+        for optimizer, state in zip(self.trainer.optimizers, self.old_optimizer_states):
+            optimizer.consolidate_state_dict()
+            self._is_equal(optimizer.state_dict(), state)
```
awaelchli, line 42: I do not understand why this is necessary. 

When we checkpoint the states, we already consolidate the state dict. When the state gets loaded for the single device, it already is consolidated. And optimizer.consolidate_state_dict should be a no-op anyway on single device. 

Please remind me where I am wrong here.

rohitgr7: looks like one can call `state_dict(all_ranks=True)` to do that.
https://github.com/facebookresearch/fairscale/blob/ee647b976cf4c8fdd37bc9ae3fd6331d225ba2a0/fairscale/optim/oss.py#L400-L404

since it relies on `self._all_states` with is an empty list by default.

tests/tests_pytorch/strategies/test_sharded_strategy.py
awaelchli: Apart from deepspeed, we don't support saving a shard of a checkpoint on each rank in any of the sharded strategyes. Issues like this one for example exist #11361. Once we support that, we will need to provide ways of resharding these checkpoints in an efficient way. 

Imagine you have a shard of N checkpoint files because you trained with N processes. Now you want to load these checkpoints into a trainer that runs with M != N processes.

The inefficient way we do it currently is to call consolidate_state_dict, but this will have to hold all shard in memory which is often not possible with the biggest models today.

tests/tests_pytorch/strategies/test_sharded_strategy.py
awaelchli: No, there must be a misunderstanding here. You set devices=1, there is only one rank. [consolidation](https://pytorch.org/docs/stable/distributed.optim.html#torch.distributed.optim.ZeroRedundancyOptimizer.consolidate_state_dict) is essentially a no-op here. Can you explain why you did it?

rohitgr7: yes, I understand that theoretically, consolidate call is not required here but that's not how fairscale is designed. it relies on `self._all_states` to generate the `state_dict`
https://github.com/facebookresearch/fairscale/blob/ee647b976cf4c8fdd37bc9ae3fd6331d225ba2a0/fairscale/optim/oss.py#L305

tests/tests_pytorch/strategies/test_sharded_strategy.py
rohitgr7: yep! agreed!

but is that blocking this PR? I understand we need to consolidate it first, so once we have a better way, we could update this test.

if you think we should close it, it's fine too :)

tests/tests_pytorch/strategies/test_sharded_strategy.py
```diff
@@ -321,3 +354,33 @@ def test_ddp_sharded_strategy_checkpoint_multi_gpu_fairscale_optimizer(tmpdir, s
     # Assert model parameters are identical after loading
     for trained_param, loaded_param in zip(model.parameters(), saved_model.parameters()):
         assert torch.equal(trained_param.to("cpu"), loaded_param)
+
+
+@RunIf(min_cuda_gpus=2, fairscale=True)
+def test_ddp_sharded_strategy_fit_ckpt_path_downsize_gpus(tmpdir):
```
awaelchli, line 360: Note: This test takes ~17 seconds to finish.

tests/tests_pytorch/strategies/test_sharded_strategy.py
awaelchli: No it's not blocking. It's just making aware that we are looking into options to generalize this and make it memory efficient.

tests/tests_pytorch/strategies/test_sharded_strategy.py
rohitgr7: oh! that's too much. wondering if we should create another one `slow-gpu` so these could run separately.

rohitgr7: cc: @carmocca @Borda

tests/tests_pytorch/strategies/test_sharded_strategy.py
rohitgr7: well, we'd still need to check other types which are not tensors and are part of this `Mapping`.

tests/tests_pytorch/strategies/test_sharded_strategy.py
carmocca: The current workflow could be ported to run on Azure with GPUs. This wasnt done in the past because the GPU bill was expensive, but now that it runs on our cluster it should be okay. I suggest you open an issue

rohitgr7: done #15372

Borda (approved):

carmocca (approved):

PR #15328, opened 2022-10-26
Parallelize package install job

carmocca:
## What does this PR do?

Part of #15105

None


Borda (changes requested):
no, you need to run in sequence as you need to upload and download all previous packages

UPDATE: note that this bottleneck is removed with transition to mirror package #15105

PR #15332, opened 2022-10-26
Create required group for app examples

carmocca:
## What does this PR do?

Revert https://github.com/Lightning-AI/lightning/pull/15324/files#r1005673357

None


Borda (dismissed):
not really, as some examples are used in testing `tests_app`, see for example `pytest_sessionstart` or `test_serve_gradio`

So this would be great to have, but we shall be fully decoupled first :otter:

.github/checkgroup.yml
```diff
@@ -233,27 +233,38 @@ subprojects:
...
-      - "examples/app_*"
+      - "examples/app_*"  # some tests_app tests call examples files
+      - ".github/workflows/ci-app-tests.yml"
+      - "setup.py"
       - ".actions/**"
     checks:
       - "App.cloud-e2e"
       - "make-doctest (app)"
       - "make-html (app)"
-      - "app-examples (macOS-11, app, 3.8, latest)"
```
Borda, line 244: but they shall stay here, as if you change source it may affect examples

carmocca: They are already in https://github.com/Lightning-AI/lightning/pull/15332/files#diff-5ca41bd8852d2c309825fa76269fb299e7a77b075918a9c56fa1990c40e48f7bR261-R266. If you modify an examples file, that group will cover it

otaj (approved):

akihironitta (approved):

.github/checkgroup.yml
otaj: @Borda are we good to resolve so that we can merge this?

Borda (approved):

PR #15336, opened 2022-10-26
Enforce PEP517 in CI

otaj:
We will enforce PEP517 in CI so that we don't have deprecation warnings during installation of dependencies


carmocca:
Can't we set a global option in `setup.cfg`/`pyproject.toml`,...?

otaj:
> Can't we set a global option in `setup.cfg`/`pyproject.toml`,...?

Don't know yet. Will investigate more.

justusschock:
closing this for now. If we find a global option we'll add that in another PR.

PR #15337, opened 2022-10-26
Replace oldest versions also for Lite

otaj:
In CI, we are not replacing oldest version for Lite. Therefore, we're not really testing Lite with oldest specified version.

cc @carmocca

### DIFF
--- a/.actions/assistant.py
+++ b/.actions/assistant.py
@@ -30,6 +30,10 @@ REQUIREMENT_FILES = {
         "requirements/app/ui.txt",
         "requirements/app/cloud.txt",
     ),
+    "lite": (
+        "requirements/lite/base.txt",
+        "requirements/lite/strategies.txt",
+    ),
 }
 REQUIREMENT_FILES_ALL = list(chain(*REQUIREMENT_FILES.values()))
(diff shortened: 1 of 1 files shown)


awaelchli (approved):

.actions/assistant.py
```diff
@@ -30,6 +30,10 @@
         "requirements/app/ui.txt",
         "requirements/app/cloud.txt",
     ),
+    "lite": (
+        "requirements/lite/base.txt",
+        "requirements/lite/strategies.txt",
```
carmocca, line 35: What would this do for strategies?

carmocca (approved):

.actions/assistant.py
otaj: `req = open(fname, encoding="utf-8").read().replace(">=", "==")` which is what `oldest` is supposed to do.

But it can only do so, if the requirement file is listed here.

Borda: some are used in CPU testing still, right?

Borda (approved):

Borda:
in fact, thinking this shall automatically parse all `.txt` files in requirements :)

Borda:
> Will you do it, @Borda, or should I?

pls do it :otter:

PR #15342, opened 2022-10-26
Create cloud docs

edenlightning:


docs/source-app/cloud/billing.rst
```diff
@@ -0,0 +1,33 @@
+.. _billing:
+
+#######
+Billing
+#######
+
+
+With Lightning.ai you payfor what you use! Lightning.ai Cloud is using a prepaid consumption billing model.
```
boodland, line 8:
```suggestion
With Lightning.ai you pay for what you use! Lightning.ai Cloud is using a prepaid consumption billing model.
```

boodland (commented):
Added a suggestion about at typo and a comment that is happening in other parts of the documentation not sure why we say we bill storage every hour unless is in our planing for the future, afaik we bill for quota requested if higher than 10GB

PR #15345, opened 2022-10-26
docker: drop pt 1.9

Borda:
## What does this PR do?

Seems that Apex does not support PT 1.9 anymore while installing from the master...
```
Warning: Torch did not find available GPUs on this system.
If your intention is to cross-compile, this is not an error.
By default, Apex will cross-compile for Pascal (compute capabilities 6.0, 6.1, 6.2),
Volta (compute capability 7.0), Turing (compute capability 7.5),
and, if the CUDA version is >= 11.0, Ampere (compute capability 8.0).
If you wish to cross-compile for a single specific architecture,
export TORCH_CUDA_ARCH_LIST="compute capability" before running setup.py.
```

Part of #15346 

## Before submitting

- [x ] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?

## PR review

## Did you have fun?


.github/workflows/ci-pytorch-dockers.yml
```diff
@@ -94,7 +94,6 @@ jobs:
         include:
           # These are the base images for PL release docker images,
           # so include at least all of the combinations in release-dockers.yml.
-          - {python_version: "3.9", pytorch_version: "1.9", cuda_version: "11.1.1"}
           - {python_version: "3.9", pytorch_version: "1.10", cuda_version: "11.3.1"}
           - {python_version: "3.9", pytorch_version: "1.11", cuda_version: "11.3.1"}
           - {python_version: "3.9", pytorch_version: "1.12", cuda_version: "11.6.1"}
```
carmocca, line 99: There's one more instance 2 lines below this comment

.github/checkgroup.yml
```diff
@@ -127,11 +127,9 @@ subprojects:
       - "setup.py"
       - ".actions/**"
     checks:
-      - "build-cuda (3.8, 1.9, 11.1.1)"
```
carmocca, line 130: Maybe replace these with 1.13 builds?

carmocca (approved):

.github/checkgroup.yml
carmocca: Feel free to ignore. Started #15347

.github/workflows/ci-pytorch-dockers.yml
```diff
@@ -94,12 +93,9 @@ jobs:
...
-          # Used in Lightning-AI/tutorials
-          - {python_version: "3.8", pytorch_version: "1.9", cuda_version: "11.1.1"}
```
otaj, line 102: Was the comment correct? Is it really used in tutorials? If so, shouldn't that be also updated?

.github/workflows/ci-pytorch-dockers.yml
```diff
@@ -34,7 +34,6 @@ jobs:
         include:
           # We only release one docker image per PyTorch version.
           # The matrix here is the same as the one in release-docker.yml.
-          - {python_version: "3.9", pytorch_version: "1.9", cuda_version: "11.1.1"}
```
otaj, line 37: Are we fine with dropping pre-built dockers for a currently supported combination of versions?

otaj:
I feel like the error message that is put in the top-level post doesn't really say what's the problem and that it is apex. It could very well be some misconfiguration on our end

.github/workflows/ci-pytorch-dockers.yml
akihironitta: Updated! https://github.com/Lightning-AI/tutorials/pull/203

akihironitta (approved):
LGTM, but I think we should wait until we confirm PyTorch 1.13 is released so that we don't breach the contract with the community: https://pytorch-lightning.readthedocs.io/en/latest/versioning.html

carmocca:
@akihironitta You are right, but these are just CI docker changes. What is the impact to the community?

carmocca:
This will require admin merge

akihironitta:
We release docker images for the community to use, so I think we should keep 1.9 until we have 1.13 even in docker CI/CD. The impact should be subtle (or even none if no one in the community actually uses the 1.9 docker image) but it technically means a breach of the contract. (Given that we've never seen anyone use the docker image (in issues/discussions/PRs), I'm also fine with merging this PR before having 1.13 in our CI/CD)

.github/workflows/ci-pytorch-dockers.yml
Borda: yes, because the only drawback is is that some dependencies are missing so they will be installed in CI, but that is fine...

Borda:
> I wouldn't say we release the docker images for the community, but for our CI use. Since they are public people are free to pull them, but we cannot guarantee their correctness or have deprecation windows etc.

exactly

.github/checkgroup.yml
```diff
@@ -141,18 +141,15 @@ subprojects:
...
       - "build-cuda (3.9, 1.11, 11.3.1)"
       - "build-cuda (3.9, 1.12, 11.6.1)"
-      - "build-cuda (3.9, 1.9, 11.1.1)"
       - "build-hpu (1.5.0, 1.11.0)"
       - "build-ipu (3.9, 1.9)"
       - "build-NGC"
       - "build-pl (3.9, 1.10, 11.3.1)"
       - "build-pl (3.9, 1.11, 11.3.1)"
       - "build-pl (3.9, 1.12, 11.6.1)"
-      - "build-pl (3.9, 1.9, 11.1.1)"
```
Borda, line 155: We could keep this as it is based on the existing image; just install PL inside...
anyway, we can drop it :otter:

akihironitta:
> And this PR only makes it so it doesn't get generated anymore right? It does not delete it from dockerhub

Yes, but if PL 1.8.0 release happens before having PyTorch 1.13 in our CI, there'll be only three PyTorch versions published to the hub. I thought docker images with `latest-*` tags etc. were for the community, but I maybe misunderstood that. Let's proceed forward with this then!

Borda:
@carmocca can you checks group check bc this need to force merge
cc: @lantiga

carmocca:
It's good to be merged! Thank you!

PR #15367, opened 2022-10-27
[App] Add env variables to desactivate pull and push of the App State

tchaton:
## What does this PR do?

This PR introduces 2 env variables to disable state manipulation through the Rest API:

```python
ENABLE_PULLING_STATE_ENDPOINT=0 # Used to disable getting the latest state
ENABLE_PUSHING_STATE_ENDPOINT=0  # Used to disable pushing deltas through the Rest API
```

Furthermore, this PR adds a layout endpoint to remove the need for the UI to fetch the AppState.

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/core/api.py
```diff
@@ -161,86 +167,114 @@ class StateUpdate(BaseModel):
...
+        if x_lightning_session_uuid is None:
+            raise Exception("Missing X-Lightning-Session-UUID header")
+        if x_lightning_session_id is None:
+            raise Exception("Missing X-Lightning-Session-ID header")
+
+        body: Dict = await request.json()
+        api_app_delta_queue.put(DeltaRequest(delta=Delta(body["delta"])))
+
+
+if ENABLE_PUSHING_STATE_ENDPOINT:
```
alecmerdler, line 246: Might be better to just raise `HTTP 405 Method Not Allowed` instead of removing the endpoint completely.

src/lightning_app/core/api.py
```diff
@@ -161,86 +167,114 @@ class StateUpdate(BaseModel):
...
-    x_lightning_type: Optional[str] = Header(None),
-    x_lightning_session_uuid: Optional[str] = Header(None),
-    x_lightning_session_id: Optional[str] = Header(None),
-) -> Mapping:
-    if x_lightning_session_uuid is None:
-        raise Exception("Missing X-Lightning-Session-UUID header")
-    if x_lightning_session_id is None:
-        raise Exception("Missing X-Lightning-Session-ID header")
-
+if ENABLE_PULLING_STATE_ENDPOINT:
```
alecmerdler, line 170: Might be better to just raise `HTTP 405 Method Not Allowed` instead of removing the endpoint completely.

src/lightning_app/core/api.py
```diff
@@ -161,86 +167,114 @@ class StateUpdate(BaseModel):
...
-    body: Dict = await request.json()
-    x_lightning_session_uuid = TEST_SESSION_UUID
+        layout = deepcopy(state["vars"]["_layout"])
+        for la in layout:
+            if la["content"].startswith("root."):
+                la["content"] = get_component_by_name(la["content"], state)
+        return layout
+
+
+if ENABLE_PULLING_STATE_ENDPOINT:
```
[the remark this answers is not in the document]
tchaton: Sounds good. I wanted to avoid folks discovering it is accessible. But I think it is fine.

src/lightning_app/core/flow.py
```diff
@@ -276,7 +276,11 @@ def state_with_changes(self):
     @property
     def flows(self):
         """Return its children LightningFlow."""
-        return {el: getattr(self, el) for el in sorted(self._flows)}
+        flows = {el: getattr(self, el) for el in sorted(self._flows)}
+        for struct_name in sorted(self._structures):
+            for flow in getattr(self, struct_name).flows:
+                flows[flow.name] = flow
+        return flows
```
manskx, line 283: This is a breaking change, everywhere the `.flows` is used it expects to get only direct children not to get nested flows in structures. maybe we can have it in a different method/property?

src/lightning_app/core/api.py
```diff
@@ -23,7 +23,13 @@
 
 from lightning_app.api.http_methods import HttpMethod
 from lightning_app.api.request_types import DeltaRequest
-from lightning_app.core.constants import CLOUD_QUEUE_TYPE, ENABLE_STATE_WEBSOCKET, FRONTEND_DIR
+from lightning_app.core.constants import (
+    CLOUD_QUEUE_TYPE,
+    ENABLE_PULLING_STATE_ENDPOINT,
+    ENABLE_PUSHING_STATE_ENDPOINT,
+    ENABLE_STATE_WEBSOCKET,
```
manskx, line 30: For the future, do you think we should leverage the `.lightning` file more to have all app-related configurations there? env variables should remain usable be can be overridden by app_config.

src/lightning_app/core/api.py
```diff
@@ -172,33 +179,71 @@ async def get_state(
...
+
+    if not ENABLE_PULLING_STATE_ENDPOINT:
+        response.status_code = status.HTTP_405_METHOD_NOT_ALLOWED
+        return {"status": "failure", "reason": "This endpoint is disabled."}
+
     global app_spec
     return app_spec or []
 
 
 @fastapi_service.post("/api/v1/delta")
```
manskx, line 239: Should `/api/v1/upload_file/{filename}` route be also disabled ?

src/lightning_app/core/api.py
```diff
@@ -172,33 +179,71 @@ async def get_state(
...
+        return {"status": "failure", "reason": "This endpoint is disabled."}
+
     with lock:
         x_lightning_session_uuid = TEST_SESSION_UUID
         state = global_app_state_store.get_app_state(x_lightning_session_uuid)
         global_app_state_store.set_served_state(x_lightning_session_uuid, state)
         return state
 
 
+def get_component_by_name(component_name: str, state):
```
manskx, line 193: Could you add a docstring of what this function does?

```suggestion
def _get_component_by_name(component_name: str, state):
```

src/lightning_app/core/api.py
lantiga: I actually think that some of these should eventually be part of the API (for instance as arguments to the App constructor, or arguments of the flow). Exposing the state or not is a concern of the app developer, rather than the one who runs the app IMO.

src/lightning_app/core/api.py
lantiga: +1

src/lightning_app/core/flow.py
lantiga: Although this is probably an unintended effect of the previous code. Semantically it was more a bug than a feature IMO.

lantiga (approved):
Good one, very much needed.
In the future we might want to get more granular and expose parts of the state, but this is a great start.

src/lightning_app/core/flow.py
tchaton: Yes. It was a bug I found while developing this feature.

src/lightning_app/core/api.py
tchaton: We can consider it later. But let's keep with env variables. It is more reversive than new arguments.

src/lightning_app/core/api.py
tchaton: Done ! I didn't do it at first as this endpoint would be removed as soon as upload pre-signed url is available.

src/lightning_app/core/api.py
tchaton: Done !

src/lightning_app/core/api.py
lantiga: Yes for sure, my comment was about future developments

lantiga (approved):

manskx (approved):

src/lightning_app/core/constants.py
```diff
@@ -63,3 +62,9 @@ def get_lightning_cloud_url() -> str:
 
 DEBUG: bool = lightning_cloud.env.DEBUG
 DEBUG_ENABLED = bool(int(os.getenv("LIGHTNING_DEBUG", "0")))
+ENABLE_PULLING_STATE_ENDPOINT = bool(int(os.getenv("ENABLE_PULLING_STATE_ENDPOINT", "1")))
+ENABLE_PUSHING_STATE_ENDPOINT = ENABLE_PULLING_STATE_ENDPOINT and bool(
+    int(os.getenv("ENABLE_PUSHING_STATE_ENDPOINT", "1"))
+)
+ENABLE_STATE_WEBSOCKET = bool(int(os.getenv("ENABLE_STATE_WEBSOCKET", "0")))
+ENABLE_UPLOAD_ENDPOINT = bool(int(os.getenv("ENABLE_UPLOAD_ENDPOINT", "1")))
```
awaelchli, line 70: As we are growing our catalog of environment variables, we should be thinking of prefixing them with `LIT_` or `LAI_` or similar. For debugging purposes (I can easily grep for the prefix) and to avoid accidental clashes (which could be incredibly difficult to debug).

tests/tests_app/conftest.py
```diff
@@ -46,6 +47,13 @@ def pytest_sessionfinish(session, exitstatus):
             continue
         child.kill()
 
+    main_thread = threading.current_thread()
+    for t in threading.enumerate():
+        if t is not main_thread:
+            t.join(0)
+
+    Popen(f"sudo py-spy top --pid {os.getpid()}", shell=True)
```
awaelchli, line 55: Could you please add a comment what this does? And what does this mean if I run it locally, it will ask for my password. And secondly, what if py-spy is not installed (I don't have it my my current lightning env). This looks like it will just silently fail, is this intentional?

awaelchli (approved):

tests/tests_app/conftest.py
tchaton: Removed.

src/lightning_app/core/constants.py
tchaton: Yes, let's organise them all in another PR.

PR #15374, opened 2022-10-27
PKG: distribute single semver

Borda:
## What does this PR do?

unify all pkg versions
in case you run from source without install provide commit hash

## Before submitting

## PR review

## Did you have fun?


src/lightning/__version__.py
```diff
@@ -1 +1,14 @@
-version = "1.8.0rc1"
+# in case you have not installed the package and have only set python path to this source
+
+import subprocess
+
+
+def get_git_revision_short_hash() -> str:
+    # https://stackoverflow.com/a/21901260/4521646
+    return subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
```
carmocca, line 8: I'm not a fan of this solution.

This creates changes to the filesystem (`os.remove` and `os.copy`). And it creates a dependency on `git`, which is not always available. Then you'd get an empty version string which would be very confusing.

I strongly suggest that we keep the different static version files as before.

And to ensure they don't diverge, add a test in CI where we import all packages and assert that the version numbers match.

justusschock: I agree, this is the reason tools like `setuptools-scm` etc. are not more widely adopted.

Borda: So you suggest that you need to remember that all sub-packages need to be updated consistently and if not release process will fail and all need to be edited and released manually because some were uploaded...
I feel that this is a bigger burden, and we have the edit in the codebase already, so this does not introduce new; it just improves it

Borda: > And it creates a dependency on git, which is not always available. Then you'd get an empty version string which would be very confusing.

If you use any way of installation, you get the semantic version, the only time you rely on git is if your IDE has set the source to `src/` folder and in such case, you have cloned this repository (the only other way would download repo as an archive and unzip it, then you get "")

Borda: > to ensure they don't diverge, add a test in CI where we import all packages and assert that the version strings match.

yes, that would be an option, but as mentioned in https://github.com/Lightning-AI/lightning/pull/15374#discussion_r1008001443 we have such edits due to the installation already anyway :/

carmocca: Which edits? `set_version_today` is not used anywhere

justusschock: > If you use any way of installation, you get the semantic version, the only time you rely on git is if your IDE has set the source to src/ folder and in such case, you have cloned this repository (the only other way would download repo as an archive and unzip it, then you get "")

But this is inconsistent and hard to test as well. E.g. the `checkout` action on GH does not preserve the `.git` folder and thereby makes it hard to actually test this. (See https://github.com/actions/checkout/issues/335 )

Borda: > `set_version_today` is not used anywhere

I don't it anywhere, it was used for `lightning` package until we unified to semver

Borda: > But this is inconsistent and hard to test as well. E.g. the `checkout` action on GH does not preserve the `.git` folder and thereby makes it hard to actually test this.

that is fine, we run install in CI

Borda:
ok, changed to have distributed the `version.info` so there is no overwrite of `__version__.py` :otter: 
@carmocca @justusschock can you please check it again :)

otaj (approved):

src/lightning/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version: str = ""
```
otaj, line 17: `mypy` is unhappy about redefinition of type.

```suggestion
        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

src/lightning_app/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version: str = ""
```
otaj, line 17: `mypy` is unhappy about redefinition of type.

```suggestion
        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

src/lightning_lite/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version: str = ""
```
otaj, line 17: `mypy` is unhappy about redefinition of type.

```suggestion
        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

src/pytorch_lightning/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version: str = ""
```
otaj, line 17: `mypy` is unhappy about redefinition of type.

```suggestion
        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

akihironitta (approved):
Is there any assertion in CI to avoid distributing packages with the wrong version info? something like:

```
python -c "assert pl.__version__ == '${{ ... }}'"
```

src/lightning/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = ""
```
otaj, line 17:
```suggestion
version: str

if os.path.isfile(_PATH_VER):
    with open(_PATH_VER, encoding="utf-8") as fo:
        version = fo.readlines()[0].strip()
else:
    try:
        import subprocess

        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

src/lightning_app/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = ""
```
otaj, line 17:
```suggestion
version: str

if os.path.isfile(_PATH_VER):
    with open(_PATH_VER, encoding="utf-8") as fo:
        version = fo.readlines()[0].strip()
else:
    try:
        import subprocess

        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

src/lightning_lite/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = ""
```
otaj, line 17:
```suggestion
version: str

if os.path.isfile(_PATH_VER):
    with open(_PATH_VER, encoding="utf-8") as fo:
        version = fo.readlines()[0].strip()
else:
    try:
        import subprocess

        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

src/pytorch_lightning/__version__.py
```diff
@@ -1 +1,17 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version: str = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = ""
```
otaj, line 17:
```suggestion
version: str

if os.path.isfile(_PATH_VER):
    with open(_PATH_VER, encoding="utf-8") as fo:
        version = fo.readlines()[0].strip()
else:
    try:
        import subprocess

        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
    except Exception:
        version = ""
```

Borda:
> Is there any assertion in CI to avoid distributing packages with the wrong version info? something like:
> 
> ```
> python -c "assert pl.__version__ == '${{ ... }}'"
> ```

good idea, we can also add in another PR, the if pypi release workflow is in action we would check if the generated version is the same as input git tag...

.actions/setup_tools.py
```diff
@@ -236,6 +236,17 @@ def create_mirror_package(src_folder: str, lit_pkg_mapping: dict) -> None:
         copy_adjusted_modules(src_folder, pkg_name, lit_name, lit_pkg_mapping)
 
 
+def distribute_version(src_folder: str, ver_file: str = "version.info") -> None:
+    """Replace the template date with today."""
```
kaushikb11, line 240: is the docstring right?

src/pytorch_lightning/__version__.py
```diff
@@ -1 +1,18 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = ""
```
kaushikb11, line 18: should we print something in the except block? or else it's pretty silent.

kaushikb11 (approved):

tchaton (approved):
LGTM !

.actions/setup_tools.py
Borda:
```suggestion
    """Copy the global version to all packages."""
```

src/pytorch_lightning/__version__.py
Borda: so like `unknown`?

kaushikb11: i think it's okay for now

lantiga (approved):

src/lightning/__version__.py
```diff
@@ -1 +1,18 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = ""
```
carmocca, line 18: I still believe that we should not have this fallback.

_PATH_VER should always exist

manskx (approved):

src/lightning/__version__.py
Borda: So the fallback shall root to the global one?
I have set default as "unknown"

src/lightning_app/__version__.py
```diff
@@ -1 +1,18 @@
...
+if os.path.isfile(_PATH_VER):
+    with open(_PATH_VER, encoding="utf-8") as fo:
+        version = fo.readlines()[0].strip()
+else:
+    try:
+        import subprocess
+
+        version = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode("ascii").strip()
+    except Exception:
+        version = "unknown"
```
carmocca, line 18: this is what I meant with my previous comment.

```suggestion
with open(_PATH_VER, encoding="utf-8") as fo:
    version = fo.read().strip()
```

Since `version.info` will exist, we can always just load it

.gitignore
```diff
@@ -52,6 +52,7 @@ wheels/
 .installed.cfg
 *.egg
 src/lightning/*/
+src/*/version.info
```
carmocca, line 55: Why do we gitignore this? We want to track it

.actions/setup_tools.py
```diff
@@ -236,6 +236,17 @@ def create_mirror_package(src_folder: str, lit_pkg_mapping: dict) -> None:
         copy_adjusted_modules(src_folder, pkg_name, lit_name, lit_pkg_mapping)
 
 
+def distribute_version(src_folder: str, ver_file: str = "version.info") -> None:
```
carmocca, line 239: Do we still need this? The `__version__` files will read the info file contents, no need to rewrite

.gitignore
Borda: no, we want to track only the global in `src/version.info` otherwise this PR does not have any value :)

src/lightning_app/__version__.py
Borda: But it will exist only if you install, not when you just clone and set `PYTHONPATH` to this folder so for example developers...

.actions/setup_tools.py
Borda: the story is about having version written only in single file not as it is now in 4 different

Borda: because that is not part of any (standalone) package :otter:

src/lightning_app/__version__.py
Borda: That is correct, also in this would allow us to understand if someone is using something else then version installed from PyPI, as installed from source can be easily version in range 100+ commits

.actions/setup_tools.py
carmocca: Since it's in the `MANIFEST.in` (https://github.com/Lightning-AI/lightning/pull/15374/files#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003R5), the standalone packages should have access to it, then we don't need one per project.

carmocca: Maybe I should try to make these changes?

carmocca: This is explained in point (4) of https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

Borda: Yes, but if it is not in project it can be overwritten by any package or user...

carmocca:
@Borda I implemented my suggestion in https://github.com/Lightning-AI/lightning/pull/15374/commits/0ef0b5b7a9cba50d3843edff6a9a56566530e3ac. I think it's easier to show the implementation as I don't think our conversation was moving forward. Tell me what you think.

I locally tested it and seemed to be working. Let's also see what CI says about it

carmocca (approved):

PR #15321, opened 2022-10-26
[WIP] Make RichProgressBar the default if rich is available

awaelchli:
## What does this PR do?

Continues #10912, #9647
Fixes #9580

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [ ] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [ ] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @borda @justusschock @kaushikb11 @rohitgr7

PR #15376, opened 2022-10-28
Fix issues when RichProgressBar disabled

awaelchli:
## What does this PR do?

Fixes issues found in #15321. When disabling the progress bar and using it in the Trainer, we get assertion errors and NoneType attribute errors. The added test reproduces several issues. 

Q: When do we have a disabled progress bar in the trainer?
A: When running with DDP, we internally call `progress_bar.disable()` on rank > 0 to avoid showing a bar for each process.

This PR fixes RichProgressBar in combination with DDP and other multi-process strategies.

No.

## Before submitting

## PR review

## Did you have fun?

cc @borda @kaushikb11 @rohitgr7


rohitgr7 (approved):

src/pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -255,9 +255,9 @@ def __init__(
         self._console: Optional[Console] = None
         self._console_kwargs = console_kwargs or {}
         self._enabled: bool = True
-        self.progress: Optional[Progress] = None
-        self.val_sanity_progress_bar_id: Optional["TaskID"] = None
+        self.progress: Optional[CustomProgress] = None
         self.main_progress_bar_id: Optional["TaskID"]
+        self.val_sanity_progress_bar_id: Optional["TaskID"] = None
```
carmocca, line 260: nit: the order here does not match the order in the properties. choose one

src/pytorch_lightning/callbacks/progress/rich_progress.py
```diff
@@ -355,9 +381,11 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
     def on_validation_batch_start(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", batch: Any, batch_idx: int, dataloader_idx: int
     ) -> None:
-        if not self.has_dataloader_changed(dataloader_idx) or self.progress is None:
```
carmocca, line 358: Isn't this equal? When `is_disabled`, `self.progress` is None. Does that mean that there are times when this is disabled but progress exist? When does that happen normally?

carmocca (approved):

src/pytorch_lightning/callbacks/progress/rich_progress.py
awaelchli: After the `on_x_start` hook runs, the value `is_disabled` and `self.progress is None` will always identical. The code change in this line is merely making it consistent with the other checks.

carmocca: The benefit of checking `self.progress is None` is that it avoids adding the `assertion` for mypy right after.

But if you prefer the current way, that's okay too. Feel free to resolve

awaelchli: We don't actually "save" that many assertions. If you look at the usages in other hooks, we still need assertions like
`assert self.test_progress_bar_id is not None`. 
I prefer the expressiveness of `is_disabled`.

PR #15383, opened 2022-10-28
[App] Fix cluster logic

ethanwharris:
## What does this PR do?

Fixes cluster logic with the following:
- if the provided cluster is `None` and there is no cluster in the `.lightning` file and the app doesn't already exist, we run the app then retrieve the cluster that it ran on and save it in the `.lightning` file
- if the provided cluster is `None` and there is no cluster in the `.lightning` file and the app already exists, we run the app on the existing cluster and save it in the `.lightning` file
- if the provided cluster is `None` and there is a cluster in the `.lightning` file, we run the app on that cluster
- if the provided cluster is not `None` and the app already exists on a different cluster, you get an error
- if the cluster doesn't exist, a confirmation dialog checks if you want to run on lightning cloud instead

## Before submitting

## PR review

## Did you have fun?

cc @borda


lantiga (changes requested):
Looks good, I'd change the error message and the related behavior.

Also, as per this discussion https://lightning-ai-corp.slack.com/archives/C020R2Q03UJ/p1666953770700849?thread_ts=1666812331.291149&cid=C020R2Q03UJ let's check app.py is at the root of all tests/examples (it should be)

tchaton (approved):
LGTM !

src/lightning_app/runners/cloud.py
```diff
@@ -293,11 +295,15 @@ def dispatch(
...
+                    # Re-run the app on the same cluster
+                    app_config.cluster_id = existing_instance.spec.cluster_id
+
                 # TODO: support multiple instances / 1 instance per cluster
-                if existing_instance.spec.cluster_id != cluster_id:
+                if existing_instance.spec.cluster_id != app_config.cluster_id:
                     raise ValueError(
-                        f"Can not start app '{name}' on cluster '{cluster_id}' "
-                        f"since this app already exists on '{existing_instance.spec.cluster_id}'. "
+                        f"Cannot start app '{app_config.name}' on cluster '{app_config.cluster_id}' "
```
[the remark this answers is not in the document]
ethanwharris: Done 😃

src/lightning_app/runners/cloud.py
```diff
@@ -326,20 +301,59 @@ def dispatch(
...
+                if app_config.cluster_id not in cluster_ids:
+                    if cluster_id:
+                        msg = f"You requested to run on cluster {cluster_id}, but that cluster doesn't exist."
+                    else:
+                        msg = (
+                            f"Your app last ran on cluster {app_config.cluster_id}, but that cluster "
+                            "doesn't exist anymore."
+                        )
+                    click.confirm(
+                        f"{msg} Do you want to run on Lightning Cloud instead?",
```
lantiga, line 317: Here the right ™️  thing to do would be to get the default cluster from the server (i.e. the cluster that corresponds to `None`) and then include that in the prompt. Not sure we have that notion though.

lantiga (approved):

src/lightning_app/runners/cloud.py
ethanwharris: Yeah, this should be updated when we have a way to figure out the default

awaelchli:
I'm surprised that we removed all the .lightning files. I hope in the future we can fix the core of the issue regarding the clusters (I'm not sure I fully understand it). My understanding is that this PR just avoids part of the problem to make it less frequent. But of course, the --name option in the cli exists and it is always possible to run into that error you describe that way.

manskx:
> 

+1

PR #15424, opened 2022-10-31
[App] Improve startup time

tchaton:
## What does this PR do?

This PR improves Lightning App startup by introducing the followings:

### 1. Start method for the Work's.

This method enables to start the work without having to run the work itself.

```python
class BoringApp(L.LightningFlow):
    def __init__(self):
        super().__init__()
        self.dict = Dict()

    def run(self):
        # create dynamically the source_work at runtime
        if "src_w" not in self.dict:
            self.dict["src_w"] = SourceFileWork()

            self.dict["dst_w"] = DestinationFileAndServeWork(
                script_path=os.path.join(os.path.dirname(__file__), "scripts/serve.py"),
                port=1111,
                parallel=False,  # runs until killed.
                cloud_compute=L.CloudCompute(),
                raise_exception=True,
            )

            self.dict["dst_w"].start() # HERE, we are starting the machine in the cloud.
```

### 2. Add a ``start`` method for the Work's. 

```python
class BoringApp(L.LightningFlow):
    def __init__(self):
        super().__init__()
        self.source_work = SourceFileWork(start_before_setup=True)
        self.dest_work = DestinationFileAndServeWork(
            script_path=os.path.join(os.path.dirname(__file__), "scripts/serve.py"),
            port=1111,
            parallel=False,  # runs until killed.
            cloud_compute=L.CloudCompute(),
            raise_exception=True,
            start=True,
        )

    def run(self):
        self.source_work.run()
        if self.source_work.has_succeeded:
            # the flow passes the file from one work to another.
            self.dest_work.run(self.source_work.boring_path)
            self._exit("Boring App End")

    def configure_layout(self):
        return {"name": "Boring Tab", "content": self.dest_work.url + "/file"}


app = L.LightningApp(BoringApp())
```

### 3. Add setup method to the flow to coordinate your machines to indicate your App is ready.

```python
class Root(LightningFlow):

    def __init__(self):
        super().__init__(self):
        self.database = Database(..., start_before_setup=True)

    def setup(self, *args, **kwargs):
        return self.database.alive()

    def run(self):
        ...
```

In the cloud, the works are going to start at the same time as the flow and speed up setup time.

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_app/core/app.py
+++ b/src/lightning_app/core/app.py
@@ -407,6 +407,9 @@ class LightningApp:
 
         try:
             self.check_error_queue()
+            if not self.has_setup:
+                self.has_setup = self.root.setup("")
+
             # Execute the flow only if:
             # - There are state changes
             # - It is the first execution of the flow
@@ -462,9 +465,6 @@ class LightningApp:
         self._reset_run_time_monitor()
 
         while not done:
-
-            if not self.has_setup:
-                self.has_setup = self.root.setup("")
 
             done = self.run_once()
 
--- a/src/lightning_app/core/app.py
+++ b/src/lightning_app/core/app.py
@@ -119,6 +119,7 @@ class LightningApp:
         self.caller_queues: Optional[Dict[str, BaseQueue]] = None
         self.work_queues: Optional[Dict[str, BaseQueue]] = None
         self.commands: Optional[List] = None
+        self.has_setup = False
 
         self.should_publish_changes_to_api = False
         self.component_affiliation = None
@@ -409,7 +410,7 @@ class LightningApp:
             # Execute the flow only if:
             # - There are state changes
             # - It is the first execution of the flow
-            if self._has_updated:
+            if self._has_updated and self.has_setup:
                 self.root.run()
         except CacheMissException:
             self._on_cache_miss_exception()
@@ -461,6 +462,10 @@ class LightningApp:
         self._reset_run_time_monitor()
 
         while not done:
+
+            if not self.has_setup:
+                self.has_setup = self.root.setup("")
+
             done = self.run_once()
 
             self._update_run_time_monitor()
--- a/src/lightning_app/core/app.py
+++ b/src/lightning_app/core/app.py
@@ -408,7 +408,9 @@ class LightningApp:
         try:
             self.check_error_queue()
             if not self.has_setup:
-                self.has_setup = self.root.setup("")
+                self._on_before_setup()
+                # TODO: Add support for multiple setup stage.
+                self.has_setup = self.root.setup(stage=None)
 
             # Execute the flow only if:
             # - There are state changes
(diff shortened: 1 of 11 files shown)


src/lightning_app/utilities/proxies.py
```diff
@@ -41,6 +42,11 @@
 _state_observer_lock = threading.Lock()
 
 
+@dataclass
+class Start:
+    has: str = str(uuid4().hex)
```
lantiga, line 47: hash?

src/lightning_app/core/work.py
```diff
@@ -78,6 +79,9 @@ def __init__(
             local_build_config: The local BuildConfig isn't used until Lightning supports DockerRuntime.
             cloud_build_config: The cloud BuildConfig enables user to easily configure machine before running this work.
             run_once: Deprecated in favor of cache_calls. This will be removed soon.
+            start_before_setup: Whether the work should be started before the event loop is triggered.
```
lantiga, line 82: I'm not sure this is needed strictly speaking. Couldn't we have a `setup` method and then just call `start` from there?
I don't think we gain much from having this flag, from a startup time perspective, since flow needs to be provisioned anyway before we warmup the works.

src/lightning_app/core/work.py
```diff
@@ -595,3 +613,11 @@ def _register_cloud_compute(self):
...
             _CLOUD_COMPUTE_STORE[internal_id] = _CloudComputeStore(id=internal_id, component_names=[])
         _CLOUD_COMPUTE_STORE[internal_id].add_component_name(self.name)
+
+
+def _default_start_before_setup() -> bool:
+    from lightning_app.utilities.app_helpers import _LightningAppRef
+
+    app = _LightningAppRef().get_current()
+
+    return not app.has_setup
```
lantiga, line 623: We may not want to start works by default if `setup` is defined. However, we want to be able to control when to call `start` on works, whenever in the lifetime of the app (i.e. not only at setup).

src/lightning_app/core/app.py
```diff
@@ -596,3 +603,11 @@ def _on_run_end(self):
...
             del os.environ["LIGHTNING_DEBUG"]
             _console.setLevel(logging.INFO)
+
+    def _on_before_setup(self):
+        for w in self.works:
+            if w._start_before_setup:
+                parallel = w.parallel
+                w._parallel = True
+                w.start()
+                w._parallel = parallel
```
lantiga, line 613: I would move the `_parallel` hack to `start`, and give the possibility to call `start` manually from within `setup`. I would then eliminate this method.

src/lightning_app/utilities/proxies.py
tchaton: Actually, let me remove it. This isn't correct.

src/lightning_app/core/work.py
lantiga: ok, let's just rename to `start`

tchaton: This is required to ensure we can start the works at the same time as the flow.

tchaton: Yes, this was my original name. Changed back. The only caveat is the start argument and the start method.

src/lightning_app/core/app.py
tchaton: Moved to the runtime. This is required only for multiprocessing locally.

PR #15413, opened 2022-10-30
Fix usage of fs.listdir in CheckpointConnector

leoleoasd:
## What does this PR do?

Fix use path in fs.listdir in checkpoint

Fixes #15379

### Does your PR introduce any breaking changes? If yes, please list them.

no

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [x] Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- [x] Did you list all the **breaking changes** introduced by this pull request?
- [x] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

PR #3805, opened 2020-10-02
Fsspec is inconsistant when doing fs.ls

f4hy:
## What does this PR do?
FSSpec, which we use for doing IO which may or may not be local, has some inconsistancies in how it lists files. On local and some remote systems `fs.ls(path)` gives a list of strings and others (hdfs for example) it gives a list of dictionaries with the paths and meta data. 

Switching to fs.listdir(path) seems to give consistent behavior on the platforms I tested. Despite fsspec listing that `ls` is an alias for `listdir` in their documentation.

Fixes # (issue)

# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you make sure your PR does only one thing, instead of bundling different changes together? Otherwise, we ask you to create a separate PR for every change.
- [ ] Did you make sure to update the documentation with your changes?
- [ ] Did you write any new necessary tests? 
- [X] Did you verify new and existing tests pass locally with your changes?
- [ ] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)?

## PR review    
Anyone in the community is free to review the PR once the tests have passed.     
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #2164, opened 2020-06-12
[Ready for review] use gfile to support remote directories

f4hy:
# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?   
- [x] Did you write any new necessary tests?  
- [x] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)?

## What does this PR do?
Fixes #2161 

## PR review    
Anyone in the community is free to review the PR once the tests have passed.     
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #15425, opened 2022-10-31
Use `sys.__stdout__` for terminal encoding check

nihir27:
## What does this PR do?
It's common to redirect `sys.std*`. Currently, if you've redirected `sys.stdout` using a custom class and that class does not have an `encoding` attribute set,  `trainer.test` can fail with an `AttributeError`.  Since the `sys.stdout.encoding` check here concerns terminals, I propose using `sys.__stdout__` instead.

## Before submitting

## PR review

## Did you have fun?


awaelchli:
@nihir27 Thanks for sending the PR. Do you have a way to demonstrate this using a small runnable code example? I'm just generally curious.

src/pytorch_lightning/loops/dataloader/evaluation_loop.py
```diff
@@ -370,8 +370,8 @@ def _print_results(results: List[_OUT_DICT], stage: str) -> None:
 
                 try:
                     # some terminals do not support this character
-                    if sys.stdout.encoding is not None:
-                        "─".encode(sys.stdout.encoding)
+                    if sys.__stdout__.encoding is not None:
```
carmocca, line 373: Given your motivation, wouldn't it be better to check `hasattr(sys.stdout, "encoding")`?

After all, `sys.stdout` is what we will use to print, and `__stdout__` is just a reference to the original handle: https://docs.python.org/3/library/sys.html#sys.__stdout__

nihir27:
@awaelchli For example, below would raise an `AttributeError`
```python
import sys

class Logger(object):
    def __init__(self):
        self.terminal = sys.__stdout__
        self.log = open("logfile.log", "a")
   
    def write(self, message):
        self.terminal.write(message)
        self.log.write(message)  

    def flush(self):
        pass    

sys.stdout = Logger()
```
https://stackoverflow.com/a/14906787

src/pytorch_lightning/loops/dataloader/evaluation_loop.py
nihir27: `hasattr(sys.stdout, "encoding") and sys.stdout.encoding is None` is an option that prevents an `AttributeError` being raised but for the case where one is writing to a terminal as well as a log file (like in https://stackoverflow.com/a/14906787) you lose the handling of the `─` character.

nihir27: That particular pattern may be actually be rare and in that case I would favour just doing `hasattr(sys.stdout, "encoding") and sys.stdout.encoding is None`.

awaelchli:
@nihir27 Would you mind adding this example as a unit test case in the file `tests/tests_pytorch/trainer/logging_/test_eval_loop_logging.py`, similar to the tests there that use `EvaluationLoop._print_results`.

leoleoasd:
Do we actually have any tests for remote file systems? Did anyone try running lightning apps locally with file:// or local://? I don't see any tests in #3805 and #2164.

image
I don't know how we organize tests here (for example, what kind of tests should be in app/lightning/pytorch), but there aren't any string that contains protocol other than HTTP or HTTPS in all PyTorch tests (apart from what I just added).

I added a test that uses fsspec's in-memory file system, and other things related to the test (_hpc_resume_path) are broken too.
I think we DO need to change EVERY usage of fsspec and every path-related function which path may be passed into fsspec.
For example, `os.path.join` is used multiple times in the code base. But we should use `fs.sep` instead of the OS path separator, since we may need to access s3fs from windows, and `os.path.join` will try to use `\` for S3FS paths on windows.

leoleoasd:
Just checked fsspec's document, it seems that it supports both `\` and `/` on windows, but all FS have `/` as sep.
https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/local.html?highlight=windows#

leoleoasd:
Sorry, I opened too many browser pages, The above two comments should be moved to #15413

awaelchli:
@leoleoasd If you find the time, would you mind taking a look at the failing test `test_native_print_results_encodings`, which is affected given your changes?

leoleoasd:
Are you mentioning the right person?

awaelchli:
Nope 🤣 Looks like you infected me in having too many browser pages open. I meant to ping @nihir27 :)

awaelchli:
@nihir27 One test is failing here:
```

    @pytest.mark.parametrize("encoding", ["latin-1", "utf-8"])
    def test_native_print_results_encodings(monkeypatch, encoding):
        import pytorch_lightning.loops.dataloader.evaluation_loop as imports
    
        monkeypatch.setattr(imports, "_RICH_AVAILABLE", False)
    
        out = mock.Mock()
        out.encoding = encoding
        with redirect_stdout(out) as out:
            EvaluationLoop._print_results(*inputs0)
    
        # Attempt to encode everything the file is told to write with the given encoding
        for call_ in out.method_calls:
            name, args, kwargs = call_
            if name == "write":
>               args[0].encode(encoding)
E               UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-119: ordinal not in range(256)

```

Could you take another look? Let us know if it doesn't work and we can help.

Borda:
@awaelchli, what is the situation here? :)

PR #15441, opened 2022-10-31
Periodically sync database to the drive

lantiga:
## What does this PR do?

Currently the database component syncs to the drive `on_exit`, which will also be triggered in case of an exception. In order to further ensure that data is not lost in case of an unhandled crash, this PR introduces periodic sync'ing of the database with a set time interval.

None

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/components/database/server.py
```diff
@@ -172,6 +180,15 @@ def db_url(self) -> Optional[str]:
...
 
-    def on_exit(self):
-        self.drive.put(self.db_filename)
+    def periodic_store_database(self, store_interval):
+        while True:
+            self.store_database()
+            time.sleep(store_interval)
+
+    def store_database(self):
+        with self._store_lock:
```
awaelchli, line 189: Is the lock here to prevent exit() and the _store_thread to interact in a race condition?

src/lightning_app/components/database/server.py
```diff
@@ -38,6 +40,7 @@ def __init__(
         self,
         models: Union[Type["SQLModel"], List[Type["SQLModel"]]],
         db_filename: str = "database.db",
+        store_interval: int = 10,
```
awaelchli, line 43: Maybe we could add this to the docstring below and say the value is in seconds

tests/tests_app/components/database/test_client_server.py
```diff
@@ -160,3 +161,40 @@ def run(self):
     sleep(2)
 
     os.remove(id)
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="currently not supported for windows.")
+@pytest.mark.skipif(not _is_sqlmodel_available(), reason="sqlmodel is required for this test.")
+def test_work_database_periodic_store():
```
awaelchli, line 168: We could use tmpdir here so that the test does not create artifacts in the working dir

awaelchli (approved):

src/lightning_app/components/database/server.py
lantiga: Yes correct

tests/tests_app/components/database/test_client_server.py
lantiga: Done

src/lightning_app/components/database/server.py
```diff
@@ -172,6 +182,15 @@ def db_url(self) -> Optional[str]:
...
 
-    def on_exit(self):
-        self.drive.put(self.db_filename)
+    def periodic_store_database(self, store_interval):
+        while True:
+            self.store_database()
+            time.sleep(store_interval)
+
+    def store_database(self):
+        with self._store_lock:
```
tchaton, line 191: The lock should also be used within the FastAPI request handler to avoid updating the database while copying it.

lantiga: This prompted to get more in depth. You should be able to copy a SQLite file at any time while it's being updated, but due to a Unix quirckiness, it can lead to issues when done from a thread on POSIX systems.

I'm going to update this PR so that we make a proper backup with the official API and then sync.

src/lightning_app/components/database/server.py
```diff
@@ -132,18 +138,44 @@ class CounterModel(SQLModel, table=True):
...
+
+            drive = Drive("lit://database", component_name=self.name, root_folder=tmpdir)
+            drive.put(os.path.basename(tmp_db_filename))
+
+        print("Stored the database to the Drive.")
+
+    def periodic_store_database(self, store_interval):
+        while not self._exit_event.is_set():
+            self.store_database()
+            self._exit_event.wait(store_interval)
```
awaelchli, line 168: I like it!

src/lightning_app/components/database/server.py
```diff
@@ -132,18 +138,44 @@ class CounterModel(SQLModel, table=True):
...
+
+            source = sqlite3.connect(self.db_filename)
+            dest = sqlite3.connect(tmp_db_filename)
+
+            source.backup(dest)
+
+            source.close()
+            dest.close()
+
+            drive = Drive("lit://database", component_name=self.name, root_folder=tmpdir)
```
awaelchli, line 160: Was there a specific reason why re-using the self.drive reference didn't work?

tests/tests_app/components/database/test_client_server.py
```diff
@@ -141,22 +144,57 @@ def run(self):
...
                 self._exit()
             else:
-                assert os.path.exists(id)
+                assert os.path.exists(self._db_filename)
                 assert len(self._client.select_all()) == 1
                 self._exit()
 
-    app = LightningApp(Flow())
-    MultiProcessRuntime(app).dispatch()
+    with tempfile.TemporaryDirectory() as tmpdir:
```
awaelchli, line 151: Another easy way to get a test-specific tmpdir is to use the predefined fixture like so:

```py
test_work_database_restart(tmpdir):
    ...
```

awaelchli (approved):

src/lightning_app/components/database/server.py
lantiga: because they had two different root folders (one tmp)

tests/tests_app/components/database/test_client_server.py
lantiga: oh I see, nice to know :-)

src/lightning_app/components/database/server.py
```diff
@@ -132,18 +138,44 @@ class CounterModel(SQLModel, table=True):
...
+            drive.put(os.path.basename(tmp_db_filename))
+
+        print("Stored the database to the Drive.")
```
Borda, line 163: Why `print` and not `logging.info`?

Borda (approved):
What will happen if the sync process takes longer than the interval?

src/lightning_app/components/database/server.py
lantiga: we convert prints to info behind the scenes, this is what we adopted throughout

lantiga:
> What will happen if the sync process takes longer than the interval?

The copy and the interval are sequential and all calls are blocking, so the interval indicates how much time will elapse before a new copy is performed.

tchaton (approved):
Nice addition !

PR #15461, opened 2022-11-02
[App] Actually fix launching apps on multiple clusters

luca3rd:
## What does this PR do?

Actually fix the cross launching apps issue.

### After
Screen Shot 2022-11-01 at 9 25 34 PM

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_app/runners/cloud.py
+++ b/src/lightning_app/runners/cloud.py
@@ -459,9 +459,9 @@ class CloudRuntime(Runtime):
             )
             if not os.path.exists(os.path.join(root, ".lightningignore")):
                 warning_msg = (
-                    warning_msg +
-                    "\nIn order to ignore some files or folder, " +
-                    "create a `.lightningignore` file and add the paths to ignore."
+                    warning_msg
+                    + "\nIn order to ignore some files or folder, "
+                    + "create a `.lightningignore` file and add the paths to ignore."
                 )
             else:
                 warning_msg += "\nYou can ignore some files or folders by adding them to `.lightningignore`."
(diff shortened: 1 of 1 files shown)


lantiga:
I don't think this is what we want.

App names must be unique per-project, not per-cluster @edenafek

rusenask:
> I don't think this is what we want.
> 
> App names must be unique per-project, not per-project + per-cluster @edenafek (I feel strongly about this).
> 
> If you try to run an app with the same name in a project, targeting a different cluster, the first app needs to be stopped, irrespective of what cluster it is running on. We need to keep the model simple, otherwise the CLI will complicate a ton (now it's not enough to address the app by name, you also need to think whether there's another instance of that app running on another cluster, and instead use the id to address it, e.g. connect, from the CLI).

:+1: I also don't think this is useful for any scenario. Why would you have the same name @edenafek ? If I wanted this as a user I would just create two apps `muse-east-1` and `muse-east-2` and then I don't have to think about it. This is almost as having two files in your directory with the same name, while possible, doesn't make things good :)

src/lightning_app/runners/cloud.py
```diff
@@ -285,14 +285,25 @@ def dispatch(
...
+            if previous_app_instances:
+                most_recent_instance = previous_app_instances[0]
                 if not app_config.cluster_id:
-                    # Re-run the app on the same cluster
-                    app_config.cluster_id = existing_instance.spec.cluster_id
+                    # Re-run the app on the most recent cluster
+                    app_config.cluster_id = most_recent_instance.spec.cluster_id
+
+                # get the existing instance on the cluster, if any
+                existing_instance = existing_instances_by_cluster.get(app_config.cluster_id, None)
```
dmitsf, line 303:
```suggestion
                existing_instance = existing_instances_by_cluster.get(app_config.cluster_id)
```

dmitsf: `None` is already the default returning value :)

PR #15471, opened 2022-11-02
Reuse existing commands when running connect more than once

lantiga:
## What does this PR do?

Everytime we connect, we download all commands from scratch, which slows down the experience. This PR copies the commands and connect file from an active connection, if it exists.

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/cli/commands/connection.py
```diff
@@ -19,6 +19,7 @@
 _PPID = str(psutil.Process(os.getpid()).ppid())
 _LIGHTNING_CONNECTION = os.path.join(_HOME, ".lightning", "lightning_connection")
 _LIGHTNING_CONNECTION_FOLDER = os.path.join(_LIGHTNING_CONNECTION, _PPID)
+_CLEAN_LIGHTNING_CONNECTION = True
```
awaelchli, line 22: I couldn't see from the code why a global variable is required. Could you explain it and/or write a small comment in the source code? Is there no way to avoid it?

tests/tests_app/cli/test_connect.py
```diff
@@ -78,8 +78,11 @@ def fn(msg):
 
     assert _retrieve_connection_to_an_app() == (None, None)
 
+    connection_module._CLEAN_LIGHTNING_CONNECTION = True
+
 
 def test_connect_disconnect_cloud(monkeypatch):
+    connection_module._CLEAN_LIGHTNING_CONNECTION = False
```
awaelchli, line 85: I suggest to use monkypatch to set this variable

```suggestion
    monkeypatch.setattr(connection_module, "_CLEAN_LIGHTNING_CONNECTION", False)
```

The benefit is that no matter how this test case exits (failure or not), monkeypatch will ensure that the variable gets reset to its original value. This helps avoid a potential cascade of failures in subsequent tests due to interactions through global variables.


Another suggestion is to monkeypatch the `_LIGHTNING_CONNECTION_FOLDER` not to be in the home folder, but instead to be under a tmpdir. For the same reason to 1) avoid interactions of tests through the file system 2) avoid leaving test artifacts behind on the developers system.

I believe these two suggestions combined would make the teardown logic

```py
    connection_module._CLEAN_LIGHTNING_CONNECTION = True
    connection_module._clean_lightning_connection()
```
at the end obsolete.

src/lightning_app/cli/commands/connection.py
manskx: If needed, it could be moved to constants as an env var and maybe renaming it to `_ENABLE_CLEAN_LIGHTNING_CONNECTION`

src/lightning_app/cli/commands/connection.py
```diff
@@ -178,9 +202,7 @@ def disconnect(logout: bool = False):
             )
 
 
-def _retrieve_connection_to_an_app() -> Tuple[Optional[str], Optional[str]]:
-    connected_file = os.path.join(_LIGHTNING_CONNECTION_FOLDER, "connect.txt")
-
+def _read_connected_file(connected_file):
```
manskx, line 205:
```suggestion
def _read_connected_file(connected_file: Union[pathlib.Path, str]) -> Tuple[Optional[str], Optional[str]]:
```

src/lightning_app/cli/commands/connection.py
```diff
@@ -284,3 +311,27 @@ def _clean_lightning_connection():
             connection = os.path.join(_LIGHTNING_CONNECTION, str(ppid))
             if os.path.exists(connection):
                 shutil.rmtree(connection)
+
+
+def _scan_lightning_connections(app_name_or_id):
```
manskx, line 316:
```suggestion
def _scan_lightning_connections(app_name_or_id: str) -> Union[str, None]:
```

src/lightning_app/cli/commands/connection.py
```diff
@@ -284,3 +311,27 @@ def _clean_lightning_connection():
...
+    if not os.path.exists(_LIGHTNING_CONNECTION):
+        return
+
+    for ppid in os.listdir(_LIGHTNING_CONNECTION):
+        try:
+            psutil.Process(int(ppid))
+        except (psutil.NoSuchProcess, ValueError):
+            continue
```
manskx, line 324: do we need these lines ?

src/lightning_app/cli/commands/connection.py
```diff
@@ -27,13 +28,16 @@ def connect(app_name_or_id: str, yes: bool = False):
     """Connect to a Lightning App."""
     from lightning_app.utilities.commands.base import _download_command
 
-    _clean_lightning_connection()
+    if _CLEAN_LIGHTNING_CONNECTION:
+        _clean_lightning_connection()
```
manskx, line 32: I see `_clean_lightning_connection` is called every time when [executing a command](https://github.com/Lightning-AI/lightning/blob/921dc1cd9afbdbf33589d8b45f390c7957072f93/src/lightning_app/cli/commands/app_commands.py#L24), and since it has a new pid it looks like it will delete it.

lantiga: No because this is going to be the pid of the terminal process, not the pid of the command itself.

tests/tests_app/cli/test_connect.py
lantiga: Totally right @awaelchli

src/lightning_app/cli/commands/connection.py
lantiga: So this is there exclusively for simplifying testing. Admittedly not ideal.
Outside of testing (or debugging) we need this constant to be True.

src/lightning_app/cli/commands/connection.py
lantiga: It’s convenient to avoid peeking into files and checking for the app name/id for all connections.
If the pid doesn’t correspond to any active terminal process, we just skip the rest.

tests/tests_app/cli/test_connect.py
lantiga: Fixed, monkeypatching should be handled properly now. Thanks a lot @awaelchli

src/lightning_app/cli/commands/connection.py
lantiga: I reworked monkeypatching, this has been removed

tchaton (approved):
LGTM !

tests/tests_app/cli/test_connect.py
```diff
@@ -19,10 +21,29 @@
...
+        monkeypatch.setattr("lightning_app.cli.commands.connection._PPID", ppid)
+        monkeypatch.setattr("lightning.app.cli.commands.connection._LIGHTNING_CONNECTION", tmpdir)
+        monkeypatch.setattr("lightning.app.cli.commands.connection._LIGHTNING_CONNECTION_FOLDER", connection_path)
+    return connection_path
+
 
+def test_connect_disconnect_local(monkeypatch):
     disconnect()
 
+    tmpdir = tempfile.mkdtemp()
```
awaelchli, line 42:
```suggestion
def test_connect_disconnect_local(tmpdir, monkeypatch):
    disconnect()

```

tests/tests_app/cli/test_connect.py
```diff
@@ -78,11 +98,19 @@ def fn(msg):
 
     assert _retrieve_connection_to_an_app() == (None, None)
 
+    shutil.rmtree(tmpdir)
```
awaelchli, line 101:
```suggestion
```

tests/tests_app/cli/test_connect.py
```diff
@@ -78,11 +98,19 @@ def fn(msg):
...
 
-def test_connect_disconnect_cloud(monkeypatch):
 
+def test_connect_disconnect_cloud(monkeypatch):
     disconnect()
 
+    tmpdir = tempfile.mkdtemp()
```
awaelchli, line 107:
```suggestion
def test_connect_disconnect_cloud(tmpdir, monkeypatch):
    disconnect()

```

tests/tests_app/cli/test_connect.py
```diff
@@ -170,13 +197,29 @@ def fn(msg):
...
+
     messages = []
     disconnect()
     assert messages == [
         "You aren't connected to any Lightning App. Please use `lightning connect app_name_or_id` to connect to one."
     ]
 
     assert _retrieve_connection_to_an_app() == (None, None)
+
+    shutil.rmtree(tmpdir)
```
awaelchli, line 225:
```suggestion
```

awaelchli (approved):

PR #15484, opened 2022-11-02
[App] Raise error when launching app on multiple clusters

luca3rd:
## What does this PR do?

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_app/runners/cloud.py
+++ b/src/lightning_app/runners/cloud.py
@@ -11,6 +11,7 @@ from typing import Any, Callable, List, Optional, Union
 
 import click
 from lightning_cloud.openapi import (
+    Externalv1LightningappInstance,
     Body3,
     Body4,
     Body7,
@@ -285,6 +286,7 @@ class CloudRuntime(Runtime):
             elif CLOUD_QUEUE_TYPE == "redis":
                 queue_server_type = V1QueueServerType.REDIS
 
+            existing_instance: Optional[Externalv1LightningappInstance] = None
             if find_instances_resp.lightningapps:
                 existing_instance = find_instances_resp.lightningapps[0]
 
@@ -329,6 +331,19 @@ class CloudRuntime(Runtime):
                         default=True,
                     )
                     app_config.cluster_id = None
+                if existing_instance and existing_instance.spec.cluster_id != app_config.cluster_id:
+                    raise ValueError(
+                        "\n".join([
+                            f"Can not run app '{app_config.name}' on cluster {app_config.cluster_id} "
+                                f"since it already exists on {existing_instance.spec.cluster_id} "
+                                "(moving apps between clusters is not supported).",
+                            "You can either:",
+                            f"a. rename app to run on {existing_instance.spec.cluster_id} with the --name option",
+                            "lightning run app script.py"
+                                f"--name (new name) --cloud --cluster-id {existing_instance.spec.cluster_id}",
+                            "b. delete the existing app in the UI before running this command."
+                        ])
+                    )
 
             if app_config.cluster_id is not None:
                 self._ensure_cluster_project_binding(project.project_id, app_config.cluster_id)
--- a/tests/tests_app/runners/test_cloud.py
+++ b/tests/tests_app/runners/test_cloud.py
@@ -92,19 +92,30 @@ def get_cloud_runtime_request_body(**kwargs) -> "Body8":
     return Body8(**default_request_body)
 
 
+@pytest.fixture
+def cloud_backend(monkeypatch):
+    cloud_backend = mock.MagicMock()
+    monkeypatch.setattr(cloud, "LocalSourceCodeDir", mock.MagicMock())
+    monkeypatch.setattr(cloud, "_prepare_lightning_wheels_and_requirements", mock.MagicMock())
+    monkeypatch.setattr(backends, "CloudBackend", mock.MagicMock(return_value=cloud_backend))
+    return cloud_backend
+
+@pytest.fixture
+def project_id():
+    return "test-project-id"
+
 class TestAppCreationClient:
     """Testing the calls made using GridRestClient to create the app."""
 
     # TODO: remove this test once there is support for multiple instances
-    @mock.patch("lightning_app.runners.backends.cloud.LightningClient", mock.MagicMock())
-    def test_new_instance_on_different_cluster(self, monkeypatch):
+    def test_new_instance_on_different_cluster(self, monkeypatch, cloud_backend, project_id):
         app_name = "test-app-name"
         original_cluster = "cluster-001"
         new_cluster = "cluster-002"
 
         mock_client = mock.MagicMock()
         mock_client.projects_service_list_memberships.return_value = V1ListMembershipsResponse(
-            memberships=[V1Membership(name="Default Project", project_id="default-project-id")]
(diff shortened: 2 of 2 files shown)


src/lightning_app/runners/cloud.py
```diff
@@ -329,6 +332,20 @@ def dispatch(
                         default=True,
                     )
                     app_config.cluster_id = None
+                if existing_instance and existing_instance.spec.cluster_id != app_config.cluster_id:
```
tchaton, line 335: I am not sure to fully understand why it is needed.

tchaton: It is a single running instance per cluster, not over all clusters.

tchaton: cc @edenlightning

tests/tests_app/runners/test_cloud.py
```diff
@@ -92,75 +94,99 @@ def get_cloud_runtime_request_body(**kwargs) -> "Body8":
...
+
+
+@pytest.fixture
+def project_id():
+    return "test-project-id"
+
+
+@contextmanager
+def does_not_raise():
+    yield
```
awaelchli, line 113: `from contextlib import nullcontext` :)

tests/tests_app/runners/test_cloud.py
```diff
@@ -92,75 +94,99 @@ def get_cloud_runtime_request_body(**kwargs) -> "Body8":
...
+            new_cluster = new_cluster or DEFAULT_CLUSTER
+            assert str(exception.value) == dedent(
+                f"""\
+                Can not run app '{app_name}' on cluster {new_cluster} since it already exists on {old_cluster}
+                    (moving apps between clusters is not supported).
+
+                You can either:
+                    a. rename the app to run on {new_cluster} with the --name option
+                        lightning run app script.py --name (new name) --cloud --cluster-id {new_cluster}
+                    b. delete the app running on {old_cluster} in the UI before running this command.
```
awaelchli, line 187: What exactly is the reason for duplicating the entire error message here? Let's not do this. Let's use the pytest match in the context manager:

```py
with pytest.raises(
    ValueError, 
    match="Can not run app '{app_name}' on cluster {new_cluster} since it already exists on {old_cluster}"
)
```
That should be enough to identify the message uniquely!

src/lightning_app/runners/cloud.py
```diff
@@ -329,6 +332,20 @@ def dispatch(
...
+                    raise ValueError(
+                        dedent(
+                            f"""\
+                            Can not run app '{app_config.name}' on cluster {app_config.cluster_id} since it already exists on {existing_instance.spec.cluster_id}
+                                (moving apps between clusters is not supported).
```
awaelchli, line 340: some terms are in quotes, others are not

tests/tests_app/runners/test_cloud.py
```diff
@@ -92,75 +94,99 @@ def get_cloud_runtime_request_body(**kwargs) -> "Body8":
...
         )
         mock_client.lightningapp_v2_service_create_lightningapp_release.return_value = V1LightningappRelease(
             cluster_id=new_cluster
         )
+
+        # backend converts "None" cluster to "litng-ai-03"
         mock_client.cluster_service_list_clusters.return_value = V1ListClustersResponse(
-            [Externalv1Cluster(id=original_cluster), Externalv1Cluster(id=new_cluster)]
+            [
+                Externalv1Cluster(id=old_cluster or DEFAULT_CLUSTER),
```
awaelchli, line 149: why are we doing `x or default` everywhere in the test? This is the same as setting default directly in the parameterization (where it is currently None).

awaelchli (approved):
LGTM but would be glad if comments in tests could be addressed :)

src/lightning_app/runners/cloud.py
luca3rd: Making this consistent. Thanks!

tests/tests_app/runners/test_cloud.py
luca3rd: I updated the comment on lines 144-146 to provide this context.

src/lightning_app/runners/cloud.py
luca3rd: It's my understanding that we want to have only one instance of the app running per project.

It's not a problem when updating apps on a single cluster since we stop any previous instance before launching a new one.

When launching an app instance onto a new cluster, this isn't the case. If we launched multiple instances of the app, it could lead to issues with `lightning connect app`, and existing data may not be available to the new cluster which may affect the app.

So instead of having multiple app instances per project / cluster, we decided to throw an error. (@lantiga and @edenlightning please correct me if I'm wrong. )

src/lightning_app/runners/cloud.py
```diff
@@ -337,6 +340,20 @@ def dispatch(
...
+                    raise ValueError(
+                        dedent(
+                            f"""\
+                            Can not run app {app_config.name} on cluster {app_config.cluster_id} since it already exists on {existing_instance.spec.cluster_id}
+                                (moving apps between clusters is not supported).
+
+                            You can either:
+                                a. rename the app to run on {app_config.cluster_id} with the --name option
+                                    lightning run app script.py --name (new name) --cloud --cluster-id {app_config.cluster_id}
+                                b. delete the app running on {existing_instance.spec.cluster_id} in the UI before running this command.
```
lantiga, line 353: Let's add the exact line here `lightning delete app --name {app_config.name}`

src/lightning_app/runners/cloud.py
lantiga: So yes, we 100% need to make the app name unique across a project, independently on whether it is running on cluster X or cluster Y.

We decided to throw an error because making a mistake with clusters may mean that an app meant to be on a VPC is now running on a public cluster. This is safer but it's kind of suboptimal from a UX standpoint IMO, because an app is an app no matter what cluster it is running on, and our journey shouldn't depend on where it is running.

But I think that's ok for now and we can change it later if we/users get annoyed or confused.

src/lightning_app/runners/cloud.py
```diff
@@ -337,6 +340,20 @@ def dispatch(
                         default=True,
                     )
                     app_config.cluster_id = None
+                if existing_instance and existing_instance.spec.cluster_id != app_config.cluster_id:
+                    raise ValueError(
+                        dedent(
+                            f"""\
+                            Can not run app {app_config.name} on cluster {app_config.cluster_id} since it already exists on {existing_instance.spec.cluster_id}
```
lantiga, line 347: I would word as 
```
An app named {app_config.name} is already running on cluster {existing_instance.spec.cluster_id}, and you now requested it to run on cluster {app_config.cluster_id}.

In order to proceed, please either:
...
```

lantiga (changes requested):

src/lightning_app/runners/cloud.py
luca3rd: @lantiga the CLI doesn't have a `lightning delete app` command. We only have `lightning delete cluster`.

lantiga: We will soon: https://github.com/Lightning-AI/lightning/pull/15783
This should land tomorrow (correct @rlizzo ?) so let's add this one in view of having the command.

lantiga (approved):

PR #15488, opened 2022-11-02
CI: exclude docs requirements changes

Borda:
## What does this PR do?

Exclude building test of the change is only to a template or other docs building setup
see: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

## Before submitting

## PR review

## Did you have fun?


.github/workflows/ci-app-examples.yml
```diff
@@ -12,7 +12,8 @@ on:
       - "src/lightning_app/**"
       - "tests/tests_app_examples/**"
       - "examples/app_*"
-      - "requirements/app/**"
+      - "requirements/app/*"
+      - "!requirements/app/docs.txt"
```
carmocca, line 16: I'm not sure that this works as you expect.

A change to `src/foo/bar` matches `!requirements/app/docs.txt`.
Whereas I believe you want to just exclude this particular file.

You have two options.

a) List all `.txt` files explicitly here without `docs.txt`
b) Use paths-ignore: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

Same comment applies to all other changes

carmocca (changes requested):

.github/workflows/ci-app-examples.yml
Borda: I believe that if it does not work, it would be a bug on their side... :confused:

carmocca: Oops I misread GitHub docs. Resolving

.github/workflows/ci-app-examples.yml
```diff
@@ -12,7 +12,8 @@ on:
       - "src/lightning_app/**"
       - "tests/tests_app_examples/**"
       - "examples/app_*"
-      - "requirements/app/**"
+      - "requirements/app/*"
```
carmocca, line 15: Why remove the **?

.github/workflows/ci-pkg-install.yml
```diff
@@ -12,6 +12,7 @@ on:
       - "setup.py"
       - "src/**"
       - "requirements/**"
+      - "!requirements/*/docs.txt"
```
carmocca, line 15: Shouldn't this have 2 `*`?

Borda: so far I know `**` is for any depth and `*` is just one folder

.github/workflows/ci-app-examples.yml
Borda: I think that the depth is just 1 not any

.github/workflows/ci-pkg-install.yml
carmocca: But in the line above you use `**`

Borda: i could be wrong, but... 
- `**` would cover files in requirements and all subfolders
- `*` would cover only files directly in requirements but no subfolder

Borda: I believe that this is the same pattern: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/pattern/PathPattern.html

carmocca (approved):

kaushikb11 (approved):

Borda:
@carmocca reverted :)

carmocca (approved):
LGTM again. Needs https://github.com/Lightning-AI/lightning/pull/15503 to be merged first

Issue #15489, opened 2022-11-03
Distribute Lite's code with PyTorch Lightning


carmocca:
## Proposed refactor

Remove this line https://github.com/Lightning-AI/lightning/blob/5025c3ec162b7a1af978bfb19424fd1bb2683012/requirements/pytorch/base.txt#L6

and distribute the Lite source code together with PL when it's installed

### Motivation

When you do
```bash
PACKAGE_NAME=pytorch pip install -e .
```
to create an editable installation of PL. Lite gets installed from PyPI:
```bash
$ pip freeze | grep lightning
lightning-lite==1.8.0.post1
-e git+https://github.com/Lightning-AI/lightning.git@5025c3ec162b7a1af978bfb19424fd1bb2683012#egg=pytorch_lightning
...
```

This means that local changes to Lite are not reflected when developing locally

One downside of this is that if you also do `pip install lightning_lite`, you might have conflicting lite installations. The PyPI version will take precedence.
______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/Lightning-AI/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/Lightning-AI/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/Lightning-AI/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging PyTorch Lightning, Transformers, and Hydra.

cc @carmocca @akihironitta @borda

carmocca:
> One downside of this is that if you also do pip install lightning_lite, you might have conflicting lite installations. The PyPI version will take precedence.

To alleviate this, @Borda suggests failing when we install Lite if PL is already installed

awaelchli:
[...]

PR #15493, opened 2022-11-03
Do not modify PACKAGE_NAME on install

carmocca:
## What does this PR do?

See title.

`lightning` is no longer installed when `PACKAGE_NAME=lite|app|pytorch` is installed.

Part of https://github.com/Lightning-AI/lightning/issues/15474

Fixes https://github.com/Lightning-AI/lightning/issues/15489 by reverting most of https://github.com/Lightning-AI/lightning/pull/14517

None


Borda:
The tricky part is that the package selection needs to be included in the build package, so we can instead of rewrite setup.py and write it to an extra file, but it still need to be dumped somewhere and distributed with sdist/wheel

.github/actions/pkg-install/action.yml
```diff
@@ -1,42 +1,32 @@
...
-    run: echo "PKG_NAME=${{ inputs.pkg-name }}" >> $GITHUB_ENV
-    shell: bash
-
   - name: Install package - archive
     working-directory: ./dist
     run: |
       pip install *.tar.gz ${{ inputs.pip-flags }}
       pip list | grep lightning
-      python -c "import ${PKG_NAME} ; print(${PKG_NAME}.__version__)"
-    shell: bash
```
Borda, line 34: `shell: bash` needs to stay, otherwise it may not be valid action...

.github/actions/pkg-install/action.yml
```diff
@@ -1,42 +1,32 @@
...
+      python -c "import ${{ inputs.pkg-name }}; print(${{ inputs.pkg-name }}.__version__)"
 
   - name: Install package - wheel
     working-directory: ./dist
     run: |
       pip install *.whl ${{ inputs.pip-flags }}
       pip list | grep lightning
-      python -c "import ${PKG_NAME} ; print(${PKG_NAME}.__version__)"
-    shell: bash
+      python -c "import ${{ inputs.pkg-name }}; print(${{ inputs.pkg-name }}.__version__)"
```
Borda, line 32:
```suggestion
      python -c "import ${{ inputs.pkg-name }}; print(${{ inputs.pkg-name }}.__version__)"
    shell: bash
```

.github/actions/pkg-install/action.yml
```diff
@@ -1,42 +1,32 @@
...
-    shell: bash
-
   - name: Install package - archive
     working-directory: ./dist
     run: |
       pip install *.tar.gz ${{ inputs.pip-flags }}
       pip list | grep lightning
-      python -c "import ${PKG_NAME} ; print(${PKG_NAME}.__version__)"
-    shell: bash
+      python -c "import ${{ inputs.pkg-name }}; print(${{ inputs.pkg-name }}.__version__)"
```
Borda, line 25:
```suggestion
      python -c "import ${{ inputs.pkg-name }}; print(${{ inputs.pkg-name }}.__version__)"
    shell: bash
```

setup.py
```diff
@@ -87,19 +69,27 @@ def _load_py_module(name: str, location: str) -> ModuleType:
...
-# https://packaging.python.org/discussions/install-requires-vs-requirements /
-# keep the meta-data here for simplicity in reading this file... it's not obvious
-# what happens and to non-engineers they won't know to look in init ...
-# the goal of the project is simplicity for researchers, don't want to add too much
-# engineer specific practices
 if __name__ == "__main__":
-    _SETUP_TOOLS = _load_py_module(name="setup_tools", location=os.path.join(".actions", "setup_tools.py"))
-
-    if _PACKAGE_NAME not in _PACKAGE_MAPPING:  # install everything
-        _SETUP_TOOLS._load_aggregate_requirements(_PATH_REQUIRE, _FREEZE_REQUIREMENTS)
```
awaelchli, line 99: While we are here, could we add an error if `PACKAGE_NAME` is an invalid name. Ran into this issue already (e.g., lightning-lite vs lightning_lite) and it's annoying 😄

carmocca: That has been added already

.github/actions/pkg-install/action.yml
```diff
@@ -1,42 +1,32 @@
...
 inputs:
   pkg-name:
-    description: package name for import
-    required: false
-    default: ""
+    description: Package name to import
+    required: true
   pip-flags:
-    description: additional pil install flags
+    description: Additional pip install flags
```
awaelchli, line 9: I think we should keep some of these typos. They add a personal touch 😅

setup.py
```diff
@@ -87,19 +69,27 @@ def _load_py_module(name: str, location: str) -> ModuleType:
...
+    setup_tools = _load_py_module(name="setup_tools", location=os.path.join(".actions", "setup_tools.py"))
+
+    package_to_install = _PACKAGE_NAME or "lightning"
+    print(f"Installing the {package_to_install} package")  # requires `-v` to appear
+    if package_to_install == "lightning":
+        # install everything
+        setup_tools._load_aggregate_requirements(_PATH_REQUIRE, _FREEZE_REQUIREMENTS)
+        setup_tools.create_mirror_package(_PATH_SRC, _PACKAGE_MAPPING)
+    elif package_to_install not in _PACKAGE_MAPPING:
+        raise ValueError(f"Unexpected package name: {_PACKAGE_NAME}")
```
awaelchli, line 82: great!

```suggestion
    elif package_to_install not in _PACKAGE_MAPPING:
        raise ValueError(f"Unexpected package name: {_PACKAGE_NAME}. Possible choices are: {list(_PACKAGE_MAPPING.keys())}")
```

awaelchli (approved):
On a high level, looks good, but no expert review!

.azure/gpu-benchmark.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
     - bash: |
         pip install -e . --find-links ${TORCH_URL}
-        git checkout -- setup.py MANIFEST.in
```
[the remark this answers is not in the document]
carmocca: No. Manifest and requirements still get modified

.github/actions/pkg-install/action.yml
```diff
@@ -1,42 +1,32 @@
...
-    run: echo "PKG_NAME=${{ inputs.pkg-name }}" >> $GITHUB_ENV
-    shell: bash
-
   - name: Install package - archive
     working-directory: ./dist
     run: |
       pip install *.tar.gz ${{ inputs.pip-flags }}
       pip list | grep lightning
-      python -c "import ${PKG_NAME} ; print(${PKG_NAME}.__version__)"
+      PKG_NAME=$(python -c "print({'app': 'lightning_app', 'lite': 'lightning_lite', 'pytorch': 'pytorch_lightning', 'lightning': 'lightning'}['${{matrix.pkg-name}}'])")
```
Borda, line 21: let's have it as a separate step and push it to env. vars

.github/workflows/ci-pkg-install.yml
```diff
@@ -68,10 +68,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04, macOS-12, windows-2022]
-        pkg-name: ["app", "lite", "pytorch", ""]
+        pkg-name: ["app", "lite", "pytorch", "lightning"]
```
Borda, line 71:
```suggestion
        pkg-name: ["app", "lite", "pytorch", "lightning", ""]
```
default fro most users installing from source is None

setup.py
```diff
@@ -87,19 +69,27 @@ def _load_py_module(name: str, location: str) -> ModuleType:
...
+    print(f"Installing the {package_to_install} package")  # requires `-v` to appear
+    if package_to_install == "lightning":
+        # install everything
+        setup_tools._load_aggregate_requirements(_PATH_REQUIRE, _FREEZE_REQUIREMENTS)
+        setup_tools.create_mirror_package(_PATH_SRC, _PACKAGE_MAPPING)
+    elif package_to_install not in _PACKAGE_MAPPING:
+        raise ValueError(f"Unexpected package name: {_PACKAGE_NAME}. Possible choices are: {list(_PACKAGE_MAPPING)}")
+
+    possible_packages = _PACKAGE_MAPPING.values() if _PACKAGE_NAME is None else [_PACKAGE_MAPPING[_PACKAGE_NAME]]
+    for pkg in possible_packages:
```
Borda, line 85: I like it! Shall we wrap it in function, so it is easier to understand this step?

tests/tests_app/cli/test_cli.py
```diff
@@ -52,7 +52,7 @@ def test_commands(command):
 
 def test_main_lightning_cli_no_arguments():
     """Validate the Lightning CLI without args."""
-    res = os.popen("python -m lightning").read()
+    res = os.popen("lightning").read()
```
Borda, line 55: I think we shall test both

.github/workflows/ci-pkg-install.yml
carmocca: If PACKAGE_NAME is not set, it defaults to "lightning" after the changes in this PR

tests/tests_app/cli/test_cli.py
carmocca: -m will fail when PACKAGE_NAME=app

setup.py
carmocca: This logic will change in followups, so I think its better to leave it unrolled for now to facilitate changes. I can add comments if a particular step is not clear

setup.py
```diff
@@ -87,19 +69,30 @@ def _load_py_module(name: str, location: str) -> ModuleType:
...
-        _SETUP_TOOLS._load_aggregate_requirements(_PATH_REQUIRE, _FREEZE_REQUIREMENTS)
-
-    _SETUP_TOOLS.create_mirror_package(os.path.join(_PATH_ROOT, "src"), _PACKAGE_MAPPING)
-
-    _SETUP_MODULE = _load_py_module(name="pkg_setup", location=_PATH_SETUP)
-    _SETUP_MODULE._adjust_manifest(pkg_name=_REAL_PKG_NAME)
-    setup(**_SETUP_MODULE._setup_args(pkg_name=_REAL_PKG_NAME))
+    setup_tools = _load_py_module(name="setup_tools", location=os.path.join(".actions", "setup_tools.py"))
+
+    package_to_install = _PACKAGE_NAME or "lightning"
```
justusschock, line 75:
```suggestion
    package_to_install = _PACKAGE_NAME.lower() or "lightning"
```

nit: just in case (remember the windows debugging :D )

justusschock (approved):
lgtm!

setup.py
carmocca: I would rather avoid this silent conversion. If uppercase is passed, an error is raised mentioning the available values.

tchaton (approved):
LGTM !

akihironitta (approved):

carmocca:
There's been a small change to the checkgroup config: https://github.com/Lightning-AI/lightning/pull/15493/commits/f81ae5984bfeec00a2020c66d3b655f985d8709e so this will need force-merge. CI was green the commit before: https://github.com/Lightning-AI/lightning/actions/runs/3393882442/jobs/5643540143

.actions/setup_tools.py
```diff
@@ -272,19 +274,3 @@ def _load_aggregate_requirements(req_dir: str = "requirements", freeze_requireme
     requires = list(chain(*requires))
     with open(os.path.join(req_dir, "base.txt"), "w") as fp:
         fp.writelines([ln + os.linesep for ln in requires])
-
-
-def set_actual_version_from_src(req_path: str, src_root: str, pkg_name: str) -> None:
-    """Setting actual version from source code for a given package."""
-    with open(req_path, encoding="utf-8") as fo:
-        lines = fo.readlines()
-    ver = parse_version_from_file(os.path.join(src_root, pkg_name.replace("-", "_")))
```
Borda, line 281: let me see how the version alignment is done :)

.github/workflows/ci-app-tests.yml
```diff
@@ -102,8 +102,7 @@ jobs:
       - name: Adjust examples
         if: ${{ matrix.pkg-name != 'lightning' }}
         run: |
-          python .actions/assistant.py copy_replace_imports --source_dir="./examples" --source_import="lightning.app" --target_import="lightning_app"
-          python .actions/assistant.py copy_replace_imports --source_dir="./examples" --source_import="lightning" --target_import="lightning_app"
+          python .actions/assistant.py copy_replace_imports --source_dir="./examples" --source_import="lightning.app,lightning" --target_import="lightning_app,lightning_app"
```
Borda, line 105: This was not needed, U remember @carmocca telling me not to edit what is unrelated the other day :D

requirements/pytorch/base.txt
```diff
@@ -3,7 +3,6 @@
 
 numpy>=1.17.2, <1.23.1
 torch>=1.9.*, <1.13.0
-lightning-lite  # actual version comes from src
```
Borda, line 6: let me see how the version alignment is done :)
ok, the issue is documented in #15489

src/lightning_app/testing/testing.py
```diff
@@ -137,9 +137,12 @@ def application_testing(
...
 
-    with mock.patch("lightning.LightningApp", lightning_app_cls), mock.patch(
-        "lightning_app.LightningApp", lightning_app_cls
-    ):
+    patch1 = mock.patch("lightning_app.LightningApp", lightning_app_cls)
+    # we need to patch both only with the mirror package
+    patch2 = (
+        mock.patch("lightning.LightningApp", lightning_app_cls) if "lightning.app" in sys.modules else nullcontext()
+    )
+    with patch1, patch2:
```
Borda, line 145:
```suggestion
    _patch1 = mock.patch("lightning_app.LightningApp", lightning_app_cls)
    # we need to patch both only with the mirror package
    _patch2 = (
        mock.patch("lightning.LightningApp", lightning_app_cls) if "lightning.app" in sys.modules else nullcontext()
    )
    with _patch1, _patch2:
```

tests/tests_app/cli/test_cli.py
Borda: can we write an issue about it? this seems we shall notify users...
cc: @manskx

Borda (approved):

.github/workflows/ci-app-tests.yml
carmocca: Fair. I'll revert it. I believe I had other changes in this file originally, but were removed

src/lightning_app/testing/testing.py
carmocca: Why this suggestion? These are local variables in a function

tests/tests_app/cli/test_cli.py
carmocca: It's not an issue but expected.

`python -m foo` bypasses the `console_scripts` entrypoint and calls the module `foo` directly. With `PACKAGE_NAME=app` we don't install the `lightning` module so the failure is working as expected

.github/workflows/ci-app-tests.yml
carmocca: I'll leave it since it avoids calling the function twice unnecesarily

src/lightning_app/testing/testing.py
Borda: ok I missed as it, too zoom code :D

PR #15508, opened 2022-11-03
CI: resolving Docs

Borda:
## What does this PR do?

Debugging another force merge without waiting on CI
Resolves #15473

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [x] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @rohitgr7 @Felonious-Spellfire @tchaton @carmocca @akihironitta

PR #15473, opened 2022-11-02
docs updates 1/n

williamFalcon:
make docs great again

PR #15510, opened 2022-11-03
Checkout the HEAD SHA for docs builds

carmocca:
## What does this PR do?

https://github.com/Lightning-AI/lightning/commit/0a3db23cc4471d00ab94d3976579b331443470dc#diff-02dbc35a6a8a8f65d291727a704a253056019b7ea277be0160c4843567def6e3R7 changed the event type to `pull_request_target`. 

This event type has the advantage sharing secrets with forks. This is important so fork PRs can build docs or they would be blocked.
By design, this event type runs using master's changes (the PR base) for security reasons.

One workaround for this (at the cost of security) is to checkout the PR head SHA.

Not doing this is what caused #15473 to appear as green when in reality the changes were broken.
**Since that has been merged and the errors now appear in master, this PR needs to be force merged. Failures here are expected and caused by #15473**

This will help unblock #15508 

None

cc @tchaton @rohitgr7 @carmocca @akihironitta @borda

### DIFF
--- a/.github/workflows/docs-checks.yml
+++ b/.github/workflows/docs-checks.yml
@@ -4,6 +4,8 @@ name: Check Docs
 on:
   push:
     branches: ["master", "release/*"]
+  # use this event type to share secrets with forks.
+  # it's important that the PR head SHA is checked out to run the changes
   pull_request_target:
     branches: ["master", "release/*"]
     paths:
@@ -30,6 +32,8 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
       - uses: actions/setup-python@v4
         with:
           python-version: "3.8"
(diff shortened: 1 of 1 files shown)


Borda (approved):

carmocca:
@Borda Turns out this solution is not complete. The contents of the repository are checked out to the latest versions, but not the workflow changes themselves (if present).

We'll need to find a different solution:
- Remove secrets by making the template public (no longer need `_target`)
- Do not allow docs PRs from contributors (remove `_target`)
- Port docs build setup to Azure

Borda:
> Turns out this solution is not complete.

It was complete, and it was known that the CI can't be edited in the same PR s the package can not be leaked, as the same applies to TPU tests...

> We'll need to choose a different solution:
a) Avoid the need for secrets by making the template public (no longer need _target)
b) Do not allow docs PRs from contributors (remove _target)
c) Port docs build setup to Azure

yes, that is what I asked @williamFalcon two days ago too :)
but corrections:
add A) we do not make a template public but the built wheel package
add B) it is not correct; your app docs as required check for add app code changes so you effectively ban any contributor editing app's codebase :/
add C) still, the package can be leaked so almost the same as A)

PR #15523, opened 2022-11-04
[App] Add start method to the LightningWork 

### LINKED ISSUE #15015: Keep a LightningWork "Hot" Until Run Method is Called
## 🚀 Feature

When running a Lightning App in the cloud, Work machines are instantiated lazily once their `run()` method is called for the first time. Spinning up a Work machine has a non-trivial time cost (>1 min). It would be nice if the Work machine could be spun up non-lazily, so that the first call to the `run()` method is fast.

### Motivation

Say I am building a speech recognition app. The architecture design that I've chosen uses a pool of dynamic `SpeechRecognizer` Works. As requests come into the app (via custom Lightning Commands), the Flow calls `self.speech_recognizer.run(request)` which kicks off the Work to run the model against the input. This means the first request will not complete until the Work machine has spun up.

### Pitch

The specific API doesn't matter much to me, but something like this would work:

```python
class App(LightningFlow):
    ...
    def run(self):
        if not self.speech_recognizer.hot():
            self.speech_recognizer.warm()

        self.speech_recognizer.run(request)
```

### Alternatives

Some would suggest using the `run()` method to create a long-running inference API endpoint, instead of running the model on each call. That defeats the purpose of some of the core value propositions of the Lightning Framework. The `run()` method is much simpler and the Framework already provides all the underlying networking between the Flow/Works. I shouldn't have to write another HTTP server on top of using Lightning when it already provides a much simpler Python interface.

### Additional context

N/A

tchaton:
## What does this PR do?

This PR adds a `start` method of the Lightning Work, enabling to get the machine warm and then actually execute its run method once ready.

```python
class WorkStart(LightningWork):
    def __init__(self):
        super().__init__(parallel=True)
        self.counter = 0

    def run(self):
        self.counter += 1


class FlowStart(LightningFlow):
    def __init__(self):
        super().__init__()
        self.w = WorkStart()

    def run(self):
        if self.w.status.stage == WorkStageStatus.NOT_STARTED:
            self.w.start()
        if self.w.status.stage == WorkStageStatus.STARTED:
            self.w.run()
        if self.w.counter == 1:
            self._exit()


def test_lightning_app_work_start():
    app = LightningApp(FlowStart())
    MultiProcessRuntime(app, start_server=False).dispatch()
```

Fixes #15015

## Before submitting

## PR review

## Did you have fun?

cc @borda

### DIFF
--- a/src/lightning_app/core/work.py
+++ b/src/lightning_app/core/work.py
@@ -539,6 +539,10 @@ class LightningWork:
 
     def start(self):
         """Starts LightingWork component via L.CloudCompute."""
+        if self.status.stage == WorkStageStatus.STOPPED:
+            raise Exception("A work can be started only once for now.")
+
+        # This enables to start the run method with a phony input and exit.
         self.run(Action(method="start"))
 
     def run(self, *args, **kwargs):
--- a/src/lightning_app/core/work.py
+++ b/src/lightning_app/core/work.py
@@ -30,7 +30,7 @@ from lightning_app.utilities.packaging.cloud_compute import (
     _maybe_create_cloud_compute,
     CloudCompute,
 )
-from lightning_app.utilities.proxies import LightningWorkSetAttrProxy, ProxyWorkRun, unwrap
+from lightning_app.utilities.proxies import Action, LightningWorkSetAttrProxy, ProxyWorkRun, unwrap
 
 
 class LightningWork:
@@ -536,6 +536,10 @@ class LightningWork:
                     if status not in final_statuses:
                         final_statuses.append(status)
                 calls[call_hash]["statuses"] = final_statuses
+
+    def start(self):
+        """Starts LightingWork component via L.CloudCompute."""
+        self.run(Action(method="start"))
 
     def run(self, *args, **kwargs):
         """Override to add your own logic.
--- a/tests/tests_app/core/test_lightning_work.py
+++ b/tests/tests_app/core/test_lightning_work.py
@@ -350,8 +350,8 @@ def test_lightning_app_with_work():
 
 
 class WorkStart(LightningWork):
-    def __init__(self):
-        super().__init__(parallel=True)
+    def __init__(self, cache_calls, parallel):
+        super().__init__(cache_calls=cache_calls, parallel=parallel)
         self.counter = 0
 
     def run(self):
@@ -359,9 +359,9 @@ class WorkStart(LightningWork):
 
 
 class FlowStart(LightningFlow):
-    def __init__(self):
+    def __init__(self, cache_calls, parallel):
         super().__init__()
-        self.w = WorkStart()
+        self.w = WorkStart(cache_calls, parallel)
         self.finish = False
 
     def run(self):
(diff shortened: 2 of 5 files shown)


src/lightning_app/core/work.py
```diff
@@ -537,6 +537,14 @@ def _cleanup_calls(calls: Dict[str, Any]):
                         final_statuses.append(status)
                 calls[call_hash]["statuses"] = final_statuses
 
+    def start(self):
+        """Starts LightingWork component via L.CloudCompute."""
+        if self.status.stage == WorkStageStatus.STOPPED:
```
lantiga, line 542: shouldn't we include anything by `NOT_STARTED`?

src/lightning_app/core/work.py
```diff
@@ -537,6 +537,14 @@ def _cleanup_calls(calls: Dict[str, Any]):
...
+            raise Exception("A work can be started only once for now.")
+
+        # This enables to start the run method with a phony input and exit.
+        self.run(Action(method="start"))
```
lantiga, line 546: ok, so instead of forcing `parallel=True` `cached=True` around the call, we are overriding the logic in `run_once` to effectively have that result, is that correct?

and we can extend the mechanism to other actions (which may have similar semantics or not, we'll see about that)

tests/tests_app/core/test_lightning_work.py
```diff
@@ -347,3 +347,38 @@ def on_before_run_once(self):
 def test_lightning_app_with_work():
     app = LightningTestAppWithWork(WorkCounter())
     MultiProcessRuntime(app, start_server=False).dispatch()
+
+
+class WorkStart(LightningWork):
+    def __init__(self):
+        super().__init__(parallel=True)
```
lantiga, line 354: let's add tests to cover all combinations of semantics (minus `parallel=False` `caching=False`)

lantiga (changes requested):
Super neat, a couple of comments

src/lightning_app/core/work.py
lantiga: I see it now, the others are handled by the implicit caching semantics

tchaton: This would break if we do so, as user won't be protecting their `.start` method. Therefore, we need to rely on the cache mechanism of the run method.

src/lightning_app/core/work.py
tchaton: Yes, this is a skateboard version to enable the feature. We would need a bigger refactor to suport this in a cleaner way, but this unblocks MultiNode example.

tests/tests_app/core/test_lightning_work.py
tchaton: Done !

lantiga (approved):
Looks good!

rohitgr7 (approved):

src/lightning_app/core/work.py
```diff
@@ -537,6 +537,14 @@ def _cleanup_calls(calls: Dict[str, Any]):
                         final_statuses.append(status)
                 calls[call_hash]["statuses"] = final_statuses
 
+    def start(self):
```
rohitgr7, line 540: qq: this is call blocking?

for eg:
```py
class Someflow(LightningFlow):
    def run(self):
        self.my_work.start(). <- is this blocking in the sense will the program move forward or wait for provisioning?
        print('something') 
```

tchaton: This depends if your work is running with `parallel=True` or not. But default, it is blocking.

manskx (approved):

PR #15525, opened 2022-11-04
remove source-lit docs

williamFalcon:
not needed

cc @borda @rohitgr7


carmocca (approved):
I guess this is fine since they are not used. Can always be re-done if necessary

Borda (approved):

carmocca:
Some leftovers were missing. Opened https://github.com/Lightning-AI/lightning/pull/15527 removing them

PR #15536, opened 2022-11-04
Add back `pytorch_lightning.utilities.distributed.rank_zero_only` but deprecate it

### LINKED ISSUE #15509: rank_zero_only removal violate backward compatibillity
cc @borda @justusschock

moghadas76:
## What does this PR do?

- Fix: Revert  lightning_lite.utilities.rank_zero_only to preserve backward compatibility

Closes #15509


awaelchli:
Haven't the rank zero utilities always been defined in `utilities/rank_zero.py`?

awaelchli:
In 1.6, we moved these and added a warning:
```py

def rank_zero_info(*args: Any, **kwargs: Any) -> Any:
    rank_zero_deprecation(
        "pytorch_lightning.utilities.distributed.rank_zero_info has been deprecated in v1.6"
        " and will be removed in v1.8."
        " Use the equivalent function from the pytorch_lightning.utilities.rank_zero module instead."
    )
    return new_rank_zero_info(*args, **kwargs)


def rank_zero_debug(*args: Any, **kwargs: Any) -> Any:
    rank_zero_deprecation(
        "pytorch_lightning.utilities.distributed.rank_zero_debug has been deprecated in v1.6"
        " and will be removed in v1.8."
        " Use the equivalent function from the pytorch_lightning.utilities.rank_zero module instead."
    )
    return new_rank_zero_debug(*args, **kwargs)

```

Is this the issue you are facing? You are upgrading to 1.8 from a version before 1.6?

moghadas76:
Yes, but violates BC! I cannot downgrade in the production

carmocca:
Exactly. Imports are not public API

moghadas76:
I love contributing to this repo, so, could you formally formulate what changes I could make?

Borda:
> Exactly. Imports are not public API

Imports do not have any separation as private or public, they are not, but what you are importing can be public/protected/private... and in this case seems we were importing public API

src/pytorch_lightning/utilities/distributed.py
```diff
@@ -27,7 +27,14 @@
 from lightning_lite.utilities.distributed import _init_dist_connection as new_init_dist_connection
 from lightning_lite.utilities.distributed import _sync_ddp as new_sync_ddp
 from lightning_lite.utilities.distributed import _sync_ddp_if_available as new_sync_ddp_if_available
-from pytorch_lightning.utilities.rank_zero import rank_zero_debug, rank_zero_deprecation, rank_zero_info
+from pytorch_lightning.utilities.rank_zero import (  # BACKWARD COMPATIBILITY
```
Borda, line 30: add when this will be removed

Borda (approved):

lantiga (approved):

carmocca:
I created https://github.com/Lightning-AI/lightning/pull/15569 to provide a better error message for those still importing these functions. The removal of them followed our regular backward-compatibility process, as they were deprecated from 1.6 to 1.8

Borda:
> I created #15569 to provide a better error message for those still importing these functions. The removal of them followed our regular backward-compatibility process, as they were deprecated from 1.6 to 1.8

ok, let's merge this so we have the fix in place and continue with #15569

carmocca:
I had a longer look. It's actually our fault as we removed `rank_zero_only` in https://github.com/Lightning-AI/lightning/commit/d2459df2ff79184c596b2f5c865ffb17c3541307#diff-f0e5bea63aa4d620b1480af939a7d0218ede26d39f889112f4ed5a87e2b41f2dL26. `rank_zero_info` still works because the module still imports it in 1.8. I'll update this PR by adding a deprecation for the breaking change

carmocca (approved):

moghadas76:
Thanks all you guys

PR #15548, opened 2022-11-05
Forbid Lite installation collision

carmocca:
## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning/issues/15489

Breaks Lite installation if PyTorch Lightning is already available

### DIFF
--- a/src/lightning_lite/__setup__.py
+++ b/src/lightning_lite/__setup__.py
@@ -1,4 +1,5 @@
 import os
+from distutils.version import LooseVersion
 from importlib.util import module_from_spec, spec_from_file_location
 from types import ModuleType
 from typing import Any, Dict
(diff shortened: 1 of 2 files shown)


src/lightning_lite/__setup__.py
```diff
@@ -60,7 +61,27 @@ def _adjust_manifest(**__: Any) -> None:
         fp.writelines(lines)
 
 
+def _check_install_collision() -> None:
+    try:
+        import pytorch_lightning
```
carmocca, line 66: Turns out we cannot import like this unless we use `pip install --no-build-isolation`: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-no-build-isolation

I'm going to try with a subprocess call

carmocca:
```py
        version = subprocess.check_output(
            "python -c 'import pytorch_lightning; print(pytorch_lightning.__version__)'",
            shell=True,
        )
```

Suffers from the same issue.

I don't think we can reliably check this.

PR #15559, opened 2022-11-06
CI: simplify groups ref/pointer

Borda:
## What does this PR do?

reduce duplicated sections which are sensitive for human error and improve readability

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/.github/checkgroup.yml
+++ b/.github/checkgroup.yml
@@ -6,63 +6,10 @@ subprojects:
 
   # SECTION: pytorch_lightning
 
-  - id: "pytorch_lightning"
-    paths:
-      # all examples don't need to be added because they aren't used in CI, but these are
-      - "examples/run_ddp_examples.sh"
-      - "examples/convert_from_pt_to_pl/**"
-      - "examples/run_pl_examples.sh"
-      - "examples/pl_basics/backbone_image_classifier.py"
-      - "examples/pl_basics/autoencoder.py"
-      - "examples/pl_loops/mnist_lite.py"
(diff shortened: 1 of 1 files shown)


akihironitta (approved):
This will be very nice if it works!

carmocca (changes requested):
Blocking until I verify if this works! Thanks for the suggestion

Borda:
> This will be very nice if it works!

in theory, it should, as it is using YAML native directive/syntax

BUT it seems that with this, the referred lists are loaded as sublists, so not expanded as a single flattened list :(
```py
import yaml
yaml.safe_load(".github/checkgroup.yml")
```

carmocca:
I checked and we use https://probot.github.io/api/7.0.1/classes/context.html#config to load the configuration file. There's no mention of any specific parsing constraints so I assume it will work just as `yaml.safe_load()`

Looking at https://stackoverflow.com/questions/24090177/how-to-merge-yaml-arrays, it looks like flattening these list is not possible.

The only solution I can think of is adding support for flattening in the javascript implementation of checkgroup so that we can use this notation: https://github.com/Lightning-AI/probot/blob/babecd77266ed8ac540de7f5e17804030174215c/src/check-group/utils/user_config_parser/populate_subprojects.ts#L63-L91

Borda:
> The only solution I can think of is adding support for flattening in the javascript implementation of checkgroup

Let's keep it simple, personally, it may be food to by default flatten all lists in check and paths as there is no meaning for a tree, but if it is not a few clicks away, let's drop it for now :rabbit:

PR #15581, opened 2022-11-08
Fix which groups requires docs builds

carmocca:
## What does this PR do?

https://github.com/Lightning-AI/lightning/pull/15317 only modifies tests, but checkgroup expects that the docs job runs. But it shouldn't since it modifies only the tests

The same change is applied to apps

None

### DIFF
--- a/.github/checkgroup.yml
+++ b/.github/checkgroup.yml
@@ -48,8 +48,6 @@ subprojects:
       #- "pl-cpu (ubuntu-20.04, lightning, 3.7, 1.9, oldest)"
       - "pl-cpu (windows-2022, lightning, 3.10, 1.12)"
       #- "pl-cpu (windows-2022, lightning, 3.7, 1.9, oldest)"
-      - "make-doctest (pytorch)"
-      - "make-html (pytorch)"
       - "pytorch-lightning (GPUs)"
       - "pytorch-lightning (HPUs)"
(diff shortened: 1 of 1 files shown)


justusschock (approved):

.github/checkgroup.yml
```diff
@@ -203,8 +202,6 @@ subprojects:
       #- "pl-cpu (ubuntu-20.04, lightning, 3.7, 1.9, oldest)"
       - "pl-cpu (windows-2022, lightning, 3.10, 1.12)"
       #- "pl-cpu (windows-2022, lightning, 3.7, 1.9, oldest)"
-      - "make-doctest (pytorch)"
-      - "make-html (pytorch)"
```
awaelchli, line 207: Are there any implications here? What if the source code removes something that is referenced in the docs. This could generate a reference error in the docs build, and it would only show up on master.

awaelchli (approved):

.github/checkgroup.yml
carmocca: That's why I added https://github.com/Lightning-AI/lightning/pull/15581/files#diff-5ca41bd8852d2c309825fa76269fb299e7a77b075918a9c56fa1990c40e48f7bR122

Borda (approved):

Issue #15388, opened 2022-10-28
Prefix Lightning App environment variables


awaelchli:
## Proposed refactor

Prefix environment variables to indicate they are set by Lightning.

### Motivation

- For debugging purposes: I can easily grep for the prefix
- Avoids accidental clashes (which could be incredibly difficult to debug).

### Pitch

Rename all environment variables by prefixing "LAI_"

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/Lightning-AI/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/Lightning-AI/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/Lightning-AI/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging PyTorch Lightning, Transformers, and Hydra.

divinit7:
I'd like to take this up. I would really appreciate some clarification on the same, I'm looking through all the export commands in the code, how do I know which specific ones to update?

PR #15590, opened 2022-11-08
[App] Enable debugger with LightningApp

tchaton:
## What does this PR do?

This PR enables running an app within a debugger with breakpoints with flows and works without your changing code :)  

https://user-images.githubusercontent.com/12861981/200654176-29fed032-dd1a-4b6f-8d11-9aded56fe59b.mp4

## Before submitting

## PR review

## Did you have fun?

cc @borda


ethanwharris (approved):
LGTM 😃

src/lightning_app/core/app.py
```diff
@@ -70,7 +70,7 @@ def __init__(
             root: The root ``LightningFlow`` or ``LightningWork`` component, that defines all the app's nested
                  components, running infinitely. It must define a `run()` method that the app can call.
             flow_cloud_compute: The default Cloud Compute used for flow, Rest API and frontend's.
-            debug: Whether to activate the Lightning Logger debug mode.
+            log_level: Whether to activate the Lightning Logger debug mode.
```
lantiga, line 73: Change log string to something like “The log level for the app, one of [‘info’, ‘debug’]”

src/lightning_app/core/app.py
```diff
@@ -162,6 +165,12 @@ def __init__(
         # this should happen once for all apps before the ui server starts running.
         frontend.update_index_file(FRONTEND_DIR, info=info, root_path=root_path)
 
+        if debugger_is_active() and not bool(int(os.getenv("LIGHTNING_DISPATCHED", "0"))):
+            os.environ["LIGHTNING_DISPATCHED"] = "1"
+            from lightning_app.runners import MultiProcessRuntime
+
+            MultiProcessRuntime(self).dispatch()
```
lantiga, line 172: should we control whether we are running on the cloud and emit an error in that case? probably it won’t happen since gettrace will be very likely to return None in that case, but better safe than sorry

lantiga (approved):
Great! Docstring change and a comment

tests/tests_app/conftest.py
```diff
@@ -20,6 +20,8 @@
     "template_react_ui": "https://github.com/Lightning-AI/lightning-template-react.git",
 }
 
+os.environ["LIGHTNING_DISPATCHED"] = "1"
+
```
awaelchli, line 24: Why is it needed to set this here?

src/lightning_app/core/app.py
```diff
@@ -162,6 +165,12 @@ def __init__(
         # this should happen once for all apps before the ui server starts running.
         frontend.update_index_file(FRONTEND_DIR, info=info, root_path=root_path)
 
+        if debugger_is_active() and not bool(int(os.getenv("LIGHTNING_DISPATCHED", "0"))):
```
awaelchli, line 168: see #15388

awaelchli (commented):
I don't understand how this is supposed to work. I checked out the branch, set a `breakpoint()` in the `examples/app_boring` in one of the works, and it exited with an error. 

Perhaps you only meant to add support for the debugger in the IDE. I verified that it works in PyCharm with the debugger, but the user will have to change their code from
```py
    app = L.LightningApp(BoringApp())
```
to
```py
    app = L.LightningApp(BoringApp())
    MultiProcessRuntime(app).dispatch()
```
in order to run it as a regular python program without cli. 
However, we never explain to the user that this is another way of launching an app locally. Using the runtime directly like that is only meant for our testing. We would need to introduce a simpler API here to do the same if we really want to support running it like that.

tchaton:
> I don't understand how this is supposed to work. I checked out the branch, set a `breakpoint()` in the `examples/app_boring` in one of the works, and it exited with an error.
> 
> Perhaps you only meant to add support for the debugger in the IDE. I verified that it works in PyCharm with the debugger, but the user will have to change their code from
> 
> ```python
>     app = L.LightningApp(BoringApp())
> ```
> 
> to
> 
> ```python
>     app = L.LightningApp(BoringApp())
>     MultiProcessRuntime(app).dispatch()
> ```
> 
> in order to run it as a regular python program without cli. However, we never explain to the user that this is another way of launching an app locally. Using the runtime directly like that is only meant for our testing. We would need to introduce a simpler API here to do the same if we really want to support running it like that.

This is meant to work if you run your script with a debugger attached to it. This enables to debug very simply your app directly within the IDE. You don't need to change single line of code for it to work, this is the entire goal here.

> in order to run it as a regular python program without cli.
This isn't what we are trying to enable. Just for users to run their app in their IDE without code modification and set breakpoints wherever they decide.

tests/tests_app/conftest.py
tchaton: It is safer in case you launch lightning with a debugger attached.

src/lightning_app/core/app.py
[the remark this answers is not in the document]
tchaton: I have checked and didn't find one.

src/lightning_app/core/app.py
tchaton: Not sure to follow you there. There is a `LIGHTNING` prefix. What would you prefer ?

src/lightning_app/core/app.py
tchaton: Done !

src/lightning_app/core/app.py
tchaton: Done !

awaelchli:
Then please could you show us how. I don't understand how this should be used. Your video starts with the debugger already running.

tchaton:
> Then please could you show us how? I don't understand how this should be used. Your video starts with the debugger already running.

Oh apology, the original video was too big. I couldn't upload it entirely.


https://user-images.githubusercontent.com/12861981/200850721-bd4de48c-ec4d-43b0-902c-d4e7339a9b95.mov

https://user-images.githubusercontent.com/12861981/200850768-e94dcfe8-d327-4f59-a548-17459b86c8d3.mov

src/lightning_app/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 - Added a `start_with_flow` flag to the `LightningWork` which can be disabled to prevent the work from starting at the same time as the flow ([#15591](https://github.com/Lightning-AI/lightning/pull/15591))
 
+- Added support for running Lightning App with IDE debugger ([#15590](https://github.com/Lightning-AI/lightning/pull/15590))
```
awaelchli, line 21: As discussed offline, let's make it clear this only works for VSCode right now (debugpy).

awaelchli (approved):

src/lightning_app/CHANGELOG.md
tchaton: Done !

PR #15611, opened 2022-11-10
Delete unused TPU CI files

carmocca:
## What does this PR do?

See title. We had some unused Dockerfiles


carmocca:
The TPU failures are because the TPU job uses `pull_request_target`, so the file does not show as moved.

akihironitta (approved):

Borda (approved):

PR #15622, opened 2022-11-10
Enable Probot CheckGroup v3

carmocca:
## What does this PR do?

Uses https://github.com/Lightning-AI/probot/pull/3

- Flatten the check lists to support the `&foo`, `*foo` notation to reuse sections (discussed in https://github.com/Lightning-AI/lightning/pull/15559#issuecomment-1307104421)
- Pull the checkgroup config from the PR itself as long as it's not from a fork (tested in #15635)

### DIFF
--- a/.github/checkgroup.yml
+++ b/.github/checkgroup.yml
@@ -6,52 +6,10 @@ subprojects:
 
   # SECTION: pytorch_lightning
 
-  - id: "pytorch_lightning"
-    paths:
-      # all examples don't need to be added because they aren't used in CI, but these are
-      - "examples/run_pl_examples.sh"
-      - "examples/pl_basics/backbone_image_classifier.py"
-      - "examples/pl_basics/autoencoder.py"
-      - "examples/pl_fault_tolerant/automatic.py"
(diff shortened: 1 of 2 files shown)


.github/checkgroup.yml
```diff
@@ -6,52 +6,10 @@ subprojects:
...
-      - "pl-cpu (slow, ubuntu-20.04, pytorch, 3.7, 1.11)"
-      - "pl-cpu (slow, windows-2022, pytorch, 3.7, 1.11)"
-      # TODO: since this job has intermittent availability, it cannot be required or it will block all PL PRs from forks
-      #- "test-on-tpus"
-
   - id: "pytorch_lightning: CPU workflow"
     paths:
       - ".github/workflows/ci-pytorch-tests.yml"
-    checks:
+    checks: &checks_pl_cpu
```
carmocca, line 12: None of these changes are necessary if #15633 lands. But at least it can be used now.

akihironitta (approved):

Borda (approved):

PR #15606, opened 2022-11-09
Checkpoint migration for `ModelCheckpoint` state-key changes

awaelchli:
## What does this PR do?

A breaking change was introduced in #15300 regarding the state loading for the model checkpoint callback. Checkpoints that have the the state saved with `save_on_train_epoch_end=True|False` can still be loaded, but the ModelCheckpoint won't load its state correctly, since the code and what's saved in the checkpoint do not match.

We can fix this issue by dropping the `save_on_train_epoch_end` from the state key. This way, all old states from ModelCheckpoint can be loaded successfully without ambiguity in what `save_on_train_epoch_end` should be set to.
An extremely rare edge case in which a migration is not possible is also handled with a user warning.

Full repro script:
```py

import os

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer
from pytorch_lightning.callbacks import ModelCheckpoint


class RandomDataset(Dataset):
    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)

    def forward(self, x):
        return self.layer(x)

    def on_train_start(self) -> None:
        print("The callback states at the beginning of training are")
        for c in self.trainer.callbacks:
            print(c.state_key, c.state_dict())

    def on_train_end(self) -> None:
        print("The callback states at the end of training are")
        for c in self.trainer.callbacks:
            print(c.state_key, c.state_dict())

    def training_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def validation_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("valid_loss1", loss)
        self.log("valid_loss2", loss)

    def test_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("test_loss", loss)

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)


def run():
    train_data = DataLoader(RandomDataset(32, 64), batch_size=2)
    val_data = DataLoader(RandomDataset(32, 64), batch_size=2)

    checkpoint1 = ModelCheckpoint(monitor="valid_loss1")  # in v1.7 -> True
[...]

PR #15634, opened 2022-11-11
Validate that state-key is unique when using multiple callbacks of the same type

awaelchli:
## What does this PR do?

If you give two different callback instances with the same state key to the Trainer, they overwrite each other when saving a checkpoint. Only one of two states will actually be in the checkpoint. Hence, the trainer would fail to resume all states correctly (silently). 

This PR validates this at the time of the trainer instantiation. 

This change is required to continue with an improved version of #15606 to make checkpoint migration better.

If someone was affected by this but weren't noticing it, when they upgrade Lighting they will get a friendly error now. Apart from that change, there are no intended breaking behaviors.

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli

### DIFF
--- a/src/pytorch_lightning/trainer/connectors/callback_connector.py
+++ b/src/pytorch_lightning/trainer/connectors/callback_connector.py
@@ -18,6 +18,8 @@ from collections import defaultdict
 from datetime import timedelta
 from typing import Dict, List, Optional, Sequence, Union
 
+from lightning_utilities.core.overrides import is_overridden
+
 import pytorch_lightning as pl
 from pytorch_lightning.callbacks import (
     Callback,
@@ -295,7 +297,7 @@ def _configure_external_callbacks() -> List[Callback]:
 
 
 def _validate_callbacks_list(callbacks: List[Callback]) -> None:
-    stateful_callbacks = [c for c in callbacks if c.state_dict()]
+    stateful_callbacks = [cb for cb in callbacks if is_overridden("state_dict", instance=cb, parent=Callback)]
     seen_callbacks = defaultdict(list)
     for callback in stateful_callbacks:
         if callback.state_key in seen_callbacks:
--- a/src/pytorch_lightning/trainer/connectors/callback_connector.py
+++ b/src/pytorch_lightning/trainer/connectors/callback_connector.py
@@ -17,7 +17,6 @@ import os
 from datetime import timedelta
 from typing import Dict, List, Optional, Sequence, Union
 
-
 import pytorch_lightning as pl
 from pytorch_lightning.callbacks import (
     Callback,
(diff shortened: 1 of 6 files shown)


src/pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -290,3 +294,18 @@ def _configure_external_callbacks() -> List[Callback]:
         )
         external_callbacks.extend(callbacks_list)
     return external_callbacks
+
+
+def _validate_callbacks_list(callbacks: List[Callback]) -> None:
+    stateful_callbacks = [cb for cb in callbacks if is_overridden("state_dict", instance=cb, parent=Callback)]
```
carmocca, line 300:
```suggestion
    stateful_callbacks = [cb for cb in callbacks if is_overridden("state_dict", cb)]
```

src/pytorch_lightning/trainer/connectors/callback_connector.py
```diff
@@ -290,3 +294,18 @@ def _configure_external_callbacks() -> List[Callback]:
...
+    seen_callbacks = defaultdict(list)
```
carmocca, line 301: Why a defaultdict instead of a plain old dict?

Why do we need a list? This enforces there's only 1 callback per state_key

carmocca (approved):

src/pytorch_lightning/trainer/connectors/callback_connector.py
awaelchli: Oh, I don't need it anymore. Originally I wanted to print more information in the error message. We can just make it a set of state_keys.

src/pytorch_lightning/trainer/connectors/callback_connector.py
awaelchli: I think we need the parent here. What if we have this inheritance: A > B > C where A is the super class of B and B is super class of C. If B implements state_dict, but C doesn't, I still want to categorize this as stateful.

carmocca: Oh, you are thinking that `is_overridden` is smarter than it is and actually checks the inheritance tree. Not the case: https://github.com/Lightning-AI/lightning/blob/43fd9c1534411a80d5364031b3e529fef02a930c/src/pytorch_lightning/utilities/model_helpers.py#L31-L32

awaelchli: Double oh, I was importing from lightning_utilities and this one isn't smart. Agree, I should be using the one from PL and then I can drop the parent! Great!

Borda (approved):

PR #15642, opened 2022-11-11
[App] Resolve bi-directional queue bug

tchaton:
## What does this PR do?

There is a bug where the Drive Component name changes during setattr to work, which is detected as a change.
This was sending a delta to work and failing as this would be assigned outside the run or init scope.

This PR ignores Drive changes as they are likely to be sent through the run method already.

```python
AttributeError: Only JSON-serializable attributes are currently supported (str, int, float, bool, tuple, list, dict etc.) to be part of <echo.components.fileserver.FileServer object at 0x144f79640> state. Found the attribute drive with {'type': '__drive__', 'id': 'echo', 'protocol': 'lit://', 'allow_duplicates': False, 'component_name': 'root.fileserver', 'root_folder': PosixPath('/Users/thomas/Documents/GitHub/LAI-Echo-App')} instead. 
HINT: Private attributes defined as follows `self._x = y` won't be shared between components and therefore don't need to be JSON-serializable. If you need to include non-JSON serializable objects in the state, you can use the `lightning.app.storage.Payload` API.
None
Exception in thread Thread-2:
  File "/Users/thomas/.pyenv/versions/3.8.5/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/Users/thomas/Documents/GitHub/lightning/src/lightning/app/utilities/proxies.py", line 258, in run
    self.run_once()
  File "/Users/thomas/Documents/GitHub/lightning/src/lightning/app/utilities/proxies.py", line 294, in run_once
    raise e
  File "/Users/thomas/Documents/GitHub/lightning/src/lightning/app/utilities/proxies.py", line 290, in run_once
    self._work.apply_flow_delta(Delta(deep_diff, raise_errors=True))
  File "/Users/thomas/Documents/GitHub/lightning/src/lightning/app/core/work.py", line 628, in apply_flow_delta
    self._default_setattr(name, value)
  File "/Users/thomas/Documents/GitHub/lightning/src/lightning/app/core/work.py", line 411, in _default_setattr
    raise AttributeError(
AttributeError: Only JSON-serializable attributes are currently supported (str, int, float, bool, tuple, list, dict etc.) to be part of <echo.components.fileserver.FileServer object at 0x144f79640> state. Found the attribute drive with {'type': '__drive__', 'id': 'echo', 'protocol': 'lit://', 'allow_duplicates': False, 'component_name': 'root.fileserver', 'root_folder': PosixPath('/Users/thomas/Documents/GitHub/LAI-Echo-App')} instead. 
HINT: Private attributes defined as follows `self._x = y` won't be shared between components and therefore don't need to be JSON-serializable. If you need to include non-JSON serializable objects in the state, you can use the `lightning.app.storage.Payload` API.
```

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

cc @borda


lantiga (approved):
This is caveat and I can see developers potentially tripping over this.
I'm wondering if we can at least emit a warning if a change is detected there.

src/lightning_app/components/serve/python_server.py
```diff
@@ -14,7 +14,7 @@
 logger = Logger(__name__)
 
 
-def image_to_base64(image_path):
+def _image_to_base64(image_path):
```
ethanwharris, line 17: Is this unused?

ethanwharris (approved):

src/lightning_app/components/serve/python_server.py
tchaton: Looks like. Let me remove it.

src/lightning_app/components/serve/python_server.py
```diff
@@ -33,7 +28,9 @@ class Image(BaseModel):
 
     @staticmethod
     def _get_sample_data() -> Dict[Any, Any]:
-        imagepath = Path(__file__).absolute().parent / "catimage.png"
+        sep = "\\" if sys.platform == "win32" else "/"
```
carmocca, line 31: You can use `os.sep` here

carmocca (approved):
The hanging e2e's were reverted in https://github.com/Lightning-AI/lightning/pull/15644

src/lightning_app/components/serve/python_server.py
Borda: But this is failing on all platforms.. :(

tchaton: Done !

PR #15651, opened 2022-11-11
Deprecate description and env parameters in LightningCLI.__init__

### LINKED ISSUE #15174: Add default_config_files argument to LightningCLI.__init__
## 🚀 Feature

I suggest to add a `default_config_files` parameter in `LightningCLI.__init__`.

### Motivation

The only way I found to change the default file to load the CLI configuration is though the `parser_kwargs` :
```python
LightningCLI(parser_kwargs={'default_config_files': ['config.yaml']})
```
Which took me a while to find because my first try was to override `add_default_arguments_to_parser` that way :

```python
def add_default_arguments_to_parser(self, parser: LightningArgumentParser) -> None:
         super().add_default_arguments_to_parser(parser)
         parser.set_defaults(config='config.yaml')
```
Which is not working for some reason (related to #15038 ?).

I think that using a configuration file is a common practice (or at least it should be), thus setting up a default path should be straightforward.

### Pitch

It seems more natural to pass this argument directly to `LightningCLI` (it seems to already work that way for `description`, `env_prefix` and `default_env`).

According to what I undestrand the change is fairly easy to make. Just by adding `default_config_files` in `LightningCLI.__init__`, then sending to `_setup_parser_kwargs` in [this section](https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/cli.py#L339):

```python
main_kwargs, subparser_kwargs = self._setup_parser_kwargs(
    parser_kwargs or {},  # type: ignore  # github.com/python/mypy/issues/6463
    {"description": description, "env_prefix": env_prefix, "default_env": env_parse, "default_config_files": default_config_files},
)
```

cc @carmocca @mauvilsa

mauvilsa:
## What does this PR do?

Fixes #15174

It includes the deprecations and the related suggestions in https://github.com/Lightning-AI/lightning/issues/15174#issuecomment-1292159835. I also increased the minimum version of jsonargparse because otherwise mypy gives an error.

Three parameters are deprecated in `LightningCLI.__init__` which should now be given through `parser_kwargs`:
- `description`
- `env_prefix`
- `env_parse`

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/cli.py
```diff
@@ -485,7 +487,7 @@ def _prepare_subcommand_parser(self, klass: Type, subcommand: str, **kwargs: Any
         parser = self.init_parser(**kwargs)
         self._add_arguments(parser)
         # subcommand arguments
-        skip = self.subcommands()[subcommand]
+        skip: Set[Union[str, int]] = set(self.subcommands()[subcommand])
```
carmocca, line 490: Doesn't `self.subcommands()[subcommand]` already return a set?

src/pytorch_lightning/cli.py
```diff
@@ -393,13 +394,14 @@ def init_parser(self, **kwargs: Any) -> LightningArgumentParser:
         )
         return parser
 
-    def setup_parser(
```
carmocca, line 396: Removing these arguments would be considered a breaking change. Can you add a deprecation cycle for them?

src/pytorch_lightning/cli.py
mauvilsa: If `set` is removed, mypy complains because the type is not `Set[Union[str, int]]`.

src/pytorch_lightning/cli.py
mauvilsa: I oversaw this. Why is `setup_parser` public? In which use cases is necessary to override it? I can change it so that we keep `_setup_parser_kwargs` and `setup_parser` still receives `main_kwargs` and `subparser_kwargs`.

src/pytorch_lightning/CHANGELOG.md
```diff
@@ -40,6 +40,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Deprecated
 
+- Deprecated `pytorch_lightning.utilities.distributed.rank_zero_only` in favor of `pytorch_lightning.utilities.rank_zero_only` ([#15536](https://github.com/Lightning-AI/lightning/pull/15536))
```
mauvilsa, line 43: Why is this in the diff?

src/pytorch_lightning/cli.py
```diff
@@ -257,9 +269,6 @@ def __init__(
         trainer_class: Union[Type[Trainer], Callable[..., Trainer]] = Trainer,
         trainer_defaults: Optional[Dict[str, Any]] = None,
         seed_everything_default: Union[bool, int] = True,
-        description: str = "pytorch-lightning trainer command line tool",
-        env_prefix: str = "PL",
-        env_parse: bool = False,
```
victor-yon, line 262: The docstring below (line 283) needs to be updated.

victor-yon (approved):
These changes solve the issue I submitted (#15174)

src/pytorch_lightning/cli.py
mauvilsa: Reverted the removal of _setup_parser_kwargs method and setup_parser parameters.

awaelchli (approved):

PR #15654, opened 2022-11-11
[App] Change app root / config path to be the `app.py` parent directory

ethanwharris:
## What does this PR do?

The logic around finding the root is complex and has some failure modes (e.g. having a .lightning file in a parent directory for any reason will cause that whole directory to get uploaded).

This PR makes it so that the root directory === the parent of the entrypoint file

## Before submitting

## PR review

## Did you have fun?


src/lightning_app/runners/cloud.py
```diff
@@ -215,6 +216,18 @@ def dispatch(
                 )
                 works.append(V1Work(name=work.name, spec=work_spec))
 
+                dict_works.append(
```
tchaton, line 219: This isn't required once my PR is merged.

tchaton (approved):
LGTM ! However, this would break for some apps.

src/lightning_app/runners/cloud.py
ethanwharris: Oh whoops, that's not required at all, accidental change haha. That's from something else 😆

src/lightning_app/CHANGELOG.md
```diff
@@ -37,6 +37,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - The `params` argument in `TracerPythonScript.run` no longer prepends `--` automatically to parameters ([#15518](https://github.com/Lightning-AI/lightning/pull/15518))
 
 
+- Changed the root directory of the app (which gets uploaded) to be the folder containing the app file, rather than any parent folder containing a `.lightning` file ([#15654](https://github.com/Lightning-AI/lightning/pull/15654))
+
```
awaelchli, line 41: This was intentional, see the test cases. The way this PR defines the root is extremely limiting. If the root folder is defined by the app.py file, then your app.py file ALWAYS has to be at the root of your project (the folder requiring all sources to run the program). 

We created the .lightning file for two purposes:
- remember the name of the app
- define the root of the project (the files that need to be uploaded)

See the specification here: https://www.notion.so/lightningai/Lightning-Dotfile-and-the-cloud-experience-cfe28f6936a642b08958cf86b458cafc#9e47b6ae4b6b48a0b9840887994353b3

How will you address the newly introduced limitations?

awaelchli (changes requested):

src/lightning_app/CHANGELOG.md
ethanwharris: I would argue a few points:
- requiring `app.py` to be at the root is no less flexible than requiring the user to create a `.lightning` file at the root
- this behaviour isn't actually documented anywhere and has led to users getting very large directories uploaded on multiple occasions
- `.lightning` files are not currently expected to be commited with source files as they also contain the cluster ID - with that new contract, the majority of apps are required to have `app.py` at the route anyway

Overall I think this approach is much closer to what you expect to happen as a user without having to learn a load about the undocumented internals.

williamFalcon:
yes this is the expected user experience. 
the other experience has already caused a ton of user problems. 

this is especially important for transparency with enterprise customers.

src/lightning_app/CHANGELOG.md
awaelchli: The user does not create it manually. It gets generated for them. If the file is not found, it creates it next to the app.py file (documented in the notion page). This means the default experience is good: Only the files next to the app.py file gets uploaded. 

I am sorry that there is misinformation around this. 

> .lightning files are not currently expected to be commited with source files as they also contain the cluster ID - with that new contract, the majority of apps are required to have app.py at the route anyway

Let's document it then? When we built the initial version of the framework, we had no time for anything.

awaelchli:
> the other experience has already caused a ton of user problems.

What were the problems? Can I take a look at them?

PR #15660, opened 2022-11-12
Simplify version distribution

carmocca:
## What does this PR do?

Further simplification of https://github.com/Lightning-AI/lightning/pull/15374

None


awaelchli (approved):

.actions/setup_tools.py
```diff
@@ -151,12 +151,12 @@ def load_readme_description(path_dir: str, homepage: str, version: str) -> str:
     return text
 
 
-def distribute_version(src_folder: str, ver_file: str = "version.info") -> None:
+def distribute_version(src_folder: str, ver_file: str = "__version__.py") -> None:
     """Copy the global version to all packages."""
-    ls_ver = glob.glob(os.path.join(src_folder, "*", "__version__.py"))
+    src_folders = glob.glob(os.path.join(src_folder, "*/"))
```
Borda, line 156:
```suggestion
    src_folders = [p for p in glob.glob(os.path.join(src_folder, "*")) if os.path.isdir(p)]
```

Borda (changes requested):
this can't be used without installation, you need to keep `__version__.py` or update `__init__.py`
```
>>> import pytorch_lightning
  File "<stdin>", line 1, in <module>
  File "/home/jirka/Workingspace/lightning/src/pytorch_lightning/__init__.py", line 10, in <module>
    from pytorch_lightning.__version__ import version as __version__  # noqa: F401
ModuleNotFoundError: No module named 'pytorch_lightning.__version__'
```

carmocca:
Good point. Then I guess this isn't worth it.

PR #15662, opened 2022-11-12
Deprecate `logger=bool`

### LINKED ISSUE #9900: Make tensorboard into an pip extra
Pytorch lightning provides a pretty convenient abstraction for writing training loops, but the tensorboard dependency ends up dramatically increasing the scope and footprint of the library. As a result, if we import pytorch-lightning into our own packages, it dramatically increases the footprint of those libraries too.

Tensorboard usage feels fairly optional for the pytorch-lightning experience, but it's listed as an unconditional dependency in the requirements.txt file. Would it be possible to move tensorboard into the pip extras instead? Seems like y'all already do this for other visualization and logging tools like `wandb`.

cc @borda @tchaton @awaelchli @edward-io @rohitgr7

carmocca:
## What does this PR do?

Fixes #9900 

None

## Before submitting

## PR review


src/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -22,6 +22,21 @@
...
 from pytorch_lightning.trainer.connectors.logger_connector.result import _METRICS, _OUT_DICT, _PBAR_DICT
 from pytorch_lightning.utilities.metrics import metrics_to_scalars
+from pytorch_lightning.utilities.rank_zero import rank_zero_deprecation
+
+
+# TODO: remove in v2.0.0
+class _NoneSentinel:
+    """Used as a sentinel value for ``None`` in the depreaction for ``Trainer(logger=bool)``.
+
+    Remove this class with the depreaction.
```
JacobHayes, line 32:
```suggestion
    Remove this class with the deprecation.
```

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -112,7 +112,8 @@ class Trainer:
     @_defaults_from_env_vars
     def __init__(
         self,
-        logger: Union[Logger, Iterable[Logger], bool] = True,
+        # TODO: Replace default for `None` in 2.0
+        logger: Optional[Union[Logger, Iterable[Logger]]] = _NoneSentinel(),  # type: ignore[assignment]
```
JacobHayes, line 116: Removing `bool` will "break" mypy users still passing a `bool` even in v1.9 (but maybe that's a nice way to signal the deprecation).

JacobHayes (approved):
Very excited for this, thanks!

src/pytorch_lightning/trainer/trainer.py
carmocca: Exactly. We kind of want to do this as a way to notify the change. Having backwards compatibility for mypy would be too restrictive

src/pytorch_lightning/cli.py
```diff
@@ -550,9 +551,17 @@ def _instantiate_trainer(self, config: Dict[str, Any], callbacks: List[Callback]
...
                 config[key] += value if isinstance(value, list) else [value]
             if self.save_config_callback and not config.get("fast_dev_run", False):
+                to_save = self.config.get(str(self.subcommand), self.config)
+                # TODO: remove this override in v2.0.0
+                if (
+                    "trainer" in to_save
+                    and "logger" in to_save.trainer
+                    and isinstance(to_save.trainer.logger, _NoneSentinel)
+                ):
+                    to_save.trainer.logger = None
```
carmocca, line 561: @mauvilsa To carry out this deprecation, I had to write a None-like sentinel value to know if the user passed something.

With this hack here, I avoid saving the sentinel in configs.

However, I need to do the same for when `--print_config` is called. I haven't easily found a way to do that.

Can you help? Thank you!

tests/tests_pytorch/test_cli.py
```diff
@@ -65,6 +68,22 @@ def mock_subclasses(baseclass, *subclasses):
         yield None
 
 
+@pytest.fixture
```
carmocca, line 71: The CLI tests were relying on a lot of silent behaviour from our loggers. This is why I had to do this many changes to the tests.

tests/tests_pytorch/test_cli.py
```diff
@@ -290,31 +314,31 @@ def test_lightning_cli_save_config_only_once(tmpdir):
...
 
-    with mock.patch("sys.argv", ["any.py", "--config", str(config_path)]):
+    with mock.patch("sys.argv", ["any.py", "--config", config_path]):
         cli = LightningCLI(
             BoringModel,
             BoringDataModule,
             subclass_mode_model=True,
             subclass_mode_data=True,
-            trainer_defaults={"callbacks": LearningRateMonitor()},
+            save_config_kwargs={"overwrite": True},
```
carmocca, line 338: I noticed one small annoyance. If a logger is not set (which would write configs to a new `version_x` directory), then passing a config forces the user to overwrite it even if the contents are the same.

Maybe we should avoid the error if they would be equal. cc @mauvilsa

tests/tests_pytorch/trainer/logging_/test_eval_loop_logging.py
```diff
@@ -765,7 +765,7 @@ def test_dataloader(self):
             return [super().test_dataloader(), super().test_dataloader()]
 
     model = CustomBoringModel()
-    trainer = Trainer(default_root_dir=tmpdir, limit_test_batches=1)
+    trainer = Trainer(default_root_dir=tmpdir, limit_test_batches=1, logger=TensorBoardLogger(tmpdir))
```
carmocca, line 768: @awaelchli Should we also look into making `save_dir` None by default so that it can be replaced with `Trainer.default_root_dir` at setup time? (not in this PR)

src/pytorch_lightning/cli.py
mauvilsa: You just need to register `_NoneSentinel` as a type that serializes to `None`, i.e.,
```python
from jsonargparse.typing import register_type
register_type(_NoneSentinel, serializer=lambda v: None)
```
No need for this custom saving.

tests/tests_pytorch/test_cli.py
mauvilsa: I am not sure what you mean. If no logger is set by the user, then in the saved config the value will be `null`. Why is it that the user is forced to overwrite?

tests/tests_pytorch/trainer/logging_/test_eval_loop_logging.py
awaelchli: Not sure what you mean. It is already like this: If user does not set save dir in the logger, trainer will log to default root dir

requirements/pytorch/test.txt
```diff
@@ -12,5 +12,6 @@ scikit-learn>0.22.1, <1.1.3
 onnxruntime<1.13.0
 psutil<5.9.4 # for `DeviceStatsMonitor`
 pandas>1.0, <1.5.2  # needed in benchmarks
-fastapi<0.87.0
-uvicorn<0.19.1
+tensorboard>=2.9.1, <2.11.0  # for `TensorBoardLogger`
```
Borda, line 15: lets move it to extras

Borda (approved):

PR #15728, opened 2022-11-18
Switch from tensorboard to tensorboardx in logger

lantiga:
## What does this PR do?

The `tensorboard` package used by the TensorBoard logger depends on `grpcio`, which is creating issues on some systems (no binary builds for M1/M2 macOS on Python > 3.9) or with other packages depending on a different `grpcio` version (i.e. `label-studio`).

This PR makes the TensorBoard logger depend on https://github.com/lanpa/tensorboardX instead, which produces equivalent TensorBoard files without extra dependencies apart from `numpy` and `protobuf`.

None

## Before submitting

## PR review

## Did you have fun?

cc @borda @awaelchli


Borda (approved):

ethanwharris (approved):

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -103,7 +103,16 @@ def __init__(
...
         self._sub_dir = None if sub_dir is None else os.fspath(sub_dir)
-        self._log_graph = log_graph
+        self._log_graph = log_graph and _TENSORBOARD_AVAILABLE
+
+        if log_graph and not _TENSORBOARD_AVAILABLE:
+            m = '''You set log_graph=True but Tensorboard is not available. Please run this command to install it:
+
+            pip install tensorboard
+
+            ```
```
lantiga, line 113:
```suggestion
            '''
```

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -103,7 +103,16 @@ def __init__(
         self._name = name or ""
         self._version = version
         self._sub_dir = None if sub_dir is None else os.fspath(sub_dir)
-        self._log_graph = log_graph
+        self._log_graph = log_graph and _TENSORBOARD_AVAILABLE
+
+        if log_graph and not _TENSORBOARD_AVAILABLE:
+            m = """You set log_graph=True but Tensorboard is not available. Please run this command to install it:
+
+            pip install tensorboard
```
carmocca, line 111: This will not be printed correctly. All the whitespace to the left will appear. I suggest using a regular string.

requirements/pytorch/test.txt
```diff
@@ -14,3 +14,4 @@ psutil<5.9.4 # for `DeviceStatsMonitor`
 pandas>1.0, <1.5.2  # needed in benchmarks
 fastapi<0.87.0
 uvicorn<0.19.1
+tensorboard>=2.9.1, <2.11.0
```
carmocca, line 17: Do we know what happens when both are installed? `tensorboardX` is still used because of the `tensorboardX` import, no?

I think we should do this

```python
if _TENSORBOARD_AVAILABLE:
    from tensorboard ...
elif _TENSORBOARDX_AVAILABLE:
    from tensorboardX ...
```

carmocca (commented):
Happy to make these suggested changes for you if you prefer! Waiting on your greenlight

lantiga:
> Happy to make these suggested changes for you if you prefer! Waiting on your greenlight

thanks @carmocca, doing now

requirements/pytorch/test.txt
lantiga: ok let's do this

Borda: I do not think so, as `tensorboardx` is importing `tensorboard` in if you request logging graph

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -23,19 +23,24 @@
...
 from lightning_lite.utilities.types import _PATH
 from pytorch_lightning.core.saving import save_hparams_to_yaml
 from pytorch_lightning.loggers.logger import Logger, rank_zero_experiment
-from pytorch_lightning.utilities.imports import _OMEGACONF_AVAILABLE
+from pytorch_lightning.utilities.imports import _OMEGACONF_AVAILABLE, _TENSORBOARD_AVAILABLE
 from pytorch_lightning.utilities.logger import _add_prefix, _convert_params, _flatten_dict
 from pytorch_lightning.utilities.logger import _sanitize_params as _utils_sanitize_params
 from pytorch_lightning.utilities.rank_zero import rank_zero_only, rank_zero_warn
 
+if _TENSORBOARD_AVAILABLE:
```
Borda, line 37: I do not think this is really needed as https://github.com/Lightning-AI/lightning/pull/15728/files#r1026799585

requirements/pytorch/test.txt
lantiga: Right, but if `tensorboard` is present then we can use it fully (not just for graph).

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -103,7 +108,14 @@ def __init__(
         self._name = name or ""
         self._version = version
         self._sub_dir = None if sub_dir is None else os.fspath(sub_dir)
-        self._log_graph = log_graph
+        self._log_graph = log_graph and _TENSORBOARD_AVAILABLE
+
+        if log_graph and not _TENSORBOARD_AVAILABLE:
+            rank_zero_warn(
+                "You set log_graph=True but TensorBoard is not available. "
+                "Please run this command to install it: pip install tensorboard."
```
Borda, line 116:
```suggestion
                "You set `log_graph=True`, but TensorBoard is not available."
                " Please install missing dependency `pip install tensorboard`"
```

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -103,7 +108,14 @@ def __init__(
         self._name = name or ""
         self._version = version
         self._sub_dir = None if sub_dir is None else os.fspath(sub_dir)
-        self._log_graph = log_graph
+        self._log_graph = log_graph and _TENSORBOARD_AVAILABLE
+
+        if log_graph and not _TENSORBOARD_AVAILABLE:
+            rank_zero_warn(
+                "You set `log_graph=True`, but TensorBoard is not available."
+                " Please install missing dependency `pip install tensorboard`"
```
lantiga, line 116:
```suggestion
                " Please install TensorBoard by running `pip install tensorboard`"
```

justusschock (approved):
lgtm in general, it would just be nice to have tests without having full tensorboard installed (which us not tested now since we always install all testing dependencies)

requirements/pytorch/test.txt
Borda: What fully you mean? Logger here is just interface to dump your data, for visualization you can use still both but independent on PL...

Borda:
> lgtm in general, it would just be nice to have tests without having full tensorboard installed (which us not tested now since we always install all testing dependencies)

All doctests are called without installing all dependencies

justusschock:
that's true but doctests for loggers are meaningless as they only test imports here.

Borda:
> that's true but doctests for loggers are meaningless as they only test imports here.

So let's add one or we may have elsewhere simple train fit in docstring which would be using default logger...

lantiga:
@Borda what is the best strategy to not have `tensorboard` installed for a test?

Borda:
> @Borda what is the best strategy to not have `tensorboard` installed for a test?

Probably, run all tests before install development requirements, but it will extend running time (note that with this case not all tests will run as some tests are patched by skip if missing)

awaelchli:
@lantiga by removing it from requirements/test.txt. IMO, if tensorboardx is going to be our new default, it should be what is running in our test suite.

requirements/pytorch/test.txt
awaelchli, line 17:
```suggestion
```
IMO, if tensorboardx becomes the new default, we should run it in our CI.

justusschock:
I think it is fine to have it installed but just mock the installation check so that we can run the tests with both options

Borda:
> @lantiga by removing it from requirements/test.txt. IMO, if tensorboardx is going to be our new default, it should be what is running in our test suite.

Yes, but then we canot test `log_graph` so we just need to remove the switch between using TensorBoard and TensorBoard X

requirements/pytorch/test.txt
Borda: Then we need to drop testing `log_graph`

carmocca: We would need to parametrize all TB tests so that they run once with each dependency. I suggest leaving this for follow-up work

carmocca (approved):
I pushed some improvements to lazily import tensorboard. Also fixed tests when only `tensorboardX` is installed.

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -97,13 +107,28 @@ def __init__(
         sub_dir: Optional[_PATH] = None,
         **kwargs: Any,
     ):
+        if not _TENSORBOARD_AVAILABLE and not _TENSORBOARDX_AVAILABLE:
```
Borda, line 110:
```suggestion
        if not _TENSORBOARD_AVAILABLE:
```
this is required depenency

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -97,13 +107,28 @@ def __init__(
...
+            if _TENSORBOARDX_AVAILABLE:
+                rank_zero_warn(
+                    "You set `TensorBoardLogger(log_graph=True)` but `tensorboard` is not available. `tensorboardX` is"
+                    " installed but it does not support this feature."
+                )
+            else:
+                rank_zero_warn(
+                    "You set `TensorBoardLogger(log_graph=True)` but `tensorboard` is not available. "
+                    + str(_TENSORBOARDX_AVAILABLE)
+                )
```
Borda, line 130:
```suggestion
            rank_zero_warn(
                "You set `TensorBoardLogger(log_graph=True)` but `tensorboard` is not available.")
```

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -36,6 +35,17 @@
...
+_TENSORBOARD_AVAILABLE = RequirementCache("tensorboard>=2.9.1")
+_TENSORBOARDX_AVAILABLE = RequirementCache("tensorboardX>=2.0")
+if TYPE_CHECKING:
+    # assumes at least one will be installed when type checking
+    if _TENSORBOARD_AVAILABLE:
+        from torch.utils.tensorboard import SummaryWriter
+    else:
+        from tensorboardX import SummaryWriter  # type: ignore[no-redef]
+else:
+    SummaryWriter = Any
```
Borda, line 47: how this can happen if TBX is a mandatory dependency?

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -97,13 +107,20 @@ def __init__(
         sub_dir: Optional[_PATH] = None,
         **kwargs: Any,
     ):
+        if not _TENSORBOARD_AVAILABLE:
+            raise ModuleNotFoundError(f"{_TENSORBOARD_AVAILABLE!s}. You can also install `tensorboardX` if you prefer.")
+
```
Borda, line 112:
```suggestion
```
this makes TB mandatory not TBX

Borda (changes requested):

carmocca (commented):
@Borda Regardless of our default installation, users might choose one over the other, so we need to check for both. Please revert the last 2 commits you pushed.

src/pytorch_lightning/loggers/tensorboard.py
Borda: but why is TBX is required rependency

lantiga:
Hey, fundamentally I think we need to avoid the duality (I agree with @Borda here).

If we trust `tensorboardX` (as we do, since we are making it the default), then we need to trust that it will be fully equivalent to `tensorboard` for anything that is not `log_graph`. And in this latter case, we correctly emit the error and `tensorboardX` will relay the call to `tensorboard`.

So, I propose we remove the dual behavior and just simplify the logic. Also I'd like to merge this before the release tomorrow.

src/pytorch_lightning/loggers/tensorboard.py
carmocca: `TensorBoardLogger` can be used with `tensorboard` or `tensorboardX` after this PR. Hence you need to check that at least one of them is installed.

If we support `tensorboard`, it should also work if the user ran `pip install lightning; pip uninstall tensorboardX; pip install tensorboard`

src/pytorch_lightning/loggers/tensorboard.py
carmocca: Importing has an overhead. You don't want to import the library unless you are going to use it. We use the `TYPE_CHECKING` flag here because forward references need to have a reference to the class.

carmocca:
@lantiga My changes do not modify the functional intention of this PR. They are just meant to decrease the overhead of import `pytorch_lightning` (see https://github.com/Lightning-AI/lightning/issues/12786) while still fixing type annotations and tests. None of this is relevant to the end user so I don't see why it needs to be "simplified". Also note that my changes were passing on CI

lantiga:
> @lantiga My changes do not modify the functional intention of this PR. They are just meant to decrease the overhead of import `pytorch_lightning` (see #12786) while still fixing type annotations and tests. None of this is relevant to the end user so I don't see why it needs to be "simplified". Also note that my changes were passing on CI

I see @carmocca, do you think import overhead is impacted with the latest approach?

carmocca:
Yes. In any case, as I said my changes are functionally equivalent so this can be merged and I'll propose them again in a separate PR

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -77,11 +81,20 @@ class TensorBoardLogger(Logger):
         sub_dir: Sub-directory to group TensorBoard logs. If a sub_dir argument is passed
             then logs are saved in ``/save_dir/name/version/sub_dir/``. Defaults to ``None`` in which
             logs are saved in ``/save_dir/name/version/``.
-        \**kwargs: Additional arguments used by :class:`SummaryWriter` can be passed as keyword
+        \**kwargs: Additional arguments used by :class:`tensorboardX.SummaryWriter` can be passed as keyword
             arguments in this logger. To automatically flush to disk, `max_queue` sets the size
             of the queue for pending logs before flushing. `flush_secs` determines how many seconds
             elapses before flushing.
 
+    Example:
```
Borda, line 89: adding example to be tested with all installation with TBX and without TB

Borda (approved):

Borda:
> They are just meant to decrease the overhead of import `pytorch_lightning` (see #12786) while still fixing type annotations and tests.

well, you shortened PL import time but then you do an import of TB/TBX is called in the method `experiment` and `log_hyperparams`

> Also note that my changes were passing on CI

no one is complaining about CI :)

leoleoasd:
This PR breaks remote file system support as it seems tensorboardX doesn't support fsspec.
image

carmocca:
@leoleoasd If you install the 1.9 release candidate (or master), you can `pip install tensorboard` and then we will use it over `tensorboardX`. Support for that was added in https://github.com/Lightning-AI/lightning/pull/15762 but it was not cherry-picked into a 1.8.x release.

leoleoasd:
Ok, thanks @carmocca!
However:
1. should we clearly document this in the `remote filesystem` section of the document?
2. it was said that this is currently a temporary solution (https://github.com/Lightning-AI/lightning/pull/15789#discussion_r1030835829). Will we still have fsspec support for the "permanent" version?

carmocca:
1. Yes, this would be an useful note. Would you like to open a PR with the docs change?
2. Remote support is important for us, so we will look to support fsspec as much as we can.

PR #15749, opened 2022-11-21
Run mypy with torch==1.13.0

carmocca:
## What does this PR do?

See title

None

### DIFF
--- a/.github/workflows/code-checks.yml
+++ b/.github/workflows/code-checks.yml
@@ -27,6 +27,8 @@ jobs:
 
   mypy:
     runs-on: ubuntu-20.04
+    env:
+      PYTORCH_VERSION: '1.3.0'
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/code-checks.yml
```diff
@@ -45,10 +47,10 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install mypy==0.982  # TODO: pull this verion deom requirements/test.txt
-        pip install torch==1.12 --find-links https://download.pytorch.org/whl/cpu/torch_stable.html
-        python ./requirements/pytorch/adjust-versions.py requirements/pytorch/extra.txt
-        # todo: adjust requirements for both code-bases
+        pip install mypy==0.982
```
Borda, line 50: shall this be part of the requirements so it is managed by dependabot?

carmocca: I suggest not. It requires manual fixes for basically all releases

Borda: yes, but this is still under PR so no merge until it is resolved

> for basically all releases

not sure what releases are you talking about... :otter:

carmocca: mypy releases

Borda: we can skip, but at least we wont be frozen for ever :)

PR #15762, opened 2022-11-21
Lazy import tensorboard

carmocca:
## What does this PR do?

Avoids importing tensorboard unless the logger is used.
Part of #12786
Part of #15662 

### Does your PR introduce any breaking changes? If yes, please list them.

None

cc @awaelchli @borda @akihironitta

Issue #14695, opened 2022-09-14
Unable to use neptune logger after a version upgrade of pytorch lightning


joshua840:
## 🐛 Bug

Unable to initialize Neptune logger after version upgrade of pytorch lightning from 1.6.5 to 1.7.0.

```
 File "/PATH/test2.py", line 14, in <module>
    cli_main()
  File "/PATH/test2.py", line 11, in cli_main
    logger = NeptuneLogger(run=run, log_model_checkpoints=False)
  File "/MYENV/envs/torch_1.12/lib/python3.10/site-packages/pytorch_lightning/loggers/neptune.py", line 278, in __init__
    self._verify_input_arguments(api_key, project, name, run, neptune_run_kwargs)
  File "/MYENV/envs/torch_1.12/lib/python3.10/site-packages/pytorch_lightning/loggers/neptune.py", line 377, in _verify_input_arguments
    if run is not None and not isinstance(run, Run):
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

In my case, the value of Run is None, which is determined in the following code.

https://github.com/Lightning-AI/lightning/blob/14b36f8109db63f4c9ed2480734ce5a0dbbc71b9/src/pytorch_lightning/loggers/neptune.py#L43

If I downgrade lightning to 1.6.5, the error does not occur.

### To Reproduce

Code
```
import pytorch_lightning as pl
from pytorch_lightning.loggers.neptune import NeptuneLogger
import neptune.new as neptune

# project_name should be "WORKSPACE_NAME/PROJECT_NAME"
PROJECT_NAME = 'YOUR_NEPTUNE_PROJECT'
API_KEY = 'YOUR_NEPTUNE_API_KEY'

def cli_main():
    run = neptune.init(api_token=API_KEY, project=PROJECT_NAME, capture_stdout=False)
    logger = NeptuneLogger(run=run, log_model_checkpoints=False)

if __name__ == "__main__":
    cli_main()
```

### Expected behavior

Test code should works regardless of the PyTorch lightning version.

### Environment

Please save the following environment information and create env by this command.
```
conda env create --file ENV.yml
```

Conda environment
```
name: torch_1.12
channels:
  - pytorch
  - conda-forge
  - defaults
dependencies:
  - _libgcc_mutex=0.1=main
  - _openmp_mutex=5.1=1_gnu
  - absl-py=1.2.0=pyhd8ed1ab_0
[...]

PR #15789, opened 2022-11-23
loggers: revert useless exception

Borda:
## What does this PR do?

revert ignored review from #15762

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -114,10 +113,6 @@ def __init__(
         sub_dir: Optional[_PATH] = None,
         **kwargs: Any,
     ):
-        if not _TENSORBOARD_AVAILABLE and not _TENSORBOARDX_AVAILABLE:
```
carmocca, line 117: This is here to provide a good error message for anybody using this class who's tb/tbX dependency is not working. Otherwise they will only find out when the code actually runs and the local imports are run. We want to fail with a good message as soon as possible

Borda: can you pls give me a simple example of how this could happen, tbX dependency is not working?
Also why we do not have the same for other mandatory packages like tqdm, fsspec, torchmetrics?

carmocca: > Also why we do not have the same for other mandatory packages like tqdm, fsspec, torchmetrics?

We want to do the same eventually. It's just work in progress tracked in #12786

> how this could happen, tbX dependency is not working?

tbX could not be available for other reasons other than that the dependency is missing. For example, it would show as unavailable if it requires a different dependency that wasn't properly installed. We had an instance of this happen to the NeptuneLogger previously: https://github.com/Lightning-AI/lightning/issues/14695#issuecomment-1247899659

carmocca: It could because a package can appear as installed but still fail the `pkg_resources.require("the_requirement")` test. One example is in the error reported in #14695.

The intention here is to be extra safe and provide a good error message for the user if this happens.

Borda: but you are referring to extra not mandatory dependency, in such case you shall path torch :otter:

carmocca: tb is functionally an extra dependency, one can use PL without tensorboard installed. Given the problematic nature of it's installation, I can see users choosing to drop it. Either after a pip installation or by modifying the local requirements in a git clone. In that case, this error helps

awaelchli: Personally I see this as a temporary solution. TBX still has dependencies we need to get rid of so it has to go eventually.

awaelchli (approved):
Either way is fine with me. This is in a temporary state anyway.

akihironitta (approved):

PR #15807, opened 2022-11-24
[App] Add Signal Handler Compose

tchaton:
## What does this PR do?

This PR enables to compose signals within the work proxies

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?


lantiga (approved):
Very cool!

lantiga:
There's apparently some interaction with spawn examples

src/lightning_app/utilities/proxies.py
```diff
@@ -420,6 +421,9 @@ def __call__(self):
     def setup(self):
         from lightning_app.utilities.state import AppState
 
+        # Note: Enable to capture all signals and call them sequentially in the order they got registered
+        signal.signal = SignalHandlerCompose(signal.signal)
```
awaelchli, line 425: What if signal.signal is already a SignalHandlerCompose? You will get `SignalHandlerCompose(SignalHandlerCompose(SignalHandlerCompose(SignalHandlerCompose(....`
since this is a global setting and so you will stack it across test runs, right?

tests/tests_app/utilities/test_proxies.py
```diff
@@ -790,3 +792,27 @@ def test_bi_directional_proxy_filtering():
...
+    def handler_fn2(signum, handler):
+        nonlocal examples
+        examples.append("handler_fn2")
+
+    signal.signal(signal.SIGTERM, handler_fn)
+    signal.signal(signal.SIGTERM, handler_fn2)
+
+    handler.run(signal.SIGTERM, None)
+
+    assert examples == ["handler_fn", "handler_fn2"]
```
awaelchli, line 818: We can see what's wrong here: The signal handlers are a global setting. The expectation is that if you call

signal.signal(x, handler) 

then it should replace the handler that was already there. In this PR, Lightning breaks this contract in all other libraries, since it patches signal.signal globally. 

I believe the intended use should be:

```
signal.signal(signum, HandlerCompose(handler1, handler2, ...))
```

awaelchli (changes requested):

src/lightning_app/utilities/signals.py
```diff
@@ -0,0 +1,25 @@
...
+        for callable in self.signums[signum]:
+            callable(signum, frame)
+
+    def __call__(self, signum, callable):
+        if signum not in self.signums:
+            self.signums[signum] = []
+
+        self.signums[signum].append(callable)
+
+        self.original_signal(signum, self.run)
```
hhsecond, line 25: we shouldn't do this for each signal registration of same `signum`

hhsecond (commented):
Looks nice! I had taken a similar route in the past and found out that the way uvloop (hence uvcorn and hence fastapi) interact with signal handling makes some conflicts with this approach.

- Async mechanism has a different way of setting handlers

    ```python
        loop = asyncio.get_running_loop()
        loop.add_signal_handler(...)
    ```

- uvloop somehow takes precedence even if we register the signal handler at the init. I haven't got around investigating what's happening there though

that being said, this is definitely the first step to making the signal handling sane

src/lightning_app/utilities/signals.py
hhsecond: Also, if there are signals registered already, we shouldn't be overwriting those. Instead we could get it using `getsignal(signum)` and keep it at the top of the stack

tests/tests_app/utilities/test_proxies.py
hhsecond: So the problem we are trying to solve here is that uvloop overwrites the signal handler we set. So if we don't patch `signal.signal`, as soon as we run `uvcorn.run`, it overwrites the handler and hence `ctrl-c` would never terminate the work process correctly

awaelchli: Then I suggest we only patch it for uvicorn/ in uvicorn

PR #15819, opened 2022-11-25
Selects random port for lightning app when running multiple apps locally

nohalon:
## What does this PR do?
Selects a random port when running a lightning app to avoid port collisions when running multiple apps locally.

No breaking changes. Most of the tests expect port 7501 as the hardcoded port for locally running lightning apps. This PR still uses 7501 as the default port for lightning apps and will only use a different port when this port is taken.

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/utilities/port.py
```diff
@@ -14,7 +26,11 @@ def _find_lit_app_port(default_port: int) -> int:
     enable_multiple_works_in_default_container = bool(int(os.getenv("ENABLE_MULTIPLE_WORKS_IN_DEFAULT_CONTAINER", "0")))
 
     if not app_id or not project_id or not enable_multiple_works_in_default_container:
-        return default_port
+        # If the default port is not available, picks any other available one
+        if not is_port_in_use(default_port):
+            return default_port
+        else:
+            return get_available_port()
```
awaelchli, line 33: @nohalon We could consider using the existing utility function `lightning_app.utilities.network.find_free_network_port`. This would be slightly faster compared to the current approach which opens a connection twice (once for checking port in use, once for getting a new one).

nohalon: perfect will use that! thanks @awaelchli

akihironitta (approved):

src/lightning_app/utilities/port.py
```diff
@@ -1,20 +1,32 @@
...
     project_id = os.getenv("LIGHTNING_CLOUD_PROJECT_ID", None)
     enable_multiple_works_in_default_container = bool(int(os.getenv("ENABLE_MULTIPLE_WORKS_IN_DEFAULT_CONTAINER", "0")))
 
     if not app_id or not project_id or not enable_multiple_works_in_default_container:
-        return default_port
+        app_port = default_port
+
+        # If the default port is not available, picks any other available one
+        if is_port_in_use(default_port):
+            app_port = find_free_network_port()
```
awaelchli, line 27: I just realized that we are calling this at import time. The function `_find_lit_app_port` here gets called in `constants.py` and assigned to the variable `APP_SERVER_PORT = _find_lit_app_port(7501)`. 

- This could slow down import time of lightning
- we still can get colliding apps if not checked at runtime

awaelchli (approved):

tchaton (approved):
LGTM !

src/lightning_app/utilities/port.py
nohalon: so @awaelchli should I revert back to my previous implementation which found a free port without `_find_lit_app_port`

akihironitta (approved):
LGTM!

src/lightning_app/cli/lightning_cli.py
```diff
@@ -267,6 +269,9 @@ def _run_app(
 
     secrets = _format_input_env_variables(secret)
 
+    port = _find_lit_app_port(constants.APP_SERVER_PORT)
+    constants.APP_SERVER_PORT = port
```
awaelchli, line 273: If there is really no other way than making it a global variable, then let's move it out of the "constants" file? Clearly, if it changes at runtime, it's no longer a constant.

Perhaps `constants.py` could contain `APP_SERVER_DEFAULT_PORT=7501` and then, I don't know, maybe `lightning_app.py` could hold the global var for the port that's being used.

nohalon: Thanks for the review @awaelchli I will update per your suggestion

PR #15853, opened 2022-11-28
Fix pip install with no PACKAGE_NAME or editable mode

carmocca:
## What does this PR do?

Fixes issue reported by @hhsecond on Slack. Repro:

```bash
pip uninstall lightning_lite pytorch_lightning lightning lightning_app -y
git clean -fxd .
pip install .
lightning
```

Missed in https://github.com/Lightning-AI/lightning/pull/15646

Current workaround is to do

```python
PACKAGE_NAME=lightning pip install .
```

None


Borda (approved):

hhsecond (approved):

setup.py
```diff
@@ -90,7 +90,15 @@ def _set_manifest_path(manifest_dir: str, aggregate: bool = False) -> Generator:
...
         for pkg in mapping.values():
-            with open(os.path.join(_PATH_SRC, pkg, "MANIFEST.in")) as fh:
+            pkg_path = os.path.join(_PATH_SRC, pkg)
+            if not os.path.exists(pkg_path):
+                # this function is getting called with `pip install .` and `pip install *.tar.gz`, however, it only
+                # should be called with the former. i haven't found a way to differentiate the two so this is the hacky
+                # solution to avoid an error
+                print(f"{pkg_path!r} does not exist")
+                yield
+                return
```
Borda, line 100: why do we exit here?

Borda (commented):
lets add test for this case

.github/workflows/ci-pkg-install.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04, macOS-12, windows-2022]
-        pkg-name: ["app", "lite", "pytorch", "lightning"]
+        pkg-name: ["app", "lite", "pytorch", "lightning", ""]
```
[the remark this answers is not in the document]
carmocca: This change is not correct. The variable needs to be unset, instead of set to empty string

setup.py
carmocca: See the comment. We want to skip the rest of the logic

akihironitta (approved):

.github/workflows/ci-pkg-install.yml
```diff
@@ -61,8 +61,7 @@ jobs:
           pkg-name: ${{ matrix.pkg-name }}
 
       - name: Run CLI
-        # todo: add testing for `lightning_app`
-        if: ${{ matrix.pkg-name == 'lightning' }}
+        if: ${{ matrix.pkg-name == 'lightning' || matrix.pkg-name == "app" }}
```
carmocca, line 64:
```suggestion
        if: ${{ matrix.pkg-name == 'lightning' || matrix.pkg-name == 'app' }}
```

PR #15875, opened 2022-11-30
[App] Support for headless apps

ethanwharris:
## What does this PR do?

- Indicate to the cloud when an app is headless
- Avoid opening the URL locally when an app is headless

### Known limitations

**Single work apps technically have UIs so aren't treated as headless by this PR - to be handled in follow-up work**

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/utilities/app_helpers.py
```diff
@@ -527,3 +528,15 @@ def _should_dispatch_app() -> bool:
...
         and "LIGHTNING_APP_STATE_URL" not in os.environ
     )
+
+
+def _is_headless(app: "LightningApp") -> bool:
+    """Utility which returns True if the given App has no ``Frontend`` objects or URLs exposed through
+    ``configure_layout``."""
+    if app.frontends:
+        return False
+    for component in breadth_first(app.root, types=(lightning_app.LightningFlow,)):
```
tchaton, line 538: Unfortunately, an app can be `headless` for some time.
This could reduce the current flexibility of the UI aspects of the framework.
I would prefer to have a property of the root flow to express this instead.

src/lightning_app/utilities/layout.py
```diff
@@ -108,8 +109,9 @@ def _collect_content_layout(layout: List[Dict], flow: "lightning_app.LightningFl
                 entry["content"] = ""
                 entry["target"] = ""
         elif isinstance(entry["content"], _MagicMockJsonSerializable):
-            # Do nothing
-            pass
+            # The import was mocked, so we just record dummy content
```
tchaton, line 112: Mind adding a comment why this is required for future readers.

src/lightning_app/utilities/app_helpers.py
ethanwharris: I don't think we want to introduce a property here, but it's entirely possible for the `is_headless` flag that we set in the lightning app instance to be mutable. So we would be able to add back support for apps which start off headless and later dynamically start pointing to a URL without needing a flag

ethanwharris: Spoke with @tchaton - we'll go for updating the `LightningAppInstance` if / when the headless state changes - that will preserve support for dynamic layouts

awaelchli (approved):

tests/tests_app/runners/test_singleprocess.py
```diff
@@ -1,3 +1,8 @@
+import os
```
hhsecond, line 1: do we even need to keep the single process runtime anymore?

hhsecond (approved):
LGTM

tests/tests_app/runners/test_singleprocess.py
ethanwharris: No, I think it could be removed

src/lightning_app/utilities/app_helpers.py
```diff
@@ -527,3 +529,46 @@ def _should_dispatch_app() -> bool:
...
+                return False
+    return True
+
+
+def _handle_is_headless(app: "LightningApp"):
+    """Utility for runtime-specific handling of changes to the ``is_headless`` property."""
+    app_id = os.getenv("LIGHTNING_CLOUD_APP_ID", None)
+    project_id = os.getenv("LIGHTNING_CLOUD_PROJECT_ID", None)
+
+    if app_id and project_id:
```
tchaton, line 551: Cleaner to return if not defined.

ethanwharris: Done!

PR #15911, opened 2022-12-05
[App] Fix bug when using structures with works

ethanwharris:
## What does this PR do?

There is a bug where works in structures can cause failures and an app retry loop when running in the cloud.

E.g. this app fails with the latest release:
```python
# app.py
# !curl https://raw.githubusercontent.com/Lightning-AI/lightning/master/examples/app_multi_node/pl_boring_script.py -o pl_boring_script.py
import lightning as L
from lightning.app.components.training import LightningTrainerScript

# run script that trains PyTorch with the Lightning Trainer
model_script = 'pl_boring_script.py'
component = LightningTrainerScript(
   model_script,
   num_nodes=1,
   cloud_compute=L.CloudCompute("gpu")
)
app = L.LightningApp(component)
```

The reason for the failure is that the `LightningTrainerScript` component populates the structure on init, before a backend is attached. As a result, the works get added without being connected to the flow, this triggers a validation error when `work.run` is called.

## Before submitting

## PR review

## Did you have fun?


awaelchli:
@ethanwharris For my understanding, how can we produce this bug? If you can show us an example, we could turn it into a test case.

src/lightning_app/structures/dict.py
```diff
@@ -64,10 +64,10 @@ def __setitem__(self, k, v):
         if isinstance(k, str) and "." in k:
             raise Exception(f"The provided name {k} contains . which is forbidden.")
 
+        _set_child_name(self, v, k)
```
tchaton, line 67: Can u share the failure trace + add a test for this example? If I understand properly, the issue was the works weren't named properly right?

ethanwharris: I can add a test, but this is broken on the latest release, so it would be best to get the fix in the release then follow-up with a test.

ethanwharris:
@awaelchli Updated the description with the failure case and some more info. We can definitely turn this into a test, but maybe best for a follow-up so we can get this in quickly. Let me know what you think :smiley:

awaelchli:
Here is a test that produces a failure on master and passes on your PR. It demonstrates the intended behavior. 

```py

def test_structures_have_name_on_init():
    """Test that the children in structures have the correct name assigned upon initialization."""
    class ChildWork(LightningWork):
        def run(self):
            pass

    class Collection(EmptyFlow):
        def __init__(self):
            super().__init__()
            self.list_structure = List()
            self.list_structure.append(ChildWork())

            self.dict_structure = Dict()
            self.dict_structure["dict_child"] = ChildWork()

    flow = Collection()
    LightningApp(flow)  # wrap in app to init all component names
    assert flow.list_structure[0].name == "root.list_structure.0"
    assert flow.dict_structure["dict_child"].name == "root.dict_structure.dict_child"
```

awaelchli (approved):

tchaton (approved):
LGTM !

Borda:
@ethanwharris why all these requirement changes?

requirements/lite/examples.txt
```diff
@@ -1,4 +1,4 @@
 # NOTE: the upper bound for the package version is only set for CI stability, and it is dropped while installing this package
 #  in case you want to preserve/enforce restrictions on the latest compatible version, add "strict" as an in-line comment
 
-torchvision>=0.10.*, <=0.13.0
+torchvision>=0.10.0, <=0.13.0
```
Borda, line 4: this shall be TV 0.11 as to be paired with PT 1.10, right?

requirements/pytorch/examples.txt
```diff
@@ -1,6 +1,6 @@
 # NOTE: the upper bound for the package version is only set for CI stability, and it is dropped while installing this package
 #  in case you want to preserve/enforce restrictions on the latest compatible version, add "strict" as an in-line comment
 
-torchvision>=0.11.*, <=0.14.0
+torchvision>=0.11.1, <=0.14.0
```
Borda, line 4: is TV 0.11.1 compatible with PT 1.10.0?

src/lightning_app/structures/list.py
```diff
@@ -52,23 +50,21 @@ def __init__(self, *items: T):
 
         self._name: t.Optional[str] = ""
         self._last_index = 0
-        self._backend: Optional[Backend] = None
+        self._backend: t.Optional[Backend] = None
```
Borda, line 53: why import with `t.` and not `from typing import ...`?

requirements/pytorch/examples.txt
ethanwharris: Yes, 0.11.0 was yanked

src/lightning_app/structures/list.py
ethanwharris: Because we already have `import typing as t` in there

requirements/pytorch/examples.txt
Borda: :+1:

src/lightning_app/structures/list.py
Borda: no I see the weird typing hack :D

PR #15921, opened 2022-12-06
[App] Add ready property to the flow

tchaton:
## What does this PR do?

This PR introduces the `ready` property on the flow to enable Open UI only once certain conditions are met.

```python
class FlowReady(LightningFlow):
    def __init__(self):
        super().__init__()
        self.w = WorkReady()

    @property
    def ready(self) -> bool:
        return self.w.has_succeeded

    def run(self):
        self.w.run()

        if self.ready:
            self._exit()
```

Example with the Boring App:

```python
    @property
    def ready(self) -> bool:
        if "dst_w" in self.dict:
            return self.dict["dst_w"].url != ""
        return False
```

https://user-images.githubusercontent.com/12861981/205928911-aadc0f12-4de8-4dba-b0cb-1c8e9a742243.mov

Fixes #\<issue_number>

## Before submitting

## PR review

## Did you have fun?

cc @borda

### DIFF
--- a/src/lightning_app/core/app.py
+++ b/src/lightning_app/core/app.py
@@ -141,6 +141,8 @@ class LightningApp:
         self.threads: List[threading.Thread] = []
         self.exception = None
         self.collect_changes: bool = True
+
+        # TODO: Enable ready locally for opening the UI.
         self.ready = False
 
         # NOTE: Checkpointing is disabled by default for the time being.  We
--- a/src/lightning_app/core/app.py
+++ b/src/lightning_app/core/app.py
@@ -489,7 +489,6 @@ class LightningApp:
         self._start_with_flow_works()
 
         if self.ready and self.should_publish_changes_to_api and self.api_publish_state_queue:
-            logger.debug("Publishing the state with changes")
             self.api_publish_state_queue.put(self.state_vars)
 
         self._reset_run_time_monitor()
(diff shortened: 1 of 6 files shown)


ethanwharris (approved):
Nice! LGTM 😃

src/lightning_app/core/flow.py
```diff
@@ -230,6 +230,12 @@ def __getattr__(self, item):
             return Path.from_dict(self._paths[item])
         return self.__getattribute__(item)
 
+    @property
+    def ready(self) -> bool:
+        """Override to customize when your App should be ready."""
```
awaelchli, line 235: The user doesn't know what it means for it to be ready. The impact of this is that the button on the admin UI becomes enabled and the user can open the UI when the app is ready. I think we should explicitly say that here. 

Also, as discussed offline, the behavior here is special. The ready property will only be re-evaluated until it turns to True and then never again. Meaning the state here can't change arbitrarily during runtime. This probably makes sense, so I suggest to document the behavior.

awaelchli (approved):

src/lightning_app/core/app.py
```diff
@@ -142,6 +142,9 @@ def __init__(
         self.exception = None
         self.collect_changes: bool = True
 
+        # TODO: Enable ready locally for opening the UI.
+        self.ready = False
```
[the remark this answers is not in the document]
tchaton: I would prefer to do this another PR to deliver faster.

src/lightning_app/core/flow.py
tchaton: This would be clarified once local dev is added properly.

PR #15922, opened 2022-12-06
Direct support for compiled models

lantiga:
## What does this PR do?

This PR introduces dynamo/inductor support to Lightning. A prior design proposed the use of a callback.

However, since compilation is going to be first-class in PyTorch, through `torch.compile`, here we add direct support for models compiled through `torch.compile`, so there's zero cognitive overhead for people who get into compiling PyTorch models. This is what it looks like, from regular modules:

```python
model = BoringModel()

trainer = Trainer()
trainer.fit(model)
```

to compiled ones:

```python
model = BoringModel()
model = torch.compile(model)

trainer = Trainer()
trainer.fit(model)
```

None

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/core/module.py
+++ b/src/pytorch_lightning/core/module.py
@@ -117,6 +117,7 @@ class LightningModule(
         self._metric_attributes: Optional[Dict[int, str]] = None
         self._should_prevent_trainer_and_dataloaders_deepcopy: bool = False
         self._register_sharded_tensor_state_dict_hooks_if_available()
+        self._compiler_ctx: Optional[Dict[str, Any]] = None
 
     @overload
     def optimizers(self, use_pl_optimizer: Literal[True] = True) -> Union[LightningOptimizer, List[LightningOptimizer]]:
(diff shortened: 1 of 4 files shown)


src/pytorch_lightning/core/module.py
```diff
@@ -1950,6 +1951,70 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
...
+            "compiler": "dynamo",
+            "dynamo_ctx": model.dynamo_ctx,
+            "original_forward": orig_module.forward,
+        }
+
+        orig_module.forward = model.dynamo_ctx(orig_module.forward)  # type: ignore[assignment]
+        return orig_module
+
+    @classmethod
+    def to_uncompiled(cls, model: Union["pl.LightningModule", "torch._dynamo.OptimizedModule"]) -> "pl.LightningModule":
```
tchaton, line 1989: This isn't used yet right ?

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -963,6 +988,18 @@ def _restore_modules_and_callbacks(self, checkpoint_path: Optional[_PATH] = None
     def _run(
         self, model: "pl.LightningModule", ckpt_path: Optional[str] = None
     ) -> Optional[Union[_EVALUATE_OUTPUT, _PREDICT_OUTPUT]]:
+        if model._compiler_ctx is not None:
```
tchaton, line 991: I think it would be cleaner for the strategy to implement a validation mechanism directly and for DDPStrategy / DDPFullyShardedNativeStrategy to uncompile the model and wrap their own forward model after wrapping.

src/pytorch_lightning/core/module.py
lantiga: Only in tests for now

tchaton (approved):
LGTM !

src/pytorch_lightning/core/module.py
```diff
@@ -1950,6 +1951,70 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
                 weakref.proxy(self), pre_load_state_dict_hook, True  # type: ignore[arg-type]
             )
 
+    @classmethod
+    def from_compiled(cls, model: Union["pl.LightningModule", "torch._dynamo.OptimizedModule"]) -> "pl.LightningModule":
```
ethanwharris, line 1955: Typed as LightningModule or OptimizedModule but then raises an error for anything that isn't an OptimizedModule:
```suggestion
    def from_compiled(cls, model: "torch._dynamo.OptimizedModule") -> "pl.LightningModule":
```

src/pytorch_lightning/core/module.py
```diff
@@ -1950,6 +1951,70 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
...
+                "`model` is expected to be a compiled LightingModule. " f"Found a compiled {type(orig_module)} instead"
+            )
+
+        orig_module._compiler_ctx = {
+            "compiler": "dynamo",
+            "dynamo_ctx": model.dynamo_ctx,
+            "original_forward": orig_module.forward,
+        }
+
+        orig_module.forward = model.dynamo_ctx(orig_module.forward)  # type: ignore[assignment]
```
tchaton, line 1985: Let's wrap `training_step`, some users might not use forward in their model.

ethanwharris (approved):
LGTM, very minor comment

justusschock (approved):
the code here looks good to me,

a general question: does ddp(compiled model) work? should we in that case call compile(ddp(orig model))? to allow for possibly better optimizations in distributed settings? what about all the other strategies?

src/pytorch_lightning/core/module.py
lantiga: great catch!

lantiga:
> the code here looks good to me,
> 
> a general question: does ddp(compiled model) work? should we in that case call compile(ddp(orig model))? to allow for possibly better optimizations in distributed settings? what about all the other strategies?

Excellent questions, probably compile(ddp(orig model)) is the way to go for distributed.
For now we want to unblock this this way and understand the best practices around distributed over the next period.

We only allow DDP and FSDP for now, which are the ones that are supported so far.

src/pytorch_lightning/trainer/trainer.py
lantiga: Agreed, this will be the way forward in the future, as soon as we understand more about dynamo and distributed.

src/pytorch_lightning/core/module.py
lantiga: of course, that makes sense

lantiga: I'm kind of hesitant because training_step contains things that dynamo doesn't seem to be tested on right now. I propose we merge this now and work on wrapping training_step later as the technology matures

src/pytorch_lightning/trainer/trainer.py
carmocca: @lantiga I agree with @tchaton. This could be implemented as a [plugin](https://github.com/Lightning-AI/lightning/tree/master/src/pytorch_lightning/plugins) that is specifically enabled for some of the strategies, Which would contain all this patching and unpatching logic.

If you agree, I would like to give it a shot

PR #15926, opened 2022-12-06
[App] Add `configure_layout` method for works

ethanwharris:
## What are you trying to solve?

Primary: We need a way for the framework to know if a given app is headless. Since many apps now rely on the implicit `configure_layout` implemented in the `_RootFlow`, we need a way to know if a work is expected to host a UI so we know whether or not to configure a layout there.

Secondary: A secondary goal is to clean up the use case where a work has an associated Frontend (e.g. the new serve component) - currently to achieve this using the Frontend in the normal way, you would need to create a Flow and implement `configure_layout` there.

## Options considered

We considered two options: adding a `work.configure_layout`, adding a flag like `has_ui` to the `LightningWork` which users would set to False if the work doesn't have a UI.

### `work.configure_layout`

- Pro: same hook name with a slightly different contract is easier to remember than a new flag
- Pro: a work can "own" its UI, so now works which have a UI on `/docs` or which are expected to be used with a particular frontend can specify that directly
- Con: maybe more boilerplate compared to the flag, users could need to implement `configure_layout` in their work for it to do what they want

### `has_ui` flag

- Pro: explicit + could help us clean up some internals if we know for sure that a work doesn't host a UI
- Con: one more thing to remember / possible bad experience if you don't remember to set it
- Con: doesn't allow for works to specify a preferred root URL or Frontend, so some use cases would still be challenging

## Design

From discussion with @williamFalcon we decided to go with the work `configure_layout` option. 

The `configure_layout` of a work can return either a string, a frontend, or None. A string is expected to be a URL. If the work returns None this is the same as not having a frontend (similar to returning an empty list from a flow, can be used to dynamically add frontends at runtime). If the work returns a frontend, then its parent flow will be responsible for hosting that frontend.

## Limitations

The main limitations is that if a flows `configure_layout` returns a tab whose content is a work with a frontend then that must be its only tab - this is something we can later address with some work (possibly a lot of work) but there are no existing use cases where this would happen.

Note: From discussion with @tchaton - one possible future evolution is to run the frontend directly in the work machine which would resolve this limitation.

No, if you don't implement `configure_layout` in your work and a flow points to it, then you get the old behaviour that we take the work URL.

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/lightning_app/utilities/layout.py
```diff
@@ -102,12 +104,42 @@ def _collect_content_layout(layout: List[Dict], flow: "lightning_app.LightningFl
...
+            work = entry["content"]
+            work_layout = _collect_work_layout(work)
+
+            if work_layout is None:
                 entry["content"] = ""
-                entry["target"] = ""
+            elif isinstance(work_layout, str):
+                entry["content"] = work_layout
+                entry["target"] = work_layout
+            else:
```
tchaton, line 115: Using elif would be cleaner.

tchaton (approved):
LGTM !

src/lightning_app/utilities/layout.py
```diff
@@ -102,12 +104,42 @@ def _collect_content_layout(layout: List[Dict], flow: "lightning_app.LightningFl
...
+                    The tab containing a `{work.__class__.__name__}` must be the only tab in the
+                    layout of this flow.
+
+                    (see the docs for `LightningWork.configure_layout`).
+                    """
+                    raise TypeError(m)
+
+                if isinstance(work_layout, Frontend):
+                    # If the work returned a frontend, treat it as belonging to the flow.
+                    frontend = work_layout
```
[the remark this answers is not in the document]
ethanwharris: Added a note in the comments

src/lightning_app/core/work.py
```diff
@@ -629,3 +632,45 @@ def apply_flow_delta(self, delta: Delta):
...
+
+            class Work(LightningWork):
+                def configure_layout(self):
+                    return [
+                        dict(name="First Tab", content=self.child0),
+                        dict(name="Second Tab", content=self.child1),
+                        dict(name="Lightning", content="https://lightning.ai"),
+                    ]
+
+        If you don't implement ``configure_layout``, Lightning will use ``self.url``.
```
hhsecond, line 669: Wouldn't this negate the original goal of figuring out whether a work has a UI or not

> Primary: We need a way for the framework to know if a given app is headless. Since many apps now rely on the implicit configure_layout implemented in the _RootFlow, we need a way to know if a work is expected to host a UI so we know whether or not to configure a layout there.

src/lightning_app/utilities/layout.py
```diff
@@ -126,3 +159,42 @@ def configure_layout(self):
...
+        return ""
+    elif isinstance(work_layout, (Frontend, _MagicMockJsonSerializable)):
+        return work_layout
+    else:
+        m = f"""
+        The value returned by `{work.__class__.__name__}.configure_layout()` is of an unsupported type.
+
+        {repr(work_layout)}
+
+        Return a `Frontend` or a URL, for example:
```
hhsecond, line 194:
```suggestion
        Return a `Frontend` or a URL string, for example:
```

src/lightning_app/utilities/layout.py
```diff
@@ -126,3 +159,42 @@ def configure_layout(self):
...
+    if is_overridden("configure_layout", work):
+        work_layout = work.configure_layout()
+    else:
+        work_layout = work.url
+
+    if work_layout is None:
+        return None
+    elif isinstance(work_layout, str):
+        url = work_layout
+        if url and not url.startswith("/"):
```
hhsecond, line 183:
```suggestion
        # The URL isn't fully defined yet. Looks something like ``self.work.url + /something``.
        if url and not url.startswith("/"):
```

hhsecond (approved):
LGTM

src/lightning_app/core/work.py
ethanwharris: In a way, the contract now is:
- if you manually point a flow to a work, we assume it has a UI (even if it doesn't have a `configure_layout`)
- if you pass a work directly to `LightningApp`, the root flow will only point to it if it overrides `configure_layout`

So there is an edge case there if your work has a UI, doesn't override `configure_layout` and is passed on its own to `LightningApp`, but this was the only way to preserve backwards compatibility

tchaton:
@awaelchli Could you review this PR ?

awaelchli (approved):

PR #16017, opened 2022-12-12
Nightly PyTorch version is now 2.0

carmocca:
## What does this PR do?

PyTorch nightlies have 2.0 as the version number, signaling that there won't be a 1.14 release. https://github.com/pytorch/pytorch/pull/90491

None


justusschock (approved):

awaelchli (approved):

Borda:
> signaling that there won't be a 1.14 release

can you share a reference to why you thin so?

src/pytorch_lightning/callbacks/quantization.py
```diff
@@ -26,8 +26,8 @@
 from torch.quantization import FakeQuantizeBase
 
 import pytorch_lightning as pl
+from lightning_lite.utilities.imports import _TORCH_GREATER_EQUAL_1_11, _TORCH_GREATER_EQUAL_1_12
```
Borda, line 29: can we do this in separate PR?

src/pytorch_lightning/utilities/imports.py
```diff
@@ -13,20 +13,11 @@
...
-import platform
-import sys
 
 import torch
 from lightning_utilities.core.imports import compare_version, module_available, package_available, RequirementCache
 
-_IS_WINDOWS = platform.system() == "Windows"
-_PYTHON_GREATER_EQUAL_3_8_0 = (sys.version_info.major, sys.version_info.minor) >= (3, 8)
-_PYTHON_GREATER_EQUAL_3_10_0 = (sys.version_info.major, sys.version_info.minor) >= (3, 10)
 _TORCH_LESSER_EQUAL_1_10_2 = compare_version("torch", operator.le, "1.10.2")
```
Borda, line 20: why this stay and others are moved?

src/lightning_lite/accelerators/cuda.py
```diff
@@ -20,7 +20,7 @@
 import torch
 
 from lightning_lite.accelerators.accelerator import Accelerator
-from lightning_lite.utilities.imports import _TORCH_GREATER_EQUAL_1_13, _TORCH_GREATER_EQUAL_1_14
+from lightning_lite.utilities.imports import _TORCH_GREATER_EQUAL_1_13, _TORCH_GREATER_EQUAL_2_0
```
Borda, line 23: I would still keep this...

Borda (commented):
seem that the PR title is misleading as it is rather using imports from Lite then changing anything to 3.0

src/lightning_lite/accelerators/cuda.py
carmocca: There's no 1.14: https://github.com/pytorch/pytorch/pull/90491

src/pytorch_lightning/utilities/imports.py
carmocca: Lite does not define this one. The others are, so they were duplicated

src/pytorch_lightning/callbacks/quantization.py
carmocca: It's just removing one duplicated variable

src/lightning_lite/accelerators/cuda.py
Borda: it was nor merged, so not sure :confused:

src/pytorch_lightning/callbacks/quantization.py
Borda: hmm but your removing is 95% of this PR :P

src/lightning_lite/accelerators/cuda.py
carmocca: It is merged. See the label in the PR

src/pytorch_lightning/callbacks/quantization.py
Borda: I would just update the PR name, but @carmocca I know may complain :)

PR #16073, opened 2022-12-15
Clarify `work.stop()` limitation

carmocca:
## What does this PR do?

See title

None

cc @borda


src/lightning_app/core/work.py
```diff
@@ -609,12 +609,12 @@ def on_exit(self):
...
 
     def stop(self):
-        """Stops LightingWork component and shuts down hardware provisioned via L.CloudCompute."""
+        """Stops LightingWork component and shuts down hardware provisioned via L.CloudCompute.
+
+        This can only be called from a ``LightningFlow``.
+        """
         if not self._backend:
-            raise Exception(
-                "Can't stop the work, it looks like it isn't attached to a LightningFlow. "
```
carmocca, line 615: I was trying this, and I had a Flow attached, so I changed the message. Tell me if I should keep the old message and mention that add that it can only be called from the flow

awaelchli (approved):

Borda (approved):

PR #16074, opened 2022-12-15
Callback support for Lite

lightningforever:
## What does this PR do?

```py
fabric = Fabric(callbacks=[callback0, callback1])
fabric.call("on_train_end")
```

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_fabric/fabric.py
+++ b/src/lightning_fabric/fabric.py
@@ -84,6 +84,7 @@ class Fabric:
         num_nodes: int = 1,
         precision: _PRECISION_INPUT = 32,
         plugins: Optional[Union[_PLUGIN_INPUT, List[_PLUGIN_INPUT]]] = None,
+        callbacks: Optional[Union[List[Any], Any]] = None,
     ) -> None:
         self._connector = _Connector(
             accelerator=accelerator,
@@ -96,6 +97,8 @@ class Fabric:
         self._strategy: Strategy = self._connector.strategy
         self._accelerator: Accelerator = self._connector.accelerator
         self._precision: Precision = self._strategy.precision
+        callbacks = callbacks if callbacks is not None else []
+        self._callbacks = callbacks if isinstance(callbacks, list) else [callbacks]
         self._models_setup: int = 0
 
         self._prepare_run_method()
@@ -517,6 +520,20 @@ class Fabric:
         if self._strategy.launcher is not None:
             return self._strategy.launcher.launch(function, *args, **kwargs)
         return function(*args, **kwargs)
+
+    def call(self, hook_name: str, *args, **kwargs: Any) -> None:
+        # TODO(fabric): Support also passing down positional arguments
+        for callback in self._callbacks:
+            method = getattr(callback, hook_name, None)
+            if method is None:
+                continue
+            method(*args, **kwargs)
+
+            # TODO(fabric): handle the following signatures
+            # method(self, fabric, x, y=1)
+            # method(self, fabric, *args, x, y=1)
+            # method(self, *args, y=1)
+            # method(self, *args, **kwargs)
 
     @staticmethod
     def seed_everything(seed: Optional[int] = None, workers: Optional[bool] = None) -> int:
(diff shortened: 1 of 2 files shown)


Borda (approved):

carmocca (approved):

src/lightning_fabric/fabric.py
```diff
@@ -518,6 +521,38 @@ def launch(self, function: Optional[Callable[["Fabric"], Any]] = None, *args: An
...
+                    ...
+
+            fabric = Fabric(callbacks=[MyCallback]))
+            fabric.call("on_train_epoch_end", results={...})
+        """
+        for callback in self._callbacks:
+            method = getattr(callback, hook_name, None)
+            if method is None:
+                continue
+            method(*args, **kwargs)
```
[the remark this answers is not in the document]
lightningforever: done

src/lightning_fabric/fabric.py
```diff
@@ -518,6 +521,44 @@ def launch(self, function: Optional[Callable[["Fabric"], Any]] = None, *args: An
             return self._strategy.launcher.launch(function, *args, **kwargs)
         return function(*args, **kwargs)
 
+    def call(self, hook_name: str, *args: Any, **kwargs: Any) -> None:
```
justusschock, line 524: what about also exposing fabric.on_batch_start? So a direct call? because this one again males it impossible to search for usage of a given function using your IDE.

I get that this one is more convenient, just wanted to raise the question:)

justusschock (approved):

src/lightning_fabric/fabric.py
lightningforever: I think we would have to do either one or the other. If we do both, it's one more thing we have to explain to the users what the difference is between both of them. 

Also, there is no callback interface, so even if you get the auto-suggestion through the type, there is no contract guaranteeing you that the hook exists on your callback (so you'd have to look it up anyway).

lightningforever: cc @lantiga

justusschock: so the benefit is that the user could call whatever hook they implemented? That's sufficient reason for me not to go with the explicit version :)

lantiga: agreed, let's stay with strings for now

personally I find the naming doesn't evoke an event being emitted, rather `call` sounds like a direct call - so one wonders why you're not calling the function directly.

I'm not against it, but I would find something like `trigger` more intuitive.

this may just be me though, I'd like to hear from the others.

lantiga (approved):
Great, just a naming proposal

src/lightning_fabric/fabric.py
justusschock: Regarding the naming: I'd stick with `call` as this basically immediately calls the callbacks with some additional checks. `trigger` might indicate that we built a fully-fledged event-based system around this, which we don't want to do.

carmocca: At the moment, it is a direct call, just done through Fabric. The user could call the function directly too if they have the reference to the callback.

"call" also fits with **call**back

but I don't mind choosing trigger or emit if anybody prefers.

src/lightning_fabric/fabric.py
```diff
@@ -518,6 +521,44 @@ def launch(self, function: Optional[Callable[["Fabric"], Any]] = None, *args: An
...
+            method = getattr(callback, hook_name, None)
+            if method is None:
+                continue
+            if not callable(method):
+                rank_zero_warn(
+                    f"Skipping the callback `{type(callback).__name__}.{hook_name}` because it is not callable."
+                )
+                continue
+
+            method(*args, **kwargs)
```
justusschock, line 554: could we do something like `return_types = [callback.method(*args, **kwargs) for callback in self._callbacks]`? So that you don't need to store it in internal state if you don't want to?

awaelchli: I think we should add this based on need, in a future iteration.

justusschock: @carmocca I agree that dict-like would be great, but then you'd also need a dict-like structure for your callbacks and in practice, execution-order of callbacks could be important although it shouldn't.

@awaelchli agreed!

src/lightning_fabric/fabric.py
lightningforever: @lantiga I agree we may need to iterate on the name, and get William Falcon's approval too. I suggest we merge the PR so we can also finish the examples, but keep it open to change the name from feedback.

PR #16080, opened 2022-12-15
Better check for programmatic lightningignore

carmocca:
## What does this PR do?

Fixes regression where the `LightningTrainerMultiNode` component fails to start because the work is re-created several times even after the app has been dispatched.

Avoiding the use of an environment variable fixes the issue. `self._backend is not None` can act as a proxy for this as suggested by @tchaton

This should get tested by re-enabling the multinode tests

None

cc @borda


awaelchli (approved):

.github/workflows/ci-app-examples.yml
```diff
@@ -86,10 +86,11 @@ jobs:
       - name: Install Yarn
         run: npm install -g yarn
 
-      - name: Install Lightning package
+      - name: Install package
```
Borda, line 89:
```suggestion
      - name: Install Lightning package
```

Borda (approved):

.github/workflows/ci-app-examples.yml
carmocca: This step also installs the lightning app package

tchaton (approved):
LGTM !

PR #16097, opened 2022-12-17
docs: drop redundant defaults

Borda:
## What does this PR do?

these defaults are already shared as the footprint of function/method so having then in docs is duplicate and also not often updated

## Before submitting

## PR review

## Did you have fun?

cc @borda


src/pytorch_lightning/trainer/trainer.py
```diff
@@ -195,111 +193,87 @@ def __init__(
                 trying to optimize initial learning for faster convergence. trainer.tune() method will
                 set the suggested learning rate in self.lr or self.learning_rate in the LightningModule.
                 To use a different key set a string instead of True with the key name.
-                Default: ``False``.
```
awaelchli, line 198: Some of these were added by contributors: #11614

As a general guideline I agree, but it also doesn't hurt having them in something like the core Trainer API (that doesn't really change their defaults that often). 

I propose to keep these removals to a minimum, and to keep them in the Trainer.

Borda: I see, not sure why we moved some back...
In my point of view, I am very fine to keep the default when there is some additional explanation or referring to the class type or just in text after a value set that the one is default... but having pure sentence "Default: None." I think is useless...

carmocca: I suggest keeping the "default:" message for the arguments that are Optional in the code only so that we can know whether the user manually passed an argument or not to then set the real default. All others could be removed

Borda: > arguments that are Optional in the code only so that we can know whether the user manually passed an argument

not sure if I understand what you mean here :confused:

carmocca: We often use the pattern

```py
x: Optional[int]
if x is None:
    # the user didn't set it
    self.x = 10
else:
    self.x = x
```
instead of

```pyt
x: int = 10
self.x = x
```

So for those, we want to still mention `Default: 10` in the docstring because saying `Default: None` is not informative unless it also explains that `None` gets converted into `10`

Borda: yes agree, but so far it was not such and we just said default None...

tchaton (approved):
LGTM 1

src/pytorch_lightning/trainer/trainer.py
awaelchli: I'm ok with adjusting the other less important files. But my opinion remains unchanged regarding trainer.py, I'd like that it stays and that we don't go back and forth #11614

awaelchli: But @Borda how about we merge the PR but remove the changes from Trainer. That's what I'm trying to say.

PR #16121, opened 2022-12-19
Logger support in Lite

lightningforever:
## What does this PR do?

This PR adds the logger base class and TensorBoardLogger implementation to Fabric.
```py
logger = TensorBoardLogger()
fabric = LightningLite(loggers=logger)

fabric.log("accuracy", acc)
fabric.log_dict({"accuracy": acc, "loss": loss})
```

The base class interface is mostly the same as in PyTorch Lightning, but with these key differences:

- Removed model checkpoint callback dependency and `after_save_checkpoint` hook. Lite won't have the same checkpointing mechanism through callbacks, and thus we will have to rework the way loggers access checkpoints.
- Removed `DummyLogger` (only needed for Trainer)
- Moved properties to the top (code organization)
- Property: `save_dir` -> `root_dir`. We are standardizing the way the log dir is defined, to be compatible with as many logger frameworks as possible. More info in #14188. 

The TensorBoard implementation changes very little. Trainer-only features and dependencies were removed:

- Removed hparams saving
- Removed OmegaConf dependency
- `save_dir` -> `root_dir`
- Removed `log_graph` argument in constructor (logging graph is still supported through `logger.log_graph()`)

**Future Work:**
- Support for `self.log` in LightningModule will be added in #16135

No

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/docs/source-pytorch/fabric/fabric.rst
+++ b/docs/source-pytorch/fabric/fabric.rst
@@ -655,4 +655,11 @@ These methods allows you to send scalar metrics to a logger registered in Fabric
     # Or send multiple metrics at once:
     fabric.log_dict({"loss": loss, "accuracy": acc})
 
-If no loggers are given to Fabric (default), `log` and `log_dict` won't do anything.
+If no loggers are given to Fabric (default), ``log`` and ``log_dict`` won't do anything.
+Here is what's happening under the hood (pseudo code) when you call ``.log()`` or ``log_dict``:
+
+.. code-block:: python
+
+    # When you call .log() or .log_dict(), we do this:
+    for logger in fabric.loggers:
+        logger.log_metrics(metrics=metrics, step=step)
--- a/docs/source-pytorch/fabric/fabric.rst
+++ b/docs/source-pytorch/fabric/fabric.rst
@@ -385,6 +385,30 @@ Then, in your training loop, you can call a hook by its name. Any callback objec
     fabric.call("on_train_epoch_end", results={...})
 
 
+loggers
+=======
+
+Attach one or several loggers/experiment trackers to Fabric for convenient logging of metrics.
+
+.. code-block:: python
+
+    # Default used by Fabric, no loggers are active
+    fabric = Fabric(loggers=[])
+
+    # Log to a single logger
+    fabric = Fabric(loggers=TensorBoardLogger(...))
+
+    # Or multiple instances
+    fabric = Fabric(loggers=[logger1, logger2, ...])
+
+Anywhere in your training loop, you can log metrics to all loggers at once:
+
+.. code-block:: python
+
+    fabric.log("loss", loss)
+    fabric.log_dict({"loss": loss, "accuracy": acc})
+
+
 ----------
 
 
@@ -613,3 +637,22 @@ It is useful when building a Trainer that allows the user to run arbitrary code 
 
     # Only the callbacks that have this method defined will be executed
     fabric.call("undefined")
+
+
+log and log_dict
+================
+
+These methods allows you to send scalar metrics to a logger registered in Fabric.
+
+.. code-block:: python
+
+    # Set the logger in Fabric
+    fabric = Fabric(loggers=TensorBoardLogger(...))
+
+    # Anywhere in your training loop or model:
+    fabric.log("loss", loss)
+
+    # Or send multiple metrics at once:
+    fabric.log_dict({"loss": loss, "accuracy": acc})
+
+If no loggers are given to Fabric (default), `log` and `log_dict` won't do anything.
--- a/src/lightning_fabric/fabric.py
+++ b/src/lightning_fabric/fabric.py
@@ -748,3 +748,4 @@ def _is_using_cli() -> bool:
 
 def _do_nothing(*_: Any) -> None:
     pass
+
--- a/src/lightning_fabric/fabric.py
+++ b/src/lightning_fabric/fabric.py
@@ -748,4 +748,3 @@ def _is_using_cli() -> bool:
 
 def _do_nothing(*_: Any) -> None:
     pass
-
--- a/src/lightning_fabric/fabric.py
+++ b/src/lightning_fabric/fabric.py
@@ -22,9 +22,12 @@ import torch
 import torch.nn as nn
 from lightning_utilities.core.apply_func import apply_to_collection
 from lightning_utilities.core.overrides import is_overridden
+from lightning_utilities.core.rank_zero import rank_zero_warn
 from torch import Tensor
 from torch.optim import Optimizer
 from torch.utils.data import BatchSampler, DataLoader, DistributedSampler, RandomSampler, SequentialSampler
+
+from lightning_fabric.loggers import Logger
 
 from lightning_fabric.plugins import Precision  # avoid circular imports: # isort: split
 from lightning_fabric.accelerators.accelerator import Accelerator
@@ -47,7 +50,6 @@ from lightning_fabric.utilities.data import (
     has_iterable_dataset,
 )
 from lightning_fabric.utilities.distributed import DistributedSamplerWrapper
-from lightning_fabric.utilities.rank_zero import rank_zero_warn
 from lightning_fabric.utilities.seed import seed_everything
 from lightning_fabric.utilities.warnings import PossibleUserWarning
 from lightning_fabric.wrappers import _FabricDataLoader, _FabricModule, _FabricOptimizer
@@ -74,6 +76,10 @@ class Fabric:
         precision: Double precision (``64``), full precision (``32``), half precision (``16``),
             or bfloat16 precision (``"bf16"``).
         plugins: One or several custom plugins
+        callbacks: A single callback or a list of callbacks. A callback can contain any arbitrary methods that
+            can be invoked through :meth:`lightning_fabric.fabric.Fabric.call` by the user.
+        loggers: A single logger or a list of loggers. See :meth:`lightning_fabric.fabric.Fabric.log` for more
+            information.
     """
 
     def __init__(
@@ -85,6 +91,7 @@ class Fabric:
         precision: _PRECISION_INPUT = 32,
         plugins: Optional[Union[_PLUGIN_INPUT, List[_PLUGIN_INPUT]]] = None,
         callbacks: Optional[Union[List[Any], Any]] = None,
+        loggers: Optional[Union[Logger, List[Logger]]] = None,
     ) -> None:
         self._connector = _Connector(
             accelerator=accelerator,
@@ -99,6 +106,8 @@ class Fabric:
         self._precision: Precision = self._strategy.precision
         callbacks = callbacks if callbacks is not None else []
         self._callbacks = callbacks if isinstance(callbacks, list) else [callbacks]
+        loggers = loggers if loggers is not None else []
+        self._loggers = loggers if isinstance(loggers, list) else [loggers]
         self._models_setup: int = 0
 
         self._prepare_run_method()
@@ -147,6 +156,16 @@ class Fabric:
     def is_global_zero(self) -> bool:
         """Whether this rank is rank zero."""
         return self._strategy.is_global_zero
+
+    @property
+    def loggers(self) -> List[Logger]:
+        """Returns all loggers passed to Fabric."""
+        return self._loggers
+
+    @property
+    def logger(self) -> Logger:
+        """Returns the first logger in the list passed to Fabric, which is considered the main logger."""
+        return self._loggers[0]
 
     def run(self, *args: Any, **kwargs: Any) -> Any:
         """All the code inside this run method gets accelerated by Fabric.
@@ -572,6 +591,28 @@ class Fabric:
             # method(self, fabric|trainer, *args, x, y=1)
             # method(self, *args, y=1)
             # method(self, *args, **kwargs)
+
+    def log(self, name: str, value: Any, step: Optional[int] = None) -> None:
+        """Log a scalar to all loggers that were added to Fabric.
+
+        Args:
+            name: The name of the metric to log.
+            value: The metric value to collect.
+            step: Optional step number. Most Logger implementations auto-increment the step value by one with every
+                log call. You can specify your own value here.
+        """
+        self.log_dict(metrics={name: value}, step=step)
+
+    def log_dict(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
+        """Log multiple scalars at once to all loggers that were added to Fabric.
+
+        Args:
+            metrics: A dictionary where the key is the name of the metric and the value the scalar to be logged.
+            step: Optional step number. Most Logger implementations auto-increment this value by one with every
+                log call. You can specify your own value here.
+        """
+        for logger in self._loggers:
+            logger.log_metrics(metrics=metrics, step=step)
 
     @staticmethod
     def seed_everything(seed: Optional[int] = None, workers: Optional[bool] = None) -> int:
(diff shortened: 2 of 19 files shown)


src/lightning_fabric/utilities/logger.py
```diff
@@ -0,0 +1,151 @@
+# Copyright The PyTorch Lightning team.
```
lightningforever, line 1: These util functions were moved without modification

src/pytorch_lightning/utilities/logger.py
```diff
@@ -13,148 +13,12 @@
...
-
-import numpy as np
-import torch
-from torch import Tensor
+from typing import List, Tuple
 
 from pytorch_lightning.callbacks import Checkpoint
 
 
-def _convert_params(params: Union[Dict[str, Any], Namespace]) -> Dict[str, Any]:
```
lightningforever, line 27: These util functions were moved to lightning_fabric/utilities/loggers.py

src/lightning_fabric/loggers/logger.py
```diff
@@ -0,0 +1,136 @@
...
+        """Return the experiment version."""
+
+    @property
+    def root_dir(self) -> Optional[str]:
+        """Return the root directory where all versions of an experiment get saved, or `None` if the logger does
+        not save data locally."""
+        return None
+
+    @property
+    def log_dir(self) -> Optional[str]:
```
lightningforever, line 47: Note: This will be what we initially proposed as "experiment_dir" in #14188

src/lightning_fabric/loggers/logger.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @abstractmethod
+    def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
+        """Records metrics. This method logs metrics as soon as it received them.
+
+        Args:
+            metrics: Dictionary with metric names as keys and measured quantities as values
+            step: Step number at which the metrics should be recorded
+        """
+        pass
```
lightningforever, line 65: This should probably become `log_dict`, and another method `log`, to align with the rest Lightning's naming.

src/lightning_fabric/loggers/tensorboard.py
```diff
@@ -0,0 +1,310 @@
...
+        logger = TensorBoardLogger("path/to/logs/rot", name="my_model")
+        logger.log_hyperparams({"epochs": 5, "optimizer": "Adam"})
+        logger.log_metrics({"acc": 0.75})
+        logger.finalize("success")
+    """
+    LOGGER_JOIN_CHAR = "-"
+
+    def __init__(
+        self,
+        root_dir: _PATH,
```
justusschock, line 84: since we don't have the default root dir from trainer here, should we combine `root_dir`, `name` and `sub_dir` to a single path?

justusschock (approved):

src/lightning_fabric/loggers/logger.py
carmocca: I don't agree. What woudl be the difference between `log` and `log_dict` for this class specifically?

`log_metrics` is explicit and avoids collision with a polymorphic `log()` method if one of the implementations support it

src/lightning_fabric/loggers/tensorboard.py
```diff
@@ -0,0 +1,310 @@
...
+_TENSORBOARDX_AVAILABLE = RequirementCache("tensorboardX")
+if TYPE_CHECKING:
+    # assumes at least one will be installed when type checking
+    if _TENSORBOARD_AVAILABLE:
+        from torch.utils.tensorboard import SummaryWriter
+    else:
+        from tensorboardX import SummaryWriter  # type: ignore[no-redef]
+
+
+class TensorBoardLogger(Logger):
```
carmocca, line 44: Don't we want the `pytorch_lightning.loggers.tensorboard` module to import this class? Without deprecations

src/lightning_fabric/loggers/logger.py
```diff
@@ -0,0 +1,136 @@
...
+from functools import wraps
+from typing import Any, Callable, Dict, Optional, Union
+
+from torch import Tensor
+from torch.nn import Module
+
+from lightning_fabric.utilities.rank_zero import rank_zero_only
+
+
+class Logger(ABC):
```
carmocca, line 27: Same comment about importing this from PL

src/lightning_fabric/fabric.py
```diff
@@ -148,6 +157,16 @@ def is_global_zero(self) -> bool:
         """Whether this rank is rank zero."""
         return self._strategy.is_global_zero
 
+    @property
+    def loggers(self) -> List[Logger]:
+        """Returns all loggers passed to Fabric."""
+        return self._loggers
+
+    @property
+    def logger(self) -> Logger:
```
Borda, line 166: shall this have index as an argument so not to be a property?

src/lightning_fabric/loggers/logger.py
lightningforever: It depends on what product wants the logger interaction and experience to be. Will a user interact mainly with the fabric object and its log methods? Or will a user mainly interact with the logger object, i.e., `logger.log_()*`, `logger.experiment.xyz()`? 
Undoubtedly, power users will want to log media datatypes and not just metrics. I don't see how that would all be enabled through the fabric object. And then users would write code with a mix of `fabric.log` and `logger.log_metrics` and `logger.experiment.log_image`.

src/lightning_fabric/loggers/tensorboard.py
lightningforever: I'm not sure we want to deviate that strongly from the design yet. I believe it's too early to tell. Eventually this logger interface will again be used by the Trainer through Lite, and we don't know whether there will be a "default_root_dir" or not. 

Btw, subdir gets concatenated after the version. A full path would look like this

`root_dir/name/version/sub_dir`

not

`root_dir/name/sub_dir/version`

src/lightning_fabric/fabric.py
lightningforever: This can already be accessed like so through an index:

```py
fabric.loggers[2]
```

and 

```py
fabric.logger == fabric.loggers[0]
```

Or did you mean something else?

src/lightning_fabric/loggers/tensorboard.py
lightningforever: I think we want to eventually. In Lite however we currently have more freedom to make minor changes without triggering us to do immediately do refactors that touch many areas of Trainer.

tchaton: Yes, let's make them as we have the freedom to do so.

tchaton (approved):
LGMT !

src/lightning_fabric/loggers/tensorboard.py
justusschock: I know, but IMO we should simplify as much as possible here. If I see a logger I expect there to be an option with one path and I'd have it log exactly at that path. TBH this is a change I would love to also make for the trainer (once we can without backwards compatibility).

lightningforever: I like that idea, I can explore it, but would like to do that in a separate PR because it would change all tests. Since the implementation is copied from PL, I'd like to keep the changes in this initial PR as minimal as possible.

src/lightning_fabric/loggers/logger.py
carmocca: Regardless of what `fabric.log()` supports, the logger interface don't need to have a matching name. This is how it is in PL at the moment.

src/lightning_fabric/loggers/tensorboard.py
carmocca: I don't think we are planning any changes to the logger objects themselves. Do you have any in mind? And any differences could be resolved by inheritance if necessary later.

lightningforever: The changes are listed in the PR summary.
Also this: https://github.com/Lightning-AI/lightning/pull/16121#discussion_r1054375539

src/lightning_fabric/loggers/logger.py
lightningforever: Not planning to change it in this PR. We should continue the discussion elsewhere, also in the context of what self.log should do and where in 2.0. Resolving

docs/source-pytorch/fabric/fabric.rst
```diff
@@ -589,3 +613,22 @@ For single-device strategies, it is a no-op. There are strategies that don't sup
 - xla
 
 For these, the context manager falls back to a no-op and emits a warning.
+
+
+log and log_dict
```
carmocca, line 618: This addition and the one above look repeated. They basically say the same thing. Keep one? Merge them?

docs/source-pytorch/fabric/fabric.rst
```diff
@@ -589,3 +613,22 @@ For single-device strategies, it is a no-op. There are strategies that don't sup
...
+    # Set the logger in Fabric
+    fabric = Fabric(loggers=TensorBoardLogger(...))
+
+    # Anywhere in your training loop or model:
+    fabric.log("loss", loss)
+
+    # Or send multiple metrics at once:
+    fabric.log_dict({"loss": loss, "accuracy": acc})
+
+If no loggers are given to Fabric (default), `log` and `log_dict` won't do anything.
```
carmocca, line 634: Can you add two lines of pseudocode showing that `fabric.log` basically routes to `logger.log_*`? This way it's clear to the users that they could also use the logger themselves.

src/pytorch_lightning/core/module.py
```diff
@@ -291,12 +291,12 @@ def truncated_bptt_steps(self, truncated_bptt_steps: int) -> None:
         self._truncated_bptt_steps = truncated_bptt_steps
 
     @property
-    def logger(self) -> Optional[Logger]:
+    def logger(self) -> Optional[Union[Logger, TensorBoardLogger]]:
```
carmocca, line 294: Did you make this change because of mypy? This shouldn't be necessary, no?

src/pytorch_lightning/loggers/tensorboard.py
```diff
@@ -19,37 +19,28 @@
...
 from lightning_fabric.utilities.types import _PATH
+from pytorch_lightning.callbacks import ModelCheckpoint
 from pytorch_lightning.core.saving import save_hparams_to_yaml
-from pytorch_lightning.loggers.logger import Logger, rank_zero_experiment
 from pytorch_lightning.utilities.imports import _OMEGACONF_AVAILABLE
 from pytorch_lightning.utilities.rank_zero import rank_zero_only, rank_zero_warn
 
 log = logging.getLogger(__name__)
 
 _TENSORBOARD_AVAILABLE = RequirementCache("tensorboard")
```
carmocca, line 38: This can be imported from Fabric

src/pytorch_lightning/loggers/logger.py
```diff
@@ -66,84 +39,12 @@ def after_save_checkpoint(self, checkpoint_callback: ModelCheckpoint) -> None:
...
-
-    def __getitem__(self, idx: int) -> "DummyExperiment":
-        # enables self.logger.experiment[0].add_image(...)
-        return self
-
-    def __setitem__(self, *args: Any, **kwargs: Any) -> None:
-        pass
-
 
 class DummyLogger(Logger):
```
carmocca, line 49: Is this not relevant for Fabric?

tests/tests_fabric/conftest.py
```diff
@@ -54,6 +54,7 @@ def restore_env_variables():
         "RANK",  # set by DeepSpeed
         "POPLAR_ENGINE_OPTIONS",  # set by IPUStrategy
         "CUDA_MODULE_LOADING",  # leaked since PyTorch 1.13
+        "CRC32C_SW_MODE",  # leaked by tensorboardX
```
carmocca, line 57: Fabric tests with tensorboard. Either the comment is not correct or this shouldn't be necessary

tests/tests_fabric/loggers/test_tensorboard.py
```diff
@@ -0,0 +1,229 @@
...
+from unittest import mock
+from unittest.mock import Mock
+
+import numpy as np
+import pytest
+import torch
+from tests_fabric.test_fabric import BoringModel
+
+from lightning_fabric.loggers import TensorBoardLogger
+from lightning_fabric.loggers.tensorboard import _TENSORBOARD_AVAILABLE
```
carmocca, line 26: Do you plan to remove any TB/logger tests from PL?

src/pytorch_lightning/core/module.py
lightningforever: Logger here is the Logger base class from PyTorch Lightning
TensorBoardLogger here is subclassing FabricTensorBoardLogger which is subclassing the base class in Fabric

So these don't have the same base class, and therefore need to be included for typing separately. This is a consequence of your request to reuse code through the inheritance.

src/pytorch_lightning/loggers/logger.py
lightningforever: Not needed for now. If it's needed we will move it. Note: If I had moved it, another version of @carmocca would have said: "This isn't used anywhere in Fabric, so we should not have it there" :))

tests/tests_fabric/conftest.py
lightningforever: What does it matter. It is the same as in pytorch's version of conftest. 
Whether the tests run one way or another depends on which package is installed. Either you have TB installed, or you have TBX installed or both, or none. The tests should run/skip accordingly no matter which configuration.

docs/source-pytorch/fabric/fabric.rst
lightningforever: The above is for the loggers Fabric argument, the one here if for the method(s). I want people to find this information regardless of where they look. This is very brief, there will be a separate guide page that goes into more details.

lightningforever: Organization will mainly happen in https://github.com/Lightning-AI/lightning/pull/16291 and following PRs.

docs/source-pytorch/fabric/fabric.rst
lightningforever: done

tests/tests_fabric/loggers/test_tensorboard.py
lightningforever: I think it is too soon now. The loggers in Fabric may change in ways that require us to do overrides again on the PL side (yes I am still against sharing the code through the subclass)

src/pytorch_lightning/loggers/logger.py
carmocca: You did the correct thing by not moving it. I was just asking if there were any future plans to use it in Fabric. I guess not because there's no concept of fast dev run

src/pytorch_lightning/core/module.py
carmocca: This is passing mypy locally for me: [`2e90ddf` (#16121)](https://github.com/Lightning-AI/lightning/pull/16121/commits/2e90ddf7d0738bb9f24c268e93a5330dd3ff4c2f)

It just changes the base import to Fabric's. Do you see any issues with this?

carmocca (approved):

src/pytorch_lightning/loggers/logger.py
lightningforever: Yes. One could build a trainer with fabric and have a fast_dev_run feature. In this case, this dummy logger would be a mere utility piece in Fabric.

PR #16147, opened 2022-12-20
Deprecate `auto_select_gpus`

### LINKED ISSUE #10535: [RFC] Deprecate `auto_select_gpus` from Trainer constructor

### LINKED ISSUE #13079: RFC: Deprecate  `auto_select_gpus` Trainer argument

awaelchli:
## What does this PR do?

Fixes #13079
Fixes #10535
Alternative to #13356

Before
```py
# Picks the first two available GPUs
Trainer(devices=2, auto_select_gpus=True)
```

After
```py
# Picks the first two available GPUs
from pytorch_lightning.accelerators import find_usable_cuda_gpus

Trainer(devices=find_usable_cuda_gpus)
```

## Before submitting

## PR review

## Did you have fun?

cc @tchaton @justusschock @awaelchli @borda


docs/source-pytorch/common/trainer.rst
```diff
@@ -317,6 +317,10 @@ before any training.
 auto_select_gpus
 ^^^^^^^^^^^^^^^^
 
+.. warning:: ``auto_select_gpus`` has been deprecated in v1.9.0 and will be removed in v1.10.0.
```
carmocca, line 320: Normally we just remove mentions of deprecated code from the docs

docs/source-pytorch/accelerators/gpu_basic.rst
```diff
@@ -88,10 +88,25 @@ The table below lists examples of possible input formats and how they are interp
...
-    For more details see the :doc:`trainer guide <../common/trainer>`.
+Find usable CUDA devices
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you want to run several experiments at the same time on your machine, for example for a hyperparameter sweep, then you can
+use the following utility function to pick GPU indices that are "accessible", without having to change your code every time.
+
+.. code-block:: python
+
+    from lightning.pytorch.accelerators import find_usable_cuda_gpus
```
carmocca, line 100: Can this be

```suggestion
    from lightning.pytorch.accelerators.cuda import find_usable_cuda_gpus
```

src/lightning_lite/accelerators/__init__.py
```diff
@@ -13,6 +13,7 @@
 from lightning_lite.accelerators.accelerator import Accelerator  # noqa: F401
 from lightning_lite.accelerators.cpu import CPUAccelerator  # noqa: F401
 from lightning_lite.accelerators.cuda import CUDAAccelerator  # noqa: F401
+from lightning_lite.accelerators.cuda import find_usable_cuda_gpus  # noqa: F401
```
carmocca, line 16: IMO this utility should be left at the `cuda` level

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,61 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
         )
 
 
-def _get_all_available_cuda_gpus() -> List[int]:
+def find_usable_cuda_gpus(num_gpus: int = -1) -> List[int]:
```
carmocca, line 81: I suggest using "devices" instead of "gpus" for consistency with the trainer flag

```suggestion
def find_usable_cuda_devices(num_devices: int = -1) -> List[int]:
```

If you agree, this should also be changed in the docstrings etc

src/pytorch_lightning/trainer/connectors/accelerator_connector.py
```diff
@@ -558,8 +558,17 @@ def _set_devices_flag_if_auto_passed(self) -> None:
...
+            rank_zero_deprecation(
+                "The Trainer argument `auto_select_gpus` has been deprecated in v1.9.0 and will be removed in v1.10.0."
+                " Please use the function `pytorch_lightning.accelerators.find_usable_cuda_gpus` instead."
+            )
         if self._auto_select_gpus and isinstance(self._gpus, int) and isinstance(self.accelerator, CUDAAccelerator):
-            self._devices_flag = pick_multiple_gpus(self._gpus)
+            self._devices_flag = pick_multiple_gpus(
+                self._gpus,
+                # we already show a deprecation message when user sets Trainer(auto_select_gpus=...)
+                __show_deprecatio=False,
```
carmocca, line 570:
```suggestion
                __show_deprecation=False,
```

carmocca (commented):
Early review

docs/source-pytorch/common/trainer.rst
awaelchli: The entire section?

src/lightning_lite/accelerators/__init__.py
awaelchli: The import would be too long:
`lightning.fabric.accelerators.cuda`

docs/source-pytorch/common/trainer.rst
carmocca: Yes. All mentions of the deprecated functionality. This way new users don't even bother with deprecated code.

Borda (approved):

justusschock (approved):

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
         )
 
 
-def _get_all_available_cuda_gpus() -> List[int]:
+def find_usable_cuda_devices(num_devices: int = -1) -> List[int]:
+    """Returns a list of all available and usable CUDA GPUs.
```
carmocca, line 82:
```suggestion
    """Returns a list of all available and usable CUDA devices.
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+
+    A GPU is considered usable if we can successfully move a tensor to the device, and this is what this function
```
carmocca, line 84:
```suggestion
    A CUDA device is considered usable if we can successfully move a tensor to the device, and this is what this function
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+    tests for each GPU on the system until the target number of usable GPUs is found.
```
carmocca, line 85:
```suggestion
    tests for each device on the system until the target number of usable devices is found.
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+
+    A subset of GPUs on the system might be used by other processes, and if the GPU is configured to operate in
```
carmocca, line 87:
```suggestion
    A subset of the CUDA devices on the system might be used by other processes, and if the device is configured to operate in
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+    'exclusive' mode (configurable by the admin), then only one process is allowed to occupy it.
+
+    Args:
+        num_devices: The number of GPUs you want to request. By default, this function will return as many as there are
```
carmocca, line 91:
```suggestion
        num_devices: The number of CUDA devices you want to request. By default, this function will return as many as there are
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+            usable GPUs available.
```
carmocca, line 92:
```suggestion
            usable devices available.
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+    Warning:
+        If multiple processes call this function at the same time, there can be race conditions in the case where
+        both processes determine that the device is unoccupied, leading into one of them crashing later on.
     """
-    Returns:
-         A list of all available CUDA GPUs
+    visible_devices = _get_all_visible_cuda_devices()
+    if not visible_devices:
+        raise ValueError(
+            f"You requested to find {num_devices} GPUs but there are no visible CUDA devices on this machine."
```
carmocca, line 101:
```suggestion
            f"You requested to find {num_devices} devices but there are no visible CUDA devices on this machine."
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+        )
+    if num_devices > len(visible_devices):
+        raise ValueError(
+            f"You requested to find {num_devices} GPUs but this machine only has {len(visible_devices)} GPUs."
```
carmocca, line 105:
```suggestion
            f"You requested to find {num_devices} devices but this machine only has {len(visible_devices)} CUDA devices."
```

src/lightning_lite/accelerators/cuda.py
```diff
@@ -78,10 +78,63 @@ def register_accelerators(cls, accelerator_registry: Dict) -> None:
...
+            continue
+
+        available_devices.append(gpu_idx)
+        if len(available_devices) == num_devices:
+            # exit early if we found the right number of GPUs
+            break
+
+    if len(available_devices) != num_devices:
+        raise RuntimeError(
+            f"You requested to find {num_devices} GPUs but only {len(available_devices)} are currently available."
```
carmocca, line 125:
```suggestion
            f"You requested to find {num_devices} CUDA devices but only {len(available_devices)} devices are currently available."
```

carmocca (commented):
A CUDA device is a GPU but a GPU might not be a CUDA device (e.g. MPS)

src/lightning_lite/accelerators/cuda.py
awaelchli: I will add some of these, but from the context it should be very clear that GPU means NVIDIA CUDA. 
GPU is not a forbidden word. We can use it and people are very familiar with it. There is nothing wrong with the current description.

> A CUDA device is a GPU but a GPU might not be a CUDA device (e.g. MPS)

You can take this further, into the absurd:

A CUDA device is a GPU but a GPU might not be a CUDA device (e.g. MPS), unless it is an AMD device then it is again pretending to be a CUDA GPU in PyTorch unless you are on MacOS with an AMD GPU then it is actually the MPS device in PyTorch.

carmocca (approved):

PR #16162, opened 2022-12-21
Avoid using the deprecated `LooseVersion`

carmocca:
## What does this PR do?

Avoids raising deprecation warnings for `distutils.version.LooseVersion`

None

cc @borda

### DIFF
--- a/src/pytorch_lightning/utilities/migration/utils.py
+++ b/src/pytorch_lightning/utilities/migration/utils.py
@@ -14,11 +14,11 @@
 import logging
 import os
 import sys
-from distutils.version import LooseVersion
 from types import ModuleType, TracebackType
 from typing import Any, Dict, List, Optional, Tuple, Type
 
 from lightning_utilities.core.rank_zero import rank_zero_warn
+from packaging.version import Version
 
 import pytorch_lightning as pl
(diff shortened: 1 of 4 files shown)


ethanwharris (approved):

awaelchli (approved):

src/pytorch_lightning/utilities/migration/utils.py
```diff
@@ -37,7 +37,7 @@ def migrate_checkpoint(checkpoint: _CHECKPOINT) -> Tuple[_CHECKPOINT, Dict[str,
         checkpoints and objects that do not support being deep-copied.
     """
     ckpt_version = _get_version(checkpoint)
-    if LooseVersion(ckpt_version) > LooseVersion(pl.__version__):
+    if Version(ckpt_version) > Version(pl.__version__):
```
Borda, line 40: do we know if it works fine with rc|dev|post inversion string?

Borda (approved):

src/pytorch_lightning/utilities/migration/utils.py
carmocca: It should. This is what `module_available` uses too

PR #16164, opened 2022-12-21
Avoid `relpath` bug on Windows

carmocca:
## What does this PR do?

Same fix as in https://github.com/ultralytics/yolov5/pull/7158
Unblocks master

Examples:
- https://github.com/Lightning-AI/lightning/actions/runs/3750245489/jobs/6369703476#step:19:7019
- https://github.com/Lightning-AI/lightning/actions/runs/3749331752/jobs/6367699855#step:16:6978

None

cc @borda

### DIFF
--- a/src/pytorch_lightning/utilities/migration/utils.py
+++ b/src/pytorch_lightning/utilities/migration/utils.py
@@ -21,6 +21,7 @@ from typing import Any, Dict, List, Optional, Tuple, Type
 from lightning_utilities.core.rank_zero import rank_zero_warn
 
 import pytorch_lightning as pl
+from lightning_lite.utilities.imports import _IS_WINDOWS
 from lightning_lite.utilities.types import _PATH
 from lightning_lite.utilities.warnings import PossibleUserWarning
 from pytorch_lightning.utilities.migration.migration import _migration_index
@@ -111,7 +112,10 @@ def _pl_migrate_checkpoint(checkpoint: _CHECKPOINT, checkpoint_path: Optional[_P
 
     # include the full upgrade command, including the path to the loaded file in the error message,
     # so user can copy-paste and run if they want
-    path_hint = os.path.relpath(checkpoint_path, os.getcwd())
+    if not _IS_WINDOWS:  # side-step bug: ValueError: path is on mount 'C:', start on mount 'D:'
+        path_hint = os.path.relpath(checkpoint_path, os.getcwd())
+    else:
+        path_hint = os.path.abspath(checkpoint_path)
     _log.info(
         f"Lightning automatically upgraded your loaded checkpoint from v{old_version} to v{new_version}."
         " To apply the upgrade to your files permanently, run"
--- a/src/pytorch_lightning/utilities/migration/utils.py
+++ b/src/pytorch_lightning/utilities/migration/utils.py
@@ -12,16 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-import os
 import sys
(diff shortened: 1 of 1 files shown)


src/pytorch_lightning/utilities/migration/utils.py
```diff
@@ -111,7 +112,10 @@ def _pl_migrate_checkpoint(checkpoint: _CHECKPOINT, checkpoint_path: Optional[_P
 
     # include the full upgrade command, including the path to the loaded file in the error message,
     # so user can copy-paste and run if they want
-    path_hint = os.path.relpath(checkpoint_path, os.getcwd())
+    if not _IS_WINDOWS:  # side-step bug: ValueError: path is on mount 'C:', start on mount 'D:'
```
awaelchli, line 115: what about `Path.cwd().relative_to(checkpoint_path)`, does it have the same issue?

awaelchli (approved):

Borda (approved):

src/pytorch_lightning/utilities/migration/utils.py
carmocca: Yes:

```python
p = Path('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_logger_reset_correctly_LR0\\.lr_find_c91f5435-ea1d-4fe7-8ad8-4229daec1a4d.ckpt')
cwd = Path('D:\\a\\lightning\\lightning\\tests\\tests_pytorch')
cwd.relative_to(p)
  File "/home/carmocca/git/py310/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3433, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-7-f6f1c7cc80c1>", line 1, in <module>
    cwd.relative_to(p)
  File "/usr/lib/python3.10/pathlib.py", line 818, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
ValueError: 'D:\\a\\lightning\\lightning\\tests\\tests_pytorch' is not in the subpath of 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_logger_reset_correctly_LR0\\.lr_find_c91f5435-ea1d-4fe7-8ad8-4229daec1a4d.ckpt' OR one path is relative and the other is absolute.
```

awaelchli: Oh but if checkpoint_path is an absolute path with a different stem from cwd, then computing the relative path should not be possible for `os.path.relpath`. This is an edge case I didn't consider when writing this function. 

The motivation was to shorten path not to make it too long, so relative to the cwd. But if the checkpoint path and cwd don't have a common stem, then that doesn't make sense. 

We can fix this by checking

```py
if checkpoint_path.is_relative_to(Path.cwd)
    ...
```

carmocca: Is this what you were thinking? [`eb51fe9` (#16164)](https://github.com/Lightning-AI/lightning/pull/16164/commits/eb51fe9fae3db542b8c784aff35a4f9d3b50981a)

carmocca: NVM. `is_relative_to` was introduced in 3.9 so we cannot use it.

awaelchli: Yes exactly, that was the original intention. No need to block the PR for the hotfix, but we might need to revisit.

PR #16187, opened 2022-12-23
Add a `trainer.ckpt_path` setter for stateful loading

carmocca:
## What does this PR do?

```python
trainer = Trainer()
trainer.ckpt_path = "my/checkpoint/file.ckpt"
trainer.fit(model)

trainer.test()  # still uses the `ckpt_path` above

# you will be in charge of resetting this
trainer.ckpt_path = None
trainer.test()  # now uses `ckpt_path="best"` as it would normally
```

This is a tool that NeMo can use in replacement of the `resume_checkpoint_path` attribute, removed in https://github.com/Lightning-AI/lightning/pull/16167: https://github.com/NVIDIA/NeMo/blob/85cc023adc530de59dd13903c4541ca008b260d3/nemo/utils/exp_manager.py#L575

cc @SeanNaren @titu1994

None

## Before submitting

## PR review


src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -391,9 +422,15 @@ def restore_lr_schedulers(self) -> None:
         for config, lrs_state in zip(self.trainer.lr_scheduler_configs, lr_schedulers):
             config.scheduler.load_state_dict(lrs_state)
 
-    # ----------------------------------
-    # PRIVATE OPS
-    # ----------------------------------
+    def _restore_modules_and_callbacks(self, checkpoint_path: Optional[_PATH] = None) -> None:
```
carmocca, line 425: This is just moved off the `trainer.py` file with no changes. Unrelated to the rest of the PR

Borda (approved):

src/pytorch_lightning/trainer/connectors/checkpoint_connector.py
```diff
@@ -73,7 +75,7 @@ def resume_start(self, checkpoint_path: Optional[_PATH] = None) -> None:
         3. from `checkpoint_path` file if provided
         4. don't restore
         """
-        self.resume_checkpoint_path = checkpoint_path
```
awaelchli, line 76: Let's keep this. It communicates that this is the checkpoint path to resume from, not to be confused with the one being saved. It gets managed by `resume_start()` and `resume_end()`

carmocca: This is still there, but renamed to `_ckpt_path`. Are you suggesting to make it public? Or to keep the old name?

I would like that this name matches the one in the trainer, so that would mean having `trainer.resume_checkpoint_path` which then would be inconsistent with `trainer.test(ckpt_path)|

awaelchli: > Are you suggesting to make it public? Or to keep the old name? 

Both. The existing attribute already sort of serves the same purpose, it is why it was public. We made `resume_checkpoint_path` so that it can be changed if needed.

awaelchli (approved):
I don't understand these changes. We were supposed to make things simpler, not more complicated.

PR #16204, opened 2022-12-26
Remove deadlock detection / process reconciliation logic

### LINKED ISSUE #12797: Refactor Strategy Script Launcher and Deadlock handling
## Proposed refactor

Currently, if the training scripts are to be spawned from the strategy itself (as in the strategies/ddp.py):
1. It uses '_SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)'
2. It handles deadlock detection and handling for the launched processes in the strategies/ddp.py file.
We can refactor this logic to allow reuse in other strategies.

### Motivation

This will enable us to share the script launch and deadlock handling logic with the FSDP Native Strategy #12447 or any other strategies in the future.

### Pitch

We can take one of the following approaches
1. Move into parallel strategy (strategies/parallel.py)
2. Move into standalone libraries that are referenced in the strategy

Open to other ideas that come up in the discussion on the issue

### Additional context

awaelchli:
## What does this PR do?

Deadlock detection / process reconciliation is a feature of the DDP strategy family. It is not documented, but it is turned on by default when ddp strategies are used, unless PL_RECONCILIATE_PROCESS=0 is set.

- It has reliability issues, such as that it relies on the current working directory be shared among all nodes, which is often the case but not always. 
- TorchElastic already has mechanisms to detect failing processes and can prevent a hang, and can restart the job automatically. TorchElastic has been upstreamed to PyTorch recently.
- The feature is implemented and tested only for the DDP strategy. Others who inherit from DDP are not tested. Other strategies that inherit from ParallelStrategy do not have this implemented.
- With process reconciliation turned off, an exception that occurs only on one rank but not others will lead to a hang, but the error message surfaces regardless. 
- The main benefit of the feature is that processes get terminated immediately and don't hang. This could save you some money for example when running in cloud instances. 

Part of #16410 
Closes #12797

Yes, for everyone using the ddp strategy, and anyone else that enables `PL_RECONCILIATE_PROCESS=1`, the feature will no longer work. However, errors 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/strategies/ddp.py
+++ b/src/pytorch_lightning/strategies/ddp.py
@@ -12,13 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-import os
-import shutil
-import signal
-import tempfile
-import time
 from datetime import timedelta
-from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
@@ -52,8 +46,7 @@ from pytorch_lightning.strategies.parallel import ParallelStrategy
 from pytorch_lightning.strategies.strategy import TBroadcast
 from pytorch_lightning.trainer.states import TrainerFn
 from pytorch_lightning.utilities.distributed import register_ddp_comm_hook
-from pytorch_lightning.utilities.exceptions import DeadlockDetectedException
-from pytorch_lightning.utilities.rank_zero import rank_zero_info, rank_zero_only, rank_zero_warn
+from pytorch_lightning.utilities.rank_zero import rank_zero_info, rank_zero_only
 from pytorch_lightning.utilities.types import PredictStep, STEP_OUTPUT, TestStep, ValidationStep
 
 if _FAIRSCALE_AVAILABLE:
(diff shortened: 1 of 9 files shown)


carmocca:
I rebased this

src/pytorch_lightning/strategies/ddp.py
```diff
@@ -145,18 +135,12 @@ def _configure_launcher(self) -> None:
         assert self.cluster_environment is not None
         if not self.cluster_environment.creates_processes_externally:
             self._launcher = _SubprocessScriptLauncher(self.cluster_environment, self.num_processes, self.num_nodes)
-            self._rank_0_will_call_children_scripts = True
```
carmocca, line 148: When this ran, `_should_run_deadlock_detection` would return True which means that the deadlock detection logic ran for most people using the subprocessscriptlauncher, not just those who set the environment.

However, it's not clear how useful this was. I just wanted to point out that this was enabled more often than we thought (if I'm not missing something)

carmocca (approved):
Pushed a commit with some leftover removals to the fsdp strategy

src/pytorch_lightning/strategies/ddp.py
awaelchli: I didn't notice that. I think regardless of this, the deadlock detection does not influence how the program runs in the normal case and so it won't be a breaking change. I will update the description of the PR.

carmocca: Yes. To me, it's a question of "how useful was this for DDP users?"

if it wasn't then we can proceed.
otherwise maybe we should just convert this logic into a plugin that strategies opt-in. Instead of through inheritance where they are forced to override or deal with this logic

awaelchli: Created #16410 as a result from our offline chat.

justusschock (approved):

Issue #5558, opened 2021-01-18
Mixed precision: scheduler and optimizer are called in the wrong order


kilianovski:
## 🐛 Bug

When using mixed-precision training, scheduler and optimizer are called in the wrong order. Warning is generated:

```
UserWarning: Detected call of `lr_scheduler.step()` before `optimizer.step()`. In PyTorch 1.1.0 and later, you should call them in the opposite order: `optimizer.step()` before `lr_scheduler.step()`.
```

## Please reproduce using the BoringModel

https://colab.research.google.com/drive/1G7pk6E9XUYq-pS41DXKhqM9Srx8sikiP?usp=sharing

There are four tests. Three of them doesn't raise the warning:

1. test_amp_scheduler(precision=16, configure_optimizers=configure_optimizers_1)
2. test_amp_scheduler(precision=32, configure_optimizers=configure_optimizers_1)
3. test_amp_scheduler(precision=32, configure_optimizers=configure_optimizers_2)

This testcase raises the warning:

1. test_amp_scheduler(precision=16, configure_optimizers=configure_optimizers_2)

### To Reproduce

1. Create model with `configure_optimizers` in a following dictionary style:

```python
def configure_optimizers_2(model):
    optimizer = torch.optim.SGD(model.layer.parameters(), lr=0.1)
    scheduler = {'scheduler':  torch.optim.lr_scheduler.StepLR(optimizer, step_size=1),
              'name': 'learning_rate',
              'interval':'step',
              'frequency': 1}
    
    return {"optimizer": optimizer, "lr_scheduler": scheduler}
```

2. Enable mixed-precision training by setting `precision=16` in a `Trainer`
3. Start training

#### Note

When scheduler is defined in another way, the issue seems to not occur:

```python
def configure_optimizers_1(model):
    optimizer = torch.optim.SGD(model.layer.parameters(), lr=0.1)
    scheduler = torch.optim.lr_scheduler.StepLR(optimizer, step_size=1)
    
    return {"optimizer": optimizer, "lr_scheduler": scheduler}
```

### Expected behavior

No warning

### Environment

* CUDA:
	- GPU:
		- Tesla P100-PCIE-16GB
	- available:         True
	- version:           10.1
* Packages:
	- numpy:             1.19.5
	- pyTorch_debug:     True
	- pyTorch_version:   1.7.0+cu101
	- pytorch-lightning: 1.1.4
	- tqdm:              4.41.1
* System:
[...]

PR #16229, opened 2023-01-03
Fix LR scheduler behaviour with AMP

### LINKED ISSUE #16228: Wrong LR scheduler behaviour when using AMP
### Bug description

When training with native AMP and a LR scheduler, we get a warning that indicates that a LR step has been taken when an optimizer step was skipped (expected at the beginning of the training with native AMP):

This can be fixed by wrapping these lines https://github.com/Lightning-AI/lightning/blob/574a9516012b4ab778254055c537f5d57e8e694f/src/pytorch_lightning/core/module.py#L1589-L1592

in `if hasattr(optimizer, '_step_count') and optimizer._step_count > 0`.

### How to reproduce the bug

```python
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum() * 100000000
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

        opt = torch.optim.SGD(self.layer.parameters(), lr=0.1)
        sched = torch.optim.lr_scheduler.StepLR(opt, 10)
        return {"optimizer": opt, 'lr_scheduler': {"scheduler": sched, "interval": "step"}}

def run():
    train_data = DataLoader(RandomDataset(32, 32), batch_size=2)

    model = BoringModel()
    trainer = Trainer(
        accelerator='gpu',
        devices=1,
        default_root_dir=os.getcwd(),
        num_sanity_val_steps=0,
        max_epochs=1,
        precision=16,
        enable_model_summary=False,
    )


if __name__ == "__main__":
    run()
```

### Error messages and logs

```
/usr/local/lib/python3.8/dist-packages/torch/optim/lr_scheduler.py:138: UserWarning: Detected call of `lr_scheduler.step()` before `optimizer.step()`. In PyTorch 1.1.0 and later, you should call them in the opposite order: `optimizer.step()` before `lr_scheduler.step()`.  Failure to do this will result in PyTorch skipping the first value of the learning rate schedule. See more details at https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate
```

### Environment

_No response_

### More info

_No response_

milesial:
## What does this PR do?

When training when native AMP and a LR scheduler, we get this warning that indicates that a LR step has been taken when an optimizer step was skipped (expected at the beginning of the training with native AMP):
```
/usr/local/lib/python3.8/dist-packages/torch/optim/lr_scheduler.py:138: UserWarning: Detected call of `lr_scheduler.step()` before `optimizer.step()`. In PyTorch 1.1.0 and later, you should call them in the opposite order: `optimizer.step()` before `lr_scheduler.step()`.  Failure to do this will result in PyTorch skipping the first value of the learning rate schedule. See more details at https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate
```

Fixes #16228 #5558

No

## Before submitting

## PR review


src/pytorch_lightning/core/module.py
```diff
@@ -1585,6 +1585,10 @@ def lr_scheduler_step(self, scheduler, optimizer_idx, metric):
                 scheduler.step(epoch=self.current_epoch)
 
         """
+        optimizer = self.trainer.optimizers[optimizer_idx]
+        if hasattr(optimizer, "_step_count") and optimizer._step_count <= 0:
```
carmocca, line 1589: This only works for the first step but the warning can appear after any optimizer step, as there's no easy way to know whether the `optimizer.step` was skipped by AMP.

milesial: This works for the N first "steps" that are skipped. True that we can't detect skipped steps passed the first true step, but PyT doesn't print any warning for those https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py#L138

milesial:
In the process of fixing tests I discovered and fixed a bug where the scheduler wouldn't match its optimizer when multiple optimizers are instantiated with frequencies. Now the optimizers and schedulers match and alternate as they should, resetting the cycle every epoch.

src/pytorch_lightning/core/module.py
akihironitta: > True that we can't easily detect skipped steps passed the first true step, but PyT doesn't print any warning for those

No, but you can detect it by comparing the scale factor before and after the `scaler.step()` as I wrote in the linked issue. IMHO, I wouldn't see the current change as a solution for the reason @carmocca stated.

milesial: Ok, I can change the implementation to a comparison of scales if that's prefered

milesial: Done

milesial:
@carmocca Ready for final review

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -390,7 +391,7 @@ def update_lr_schedulers(self, interval: str, update_plateau_schedulers: bool) -
         if interval == "step" and self._should_accumulate():
             return
         active_optimizers = _get_active_optimizers(
-            self.trainer.optimizers, self.trainer.optimizer_frequencies, self.total_batch_idx
+            self.trainer.optimizers, self.trainer.optimizer_frequencies, self.batch_idx
```
tchaton, line 394: Could you add a test to verify this works properly ?

milesial: I modified the third case of `test_step_scheduling_for_multiple_optimizers_with_frequency` so that it tests that

setup.cfg
```diff
@@ -34,6 +34,7 @@ markers =
     cloud:Run the cloud tests for example
 filterwarnings =
     error::FutureWarning
+    error:Detected call of `lr_scheduler.step\(\)` before `optimizer.step\(\)`:UserWarning
```
carmocca, line 37: I added this line so that our CI fails if this warning appears. This way it tests that your patch works as expected.

carmocca (commented):
Can you check the failing tests?

setup.cfg
milesial: Thanks, but this also makes IPU tests fail, this PR is focused on GPU. Not sure where to fix the issue on IPUs

milesial:
I also modified `native_amp.py` in both lightning_fabric and pytorch_lightning. It doesn't seem like the lightning_fabric one is called for a typical workflow, so I don't know if it's the right way

setup.cfg
akihironitta: @milesial IMHO, we could just skip the IPU test failures as done in my old PR #11755 for now.

Borda (approved):

mrembalski:
Hi @Borda, I also encounter the same issue. Will this be merged?

PR #16251, opened 2023-01-04
Fix app e2e race condition. Add more logs verbosity

carmocca:
## What does this PR do?

Add logs for when the e2e infinite loops fail to stop.

None

cc @borda

### DIFF
--- a/src/lightning_app/testing/testing.py
+++ b/src/lightning_app/testing/testing.py
@@ -389,34 +389,33 @@ def run_app_in_cloud(
             process = Process(target=_print_logs, kwargs={"app_id": app_id})
             process.start()
 
-        # Wait until the app is running
         i = 1
         while True:
-            sleep(1)
             app = _fetch_app_by_name(client, project_id, name)
+            # wait until the app is running
             if app.status.phase == V1LightningappInstanceState.RUNNING:
                 print("App is running, continuing with testing...")
+                view_page = context.new_page()
+                app_url = app.status.url
+                view_page.goto(f"{app_url}/view")
+                j = 1
+                status_code = None
+                while status_code != 200:
+                    status_code = requests.get(app_url + "/openapi.json").status_code
+                    if debug and j % 30 == 0:
+                        print(f"Received status code {status_code} at {app_url!r}, continuing infinite loop...")
+                    j += 1
+                    sleep(1)
                 break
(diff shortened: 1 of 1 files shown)


src/lightning_app/testing/testing.py
```diff
@@ -387,47 +390,41 @@ def run_app_in_cloud(
...
-            ]
-            app = lit_apps[0]
-
+            app = _fetch_app_by_name(client, project_id, name)
             if app.status.phase == V1LightningappInstanceState.RUNNING:
+                print("App is running, continuing with testing...")
+                break
+            if app.status.phase == V1LightningappInstanceState.STOPPED:
+                print("App is stopped, continuing with testing...")
                 break
```
carmocca, line 402: Maybe this should return directly?

src/lightning_app/testing/testing.py
```diff
@@ -366,68 +379,51 @@ def run_app_in_cloud(
...
-            ).lightningapps
-            if app.name == name
-        ]
-        app = lit_apps[0]
-        app_url = app.status.url
-        while True:
-            sleep(1)
-            resp = requests.get(app_url + "/openapi.json")
-            if resp.status_code == 200:
+            if app.status.phase == V1LightningappInstanceState.STOPPED:
```
ethanwharris, line 410: Maybe this should be:
```suggestion
            elif app.status.phase != V1LightningappInstanceState.PENDING:
```
If it isn't pending then it started at some point then either stopped or failed - either way we can break out of the loop

ethanwharris (approved):
Nice!

src/lightning_app/testing/testing.py
```diff
@@ -366,68 +379,51 @@ def run_app_in_cloud(
...
+                # there's a race condition if the app goes from pending to running to stopped before we evaluate the
+                # condition above. avoid it by checking stopped explicitly
+                print("App is stopped, finished testing...")
                 break
```
ethanwharris, line 414: I think view page is already typed as optional so we can just set it to None here - if apps exit that quick then the tests can't be making any checks on the UI
```suggestion
                print("App is stopped, finished testing...")
                view_page = None
                break
```

carmocca: I created a new page before we start the loop. I believe it's okay to the same page instance each iteration

awaelchli (approved):

PR #16123, opened 2022-12-20
Update CI to CUDA 11.7.1

Borda:
## What does this PR do?

we can also consider bumping to 11.7.1

Closes #16193 

## Before submitting

- [ ] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @borda @carmocca @akihironitta

PR #16274, opened 2023-01-06
WIP: Try revert "Update CI to CUDA 11.7.1 (#16123)" 

awaelchli:
This reverts commit 3326e65bb2f7de2040bdf300dca062c627cdddb1 (#16123)

WIP: For debugging, we attempt to revert these changes to determine whether they were the cause of the Horovod failures we see on master.


awaelchli:
The horovod failures show even though we are using the previous pytorchlightning/pytorch_lightning:base-cuda-py3.9-torch1.13-cuda11.6.1 here. So we can rule this out as the source for the issue. Closing. 

cc @carmocca @akihironitta @Borda

PR #16275, opened 2023-01-06
Inform user about `set_to_none` when using DeepSpeed in Lite

awaelchli:
## What does this PR do?

The deepspeed optimizer does not accept the argument `optimizer.zero_grad(set_to_none=...)`. This would create an error, but in Fabric, we can automatically drop it and inform the user.

Should we do this? Perhaps it should be left to the user to resolve any deepspeed errors?

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning_fabric/wrappers.py
+++ b/src/lightning_fabric/wrappers.py
@@ -27,7 +27,7 @@ from lightning_fabric.strategies import Strategy, DeepSpeedStrategy
 from lightning_fabric.utilities import move_data_to_device
 from lightning_fabric.utilities.device_dtype_mixin import _DeviceDtypeModuleMixin
 from lightning_fabric.utilities.types import Optimizable
-from lightning_fabric.utilities.warnings import PossibleUserWarning
+from lightning_fabric.utilities import rank_zero_info
 
 
 T_destination = TypeVar("T_destination", bound=Dict[str, Any])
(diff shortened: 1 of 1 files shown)


Borda (approved):

tchaton (approved):
LGTM !

src/lightning_fabric/wrappers.py
```diff
@@ -68,6 +71,10 @@ def step(self, closure: Optional[Callable] = None) -> Any:
             **kwargs,
         )
 
+    def zero_grad(self, **kwargs: Any) -> None:
```
carmocca, line 74: I agree with the purpose of the PR, but the implementation is different from what we have always done.

Historically, for this change, we would've rerouted all `zero_grad` calls through the strategy interface so that the `DeepSpeedStrategy` can override this hook and do this parsing.

In this case, you've skipped all that by just doing it directly in our optimizer wrapper.

Maybe this argument is more generic than just deepspeed and it makes sense to have it here. Or maybe this will eventually need to be refactored into a strategy.zero_grad().

I'm okay with this because it's a very simple validation and deepspeed themselves should fix the inconsistency, but this might go away eventually. Up to you if you want to do the full strategy change or keep it like this.

carmocca (approved):

src/lightning_fabric/wrappers.py
awaelchli: Yes exactly, the choice here was based on your last argument. The `zero_grad` is such a simple method right now and there is currently no need for the strategy to abstract anything. If this should change in the future, we will add a public method to the strategy. Until then, Fabric and its wrappers can be that thin "frontend" translation layer.

Issue #16285, opened 2023-01-06
Use SIGTERM instead of SIGKILL on DDP Errors


alanhdu:
### Description & Motivation

In DDP training, when a process has an exception then a SIGKILL is issued to all other processes to tear them down:
https://github.com/Lightning-AI/lightning/blob/4c3ce605ad814155e309b6bb0737db6b428a2a0c/src/pytorch_lightning/strategies/ddp.py#L458

Because it's a SIGKILL, everything is immediately torn down and no finalizers an run. This is a problem for us, because we have code setup on the main process to report errors (e.g. log the stack trace to our job tracker, upload any dumped files / logs to S3 for permanent storage + post-hoc debugging) which will not run.

### Pitch

The simplest alternative would be to issue a SIGTERM instead of SIGKILL -- this would let us install a signal handler on the main process to do all of the "finalizers" for us before exiting.

### Alternatives

There are other things we could do, of course:

- We could always do something like send SIGTERM, sleep for some time, check that the process is killed, if process is not killed within timeout send SIGKILL.
- We could use a `SIGINT` (which raises a `KeyboardInterrupt`) -- that way, all of the `try/except`s and `with` statements will naturally run cleanup without having to write a special handler
- We could make a Callback that tries to do things in the `on_exception` handler. This should be mostly possible, but would require more work on our side because right now only the main process establishes connections with our services for error handling + logging

awaelchli:
@alanhdu I think there shouldn't be an issue. I'm wondering, maybe the sigkill was chosen because sigterm was not effective enough to terminate all hanging processes. 
But feel free to send a PR with your proposal if you have interest in contributing it.

Borda:
@alanhdu, what happened with your PR, I somehow can't open it, GH gives me a unicorn... :unicorn:

PR #16307, opened 2023-01-09
Send a SIGINT before sending a SIGKILL in DDP

alanhdu:
## What does this PR do?

This is a proposal for fixing #16285. When shutting down DDP training, we first try sending a SIGINT to the other processes -- this (by default) raises a `KeyboardInterrupt` exception, and since that's an exception all context managers + `try`/`finally` blocks will run as normal for graceful termination. If the processes don't shut down "soon", then we send a SIGKILL later.

Let me know if this is an acceptable approach!

TODO:
- [ ] Update the CHANGELOG


awaelchli (approved):

alanhdu:
Hm... this might be a bit of a dead-end. I tried running the test as an independent script, and the main process just hangs at the following backtrace (obtained via py-spy) 

```
    __futex_abstimed_wait_common (libc.so.6)
    pthread_cond_wait@@GLIBC_2.3.2 (libc.so.6)
    std::condition_variable::wait (libstdc++.so.6.0.30)
    std::condition_variable::wait (libstdc++.so.6.0.30)
    c10::ivalue::Future::wait (libtorch_cpu.so)
    c10d::Reducer::finalize_backward (libtorch_cpu.so)
    std::_Function_handler<void (), c10d::Reducer::mark_variable_ready(unsigned long)::{lambda()#1}>::_M_invoke (libtorch_cpu.so)
    torch::autograd::GraphTask::exec_post_processing (libtorch_cpu.so)
    torch::autograd::GraphTask::mark_as_completed_and_run_post_processing (libtorch_cpu.so)
    torch::autograd::Engine::thread_main (libtorch_cpu.so)
    torch::autograd::Engine::execute_with_graph_task (libtorch_cpu.so)
    torch::autograd::python::PythonEngine::execute_with_graph_task (libtorch_python.so)
    torch::autograd::Engine::execute (libtorch_cpu.so)
    torch::autograd::python::PythonEngine::execute (libtorch_python.so)
    THPEngine_run_backward (libtorch_python.so)
    backward (torch/autograd/__init__.py:197)
    backward (torch/_tensor.py:488)
    backward (pytorch_lightning/core/module.py:1452)
    backward (pytorch_lightning/plugins/precision/precision_plugin.py:67)
    backward (pytorch_lightning/strategies/strategy.py:207)
    _call_strategy_hook (pytorch_lightning/trainer/trainer.py:1485)
    backward_fn (pytorch_lightning/loops/optimization/optimizer_loop.py:305)
    closure (pytorch_lightning/loops/optimization/optimizer_loop.py:144)
    __call__ (pytorch_lightning/loops/optimization/optimizer_loop.py:149)
    _wrap_closure (pytorch_lightning/plugins/precision/precision_plugin.py:105)
    step (torch/optim/sgd.py:130)
    _use_grad (torch/optim/optimizer.py:23)
    wrapper (torch/optim/optimizer.py:140)
    wrapper (torch/optim/lr_scheduler.py:68)
    optimizer_step (pytorch_lightning/plugins/precision/precision_plugin.py:119)
    optimizer_step (pytorch_lightning/strategies/strategy.py:234)
    optimizer_step (pytorch_lightning/strategies/ddp.py:280)
    step (pytorch_lightning/core/optimizer.py:169)
    optimizer_step (pytorch_lightning/core/module.py:1708)
    _call_lightning_module_hook (pytorch_lightning/trainer/trainer.py:1347)
    _optimizer_step (pytorch_lightning/loops/optimization/optimizer_loop.py:370)
    _run_optimization (pytorch_lightning/loops/optimization/optimizer_loop.py:249)
    advance (pytorch_lightning/loops/optimization/optimizer_loop.py:202)
    run (pytorch_lightning/loops/loop.py:199)
    advance (pytorch_lightning/loops/batch/training_batch_loop.py:88)
    run (pytorch_lightning/loops/loop.py:199)
    advance (pytorch_lightning/loops/epoch/training_epoch_loop.py:213)
    run (pytorch_lightning/loops/loop.py:199)
    advance (pytorch_lightning/loops/fit_loop.py:267)
    run (pytorch_lightning/loops/loop.py:199)
    _run_train (pytorch_lightning/trainer/trainer.py:1205)
    _run_stage (pytorch_lightning/trainer/trainer.py:1182)
    _run (pytorch_lightning/trainer/trainer.py:1103)
    _fit_impl (pytorch_lightning/trainer/trainer.py:650)
    launch (pytorch_lightning/strategies/launchers/subprocess_script.py:88)
    _call_and_handle_interrupt (pytorch_lightning/trainer/call.py:36)
    fit (pytorch_lightning/trainer/trainer.py:608)
    <module> (script.py:30)
    __libc_start_call_main (libc.so.6)
    __libc_start_main_impl (libc.so.6)
```

This is weird, because I can manually hit CTRL-C on the script, which *will* successfully raise a KeyboardInterrupt and exit out.

Similarly, I tried sending a `SIGTERM` instead of a `SIGKILL` -- this *does* successfully shut the process down. But as soon as I register a Python signal handler, the process hangs again!

I'm not sure what to make of this behavior -- do you have any ideas what could explain what this?

src/pytorch_lightning/strategies/ddp.py
```diff
@@ -453,9 +453,23 @@ def reconciliate_processes(self, trace: str) -> None:
         if len(os.listdir(sync_dir)) == (self.world_size // self.num_nodes):
             return
 
+        # Shutdown all other processes. First, we try sending a SIGINT since
+        # that triggers a KeyboardInterrupt exception by default, which should
+        # allow any context managers and try/finally logic to run.
         for pid in self._pids:
             if pid != os.getpid():
-                os.kill(pid, signal.SIGKILL)
+                os.kill(pid, signal.SIGINT)
```
awaelchli, line 461: I see you tried to add an end to end test. But this can get really tricky. Think of this:

If your logic is correct and it sends the sigkill correctly, then that would mean that the test process testing this would be killed. But then this is indistinguishable from not running the test at all. 

To test it end-to-end, we would somehow need to run everything from a separate process that waits for the result, then checks whether it's subprocess, the actual test, has stopped and was killed. 

This seems too involved. So I'm suggesting a simpler test that mocks the `os.kill()` call here. We can then verify that the mock was called correctly. Both worker processes would exit the simulation with success. What do you think?

awaelchli:
The code from your https://github.com/Lightning-AI/lightning/pull/16307#issuecomment-1377772171 doesn't hang for me. It exits with sigkill for me (macos).

alanhdu:
> The code from your https://github.com/Lightning-AI/lightning/pull/16307#issuecomment-1377772171 doesn't hang for me. It exits with sigkill for me (macos).

Sorry to clarify:
- The SIGKILL after the `time.sleep` indeed works correctly
- But if you comment that out (or increase the sleep to something large, like 120 second)  then the SIGINT never seems to get delivered. SIGTERMs still get delivered, but never seem to actually reach a custom signal handler if you install it in Python (i.e. the process is torn down if you use the default signal handler, but hangs if you install you're own signal handler).

Borda (approved):

awaelchli:
@alanhdu We removed the process reconciliation logic recently (#16204). We are looking into #16410 as an alternative instead, and your code modification could eventually be included there.

PR #16311, opened 2023-01-10
Lite: Support `self.log` from a LightningModule

lightningforever:
## What does this PR do?

WIP

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/pytorch_lightning/core/module.py
+++ b/src/pytorch_lightning/core/module.py
@@ -403,6 +403,14 @@ class LightningModule(
             rank_zero_only: Whether the value will be logged only on rank 0. This will prevent synchronization which
                 would produce a deadlock as not all processes would perform this log call.
         """
+        if self._fabric is not None and logger is not False:
+            apply_to_collection(
+                value, object, self.__check_allowed, name, value, wrong_dtype=(numbers.Number, Tensor)
+            )
+            # TODO(fabric): Warn if on_epoch, on_step, etc. are set
+            self._fabric.log(name=name, value=value)
+            return
+
         # check for invalid values
         apply_to_collection(value, dict, self.__check_not_nested, name)
         apply_to_collection(
@@ -554,6 +562,12 @@ class LightningModule(
             rank_zero_only: Whether the value will be logged only on rank 0. This will prevent synchronization which
                 would produce a deadlock as not all processes would perform this log call.
         """
+        if self._fabric is not None and logger is not False:
+            # TODO(fabric): Warn if on_epoch, on_step, etc. are set
+            # TODO(fabric): Restrict the type, check that it's not nested
+            self._fabric.log_dict(dictionary)
+            return
+
         for k, v in dictionary.items():
             self.log(
                 name=k,


justusschock (approved):
LGTM in general. One thing: should we warn the user the first time they pass in unsupported arguments? Otherwise, this may silently lead to unexpected behavior

src/pytorch_lightning/core/module.py
```diff
@@ -554,22 +558,47 @@ def log_dict(
...
+
+    def _log_through_fabric(self, name: str, value: _METRIC_COLLECTION, logger: Optional[bool] = None) -> None:
+        if logger is False:
+            # Passing `logger=False` with Fabric does not make much sense because there is no other destination to
+            # log to, but we support it in case the original code was written for Trainer use
+            return
+        apply_to_collection(value, object, self.__check_allowed, name, value, wrong_dtype=(numbers.Number, Tensor))
+
+        assert self._fabric is not None
+        self._fabric.log(name=name, value=value)
```
carmocca, line 589: This could just call `_log_dict_through_fabric` (or the other way around) to reuse the code. Just as you did in the Fabric implementations

src/lightning_fabric/fabric.py
```diff
@@ -608,9 +609,17 @@ def log_dict(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
 
         Args:
             metrics: A dictionary where the key is the name of the metric and the value the scalar to be logged.
+                Any :class:`torch.Tensor`s in the dictionary get detached from the graph automatically.
             step: Optional step number. Most Logger implementations auto-increment this value by one with every
                 log call. You can specify your own value here.
         """
+
+        def to_item(value: Tensor) -> int | float | bool:
```
carmocca, line 617: The new style annotations won't work without the __future__ import. But it doesn't work with libraries that inspect signatures (e.g. jsonarpgarse). Use good old Union

```suggestion
        def to_item(value: Tensor) -> Union[int, float, bool]:
```

src/lightning_fabric/fabric.py
```diff
@@ -608,9 +609,17 @@ def log_dict(self, metrics: Dict[str, Any], step: Optional[int] = None) -> None:
...
+            if value.numel() != 1:
+                raise ValueError("Logging tensors with more than one element is not supported.")
+            return value.item()
```
carmocca, line 620: Should we pull `LightningModule.__to_tensor` to an utility function to use it here? It also detaches and squeezes

awaelchli:
> One thing: should we warn the user the first time they pass in unsupported arguments? Otherwise, this may silently lead to unexpected behavior

I initially wanted to do that. But it would add quite some code to do this correctly and nicely, with added complexity. As a consequence, all types and default values would change in the signature to `Optional[None] = None` etc. I didn't want to go down that path without knowing this is actually worth it.

src/lightning_fabric/fabric.py
awaelchli: I copied this code from the utilities, it is like this there too 😅

src/lightning_fabric/fabric.py
awaelchli: We already have this in yet another utility function called `metrics_to_scalars`. But we (I think) want a different error message.

carmocca: That function's message looks generic enough to me so that it works here. Keep in mind the error would be accompanied by a stacktrace. I was proposing `__to_tensor` because it does a few more things

src/lightning_fabric/fabric.py
carmocca: But you didn't copy the `from __future__ import annotations`. The problem is that it breaks https://github.com/Lightning-AI/lightning/issues/11205. So better not use it.

In `metrics_to_scalars` it's not a problem because it's not parsed by the CLI

src/lightning_fabric/fabric.py
awaelchli: I'm ok with reusing/moving metrics_to_scalars. But pretty sure `__to_tensor` is not suitable, because here in fabric, we want to send the result of `.item()` to the logger. Converting first to a tensor and moving to the device is redundant in this case.

lightningforever: See 789d413

carmocca (approved):

PR #16327, opened 2023-01-10
Fabric: Remove `_Connector.is_distributed`

carmocca:
## What does this PR do?

Resolves TODO in the Fabric connector

None.

### DIFF
--- a/src/lightning_fabric/strategies/parallel.py
+++ b/src/lightning_fabric/strategies/parallel.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from abc import ABC, abstractmethod
+from abc import ABC
 from typing import Any, Dict, List, Optional
 
 import torch
@@ -42,9 +42,8 @@ class ParallelStrategy(Strategy, ABC):
         self.cluster_environment: Optional[ClusterEnvironment] = cluster_environment
 
     @property
-    @abstractmethod
-    def root_device(self) -> torch.device:
-        """Return the root device."""
+    def is_distributed(self) -> bool:
+        return True
 
     @property
     def global_rank(self) -> int:
--- a/src/lightning_fabric/strategies/parallel.py
+++ b/src/lightning_fabric/strategies/parallel.py
@@ -42,10 +42,6 @@ class ParallelStrategy(Strategy, ABC):
         self.cluster_environment: Optional[ClusterEnvironment] = cluster_environment
 
     @property
-    def is_distributed(self) -> bool:
-        return True
-
-    @property
     def global_rank(self) -> int:
         return self.cluster_environment.global_rank() if self.cluster_environment is not None else 0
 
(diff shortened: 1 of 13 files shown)


src/lightning_fabric/strategies/parallel.py
```diff
@@ -42,9 +42,8 @@ def __init__(
         self.cluster_environment: Optional[ClusterEnvironment] = cluster_environment
 
     @property
-    @abstractmethod
-    def root_device(self) -> torch.device:
```
carmocca, line 46: This is already in the parent definition

src/lightning_fabric/strategies/strategy.py
```diff
@@ -62,6 +62,11 @@ def root_device(self) -> torch.device:
     def is_global_zero(self) -> bool:
         """Whether the current process is the rank zero process not only on the local node, but for all nodes."""
 
+    @property
+    @abstractmethod
+    def is_distributed(self) -> bool:
+        """Whether the strategy runs in a distributed environment."""
```
awaelchli, line 68: There is a design where `isinstance(x, ParallelStrategy)` (and perhaps world_size>0) is equivalent with `is_distributed()=True`. In its current implementation, it is mainly needed to determine whether a special sampler is needed. There is also some outdated code in the TPU strategy. I believe long-term, this property would disappear again. 

I would consider keeping the PR as is, but remove it from the base class and do the `hasattr(self.strategy, "is_distributed")` call instead, so we don't have to deprecate it later on.

justusschock: @awaelchli the worldsize check would definitely be required, otherwise DP would falsely be marked as distributed as well (not sure we do support it with fabric today, but it would be a valid usecase). I think that I would like this API to exist, but later just return a check whether the collective is an actual collective or just the no-op one (which is what this ultimately comes to).

justusschock (approved):

src/lightning_fabric/strategies/strategy.py
carmocca: I wouldn't go with `isinstance(x, ParallelStrategy)` because then it forces strategy authors to subclass it to correctly set the sampler.

A world size check would work, but maybe we want to keep `world_size==1` to return True for debugging purposes. For example, testing an interaction with the `DistributedSampler` with `strategy="ddp", devices=1`.

@justusschock Good catch with DataParallel, it needs to return `False` given the current use of `strategy.is_distributed`

awaelchli: > I wouldn't go with isinstance(x, ParallelStrategy) because then it forces strategy authors to subclass it to correctly set the sampler.

Both ways, the developer would need to override and return the correct value. Maybe my point was not clear. I'm saying there is a design where there is no need for an attribute like this, where the distributed sampler creating is completely encapsulated by the strategy. I'm saying that, you can update the PR in a way that would allow us to later switch to this design without needing to deprecate anything.

carmocca: Okay. Actualyl we can use `hasattr(self.strategy, "distributed_sampler_kwargs")` which is already defined for the correct strategies

PR #16353, opened 2023-01-13
Deprecate the FairScale integration

carmocca:
## What does this PR do?

See title.

I haven't added deprecation tests as they are extensively covered in the sharded tests.
All deprecations use the exact same message, so there's no need to check whether it has been shown already. (warnings do this by default)

None

cc @borda @tchaton @awaelchli

### DIFF
--- a/docs/source-pytorch/advanced/model_parallel.rst
+++ b/docs/source-pytorch/advanced/model_parallel.rst
@@ -151,9 +151,9 @@ Here's an example of changing the placement policy to "cpu".
 
 .. _sharded-training:
 
-**************************
+****************
 Sharded Training
-**************************
+****************
 
 The technique can be found within `DeepSpeed ZeRO <https://arxiv.org/abs/1910.02054>`_ and
 `ZeRO-2 <https://www.microsoft.com/en-us/research/blog/zero-2-deepspeed-shattering-barriers-of-deep-learning-speed-scale/>`_,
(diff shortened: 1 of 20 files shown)


src/pytorch_lightning/strategies/sharded.py
```diff
@@ -40,6 +41,15 @@ class DDPShardedStrategy(DDPStrategy):
     strategy_name = "ddp_sharded"
     _REDUCE_BUFFER_SIZE_DEFAULT: int = 2**23  # 8M
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        rank_zero_deprecation(
+            "PyTorch Lightning's sharded implementation using FairScale has been deprecated in v1.9.0 and will be"
+            " removed in v2.0.0. You can try using the `Trainer(strategy='fsdp_native')` instead."
+            " The difference is that native FSDP uses PyTorch's implementation and the current strategy uses"
+            " FairScale's implementation (which was upstreamed to PyTorch). After removal, `strategy='fsdp'` will use"
+            " the native version by default."
```
awaelchli, line 50: Great, this aligns well with Fabric

docs/source-pytorch/advanced/model_parallel.rst
```diff
@@ -171,178 +168,25 @@ these benefits in multi-GPU setups are almost free and throughput scales well wi
...
-            self.block_1 = checkpoint_wrapper(nn.Sequential(nn.Linear(32, 32), nn.ReLU()))
-            self.block_2 = nn.Linear(32, 2)
-
-----
-
-.. _fully-sharded-native-training:
-
 ******************************
-PyTorch Fully Sharded Training
+Fully Sharded Training
```
awaelchli, line 184: needs to adjust the underline for the title

awaelchli (approved):

Borda (approved):

carmocca:
There's an issue with standalone tests hanging, marking as a draft...

carmocca:
The hang is caused by
```python
with pytest.deprecated_call():
    trainer = Trainer()
trainer.fit()
```
when multiple processes are used

because the deprecation message is only shown on rank 0, the other ranks fail the `deprecate_call` and then `fit` hangs after because some processes didn't make it.

We would need a custom `rank_zero_deprecated_call` testing utility to avoid this.

PR #16365, opened 2023-01-15
Disable strict loading in multiprocessing launcher

### LINKED ISSUE #14534: Setting up submodules in `setup` doesn't work correctly with DDP
## 🐛 Bug

When setting up part of the module in `setup` while using DDP, loading the checkpoint here reports unexpected keys for all submodules that are initialized in `setup`.
https://github.com/Lightning-AI/lightning/blob/fafd2546782c8d2c249c9d7546f74b416e803af7/src/pytorch_lightning/strategies/launchers/multiprocessing.py#L151

Similar to https://github.com/Lightning-AI/lightning/issues/5410

### To Reproduce

Made only two changes, marked with `### HERE ###`.

```python
import os

import torch

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()
    
    ### HERE ###
    def setup(self, stage):
        self.foo = torch.nn.Linear(32, 3)

    def forward(self, x):
        return self.layer(x)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

        loss = self(batch).sum()
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        self.log("test_loss", loss)

def run():

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        limit_val_batches=1,
        limit_test_batches=1,
        num_sanity_val_steps=0,
        max_epochs=1,
        enable_model_summary=False,
        ### HERE ###
        devices=2,
    )


if __name__ == "__main__":
    run()
```

### Expected behavior

Should not report any errors.

cc @justusschock @kaushikb11 @awaelchli @akihironitta @rohitgr7

awaelchli:
## What does this PR do?

Fixes #14534

Sets `strict=False` when loading the state dict of the model back into the main process. The model in the main process may have a different architecture than the one trained in the worker processes:

```py

    def __init__(self):
        super().__init__()
        self.layer1 = torch.nn.Linear(32, 2)

    def setup(self, stage=None):
        self.layer2 = torch.nn.Linear(32, 2)  # this layer does not exist in main process


model = BoringModel()  # layer2 does not exist yet
trainer = Trainer(strategy="ddp_spawn", ...)
trainer.fit(model) 
# here, at the end of fit, we load model weights back into main process
# but we can only do so for layer1, because layer2 does not exist in main process

```

This is a limitation of this type of training with the "spawn" method. Since we don't know what the user will do with the model after `fit()`, the best we can do is load the weights that match.

## Before submitting

## PR review

## Did you have fun?


tests/tests_pytorch/strategies/test_ddp_spawn_strategy.py
```diff
@@ -126,30 +124,6 @@ def test_ddp_spawn_configure_ddp(tmpdir):
     trainer.predict(model, dataloaders=model.predict_dataloader())
 
 
-@pytest.mark.parametrize("trainer_fn", [TrainerFn.FITTING, "other"])
-def test_ddp_spawn_transfer_weights(tmpdir, trainer_fn):
```
awaelchli, line 130: test moved to the other file

src/pytorch_lightning/strategies/launchers/multiprocessing.py
```diff
@@ -153,7 +153,7 @@ def _recover_results_in_main_process(self, worker_output: "_WorkerOutput", train
         # load last weights
         if worker_output.weights_path is not None:
             ckpt = self._strategy.checkpoint_io.load_checkpoint(worker_output.weights_path)
-            trainer.lightning_module.load_state_dict(ckpt)
+            trainer.lightning_module.load_state_dict(ckpt, strict=False)
```
carmocca, line 156: (not for this PR). Is there a way we could know which modules were not loaded? so that we tell the user about this. Otherwise it could be confusing if they notice part is not loaded.

src/pytorch_lightning/strategies/launchers/multiprocessing.py
```diff
@@ -153,7 +153,7 @@ def _recover_results_in_main_process(self, worker_output: "_WorkerOutput", train
         # load last weights
         if worker_output.weights_path is not None:
             ckpt = self._strategy.checkpoint_io.load_checkpoint(worker_output.weights_path)
```
carmocca, line 155: Can you add a small comment explaining the "why" of having strict false hardcoded here?

carmocca (approved):

src/pytorch_lightning/strategies/launchers/multiprocessing.py
awaelchli: Yes there is. See the earlier commit where I had this: https://github.com/Lightning-AI/lightning/pull/16365/commits/cefa66e21c6bcced4ed69e0c1b01fecbf966350c

But the issue is with us know knowing whether anything meaningful happens after `trainer.fit()` where the user relies on the full model. If not, this would just be a false positive warning. 


Case 1:
```py
trainer.fit(model)
# program exits
```
All good. No warnings needed.

Case 2:
```py
trainer.fit(model)
trainer.test(model, ckpt_path="best")
# or
trainer.test()  # no model instance given
```
All good. setup() gets called and all weights reloaded for test. No warnings needed.

Case 3:
```py
# or
trainer.test(model)  # use the current weights (no ckpt_path=...)
# when setup() gets called, the model weights would get overwritten
```
A warning may or may not be appropriate. 

Case 4:
```py
trainer.fit(model)

# do something with the model later
model.forward(...)  # some layers are not available, this would error 
```
A warning may or may not be appropriate. The warning would provide additional information as to why the error occurs.

carmocca: I prefer a false positive (we have `PossibleUserWarning` for this) than Case 3 creating a footgun for our users.

src/pytorch_lightning/strategies/launchers/multiprocessing.py
Borda: was this added? :chipmunk:

src/pytorch_lightning/strategies/launchers/multiprocessing.py
justusschock: I tend to agree with @awaelchli here. I think a prominent warning in the docs should be sufficient. Let's not spam warnings with inherent properties of our strategies

justusschock (approved):

src/pytorch_lightning/strategies/launchers/multiprocessing.py
awaelchli: added the requested comment

vitkl:
It is possible to install `pytorch-lightning` with this fix? I don't see any way to do this listed here https://pytorch-lightning.readthedocs.io/en/stable/starter/installation.html

Screenshot 2023-01-23 at 21 54 27

vitkl:
@awaelchli Thank you for quickly implementing this!

It is possible to install `pytorch-lightning` with this fix? I don't see any way to do this listed here https://pytorch-lightning.readthedocs.io/en/stable/starter/installation.html

Screenshot 2023-01-23 at 21 54 27

carmocca:
Try this:

```python
PACKAGE_NAME=pytorch pip install https://github.com/Lightning-AI/lightning/archive/refs/heads/master.zip -U
```

vitkl:
This works, thank you!

vitkl:
Although the solution itself doesn't work - the module ends up with no parameters. Is there any point downstream of this change there the contents of `state_dict` can be ignored or overwritten?

awaelchli:
This is probably because you don't create your layers at the time of instantiation, only later. You will always run into this limitation with the "ddp_spawn" strategy, it's a result of the design. In this case, you should choose `strategy="ddp"` and you will never have this issue.

vitkl:
I see, thanks for explaining! 

And there is no way to modify the strategy - e.g. run Callback setup before loading the parameters in the main process?

awaelchli:
If you want, you can always call the setup() method yourself in the main process:
```py
my_callback.setup("fit") # call so that layers exist after fit
my_model.setup("fit")  # call so that layers exist after fit
trainer.fit(my_model, ...)
```
But before falling back to this workaround, I suggest just using the regular ddp strategy.

vitkl:
Thanks @awaelchli! Both solutions `ddp` and `dpp_notebook + running callback in the main process` seem to work (code runs). However, I see that both solutions lead to identical batches loaded to both devices (the data-loaded tensors such as the pyro plate indices on both devices are identical). I read this https://github.com/Lightning-AI/lightning/discussions/7186 and other related issues - but I don't understand why this should be happening. Is it possible to check if DistributedSampler (`replace_sampler_ddp=True`) was created successfully?

Screenshot 2023-01-29 at 02 31 28

awaelchli:
> Is it possible to check if DistributedSampler (replace_sampler_ddp=True) was created successfully?

You can check `isinstance(trainer.train_dataloaders[0].sampler, DistributedSampler)` for example. 

> It appears that worker_init_fn=pl_worker_init_function leads to all workers being initialised in all processes using the same seed. Is this expected?

Call `seed_everything(1, workers=True)` to seed **dataloader** workers based on global rank.

For further questions, please consider posting in the forum or if you find a bug, a new issue would be appreciated (since this topic here is about strict loading of weights).

PR #16369, opened 2023-01-15
Fix `save_hyperparameters` for multiple inheritance and mixins

### LINKED ISSUE #16206: `save_hyperparameters` and mixins/inheritance
### Bug description

Consider the following piece of code:

```python

class A:
    def __init__(self, bar="abc", **kwargs):
        self.save_hyperparameters()
        super().__init__(**kwargs)

        print("A")
        print(self.hparams)

class B:
    def __init__(self, foo=123, **kwargs):
        self.save_hyperparameters()
        super().__init__(**kwargs)

        print("B")
        print(self.hparams)

class M(A, B, pl.LightningModule):
    pass

m = M()
```

Prior to 1.7.2, this used to print

```
B
"bar": abc
"foo": 123
A
"bar": abc
"foo": 123
```

but starting with 1.7.2 this now prints

```
B

A

```

This change was probably unintentionally caused by #14151.

Given that the `pytorch-lightning` codebase itself heavily uses Mixins, I would expect that this should continue working as before. :laughing: 

### How to reproduce the bug

_No response_

### Error messages and logs

_No response_

### Environment

I have bisected the breaking change to `1.7.2`.
```
pip install 'pytorch-lightning<1.7.2'  # good
pip install 'pytorch-lightning==1.7.2' # bad
```

### More info

_No response_

cc @borda

RuRo:
## What does this PR do?

This PR slightly changes the `save_hyperparameters` logic.

Originally, `save_hyperparameters` captured the arguments from all stack frames. #14151 changed this behaviour. I believe, the intention was to only capture the arguments from the stack frames that correspond to `LightningModule`/`LightningDataModule`/etc children.

This was originally implemented by checking if the current class was a subclass of `LightningModule`/`LightningDataModule`/etc. Unfortunately, this turned out to be too restrictive of a requirement (imho) and it broke multiple inheritance and mixins in particular. (See #16206 and the added tests in `tests/tests_pytorch/models/test_hparams.py`)

This PR slightly relaxes this check. Instead of checking `issubclass(current_class, ...)` we now check `isinstance(current_self, ...)` which correctly accounts for multiple inheritance MRO shenanigans.

Fixes #16206

I modified the `get_init_args` function in `src/pytorch_lightning/utilities/parsing.py` to also return the self argument alongside the `local_args` dict. I am assuming that `get_init_args` is not part of the public API. If it is, then this is a breaking change. Please, let me know if this was a breaking change.

## Before submitting

Not sure, where is the "unreleased section" in the CHANGELOG. I am assuming, that it's currently missing because you are going through the release process for 1.9.0?

## PR review

## Did you have fun?

Yes, mother, I had fun.


awaelchli (commented):
I'm ok with this change since it doesn't contradict #14151 and the existing tests :) 
However, I believe in the future we need to be able to guarantee that calling `save_hyperparameters` leads to a configuration that always can restore the model instance. We can't guarantee it at the moment, because it is allowed to call `save_hyperparameters` multiple times (for example like here in multiple-inheritance) and this can easily break the contract unless the user handles the passing of args to super-inits in particular ways (like in the test).

RuRo:
> However, I believe in the future we need to be able to guarantee that calling `save_hyperparameters` leads to a configuration that always can restore the model instance. We can't guarantee it at the moment, because it is allowed to call `save_hyperparameters` multiple times (for example like here in multiple-inheritance) and this can easily break the contract unless the user handles the passing of args to super-inits in particular ways (like in the test).

To be perfectly honest, I think that the current behaviour in terms of how `save_hyperparameters` and inheritance interact is already fine.

```python
class A(pl.LightningModule):
    def __init__(self, foo, bar):
        print(f"A({foo=}, {bar=})")
        super().__init__()
        self.save_hyperparameters()

class B(A):
    def __init__(self, bam):
        print(f"B({bam=})")
        super().__init__(foo=2, bar=3)
        self.save_hyperparameters()

model = B(bam=1)
print(model.hparams)
model_copy = B(**model.hparams) # Boom!
```

This code is broken. But I don't think that this code is broken **because of** our `save_hyperparameters` implementation. The reason, why it's broken, is that the `__init__` method in `B` doesn't follow the Liskov substitution principle. If `B` inherits from `A`, then you should be able to pass instances of `B` to any place that expects an instance of `A` and it should work. This is not the case for the above classes.

Note, that this doesn't have anything to do with `pytorch_lightning`. This kind of code is broken even in pure python:

```python
class A:
    def __init__(self, foo, bar):
        print(f"A({foo=}, {bar=})")
        super().__init__()

class B(A):
    def __init__(self, bam):
        print(f"B({bam=})")
        super().__init__(foo=2, bar=3)

kwargs_for_A = {"foo": 2, "bar": 3}
model = A(**kwargs_for_A)
model_b = B(bam=1)

# This should work under Liskov substitution:
assert isinstance(model_b, A)
model_specialization = type(model_b)(**kwargs_for_A) # Boom!
```

So in my opinion, we shouldn't aim to support this kind of broken inheritance. IMHO, it should be sufficient to warn the user or maybe raise an Exception, if we are able to detect such situations, but I don't think that we should try to do anything "smart" to try and "support" this kind of broken inheritance.

src/pytorch_lightning/utilities/parsing.py
```diff
@@ -179,8 +180,8 @@ def collect_init_args(
     if not isinstance(frame.f_back, types.FrameType):
         return path_args
 
-    if "__class__" in local_vars and (not classes or issubclass(local_vars["__class__"], classes)):
-        local_args = get_init_args(frame)
+    local_self, local_args = get_init_args(frame)
```
Borda, line 183:
```suggestion
    local_vars, local_args = get_init_args(frame)
```
is this a typo?

RuRo: No, why? `local_self` is the value of `self` (or whatever the first argument is called) in the current frame.

awaelchli (approved):

carmocca (approved):

src/pytorch_lightning/utilities/parsing.py
```diff
@@ -137,10 +137,10 @@ def _get_first_if_any(
     return n_self, n_args, n_kwargs
 
 
-def get_init_args(frame: types.FrameType) -> Dict[str, Any]:
+def get_init_args(frame: types.FrameType) -> Tuple[Optional[Any], Dict[str, Any]]:
```
carmocca, line 140: Can you add a CHANGELOG entry?

carmocca:
> If it is, then this is a breaking change. Please, let me know if this was a breaking change.

This was a breaking change and it broke https://github.com/jdb78/pytorch-forecasting/blob/4ba7a6ec213c395e0903a38a1a6429517dea3464/pytorch_forecasting/models/base_model.py#L260

I'll open an issue adding back compatibility for this

PR #16389, opened 2023-01-17
Remove support for logging multiple metrics together

carmocca:
## What does this PR do?

Removes support for `log("key", {"something": 123})`

If the logger instance supports this, we suggest logging directly into it, instead of going through the `self.log` mechanism.

These logged keys couldn't be monitored by `ModelCheckpoint` or `EarlyStopping`.

The objective of this is to simplify the logging internals. As this is a very niche feature.

Removes the above.

cc @justusschock @awaelchli @carmocca @Blaizzy @borda


awaelchli (approved):

tchaton (approved):
LGTM !

PR #16401, opened 2023-01-17
fix `dirpath` in `log_dir` for CSVLogger

Borda:
## What does this PR do?

replace the rather more robust way of obtaining `log_dir`
see failing legacy testing: https://github.com/Lightning-AI/lightning/actions/runs/3938658456/jobs/6737646117

## Before submitting

- [x] Was this **discussed/approved** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- [ ] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ ] Did you write any **new necessary tests**? (not for typos and docs)
- [ ] Did you verify new and **existing tests pass** locally with your changes?
- [ ] Did you list all the **breaking changes** introduced by this pull request?
- [ ] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

- [x] Is this pull request ready for review? (if not, please submit in draft mode)
- [x] Check that all items from **Before submitting** are resolved
- [x] Make sure the title is self-explanatory and the description concisely explains the PR
- [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?

Make sure you had fun coding 🙃

cc @awaelchli @borda @Blaizzy @tchaton

PR #16405, opened 2023-01-17
Revert #16401 and user proper `CSVLogger`

carmocca:
## What does this PR do?
This reverts commit d46091c.

The true cause was that we were mistakenly importing the CSVLogger from Fabric

None

cc @borda @tchaton

### DIFF
--- a/src/pytorch_lightning/loggers/csv_logs.py
+++ b/src/pytorch_lightning/loggers/csv_logs.py
@@ -39,7 +39,7 @@ class ExperimentWriter(_FabricExperimentWriter):
     r"""
     Experiment writer for CSVLogger.
 
-    Currently supports to log hyperparameters and metrics in YAML and CSV
+    Currently, supports to log hyperparameters and metrics in YAML and CSV
     format, respectively.
 
     Args:
--- a/src/pytorch_lightning/loggers/csv_logs.py
+++ b/src/pytorch_lightning/loggers/csv_logs.py
@@ -39,7 +39,7 @@ class ExperimentWriter(_FabricExperimentWriter):
     r"""
     Experiment writer for CSVLogger.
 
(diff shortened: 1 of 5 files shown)


src/pytorch_lightning/loggers/csv_logs.py
```diff
@@ -39,7 +39,7 @@ class ExperimentWriter(_FabricExperimentWriter):
     r"""
     Experiment writer for CSVLogger.
 
-    Currently, supports to log hyperparameters and metrics in YAML and CSV
+    Currently supports to log hyperparameters and metrics in YAML and CSV
```
carmocca, line 42:
```suggestion
    Currently, supports to log hyperparameters and metrics in YAML and CSV
```

src/pytorch_lightning/trainer/connectors/logger_connector/logger_connector.py
```diff
@@ -18,12 +18,11 @@
 from torch import Tensor
 
 import pytorch_lightning as pl
-from lightning_fabric.loggers import CSVLogger
 from lightning_fabric.loggers.tensorboard import _TENSORBOARD_AVAILABLE, _TENSORBOARDX_AVAILABLE
 from lightning_fabric.plugins.environments import SLURMEnvironment
 from lightning_fabric.utilities import move_data_to_device
 from lightning_fabric.utilities.apply_func import convert_tensors_to_scalars
-from pytorch_lightning.loggers import Logger, TensorBoardLogger
+from pytorch_lightning.loggers import CSVLogger, Logger, TensorBoardLogger
```
carmocca, line 25: This is what was causing the issue. Fabric loggers do not work with PL

awaelchli (approved):
Great, this also addresses my comment on the source PR

Issue #16183, opened 2022-12-22
Hivemind strategy fails


the-beee:
### Bug description

I'm trying to use `HivemindStrategy` to train a ResNet model on Cifar-10 using two machines (one w/ a gpu and the other no).
I start the CPU machine first, and the training starts without a problem. Then, I copy the initial_peers value to the GPU machine, start training but it fails.

### How to reproduce the bug

```python
import os

import pandas as pd
import seaborn as sn
import torch
import torch.nn as nn
import torch.nn.functional as F
import torchvision
from pl_bolts.datamodules import CIFAR10DataModule
from pl_bolts.transforms.dataset_normalizations import cifar10_normalization
from pytorch_lightning import LightningModule, Trainer, seed_everything
from pytorch_lightning.callbacks import LearningRateMonitor
from pytorch_lightning.callbacks.progress import TQDMProgressBar
from pytorch_lightning.loggers import CSVLogger
from torch.optim.lr_scheduler import OneCycleLR
from torch.optim.swa_utils import AveragedModel, update_bn
from torchmetrics.functional import accuracy

seed_everything(7)

PATH_DATASETS = os.environ.get("PATH_DATASETS", ".")
BATCH_SIZE = 256 if torch.cuda.is_available() else 16
NUM_WORKERS = int(os.cpu_count() / 2)


train_transforms = torchvision.transforms.Compose(
    [
        torchvision.transforms.RandomCrop(32, padding=4),
        torchvision.transforms.RandomHorizontalFlip(),
        torchvision.transforms.ToTensor(),
        cifar10_normalization(),
    ]
)

test_transforms = torchvision.transforms.Compose(
    [
        torchvision.transforms.ToTensor(),
        cifar10_normalization(),
    ]
)

cifar10_dm = CIFAR10DataModule(
    data_dir=PATH_DATASETS,
    batch_size=BATCH_SIZE,
    num_workers=NUM_WORKERS,
    train_transforms=train_transforms,
    test_transforms=test_transforms,
    val_transforms=test_transforms,
)

def create_model():
    model = torchvision.models.resnet18(pretrained=False, num_classes=10)
    model.conv1 = nn.Conv2d(3, 64, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1), bias=False)
    model.maxpool = nn.Identity()
    return model


class LitResnet(LightningModule):
    def __init__(self, lr=0.05):
        super().__init__()

        self.save_hyperparameters()
[...]

PR #16407, opened 2023-01-17
Remove the `HivemindStrategy`

carmocca:
## What does this PR do?

Removes the `HivemindStrategy()`

The code for this strategy is conveniently self-contained. We will move the implementation to a separate repository (TBD) which tests support with newer PL versions.

Related #16183 (to be transferred) 

Removes the `HivemindStrategy()`


awaelchli (approved):

justusschock (approved):
I vote for moving it in a separate repository as it is the only thing we have that supports training on heterogeneous clusters of variable sizes (spot instances).

awaelchli:
In addition to @justusschock suggestion, this could also make a great tutorial how to build a custom strategy, integrating a library like this, perhaps in the context of Fabric.

Borda:
> I vote for moving it in a separate repository as it is the only thing we have that supports training on heterogeneous clusters of variable sizes (spot instances).

I think that shall be for the time being preserved in separate repo

lantiga:
I also vote for moving it out (and having it in ecosystem-ci), and second @awaelchli 's suggestion about demonstrating how to maintain a strategy for others that will want/need to do it.

PR #16429, opened 2023-01-18
Loop flattening: remove the default .run() implementation

carmocca:
## What does this PR do?

Reduce base interface as there's no longer a `run` method in the base
Remove the docs.

Further PRs:
- Mark as protected
- Look for opportunities for simplification

Removes support for loop customization


src/pytorch_lightning/loops/optimization/manual_loop.py
```diff
@@ -81,17 +81,16 @@ def __init__(self) -> None:
         # `OptimizationProgress`
         self.optim_step_progress = Progress.from_defaults(ReadyCompletedTracker)
 
-        self._done: bool = False
```
carmocca, line 84: In many loops we had to rely on instance attributes to share information between methods to conform to the base `run` interface.

This is no longer required so we can look for and remove these quirks. This is one example

awaelchli (approved):
Thanks for adding good docstrings to the classes 🎉

Borda (approved):

PR #16434, opened 2023-01-19
Fabric checkpointing 1/n: base implementation

awaelchli:
## What does this PR do?

Adds the base implementation for checkpoint saving and loading, with the new API. Details of how it is used are in the docs pages of this PR. 
Follow up PRs will add the implementation in the DeepSpeed and FSDP strategies.

Part of #14816

```py
# Saving
state = {
    "model1": model1,
    "model2": model2,
    "optimizer": optimizer,
    "iteration": iteration,
}
fabric.save("./checkpoints/version_2/checkpoint.ckpt", state)

# Loading (selective, in-place)
state = {"model1": model1}
fabric.load("./checkpoints/version_2/checkpoint.ckpt", state)

# Loading (all)
all_states = fabric.load("./checkpoints/version_2/checkpoint.ckpt")
```

## Before submitting
- [x] Was this discussed/approved via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your PR does only one thing, instead of bundling different changes together?
- [x] Did you make sure to update the documentation with your changes? (if necessary)
- [x] Did you write any new necessary tests? (not for typos and docs)
- [x] Did you verify new and existing tests pass locally with your changes?
- [x] Did you update the [CHANGELOG](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/CHANGELOG.md)? (not for typos, docs, test updates, or internal minor changes/refactorings)

## PR review
Anyone in the community is free to review the PR once the tests have passed.
Before you start reviewing make sure you have read [Review guidelines](https://github.com/PyTorchLightning/pytorch-lightning/wiki/Review-guidelines). In short, see the following bullet-list:

 - [x] Is this pull request ready for review? (if not, please submit in draft mode)
 - [x] Check that all items from **Before submitting** are resolved
 - [x] Make sure the title is self-explanatory and the description concisely explains the PR
 - [x] Add labels and milestones (and optionally projects) to the PR so it can be classified

## Did you have fun?
I made sure I had fun coding 🙃

cc @borda @awaelchli @carmocca @justusschock

PR #16452, opened 2023-01-20
Fabric checkpointing 2/n: DeepSpeed implementation

awaelchli:
## What does this PR do?

Implement checkpoint saving and loading methods on the DeepSpeedStrategy in Fabric.
Follow up to #16434
Part of https://github.com/Lightning-AI/lightning/issues/14816

## Before submitting

## PR review

## Did you have fun?


justusschock (approved):

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
         self, path: _PATH, state: Dict[str, Union[Module, Optimizer, Any]], storage_options: Optional[Any] = None
     ) -> None:
-        raise NotImplementedError
+        """Save model, optimizer, and other state in a checkpoint directory.
+
+        Args:
+            path: A path to where the files should be saved
+            state: A dictionary with contents to be saved. If the dict contains modules or optimizers, their
+                state-dict will be retrieved and converted automatically.
+            storage_options: Unused by this strategy, since it doesn't use a ``CheckpointIO`` plugin.
```
carmocca, line 375: I was looking at the other implementations. Where do we even use `storage_options`? Why did we add them?

I couldn't find it, maybe we should remove it

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
+
+        # broadcast the path from rank 0 to ensure all the states are saved in a common path
+        path = self.broadcast(path)
+
+        # split the checkpoint into two parts:
+        # 1) the deepspeed engine encapsulating both the model and optionally the optimizer(s)
+        # 2) the rest of the user's state, which in deepspeed is called `client state`
+        excluded_objects = (engine, engine.optimizer)
+        state = {k: v for k, v in state.items() if v not in excluded_objects}
+        # there might be other stateful objects unrelatd to the deepspeed engine - convert them to a state_dict
```
carmocca, line 413:
```suggestion
        # there might be other stateful objects unrelated to the deepspeed engine - convert them to a state_dict
```

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
+        state = self._convert_stateful_objects_in_state(state)
```
carmocca, line 414: I believe deepspeed does this internally:

https://github.com/microsoft/DeepSpeed/blob/867da307d047c8122454ac8b0880cd23d38949e7/deepspeed/runtime/engine.py#L3290-L3311

By passing our own state we run the risk of duplication if the keys are different for the same objects. Maybe add more references to `exclude_objects`? (lr scheduler for example)

still this isn't bullet proof

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
+                When no :class:`deepspeed.DeepSpeedEngine` objects were found in the state, or when multiple
+                :class:`deepspeed.DeepSpeedEngine` objects were found.
+        """
+        if storage_options is not None:
+            raise TypeError(
+                "`DeepSpeedStrategy.save_checkpoint(..., storage_options=...)` is not supported because"
+                " `DeepSpeedStrategy` does not use the `CheckpointIO`."
+            )
+
+        engines = _get_deepspeed_engines_from_state(state)
```
carmocca, line 390: I think this is misleading. The engine is used to save but it doesn't save itself, only it's components: https://github.com/microsoft/DeepSpeed/blob/867da307d047c8122454ac8b0880cd23d38949e7/deepspeed/runtime/engine.py#L3276

So asking to pass the engine through the `state` makes it look like we are going to save it, when in reality it saves what's inside of it. Why not use the engine reference we have in `self` instead?

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
-
-    def load_module_state_dict(self, module: Module, checkpoint: Mapping[str, Any]) -> None:
-        # override to do nothing, deepspeed engine already loaded the weights in `load_checkpoint()`
+        Args:
+            path: A path to where the file is located
+            state: A dictionary of objects whose state will be restored in-place from the checkpoint path.
+                This should contain exactly one model, and the model must already be set up by DeepSpeed.
+
+        Returns:
+            Dictionary with metadata from DeepSpeed
```
carmocca, line 429:
```suggestion
            Dictionary with the state inside DeepSpeed's engine
```

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
+            )
+        engine = engines[0]
+        optimzer_state_requested = bool(len([item for item in state.values() if isinstance(item, Optimizer)]))
+
+        torch.cuda.empty_cache()
+        _, client_state = engine.load_checkpoint(
+            path,
+            tag="checkpoint",
+            load_optimizer_states=optimzer_state_requested,
+            load_lr_scheduler_states=False,
```
carmocca, line 473: Doesn't this need similar logic to the optimizer flag?

src/lightning_fabric/strategies/deepspeed.py
```diff
@@ -365,24 +366,123 @@ def module_sharded_context(self) -> Generator[None, None, None]:
...
+        if client_state is None:
+            raise RuntimeError(
+                "DeepSpeed was unable to load the checkpoint. Ensure you passed in a DeepSpeed compatible checkpoint"
+                " or a single checkpoint file by setting `DeepSpeedStrategy(..., load_full_weights=True)`."
+            )
+        for k, v in client_state.copy().items():
+            if k not in state:
+                continue
+            state[k] = client_state.pop(k)
+        return client_state
```
carmocca, line 485: This could use dictionary difference instead of for-loop and popping

src/lightning_fabric/strategies/deepspeed.py
awaelchli: PyTorch Lightning imports these checkpoint io implementations, and storage options is used there from trainer.save_checkpoint() and passed down. Feature request by Meta

carmocca: Right but none of our CheckpointIO implementations support it. I'd advocate for removal (in a separate PR). Do you know where this was discussed originally?

src/lightning_fabric/strategies/deepspeed.py
carmocca: Offline discussion: We don't want to separate what the user saves from the deepspeed checkpoint, so as an in-between we are going to warn about conflicting keys

src/lightning_fabric/strategies/deepspeed.py
carmocca: Answer: Setup does not support lr_schedulers, so there's nothing to do here.

src/lightning_fabric/strategies/deepspeed.py
carmocca: Answer: My suggestion wouldn't work in examples that use multiple wrapped modules.

docs/source-pytorch/fabric/guide/checkpoint.rst
```diff
@@ -35,7 +35,7 @@ To save the state to the filesystem, pass it to the :meth:`~lightning_fabric.fab
 
     fabric.save("path/to/checkpoint.ckpt", state)
 
-This will unwrap your model and optimizer and convert their `state_dict` automaticall for you.
+This will unwrap your model and optimizer and convert their `state_dict` automatically for you.
```
awaelchli, line 38:
```suggestion
This will unwrap your model and optimizer and convert their `state_dict` automaticall for you.
```
Typo fix moved to #16480 so it doesn't require docs codeowners review for this PR

carmocca (approved):

PR #16469, opened 2023-01-22
Bump max deepspeed version to 0.8.0

awaelchli:
## What does this PR do?

Requires #16458

## Before submitting

## PR review

## Did you have fun?

cc @carmocca @akihironitta @borda @awaelchli @justusschock


justusschock (approved):

carmocca (approved):

Borda (approved):

awaelchli:
CI still installs 0.7.0 😡

Borda:
@awaelchli using `eager` helped :)

awaelchli:
In the PyTorch GPU CI, the package is still 0.7.0
image

Borda:
@awaelchli, as collateral it also updated colossal and all other strategies to the latest in the given range...

akihironitta (approved):

.azure/gpu-tests-pytorch.yml
```diff
@@ -91,7 +91,9 @@ jobs:
         done
       displayName: 'Adjust dependencies'
 
-    - bash: pip install -e .[extra,test,examples] --find-links ${TORCH_URL}
+    - bash: |
+        pip install -e .[extra,test,examples] \
+          -U --upgrade-strategy=eager --find-links ${TORCH_URL}
```
carmocca, line 96: I suggest reverting these "eager" install changes.

If you want to have GPU ci use the latest version in this PR, you will have to push a new cuda docker image to DockerHub so the Azure GPU ci job uses it.

Otherwise yolo it, let it happen overnight, and pray that the upgrade doesn't break anything

awaelchli: The docker image doesn't play a role here I think, since we install the strategies manually inside the docker image (with the command above). The eager argument was added because the pip install command you see here did not upgrade the deepspeed version even with the -U command.

carmocca: My point is that setting `--eager` opens a can of worms that you don't want to get into. ([i tried already](https://github.com/Lightning-AI/lightning/pull/13198)!)

and `-U` won't do anything because it is already installed from the docker image

so docker images do play a role. You need to get the new version in the docker image, which is what I suggest in my previous comment

Borda: which is happening every day... https://github.com/Lightning-AI/lightning/blob/dc298f23405550c46b44cddf04113d4069000a61/.github/workflows/ci-dockers-pytorch.yml#L22

Borda: as you can see in builds: https://github.com/Lightning-AI/lightning/actions/workflows/ci-dockers-pytorch.yml?query=event%3Aschedule

awaelchli:
I pushed and pulled the docker image per your instructions and now Fabric and PL are installing 0.8.0.

carmocca:
Cool! Let's merge this then

PR #16501, opened 2023-01-25
SIGTERM handling is now unrelated to fault tolerance

carmocca:
## What does this PR do?

Fault-tolerance is composed of several items:
1. Automatic DataLoader state serialization
2. Manual DataLoader state serialization
3. Metrics/Logging results state
4. Sigterm handling to signal to terminate gracefully
5. Save a checkpoint when it is signaled to terminate gracefully

This PR changes (4) so it is not tied to the fault-tolerance feature, as it doesn't need to be and could be used for other purposes, or for manual fault-tolerance.

None

### DIFF
--- a/src/pytorch_lightning/loops/epoch/training_epoch_loop.py
+++ b/src/pytorch_lightning/loops/epoch/training_epoch_loop.py
@@ -29,7 +29,7 @@ from pytorch_lightning.trainer.connectors.logger_connector.result import _Result
 from pytorch_lightning.trainer.progress import BatchProgress, SchedulerProgress
 from pytorch_lightning.trainer.supporters import CombinedLoader
 from pytorch_lightning.utilities.auto_restart import _collect_states_on_rank_zero_over_collection
-from pytorch_lightning.utilities.exceptions import ExitGracefullyException, MisconfigurationException
+from pytorch_lightning.utilities.exceptions import MisconfigurationException, SIGTERMException
 from pytorch_lightning.utilities.fetching import AbstractDataFetcher, DataLoaderIterDataFetcher
 from pytorch_lightning.utilities.model_helpers import is_overridden
 from pytorch_lightning.utilities.rank_zero import rank_zero_warn, WarningCache
@@ -293,7 +293,7 @@ class _TrainingEpochLoop(loops._Loop):
         # if training finished, defer exit to the parent. this assumes there will be enough time in between
         # which might not be the case depending on what's in the `*_epoch_end` hooks
         if not self._is_training_done and self.trainer.received_sigterm:
-            raise ExitGracefullyException
+            raise SIGTERMException
 
     def on_run_end(self) -> _OUTPUTS_TYPE:
         outputs, self._outputs = self._outputs, []
--- a/src/pytorch_lightning/loops/epoch/training_epoch_loop.py
+++ b/src/pytorch_lightning/loops/epoch/training_epoch_loop.py
@@ -29,7 +29,7 @@ from pytorch_lightning.trainer.connectors.logger_connector.result import _Result
 from pytorch_lightning.trainer.progress import BatchProgress, SchedulerProgress
 from pytorch_lightning.trainer.supporters import CombinedLoader
 from pytorch_lightning.utilities.auto_restart import _collect_states_on_rank_zero_over_collection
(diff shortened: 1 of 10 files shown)


src/pytorch_lightning/trainer/connectors/signal_connector.py
```diff
@@ -37,17 +37,15 @@ def __call__(self, signum: _SIGNUM, frame: FrameType) -> None:
...
     def register_signal_handlers(self) -> None:
+        self.trainer._received_sigterm = False
         self._original_handlers = self._get_current_signal_handlers()
 
         sigusr_handlers: List[_HANDLER] = []
-        sigterm_handlers: List[_HANDLER] = []
-
-        if _fault_tolerant_training():
-            sigterm_handlers.append(self.fault_tolerant_sigterm_handler_fn)
+        sigterm_handlers: List[_HANDLER] = [self._sigterm_notifier_fn]
```
carmocca, line 48: This has the implication that the SIGTERM signal will be ignored until the loops reach a point that check for it. After that, they will raise the `ExitGracefullyException`

src/pytorch_lightning/trainer/trainer.py
```diff
@@ -1894,6 +1894,14 @@ def sanity_checking(self, val: bool) -> None:
         elif self.sanity_checking:
             self.state.stage = None
 
+    @property
+    def received_sigterm(self) -> bool:
+        """Whether a SIGTERM signal was received (in at least one process)
+
+        For example, this can be checked to exit gracefully.
+        """
+        return self.strategy.reduce_boolean_decision(self._received_sigterm, all=False)
```
awaelchli, line 1903: Should we init this attribute as False in the Trainer's init?
How is it even possible mypy doesn't complain here?

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -292,9 +292,8 @@ def on_advance_end(self) -> None:
 
         # if training finished, defer exit to the parent. this assumes there will be enough time in between
         # which might not be the case depending on what's in the `*_epoch_end` hooks
-        if not self._is_training_done:
-            # if fault tolerant is enabled and process has been notified, exit.
-            self.trainer._exit_gracefully_on_signal()
+        if not self._is_training_done and self.trainer.received_sigterm:
```
awaelchli, line 295: This logic is not equivalent to what we had before. Correct me if I'm wrong, but this looks like it will be (much?) slower than master because we are making an reduce collective call on every training step. This isn't the case in master, because in the previous logic we were testing whether fault tolerant training was enabled through `if _fault_tolerant_training()`.

src/pytorch_lightning/trainer/trainer.py
carmocca: It is defined in the SignalConnector.

mypy does not understand self attributes defined outside the module. This is one issue caused by using connectors.

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: You are correct. The purpose of this PR is to untie this signal handling logic from fault tolerance, so this is expected (not meant to be equivalent).

These checks are run at the epoch end.

I don't have any good suggestion here. Would you suggest that we don't all_reduce?

awaelchli: > These checks are run at the epoch end.

The check runs with the condition `not self._is_training_done` which is True in most training steps. So it's not just at the end of the epoch.

> Would you suggest that we don't all_reduce?

Need to think about this. I want to ask @tchaton. In theory, every process should get the signal regardless of where the process was created. But I'm no sure. We should be able to verify this.

carmocca: Based on this blog: https://the-fonz.gitlab.io/posts/python-multiprocessing/ I think we don't need the reduction as each spawned process will register the signal handler.

tchaton: From what I remember, it was required as SIGTERM won't always be passed or they could be race conditions. I don't think it harms to have it.
I would let you be the judge.

tchaton (approved):
LGTM !

awaelchli (approved):

PR #16520, opened 2023-01-26
Remove memory-retaining epoch-end hooks

carmocca:
## Migration guide

<details>
<summary>
training_epoch_end -> on_train_epoch_end
</summary>

```diff
 class MyLightningModule(L.LightningModule):
+    def __init__(self):
+        super().__init__()
+        self.training_step_outputs = []

     def training_step(self, ...):
         loss = ...
+        self.training_step_outputs.append(loss)
         return loss

-    def training_epoch_end(self, outputs):
-        epoch_average = torch.stack([output["loss"] for output in outputs]).mean()
+    def on_train_epoch_end(self):
+        epoch_average = torch.stack(self.training_step_outputs).mean()
         self.log("training_epoch_average", epoch_average)
+        self.training_step_outputs.clear()  # free memory
```

The same suggestions apply to those implementing `Callback.training_epoch_end`

</details>

<details>
<summary>
validation_epoch_end -> on_validation_epoch_end
</summary>

```diff
 class MyLightningModule(L.LightningModule):
+    def __init__(self):
+        super().__init__()
+        self.validation_step_outputs = []

     def validation_step(self, ...):
         loss = ...
+        self.validation_step_outputs.append(loss)
         return loss

-    def validation_epoch_end(self, outputs):
-        epoch_average = torch.stack(outputs).mean()
+    def on_validation_epoch_end(self):
+        epoch_average = torch.stack(self.validation_step_outputs).mean()
         self.log("validation_epoch_average", epoch_average)
+        self.validation_step_outputs.clear()  # free memory
```

The same suggestions apply to those implementing `Callback.validation_epoch_end`

</details>

<details>
<summary>
test_epoch_end -> on_test_epoch_end
</summary>

```diff
 class MyLightningModule(L.LightningModule):
+    def __init__(self):
+        super().__init__()
+        self.test_step_outputs = []

     def test_step(self, ...):
         loss = ...
+        self.test_step_outputs.append(loss)
         return loss

-    def test_epoch_end(self, outputs):
-        epoch_average = torch.stack(outputs).mean()
+    def on_test_epoch_end(self):
+        epoch_average = torch.stack(self.test_step_outputs).mean()
         self.log("test_epoch_average", epoch_average)
+        self.test_step_outputs.clear()  # free memory
```

The same suggestions apply to those implementing `Callback.test_epoch_end`

</details>

<details>
<summary>
Example with two DataLoaders
</summary>

```diff
 class MyLightningModule(L.LightningModule):
+    def __init__(self):
+        super().__init__()
+        self.test_step_outputs = [[], []]  # two dataloaders

     def test_step(self, batch, batch_idx, dataloader_idx=0):
         loss = ...
+        self.test_step_outputs[dataloader_idx].append(loss)
         return loss

-    def test_epoch_end(self, outputs):
+    def on_test_epoch_end(self):
-        for dl_idx in range(len(outputs)):
+        for dl_idx in range(len(self.test_step_outputs)):
-            dataloader_epoch_average = torch.stack(outputs[dl_idx]).mean()
+            dataloader_epoch_average = torch.stack(self.test_step_outputs[dl_idx]).mean()
             self.log(f"test_epoch_average_dl_{dl_idx}", dataloader_epoch_average)
-            outputs[dl_idx].clear()
+            self.test_step_outputs[dl_idx].clear()

     def test_dataloader(self):
         dl1 = DataLoader(RandomDataset(32, 64), batch_size=2)
         dl2 = DataLoader(RandomDataset(32, 64), batch_size=2)
         return dl1, dl2
```

</details><details>
<summary>
Example with strategy="dp" (DataParallel)
</summary>

```diff
 class MyLightningModule(L.LightningModule):
+    def __init__(self):
+        super().__init__()
+        self.training_step_outputs = []
+        self.validation_step_outputs = []

     def training_step(self, batch, batch_idx):
         output = ...
         return output
 
     def validation_step(self, batch, batch_idx):
         output = ...
         return output

+    def training_step_end(self, training_step_output):
+        training_step_output = self.trainer.strategy.reduce(training_step_output)
+        self.training_step_outputs.append(training_step_output)
+        return training_step_output

+    def validation_step_end(self, validation_step_output):
+        self.validation_step_outputs.append(validation_step_output)
 
-    def training_epoch_end(self, outputs):
-        epoch_average = torch.stack([output["loss"] for output in outputs]).mean()
+    def on_train_epoch_end(self):
+        epoch_average = torch.stack(self.training_step_outputs).mean()
         self.log("training_epoch_average", epoch_average)
+        self.training_step_outputs.clear()  # free memory
 
-    def validation_epoch_end(self, outputs):
+    def on_validation_epoch_end(self):
         epoch_average = torch.stack(self.validation_step_outputs).mean()
         self.log("validation_epoch_average", epoch_average)
+        self.validation_step_outputs.clear()  # free memory
```

</details>

If you have questions about how to migrate your use case, you can ask in this PR.

## What does this PR do?

Removes the `training_epoch_end`, `validation_epoch_end`, and `test_epoch_end` hooks.
In favor of `on_train_epoch_end`, `on_validation_epoch_end`, and `on_test_epoch_end`.

These hooks were becoming problematic as just implementing them could lead to memory issues if the user was unaware of their implementation.
They also increased the loop's complexity and were hard to hack or customize externally.

At runtime, we check whether the old hooks are overridden, and fail if they are with an error message that points to the migration guide above

Blocked by https://github.com/Lightning-AI/lightning/pull/16567

Fixes https://github.com/Lightning-AI/lightning/issues/8731
Closes https://github.com/Lightning-AI/lightning/issues/9380
Closes https://github.com/Lightning-AI/lightning/issues/9968
Closes https://github.com/Lightning-AI/lightning/pull/10878
Closes https://github.com/Lightning-AI/lightning/issues/11554

Follow-up things to address:
https://github.com/Lightning-AI/lightning/issues/8479: need to remove outputs from `on_predict_epoch_end`

Removes the hooks described above.


tests/tests_pytorch/strategies/test_dp.py
```diff
@@ -91,6 +91,12 @@ def test_multi_gpu_model_dp(tmpdir):
 
 
 class ReductionTestModel(BoringModel):
+    def __init__(self):
+        super().__init__()
+        self.train_outputs = []
+        self.val_outputs = []
+        self.tests_outputs = []
```
awaelchli, line 98:
```suggestion
        self.test_outputs = []
```

awaelchli (approved):
Very nice!

tests/tests_pytorch/strategies/test_dp.py
```diff
@@ -123,17 +129,32 @@ def test_step(self, batch, batch_idx):
         self.add_outputs(output, batch.device)
         return output
 
-    def training_epoch_end(self, outputs):
-        assert outputs[0]["loss"].shape == torch.Size([])
-        self._assert_extra_outputs(outputs)
+    def training_step_end(self, training_step_output):
+        # the strategy does this automatically, but since we want to store these in memory, we need to manually do it
+        # so that we can append the reduced value and not the per-rank value
+        training_step_output["loss"] = self.trainer.strategy.reduce(training_step_output["loss"])
```
carmocca, line 135: Storing step-outputs in memory are not as simple with DP

If you do `self.train_outputs.append(output)` in the `training_step`, the outputs are per-rank and not reduced yet

So to replicate the old behaviour, one needs to manually reduce in `training_step_end` (which overrides the automatic reduction by the strategy: https://github.com/Lightning-AI/lightning/blob/fd61ed065a19b89f05b1d751ac1b9106cf68145c/src/lightning/pytorch/strategies/dp.py#L152-L153) so that you have a reference to append.

For evaluation, you still have to override `_step_end` but there's no need to reduce.

src/lightning/pytorch/callbacks/callback.py
```diff
@@ -95,10 +95,29 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
     def on_train_epoch_end(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         """Called when the train epoch ends.
 
-        To access all batch outputs at the end of the epoch, either:
+        To access all batch outputs at the end of the epoch, you can cache step outputs as an attribute of the
```
carmocca, line 98: cc @edenafek for a review

src/lightning/pytorch/trainer/configuration_validator.py
```diff
@@ -98,11 +88,25 @@ def __verify_train_val_loop_configuration(trainer: "pl.Trainer", model: "pl.Ligh
             category=PossibleUserWarning,
         )
 
+    # check legacy hooks are not present
+    if callable(getattr(model, "training_epoch_end", None)):
```
carmocca, line 92: cc @edenafek for a review

docs/source-pytorch/common/lightning_module.rst
```diff
@@ -400,21 +373,30 @@ Validation Epoch-level Metrics
 ==============================
 
 If you need to do something with all the outputs of each :meth:`~pytorch_lightning.core.module.LightningModule.validation_step`,
-override the :meth:`~pytorch_lightning.core.module.LightningModule.validation_epoch_end` method. Note that this method is called before :meth:`~pytorch_lightning.core.module.LightningModule.training_epoch_end`.
+override the :meth:`~pytorch_lightning.core.module.LightningModule.on_validation_epoch_end` method.
```
carmocca, line 376: cc @edenafek for a review

Borda (approved):

justusschock (approved):

src/lightning/pytorch/callbacks/callback.py
```diff
@@ -95,10 +95,29 @@ def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningMo
...
+        .. code-block:: python
+
+            class MyLightningModule(L.LightningModule):
+                def __init__(self):
+                    super().__init__()
+                    self.training_step_outputs = []
+
+                def training_step(self):
+                    loss = ...
+                    selgf.training_step_outputs.append(loss)
```
edenlightning, line 110:
```suggestion
                    self.training_step_outputs.append(loss)
```

src/lightning/pytorch/trainer/configuration_validator.py
edenlightning: Maybe "Support for training_epoch_end has been removed in v2.0.0, use the on_train_epoch_end hook instead. To access outputs, save them in-memory an instance attributes."

edenlightning (approved):

lantiga (approved):
Looks great

fishbotics:
Hi,

I have a question about the migration. I've been doing averaging of some metrics at the end of a validation epoch (nearly identical to the example in the migration description above). I'm training with DDP across 8 devices and I'm wondering whether the averaging in the example above accounts for multiple devices. 

If I understand correctly, the model is instantiated separately for each of the 8 devices. If I append to a metrics list for each model, there would be 8 lists (for each of the 8 models). What is the right way to average the metrics from all 8 devices at the end of the validation epoch?

Thanks a lot!

awaelchli:
Yes correct, the metrics you collect are separete in each process. You can `self.all_gather` your data in the "8 lists" if you want, and then compute the final metric on the data from all processes. This will very likely be inefficient and error prone, so before you do that I recommend checking out [TorchMetrics](https://torchmetrics.readthedocs.io/en/stable/). Chances are that they already have the metric you need implemented and it will work flawlessly and correctly with DDP.

JuanFMontesinos:
Sincerely I find this a step back. PL is supposed to reduce boilerplate code.
Now we need to be aware of how many dataloaders are we planning to use. That's not known until we call trainer.fit, therefore cannot be smoothly defined in `__init__`
Memory cleaning as well.

lantiga:
Thank you @JuanFMontesinos and @fishbotics for the feedback.

@JuanFMontesinos can you provide a minimal working example of what you are describing, so we can reason about it?

The hook required you to be aware of how memory was handled, and it has caused multiple problems for users in the past (e.g. https://github.com/Lightning-AI/lightning/issues/15514). At the same time I’m all for figuring out how to keep things simpler for users.

mwalmsley:
> What is the right way to average the metrics from all 8 devices at the end of the validation epoch?

> I recommend checking out [TorchMetrics](https://torchmetrics.readthedocs.io/en/stable/). Chances are that they already have the metric you need implemented and it will work flawlessly and correctly with DDP.

I am also hoping to aggregate metrics across devices, and currently use TorchMetrics. The TM [docs](https://torchmetrics.readthedocs.io/en/stable/pages/lightning.html) currently show examples using `training_epoch_end` (in the main text) as well as `training_step_end` (specifically for DP, now deprecated so less of an issue). These now no longer work.

For now, I'm switching to `on_train_batch_end` with `self.log(metric, value , on_step=False, on_epoch=True)` instead, and this seems to work well. Is this the intended new method? 

(I have raised a TM [issue](https://github.com/Lightning-AI/metrics/issues/1632) to update their docs)

Borda:
> I am also hoping to aggregate metrics across devices, and currently use TorchMetrics. The TM [docs](https://torchmetrics.readthedocs.io/en/stable/pages/lightning.html) currently show examples using `training_epoch_end` (in the main text) as well as `training_step_end` (specifically for DP, now deprecated so less of an issue). These now no longer work.

@justusschock @SkafteNicki can you please update TM docs/examples

xingzhehe:
I really don't like this either. I am doing GAN, so I need to calculate FID in `validation_epoch_end`. I am also doing some other metrics which require linear regression and svm from processed validation data. Removal of `validation_epoch_end` makes my life much harder.

JuanFMontesinos:
Let me be more proactive. 
Having backwards compatibility whenever possible is a desiderable coding principle . These automatic hooks where a nice feature, where users don't  need to store the desired outputs themselves. They serve for the purpose PL exists: reducing boilerplate and adding functionality. 
I've been diagonally reading the issues that lead to the deprecation of this feature and, if I'm now wrong, the main problem is users don't know how pytorch works. 

It's nice that you try to reach a general public but PL is not an AutoML framework. PL is built on top of Pytorch, and so a user should know how dynamic graphs and memory management work. In the worst case, the OOM issue deserves a warning or some explanations in the docs, as it's done in the Pytorch's docs. 
Instead the decision has been breaking backwards compatibility and a nice feature. Is it something silly? Yes, it takes some decorators to recover exactly the same behaviour. But note that users following the migration guide will still find the same OOM issue. 

Here the damage is already done. But please, note that Pl is used worldwide now. You cannot be breaking backwards compatibility like this if there is no major reason. And it's something that is happening more and more within pytorch ecosystem. 

W.r.t. `num_sanity_val_steps `, N validation steps are run at the beggining (before training). With user-coded hooks, outputs from this sanity check process are stored in the hook the first epoch. Since (I think) on_val_epoch_end is not called afterwards, these two check-values count on the ouputs of the first epoch. Guess this was taken care before. I've tried to run the new lightning on colab but got issues so cannot provide a test.

awaelchli:
@xingzhehe you can definitely still do all of that. You can move your code to the `on_validation_epoch_end` hook. 

Regarding backward compatibility: This is why we waited until 2.0. If you look at the linked issue, it was proposed nearly 2 years ago but we couldn't do it because many users depend on this behavior. With a major version update to 2.0, we found now was the right time to do it. And with the commitment to long-term support of the 1.9.x series, users can still continue to use the existing mechanism without breaking functionality. 

> But note that users following the migration guide will still find the same OOM issue.

The key here is that these users will now be able to debug and identify the issue inside their code immediately. Before, this was difficult or even impossible to do for users digging through the Trainer's implementation. 


Regarding num_val_steps: This feature should not at all impact in any way how the outputs are collected. The sanity check runs the loop the exact same way as it would run when validation runs normally during training, just only N steps. The user should not at all have any code that is conditional on whether the sanity check runs or not, or it defeats the purpose of this feature. If there is an issue with this feature that fails after migrating to 2.0, we should open a bug report and investigate this.

JuanFMontesinos:
@carmocca Yep, you are right. My apologies. I had the feeling `on_validation_epoch_end` wasn't called after the sanity check.

cs-mshah:
The following tutorial doesn't work due to the breaking changes of `validation_epoch_end`: [lightning_examples/text-transformers](https://lightning.ai/docs/pytorch/latest/notebooks/lightning_examples/text-transformers.html).

awaelchli:
@cs-mshah We have almost updated all of these (see https://github.com/Lightning-AI/tutorials/pull/237 for the transformers example). We will include them soon in the latest docs as well. cc @Borda

juliawilkins:
> Hi,
> 
> I have a question about the migration. I've been doing averaging of some metrics at the end of a validation epoch (nearly identical to the example in the migration description above). I'm training with DDP across 8 devices and I'm wondering whether the averaging in the example above accounts for multiple devices.
> 
> If I understand correctly, the model is instantiated separately for each of the 8 devices. If I append to a metrics list for each model, there would be 8 lists (for each of the 8 models). What is the right way to average the metrics from all 8 devices at the end of the validation epoch?
> 
> Thanks a lot!

@fishbotics Following up on this, do you have a simple working example of how to best log with 2.0 + DDP? I'm still a bit confused on the best way to go about this - with maintaining multiple lists (manually?) or does something like `sync_dist=True`do the trick out of the box? Thanks!

TheShadow29:
If I am understanding correctly, to regain previous functionality we need to do:

```
class Module(pl.LightningModule):
   def on_validation_epoch_start(self) -> None:
        super().on_validation_epoch_start()
        self.val_output_list = []
        return
   def validation_step(self, batch):
        ....
        self.val_output_list += results

   def on_validation_epoch_end(self):
       # use val_output_list as needed
```

carmocca:
@juliawilkins What exactly do you want to log? The "keeping outputs in memory" feature was completely separated from `self.log()`. You can log any scalar value. `sync_dist` will reduce the scalar across processes

JackMacs:
Which file should i modify, i just couldn't find the “class Mylightning"

edmcman:
Is there any way to get the old behavior without adding boiler plate?

42elenz:
I am wondering the following:
I Am using multiple GPU - Training so I assumed that I am using some sort of DP. This is the reason I followed the DP-example. This is my implementation:

`class Contrastive_Training_Model(MultimodalBasis):
    def __init__(self, hparams, fold=''):
        super().__init__(hparams)

        self.save_hyperparameters(hparams)
        self.train_criterion_type = hparams.model.clip_pretrain_train_criterion
        self.label_type_fn = hparams.data.label_type_false_negative_class
        self.train_criterion, self.validation_criterion = choose_contrastive_loss_fct(self.train_criterion_type, hparams)
        self.debugging = hparams.logging.debug_level
        self.correct_val_ids_file = hparams.logging.correct_val_ids_file
        self.fold = fold
        self.training_step_outputs = []
        self.validation_step_outputs = []

        #sanity_check(self.train_criterion_type, self.label_type_fn)
    
    #This is called automatically in trainer class
        """
        Trains contrastive model
        """
        train_mri = batch['cor_mri'] #can be courrpted depends on the settings
        train_questionnaire = batch['cor_questionnaire'] #can be courrpted depends on the settings
        mri_data = batch['mri']
        questionnaire_data = batch['questionnaire']
        id = batch['id']
        fn_label_class = batch['fn_label_class']
        ds_label_class = batch['ds_label_class']

        if self.label_type_fn == "binary":
            fn_label_class = fn_label_class.bool()
        mri_embeddings_projected = self.forward_mri(train_mri) 
        questionnaire_embeddings_projected = self.forward_quest(train_questionnaire)
        loss, logits, labels = self.train_criterion(mri_embeddings_projected, questionnaire_embeddings_projected, fn_label_class)
        #self.training_step_outputs.append({'loss':loss, 'logits': logits, 'labels': labels, "ds_label_class": ds_label_class, 'ID': id, 'mri_embeddings': mri_embeddings_projected, 'questionnaire_embeddings': questionnaire_embeddings_projected, 'questionaire_data': questionnaire_data})
        # self.log(f"multimodal.train.loss", loss, on_epoch=True, on_step=False) implemente later
        #if len(im_views[0])==self.hparams.batch_size:
        #self.calc_and_log_train_embedding_acc(logits=logits, labels=labels, modality='multimodal')

        return {'loss':loss,
                'logits': logits, 
                "ds_label_class": ds_label_class,
                'fn_label_class': fn_label_class, 
                'ID': id,
                'mri_embeddings': mri_embeddings_projected,
                'questionnaire_embeddings': questionnaire_embeddings_projected,
                'questionaire_data': questionnaire_data,}
    
    def training_step_end(self, training_step_output):
        training_step_output = self.trainer.strategy.reduce(training_step_output)
        self.training_step_outputs.append(training_step_output)
        return training_step_output
    

        """
        Validates contrastive model
        """
        val_mri = batch['cor_mri']
        val_questionnaire = batch['cor_questionnaire']
        mri_data = batch['mri']
        questionnaire_data = batch['questionnaire']
        id = batch['id']
        fn_label_class = batch['fn_label_class']
        ds_label_class = batch['ds_label_class']

        mri_embeddings_projected = self.forward_mri(val_mri)
        questionnaire_embeddings_projected = self.forward_quest(val_questionnaire)
        loss, logits, quest_logits, labels = self.validation_criterion(mri_embeddings_projected, questionnaire_embeddings_projected, fn_label_class)
        #self.validation_step_outputs.append({'loss':loss, 'mri_logits': logits, 'quest_logits': quest_logits,'logits': logits, "ds_label_class": ds_label_class, 'fn_label_class': fn_label_class, 'ID': id, 'mri_embeddings': mri_embeddings_projected, 'questionnaire_embeddings': questionnaire_embeddings_projected,'questionaire_data': questionnaire_data})
        return {'loss':loss,
                'mri_logits': logits,
                'quest_logits': quest_logits,
                'logits': logits,
                "ds_label_class": ds_label_class,
                'fn_label_class': fn_label_class, 
                'ID': id,
                'mri_embeddings': mri_embeddings_projected,
                'questionnaire_embeddings': questionnaire_embeddings_projected,
                'questionaire_data': questionnaire_data,}

    def validation_step_end(self, validation_step_output):
        self.validation_step_outputs.append(validation_step_output)

    #At the end of the epcoh. All outputs are in a list.
    def on_train_epoch_end(self):
        train_outputs = self.training_step_outputs
        epoch_loss, epoch_accuracy = evaluation_of_contrastive_outputs(train_outputs,self.debugging, evaluation_type="train")
        fold = self.fold
        self.log(f"cont.train.loss{fold}", epoch_loss, on_epoch=True, on_step=False)
        self.log(f"cont.train.acc{fold}", epoch_accuracy, on_epoch=True, on_step=False)
        self.training_step_outputs.clear()

    def on_validation_epoch_end(self):
        val_outputs = self.validation_step_outputs
        fold = self.fold
        epoch_loss, epoch_accuracy, mri_accuracy, questionnaire_accuracy, mri_quest_accuracy = evaluation_of_contrastive_outputs(val_outputs,self.debugging, evaluation_type="validation", correct_val_ids_file=self.correct_val_ids_file)
        self.log(f"cont.val.loss", epoch_loss, on_epoch=True, on_step=False)
        self.log(f"cont.val.accuracy{fold}", epoch_accuracy, on_epoch=True, on_step=False)
        self.log(f"cont.val.mri_accuracy{fold}", mri_accuracy, on_epoch=True, on_step=False)
        self.log(f"cont.val.questionnaire_accuracy{fold}", questionnaire_accuracy, on_epoch=True, on_step=False)
        self.log(f"cont.val.mri_questionnaire_accuracy{fold}", mri_quest_accuracy, on_epoch=True, on_step=False)
        self.validation_step_outputs.clear()
        
        optimizer = torch.optim.Adam(
            self.parameters(), 
            lr=self.hparams.training.lr,
            weight_decay=self.hparams.training.weight_decay)
        return optimizer`
        
Unfourtnatly in the prerrunning, when the basic validation is calculated, the valdiation_step_end() is not called. So I get an error (Division by Zero in the evaluation_of_contrastive_outputs).

What else can I do than just check for this case?

42elenz:
I am wondering the following:
I Am using multiple GPU - Training so I assumed that I am using some sort of DP. This is the reason I followed the DP-example. This is my implementation:

`class Contrastive_Training_Model(MultimodalBasis):
def init(self, hparams, fold=''):
super().init(hparams)

    self.save_hyperparameters(hparams)
    self.train_criterion_type = hparams.model.clip_pretrain_train_criterion
    self.label_type_fn = hparams.data.label_type_false_negative_class
    self.train_criterion, self.validation_criterion = choose_contrastive_loss_fct(self.train_criterion_type, hparams)
    self.debugging = hparams.logging.debug_level
    self.correct_val_ids_file = hparams.logging.correct_val_ids_file
    self.fold = fold
    self.training_step_outputs = []
    self.validation_step_outputs = []

    #sanity_check(self.train_criterion_type, self.label_type_fn)

#This is called automatically in trainer class
    """
    Trains contrastive model
    """
    train_mri = batch['cor_mri'] #can be courrpted depends on the settings
    train_questionnaire = batch['cor_questionnaire'] #can be courrpted depends on the settings
    mri_data = batch['mri']
    questionnaire_data = batch['questionnaire']
    id = batch['id']
    fn_label_class = batch['fn_label_class']
    ds_label_class = batch['ds_label_class']

    if self.label_type_fn == "binary":
        fn_label_class = fn_label_class.bool()
    mri_embeddings_projected = self.forward_mri(train_mri) 
    questionnaire_embeddings_projected = self.forward_quest(train_questionnaire)
    loss, logits, labels = self.train_criterion(mri_embeddings_projected, questionnaire_embeddings_projected, fn_label_class)
    #self.training_step_outputs.append({'loss':loss, 'logits': logits, 'labels': labels, "ds_label_class": ds_label_class, 'ID': id, 'mri_embeddings': mri_embeddings_projected, 'questionnaire_embeddings': questionnaire_embeddings_projected, 'questionaire_data': questionnaire_data})
    # self.log(f"multimodal.train.loss", loss, on_epoch=True, on_step=False) implemente later
    #if len(im_views[0])==self.hparams.batch_size:
    #self.calc_and_log_train_embedding_acc(logits=logits, labels=labels, modality='multimodal')

    return {'loss':loss,
            'logits': logits, 
            "ds_label_class": ds_label_class,
            'fn_label_class': fn_label_class, 
            'ID': id,
            'mri_embeddings': mri_embeddings_projected,
            'questionnaire_embeddings': questionnaire_embeddings_projected,
            'questionaire_data': questionnaire_data,}

def training_step_end(self, training_step_output):
    training_step_output = self.trainer.strategy.reduce(training_step_output)
    self.training_step_outputs.append(training_step_output)
    return training_step_output


    """
    Validates contrastive model
    """
    val_mri = batch['cor_mri']
    val_questionnaire = batch['cor_questionnaire']
    mri_data = batch['mri']
    questionnaire_data = batch['questionnaire']
    id = batch['id']
    fn_label_class = batch['fn_label_class']
    ds_label_class = batch['ds_label_class']

    mri_embeddings_projected = self.forward_mri(val_mri)
    questionnaire_embeddings_projected = self.forward_quest(val_questionnaire)
    loss, logits, quest_logits, labels = self.validation_criterion(mri_embeddings_projected, questionnaire_embeddings_projected, fn_label_class)
    #self.validation_step_outputs.append({'loss':loss, 'mri_logits': logits, 'quest_logits': quest_logits,'logits': logits, "ds_label_class": ds_label_class, 'fn_label_class': fn_label_class, 'ID': id, 'mri_embeddings': mri_embeddings_projected, 'questionnaire_embeddings': questionnaire_embeddings_projected,'questionaire_data': questionnaire_data})
    return {'loss':loss,
            'mri_logits': logits,
            'quest_logits': quest_logits,
            'logits': logits,
            "ds_label_class": ds_label_class,
            'fn_label_class': fn_label_class, 
            'ID': id,
            'mri_embeddings': mri_embeddings_projected,
            'questionnaire_embeddings': questionnaire_embeddings_projected,
            'questionaire_data': questionnaire_data,}

def validation_step_end(self, validation_step_output):
    self.validation_step_outputs.append(validation_step_output)

#At the end of the epcoh. All outputs are in a list.
def on_train_epoch_end(self):
    train_outputs = self.training_step_outputs
    epoch_loss, epoch_accuracy = evaluation_of_contrastive_outputs(train_outputs,self.debugging, evaluation_type="train")
    fold = self.fold
    self.log(f"cont.train.loss{fold}", epoch_loss, on_epoch=True, on_step=False)
    self.log(f"cont.train.acc{fold}", epoch_accuracy, on_epoch=True, on_step=False)
    self.training_step_outputs.clear()

def on_validation_epoch_end(self):
    val_outputs = self.validation_step_outputs
    fold = self.fold
    epoch_loss, epoch_accuracy, mri_accuracy, questionnaire_accuracy, mri_quest_accuracy = evaluation_of_contrastive_outputs(val_outputs,self.debugging, evaluation_type="validation", correct_val_ids_file=self.correct_val_ids_file)
    self.log(f"cont.val.loss", epoch_loss, on_epoch=True, on_step=False)
    self.log(f"cont.val.accuracy{fold}", epoch_accuracy, on_epoch=True, on_step=False)
    self.log(f"cont.val.mri_accuracy{fold}", mri_accuracy, on_epoch=True, on_step=False)
    self.log(f"cont.val.questionnaire_accuracy{fold}", questionnaire_accuracy, on_epoch=True, on_step=False)
    self.log(f"cont.val.mri_questionnaire_accuracy{fold}", mri_quest_accuracy, on_epoch=True, on_step=False)
    self.validation_step_outputs.clear()
    
    optimizer = torch.optim.Adam(
        self.parameters(), 
        lr=self.hparams.training.lr,
        weight_decay=self.hparams.training.weight_decay)
    return optimizer`
Unfourtnatly in the prerrunning, when the basic validation is calculated, the valdiation_step_end() is not called. So I get an error (Division by Zero in the evaluation_of_contrastive_outputs).

What else can I do than just check for this case?

bnestor:
@42elenz I had to add `self.validation_step_outputs.append(validation_step_output)` to my validation_step. Then everything worked fine.

PR #16525, opened 2023-01-27
Cascade SIGTERM to subprocesses

carmocca:
## What does this PR do?

Follow-up to https://github.com/Lightning-AI/lightning/pull/16501

Part of https://github.com/Lightning-AI/lightning/issues/16410

None


carmocca:
@awaelchli I ran a manual test and this is necessary and works as expected.

Should I just add a test with mocks, or add a complete test to https://github.com/Lightning-AI/lightning/blob/c854e4d1e230133b80e92c5dd088f2289219772f/tests/tests_pytorch/run_standalone_tasks.sh?

awaelchli:
@carmocca This looks great. A simple test through unitest.mock would be great!

src/pytorch_lightning/strategies/launchers/launcher.py
```diff
@@ -0,0 +1,23 @@
...
+from abc import ABC, abstractmethod
+
+from lightning_fabric.strategies.launchers.launcher import _Launcher as _FabricLauncher
+from pytorch_lightning.trainer.connectors.signal_connector import _SIGNUM
+
+
+class _Launcher(_FabricLauncher, ABC):
+    @abstractmethod
+    def kill(self, signum: _SIGNUM) -> None:
+        """Kill existing alive processes."""
```
awaelchli, line 23: @carmocca You could also push this to fabric and raise NotImplementedError. To avoid having to define a new base class in PL.

awaelchli (approved):

src/pytorch_lightning/strategies/launchers/launcher.py
carmocca: But do you see us implementing signal handling for fabric in the future? Otherwise I wouldn't do it

awaelchli: I think at the core, fabric would expose these primitives yes. You could call `fabric.strategy.launcher.kill()` for example if that's useful to you. The launcher implementation is also very PL-independent. The only reason why they are subclassed in PL is because of some extra state management and hydra support.

carmocca: I'd rather not do this as it would be unused code in Fabric. In the future, we can do that and add docs for it, or try to merge the classes

carmocca: Resolving. We can do this as a follow-up

PR #16533, opened 2023-01-27
Remove DataLoader serialization (under fault tolerance)

carmocca:
## What does this PR do?

Fault-tolerance is composed of several items:
1. Automatic DataLoader state serialization (this)
2. Manual DataLoader state serialization (this)
3. Metrics/Logging results state (#16516)
4. Sigterm handling to signal to terminate gracefully (#16501)
5. Save a checkpoint when it is signaled to terminate gracefully (#16512)

This PR removes (1) and (2). Parts of this might re-land in the future

This is the last of the fault-tolerance items.

Removes support for serializing the DataLoader state under the `PL_FAULT_TOLERANCE_TRAINING` environment variable

cc @carmocca @akihironitta @borda @justusschock @awaelchli


tests/tests_pytorch/utilities/test_auto_restart.py
```diff
@@ -1061,429 +130,4 @@ def test_auto_restart_under_signal(on_last_batch, val_check_interval, failure_on
...
-    model = TestModel()
-    trainer = Trainer(default_root_dir=tmpdir, max_epochs=1, val_check_interval=val_check_interval)
-    trainer.fit(model, ckpt_path=checkpoint_path)
-    trainer.train_dataloader = None
-    restart_batches = model.batches
-
-    torch.testing.assert_close(total_batches, failed_batches + restart_batches)
-    assert not torch.equal(total_weight, failed_weight)
-    assert torch.equal(total_weight, model.layer.weight)
+    _fit_model(tmpdir, True, val_check_interval, failure_on_step, failure_on_training, on_last_batch, status=status)
```
carmocca, line 133: This tests where in the loops we check if the SIGTERM signal was received to raise an exception. It could be moved to signal connector or a loops test file.

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -306,27 +299,9 @@ def teardown(self) -> None:
...
-            state = loader.state_dict(has_completed=self._has_completed())
-            if state:
-                state_dict["dataloader_state_dict"] = _collect_states_on_rank_zero_over_collection(state)
-
         return state_dict
 
     def on_load_checkpoint(self, state_dict: Dict) -> None:
-        # cache the dataloader state dict until the dataloader objects are available
-        self._dataloader_state_dict = state_dict.get("dataloader_state_dict", {})
         self._batches_that_stepped = state_dict.get("_batches_that_stepped", 0)
```
awaelchli, line 305: @carmocca should we implement a migration that drops these keys? The feature was experimental, we could also just leave the keys in the checkpoint unchanged.

awaelchli (approved):

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
carmocca: What would be the benefit? Loading should work without issues as it is. They would just be unused.

justusschock (approved):

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
awaelchli: Just for your awareness. They would be unused yes. The user could do something with it as well.

akihironitta (approved):
The change in CI looks good to me!

src/pytorch_lightning/utilities/fetching.py
```diff
@@ -39,7 +28,7 @@ def _profile_nothing() -> None:
 
 class AbstractDataFetcher(ABC):
```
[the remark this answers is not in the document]
carmocca: It's still necessary for iterable dataset support

PR #16539, opened 2023-01-28
Make manual optimization mandatory for multiple optimizers

### LINKED ISSUE #12169: lr scheduler doesn't follow optimizer's frequency [in an edge case]
## 🐛 Bug
This is an unlikely case, but `UserWarning: Detected call of lr_scheduler.step() before optimizer.step()` gets raised in the following condition:

- using two optimizers and (at least one) lr schedulers
- AND the first optimizer's frequency is more than or equal to the number of batches in an epoch (so the second optimizer's step is never called)

### To Reproduce
Running the script below with any values that satisfy `LIMIT_TRAIN_BATCHES<=FREQUENCY1`.

```python
import torch

# raises the warning if LIMIT_TRAIN_BATCHES<=FREQUENCY1
LIMIT_TRAIN_BATCHES = 4
FREQUENCY1 = 5
FREQUENCY2 = 3

class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx, optimizer_idx):
        return self(batch).sum()

        optimizer1 = torch.optim.Adam(self.parameters(), lr=0.01)
        optimizer2 = torch.optim.Adam(self.parameters(), lr=0.01)
        scheduler1 = torch.optim.lr_scheduler.StepLR(optimizer1, step_size=5)
        scheduler2 = torch.optim.lr_scheduler.CosineAnnealingLR(
            optimizer2, T_max=2
        )
        lr_scheduler_config_1 = {"scheduler": scheduler1, "interval": "epoch"}
        lr_scheduler_config_2 = {"scheduler": scheduler2, "interval": "epoch"}
        return [
            {
                "optimizer": optimizer1,
                "frequency": FREQUENCY1,
                "lr_scheduler": lr_scheduler_config_1,
            },
            {
                "optimizer": optimizer2,
                "frequency": FREQUENCY2,
                "lr_scheduler": lr_scheduler_config_2,
            },
        ]

    def optimizer_step(
        self,
        epoch,
        batch_idx,
        optimizer,
        optimizer_idx,
        optimizer_closure=None,
        on_tpu: bool = False,
        using_native_amp: bool = False,
        using_lbfgs: bool = False,
    ) -> None:
        print(f"batch{batch_idx} optimizer{optimizer_idx}.step()")
        return super().optimizer_step(
            epoch,
            batch_idx,
            optimizer,
            optimizer_idx,
            optimizer_closure,
            on_tpu,
            using_native_amp,
            using_lbfgs,
        )

    def lr_scheduler_step(self, scheduler, optimizer_idx, metric):
        print(f"       scheduler{optimizer_idx}.step()")
        return super().lr_scheduler_step(scheduler, optimizer_idx, metric)

def main():
    model = BoringModel()
    trainer = Trainer(
        limit_train_batches=LIMIT_TRAIN_BATCHES,
        max_epochs=2,
        enable_progress_bar=False,
        enable_model_summary=False,
        enable_checkpointing=False,
        logger=False,
    )

if __name__ == "__main__":
    main()
```

```shell
batch0 optimizer0.step()
batch1 optimizer0.step()
batch2 optimizer0.step()
batch3 optimizer0.step()
       scheduler0.step()
batch0 optimizer0.step()
batch1 optimizer0.step()
batch2 optimizer0.step()
batch3 optimizer0.step()
       scheduler1.step()  # <- should be `scheduler0.step()`
```

### Expected behavior

`scheduler1.step()` should never be called in the above script.

```shell
batch0 optimizer0.step()
batch1 optimizer0.step()
batch2 optimizer0.step()
batch3 optimizer0.step()
       scheduler0.step()
batch0 optimizer0.step()
batch1 optimizer0.step()
batch2 optimizer0.step()
batch3 optimizer0.step()
       scheduler0.step()
```

### Environment

- PyTorch Version (e.g., 1.10):
- Python version (e.g., 3.9):
- OS (e.g., Linux):
- CUDA/cuDNN version:

### Additional context
Found in #11755 where the following test case throws the warning:
https://github.com/PyTorchLightning/pytorch-lightning/blob/9e7bd9c72d0150940138e7ae8f272afcfe56cd01/tests/trainer/optimization/test_optimizers.py#L273-L282

If you think this is a critical issue, please assign this issue to me, and I'll prioritize this :)

cc @tchaton @rohitgr7

awaelchli:
## What does this PR do?

**WARNING** Estimated time to review: 2 hours + 🤣. 

This PR reworks how optimization with multiple optimizers works.

Closes #12169

### Before 

User returns multiple optimizers from the `configure_optimizers` hook
The training step gets a `optimizer_idx` argument and gets called as many times with the same batch as there are optimizers.

Pro:
- User doesn't have to write the `zero_grad` and `step` calls
- Toggling of optimizers handled automatically

Cons:
- Users struggle to understand how it works under the hood
- Stepping optimizers at different intervals is possible, but cumbersome to configure
- Loop implementation is complex, hard to follow and debug for contributors
- Non-standard optimization requires the user to override several hooks (backward, optimizer_step) etc. which increases the surface for bugs (simplest example is GAN).
- Complex output post-processing for the batch-end and epoch-end hooks required (lists of lists of lists)

### Now

Returning multiple optimizers from `configure_optimizers` hook will raise an error. You need to set `self.automatic_optimization=False` and implement manual optimization in your training step.

Pro:
- Non-standard optimization is more intuitive to implement (simplest example is GAN)
- Stepping optimizers at custom intervals is easy to understand and debug because logic is inlined in training step by user
- Loops become simpler
- No special outputs processing needed for batch-end and epoch-end hooks

Cons:
- User has to call the `zero_grad` and `step` and `manual_backward` manually.
- Gradient accumulation is not supported out of the box for manual optimization. You have to accumulate yourself. For the highest flexibility, Fabric should be considered.

  * Removed `opt_idx` argument from `BaseFinetuning.finetune_function` callback method
  * Removed `opt_idx` argument from `Callback.on_before_optimizer_step` callback method
  * Removed `optimizer_idx` as an optional argument in `LightningModule.training_step`
  * Removed `optimizer_idx` argument from `LightningModule.on_before_optimizer_step`
  * Removed `optimizer_idx` argument from `LightningModule.configure_gradient_clipping`
  * Removed `optimizer_idx` argument from `LightningModule.optimizer_step`
  * Removed `optimizer_idx` argument from `LightningModule.optimizer_zero_grad`
  * Removed `optimizer_idx` argument from `LightningModule.lr_scheduler_step`
  * Removed support for declaring optimizer frequencies in the dictionary returned from `LightningModule.configure_optimizers`
  * Removed arguments `optimizer` and `optimizer_idx` from `LightningModule.backward`
  * Removed `optimizer_idx` argument from `PrecisionPlugin.optimizer_step` and all of its overrides in subclasses
  * Removed `optimizer_idx` argument from `PrecisionPlugin.{optimizer_step,backward}` and all of its overrides in subclasses
  * Removed `optimizer_idx` argument from `Strategy.{optimizer_step,backward}` and all of its overrides in subclasses
  * Removed `Trainer.optimizer_frequencies` attribute
  

Follow ups to this PR:
- Documentation updates
- Demote optimizer loop class from being a loop

## Before submitting

## PR review

## Did you have fun?


src/pytorch_lightning/loops/progress.py
```diff
@@ -259,31 +259,22 @@ class OptimizationProgress(BaseProgress):
 
     Args:
         optimizer: Tracks optimizer progress.
-        optimizer_position: The index of the current optimizer amongst the currently active optimizers.
```
awaelchli, line 262: TODO: migration in checkpoint

tests/tests_pytorch/utilities/migration/test_migration.py
```diff
@@ -65,9 +65,9 @@ def test_migrate_loop_global_step_to_progress_tracking():
     updated_checkpoint, _ = migrate_checkpoint(old_checkpoint, target_version="1.6.0")
```
awaelchli, line 65: TODO: undo this auto-refactor changes. 
Add migration for optimizer_loop state

src/pytorch_lightning/core/module.py
```diff
@@ -1478,16 +1400,15 @@ def toggle_optimizer(self, optimizer: Union[Optimizer, LightningOptimizer], opti
...
-        This is only called automatically when automatic optimization is enabled and multiple optimizers are used.
-
         Args:
-            optimizer_idx: The index of the optimizer to untoggle.
+            optimizer: The optimizer to untoggle.
         """
-        for opt_idx, opt in enumerate(self.trainer.optimizers):
-            if optimizer_idx != opt_idx:
+        for opt in self.trainer.optimizers:
+            # TODO: handle comparison when LightningOptimizer
```
awaelchli, line 1410: TODO

src/pytorch_lightning/core/hooks.py
```diff
@@ -73,6 +73,7 @@ def on_train_batch_start(self, batch: Any, batch_idx: int) -> Optional[int]:
             batch_idx: the index of the batch
         """
 
+    # TODO: Should 'outputs' be renamed to 'output' (singular)?
```
awaelchli, line 76: TODO

tests/tests_pytorch/trainer/optimization/test_optimizers.py
```diff
@@ -49,36 +48,39 @@ def test_optimizer_with_scheduling(tmpdir):
...
-
-    model = TestModel()
-    model.training_epoch_end = None
-    trainer = Trainer(default_root_dir=tmpdir, max_epochs=1, limit_val_batches=0.1, limit_train_batches=0.2)
-    trainer.fit(model)
-    assert trainer.state.finished, f"Training failed with {trainer.state}"
-
-    adjusted_lr1 = [pg["lr"] for pg in trainer.optimizers[0].param_groups]
-    adjusted_lr2 = [pg["lr"] for pg in trainer.optimizers[1].param_groups]
+# TODO: Does it make sense to support returning [optimizer1], [lr_scheduler1, lr_scheduler2]?
```
awaelchli, line 51: TODO

tests/tests_pytorch/trainer/optimization/test_multiple_optimizers.py
```diff
@@ -19,69 +19,27 @@
 from pytorch_lightning.demos.boring_classes import BoringModel
 
 
+# TODO:
```
awaelchli, line 22: TODO

tests/tests_pytorch/loops/test_training_loop_flow_scalar.py
```diff
@@ -328,7 +324,7 @@ def train_dataloader(self):
 
         def on_train_batch_end(self, outputs, batch, batch_idx):
```
awaelchli, line 325: TODO: rename outputs to output (singular)?

src/pytorch_lightning/core/optimizer.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, optimizer: Optimizer):
     def optimizer(self) -> Optimizer:
         return self._optimizer
 
+    # TODO:
```
awaelchli, line 56: TODO

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -232,10 +229,10 @@ def advance(self, data_fetcher: AbstractDataFetcher) -> None:
             with self.trainer.profiler.profile("run_training_batch"):
                 # choose which loop will run the optimization
                 if self.trainer.lightning_module.automatic_optimization:
-                    optimizers = _get_active_optimizers(
-                        self.trainer.optimizers, self.trainer.optimizer_frequencies, kwargs.get("batch_idx", 0)
-                    )
-                    batch_output = self.optimizer_loop.run(optimizers, kwargs)
+                    optimizers = self.trainer.optimizers
+                    # TODO
+                    assert len(optimizers) == 1
```
awaelchli, line 234: TODO

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
```diff
@@ -550,87 +473,9 @@ def _build_kwargs(self, kwargs: OrderedDict, batch: Any, batch_idx: int) -> Orde
         """
         kwargs["batch"] = batch
         training_step_fx = getattr(self.trainer.lightning_module, "training_step")
+        # TODO
         # the `batch_idx` is optional, however, when there's more than 1 argument we cannot differentiate whether the
         # user wants the `batch_idx` or another key like `optimizer_idx` as we are not strict about the argument names
         if is_param_in_hook_signature(training_step_fx, "batch_idx", min_args=2):
             kwargs["batch_idx"] = batch_idx
         return kwargs
```
awaelchli, line 481: TODO

src/pytorch_lightning/loops/fit_loop.py
```diff
@@ -283,11 +282,12 @@ def on_advance_end(self) -> None:
...
-                self._outputs,
-                lightning_module=model,
-                num_optimizers=len(self.trainer.optimizers),
-            )
+            epoch_end_outputs = self._outputs
+            # self.epoch_loop._prepare_outputs_training_epoch_end(
+            #     self._outputs,
+            #     lightning_module=model,
+            #     num_optimizers=len(self.trainer.optimizers),
+            # )
```
awaelchli, line 290: TODO

src/pytorch_lightning/strategies/strategy.py
```diff
@@ -109,6 +108,7 @@ def optimizers(self) -> List[Optimizer]:
     @optimizers.setter
     def optimizers(self, optimizers: List[Optimizer]) -> None:
         self._optimizers = optimizers
+        # TODO
```
awaelchli, line 111: TODO

src/pytorch_lightning/loops/epoch/training_epoch_loop.py
awaelchli: Update comment, keep check (dataloader_idx)

tests/tests_pytorch/trainer/optimization/test_optimizers.py
```diff
@@ -640,6 +493,7 @@ def on_save_checkpoint(self, checkpoint):
     assert model.on_save_checkpoint_called
 
 
+# TODO: Error on scheduler interval config when using multiple optimizers + manual opt
```
awaelchli:
```suggestion
```

tests/tests_pytorch/trainer/optimization/test_optimizers.py
```diff
@@ -736,14 +593,14 @@ def configure_optimizers(self):
         limit_val_batches=0,
     )
 
-    with patch.object(CustomEpochScheduler, "step") as mock_method_epoch, patch.object(
-        torch.optim.lr_scheduler.StepLR, "step"
-    ) as mock_method_step:
-        trainer.fit(model)
+    # with patch.object(CustomEpochScheduler, "step") as mock_method_epoch, patch.object(
+    #     torch.optim.lr_scheduler.StepLR, "step"
+    # ) as mock_method_step:
```
awaelchli: TODO

tests/tests_pytorch/trainer/optimization/test_manual_optimization.py
```diff
@@ -932,28 +932,28 @@ def training_step(self, batch, batch_idx):
...
+            self.toggle_optimizer(optimizer)
 
             loss_g = self.step(batch)
             self.log("loss_g", loss_g, prog_bar=True)
 
             optimizer.zero_grad()
             self.manual_backward(loss_g)
             optimizer.step()
-            self.untoggle_optimizer(optimizer_idx)
+            self.untoggle_optimizer(optimizer)
```
awaelchli, line 956: TODO: split out toggle/untoggle PR

src/pytorch_lightning/core/optimizer.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, optimizer: Optimizer):
...
     @classmethod
     def _to_lightning_optimizer(
         cls, optimizer: Union[Optimizer, "LightningOptimizer"], strategy: "pl.strategies.Strategy", opt_idx: int
```
awaelchli, line 59: Here is also an opt_idx argument. But this is only used internally for bookkeeping (we loop over optimizers and enumerate them to call this method). It would be better to remove this in a follow up, separate from this PR.

src/pytorch_lightning/core/optimizer.py
```diff
@@ -349,27 +343,25 @@ def _validate_scheduler_api(lr_scheduler_configs: List[LRSchedulerConfig], model
...
 
-def _set_scheduler_opt_idx(optimizers: List[Optimizer], lr_scheduler_configs: List[LRSchedulerConfig]) -> None:
-    for config in lr_scheduler_configs:
+def _validate_multiple_optimizers_support(optimizers: List[Optimizer], model: "pl.LightningModule") -> None:
+    if model.automatic_optimization and len(optimizers) > 1:
+        raise RuntimeError(
+            "Training with multiple optimizers is only supported with manual optimization. Set"
+            " `self.automatic_optimization = False`, then access your optimizers in `training_step` with"
+            " `opt1, opt2, ... = self.optimizers()`."
+        )
```
awaelchli, line 352: This is the new error message we raise.

tests/tests_pytorch/trainer/dynamic_args/test_multiple_eval_dataloaders.py
```diff
@@ -104,62 +104,3 @@ def val_dataloader(self):
     )
 
     trainer.fit(model)
-
-
-def test_multiple_optimizers_multiple_dataloaders(tmpdir):
```
awaelchli, line 109: Is this test meaningful? If I would convert this to manual optimization, what exactly would there be to assert?

tests/tests_pytorch/trainer/optimization/test_optimizers.py
```diff
@@ -355,52 +268,6 @@ def configure_optimizers(self):
 
     assert len(trainer.lr_scheduler_configs) == 0
     assert len(trainer.optimizers) == 0
-    assert len(trainer.optimizer_frequencies) == 0
-
-
-def test_multiple_optimizers_callbacks(tmpdir):
-    """Tests that multiple optimizers can be used with callbacks."""
```
awaelchli, line 362: This is kind of pointless. There is no interaction between the two features.

src/pytorch_lightning/callbacks/finetuning.py
```diff
@@ -290,18 +290,13 @@ def _store(
 
     def on_train_epoch_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule") -> None:
         """Called when the epoch begins."""
-        # import is here to avoid circular imports
-        from pytorch_lightning.loops.utilities import _get_active_optimizers
-
-        for opt_idx, optimizer in _get_active_optimizers(trainer.optimizers, trainer.optimizer_frequencies, 0):
+        for opt_idx, optimizer in enumerate(trainer.optimizers):
```
carmocca, line 293: This loop is left for manual optimization? Do we have tests covering manual with this callback?

src/pytorch_lightning/core/module.py
```diff
@@ -682,23 +680,25 @@ def training_step(self, batch, batch_idx):
...
-        If you define multiple optimizers, this step will be called with an additional
-        ``optimizer_idx`` parameter.
+        To use multiple optimizers, you can switch to 'manual optimization' and control their stepping:
 
         .. code-block:: python
 
+            def __init__(self):
+                super().__init__()
+                self.automatic_optimization = False
+
```
carmocca, line 690: nit:
```suggestion
```

src/pytorch_lightning/plugins/precision/precision_plugin.py
```diff
@@ -59,12 +58,11 @@ def backward(  # type: ignore[override]
             tensor: the loss value obtained from the closure
             model: the model to be optimized
             optimizer: current optimizer being used. ``None`` if using manual optimization
-            optimizer_idx: the index of the current optimizer. ``None`` if using manual optimization
             \*args: Positional arguments intended for the actual function that performs the backward, like
                 :meth:`~torch.Tensor.backward`.
             \**kwargs: Keyword arguments for the same purpose as ``*args``.
         """
-        model.backward(tensor, optimizer, optimizer_idx, *args, **kwargs)
+        model.backward(tensor, *args, **kwargs)
```
carmocca, line 65: where did optimizer go?

tests/tests_pytorch/trainer/dynamic_args/test_multiple_eval_dataloaders.py
carmocca: I agree with removing it

carmocca (approved):
LGTM. Huge effort!

src/pytorch_lightning/core/optimizer.py
```diff
@@ -176,21 +176,22 @@ def _init_optimizers_and_lr_schedulers(
...
         _configure_schedulers_automatic_opt(lr_schedulers, monitor)
         if model.automatic_optimization
         else _configure_schedulers_manual_opt(lr_schedulers)
     )
-    _set_scheduler_opt_idx(optimizers, lr_scheduler_configs)
+    _validate_multiple_optimizers_support(optimizers, model)
+    _validate_optimizers_attached(optimizers, lr_scheduler_configs)
     _validate_scheduler_api(lr_scheduler_configs, model)
-    return optimizers, lr_scheduler_configs, optimizer_frequencies
+    return optimizers, lr_scheduler_configs
```
justusschock, line 188: couldn't you still have a single optimizer frequency which would then be equivalent to gradient accumulation?

src/pytorch_lightning/loops/optimization/optimizer_loop.py
```diff
@@ -142,91 +142,26 @@ def __call__(self, *args: Any, **kwargs: Any) -> Optional[Tensor]:
...
-
-    def on_run_end(self) -> _OUTPUTS_TYPE:
-        outputs, self._outputs = self._outputs, {}  # free memory
-        self._indices = tuple()
-        self._optimizers = tuple()
-        return outputs
-
-    def _run_optimization(self, kwargs: OrderedDict, optimizer: torch.optim.Optimizer) -> ClosureResult:
+
+    def run(self, optimizer: Optimizer, kwargs: OrderedDict) -> _OUTPUTS_TYPE:
```
justusschock, line 158: why do we need an entire loop for this if it isn't even looping? Can we either remove the `_Loop` baseclass as it is misleading or move this up in the loop hierarchy and have it as a simple function?

src/pytorch_lightning/loops/progress.py
justusschock: why? having leftover keys should be fine, no? And since 2.0 is breaking , we don't need 2.0 checkpoints to be supported with 1.9.X

justusschock (approved):

src/pytorch_lightning/core/optimizer.py
awaelchli: If we did this, there would be two ways of enabling gradient accumulation. Isn't the Trainer argument more intuitive? It's an interesting idea though.

src/pytorch_lightning/loops/optimization/optimizer_loop.py
awaelchli: Yes, this would be the direct follow up. I thought about doing it in this PR but it would only make it much bigger. There would be three follow up PRs:

- Remove optimizer loop as a loop class
- All documentation updates
- Clean up opt_idx in lightning_optimizer utility functions

src/pytorch_lightning/loops/progress.py
awaelchli: I think it's better to clean things up. 
In a follow up, we will probably collapse the optimizer loop to flatten the hierarchy. In this case, we would extend this migration here to move the optimizer progress tracking state up one level (similar to tbptt). A remapping of the key would anyway be necessary, and then this clean up here would be just an extra.

src/pytorch_lightning/core/optimizer.py
justusschock: I think the trainer argument is more intuitive, however, for me this is an inherent property of the module...

src/pytorch_lightning/plugins/precision/precision_plugin.py
awaelchli: This is the `LightningModule.backward()` call. Now that we don't loop over optimizers anymore, what is the point of feeding the optimizer to this hook? Is it useful? I would drop it.

src/pytorch_lightning/callbacks/finetuning.py
awaelchli: There isn't a test for it with manual optimization, but it should work because all the callback does is freeze and unfreeze parameters, which the user (who implements the finetune function) could do on a per-optimizer param group level.

Borda (approved):
quite long but over all looks good :+1:

src/pytorch_lightning/core/optimizer.py
awaelchli: I thought about this and it actually doesn't work with only a single optimizer. Since we execute the training_step as part of the closure, Not doing the step would mean that the entire training step doesn't get executed. 

Frequency made sense when we had multiple, because the training_step closure would always be executed by either one of the optimizers, and because the user had guarded their code with the optimizer_idx, this would work out.

PR #16552, opened 2023-01-30
store: download model before load

Borda:
## What does this PR do?

the actual store requires a user to call download manually before loading, so we will do it for him... as an improvement, resolving https://github.com/Lightning-AI/lightning/pull/15811#discussion_r1088078197

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/src/lightning/store/cloud_api.py
+++ b/src/lightning/store/cloud_api.py
@@ -333,52 +333,52 @@ def load_model(
             " it's expected that only one of them are requested in a single call."
         )
 
-    if os.path.exists(_LIGHTNING_STORAGE_FILE):
-        version = version or "latest"
-        model_data = _get_model_data(name, version)
-        output_dir = model_data["output_dir"]
-        linked_output_dir = model_data["linked_output_dir"]
-        meta_data = model_data["metadata"]
-        stored = {"code": {}}
-
-        for key, val in meta_data.items():
-            if key.startswith("stored_"):
-                if key.startswith("stored_code_"):
-                    stored["code"][key.split("_code_")[1]] = val
-                else:
-                    stored[key.split("_")[1]] = val
-
-        _validate_output_dir(output_dir)
-        if linked_output_dir:
-            _validate_output_dir(linked_output_dir)
-
-        if load_weights:
-            # This first loads the model - and then the weights
-            if not model:
-                raise ValueError(
-                    "Expected model=... to be passed for loading weights, please pass"
-                    f" your model object to load_from_lightning_cloud({name}, {version}, model=ModelObj)"
-                )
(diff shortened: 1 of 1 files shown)


src/lightning/store/cloud_api.py
```diff
@@ -323,59 +319,57 @@ def load_model(
...
-        else:
-            return _load_model(stored, linked_output_dir or output_dir, *args, **kwargs)
-    else:
-        raise ValueError(
-            f"Could not find the model (for {name}:{version}) in the local system."
-            " Did you make sure to download the model using: `download_model(...)`"
-            " before calling `load_model(...)`?"
-        )
+    if not os.path.exists(_LIGHTNING_STORAGE_FILE):
+        download_model(name=name, version=version, output_dir=_LIGHTNING_STORAGE_DIR, progress_bar=progress_bar)
```
ethanwharris, line 332: This looks weird, we only want to download it if we haven't before right? Does `download_model` handle that? Or can you only download one at a time?

Also, after you do this one that file will exist, but then if you change the model the file still exists no? So probably this only works the first time you do it...

ethanwharris (commented):
Not sure the logic here makes sense

src/lightning/store/cloud_api.py
Borda: > This looks weird, we only want to download it if we haven't before right? Does `download_model` handle that? Or can you only download one at a time?

good point, but I think this shall be handled on the download side...

Borda: Isn't the condition just before solving the issue?
```py
if not os.path.exists(_LIGHTNING_STORAGE_FILE):
```

lantiga: @ethanwharris 's point is that this has only a chance to run once, after which `_LIGHTNING_STORAGE_FILE` will exist.

Note that `_LIGHTNING_STORAGE_FILE` is defined as

```
_LIGHTNING_DIR = os.path.join(Path.home(), ".lightning")
__STORAGE_FILE_NAME = ".model_storage"
_LIGHTNING_STORAGE_FILE = os.path.join(_LIGHTNING_DIR, __STORAGE_FILE_NAME)
```

so it has nothing to do with the actual location where the specific model file has been saved. It only contains metadata, see:

https://github.com/Lightning-AI/lightning/blob/21a7aa2735d1617fc0ed581b318c6a8b5ca09d89/src/lightning/store/cloud_api.py#L284

The actual solution here is to peek into this file, extract the actual location and see if the specific model file exists. Then, if version is `latest`, we need to resolve it and see if it increased with respect to what it's contained in the storage file.

Borda:
not needed in light of major refactor in #18234

PR #16558, opened 2023-01-30
Wraps sharded model for proper access to it `state_dict` in `FSDP` strategy

### LINKED ISSUE #16526: Proper extraction of `state_dict` for `fsdp` strategy
### Outline & Motivation

For now, [`FSDP`](https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/strategies/fsdp.py#L74) uses default implementation from `Strategy` to get model state dict: [`Strategy.lightning_module_state_dict`](https://github.com/Lightning-AI/lightning/blob/master/src/pytorch_lightning/strategies/strategy.py#L450). But, like creating an optimizer, this returns an empty ordered dict. This affects, for example, the ModelCheckpoint callback, there is no model weight in the saved checkpoints.

### Pitch

We need a custom implementation to get model state dict when the user uses the `FSDP` strategy (add the `lightning_module_state_dict` method into the `FSDPStrategy` class). I found 2 workarounds, first one is pretty straightforward and replicates the optimizer initialization behavior:
```python
def lightning_module_state_dict(self) -> Dict[str, Any]:
    """Returns model state."""
    assert self.lightning_module.trainer.model is not None
    wrapped_state_dict = self.lightning_module.trainer.model.state_dict()
    return {k.replace("_forward_module.", ""): v for k, v in wrapped_state_dict}
```

But I'm not sure if this is the correct way to get sharded state dict together. Following documentation from [PyTorch](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.state_dict), we need to use the `state_dict_type` context manager to unshard the model. This implementation should be like this:
```python
def lightning_module_state_dict(self) -> Dict[str, Any]:
    """Returns model state."""
    assert self.lightning_module.trainer.model is not None
    model = self.lightning_module.trainer.model
    full_state_dict_config = FullStateDictConfig(rank0_only=True)
    with FSDP.state_dict_type(model, StateDictType.FULL_STATE_DICT, full_state_dict_config):
        wrapped_state_dict = sharded_module.state_dict()
    return {k.replace("_forward_module.", ""): v for k, v in wrapped_state_dict}
```

### Additional context

1. For both suggested implementations, I remove the prefix `_forward_module` from the state dict. This is due to wrapping the user model into `_LightningModuleWrapperBase`.
2. For the second option there is also an option to specify offloading into the CPU. This is useful when a model is too large to fit on a single GPU.

cc @awaelchli @carmocca

SpirinEgor:
## What does this PR do?

Fixes #16526 by following previously deleted `DDPFullyShardedStrategy`

### Does your PR introduce any breaking changes?

No, it doesn't.


src/pytorch_lightning/strategies/fsdp.py
```diff
@@ -71,6 +71,20 @@
...
 
 
+class _FSDPStrategyModuleWrapper(_LightningModuleWrapperBase):
+    def state_dict(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:  # type: ignore[override]
+        # this is required because with FSDP lightning_module is empty because weights are sharded.
+        # So we need to call self.trainer.model.state_dict (wrapped version) and use this wraper to
+        # avoid extra keys `_forward_module.layer.weight.` since we want `layer.weight.` in state_dict.
+        return self._forward_module.state_dict(*args, **kwargs)
+
+    def named_modules(self, *args: Any, **kwargs: Any) -> Iterator[Tuple[str, Module]]:
```
carmocca, line 81: I wonder if it would be useful to put this in the parent (`_LightningModuleWrapperBase`)

carmocca (commented):
Thanks for looking into this!

This is only an issue in master, not 1.9, correct? In that case, this doesn't need a CHANGELOG entry.

Can you write a test?

src/pytorch_lightning/strategies/fsdp.py
SpirinEgor: Are there any other strategies or another classes that already somehow handles `_forward_module`? If yes, then directly changing `_LightningModuleWrapperBase` would break them.

SpirinEgor:
@carmocca I wrote tests for this. And I notice that the problem is actually occurred only for layers that is not wrapped, e.g. small layers. So, I parametrized tests with different wrapping policies and slightly changed `BoringModel` for correct asserts.

My previous build has one failed test, but it seems not my fault. What should I do?

And do you have any suggestions about modifying `_LightningModuleWrapperBase`?

SpirinEgor:
@carmocca Hi!
What should I do with this PR? Should I cherry-pick it for `1.9.2`? I guess I need your approval for it...

src/lightning/pytorch/strategies/fsdp.py
```diff
@@ -151,6 +165,11 @@ def __init__(
         )
         self.kwargs = kwargs
 
+    def lightning_module_state_dict(self) -> Dict[str, Any]:
+        """Returns model state."""
+        assert self.model is not None
+        return self.model.state_dict()
```
awaelchli, line 171: Let's do this instead:

```py
if isinstance(model, FullyShardedDataParallel):
    self.model.module.state_dict()
```

Instead of adding another wrapping layer.

SpirinEgor: So, you don't want to add `_FSDPStrategyModuleWrapper`? For some reason, your approach doesn't work, there are only shared layers in state dict. For example, if model contains 2 layers with 5 and 10 parameters and we are using `size_based_auto_wrap_policy` with 7 as `min_num_params`, then state dict will only contain second layer.

I guess this is because `FullyShardedDataParallel.state_dict()` aggregate states from all nodes, but `model.module` is a regular torch module without any additional logic.

If you don't want to add additional wrapping, then we can do something like this:
```python
if isinstance(model, FullyShardedDataParallel):
    raw_state_dict = self.model.state_dict()
    state_dict = OrderedDict({k.removeprefix("_forward_module"): v for k, v in raw_state_dict.items()})
```

awaelchli: I would prefer to avoid the extra layer on top, yes. Then you can call self.model.state_dict(). Just a suggestion. But be aware that it will be only a temporary solution. We will want to eventually rework this for implementing checkpointing with local state dict.

SpirinEgor: So, you are okay with just removing this prefix from state dict? `.removeprefix()` is available only in Python 3.9+, is it okay or it is better to write custom function?

I also can implement dumping state dict according to `state_dict_type`, like in torch [documentation](https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel.state_dict_type). We can add argument to `FSDPStrategy` init method and use it in this function.

carmocca: Interesting overrides I saw in the wild: https://github.com/declare-lab/flan-alpaca/blob/97b07b41041acf6606ba253642d07cc7f18eeba2/training.py#L27-L69

SpirinEgor: Yes, I tried to suggest something like this. I'm just curious if we want to support different types of state dicts? If not, and we always need `FULL_STATE`, then the solution is simple and straightforward.

awaelchli: @SpirinEgor Yes feel free to just iterate over the keys and remove the `forward_module.` part in front of all keys. 
I'm also fine with the current wrapper but leaning towards the removal of the prefix. 

See also my comment here: https://github.com/Lightning-AI/lightning/issues/16526#issuecomment-1503541511
This PR takes care of the saving of full-state-dict. We will also need to implement the sharded state eventually and take care of loading.

awaelchli:
There is nothing you need to do. The PR will be merged to master and then cherry-picked to the 1.9.x release branch manually (as we do for all bugfixes).

SpirinEgor:
I rethought this logic with offloading to the CPU. It's not good to reuse this variable as it's intended for a completely different purpose. We need to figure out why this doesn't work on CI. Because it works on my setup (2xA100).

awaelchli:
I'm looking into it!

awaelchli (approved):

Borda (approved):

PR #16585, opened 2023-01-31
Sized iterable typing improvements

carmocca:
## What does this PR do?

Part of https://github.com/Lightning-AI/lightning/issues/10696

None


src/pytorch_lightning/utilities/data.py
```diff
@@ -94,41 +87,39 @@ def extract_batch_size(batch: BType) -> int:
...
 def has_len_all_ranks(
-    dataloader: Union[DataLoader, CombinedLoader],
+    dataloader: object,
     strategy: "pl.strategies.Strategy",
     model: Union["pl.LightningModule", "pl.LightningDataModule"],
-) -> bool:
-    """Checks if a given Dataloader has ``__len__`` method implemented i.e. if it is a finite dataloader or
-    infinite dataloader."""
-    try:
-        local_length = len(dataloader)  # type: ignore [arg-type] # we are checking with duck-typing
```
carmocca, line 104: I believe this solves a deadlock that could happen if one rank does not define `__len__`. That rank would jump to the `except` block while the others would try to `reduce`

At least now it raises an exception in that case

awaelchli: @carmocca If len is not defined on one of the ranks, it would mean we would be running with different source code. But that's not possible and violates the assumptions of how a program is run.

carmocca: This is cherry-picked, but the following is possible as this is a public function

```python
thing = SizedThing()
if rank == 1:
   thing = UnsizedThing()
has_len_all_ranks(thing, ...)
```

In any case, this was a minor observation. As you suggest, it shouldn't happen within our own usage.

src/lightning_fabric/utilities/data.py
```diff
@@ -42,31 +43,34 @@ def __call__(self, *args: Any) -> None:
...
-def has_iterable_dataset(dataloader: DataLoader) -> bool:
-    return hasattr(dataloader, "dataset") and isinstance(dataloader.dataset, IterableDataset)
+def has_iterable_dataset(dataloader: object) -> TypeGuard[DataLoader]:
+    return isinstance(dataloader, DataLoader) and isinstance(dataloader.dataset, IterableDataset)
 
 
-def has_len(dataloader: Union[DataLoader, Iterable, Dataset]) -> bool:
-    """Checks if a given Dataloader has ``__len__`` method implemented i.e. if it is a finite dataloader or
-    infinite dataloader."""
+def sized_len(dataloader: object) -> Optional[int]:
```
justusschock, line 50: nit: instead of `object` I'd prefer `Any`, which should be the same (applies to other parts of this PR as well)

justusschock (approved):

src/lightning_fabric/utilities/data.py
carmocca: object is more informative, as Any signals the type checker to ignore this path entirely: https://stackoverflow.com/a/39817126

src/lightning_fabric/utilities/data.py
```diff
@@ -42,31 +43,34 @@ def __call__(self, *args: Any) -> None:
...
 
-def has_iterable_dataset(dataloader: DataLoader) -> bool:
+def has_iterable_dataset(dataloader: object) -> bool:
     return hasattr(dataloader, "dataset") and isinstance(dataloader.dataset, IterableDataset)
 
 
-def has_len(dataloader: Union[DataLoader, Iterable, Dataset]) -> bool:
-    """Checks if a given Dataloader has ``__len__`` method implemented i.e. if it is a finite dataloader or
-    infinite dataloader."""
+def sized_len(dataloader: object) -> Optional[int]:
```
awaelchli, line 50: It's public, so maybe needs a docstring?

awaelchli (approved):

PR #16628, opened 2023-02-03
Fix TPU tests

carmocca:
## What does this PR do?

https://github.com/Lightning-AI/lightning/pull/16571 was merged with failing TPU tests


tests/tests_fabric/strategies/launchers/test_xla.py
```diff
@@ -1,137 +0,0 @@
...
-
-_loader = DataLoader(RandomDataset(32, 64))
-_iterable_loader = DataLoader(RandomIterableDataset(32, 64))
-_loader_no_len = CustomNotImplementedErrorDataloader(_loader)
-
-
-@RunIf(tpu=True)
-@pytest.mark.parametrize("dataloader", [None, _iterable_loader, _loader_no_len])
-@mock.patch("lightning.fabric.strategies.xla.XLAStrategy.root_device")
-def test_xla_validate_unsupported_iterable_dataloaders(_, dataloader, monkeypatch):
```
carmocca, line 108: This entire file is duplicated

Borda (approved):

tests/tests_fabric/strategies/test_xla.py
```diff
@@ -86,7 +86,7 @@ def test_tpu_reduce():
 def test_xla_mp_device_dataloader_attribute(_, monkeypatch):
     import torch_xla.distributed.parallel_loader as parallel_loader
 
-    mp_loader_mock = Mock()
+    mp_loader_mock = Mock(autospec=parallel_loader.MpDeviceLoader)
```
awaelchli, line 89: @carmocca I tried spec as well in the original PR, but it's not enough because here we are attempting to mock a type. The isinstance check won't be so easy to pass `isinstance(x, mock)`. I'm not sure how to properly do this.

carmocca: Works by overriding isisntancecheck

awaelchli: That's it! I completely forgot that I used the same trick in test_ddp_module_state_dict

awaelchli (approved):

PR #16611, opened 2023-02-02
Set `find_unused_parameters=False` as the default

awaelchli:
## What does this PR do?

Fixes #7330
Fixes #14486
Closes https://github.com/Lightning-AI/lightning/issues/12445
Part of #12398

The default value for DDP's `find_unused_parameters` in PyTorch is `False`. So far, Lightning has always flipped that to `True` because it tried to minimize the confusion of users who would otherwise see an error from PyTorch: 

> RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss. You can enable unused parameter detection by passing the keyword argument `find_unused_parameters=True` to `torch.nn.parallel.DistributedDataParallel`, and by 
> making sure all `forward` function outputs participate in calculating loss. 
> If you already have done the above, then the distributed data parallel module wasn't able to locate the output tensors in the return value of your module's `forward` function. Please include the loss function and the structure of the return value of `forward` of your module when reporting this issue (e.g. list, dict, iterable).
> Parameter indices which did not receive grad for rank 0: 0 1
> In addition, you can set the environment variable TORCH_DISTRIBUTED_DEBUG to either INFO or DETAIL to print out information about which particular parameters did not receive gradient on this rank as part of this error
> 

This error would often occur when users dealt multiple optimizers (GAN) etc. After #16539, this issue is less significant. However, the drawback of setting `find_unused_parameters` is that you compromise on performance. PyTorch shows a warning:

> Warning: find_unused_parameters=True was specified in DDP constructor, but did not find any unused parameters in the forward pass. This flag results in an extra traversal of the autograd graph every iteration,  which can adversely affect performance. If your model indeed never has any unused parameters in the forward pass, consider turning this flag off. Note that this warning may be a false positive if your model has flow control causing later iterations to have unused parameters. (function operator())

But obviously, the warning is pointing at the PyTorch DDP API that the Lightning user doesn't have access to. Thus, the warning is not very helpful. 

Since Trainer 2.0 favors performance and parity with PyTorch, we will switch `find_unused_parameters` to `False` to align with PyTorch's defaults. Discussion from the past: https://github.com/Lightning-AI/lightning/discussions/6219

### Does your PR introduce any breaking changes? If yes, please list them.

In summary, these are the possible configurations a user can run into:

**Case 1: User has unused parameters in their forward-backward**
- Trainer < 2.0: No errors, no warnings, expected performance hit
[...]

Issue #16410, opened 2023-01-17
Process reconciliation as a plugin


awaelchli:
### Outline & Motivation

The DDP strategy and its subclasses have a feature called "deadlock detection and process reconciliation". It can ensure that all processes terminate properly when an error occurs on a subset of the ranks. Without this feature, the processes where no errors occur would continue to run and hang/wait at the collectives. 

Pro:
- Can save you costs when running in the cloud.
- No zombie processes you have to manually kill

Con:
- Implementation is hardcoded into the DDPStrategy, does not work well with inheritance
- Makes [trainer exception handling](https://github.com/Lightning-AI/lightning/blob/fc195b95405e9e2629466e5b28c6a9243209d596/src/pytorch_lightning/trainer/call.py#L57-L59) complex 
- Assumes a shared filesystem

### Pitch

1. Remove the feature from PL strategies (#16204)
2. Introduce it as a plugin under fabric
3. Introduce `Strategy.on_exception` that the exception handler can call in a standardized way
4. Re-introduce it in PL strategies once flattening preparations for Fabric integration are done
- [ ] Subtask: Add the plugin to other strategies such as deepspeed:  #16518

A strategy can enable the plugin like so:
```py
if not torch_greater_equal_foo and has_shared_filesystem:
    enable_plugin()
```

and by implementing 
```py
def on_exception(self, exception):
    self.plugin.reconciliate_processes(...)
```

### Additional context

Credit for the ideas @carmocca 

_No response_

cc @justusschock @awaelchli @carmocca

carmocca:
In #16525 I implemented a limited version of this proposal. It's intra-node only and only runs under the SIGTERM signal, but it could be extended to support inter-node, more signals, or any exception type.

My code was loosely based on https://github.com/pytorch/pytorch/blob/v1.13.1/torch/distributed/elastic/multiprocessing/api.py

Issue #14652, opened 2022-09-12
Support Python 3.11


akihironitta:
## 🚀 Feature
This is a tracking issue to support Python 3.11 released on 2022-10-03.

### Additional context
- [PEP 664 – Python 3.11 Release Schedule](https://peps.python.org/pep-0664/)
- [What’s New In Python 3.11](https://docs.python.org/3.11/whatsnew/3.11.html)

______________________________________________________________________

#### If you enjoy Lightning, check out our other projects! ⚡

- [**Metrics**](https://github.com/Lightning-AI/metrics): Machine learning metrics for distributed, scalable PyTorch applications.

- [**Lite**](https://pytorch-lightning.readthedocs.io/en/latest/starter/lightning_lite.html): enables pure PyTorch users to scale their existing code on any kind of device while retaining full control over their own loops and optimization logic.

- [**Flash**](https://github.com/Lightning-AI/lightning-flash): The fastest way to get a Lightning baseline! A collection of tasks for fast prototyping, baselining, fine-tuning, and solving problems with deep learning.

- [**Bolts**](https://github.com/Lightning-AI/lightning-bolts): Pretrained SOTA Deep Learning models, callbacks, and more for research and production with PyTorch Lightning and PyTorch.

- [**Lightning Transformers**](https://github.com/Lightning-AI/lightning-transformers): Flexible interface for high-performance research using SOTA Transformers leveraging PyTorch Lightning, Transformers, and Hydra.

cc @borda @carmocca

awaelchli:
Regarding PyTorch's support of 3.11, updates can be found here: https://github.com/pytorch/pytorch/issues/86566

akihironitta:
Looking at the linked PyTorch's GH issue, I think we can start working on this. The instability issue they mention only applies to their conda CI.

mrembalski:
Hi there! Does this have a planned date of release?

Borda:
> Hi there! Does this have a planned date of release?

We are happy to add it, but in this case, we are very dependent on PyToch, see: https://github.com/pytorch/pytorch/issues/86566

mrembalski:
> > Hi there! Does this have a planned date of release?
> 
> We are happy to add it, but in this case, we are very dependent on PyToch, see: [pytorch/pytorch#86566](https://github.com/pytorch/pytorch/issues/86566)

Sorry, I did not know that. Thank you for explaining.

eric-01001000:
[...]

PR #16646, opened 2023-02-06
Add `Strategy.on_exception`

### LINKED ISSUE #12445: Adapt `find_unused_parameters` warning
## 🚀 Feature

Latter PyTorch versions introduced the following warning

https://github.com/pytorch/pytorch/blob/d0adb5ff264df8e0e057ce1178feb30198c601d2/torch/csrc/distributed/c10d/reducer.cpp#L1256-L1264

which appears when `find_unused_parameters=True` but it's not necessary.

### Motivation

We set `find_unused_parameters=True` by default, so this warning will appear for a large percentage of users.

### Pitch

Catch this warning and produce another one that explicitly mentions how to modify the strategy passed.

cc @borda @akihironitta

awaelchli:
## What does this PR do?

Follow up to #16611
Closes #12445

This PR adds

```py
def on_exception(self, exception):
    ...
```

to the Strategy base interface. An implementation can optionally override this method and process exceptions when they occur in Trainer. This will also be useful for #16410.

This makes it possible that the strategy can replace/edit an exception. For example, the following error emitted by DDP

> RuntimeError: Expected to have finished reduction in the prior iteration before starting a new one. This error indicates that your module has parameters that were not used in producing loss. You can enable unused parameter detection by passing the keyword argument find_unused_parameters=True to torch.nn.parallel.DistributedDataParallel, and by
> making sure all forward function outputs participate in calculating loss.
> If you already have done the above, then the distributed data parallel module wasn't able to locate the output tensors in the return value of your module's forward function. Please include the loss function and the structure of the return value of forward of your module when reporting this issue (e.g. list, dict, iterable).
> Parameter indices which did not receive grad for rank 0: 0 1
> In addition, you can set the environment variable TORCH_DISTRIBUTED_DEBUG to either INFO or DETAIL to print out information about which particular parameters did not receive gradient on this rank as part of this error

gets replaced with a more useful one for Lightning users:

> RuntimeError: It looks like your LightningModule has parameters that were not used in producing the loss returned by training_step. If this is intentional, you must enable the detection of unused parameters in DDP, either by setting the string value `strategy='ddp_find_unused_parameters_true'` or by setting the flag in the strategy with `strategy=DDPStrategy(find_unused_parameters=True)`.

## Before submitting

## PR review

## Did you have fun?


tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -276,3 +276,23 @@ def test_ddp_strategy_checkpoint_zero_redundancy_optimizer(tmpdir, strategy):
...
+        return super().training_step(batch, batch_idx)
+
+
+def test_ddp_strategy_find_unused_parameters_exception():
+    """Test that the DDP strategy can change PyTorch's error message so that it's more useful for Lightning
+    users."""
+    # TODO: Update default
+    trainer = Trainer(accelerator="cpu", devices=1, strategy="ddp_find_unused_parameters_false", max_steps=2)
+    with pytest.raises(RuntimeError, match="It looks like your LightningModule has parameters that were not used in"):
+        trainer.fit(UnusedParametersModel())
```
awaelchli, line 298: We add a full integration test so that in case PyTorch changes the message, this test will fail.

tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -276,3 +276,23 @@ def test_ddp_strategy_checkpoint_zero_redundancy_optimizer(tmpdir, strategy):
...
+    def training_step(self, batch, batch_idx):
+        with torch.no_grad():
+            batch = self.intermediate_layer(batch)
+        return super().training_step(batch, batch_idx)
+
+
+def test_ddp_strategy_find_unused_parameters_exception():
+    """Test that the DDP strategy can change PyTorch's error message so that it's more useful for Lightning
+    users."""
+    # TODO: Update default
```
awaelchli, line 295: TODO after #16611 gets merged

tests/tests_pytorch/strategies/test_ddp_spawn.py
```diff
@@ -73,3 +77,14 @@ def test_ddp_all_dataloaders_passed_to_fit(tmpdir):
     )
     trainer.fit(model, train_dataloaders=model.train_dataloader(), val_dataloaders=model.val_dataloader())
     assert trainer.state.finished, "DDP doesn't work with dataloaders passed to fit()."
+
+
+def test_ddp_spawn_find_unused_parameters_exception():
+    """Test that the DDP strategy can change PyTorch's error message so that it's more useful for Lightning
+    users."""
+    # TODO: Update default
+    trainer = Trainer(accelerator="cpu", devices=1, strategy="ddp_spawn_find_unused_parameters_false", max_steps=2)
```
awaelchli, line 86: TODO after https://github.com/Lightning-AI/lightning/pull/16611 gets merged
```suggestion
    trainer = Trainer(accelerator="cpu", devices=1, strategy="ddp", max_steps=2)
```

tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -276,3 +276,23 @@ def test_ddp_strategy_checkpoint_zero_redundancy_optimizer(tmpdir, strategy):
...
+    trainer = Trainer(accelerator="cpu", devices=1, strategy="ddp_find_unused_parameters_false", max_steps=2)
```
awaelchli, line 296: TODO after https://github.com/Lightning-AI/lightning/pull/16611 gets merged
```suggestion
    # TODO: Update default
    trainer = Trainer(accelerator="cpu", devices=1, strategy="ddp", max_steps=2)
```

src/lightning/pytorch/utilities/exceptions.py
```diff
@@ -27,3 +29,9 @@ class SIGTERMException(SystemExit):
 
 class _TunerExitException(Exception):
     """Exception used to exit early while tuning."""
+
+
+def _replace_message(exception: BaseException, pattern: str, new_message: str) -> None:
+    exception.args = tuple(
```
carmocca, line 35: Did you explore playing with the `__cause__` attribute?

https://legacy.python.org/dev/peps/pep-3134/
https://stackoverflow.com/questions/24752395/python-raise-from-usage

We should keep the original message too. As `raise ... from ...` does

src/lightning/pytorch/trainer/call.py
```diff
@@ -53,6 +53,7 @@ def _call_and_handle_interrupt(trainer: "pl.Trainer", trainer_fn: Callable, *arg
     except BaseException as exception:
         trainer.state.status = TrainerStatus.INTERRUPTED
         trainer._call_callback_hooks("on_exception", exception)
+        trainer.strategy.on_exception(exception)
```
carmocca, line 56: Was this not added to the except block above on purpose?

awaelchli: Yes, because KeyboardInterrupt is also not passed to callbacks.

src/lightning/pytorch/utilities/exceptions.py
awaelchli: Are you suggesting to return a new exception from the on_exception method?

carmocca: No. I was just asking if you had tried hacking around the `__cause__` attribute. I don't know if this is possible.

This would also be replaced by https://peps.python.org/pep-0678/ when 3.11 is our minimum version. Can you add a note about that?

awaelchli: I unfortunately don't see how I could manipulate `__cause__`. The value of `__cause__` needs to be another exception.

src/lightning/pytorch/trainer/call.py
carmocca: Are you referring to the `trainer._call_callback_hooks("on_exception", exception)` call? It's there too

src/lightning/pytorch/utilities/exceptions.py
carmocca: And what's the issue with adding another exception?

```python
try:
    raise ValueError("foobar")
except BaseException as err:
    if _PYTHON_GREATER_EQUAL_3_11:
        err.add_note("barfoo")
    else:
        err.__cause__ = type(err)("barfoo")
    raise
```

awaelchli: That's what I tried. If you look at the ddp_spawn test I added, there the exception type is ProcessRaisedException, and it can't be instantiated with just `type(err)(*exception.args)`. It just doesn't generalize like that.

justusschock (approved):

carmocca (approved):

src/lightning/pytorch/utilities/exceptions.py
awaelchli: It requires #14652 before it can be properly tested.

PR #16656, opened 2023-02-06
Allow frozen data classes in optimizer state dict

janEbert:
## What does this PR do?

We use [the (now merged) PR in `lightning-utilities` that allows frozen dataclasses to be ignored in `apply_to_collection`](https://github.com/Lightning-AI/utilities/pull/98). With this, we can now allow saving optimizers for which the `state` includes frozen dataclasses that we just want to pass through.
Any optimizer that includes a frozen dataclass in its `state`, is not supported in Lightning without this, instead causing errors.

This should be backward-compatible, since previously, frozen dataclasses in optimizers simply caused an error. Other than that, nothing changed.

<details>

- [ ] Was this **discussed/agreed** via a GitHub issue? (not for typos and docs)
  I tried, tests were broken for me even without the changes. Manually running `tests/tests_fabric/utilities/test_optimizer.py` does not complain.
  Did not seem necessary.

</details>


janEbert:
Converted to a draft for safety regarding the other merge happening first.

janEbert:
Other merge happened, this is good to go from my side.

Borda (approved):

src/lightning/fabric/utilities/optimizer.py
```diff
@@ -31,4 +31,4 @@ def _optimizers_to_device(optimizers: Iterable[Optimizer], device: _DEVICE) -> N
 def _optimizer_to_device(optimizer: Optimizer, device: _DEVICE) -> None:
     """Moves the state of a single optimizer to the device."""
     for p, v in optimizer.state.items():
-        optimizer.state[p] = apply_to_collection(v, Tensor, move_data_to_device, device)
+        optimizer.state[p] = apply_to_collection(v, Tensor, move_data_to_device, device, allow_frozen=True)
```
awaelchli, line 34: @janEbert Before this can be used, we need to bump the version in the requirements file once the lightning-utilities has made a release. cc @Borda

janEbert:
Tests fail because it expects the `lightning-utiliities` master, I guess. Would it make sense to tag a new version in there? Seems a bit overkill for the small change, but it would be a solution.

src/lightning/fabric/utilities/optimizer.py
janEbert: You were faster ;p

awaelchli: @Borda Is lightning utilities ready to be used here now?

awaelchli (approved):

carmocca (approved):

janEbert:
Thanks for your work on fixing everything up!

PR #16680, opened 2023-02-07
Move logger utilities with the loggers

carmocca:
## What does this PR do?

See title


src/lightning/pytorch/callbacks/progress/base.py
```diff
@@ -267,11 +266,11 @@ def get_standard_metrics(trainer: "pl.Trainer") -> Dict[str, Union[int, str]]:
     Return:
         Dictionary with the standard metrics to be displayed in the progress bar.
     """
-
     items_dict: Dict[str, Union[int, str]] = {}
     if trainer.loggers:
-        version = _version(trainer.loggers)
-        if version is not None:
+        from lightning.pytorch.loggers.utilities import _version
```
awaelchli, line 271: @carmocca The motivation behind having these functions in the utilities module was exactly to avoid such circular imports that could occur.

awaelchli (approved):

justusschock (approved):

PR #16702, opened 2023-02-09
Group trainer call methods as functions

carmocca:
## What does this PR do?


src/lightning/pytorch/core/module.py
```diff
@@ -42,6 +42,7 @@
 from lightning.pytorch.core.optimizer import LightningOptimizer
 from lightning.pytorch.core.saving import ModelIO
 from lightning.pytorch.loggers import Logger
+from lightning.pytorch.trainer import call
```
Borda, line 45:
```suggestion
from lightning.pytorch.trainer import _calls
```
 seems all func there protected so let make whole module protected

src/lightning/pytorch/loops/dataloader/evaluation_loop.py
```diff
@@ -26,6 +26,7 @@
 from lightning.pytorch.loops.epoch import _EvaluationEpochLoop
 from lightning.pytorch.loops.fetchers import _DataFetcher
 from lightning.pytorch.loops.utilities import _select_data_fetcher, _set_sampler_epoch
+from lightning.pytorch.trainer import call
```
Borda, line 29: Then also rather import each function then this module call which may confuse as being another object...

Borda (approved):

src/lightning/pytorch/loops/dataloader/evaluation_loop.py
carmocca: Doing it like this follows the pattern started by William at https://github.com/Lightning-AI/lightning/pull/14871 and https://github.com/Lightning-AI/lightning/pull/14862

awaelchli (approved):
PR LGTM, but not happy about the pattern that was started prior to this :)

src/lightning/pytorch/core/module.py
carmocca: I'll leave this for a follow-up if we want to do it. We don't have any other protected modules and I don't want to stray away from the current pattern used in the trainer unless there's consensus

Borda:
> PR LGTM, but not happy about the pattern that was started prior to this :)

the same here :(

PR #16704, opened 2023-02-09
Move `inference_mode` logic to the loops

carmocca:
## What does this PR do?

Part of the `trainer.py` cleanup

cc @justusschock @awaelchli @carmocca @borda

### DIFF
--- a/src/lightning/pytorch/loops/dataloader/evaluation_loop.py
+++ b/src/lightning/pytorch/loops/dataloader/evaluation_loop.py
@@ -25,7 +25,7 @@ from lightning.pytorch.callbacks.progress.rich_progress import _RICH_AVAILABLE
 from lightning.pytorch.loops.dataloader import _DataLoaderLoop
 from lightning.pytorch.loops.epoch import _EvaluationEpochLoop
 from lightning.pytorch.loops.fetchers import _DataFetcher
-from lightning.pytorch.loops.utilities import _select_data_fetcher, _set_sampler_epoch
+from lightning.pytorch.loops.utilities import _no_grad_context, _select_data_fetcher, _set_sampler_epoch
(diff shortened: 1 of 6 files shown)


src/lightning/pytorch/loops/dataloader/evaluation_loop.py
```diff
@@ -88,6 +89,7 @@ def skip(self) -> bool:
         max_batches = self._get_max_batches()
         return sum(max_batches) == 0
 
+    @_no_grad_context
```
awaelchli, line 92: Does it have to be a decorator? Can't we just keep it as a context manager, moving it directly over the logic in run()?

IMO the decorator approach is more complicated than it has to be. It requires this validation logic even though it is just internal.

carmocca: It doesnt have to be. I just thought it would be more readable like this, as it doesnt add an extra indentation level to the run implementation. Ccing @justusschock for the tiebreaker

Borda (approved):

src/lightning/pytorch/loops/dataloader/evaluation_loop.py
justusschock: I agree with @carmocca here that this is more readable.

justusschock (approved):

awaelchli (approved):

src/lightning/pytorch/loops/dataloader/evaluation_loop.py
awaelchli: Okay, so we can do the current way. 

I'm still convinced that the ctx manager way is more readable as it has less code and nesting. The extra indentation level is just how it works in Python, so I'm not seeing it as a strong argument.

PR #16706, opened 2023-02-09
ci: enable lightning on TPU

Borda:
## What does this PR do?

add parametrization for running TPU testing for the lightning mono package

Fixes #\<issue_number>

<details>

</details>

## PR review

<details>

</details>


.github/workflows/tpu-tests.yml
```diff
@@ -3,7 +3,7 @@ name: Test PyTorch - TPU
 on:
   push:
     branches: [master, "release/*"]
-  pull_request_target:
+  pull_request:  # fixme: pull_request_target
```
akihironitta, line 6: Should this be reverted before merging this PR?
```suggestion
  pull_request:  # fixme: pull_request_target
```

akihironitta (approved):

Borda:
> We need to decrease the TPU CI load, and adding one more parametrization will make things worse.

so you suggest rather don't know if it works?

carmocca:
Yes, exactly :wink:

Now, why?
- TPU capacity is low.
- We already test `lightining` with the CPU and CUDA workflows
- We effectively already don't know if anything works because of the low capacity.
- Developments happen with `lightning` (not the standalone versions) so we'd catch issues when running on TPUs as we develop.

Alternatively, we could test `lightning.pytorch` and `lightning.fabric`, but my last point suggest that the current approach is better.

Borda:
> * We already test `lightining` with the CPU and CUDA workflows

sounds like we do not need to test the motor-bike when we are already testing a bicycle just because both have two wheels... :chipmunk: 

> * TPU capacity is low.

then leave it as standalone extension when the load will be limited

> * We effectively already don't know if anything works because of the low capacity.

it is not a reason to don't test it in future...

> * Developments happen with `lightning` (not the standalone versions) so we'd catch issues when running on TPUs as we develop.

as in some of your PR you limit the testing/triggering events making it the same case standalone... so when you edit TPU code it would be in the other place and if general utils (you already suggested skipping TPU testing so no change in fact) and we have compatibility repo for this case...

Borda:
closing as @carmocca does not think testing `lightning` is needed, lets's just whole all will work somehow... :chipmunk:

PR #16711, opened 2023-02-10
Group `torch.compile` utilities together

carmocca:
## What does this PR do?

### DIFF
--- a/src/lightning/pytorch/core/module.py
+++ b/src/lightning/pytorch/core/module.py
@@ -34,7 +34,7 @@ from lightning.fabric.utilities.apply_func import convert_to_tensors
 from lightning.fabric.utilities.cloud_io import get_filesystem
 from lightning.fabric.utilities.device_dtype_mixin import _DeviceDtypeModuleMixin
 from lightning.fabric.utilities.distributed import _distributed_available, _sync_ddp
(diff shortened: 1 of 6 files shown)


src/lightning/pytorch/core/module.py
```diff
@@ -1640,86 +1640,6 @@ def _register_sharded_tensor_state_dict_hooks_if_available(self) -> None:
                 weakref.proxy(self), pre_load_state_dict_hook, True  # type: ignore[arg-type]
             )
 
-    @classmethod
-    def from_compiled(cls, model: "torch._dynamo.OptimizedModule") -> "pl.LightningModule":
```
awaelchli, line 1644: Weren't these meant to be actual methods on the LM? Maybe keep but call the function below?

awaelchli (approved):

src/lightning/pytorch/core/module.py
carmocca: cc @lantiga: Could you let us know if you wrote these `classmethod`s for a specific benefit? The implementation does not need it. The motivation for moving it is just to improve the readability of `module.py`

Borda (approved):

src/lightning/pytorch/core/module.py
lantiga: No specific benefit, just grouping. It’s 100% fine to move it.

lantiga (approved):

PR #16712, opened 2023-02-10
Move `Trainer._log_hyperparams` to an utility

carmocca:
## What does this PR do?

### DIFF
--- a/src/lightning/pytorch/loggers/utilities.py
+++ b/src/lightning/pytorch/loggers/utilities.py
@@ -16,6 +16,9 @@
 from pathlib import Path
 from typing import Any, List, Tuple, Union
 
+from torch import Tensor
+
+import lightning.pytorch as pl
 from lightning.pytorch.callbacks import Checkpoint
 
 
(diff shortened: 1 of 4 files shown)


awaelchli (approved):

Borda (approved):

src/lightning/pytorch/loggers/utilities.py
```diff
@@ -52,3 +55,42 @@ def _scan_checkpoints(checkpoint_callback: Checkpoint, logged_model_time: dict)
...
     checkpoints = [c for c in checkpoints if c[1] not in logged_model_time.keys() or logged_model_time[c[1]] < c[0]]
     return checkpoints
+
+
+def _log_hyperparams(trainer: "pl.Trainer") -> None:
+    if not trainer.loggers:
+        return
+
+    pl_module = trainer.lightning_module
+    datamodule_log_hyperparams = trainer.datamodule._log_hyperparams if trainer.datamodule is not None else False
```
tbenst, line 65: datamodule is not required to have `_log_hyperparams`, so I think this needs to check if `_log_hyperparams` is an attribute of `trainer.datamodule`. @carmocca @Borda

tbenst: oops please disregard! I forgot to call `super().__init__()` in my subclass of pl.LightningDataModule

PR #16715, opened 2023-02-10
Adds Gradient Clipping to Fabric

### LINKED ISSUE #13266: [Fabric] Gradient Clipping with AMP
## 🚀 Feature

Currently, there doesn't seem to be a way to do gradient clipping when using AMP with LightningLite. This is because the grad scalar code is behind the NativeMixedPrecisionPlugin, and there doesn't seem to be any way of changing this for LightningLite.

### Motivation

Grad clipping is very popular across DL, and having some way to do this is very important.

### Pitch

I see that the NativeMixedPrecisionPlugin has functionality for doing this for LightningModules, but not for Lite models. 

The method used for LightningModules (using `configure_gradient_clipping`) -- if extended to work with Lite modules -- also seems to not be in the spirit of Lite (but it's better than nothing, I suppose).

cc @borda @carmocca @justusschock @awaelchli @akihironitta @rohitgr7

justusschock:
## What does this PR do?

Adds Gradient Clipping to Fabric.

Fixes  #13266

<details>

</details>

## PR review

<details>

</details>


src/lightning/fabric/strategies/strategy.py
```diff
@@ -305,6 +306,39 @@ def teardown(self) -> None:
...
         self.checkpoint_io.teardown()
 
+    def _unscale_params(self, optimizer):
+        scaler = getattr(self.precision, "scaler", None)
+        if scaler is not None:
+            if _optimizer_handles_unscaling(optimizer):
+                raise NotImplementedError("Gradient clipping is not implemented for optimizers handling the unscaling.")
+            scaler.unscale_(optimizer)
+
+    def clip_gradients_norm(
```
awaelchli, line 316: IMO, the base implementation should be in the precision plugin base, and the strategy here just calls `self.precision.clip_...`

A strategy implementing the gradient clipping can decide whether to override or call into its own precision plugin.

src/lightning/fabric/strategies/strategy.py
```diff
@@ -305,6 +306,39 @@ def teardown(self) -> None:
...
+        from lightning.fabric.wrappers import _unwrap_objects
+
+        optimizer = _unwrap_objects(optimizer)
+        self._unscale_params(optimizer)
+        parameters = self.precision.main_params(optimizer)
+        return torch.nn.utils.clip_grad_norm_(
+            parameters, max_norm=max_norm, norm_type=norm_type, error_if_nonfinite=error_if_nonfinite
+        )
+
+    def clip_gradients_value(self, optimizer, clip_val: Union[float, int]):
```
awaelchli, line 333: As briefly chatted offline, we may want to consider a single clip_gradients method, and the mode as a string. The strategy implementations would handle this similarly to the collectives, where sometimes not all possible inputs are supported.

awaelchli (commented):
Quick initial review

src/lightning/fabric/strategies/fsdp.py
```diff
@@ -267,6 +270,28 @@ def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
...
+        self,
+        optimizer,
+        max_norm: Union[float, int],
+        norm_type: Union[float, int] = 2.0,
+        error_if_nonfinite: bool = True,
+    ):
+        """Clip gradients by norm."""
+        from lightning.fabric.wrappers import _unwrap_objects
+
+        optimizer = _unwrap_objects(optimizer)
```
awaelchli, line 283: @justusschock We will need a `Fabric.clip_gradients()` method too. Its responsibility will be to pass the real optimizer to the strategy instead of the _FabricOptimizer. The strategy is designed to always operate on the non-fabric wrapped objects.

src/lightning/fabric/strategies/strategy.py
justusschock: I disagree here. IMO this should live in the strategy itself as it's not related to precision handling. Yes, it may need to unscale gradients, but that's all there is. And that's exactly why the precision plugin lives in the strategy now and they aren't on the same level within the accelerator anymore.

This is especially true as some strategies (e.g. FSDP) require collective communications here and I don't like the implementations of this to be scattered between precision plugins and strategies.

src/lightning/fabric/strategies/strategy.py
justusschock: I know, this just wasn't added yet, but it will be :)

src/lightning/fabric/strategies/fsdp.py
justusschock: I will move the unwrapping out of here then :)

src/lightning/fabric/strategies/strategy.py
justusschock: after discussing offline with @awaelchli we agreed to move `unscale_params` as `unscale_gradients` to the precision plugin which can then be called here.

src/lightning/fabric/plugins/precision/amp.py
```diff
@@ -90,3 +90,13 @@ def _autocast_context_manager(self) -> torch.autocast:
...
         # https://github.com/pytorch/pytorch/issues/67233
         return torch.autocast(self.device, dtype=torch.bfloat16 if self.precision == "bf16" else torch.half)
+
+
+def _optimizer_handles_unscaling(optimizer: Any) -> bool:
+    """Determines whether a PyTorch optimizer handles unscaling gradients in the step method rather than through the
+    :class:`torch.cuda.amp.GradScaler`.
+
+    Since, the current implementation of this function checks a PyTorch internal variable on the optimizer, the return
+    value will only be reliable for built-in PyTorch optimizers.
```
carmocca, line 100: If you assume this, then you could use `TypeGuard[Optimizer]` as the return annotation

src/lightning/fabric/strategies/fsdp.py
```diff
@@ -127,6 +127,7 @@ def __init__(
         self.cpu_offload = _init_cpu_offload(cpu_offload)
         self.backward_prefetch = backward_prefetch
         self.mixed_precision = mixed_precision
+        self._module: "FullyShardedDataParallel" = ...
```
carmocca, line 130: Initializing to ellipsis is not a good idea. Use None instead

justusschock: I wanted to avoid the typing issue here, but yeah will do and add type casts.

src/lightning/fabric/plugins/precision/amp.py
justusschock: well actually not really since this wouldn't allow for arbitrary custom optimizers. This is also why we use `getattr` instead of `optimizer._steps_supports_amp_scaling

awaelchli:
@justusschock For the misbehaving collectives test, I vote for adding `Runif(skip_windows=True)`.

src/lightning/fabric/plugins/precision/precision.py
```diff
@@ -97,6 +97,9 @@ def main_params(self, optimizer: Optimizer) -> _PARAMETERS:
         for group in optimizer.param_groups:
             yield from group["params"]
 
+    def unscale_gradients_(self, optimizer: Optimizer) -> None:
```
awaelchli, line 100: I wouldn't add the underscore at the end. We don't follow torch's "inplace" operation notation anywhere else.

tests/tests_fabric/strategies/test_grad_clipping.py
```diff
@@ -0,0 +1,125 @@
...
+from lightning.fabric.strategies import DeepSpeedStrategy, FSDPStrategy
+from lightning.fabric.wrappers import _FabricModule, _FabricOptimizer
+
+
+class _MyFabricGradNorm(BoringFabric):
+    def after_backward(self, model: _FabricModule, optimizer: _FabricOptimizer):
+        self.clip_gradients(model, optimizer, max_norm=0.05, error_if_nonfinite=True)
+
+        with model._forward_module.summon_full_params(model._forward_module) if isinstance(
+            self.strategy, FSDPStrategy
```
awaelchli, line 19: Is this a leftover from debugging fsdp?

tests/tests_fabric/strategies/test_grad_clipping.py
```diff
@@ -0,0 +1,125 @@
...
+    "strategy,num_devices,precision",
+    [
+        pytest.param("auto", 1, "16-mixed"),
+        pytest.param("ddp", 2, "32-true"),
+        pytest.param("dp", 2, "16-mixed", marks=RunIf(min_cuda_gpus=2)),
+        pytest.param(
+            "fsdp",
+            2,
+            "32-true",
+            marks=[RunIf(min_cuda_gpus=2), pytest.mark.xfail(reason="Testing with FSDP is not yet correct")],
```
awaelchli, line 57: how can this work if we raise NotImplementedError anyway? :)

tests/tests_fabric/strategies/test_grad_clipping.py
```diff
@@ -0,0 +1,125 @@
+from contextlib import nullcontext
```
awaelchli, line 1: Ok with this, since the tests are already written now. In general, minor suggestion would be to have the grad clipping tested in each strategy specific file, e.g. test/tests_fabric/strategies/xzy etc. This way, when adding an new strategy, one can easily look at the others and know what needs to be tested and to what extent. We could refactor it in a follow up.

src/lightning/fabric/fabric.py
```diff
@@ -363,6 +363,35 @@ def backward(self, tensor: Tensor, *args: Any, model: Optional[_FabricModule] =
 
         self._precision.backward(tensor, module, *args, **kwargs)
 
+    def clip_gradients(
```
awaelchli, line 366: would be nice to have this covered in `tests/tests_fabric/test_fabric.py` with a strategy mock.

awaelchli (commented):
LGTM, minor comments

tests/tests_fabric/strategies/test_grad_clipping.py
justusschock: I can still add this :)

tests/tests_fabric/strategies/test_grad_clipping.py
justusschock: we don't :) NotImplemented is only raised for clipping by value (this function tests clip by val :) ) For clip_gradients_norm, we only raise a warning currently (which will be gone after we figured out testing).

awaelchli (approved):
🎉

src/lightning/fabric/strategies/fsdp.py
```diff
@@ -202,6 +202,8 @@ def setup_module(self, module: Module) -> "FullyShardedDataParallel":
         if _TORCH_GREATER_EQUAL_1_13 and self._activation_checkpointing:
             _setup_activation_checkpointing(module=wrapped_module, layers=self._activation_checkpointing)
 
+        self._module = module
```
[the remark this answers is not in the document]
awaelchli: This is a left-over we wanted to avoid, right @justusschock?

tests/tests_fabric/strategies/test_single_device.py
```diff
@@ -52,3 +55,69 @@ def test_single_device_module_to_device():
...
+        grad_max = torch.max(torch.stack(grad_max_list))
+        torch.testing.assert_close(grad_max.abs(), torch.tensor(1e-10, device=self.device))
+
+    def run(self):
+        while True:
+            try:
+                super().run()
+                break
+            except RuntimeError:  # nonfinite grads -> skip and continue
+                pass
```
carmocca, line 99: Can we use a different pattern? As I understand, you are trying to assert that nonfinite grads is not raised?

But if it is, then this will run endlessly?

carmocca (commented):
Pushed minor test improvements

src/lightning/fabric/fabric.py
```diff
@@ -363,6 +363,35 @@ def backward(self, tensor: Tensor, *args: Any, model: Optional[_FabricModule] =
...
+    ) -> Optional[torch.Tensor]:
+        if clip_val is not None and max_norm is not None:
+            raise ValueError(
+                "Only one of `clip_val` or `max_norm` can be set as this specifies the underlying clipping algorithm!"
+            )
+
+        if clip_val is not None:
+            self.strategy.clip_gradients_value(_unwrap_objects(module), _unwrap_objects(optimizer), clip_val=clip_val)
+            return None
+
```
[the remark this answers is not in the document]
justusschock: I did this initially, but then mypy complained, that this doesn't have a return value.

src/lightning/fabric/plugins/precision/amp.py
```diff
@@ -93,3 +93,20 @@ def _autocast_context_manager(self) -> torch.autocast:
         # the dtype could be automatically inferred but we need to manually set it due to a bug upstream
         # https://github.com/pytorch/pytorch/issues/67233
         return torch.autocast(self.device, dtype=torch.bfloat16 if self.precision == "bf16-mixed" else torch.half)
+
+    def unscale_gradients(self, optimizer: Optimizer) -> None:
+        scaler = getattr(self.precision, "scaler", None)
```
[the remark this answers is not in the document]
justusschock: whoops, yes. Thanks!

src/lightning/fabric/strategies/fsdp.py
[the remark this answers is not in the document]
justusschock: yes, this is a left-over I overlooked

src/lightning/fabric/strategies/fsdp.py
```diff
@@ -268,6 +270,30 @@ def broadcast(self, obj: TBroadcast, src: int = 0) -> TBroadcast:
...
+        self,
+        module: "FullyShardedDataParallel",
+        optimizer: Optimizer,
+        max_norm: Union[float, int],
+        norm_type: Union[float, int] = 2.0,
+        error_if_nonfinite: bool = True,
+    ) -> None:
+        """Clip gradients by norm."""
+
+        rank_zero_warn("Gradient Clipping by Norm is currently experimental for FSDP. Proceed with Caution!")
```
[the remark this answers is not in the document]
justusschock: This warning is basically here, because we haven't yet figured out a proper way to test fsdp. So even though we make the calls as they should be, we cannot guarantee correctness here, which is why I added this warning.

tests/tests_fabric/strategies/test_single_device.py
justusschock: The issue here is, that with calling backward in an fp16 setting, the scaling-factor may be too large so that it produces an overflow.

In following steps the scaler will then automatically reduce this factor until no more overflow occurs so it shouldn't run endlessly.

src/lightning/fabric/strategies/fsdp.py
justusschock, line 205:
```suggestion
```

tests/tests_fabric/strategies/test_single_device.py
carmocca:
```suggestion
            except RuntimeError:
                # nonfinite grads -> skip and continue
                # this may repeat until the sacler finds a factor where overflow is avoided, so the while loop should eventually break
                pass
```

src/lightning/fabric/plugins/precision/precision.py
```diff
@@ -97,6 +97,9 @@ def main_params(self, optimizer: Optimizer) -> _PARAMETERS:
         for group in optimizer.param_groups:
             yield from group["params"]
 
+    def unscale_gradients(self, optimizer: Optimizer) -> None:
```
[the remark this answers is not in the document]
carmocca: Okay. I'm fine with this, but if we start adding adding logic like this that leaks to the base interface, we should then write an `UnscalingPrecision` plugin

carmocca (approved):

src/lightning/fabric/plugins/precision/amp.py
```diff
@@ -93,3 +93,20 @@ def _autocast_context_manager(self) -> torch.autocast:
         # the dtype could be automatically inferred but we need to manually set it due to a bug upstream
         # https://github.com/pytorch/pytorch/issues/67233
         return torch.autocast(self.device, dtype=torch.bfloat16 if self.precision == "bf16-mixed" else torch.half)
+
+    def unscale_gradients(self, optimizer: Optimizer) -> None:
+        scaler = getattr(self, "scaler", None)
```
carmocca, line 98: It's always defined

```suggestion
        scaler = self.scaler
```

PR #16733, opened 2023-02-12
Generalize `Optimizer` validation to accommodate both FSDP 1.x and 2.x

### LINKED ISSUE #16732: Generalize `Optimizer` validation to accommodate both FSDP 1.x and 2.0
### Generalize `Optimizer` validation to accommodate both FSDP 1.x and 2.x

In preparing the downstream package I maintain ([finetuning-scheduler](https://github.com/speediedan/finetuning-scheduler)) to support PyTorch 2.0's version of FSDP, I noticed that Lightning has yet to be adapted to accommodate FSDP 2.x's useful `use_orig_params` params mode. 

Currently, false-positive "The optimizer does not seem to reference any FSDP parameters" `ValueError`s will be triggered when the `use_orig_params` FSDP configuration is used because such a configuration does not require `FlatParameter` instances be present in the `Optimizer`. 

https://github.com/Lightning-AI/lightning/blob/c35fee92f4ffdcfe5137c762f614b44ca7f8b7d0/src/lightning/fabric/strategies/fsdp.py#L344-L347

https://github.com/Lightning-AI/lightning/blob/c35fee92f4ffdcfe5137c762f614b44ca7f8b7d0/src/lightning/pytorch/strategies/fsdp.py#L282-L287

FSDP versions `1.13` and `2.0` both ensure FSDP flattened parameters have the `_fsdp_flattened` attribute set, allowing for a single check to work across both versions (this is the check used internally by PyTorch [FSDP](https://github.com/pytorch/pytorch/blob/54ad459fb021c378dc950863c790f7a85c887f0b/torch/distributed/fsdp/_common_utils.py#L178-L181)). For BC, we need to keep the existing `FlatParameter`-based check for versions below `1.13`, but we should be able to remove that version guard once `1.12` is deprecated.

```python
def _optimizer_has_flat_params(optimizer: Optimizer) -> bool:
    if _TORCH_GREATER_EQUAL_1_13:
        return any(getattr(param, _FSDP_FLATTENED, False) for param in optimizer.param_groups[0]["params"])
    else:
        from torch.distributed.fsdp import FlatParameter

        return any(isinstance(param, FlatParameter) for param in optimizer.param_groups[0]["params"])
```

I'm sure support for this (along with other FSDP 2.x features) will be arriving as the team's priorities dictate but I'm hoping this issue and the PR I'll be opening momentarily will allow PyTorch 2.x development/testing to continue for `use_orig_params` in a BC-compatible manner while also adding some useful 1.x and 2.x FSDP tests.

### Pitch

The PR I'll be opening for this feature makes 4 minor changes:

1. Generalizes `_optimizer_has_flat_params` to accommodate FSDP 1.x and 2.0 usage. 
	- To avoid directly depending upon PyTorch's protected `_is_fsdp_flattened` method and protected constant (which have both changed locations between versions), I've included the relevant constant and attribute check in `fabric/strategies/fsdp.py` directly rather than relying on a protected attribute whose location is version dependent.
2. Adds an FSDP 2.0 compatible `auto_wrap_policy` test
3. Adds an FSDP 2.0 `_FSDPPolicy`-based test that uses `use_orig_params` to validate the multi-version `Optimizer` validation while also testing the preferred  `_FSDPPolicy`-based  auto-wrapping approach moving forward
4. Guards the 1.x compatible `auto_wrap_policy` behind the necessary `max_torch` mark

### Alternatives

One could use an FSDP instance-based check but that seems both more complicated and brittle. The `_fsdp_flattened` attribute is used extensively internally and was the best candidate attribute for this check IMHO.

### Additional context

_No response_

cc @borda @awaelchli @carmocca

speediedan:
## What does this PR do?

Resolves #16732

### Generalize `Optimizer` validation to accommodate both FSDP 1.x and 2.x

In preparing the downstream package I maintain ([finetuning-scheduler](https://github.com/speediedan/finetuning-scheduler)) to support PyTorch 2.0's version of FSDP, I noticed that Lightning has yet to be adapted to accommodate FSDP 2.x's useful `use_orig_params` params mode. 

Currently, false-positive "The optimizer does not seem to reference any FSDP parameters" `ValueError`s will be triggered when the `use_orig_params` FSDP configuration is used because such a configuration does not require `FlatParameter` instances be present in the `Optimizer`. 

https://github.com/Lightning-AI/lightning/blob/c35fee92f4ffdcfe5137c762f614b44ca7f8b7d0/src/lightning/fabric/strategies/fsdp.py#L344-L347

https://github.com/Lightning-AI/lightning/blob/c35fee92f4ffdcfe5137c762f614b44ca7f8b7d0/src/lightning/pytorch/strategies/fsdp.py#L282-L287

FSDP versions `1.13` and `2.0` both ensure FSDP flattened parameters have the `_fsdp_flattened` attribute set, allowing for a single check to work across both versions (this is the check used internally by PyTorch [FSDP](https://github.com/pytorch/pytorch/blob/54ad459fb021c378dc950863c790f7a85c887f0b/torch/distributed/fsdp/_common_utils.py#L178-L181)). For BC, we need to keep the existing `FlatParameter`-based check for versions below `1.13`, but we should be able to remove that version guard once `1.12` is deprecated.

```python
def _optimizer_has_flat_params(optimizer: Optimizer) -> bool:
    if _TORCH_GREATER_EQUAL_1_13:
        return any(getattr(param, _FSDP_FLATTENED, False) for param in optimizer.param_groups[0]["params"])
    else:
        from torch.distributed.fsdp import FlatParameter

        return any(isinstance(param, FlatParameter) for param in optimizer.param_groups[0]["params"])
```

I'm sure support for this (along with other FSDP 2.x features) will be arriving as the team's priorities dictate but I'm hoping this PR will allow PyTorch 2.x development/testing to continue for `use_orig_params` in a BC-compatible manner while also adding some useful 1.x and 2.x FSDP tests.

This PR makes 4 minor changes:

1. Generalizes `_optimizer_has_flat_params` to accommodate FSDP 1.x and 2.0 usage. 
	- To avoid directly depending upon PyTorch's protected `_is_fsdp_flattened` method and protected constant (which have both changed locations between versions), I've included the relevant constant and attribute check in `fabric/strategies/fsdp.py` directly rather than relying on a protected attribute whose location is version dependent.
2. Adds an FSDP 2.0 compatible `auto_wrap_policy` test
3. Adds an FSDP 2.0 `_FSDPPolicy`-based test that uses `use_orig_params` to validate the multi-version `Optimizer` validation while also testing the preferred  `_FSDPPolicy`-based  auto-wrapping approach moving forward
4. Guards the 1.x compatible `auto_wrap_policy` behind the necessary `max_torch` mark

Tested with: PyTorch 1.12.1, 1.13.1 and 2.0.0.dev20230211

#### Noted Alternatives

One could use an FSDP instance-based check but that seems both more complicated and brittle. The `_fsdp_flattened` attribute is used extensively internally and was the best candidate attribute for this check IMHO.

<details>

- [X] Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

</details>

## PR review

<details>

</details>


src/lightning/fabric/strategies/fsdp.py
```diff
@@ -49,6 +49,7 @@
     )
 
 _FSDP_ALIASES = ("fsdp", "fsdp_cpu_offload")
+_FSDP_FLATTENED = "_fsdp_flattened"
```
carmocca, line 52: What's the benefit of making this a global if it's only used in `_optimizer_has_flat_params`?

src/lightning/fabric/strategies/fsdp.py
```diff
@@ -342,6 +343,9 @@ def _init_cpu_offload(cpu_offload: Optional[Union[bool, "CPUOffload"]]) -> "CPUO
 
 
 def _optimizer_has_flat_params(optimizer: Optimizer) -> bool:
-    from torch.distributed.fsdp import FlatParameter
+    if _TORCH_GREATER_EQUAL_1_13:
```
carmocca, line 346: CHANGELOG mentions 2.x, but here you are checking 1.13. Is this an issue on 2.0+ or on 1.13+? Should it be using `_TORCH_GREATER_EQUAL_2_0` instead?

tests/tests_pytorch/strategies/test_fsdp.py
```diff
@@ -117,17 +125,18 @@ def _run_multiple_stages(trainer, model, model_path: Optional[str] = None):
 
     _assert_save_equality(trainer, model_path, cls=model.__class__)
 
-    # Test entry point
-    trainer.test(model)  # model is wrapped, will not call `configure_sharded_model`
+    with torch.inference_mode():
```
carmocca, line 128: This context manager is used automatically by the test and predict loops. Did you add this because something was failing?

src/lightning/fabric/strategies/fsdp.py
speediedan: My thought was that since it was a widely used attribute, it might be worth making accessible to other functions since other functions will likely use it in the future and it would prevent the need to move it. I'm happy to move it to make it function scoped if you think that's better though.

src/lightning/fabric/strategies/fsdp.py
speediedan: While this is an issue on 2.x, this function runs in the context of 1.12+. The attribute being checked is only present in 1.13+ though so for version 1.12 we need to retain the `FlatParameter` based version of the check.

tests/tests_pytorch/strategies/test_fsdp.py
speediedan: The context manager used automatically by the test and predict loops doesn't enter until invoked in [`run_stage`](https://github.com/speediedan/lightning/blob/1d5dabd61f5ea2d782f88c28c68094693a0e5d19/src/pytorch_lightning/trainer/trainer.py#L884-L885)
but with 2.x FSDP in auto-wrap with non-`use_orig_params` mode, a `_pre_load_state_dict_hook` is invoked during model restoration (before the Lightning nograd context manager is entered) which fails trying to use an inference tensor in non-inference mode. 

I'll have to dig into a bit more deeply to verify if it's a bug on the PyTorch side but until that is resolved (or a fix deemed necessary within Lightning which IMHO would be done in another PR), the outer context manager I added was necessary for the 2.0 non-`use_orig_params` auto-wrap test . `use_orig_params` does not trigger the error since it doesn't call
[`torch/distributed/fsdp/flat_param.py:1401`  ](https://github.com/pytorch/pytorch/blob/1f7448eeda824de2c15181997f16376803578529/torch/distributed/fsdp/flat_param.py#L1401)

An alternate solution would be to temporarily comment out the 2.0 auto-wrap test until the issue is addressed, we could keep the rest of the tests without the outer context manager.

FYI, here's the relevant portion of the call-stack:

```
split (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/_tensor.py:804)
split (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/functional.py:189)
_get_unflat_views (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/flat_param.py:1364)
_use_unsharded_views (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/flat_param.py:1384)
unflatten_as_params (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/flat_param.py:1536)
__enter__ (/opt/miniconda/envs/lightning_dev/lib/python3.10/contextlib.py:135)
_unflatten_as_params (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/_unshard_param_utils.py:116)
__enter__ (/opt/miniconda/envs/lightning_dev/lib/python3.10/contextlib.py:135)
enter_context (/opt/miniconda/envs/lightning_dev/lib/python3.10/contextlib.py:492)
_unshard_fsdp_state_params (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/_unshard_param_utils.py:224)
__enter__ (/opt/miniconda/envs/lightning_dev/lib/python3.10/contextlib.py:135)
_enter_unshard_params_ctx (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/_state_dict_utils.py:109)
_full_pre_load_state_dict_hook (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/_state_dict_utils.py:332)
_pre_load_state_dict_hook (/opt/miniconda/envs/lightning_dev/lib/python3.10/site-packages/torch/distributed/fsdp/_state_dict_utils.py:697)
...
```
You can see FSDP pre_load_state_dict hook unsharding the flattened parameters in the FSDP module (to prepare for loading the unflattened state_dict), but when it goes to register an unflattened view into the unsharded flat_param, the view it returns is in inference mode. If you debug at the top of the provided stack, you can see the call that returns the inference tensor:
```
inf_mode_tensors = torch._VF.split_with_sizes(self, split_size, dim)
inf_mode_tensors[0]._version
  File "<string>", line 1, in <module>
RuntimeError: Inference tensors do not track version counter.
```
There are a number of ways we could work around this if it isn't going to be addressed on the PyTorch side, but IMHO they'd be best tackled in a separate PR. For the sake of this PR, it makes sense to me to either comment out the 2.0 auto-wrap test for now or to leave in the torch.inference_mode context manager but add the relevant TODO followup possibly pointing to a new PyTorch issue. Happy to help out with any other approaches you suggest though. Thanks again for the review!

speediedan: I had chance today to dig into PyTorch's implementation of the `torch.inference_mode` context manager to verify the behavior that had necessitated the outer `torch.inference_mode` context manager in these tests is not a bug but rather the intended behavior.

When the first test stage execution https://github.com/Lightning-AI/lightning/blob/c35fee92f4ffdcfe5137c762f614b44ca7f8b7d0/tests/tests_pytorch/strategies/test_fsdp.py#L121 begins, within the new ``_no_grad_context`` ``inference_mode`` context manager:
https://github.com/Lightning-AI/lightning/blob/c35fee92f4ffdcfe5137c762f614b44ca7f8b7d0/src/lightning/pytorch/loops/utilities.py#L175, FSDP [initializes the flat parameters](https://github.com/pytorch/pytorch/blob/a17a7ccc92144452e0fe51e02f21f1f1ba88118a/torch/distributed/fsdp/flat_param.py#L826-L831).

```
init_flat_param_attributes (/home/speediedan/repos/pytorch/torch/distributed/fsdp/flat_param.py:826)
decorate_context (/home/speediedan/repos/pytorch/torch/utils/_contextlib.py:115)
_share_state_and_init_handle_attrs (/home/speediedan/repos/pytorch/torch/distributed/fsdp/_runtime_utils.py:258)
_lazy_init (/home/speediedan/repos/pytorch/torch/distributed/fsdp/_runtime_utils.py:174)
_root_pre_forward (/home/speediedan/repos/pytorch/torch/distributed/fsdp/_runtime_utils.py:516)
forward (/home/speediedan/repos/pytorch/torch/distributed/fsdp/fully_sharded_data_parallel.py:735)
_call_impl (/home/speediedan/repos/pytorch/torch/nn/modules/module.py:1501)
test_step (/home/speediedan/repos/lightning/src/lightning/pytorch/strategies/fsdp.py:353)
```
At ``L826``, we still have a normal tensor from the existing model:

```python
flat_param._full_param_padded._version
0
```

Diving into the ``inference_mode`` [behavior](https://github.com/pytorch/pytorch/blob/a17a7ccc92144452e0fe51e02f21f1f1ba88118a/c10/core/TensorImpl.cpp#L167-L182):

Yes we're in inference mode...
```cpp
Thread 1 "python" hit Breakpoint 5, c10::TensorImpl::TensorImpl (this=0x7f3a530, storage=..., key_set=..., data_type=..., 
    device_opt=...) at /home/speediedan/repos/pytorch/c10/core/TensorImpl.cpp:167
167   if (inference_mode) {
(gdb) s
```

So we remove any autograd and ADInPlace/View dispatch keys from our dispatch key set...
```cpp
Thread 1 "python" hit Breakpoint 6, c10::TensorImpl::TensorImpl (this=0x7f3a530, storage=..., key_set=..., data_type=..., 
    device_opt=...) at /home/speediedan/repos/pytorch/c10/core/TensorImpl.cpp:171
171     key_set_ = key_set - c10::autograd_dispatch_keyset_with_ADInplaceOrView;
```

and don't include autograd keys that we normally would
```cpp
    key_set_ = key_set | getAutogradRelatedKeySetFromBackend(k);
  }
```

After this is done, the fact that we have no autograd or ADInplace/View keys (no more autograd and view tracking overhead!)...
```cpp
  c10::TensorImpl::is_inference (this=0x7f3a530) at /home/speediedan/repos/pytorch/c10/core/TensorImpl.h:1219
1219        bool no_Autograd = !key_set_.has_any(c10::autograd_dispatch_keyset);
(gdb) s
1220        TORCH_INTERNAL_ASSERT_DEBUG_ONLY(
(gdb) s
1223        return no_ADInplaceOrView && no_Autograd;
(gdb) s
1224      }
(gdb) p no_ADInplaceOrView
$3 = true
(gdb) p no_ADInplaceOrView && no_Autograd
$4 = true
```
... signifies that this new ``_full_param_padded`` tensor is an inference tensor and will remain so (technically, the absence of a ``version_counter_`` is part of the definition as well). Finally, we then also skip configuring a ``version_counter_`` which is a way to detect the inference tensor on the python side and also imposes constraints on inference tensors once exiting inference mode:
```cpp
  if (!is_inference()) {
    version_counter_ = VariableVersion(/*version=*/0);
  }
```
So when we return to [``L831``](https://github.com/pytorch/pytorch/blob/a17a7ccc92144452e0fe51e02f21f1f1ba88118a/torch/distributed/fsdp/flat_param.py#L826-L831), ``_full_param_padded`` is an inference tensor... 

```python
flat_param._full_param_padded._version
  File "<string>", line 1, in <module>
RuntimeError: Inference tensors do not track version counter.
```
... and we'll hit the error we get when creating a ``Parameter`` from a view based on it as this same tensor is used in the next test stage during the ``pre_load_state_dict`` hook. 

Given the described behavior, I think we should either keep the outer ``inference_mode`` guard that is currently in this PR or add a minor refactoring (like below) so that it is included for only the tests that require it (that trigger the `pre_load_state_dict` hook on existing inference tensors).
```python
    # Test entry point
    trainer.test(model)  # model is wrapped, will not call `configure_sharded_model`

    # Predict entry point
    trainer.predict(model)  # model is wrapped, will not call `configure_sharded_model`

    with torch.inference_mode():

        trainer.test(ckpt_path=model_path)

        trainer.predict(ckpt_path=model_path)
```

Let me know which direction you think we should go (or if you would like another approach).

At some point in a separate PR, it might be worth offering the user the configuration option to use the less-restrictive ``no_grad`` guard for the test and predict stage loops in case they want to avoid nested ``inference_mode`` guards for cases like this.

I know you are swamped, thanks again for the review!
@carmocca @awaelchli

tests/tests_pytorch/strategies/test_fsdp.py
```diff
@@ -221,19 +261,38 @@ def custom_auto_wrap_policy(
...
+            marks=RunIf(min_torch="2.0.0"),
+        ),
+        pytest.param(
+            TestFSDPModelAutoWrapped(),
+            FSDPStrategy,
+            {"auto_wrap_policy": custom_fsdp_policy, "use_orig_params": True},
+            marks=RunIf(min_torch="2.0.0"),
+        ),
     ],
+    ids=["manually_wrapped", "autowrap_1x", "autowrap_2x", "autowrap_use_orig_params"],
```
awaelchli, line 286: Minor nit: if specifying ids, then maybe let's move them to the pytest.param() above so it's easier to read/track.

tests/tests_pytorch/strategies/test_fsdp.py
```diff
@@ -142,6 +151,37 @@ def _assert_save_equality(trainer, ckpt_path, cls=TestFSDPModel):
...
+def custom_auto_wrap_policy(
+    module,
+    recurse,
+    unwrapped_params: int,
+    min_num_params: int = int(1e8),
+) -> bool:
+    return unwrapped_params >= 2
+
+
+def custom_auto_wrap_policy_2x(
```
awaelchli, line 163: Nit: can we name it more explicitly like `_torch_gte_2_0`

src/lightning/fabric/strategies/fsdp.py
awaelchli: I would remove it since it is only used in that one function. But either way is fine with me too

awaelchli (approved):
Thanks for the patience on this!

tests/tests_pytorch/strategies/test_fsdp.py
```diff
@@ -142,6 +151,37 @@ def _assert_save_equality(trainer, ckpt_path, cls=TestFSDPModel):
             assert torch.equal(ddp_param.float().cpu(), shard_param)
 
 
+def custom_auto_wrap_policy(
```
awaelchli, line 154: could we move the definition of these to above the test_fsdp_checkpoint_multi_gpus function where they are used?

tests/tests_pytorch/strategies/test_fsdp.py
speediedan: Kept the new conditional definition from [16764](https://github.com/Lightning-AI/lightning/pull/16764).

src/lightning/fabric/strategies/fsdp.py
speediedan: Narrowed the ``_fsdp_flattened`` suffix definition to be function scoped as suggested.

PR #16747, opened 2023-02-13
CI: Update colossalai version

awaelchli:
## What does this PR do?

See title. The version we use is no longer on pypi: https://pypi.org/project/colossalai/#history
We are now setting the minimum version to 0.2.0, which drops a few lines of old code that was guarded for the older version.

An unrelated deepspeed error popped up and we temporarily downgrade it. See the fix here:
https://github.com/microsoft/DeepSpeed/issues/2697


.azure/ipu-tests.yml
```diff
@@ -77,6 +77,12 @@ jobs:
           python ./requirements/pytorch/adjust-versions.py $fpath; \
         done
         pip install -e .[dev]
+
+        # TODO: Remove this uninstall command
+        # https://github.com/microsoft/DeepSpeed/commit/b587c7e85470329ac25df7c7c2521ff9b2833db7
+        # gets released
+        pip uninstall deepspeed
```
carmocca, line 84: I suggest doing this instead, so that deepspeed is not even installed

```suggestion
        pip install -e .[extra,examples]
```

.azure/ipu-tests.yml
```diff
@@ -76,7 +76,8 @@ jobs:
         for fpath in `ls requirements/**/*.txt`; do \
           python ./requirements/pytorch/adjust-versions.py $fpath; \
         done
-        pip install -e .[dev]
+        pip install -e .[extra,examples]
```
carmocca, line 79:
```suggestion
        pip install -e .[extra,examples,test]
```

requirements/pytorch/strategies.txt
```diff
@@ -1,3 +1,4 @@
 # NOTE: the upper bound for the package version is only set for CI stability, and it is dropped while installing this package
 #  in case you want to preserve/enforce restrictions on the latest compatible version, add "strict" as an in-line comment
 deepspeed>=0.6.0, <=0.8.0
```
carmocca, line 3: Do you think the deepspeed issue you saw would also impact our users? If yes, we should also do

```suggestion
deepspeed>=0.6.0, <0.8.0
```

carmocca (approved):

akihironitta (approved):

.azure/ipu-tests.yml
```diff
@@ -76,7 +76,8 @@ jobs:
         for fpath in `ls requirements/**/*.txt`; do \
           python ./requirements/pytorch/adjust-versions.py $fpath; \
         done
-        pip install -e .[dev]
+        pip install -e .[extra,examples,test]
```
carmocca, line 79: Looks like we should also avoid installing the strategies for the tpu tests

carmocca:
I don't understand why these deepspeed errors appear here and not in master. Both are using `0.8.0`

.azure/gpu-tests-pytorch.yml
```diff
@@ -108,19 +108,20 @@ jobs:
...
-        CUDA_VERSION_MM_COLOSSALAI=$(python -c "import torch ; print(''.join(map(str, torch.version.cuda)))")
-        CUDA_VERSION_COLOSSALAI=$(python -c "print([ver for ver in [11.3, 11.1] if $CUDA_VERSION_MM_COLOSSALAI >= ver][0])")
-        pip install "colossalai==0.1.12+torch${PYTORCH_VERSION}cu${CUDA_VERSION_COLOSSALAI}" --find-links https://release.colossalai.org
-
         pip install -r requirements/pytorch/strategies.txt --find-links ${TORCH_URL}
 
+        # TODO: Lift the restriction after
+        # https://github.com/microsoft/DeepSpeed/commit/b587c7e85470329ac25df7c7c2521ff9b2833db7
+        # gets released
+        pip install "deepspeed<0.8.0"
```
Borda, line 123: lets pin it rather here: https://github.com/Lightning-AI/lightning/pull/16747/files#r1105178273

Borda (approved):

tests/tests_pytorch/strategies/test_colossalai.py
```diff
@@ -231,18 +231,26 @@ def test_multi_gpu_checkpointing(tmpdir):
         precision=16,
         strategy="colossalai",
         callbacks=[ck],
+        num_sanity_val_steps=0,
```
awaelchli, line 234: This and the test below (note the xfail) are due to a breaking change in the new colossal version: You can't run test/val after fit due to the requirement of the warmup phase. 

`AssertionError: You should run a completed iteration as your warmup iter`

We can fix this or improve the error messaging in a follow up.

tests/tests_pytorch/strategies/test_colossalai.py
```diff
@@ -231,18 +231,26 @@ def test_multi_gpu_checkpointing(tmpdir):
...
+def test_test_without_fit(tmpdir):
     model = ModelParallelClassificationModel()
+    dm = ClassifDataModule()
     trainer = Trainer(default_root_dir=tmpdir, accelerator="gpu", devices=2, precision=16, strategy="colossalai")
-    saved_results = trainer.test(model, datamodule=dm, ckpt_path=ck.best_model_path)
-    assert saved_results == results
+
+    # Colossal requires warmup, you can't run validation/test without having fit first
+    # This is a temporary limitation
+    # Raises: `AssertionError: You should run a completed iteration as your warmup iter`
```
carmocca, line 252: You can put this "match" in the xfail raises

.azure/ipu-tests.yml
awaelchli: I knew all this is a trap...
There is no indication whatsoever in the tpu-tests.yml that we are installing anything.

carmocca: It's here: https://github.com/Lightning-AI/lightning/blob/fbbbbf6ad7aef07f418df198ffde26276419fdd6/dockers/base-xla/tpu_workflow_pytorch.jsonnet#L45

dockers/base-xla/tpu_workflow_pytorch.jsonnet
```diff
@@ -42,7 +42,7 @@ local tputests = base.BaseTest {
       for fpath in `ls requirements/**/*.txt`; do
         python requirements/pytorch/adjust-versions.py $fpath {PYTORCH_VERSION};
       done
-      PACKAGE_NAME=pytorch pip install .[dev]
+      PACKAGE_NAME=pytorch pip install .[extra,examples,test]
```
carmocca, line 45: TPUs don't run any examples

```suggestion
      PACKAGE_NAME=pytorch pip install .[extra,test]
```

carmocca (approved):
Pushed a commit to ignore the standalone false positive

PR #16751, opened 2023-02-14
fix warning so the user has a clear next step

nohalon:
tell the user exactly how to install `lightning-api-access` directly from the warning

cc @borda

### DIFF
--- a/src/lightning/app/components/serve/auto_scaler.py
+++ b/src/lightning/app/components/serve/auto_scaler.py
@@ -453,7 +453,7 @@ class _LoadBalancer(LightningWork):
         try:
             from lightning_api_access import APIAccessFrontend
         except ModuleNotFoundError:
-            logger.warn("APIAccessFrontend not found. Please install lightning-api-access to enable the UI")
+            logger.warn("APIAccessFrontend not found. To enable the UI run: pip install lightning-api-access")
             return
 
         if is_running_in_cloud():
--- a/src/lightning/app/components/serve/auto_scaler.py
+++ b/src/lightning/app/components/serve/auto_scaler.py
@@ -453,7 +453,9 @@ class _LoadBalancer(LightningWork):
         try:
             from lightning_api_access import APIAccessFrontend
         except ModuleNotFoundError:
-            logger.warn("Some dependencies to run the UI are missing. To resolve, run `pip install lightning-api-access`")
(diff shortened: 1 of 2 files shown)


src/lightning/app/components/serve/auto_scaler.py
```diff
@@ -453,7 +453,7 @@ def _get_endpoint_info_page(self) -> Optional["APIAccessFrontend"]:  # noqa: F82
         try:
             from lightning_api_access import APIAccessFrontend
         except ModuleNotFoundError:
-            logger.warn("APIAccessFrontend not found. Please install lightning-api-access to enable the UI")
+            logger.warn("APIAccessFrontend not found. To enable the UI run: pip install lightning-api-access")
```
tchaton, line 456:
```suggestion
            logger.warn("You are missing some dependencies. Run `pip install lightning-api-access` to resolve the UI.")
```

tchaton (approved):
LGTM !

src/lightning/app/components/serve/python_server.py
```diff
@@ -293,7 +293,7 @@ def configure_layout(self) -> Optional["Frontend"]:
         try:
             from lightning_api_access import APIAccessFrontend
         except ModuleNotFoundError:
-            logger.warn("APIAccessFrontend not found. Please install lightning-api-access to enable the UI")
+            logger.warn("APIAccessFrontend not found. To enable the UI run: pip install lightning-api-access")
```
awaelchli, line 296:
```suggestion
            logger.warn("APIAccessFrontend not found. To enable the UI, run: pip install lightning-api-access")
```
Not sure, but I think the comma is missing.

awaelchli (approved):

src/lightning/app/components/serve/auto_scaler.py
nohalon, line 456:
```suggestion
            logger.warn("Some dependencies to run the UI are missing. To resolve, run `pip install lightning-api-access`")
```

src/lightning/app/components/serve/python_server.py
nohalon, line 296:
```suggestion
            logger.warn("Some dependencies to run the UI are missing. To resolve, run `pip install lightning-api-access`")
```

Issue #4206, opened 2020-10-17
Add warning if {}_idx is not present for multiple val dataloaders / optimisers


tchaton:
### Motivation

Currently, if a user uses several val data loaders and forgot to add dataloader_idx, they will get the following error:
```
  File "/home/some/folder/miniconda3/envs/mac_slurmPL1/lib/python3.8/site-packages/pytorch_lightning/overrides/data_parallel.py", line 182, in forward
    output = self.module.validation_step(*inputs[0], **kwargs[0])
TypeError: validation_step() takes 3 positional arguments but 4 were given
```
From Slack.
https://pytorch-lightning.slack.com/archives/CRBLFHY79/p1602879838197200

### Pitch

Add a warning which checks the functions signature and validate the number of right number of arguments.
If not, add warning indicating `dataloader_idx` should be provided.

sanjeevk-os:
@tchaton I would like to work on this issue. Thanks.

bipinKrishnan:
I am a new contributor, can I provide any help in fixing this issue @tchaton

carmocca:
@SkafteNicki I unassigned you since you closed your PR, but if you still want to finish it, feel free to assign yourself back!

PR #16753, opened 2023-02-14
Make dataloader_idx optional for batch start/end hooks

carmocca:
## What does this PR do?

When you don't use multiple dataloaders, `training_step` does not take a `dataloader_idx` argument.
However, this is not the case for `on_{validation,test,predict}_batch_{start,end}`.

This PR makes it optional for them too.

This is a breaking change, as the users who define this argument and don't use multiple dataloaders now need to set `dataloader_idx=0` in their signature, or remove it if they don't need it.

### DIFF
--- a/src/lightning/app/cli/pl-app-template/core/callbacks.py
+++ b/src/lightning/app/cli/pl-app-template/core/callbacks.py
@@ -110,7 +110,7 @@ class PLAppProgressTracker(Callback):
         outputs: Any,
         batch: Any,
         batch_idx: int,
-        dataloader_idx: int,
+        dataloader_idx: int = 0,
     ) -> None:
         self._state.metrics = self._progress_bar_metrics(trainer, pl_module)
         current = self._test_batch_idx(trainer)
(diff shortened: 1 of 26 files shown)


awaelchli:
Does this close the age old https://github.com/Lightning-AI/lightning/issues/4206?

awaelchli (approved):

carmocca:
> Does this close the age old #4206 as well?

No it doesn't.

awaelchli:
Why not, the error there was 

TypeError: validation_step() takes 3 positional arguments but 4 were given

which you would get when not passing the arg. But you are making that optional now.

carmocca:
The linked issue references `validation_step`, which this PR does not touch. `dataloader_idx` is already optional for that method. As I understand, that PR suggest that we add validation logic at the beginning of the run to check whether `*_step` methods have this argument to avoid the Python exception. With the changes in this PR, the `on_{validation,test,predict}_batch_{start,end}` hooks would need this validation too.

PR #16757, opened 2023-02-15
Replace ColossalAIStrategy with external implementation

awaelchli:
## What does this PR do?

The Colossal AI Strategy moves to an external repo where it will be maintained: https://github.com/Lightning-AI/lightning-colossalai/pull/1. A corresponding package will be published to PyPi from which users can install the integration. 

- [x] Replaces the implementation with the one in the lightning-colossalai package. 
- [x] Registeres the strategy="colossal" as a strategy if the package is installed
- [x] Produces an error message if you request the strategy but don't have the package installed

**Before:**

```py
trainer = Trainer(strategy="colossalai")

# OR

from lightning.pytorch.strategies import ColossalAIStrategy
trainer = Trainer(strategy=ColossalAIStrategy(...))
```

**Now:**

```bash
pip install lightning-colossalai
```

```py
trainer = Trainer(strategy="colossalai")  # only works if  `lightning-colossalai` is installed.

# OR

from lightning_colossalai import ColossalAIStrategy
trainer = Trainer(strategy=ColossalAIStrategy(...))
```

Follow up:
- Update docs once the lightning-colossalai package is finalized and public

cc @borda @justusschock


requirements/pytorch/strategies.txt
```diff
@@ -2,4 +2,4 @@
 #  in case you want to preserve/enforce restrictions on the latest compatible version, add "strict" as an in-line comment
 
 deepspeed>=0.6.0, <0.8.0  # TODO: Include 0.8.x after https://github.com/microsoft/DeepSpeed/commit/b587c7e85470329ac25df7c7c2521ff9b2833db7 gets released
-colossalai>=0.2.0, <=0.2.4
+# https://${GITHUB_TOKEN}@github.com/Lightning-AI/lightning-colossalai@poc-dont-touch-this-branch-thank-you
```
awaelchli, line 5: TODO: Just here for debugging. Remove

justusschock: I like that branch name :D

justusschock (approved):
lgtm. Will the external strategies be registered to pypi?

Borda (approved):

awaelchli:
> lgtm. Will the external strategies be registered to pypi?
That's the plan yes. Right now there is a chicken and egg problem but once the two PRs land we can complete the integration.

src/lightning/pytorch/callbacks/gradient_accumulation_scheduler.py
```diff
@@ -121,17 +122,21 @@ def on_train_start(self, trainer: "pl.Trainer", pl_module: "pl.LightningModule")
...
-        from lightning.pytorch.strategies import ColossalAIStrategy, DeepSpeedStrategy
+        from lightning.pytorch.strategies import DeepSpeedStrategy
 
-        unsupported_strategies = (DeepSpeedStrategy, ColossalAIStrategy)
         unsupported_accelerators = (IPUAccelerator,)
+        unsupported_strategies = [DeepSpeedStrategy]
+        if _LIGHTNING_COLOSSALAI_AVAILABLE:
+            from lightning_colossalai import ColossalAIStrategy
+
+            unsupported_strategies.append(ColossalAIStrategy)
```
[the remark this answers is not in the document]
awaelchli: We decoupled the callback from the trainer/strategy in #16729. I would prefer to keep it that way so that it doesn't matter where the callback lives.

src/lightning/pytorch/trainer/connectors/accelerator_connector.py
```diff
@@ -206,6 +206,12 @@ def _check_config_and_set_final_flags(
         if strategy is not None:
             self._strategy_flag = strategy
 
+        if strategy == "colossalai" and not _LIGHTNING_COLOSSALAI_AVAILABLE:
+            raise ModuleNotFoundError(
+                "The `lightning-colossalai` package is required to run the ColossalAI strategy."
+                " HINT: Try running `pip install -U lightning-colossalai`"
```
carmocca, line 212: This could be

```suggestion
                str(_LIGHTNING_COLOSSALAI_AVAILABLE)
```

src/lightning/pytorch/utilities/imports.py
```diff
@@ -32,6 +32,7 @@
 _RICH_AVAILABLE = package_available("rich") and compare_version("rich", operator.ge, "10.2.2")
 _TORCH_QUANTIZE_AVAILABLE = bool([eg for eg in torch.backends.quantized.supported_engines if eg != "none"])
 _TORCHVISION_AVAILABLE = RequirementCache("torchvision")
+_LIGHTNING_COLOSSALAI_AVAILABLE = package_available("lightning_colossalai")
```
carmocca, line 35: RequirementCache plz?

carmocca (approved):

src/lightning/pytorch/utilities/imports.py
awaelchli: It wasn't working when I was tinkering with the package. Let's try again.

src/lightning/pytorch/trainer/connectors/accelerator_connector.py
awaelchli: Fine with me. The error message will be more generic and the test will become less nice but it works.

PR #16760, opened 2023-02-15
Use the local batch_idx to update the progress bar

carmocca:
## What does this PR do?

Part of https://github.com/Lightning-AI/lightning/pull/16726. That PR will require using these over the progress tracking values.

The unused batch_idx properties are removed.

Fixes https://github.com/Lightning-AI/lightning/issues/12901


src/lightning/pytorch/callbacks/progress/rich_progress.py
```diff
@@ -485,7 +485,7 @@ def on_predict_batch_start(
     def on_train_batch_end(
         self, trainer: "pl.Trainer", pl_module: "pl.LightningModule", outputs: STEP_OUTPUT, batch: Any, batch_idx: int
     ) -> None:
-        self._update(self.train_progress_bar_id, self.train_batch_idx)
+        self._update(self.train_progress_bar_id, batch_idx + 1)
```
awaelchli, line 488: Have you checked that this is resuming the bar correctly when resuming from a checkpoint?

src/lightning/pytorch/callbacks/progress/tqdm_progress.py
```diff
@@ -190,7 +192,6 @@ def init_train_tqdm(self) -> Tqdm:
         """Override this to customize the tqdm bar for training."""
         bar = Tqdm(
             desc=self.train_description,
-            initial=self.train_batch_idx,
```
awaelchli, line 193: What will we do about this? Don't you need to add an update to the on_train_start hooks etc?

carmocca: I don't think this is necessary:
- `on_train_epoch_start` already sets initial=0.
- When any batch runs, the `n` will be updated
- The other bars did not set it.

src/lightning/pytorch/callbacks/progress/rich_progress.py
carmocca: Why wouldn't it? As long as the batch_idx is correct, this would be too. I don't think we have any tests of this specifically for the tqdm bar, but we have for the loops, and this index comes directly from the loops.

If you have a specific suggestion, I'd be happy to write a test for it

awaelchli: It looks good. Yes, we could rely on the tests but there can be visual glitches with the bar that are hard to test. There are countless regressions in the past. Which is why I prefer to do a visual inspection on top of the existing testing we have. 

The `initial=x` didn't seem to do much. Both on master and on this branch, the bar doesn't actually take the proper initial value until the first batch has run through. I checked using this script and saving a mid-epoch checkpoint:

```py
import os
from time import sleep

import torch
from lightning.pytorch.callbacks import ModelCheckpoint

from lightning.pytorch import LightningModule, Trainer


class RandomDataset(Dataset):
        self.len = length

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


    def __init__(self):
        super().__init__()

    def forward(self, x):
        return self.layer(x)

        print("observe state of progress on train start")
        sleep(5)

        loss = self(batch).sum()
        self.log("train_loss", loss)
        if batch_idx == 1 and self.current_epoch == 1:
            print("saving mid epoch ckpt")
            self.trainer.save_checkpoint("mid-epoch.ckpt")
        sleep(1)
        return {"loss": loss}

        loss = self(batch).sum()
        sleep(0.5)
        self.log("valid_loss", loss)

        loss = self(batch).sum()
        sleep(0.5)
        self.log("test_loss", loss)

def run():

    checkpoint = ModelCheckpoint(dirpath=".", save_top_k=-1)

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=3,
        limit_val_batches=4,
        num_sanity_val_steps=10,
        max_epochs=3,
        enable_model_summary=False,
        callbacks=checkpoint,
    )

    print("resuming")

    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=3,
        limit_val_batches=4,
        num_sanity_val_steps=10,
        max_epochs=3,
        enable_model_summary=False,
        callbacks=checkpoint,
    )
    # trainer.fit(model, train_dataloaders=train_data, val_dataloaders=val_data, ckpt_path="epoch=1-step=6.ckpt")
    trainer.fit(model, train_dataloaders=train_data, val_dataloaders=val_data, ckpt_path="mid-epoch.ckpt")


if __name__ == "__main__":
    run()

```

src/lightning/pytorch/callbacks/progress/tqdm_progress.py
awaelchli: See comment above.

awaelchli (approved):

justusschock (approved):

PR #16764, opened 2023-02-15
PL: Test PyTorch 2.0 pre-release on CPU and CUDA

justusschock:
## What does this PR do?

Adds CI for PT 2.0 pre release for CPU and GPU

A follow-up will do the same for Fabric


.github/workflows/ci-tests-pytorch.yml
```diff
@@ -61,6 +61,10 @@ jobs:
           - {os: "macOS-11", pkg-name: "pytorch", python-version: "3.8", pytorch-version: "1.13"}
           - {os: "ubuntu-20.04", pkg-name: "pytorch", python-version: "3.8", pytorch-version: "1.13"}
           - {os: "windows-2022", pkg-name: "pytorch", python-version: "3.8", pytorch-version: "1.13"}
+          # pytorch 2.9 Tests (currently pre-release, change this once it's released)
```
akihironitta, line 64:
```suggestion
          # pytorch 2.0 Tests (currently pre-release, change this once it's released)
```

akihironitta (approved):

.github/workflows/ci-tests-pytorch.yml
```diff
@@ -87,13 +91,21 @@ jobs:
       run: bash .actions/pull_legacy_checkpoints.sh
 
     - name: Adjust PyTorch versions in requirements files
-      if: ${{ matrix.requires != 'oldest' }}
+      if: ${{ matrix.requires != 'oldest' }} && ${{ matrix.release != 'pre' }}
```
Borda, line 94:
```suggestion
      if: ${{ matrix.requires != 'oldest' && matrix.release != 'pre' }}
```

dockers/base-cuda/Dockerfile
```diff
@@ -21,6 +21,7 @@ FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu${UBUNTU_VERSION}
 ARG PYTHON_VERSION=3.9
 ARG PYTORCH_VERSION=1.13
 ARG MAX_ALLOWED_NCCL=2.16.2
+ARG PRE_RELEASE
```
Borda, line 24: I would not include pre option to dockers, lets keep it simple :)

justusschock: This is still left from when I wanted to follow another path for GPU CI.

requirements/pytorch/adjust-versions.py
```diff
@@ -29,34 +28,40 @@ def find_latest(ver: str) -> Dict[str, str]:
...
     return req
 
 
 if __name__ == "__main__":
-    if len(sys.argv) == 3:
-        requirements_path, torch_version = sys.argv[1:]
-    else:
-        requirements_path, torch_version = sys.argv[1], None
-    print(f"requirements_path='{requirements_path}' with torch_version='{torch_version}'")
+    import argparse
```
carmocca, line 53: Can you explain why the changes to this file are necessary?

Can't we just skip the step where they are initially fixed, instead of removing them again? I feel like this is a difficult solution in the code to what would be simple in CI

.github/workflows/ci-tests-pytorch.yml
```diff
@@ -61,6 +61,10 @@ jobs:
           - {os: "macOS-11", pkg-name: "pytorch", python-version: "3.8", pytorch-version: "1.13"}
           - {os: "ubuntu-20.04", pkg-name: "pytorch", python-version: "3.8", pytorch-version: "1.13"}
           - {os: "windows-2022", pkg-name: "pytorch", python-version: "3.8", pytorch-version: "1.13"}
+          # pytorch 2.0 Tests (currently pre-release, change this once it's released)
+          - {os: "macOS-11", pkg-name: "lightning", python-version: "3.9", pytorch-version: "2.0", release: "pre"}
+          - {os: "ubuntu-20.04", pkg-name: "lightning", python-version: "3.9", pytorch-version: "2.0", release: "pre"}
+          - {os: "windows-2022", pkg-name: "lightning", python-version: "3.9", pytorch-version: "2.0", release: "pre"}
```
carmocca, line 67: These should be added to the checkgroup config

.azure/gpu-tests-pytorch.yml
```diff
@@ -54,6 +54,10 @@ jobs:
           image: "pytorchlightning/pytorch_lightning:base-cuda-py3.9-torch1.13-cuda11.7.1"
           scope: ""
           PACKAGE_NAME: "pytorch"
+        'PyTorch | nightly':
+          image: "pytorchlightning/pytorch_lightning:base-cuda-py3.9-torch1.13-cuda11.7.1"
```
carmocca, line 58: Why not create a 2.0 docker image? We'll need it later anyways

Borda: That sounds good to me =)

justusschock: I wanted to but @Borda didn't like changes to the dockerfile

.github/workflows/ci-tests-pytorch.yml
justusschock: shouldn't we wait with this until they have an official rc?

requirements/pytorch/adjust-versions.py
justusschock: This is necessary since a won't be selected if there is any pinned version (even with `--pre`). So I still had to remove any version specifiers.

justusschock: I'll revert this once I have a working docker image

.azure/gpu-tests-pytorch.yml
Borda: True, I did revert it, but maybe it would be good call; sorry for moving there and back...

Borda: as it is solution just for next a few weeks I would stick with the simple way and don't build special images as that would also mean additional adjustments on workflow side so the RC are not overwritten due to `FREEZE_REQUIREMENTS`

Borda (approved):
LGTM from the CI stand of view :tada:

.github/workflows/ci-tests-pytorch.yml
```diff
@@ -87,13 +93,21 @@ jobs:
...
         cat requirements/pytorch/base.txt
 
+    # - name: Adjust PyTorch versions in requirements files
+    #   if: ${{matrix.release == 'pre'}}
+    #   run: |
+    #     for fpath in `ls requirements/**/*.txt`; do \
+    #       python ./requirements/pytorch/adjust-versions.py $fpath ${{ matrix.pytorch-version }} --remove_torch_ver; \
+    #     done
+    #     cat requirements/pytorch/base.txt
+
```
justusschock, line 110:
```suggestion
```

awaelchli (approved):

dockers/base-cuda/Dockerfile
```diff
@@ -91,18 +93,18 @@ RUN \
...
-    pip install -r requirements/pytorch/base.txt --no-cache-dir --find-links https://download.pytorch.org/whl/cu${CUDA_VERSION_MM}/torch_stable.html && \
+    pip install -r requirements/pytorch/base.txt --no-cache-dir --find-links ${TORCH_URL} ${TORCH_WHEEL_OPTIONS} && \
     rm assistant.py
 
 
 RUN \
     # install rest of strategies
     cat requirements/pytorch/strategies.txt && \
-    pip install -r requirements/pytorch/devel.txt -r requirements/pytorch/strategies.txt --no-cache-dir --find-links https://download.pytorch.org/whl/cu${CUDA_VERSION_MM}/torch_stable.html
+    pip install -r requirements/pytorch/devel.txt --no-cache-dir --find-links ${TORCH_URL}
```
carmocca, line 107: Where did the strategy install go?

dockers/base-cuda/Dockerfile
```diff
@@ -91,18 +93,18 @@ RUN \
     pip config set global.cache-dir false && \
     # set particular PyTorch version \
     for fpath in `ls requirements/**/*.txt`; do \
-      python ./requirements/pytorch/adjust-versions.py $fpath ${PYTORCH_VERSION}; \
+        python ./requirements/pytorch/adjust-versions.py $fpath ${PYTORCH_VERSION} ; \
     done && \
 
     # Install base requirements \
-    pip install -r requirements/pytorch/base.txt --no-cache-dir --find-links https://download.pytorch.org/whl/cu${CUDA_VERSION_MM}/torch_stable.html && \
+    pip install -r requirements/pytorch/base.txt --no-cache-dir --find-links ${TORCH_URL} ${TORCH_WHEEL_OPTIONS} && \
```
carmocca, line 100: This is unused?

```suggestion
    pip install -r requirements/pytorch/base.txt --no-cache-dir --find-links ${TORCH_URL} && \
```

dockers/base-cuda/Dockerfile
```diff
@@ -91,18 +93,18 @@ RUN \
     pip config set global.cache-dir false && \
     # set particular PyTorch version \
     for fpath in `ls requirements/**/*.txt`; do \
-      python ./requirements/pytorch/adjust-versions.py $fpath ${PYTORCH_VERSION}; \
+        python ./requirements/pytorch/adjust-versions.py $fpath ${PYTORCH_VERSION} ; \
```
carmocca, line 96: Dockers use 2 spaces

```suggestion
      python ./requirements/pytorch/adjust-versions.py $fpath ${PYTORCH_VERSION} ; \
```

.github/workflows/ci-tests-pytorch.yml
carmocca: It's out now

.azure/gpu-tests-pytorch.yml
```diff
@@ -102,13 +106,21 @@ jobs:
         for fpath in `ls requirements/**/*.txt`; do \
           python ./requirements/pytorch/adjust-versions.py $fpath ${PYTORCH_VERSION}; \
         done
+      condition: ne(variables['scope'], 'future')
       displayName: 'Adjust dependencies'
 
-    - bash: pip install -e .[extra,test,examples] --find-links ${TORCH_URL}
+    - bash: pip install -e .[extra,test,examples] -U --find-links ${TORCH_URL} ${INSTALL_FLAG}
```
carmocca, line 112: Unused?

```suggestion
    - bash: pip install -e .[extra,test,examples] -U --find-links ${TORCH_URL}
```

carmocca:
Looks like `onnx` is now a required dependency for the onnx tests.

.github/workflows/ci-tests-pytorch.yml
justusschock: yes, as @Borda shared on slack, we would merge it without adding it here, to first make sure they pass and once we're confident it works as it should, we would add it to checkgroup.
This PR only adds CI but does not aim at fixing issues.

carmocca: Please don't merge this with failing jobs, even if they are unmarked as required. CI should always be green.

I'm fine with fixing the issues separately, but then I would fix those first. You can work here to validate the changes and pull them to a separate PR once they are passing

dockers/base-cuda/Dockerfile
Borda:
```suggestion
    pip install -r requirements/pytorch/devel.txt -r requirements/pytorch/strategies.txt --no-cache-dir --find-links ${TORCH_URL}
```

.azure/gpu-tests-pytorch.yml
Borda: can be dropped

.github/workflows/ci-tests-pytorch.yml
Borda: > Please don't merge this with failing jobs, even if they are unmarked as required. CI should always be green.

ls help us to make it green :)

requirements/pytorch/base.txt
```diff
@@ -10,3 +10,4 @@ torchmetrics>=0.7.0, <0.10.1  # needed for using fixed compare_version
 packaging>=17.1, <=23.0
 typing-extensions>=4.0.0, <=4.4.0
 lightning-utilities>=0.6.0.post0, <0.7.0
+onnx<1.14.0
```
carmocca, line 13: Why not keep this in the test requirements?

Borda:
seems some size leaking:
```
>       assert cuda_memory < model.start_cuda_memory
E       assert 17039872 < 2048
E        +  where 2048 = BoringModelGPU(\n  (layer): Linear(in_features=32, out_features=2, bias=True)\n).start_cuda_memory
```

requirements/pytorch/base.txt
Borda: becase it is needed for PT 2.0, they have it as optional package so it crash with error you need to install it :)

Borda: and in tets we still keep runitme version... :chipmunk:

carmocca (changes requested):
Please don't merge this until 2.0 failures are resolved

requirements/pytorch/base.txt
carmocca: I don't think that's right. The onnx installation should still be optional: https://github.com/Lightning-AI/lightning/pull/16891

Borda: ok, lets merge #16764 here instead of master :otter:

tests/tests_pytorch/strategies/test_ddp.py
```diff
@@ -56,22 +56,6 @@ def test_multi_gpu_model_ddp_fit_test(tmpdir):
...
-    model = BoringModel()
-    trainer = Trainer(
-        default_root_dir=tmpdir,
-        fast_dev_run=True,
-        strategy=DDPStrategy(process_group_backend="undefined"),
-        accelerator="cuda",
-        devices=2,
-        logger=False,
-    )
-    with pytest.raises(ValueError, match="Invalid backend: 'undefined'"):
```
carmocca, line 71: PyTorch 2.0 changed the backend validation logic, and this is now hanging instead of raising the ValueError

Either way, we can remove this test because the exception is raised inside Torch, so we don't need to test that

carmocca (approved):
All green now

PR #16726, opened 2023-02-11
Sequential `CombinedLoader` to flatten the eval and predict loops

carmocca:
## What does this PR do?

- Removes the `_EvaluationEpochLoop`
- Removes the `_PredictionEpochLoop`
- Removes the `_DataLoaderLoop`
- Moves the `_EvaluationLoop` out of the dataloader directory
- Moves the `_PredictionLoop` out of the dataloader directory
- Adds a `loop.setup_data()` method to the top-level loops. This method is called at the beginning of `loop.run()`. This replaces the `trainer.reset_*_dataloader` methods.
- Removes the `trainer.reset_*_dataloader` methods and related trainer.py simplifications
- The top-level loops now own their `_DataLoaderSource` and `CombinedLoader`
- The `trainer.*_dataloader` properties now return what the user returned in their `Module.*_dataloader()` hook. Nit: validation dataloaders are casted to list, see follow-up section
- Adds support for `predict_step(dataloader_iter)`. this is not an important objective of the PR, but it's implemented for free.

Fixes https://github.com/Lightning-AI/lightning/issues/10696
Fixes https://github.com/Lightning-AI/lightning/issues/11845

Follow-ups:

- Refactor the `setup_data` implementations
- Reduce the data connector
- Move the `num_*_batches` ownership to the loops
- `Trainer.evaluation_dataloaders` should not be a forced list
- Allow passing CombinedLoaders, remove trainer flag for the mode
- Add a checkpoint migration for the loop state
- Review docs

cc @justusschock @awaelchli @borda @carmocca

Issue #10696, opened 2021-11-23
Properly support Iterables as an alternative to DataLoaders in `Trainer.{fit, validate, ...}`


awaelchli:
## 🚀 Feature

1. Support passing dataloaders of type `Iterable` to `Trainer.{fit, validate, test, predict}`
2. Support returning dataloaders of type `Iterable` in `LightningModule.*_dataloder()` and `LightningDataModule.*_dataloder()`

### Motivation

Lightning today supports iterating over a DataLoader (when creating an iterator from it) or multiple dataloaders. Some claims have been made in the past that the Trainer also supports `Iterable`, but this is not true today because a trivial example can demonstrate that it fails. In any case, our `Trainer` methods as well as the `Lightning{Data}Module` methods do not reflect this in their signature types.
Some applications require operating directly on an instance of Iterable and wrapping them inside a torch DataLoader/Dataset would not be desirable or feasible. An example from the medical domain: https://github.com/MIC-DKFZ/batchgenerators 

Furthermore, the introduction of Iterables would most likely facilitate the integration of [torch/data DataPipes](https://github.com/pytorch/data) which are soon to be included in PyTorch.

### Pitch

Support the Iterable type. 

Usage example: 

```py
trainer = Trainer()
model = Model()
data = [batch0, batch1, batch2]  # a list of batches
trainer.fit(model, train_dataloaders=data)
```

Several places inside [data_loading.py](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/trainer/data_loading.py) would have to be updated with branching logic that excludes setup steps normally done for DataLoaders (see limitations section below).

### Limitations and Challenges

Some features in Lightning will not work for Iterables. Most notably:
- Automatic replacement of samplers for distributed data-parallel: Samplers do not apply here for Iterable datastructures. However, the user will still have to take care of correctly returning the data in case of distributed training.
- Fault-tolerant training: Not supported out of the box. The user may have to switch to manual-mode fault-tolerant training #10568 
- Ambiguity of the input type for `Trainer.fit()`: The .fit() method and it's sisters currently support these types: 
   https://github.com/PyTorchLightning/pytorch-lightning/blob/48cf1adfd3ad9c7e659083a4afc334dafb331f28/pytorch_lightning/utilities/types.py#L36-L43
   However, if we now add `Iterable`, we run into an ambiguous case because `Dict` and `Sequence` are subtypes of `Iterable`. 

### Alternatives

[...]

PR #16775, opened 2023-02-15
Run XLA's dataloader validation per dataloader

carmocca:
## What does this PR do?

Part of #10696 
Part of #16726

The validation logic for dataloaders should be applied to a single dataloader.

### DIFF
--- a/src/lightning/fabric/strategies/xla.py
+++ b/src/lightning/fabric/strategies/xla.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import io
 import os
-from typing import Any, Dict, List, Mapping, Optional, Sequence, TYPE_CHECKING, Union
+from typing import Any, Dict, List, Optional, TYPE_CHECKING, Union
 
 import torch
 from torch import Tensor
@@ -30,7 +30,6 @@ from lightning.fabric.plugins.precision import Precision
 from lightning.fabric.strategies import ParallelStrategy
 from lightning.fabric.strategies.launchers.xla import _XLALauncher
 from lightning.fabric.strategies.strategy import TBroadcast
-from lightning.fabric.utilities.apply_func import apply_to_collection
 from lightning.fabric.utilities.data import has_len
 from lightning.fabric.utilities.rank_zero import rank_zero_only
 from lightning.fabric.utilities.types import _PATH, ReduceOp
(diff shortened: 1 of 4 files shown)


tests/tests_pytorch/strategies/test_tpu_spawn.py
```diff
@@ -45,39 +44,10 @@ def predict_dataloader(self):
...
-    "train_dataloaders, val_dataloaders, test_dataloaders, predict_dataloaders",
-    [
-        (_loader_no_len, None, None, None),
-        (None, _loader_no_len, None, None),
-        (None, None, _loader_no_len, None),
-        (None, None, None, _loader_no_len),
-        (None, [_loader, _loader_no_len], None, None),
-    ],
-)
-def test_error_iterable_dataloaders_passed_to_fit(
```
carmocca, line 58: This test will be added back in https://github.com/Lightning-AI/lightning/pull/16726/files#diff-6aedcbda38b84cd2299e850355bf10dccc690b4bc71fa8774410269ee56d1bf5L58. Fixing it here would require major changes that will land later.

The test below still tests that the exception is raised.

And the test that will be added asserts that `strategy.process_dataloader` is called as many times as we have dataloaders

src/lightning/pytorch/strategies/tpu_spawn.py
```diff
@@ -98,34 +94,14 @@ def root_device(self) -> torch.device:
...
-                TPUSpawnStrategy._validate_dataloader(source.instance)
+    def _validate_dataloader(dataloader: object) -> None:
+        if not has_len(dataloader):
+            raise TypeError(
+                "TPUs do not currently support IterableDataset objects, the dataset must implement `__len__`."
+                " HINT: You can mock the length on your dataset to bypass this error."
+            )
 
     def connect(self, model: "pl.LightningModule") -> None:
-        TPUSpawnStrategy._validate_patched_dataloaders(model)
```
carmocca, line 128: This validation will be run by the loops as they setup their dataloaders. The strategy is not responsible for it, just needs to do it in `process_dataloader`

src/lightning/fabric/strategies/xla.py
```diff
@@ -222,12 +221,9 @@ def _set_world_ranks(self) -> None:
...
     @staticmethod
-    def _validate_dataloader(dataloaders: DataLoader) -> None:
-        def check_has_len(dataloader: DataLoader) -> None:
-            if not has_len(dataloader):
-                raise TypeError(
-                    "TPUs do not currently support IterableDataset objects, the dataset must implement `__len__`."
-                    " HINT: You can mock the length on your dataset to bypass this MisconfigurationException."
-                )
-
-        apply_to_collection(dataloaders, dtype=object, wrong_dtype=(Sequence, Mapping), function=check_has_len)
```
awaelchli, line 233: This makes sense. In fabric, this was redundant. This was overlooked when adapting the PL code to the fabric strategy.

awaelchli (approved):

justusschock (approved):

PR #16780, opened 2023-02-16
Make DDP subprocess the default launcher for multi-device

awaelchli:
## What does this PR do?

When using multiple devices, the strategy now defaults to "ddp" instead of "ddp_spawn" when none is set.

Before:
```py
# in non-interactive env, implies strategy=ddp_spawn
# in Jupyter notebooks, implies strategy=ddp_notebook
trainer = Trainer(devices=2)
```

Now:
```py
# in non-interactive env, implies strategy=ddp
# in Jupyter notebooks, implies strategy=ddp_notebook
trainer = Trainer(devices=2)

# Still works:
trainer = Trainer(devices=2, strategy="ddp_spawn")
```

DDP-spawn is good for debugging and testing, but it has "invisible" process boundaries and especially beginner users don't know that. Since the processes join after `Trainer.fit()` etc., the state is not updated (only the model params are transferred), and this can lead to unexpected behavior. Very early on in Lightnings life, spawn also used to work in notebooks which was the primary reason why it was chosen as default. But support for that was later dropped in PyTorch and we introduced ddp-fork as an alternative later on. 

Over time we started discouraging ddp-spawn in docs, tutorials etc. and always promoted explicitly setting strategy=ddp. With Lightning 2.0, we are now ready to switch fully over to ddp as the default multi-device strategy for small to medium sized models. 


carmocca (approved):

docs/source-pytorch/accelerators/gpu_intermediate.rst
```diff
@@ -196,13 +196,9 @@ Comparison of DDP variants and tradeoffs
      - No
      - Yes
      - No
-   * - Is the guard ``if __name__=="__main__"`` required?
```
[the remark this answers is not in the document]
awaelchli: I wrote this, but I found it misleading reading this again. Technically it is possible to not specify `if __name__=="__main__"` for the fork method but then if you use DataLoader num_workers > 0 it's anyway needed. So overall, this is not so helpful to have in the docs. 

But the docs change here is completely unrelated to the rest of the PR, I just stumbled over this and felt it's ok to squeeze this in.

justusschock (approved):

src/lightning/pytorch/trainer/connectors/accelerator_connector.py
```diff
@@ -458,12 +458,9 @@ def _choose_strategy(self) -> Union[Strategy, str]:
...
             # TODO: lazy initialized device, then here could be self._strategy_flag = "single_device"
             return SingleDeviceStrategy(device=device)  # type: ignore
-        if len(self._parallel_devices) > 1:
-            if _IS_INTERACTIVE:
-                return "ddp_fork"
-            return "ddp_spawn"
-
-        return DDPStrategy.strategy_name
+        if len(self._parallel_devices) > 1 and _IS_INTERACTIVE:
+            return "ddp_fork"
```
Borda, line 462: in past we had a few silly bugs with typos, so how about using enums internally?

Borda (approved):

src/lightning/pytorch/trainer/connectors/accelerator_connector.py
awaelchli: We had enums in the past and we decided to remove them for better readability. Also, since then we introduced the strategy registry. We have many tests that, so if there was a typo, they would fail.

PR #16748, opened 2023-02-14
Remove DP

justusschock:
## What does this PR do?

Removes DataParallel as it is [discouraged by PyTorch](https://github.com/pytorch/pytorch/issues/65936) and requires a lot of workarounds due to a different data-flow.

<details>
  <summary><b>Before submitting</b></summary>

- Was this **discussed/agreed** via a GitHub issue? (not for typos and docs)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/lightning/blob/master/.github/CONTRIBUTING.md), **Pull Request** section?
- [x] Did you make sure your **PR does only one thing**, instead of bundling different changes together?
- Did you make sure to **update the documentation** with your changes? (if necessary)
- Did you write any **new necessary tests**? (not for typos and docs)
- [x] Did you verify new and **existing tests pass** locally with your changes?
- Did you list all the **breaking changes** introduced by this pull request?
- Did you **update the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

</details>

## PR review

Anyone in the community is welcome to review the PR.
Before you start reviewing, make sure you have read the [review guidelines](https://github.com/Lightning-AI/lightning/wiki/Review-guidelines). In short, see the following bullet-list:

<details>
  <summary>Reviewer checklist</summary>

- [ ] Is this pull request ready for review? (if not, please submit in draft mode)
- [ ] Check that all items from **Before submitting** are resolved
- [ ] Make sure the title is self-explanatory and the description concisely explains the PR
- [ ] Add labels and milestones (and optionally projects) to the PR so it can be classified

</details>

cc @justusschock @awaelchli @borda

PR #16791, opened 2023-02-17
Remove the `*_step_end` hooks

awaelchli:
## What does this PR do?

The `training_step_end`, `validation_step_end`, and `test_step_end` are no longer required after the removal of DP in #16748. Since DP does not exist anymore, there won't be a use case anymore to reduce the tensors from training_step. For any other post-training step processing of outputs, the user can override the `on_*_batch_end` hooks.

Part of the Lightning 2.0 simplification initiative.


src/lightning/pytorch/strategies/xla.py
```diff
@@ -225,18 +225,6 @@ def predict_step(self, *args: Any, **kwargs: Any) -> STEP_OUTPUT:
         with self.precision_plugin.predict_step_context():
             return self.model(*args, **kwargs)
 
-    def training_step_end(self, output: STEP_OUTPUT) -> STEP_OUTPUT:
-        self._pod_progress_bar_force_stdout()
-        return output
```
awaelchli, line 230: TODO: where could I move this 🤔

tests/tests_pytorch/loops/test_evaluation_loop_flow.py
```diff
@@ -74,63 +71,6 @@ def backward(self, loss):
...
-                out = ["1", 2, torch.tensor(2)]
-            if batch_idx > 0:
-                out = {"something": "random"}
-            self.last_out = out
-            return out
-
-        def validation_step_end(self, out):
-            self.validation_step_end_called = True
-            assert self.last_out == out
-            return out
```
awaelchli, line 99: Since this does not apply anymore, the test would be come equivalent to the one above. So removing it.

tests/tests_pytorch/loops/test_fetchers.py
```diff
@@ -156,50 +156,6 @@ def measure() -> float:
...
-
-class RandomIndicesDataset(Dataset):
-    def __getitem__(self, index):
-        return torch.randint(EMB_DIM, [BATCH_SIZE])
-
-    def __len__(self):
-        return 16
-
-
-class RecommenderModel(BoringModel):
```
awaelchli, line 171: This was unused. cc @carmocca

tests/tests_pytorch/loops/test_training_loop_flow_scalar.py
```diff
@@ -123,62 +118,6 @@ def backward(self, loss):
 
     # make sure correct steps were called
     assert model.training_step_called
-    assert not model.training_step_end_called
```
awaelchli, line 126: GitHub shows these lines as deleted but that's just an illusion because they are duplicated in the test below and so the diff is just shifted.

justusschock (approved):

src/lightning/pytorch/strategies/xla.py
justusschock: couldn't this go to on_training_batch_start in the following step? It might be a bit off, but should be fine I think

ethanwharris (approved):
Approving app bit 😆

src/lightning/pytorch/core/hooks.py
```diff
@@ -137,7 +137,7 @@ def on_predict_batch_end(self, outputs: Optional[Any], batch: Any, batch_idx: in
         """Called in the predict loop after the batch.
 
         Args:
-            outputs: The outputs of predict_step_end(test_step(x))
+            outputs: The outputs of test_step(x)
```
carmocca, line 140:
```suggestion
            outputs: The outputs of predict_step(x)
```

src/lightning/pytorch/core/hooks.py
```diff
@@ -137,7 +137,7 @@ def on_predict_batch_end(self, outputs: Optional[Any], batch: Any, batch_idx: in
...
             batch: The batched data as it is returned by the test DataLoader.
```
carmocca, line 141:
```suggestion
            batch: The batched data as it is returned by the prediction DataLoader.
```

src/lightning/pytorch/core/module.py
```diff
@@ -762,9 +701,6 @@ def validation_step(self, *args: Any, **kwargs: Any) -> Optional[STEP_OUTPUT]:
             # pseudocode of order
```
carmocca, line 701: Remove this code-block entirely?

src/lightning/pytorch/strategies/hpu_parallel.py
```diff
@@ -123,16 +122,6 @@ def optimizer_step(
...
         return optimizer_output
 
-    def validation_step_end(self, step_output: STEP_OUTPUT) -> STEP_OUTPUT:
-        # Break lazy accumulation of graph after every step
-        htcore.mark_step()
-        return step_output
-
-    def test_step_end(self, step_output: STEP_OUTPUT) -> STEP_OUTPUT:
-        # Break lazy accumulation of graph after every step
-        htcore.mark_step()
```
carmocca, line 133: Isn't this still necessary?

src/lightning/pytorch/strategies/xla.py
```diff
@@ -299,3 +280,13 @@ def register_strategies(cls, strategy_registry: Dict) -> None:
             cls,
             description=f"{cls.__class__.__name__}",
         )
+
+    def _pod_progress_bar_force_stdout(self) -> None:
+        # Why is it required? The way `pytorch_xla.distributed` streams logs
```
carmocca, line 285: pre-existing issue: this should check whether a TQDMPbar is present and active

carmocca (approved):

src/lightning/pytorch/strategies/xla.py
awaelchli: I can't really do that right now, I'd need access to a trainer reference.

src/lightning/pytorch/strategies/hpu_parallel.py
awaelchli: No, since we don't go through the step_end code path anymore, so there aren't any graph operations there to mark a step for. 

mark_step is called also in on_after_backward and optimizer_step() hook as required by habana. See: https://docs.habana.ai/en/latest/PyTorch/PyTorch_Model_Porting/Porting_PyTorch_Models_to_Gaudi.html?highlight=mark_step#porting-pytorch-models-to-gaudi

Borda (approved):

PR #16795, opened 2023-02-17
Incorporate pytorch's fixes in device_count_nvml

### LINKED ISSUE #16590: `CUDAAccelerator` can not run on your system since the accelerator is not available.
### Bug description

So, in my environment `torch.cuda.is_available()` is `True` but `torch.cuda.device_count()` is `0`. This issue is probably linked with a [pytorch issue](https://github.com/pytorch/pytorch/issues/90543). Since I was planning on using lightning for a new project, I am unable to use GPU using the `pl.Trainer(accelerator='cuda', devices=1)`. 

Not sure if this is a bug on your end. Any suggestion to go about this would be great.

### How to reproduce the bug

_No response_

### Error messages and logs

```
# Error messages and logs here please
```

### Environment

<details>
  <summary>Current environment</summary>

```
#- Lightning Component (e.g. Trainer, LightningModule, LightningApp, LightningWork, LightningFlow):
#- PyTorch Lightning Version (e.g., 1.5.0):
#- Lightning App Version (e.g., 0.5.2):
#- PyTorch Version (e.g., 1.10):
#- Python version (e.g., 3.9):
#- OS (e.g., Linux):
#- CUDA/cuDNN version:
#- GPU models and configuration:
#- How you installed Lightning(`conda`, `pip`, source):
#- Running environment of LightningApp (e.g. local, cloud):
```

</details>

### More info

_No response_

cc @tchaton

awaelchli:
## What does this PR do?

Adopts the changes/fixes from https://github.com/pytorch/pytorch/commit/e7e51b3a5c1d134bb6334da49cbf89221a7a63a0

Fixes #16590


src/lightning/fabric/accelerators/cuda.py
```diff
@@ -181,62 +181,168 @@ def is_cuda_available() -> bool:
 
 
 # TODO: Remove once minimum supported PyTorch version is 1.13
```
carmocca, line 183:
```suggestion
# TODO: Remove once minimum supported PyTorch version is 2.0
```

carmocca: L164 now needs to check for 2_0

justusschock (approved):

src/lightning/fabric/accelerators/cuda.py
awaelchli: No, why? If every time there is a fix we need to bump the version, we will never be able to drop the code and rely on pytorch.

carmocca: These fixes that you are incorporating are only part of PyTorch 2.0

So we can only rely on pytorch after that. This is when our minimum supported PyTorch version is 2.0

Or am I missing something?

carmocca (approved):

PR #16799, opened 2023-02-17
add launch() call

rasbt:
Adding `fabric.launch()` here. Not sure if it is strictly required in this tutorial, but maybe it's best practice (so it would work if someone uses multiple GPUs). At least this way it's consistent with the overview page at https://pytorch-lightning.readthedocs.io/en/stable/fabric/fabric.html then.

### DIFF
--- a/docs/source-pytorch/fabric/fundamentals/convert.rst
+++ b/docs/source-pytorch/fabric/fundamentals/convert.rst
@@ -13,6 +13,7 @@ Here are five easy steps to let :class:`~lightning_fabric.fabric.Fabric` scale y
     from lightning.fabric import Fabric
 
     fabric = Fabric()
+    fabric.launch()
 
 **Step 2:** Call :meth:`~lightning_fabric.fabric.Fabric.setup` on each model and optimizer pair and :meth:`~lightning_fabric.fabric.Fabric.setup_dataloaders` on all your data loaders.
 


docs/source-pytorch/fabric/fundamentals/convert.rst
```diff
@@ -13,6 +13,7 @@ Here are five easy steps to let :class:`~lightning_fabric.fabric.Fabric` scale y
     from lightning.fabric import Fabric
 
     fabric = Fabric()
+    fabric.launch()
```
awaelchli, line 16: Thanks for looking at the docs @rasbt.

So adding it here would be somewhat wrong because in step 5 below, we launch using the command line utility:

`  lightning run model path/to/train.py`

There is "CLI launch", and there is "programmatic launch". They have intersecting functionality but have different use cases. This is explained in the launch distributed training section: https://pytorch-lightning.readthedocs.io/en/stable/fabric/fundamentals/launch.html#programmatic-launch

So you either you do launch through .launch() in the code (and run the script with python train.py), or you do it through `lightning run` command. Doing both will result in an error. 

If this is confusing, maybe we should consider removing it from the homepage instead of adding here. Or we could be putting a note here. Just an idea

rasbt: Oh, got it, I didn't see the `lightning run model path/to/train.py` below.

Tbh, wouldn't it be cleaner/less ambiguous to just have one way to do this, i.e., via always using `.launch()` and `python train.py`? This way, people who just want to use fabric don't have to remember the `lightning run model` command and can use their scripts as they usually do.

(But that would probably require ignoring `.launch` if someone runs it in a lightning app)

awaelchli: Ok, I can bring this up to Eden. Maybe it is confusing to others as well. 

But if we removed one, then we'd remove functionality that the other doesn't have, so I'm not sure where that'd leave us. Need to think about it.

rasbt: @awaelchli So, I would say always require `.launch()` whether you use it at 

    lightning run model path/to/train.py

or

    python path/to/train.py

It's one more line of code, but this way we can simplify the documentation and maybe it also prevents some weird bugs where users take `lightning run` code and try to run it via regular python where it then doesn't work.

awaelchli: Talked to Eden about this. We are considering to feature the non-CLI way as the primary way in all docs, and have the CLI as an optional, secondary way to run programs. This would require us to rewrite the launch.rst and getting started page a bit, but would probably overall work better. WDYT?

rasbt: I.e., including `.launch()`? I think that makes sense. This way users won't shoot themselves in the food when they call the code via `python` instead of `lightning` from the CLI or if they use Notebooks

PR #16800, opened 2023-02-17
Update

carmocca:
## What does this PR do?

Resolves these follow-ups from https://github.com/Lightning-AI/lightning/pull/16726

- Trainer.evaluation_dataloaders should not be a forced list
- Allow passing CombinedLoaders, remove trainer flag for the mode

And some TODOs I added.

### DIFF
--- a/src/lightning/pytorch/trainer/connectors/data_connector.py
+++ b/src/lightning/pytorch/trainer/connectors/data_connector.py
@@ -17,7 +17,6 @@ from dataclasses import dataclass, field
 from typing import Any, Iterable, List, Optional, Tuple, Union
 from weakref import proxy
 
-from lightning_utilities.core.apply_func import apply_to_collection
 from torch.utils.data import BatchSampler, DataLoader, Sampler, SequentialSampler
 from torch.utils.data.distributed import DistributedSampler
 
@@ -29,7 +28,7 @@ from lightning.pytorch.overrides.distributed import UnrepeatedDistributedSampler
 from lightning.pytorch.strategies import DDPSpawnStrategy
 from lightning.pytorch.trainer import call
 from lightning.pytorch.trainer.states import RunningStage, TrainerFn
-from lightning.pytorch.trainer.supporters import _LITERAL_SUPPORTED_MODES, CombinedLoader
+from lightning.pytorch.trainer.supporters import CombinedLoader
 from lightning.pytorch.utilities.data import _is_dataloader_shuffled, _update_dataloader, has_len_all_ranks
 from lightning.pytorch.utilities.exceptions import MisconfigurationException
 from lightning.pytorch.utilities.model_helpers import is_overridden
@@ -41,9 +40,8 @@ warning_cache = WarningCache()
 
 
 class DataConnector:
-    def __init__(self, trainer: "pl.Trainer", multiple_trainloader_mode: _LITERAL_SUPPORTED_MODES = "max_size_cycle"):
+    def __init__(self, trainer: "pl.Trainer"):
         self.trainer = trainer
-        self.multiple_trainloader_mode = multiple_trainloader_mode
         self._datahook_selector: Optional[_DataHookSelector] = None
 
     @property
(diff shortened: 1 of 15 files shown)


tests/tests_pytorch/trainer/test_supporters.py
```diff
@@ -339,11 +355,15 @@ def test_combined_data_loader_with_max_size_cycle_and_ddp(accelerator, replace_s
...
-        assert len(dataloader) == length // 2 if replace_sampler_ddp else length
+        assert len(combined_loader) == length
+
+        trainer._data_connector.attach_data(model, train_dataloaders=combined_loader)
+        trainer.fit_loop.setup_data(shuffle=False)
+
+        assert len(combined_loader) == length // 2 if replace_sampler_ddp else length
         if replace_sampler_ddp:
-            last_batch = list(dataloader)[-1]
+            print(list(combined_loader))
```
carmocca, line 365:
```suggestion
```

tests/tests_pytorch/callbacks/test_prediction_writer.py
```diff
@@ -108,38 +107,6 @@ def test_prediction_writer_batch_indices(num_workers):
     )
 
 
-def test_prediction_writer_partial_support_for_combined_loader():
-    """Test partial support for CombinedLoader: prediction works but sample indices don't get tracked."""
-
-    class PredictionModel(BoringModel):
-        def predict_dataloader(self):
-            # TODO(carmocca): this should work
```
carmocca, line 116: This is now supported. So the test is not necessary.

tests/tests_pytorch/models/test_cpu.py
```diff
@@ -239,14 +239,13 @@ def test_running_test_no_val(tmpdir):
     """
 
     class ModelTrainTest(BoringModel):
-        def val_dataloader(self):
-            pass
```
carmocca, line 243: This would be treated as `iterable=None`, which will fail as you'd expect

src/lightning/pytorch/trainer/connectors/data_connector.py
```diff
@@ -333,83 +318,80 @@ def _reset_eval_dataloader(
...
+        trainer = self.trainer
 
-        # TODO(carmocca): list conversion shouldn't be forced
-        if not isinstance(dataloaders, list):
-            dataloaders = [dataloaders]  # type: ignore[assignment]
+        if trainer.overfit_batches > 0:
+            self._resolve_overfit_batches(combined_loader, mode)
 
-        if any(dl is None for dl in dataloaders):
-            rank_zero_warn("One of given dataloaders is None and it will be skipped.")
```
carmocca, line 347: I didn't port this skip logic. I think it can lead to bugs where the user accidentally returns `None` and then training continues normally when it should fail.

tests/tests_pytorch/trainer/test_dataloaders.py
```diff
@@ -174,19 +175,27 @@ def test_dataloaders_passed_to_fn(tmpdir, ckpt_path, n):
...
     trainer = Trainer(default_root_dir=tmpdir, max_epochs=1, limit_val_batches=0.1, limit_train_batches=0.2)
-    trainer.fit(model, train_dataloaders=train_dataloaders, val_dataloaders=eval_dataloaders)
 
-    assert trainer.state.finished, f"Training failed with {trainer.state}"
-    assert len(trainer.val_dataloaders) == n
+    trainer.fit(model, train_dataloaders=train_dataloaders, val_dataloaders=eval_dataloaders)
+    if n > 1:
+        assert len(trainer.val_dataloaders) == n
+    else:
+        assert isinstance(trainer.val_dataloaders, DataLoader)
```
awaelchli, line 183: Can you document this change in the changelog?

awaelchli (approved):

tests/tests_pytorch/trainer/test_dataloaders.py
carmocca: It's already mentioned: "- The `trainer.*_dataloader` properties now return what the user returned in their `LightningModule.*_dataloader()` hook ([#16726](https://github.com/Lightning-AI/lightning/pull/16726), [#16800](https://github.com/Lightning-AI/lightning/pull/16800))"

src/lightning/pytorch/trainer/connectors/data_connector.py
```diff
@@ -333,83 +318,80 @@ def _reset_eval_dataloader(
...
+        dataloaders = []
+        for i, dl in enumerate(combined_loader.flattened):
+            is_shuffled = _is_dataloader_shuffled(dl)
+            # limit this warning only for samplers assigned automatically when shuffle is set
+            if is_shuffled:
+                rank_zero_warn(
+                    f"Your `{mode.dataloader_prefix}_dataloader`'s sampler has shuffling enabled,"
+                    " it is strongly recommended that you turn shuffling off for val/test/predict dataloaders.",
+                    category=PossibleUserWarning,
+                )
```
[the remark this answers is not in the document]
carmocca: This needs to compute `is_shuffled` before for so that we pass it to `_prepare_dataloader`, so what was `_check_eval_shuffling` before would now just be a wrapper for the `rank_zero_warn` (it would also compute `is_shuffled` before). Which I thought was unnecessary to have separate as a staticmethod.

justusschock (approved):

PR #16803, opened 2023-02-18
Resolve FitLoop setter TODOs

carmocca:
## What does this PR do?

See title. Two setters are removed

### DIFF
--- a/src/lightning/pytorch/trainer/setup.py
+++ b/src/lightning/pytorch/trainer/setup.py
@@ -72,7 +72,7 @@ def _init_debugging_flags(
 
         trainer.limit_test_batches = num_batches
         trainer.limit_predict_batches = num_batches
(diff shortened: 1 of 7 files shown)


src/lightning/pytorch/trainer/setup.py
```diff
@@ -72,7 +72,7 @@ def _init_debugging_flags(
 
         trainer.limit_test_batches = num_batches
         trainer.limit_predict_batches = num_batches
-        trainer.fit_loop.max_steps = num_batches
+        trainer.fit_loop.epoch_loop.max_steps = num_batches
```
awaelchli, line 75: We can do this because the value gets already checked on line 54. So the setter is redundant.

awaelchli (approved):

PR #16809, opened 2023-02-19
Merge DDPStrategy and DDPSpawnStrategy in PL

awaelchli:
## What does this PR do?

Follows the same motivations as #14952. At the beginning, even prior to launchers being introduced, the two implementations were very different and had to be clearly separated. Over time, strategies became more modular and the two grew closer in functionality. Today, they only differ in the way processes are launched. 

This refactor joins them together to remove the code duplication. 


tests/tests_pytorch/strategies/test_ddp_spawn_strategy.py
```diff
@@ -20,7 +20,7 @@
 
 from lightning.pytorch import LightningModule, Trainer
```
awaelchli, line 21: I will rename/merge and deduplicate the test files in a follow up. To simplify the work of the reviewer

src/lightning/pytorch/strategies/ddp.py
```diff
@@ -129,8 +138,10 @@ def process_group_backend(self) -> Optional[str]:
 
     def _configure_launcher(self) -> None:
         assert self.cluster_environment is not None
-        if not self.cluster_environment.creates_processes_externally:
```
awaelchli, line 132: This gets checked inside the `_SubprocessScriptLauncher`

awaelchli:
@carmocca @justusschock This is now ready for review :)

src/lightning/pytorch/strategies/ddp.py
```diff
@@ -347,23 +362,38 @@ def post_training_step(self) -> None:
...
+        for name, start_method in entries:
+            strategy_registry.register(
+                name,
+                cls,
+                description=f"DDP strategy with `start_method` '{start_method}'",
+                start_method=start_method,
+            )
+
+        entries = (
+            ("ddp_find_unused_parameters_false", False, "popen"),
```
Borda, line 380: can we have it rather as dict instead of args?

Borda (approved):

justusschock (approved):

carmocca (approved):
Nicely done

src/lightning/pytorch/strategies/ddp.py
awaelchli: I tried making it a dict, but I'm not sure how this should look like. It could be a tuple of dicts or a dict of dicts. I think the current way is pretty readable and short.

PR #16819, opened 2023-02-20
Move the `CombinedLoader` to an utility file

carmocca:
## What does this PR do?

The supporters file only contains the public `CombinedLoader`. So rename it to that
It also does not depend on anything inside PL core, so it can become an utility. It could be used with Fabric or externally.


awaelchli (commented):
@carmocca Could we import it in `utilities/__init__.py`?

carmocca:
Good point. Done

awaelchli (approved):

PR #16837, opened 2023-02-22
Check for dataloader_idx presence in the hooks

carmocca:
## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning/issues/4206

This should improve the user experience with multiple dataloaders

cc @borda @carmocca @awaelchli @justusschock


tests/tests_pytorch/models/test_hooks.py
```diff
@@ -762,15 +762,17 @@ def validation_step(self, x2, batch_nb2):
             self.assert_args(x2, batch_nb2)
             return super().validation_step(x2, batch_nb2)
 
-        def test_step(self, x3, batch_nb3, dl_idx3):
+        # we don't support a different name for `dataloader_idx`
```
carmocca, line 765: This is a breaking change. I don't see a simple way to support it

awaelchli: I think it would go too far. I prefer to establish a clear, unambiguous contract here.

awaelchli (approved):

justusschock (approved):

PR #16847, opened 2023-02-23
Trainer: auto default

carmocca:
## What does this PR do?

Fixes https://github.com/Lightning-AI/lightning/issues/10606

Follow-up questions:
- CUDA CI only runs the "min_cuda_gpus=N" tests. This helped greatly reduce the CI load in the past. Since we default to `"auto"` now, should we have CUDA CI run all tests?
- The same question above applies to HPU, IPU, TPU, ...


docs/source-pytorch/accelerators/hpu_basic.rst
```diff
@@ -25,28 +25,6 @@ For more information, check out `Gaudi Architecture <https://docs.habana.ai/en/l
 
 ----
 
-Run on 1 Gaudi
```
carmocca, line 28: I thought these accelerator-specific sections were a bit repetitive. There's no logic here specific to the accelerator, and this is already described in the Trainer section. So I went ahead and removed it.

Any other docs suggestions are welcome

justusschock (approved):

justusschock:
> CUDA CI only runs the "min_cuda_gpus=N" tests. This helped greatly reduce the CI load in the past. Since we default to "auto" now, should we have CUDA CI run all tests?

I wouldn't do this. I think the current way would be preferred.

docs/source-pytorch/accelerators/hpu_basic.rst
awaelchli: I think the intention was more that it is self-contained and that you know that accelerator=hpu exists.

docs/source-pytorch/accelerators/tpu_basic.rst
```diff
@@ -32,26 +32,6 @@ some subset of those 2048 cores.
 
 ----
 
-Run on 1 TPU core
```
awaelchli, line 35: These sections are still relevant. It is still 100% ok to set these arguments, even if they default to auto now.

tests/tests_pytorch/strategies/test_ddp_strategy.py
```diff
@@ -115,7 +115,7 @@ def creates_processes_externally(self):
 
 
 @RunIf(skip_windows=True)
-def test_ddp_configure_ddp(cuda_count_2, mps_count_0):
```
awaelchli, line 118: mps_count_0 needs to stay. ddp is not supported for mps (accelerator=auto)

tests/tests_pytorch/trainer/test_trainer.py
```diff
@@ -1871,13 +1873,13 @@ def training_step(self, batch, batch_idx):
...
     ["trainer_kwargs", "strategy_cls", "strategy_name", "accelerator_cls", "devices"],
     [
-        ({"strategy": None}, SingleDeviceStrategy, "single_device", CPUAccelerator, 1),
+        ({}, SingleDeviceStrategy, "single_device", CPUAccelerator, 1),
         pytest.param({"strategy": "ddp"}, DDPStrategy, "ddp", CPUAccelerator, 1, marks=RunIf(mps=False)),
         pytest.param(
             {"strategy": "ddp", "num_nodes": 2}, DDPStrategy, "ddp", CPUAccelerator, 1, marks=RunIf(mps=False)
         ),
         (
-            {"strategy": None, "accelerator": "cuda", "devices": 1},
```
awaelchli, line 1880: i would keep it explicit here with auto

docs/source-pytorch/accelerators/hpu_basic.rst
carmocca: Added back

awaelchli (approved):

carmocca:
I added one conenctor test that asserts "auto" under each accelerator availability. Just as the one in #16842

PR #16854, opened 2023-02-23
Introduce `Trainer(barebones=True)`

carmocca:
## What does this PR do?

Introduces a flag that disables all the "magic" inside the Trainer that can have a negative speed impact compared to alternatives.


src/lightning/pytorch/trainer/trainer.py
```diff
@@ -286,6 +291,76 @@ def __init__(
         if default_root_dir is not None:
             default_root_dir = os.fspath(default_root_dir)
 
+        self.barebones = barebones
+        if barebones:
```
lantiga, line 295: I would also print a nice message saying something like
```
You are running in barebones mode. All features that may impact raw performance have been disabled to facilitate running benchmarks. Specifically, the following trainer flags have been set:
- enable_checkpointing=False
- ...
```

lantiga (commented):
Looks good, I added a few comments on messaging towards the user

src/lightning/pytorch/trainer/trainer.py
```diff
@@ -278,6 +279,10 @@ def __init__(
 
             inference_mode: Whether to use :func:`torch.inference_mode` or :func:`torch.no_grad` during
                 evaluation (``validate``/``test``/``predict``).
+
+            barebones: Whether to run in "barebones mode", where all features that could impact performance are
```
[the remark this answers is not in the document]
justusschock: can we actually call it "speed" or something similar here? The term performance might be confusing as it could also related to predictive performance (aka "accuracy")

src/lightning/pytorch/trainer/trainer.py
```diff
@@ -286,6 +291,76 @@ def __init__(
...
+                )
+            enable_model_summary = False
+            if num_sanity_val_steps is not None and num_sanity_val_steps != 0:
+                raise ValueError(
+                    f"`Trainer(barebones=True, num_sanity_val_steps={num_sanity_val_steps!r})` was passed."
+                    " Sanity checking can impact performance so it is disabled in barebones mode."
+                )
+            num_sanity_val_steps = 0
+            # opt-ins
+            if fast_dev_run is not False and fast_dev_run != 0:
```
justusschock, line 334: personally, I think `fast_dev_run` is the only flag that should have priority over barebones. If you set this you anyways don't want to train a real workload.

src/lightning/pytorch/trainer/trainer.py
```diff
@@ -286,6 +291,76 @@ def __init__(
...
+        self.barebones = barebones
+        if barebones:
+            # opt-outs
+            if enable_checkpointing:
+                raise ValueError(
+                    f"`Trainer(barebones=True, enable_checkpointing={enable_checkpointing!r})` was passed."
+                    " Checkpointing can impact performance so it is disabled in barebones mode."
+                )
+            enable_checkpointing = False
+            if logger is not None and logger is not False:
```
justusschock, line 303: does this also disable the internal handling of logged values (and their aggregation) or just the flushing to the logger?

justusschock (approved):

src/lightning/pytorch/trainer/trainer.py
carmocca: Yes it does. See https://github.com/Lightning-AI/lightning/pull/16854/files#diff-4fdfbdac03762c6e625f1289e98750edd4c585a471e049a02f41d56922bf8be9R392-R397

src/lightning/pytorch/trainer/trainer.py
carmocca: Can you clarify what do you mean precisely?

Is it that fast_dev_run ignores barebones?
Or to allow them run run together?

src/lightning/pytorch/trainer/trainer.py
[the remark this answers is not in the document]
carmocca: Used "raw speed" but happy to change it

src/lightning/pytorch/trainer/trainer.py
```diff
@@ -285,6 +301,92 @@ def __init__(
...
+                " - Progress bar: `Trainer(enable_progress_bar=True)`",
+                " - Model summary: `Trainer(enable_model_summary=True)`",
+                " - Logging: `Trainer(logger=True)`, `Trainer(log_every_n_steps>0)`,"
+                " `LightningModule.log(...)`, `LightningModule.log_dict(...)`",
+                " - Sanity checking: `Trainer(num_sanity_val_steps>0)`",
+                " - Development run: `Trainer(fast_dev_run=True)`",
+                " - Anomaly detection: `Trainer(detect_anomaly=True)`",
+                " - Profiling: `Trainer(profiler=...)`",
+            )
+            rank_zero_info(
```
carmocca, line 370: Looks like

```python
You are running in `Trainer(barebones=True)` mode. All features that may impact raw speed have been disabled to facilitate analyzing the Trainer overhead. Specifically, the following features are deactivated:
 - Checkpointing: `Trainer(enable_checkpointing=True)`
 - Progress bar: `Trainer(enable_progress_bar=True)`
 - Model summary: `Trainer(enable_model_summary=True)`
 - Logging: `Trainer(logger=True)`, `Trainer(log_every_n_steps>0)`, `LightningModule.log(...)`, `LightningModule.log_dict(...)`
 - Sanity checking: `Trainer(num_sanity_val_steps>0)`
 - Development run: `Trainer(fast_dev_run=True)`
 - Anomaly detection: `Trainer(detect_anomaly=True)`
 - Profiling: `Trainer(profiler=...)`
```

src/lightning/pytorch/trainer/trainer.py
justusschock: fast_dev_run ignores barebones

src/lightning/pytorch/trainer/trainer.py
```diff
@@ -112,13 +112,13 @@ def __init__(
         limit_test_batches: Optional[Union[int, float]] = None,
         limit_predict_batches: Optional[Union[int, float]] = None,
         val_check_interval: Optional[Union[int, float]] = None,
-        log_every_n_steps: int = 50,
+        log_every_n_steps: Optional[int] = None,
```
awaelchli, line 115: Why can't we leave these defaults? This makes everything less readable, a step back again from the goals of 2.0 in my opinion. I don't think we need to do a special validation below for every flag. IMO, if `barebones=True`, we show the warning/info message that all other settings get bypassed, regardless of their setting. The same principle applied to fast_dev_run previously.

carmocca: To avoid users mistakenly thinking that `Trainer(logger=TensorBoardLogger(), barebones=True)` works.

I agree it's less readable, but the current implementation makes it more fool-proof.

If we did a `BarebonesTrainer`, it wouldn't expose these arguments, and the Trainer argument defaults would stay as they are. But that's a different solution

src/lightning/pytorch/trainer/trainer.py
carmocca: What would be the benefit of that? I think it could make things more error prone if `Trainer(fast_dev_run=True, barebones=True)` silently ignores barebones

src/lightning/pytorch/trainer/trainer.py
awaelchli: Barebones is False by default, so a user has to activate it manually for it to run. So they learn what it does by reading the docstring and then through the info message that shows up. So it's not like someone would blindly enable a feature they don't know what it does, and then be surprised that tensorboard shows no logs.

carmocca: I disagree. Several times people have left trainer flags enabled and then complained about issues related to them, without noticing that the flag is active.

Maybe it was an accident, maybe it's a new user who copy-pasted an example, maybe an issue in your CLI configuration, ... or maybe they just don't really understand exactly what it does!

src/lightning/pytorch/trainer/trainer.py
```diff
@@ -277,6 +278,21 @@ def __init__(
 
             inference_mode: Whether to use :func:`torch.inference_mode` or :func:`torch.no_grad` during
                 evaluation (``validate``/``test``/``predict``).
+
+            barebones: Whether to run in "barebones mode", where all features that may impact raw speed are
+                disabled. This is meant for analyzing the Trainer overhead and is discouraged during regular trainer
```
lantiga, line 283:
```suggestion
                disabled. This is meant for analyzing the Trainer overhead and is discouraged during regular training
```

src/lightning/pytorch/trainer/trainer.py
lantiga: I agree with @carmocca here. `barebones=True` should basically take over the Trainer, combinations should be disallowed to reinforce the fact that it needs to be used intentionally and disabled in all other cases.

src/lightning/pytorch/trainer/trainer.py
lantiga: Looks great

src/lightning/pytorch/trainer/trainer.py
lantiga: Yeah not ideal but I agree it's the safest. If we could inspect what gets passed into the constructor that would be great but there doesn't seem to be a good way.

lantiga (approved):

awaelchli (approved):
unblock :)

PR #16867, opened 2023-02-24
move changelogs to standalone folders

Borda:
## What does this PR do?

pesonaly I would rather see the changelog outside the package/python tree

<details>

</details>

## PR review

<details>

</details>


justusschock (changes requested):
I am very much against this change.
If you have a local installation of lightning, the default folders do exist and are on python path if you installed in editable mode.

Things like https://github.com/Lightning-AI/lightning/blob/master/src/lightning/app/components/multi_node/trainer.py#L51 where we actually need to import from both packages when present to make sure it also works with standalone PL (or Fabric), require me to delete the standalone folders to not get ImportErrors just because of the folder existence.

If we move the changelog there, I cannot do this anymore while still editing the changelog locally.

I would appreciate a  lot if we could actually get rid of standalone folders

awaelchli:
I'm also for getting rid of the standalone folders entirely. We can already see that contributors get confused, and make source code modifications in the [wrong files](https://github.com/Lightning-AI/lightning/pull/16866/commits/38283a59fb890f19e133dc725b8a21c97387bdcf) (lightning_fabric instead of lightning/fabric).

awaelchli:
Since the standalone packages get generated anyway, one could just move/generate these files and if you want, also remove the changelogs from the subfolder. Just an idea.

Borda:
> We can already see that contributors get confused, and make source code modifications in the [wrong files](https://github.com/Lightning-AI/lightning/pull/16866/commits/38283a59fb890f19e133dc725b8a21c97387bdcf) (lightning_fabric instead of lightning/fabric).

This won't change with del dir as you can see, it was one line cnage, so user created PR from state before refactoring

> Things like [`master`/src/lightning/app/components/multi_node/trainer.py#L51](https://github.com/Lightning-AI/lightning/blob/master/src/lightning/app/components/multi_node/trainer.py?rgh-link-date=2023-02-24T14%3A45%3A27Z#L51) where we actually need to import from both packages when present to make sure it also works with standalone PL (or Fabric), require me to delete the standalone folders to not get ImportErrors just because of the folder existence.

Not sure how delete folder helps here as this line need to stay uncless we abendent stanlone packages (not only folder for it)

PR #16891, opened 2023-02-28
PyTorch 2.0 requires ONNX to be available to use it

carmocca:
## What does this PR do?

See title. Raise an exception if ONNX is used but it's not available.

cc @justusschock @awaelchli


src/lightning/pytorch/core/module.py
```diff
@@ -1299,22 +1302,23 @@ def to_onnx(self, file_path: Union[str, Path], input_sample: Optional[Any] = Non
             **kwargs: Will be passed to torch.onnx.export function.
 
         Example:
-            >>> class SimpleModel(LightningModule):
```
carmocca, line 1302: Removed as doctest because it's not easy to skip based on a condition (onnx available)

awaelchli: You could convert it to a regular code example, and/or move it to the .rst doc

awaelchli (approved):

src/lightning/pytorch/core/module.py
carmocca: Isn't the current change already a conversion to a regular code example?

carmocca:
@Borda I merged this into the other PR and it's passing. Can we move forward with this now?

carmocca:
Included in https://github.com/Lightning-AI/lightning/pull/16764

PR #16898, opened 2023-02-28
Revert "import neptune instead of import neptune.new"

wouterzwerink:
Reverts Lightning-AI/lightning#16888

See https://github.com/Lightning-AI/lightning/pull/16888#issuecomment-1448032165 for further information


awaelchli (approved):

carmocca (changes requested):
Let's see if https://github.com/Lightning-AI/lightning/pull/16761 lands before the 2.0 release. Otherwise we'll merge this to avoid introducing the breaking change. Requesting changes to block in the meantime

carmocca:
Looks like it's not happening. Unblocking

carmocca (approved):

